PR #1, opened 2019-10-18
docs(README): campatibility -> compatibility

samouss:


bufdev:
Thanks for this - I think it's time to find a Vim plugin for spellcheck.

The README.md is automatically created from an internal repository and sync'ed here (and we haven't gotten around to making it bidirectional yet), so I'll manually sync this back for now, but if you find any more, let us know at dev@buf.build for now if that's OK :-)

PR #13, opened 2020-01-04
Add git, openssh-client, tzdata to docker images

kevinconway:
The missing git binary results in failures when running the `check
breaking` feature from within a container because it relies on git to
clone branches/versions. This patch enables use of `check breaking` from
the public container.

Adding openssh-client enables cloning over SSH and enables Linux and WSL
users to forward/mount the local SSH agent. This enables cloning private
repositories over SSH through the docker image.

The tzdata dependency is required for Go binaries that manage time.
There doesn't appear to be any current use case where timezone data are
manipulated but this will prevent broken builds and panics should a
timezone operation make its way into buf.

### DIFF
--- a/Dockerfile.buf
+++ b/Dockerfile.buf
@@ -9,7 +9,7 @@ RUN go build -o /go/bin/buf ./cmd/buf
 FROM alpine:3.10
 
 RUN apk add --update --no-cache \
-    ca-certificates && \
+    ca-certificates tzdata git openssh-client && \
   rm -rf /var/cache/apk/*
 
 COPY --from=builder /go/bin/buf /usr/local/bin/buf
--- a/Dockerfile.workspace
+++ b/Dockerfile.workspace
@@ -17,6 +17,8 @@ RUN apk add --update --no-cache \
     bash \
     build-base \
     ca-certificates \
+    tzdata \
+    openssh-client \
     curl \
     git \
     unzip \


bufdev:
The git binary isn’t required for clones - buf uses https://godoc.org/gopkg.in/src-d/go-git.v4, so if the git binary is needed, this is a bug. We will test this internally but can you provide a reproduction of a case that fails with git with the Docker container?

Buf is meant to be self-contained, so besides certs, I don’t think these should be necessary, but if there’s a reproduction of your issue we will certainly check it out!

kevinconway:
Reproduction steps:

```bash
$ docker run \
--volume="/home/user/code/myproject/buf-repro:/work" \
--workdir="/work" \
bufbuild/buf:latest \
check breaking --against-input '.git#branch=master'
could not clone .git: exec: "git": executable file not found in $PATH
```
I've attached a tarball of a basic repository that reproduces this issue. The included `Makefile` has both a `lint` and `test` targets. The `lint` target runs `check lint` to demonstrate that the repository is valid and the tool works from the container. The `test` target runs `check breaking` to demonstrate the issue.

[repro-example.tar.gz](https://github.com/bufbuild/buf/files/4022603/repro-example.tar.gz)

kevinconway:
> Buf is meant to be self-contained, so besides certs, I don’t think these should be necessary

`tzdata` is still a system dependency for which the Go runtime has no internal equivalent. So long as there are no timezone operations then it doesn't matter if it is missing. I don't have a current requirement for it. As for `openssh-client`, I'm also happy to maintain my own per-project extensions of the public image where this is needed rather than bloat the public image.

It seems like this PR is not needed and can be closed. Would you prefer I open an issue to track the `git` binary bug?

bufdev:
Happy to keep the PR open - this is clearly an issue (and we sincerely appreciate you filing this) - if you can file an issue as well, we will look at it, the above is a bug and not what is intended on our end, so we will make sure this issue is solved.

bufdev:
Just an update for over here:

- We may have to add `git` until https://github.com/src-d/go-git/issues/847 is fixed.
- Adding `openssh-client` is likely a step too far - we provide the ability to use public key authentication only, and don't intend to provide full ssh capabilities here.
- I don't see a need for `tzdata`, we're not going to be manipulating timezone data.

The goal is for the Docker image to be as small as possible, and optimally a single binary only. There is the need to have the certs for HTTPS requests, but beyond that the goal is for buf to be self-contained. The above adds almost double the image size from ~27MB to ~50MB, and make it so we have more external deps to tell users to intall.

kevinconway:
That all sounds pretty reasonable to me. I'll close this PR and defer when/if to add `git` to you. I appreciate you tracking down the source of the issue.

PR #45, opened 2020-03-27
Allow all input types to image build

### LINKED ISSUE #44: Add functionality to convert Image formats
I am dealing with a set of proto files from a couple legacy repos that I cannot change. I can successfully build an image from these protos, but I cannot generate stubs from them with protoc. To generate stubs correctly I would have to write some nasty commands to automatically fixup the proto files before building the image. However, if I could build a JSON image, modify that really easily with jq, and then convert that to a binary image with buf to use as input to protoc, I wouldn't have to resort to hack-ily sed-ing the proto files. Is this possible?

P.S. Converting binary images to JSON would also be useful for introspection, but that is not necessary for me right now.

bufdev:


bufdev:
Closing for now per discussion on #44

PR #46, opened 2020-03-27
Add recurse_submodules input option

### LINKED ISSUE #42: Add support for submodules in check breaking
In my repo where I've vendored in e.g. googleapis as a submodule, the command for comparing breaking against a git repo doesn't work. I suggest adding a new flag to the format ref for git that, when activated, will update init submodules recursively.

## Expected

I expect the following to work for a repo with submodules:

```
buf check breaking --against-input '.git#branch=master'
```

## Actual

It did not work, since the submodules will be empty after cloning, and therefore the breaking check will complain about missing imports (e.g. google/api/annotations.proto).

## Suggestion

Add a submodule flag, e.g.:

```
 buf check breaking --against-input '.git#branch=master#submodules=true'
```

When the submodules flag is active, the breaking check will initialize submodules recursively in the (temporary) clone.

bufdev:
Fixes #42.

Example:

```
$ buf check lint --input https://github.com/foo/bar.git#branch=master,recurse_submodules=true
```

### DIFF
--- a/internal/buf/bufos/env_reader.go
+++ b/internal/buf/bufos/env_reader.go
@@ -518,6 +518,7 @@ func (e *envReader) getReadBucketCloser(
 			getenv,
 			inputRef.Path,
 			inputRef.GitRefName,
+			inputRef.GitRecurseSubmodules,
 		)
 	default:
 		return nil, fmt.Errorf("unknown format outside of parse: %v", inputRef.Format)
@@ -597,6 +598,7 @@ func (e *envReader) getReadBucketCloserFromGitRepo(
 	getenv func(string) string,
 	gitRepo string,
 	gitRefName storagegitplumbing.RefName,
+	gitRecurseSubmodules bool,
 ) (_ storage.ReadBucketCloser, retErr error) {
 	defer utillog.Defer(e.logger, "get_git_bucket_memory")()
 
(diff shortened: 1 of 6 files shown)


fsommar:
Thanks a lot! This unfortunately times out on my repo. I suggest lowering the default recursiveness to 1, and/or make it configurable.

bufdev:
This wouldn’t match with git. If it’s timing out, you probably have a bunch of recursive repos, increase the —timeout flag.

fsommar:
Ok, do you know if go-git reads .gitmodules to make my submodule clone shallow? Either way, it worked with increasing the timeout (in total `buf --timeout 120s check breaking --against-input '.git#branch=master,recurse_submodules=true'` takes roughly a minute). If I do a regular `git clone --recurse-submodules` it takes ca 30 seconds. So 60 seconds is within an acceptable time frame for sure.

Again, thanks a lot for implementing this!

bufdev:
May add `recurse_submodules=shallow`, as there is a corresponding git binary flag for this.

Also filed https://github.com/go-git/go-git/issues/15 to see if we can get cloning at depth 1.

bufdev:
FYI, waiting on https://github.com/go-git/go-git/issues/15 to see if we can improve this before merging. Will keep this branch here for now.

bufdev:
Going to merge this, however @fsommar note that we may change the name of the option and/or add functionality before the next release, and before v1.0.

PR #53, opened 2020-05-08
Move to golang/protobuf v2

bufdev:


bufdev:
Currently pinned to https://github.com/jhump/protoreflect/pull/319, will merge when that is merged to master (no rush Josh, all good :-) )

Issue #105, opened 2020-07-12
Zsh/Fish completion failing


amenzhinsky:
Hi there,

zsh completion is not working, I'm getting `_arguments:comparguments:325: invalid option definition: --log-format[The log format [text,color,json].]:`

Generated completion looks like:

```
function _buf {
  local -a commands

  _arguments -C \
    '--log-format[The log format [text,color,json].]:' \
    '--log-level[The log level [debug,info,warn,error].]:' \
    '--timeout[The duration until timing out.]:' \
    "1: :->cmnds" \
    "*::arg:->args"
```

It seems cobra doesn't escape flag description properly: https://github.com/spf13/cobra/blob/v1.0.0/zsh_completions.go#L334 and braces from flag descriptions with permitted values break it.

I'm not sure whether it's done on purpose for some flexibility or not.

```
% zsh --version
zsh 5.8 (x86_64-pc-linux-gnu)
```

bufdev:
See https://github.com/spf13/cobra/issues/1129

bufdev:
See https://github.com/spf13/cobra/issues/1129#issuecomment-657811554 - we're going to update to cobra master for now.

amenzhinsky:
@bufdev the error is fixed, but completion is gone now btw :)
Didn't have much time to figure out  why, but generated code does nothing now, it doesn't contain nether command names nor flags.

```
#compdef _buf buf

# zsh completion for buf                                  -*- shell-script -*-

__buf_debug()
{
    local file="$BASH_COMP_DEBUG_FILE"
    if [[ -n ${file} ]]; then
        echo "$*" >> "${file}"
    fi
}

_buf()
{
    local shellCompDirectiveError=1
    local shellCompDirectiveNoSpace=2
    local shellCompDirectiveNoFileComp=4
    local shellCompDirectiveFilterFileExt=8
    local shellCompDirectiveFilterDirs=16

    local lastParam lastChar flagPrefix requestComp out directive compCount comp lastComp
    local -a completions

    __buf_debug "\n========= starting completion logic =========="
    __buf_debug "CURRENT: ${CURRENT}, words[*]: ${words[*]}"

    # The user could have moved the cursor backwards on the command-line.
    # We need to trigger completion from the $CURRENT location, so we need
    # to truncate the command-line ($words) up to the $CURRENT location.
    # (We cannot use $CURSOR as its value does not work when a command is an alias.)
    words=("${=words[1,CURRENT]}")
[...]

PR #106, opened 2020-07-12
Fix zsh autocompletion

amenzhinsky:
Fixes: #105 

Probably not the nicest fix, but makes zsh autocompletion work.

Unfortunately there's no single styleguide for cli apps to follow, but docker, systemctl and kubectl declare available option values in parentheses.

### DIFF
--- a/internal/pkg/app/appflag/builder.go
+++ b/internal/pkg/app/appflag/builder.go
@@ -51,8 +51,8 @@ func newBuilder(options ...BuilderOption) *builder {
 }
 
 func (b *builder) BindRoot(flagSet *pflag.FlagSet) {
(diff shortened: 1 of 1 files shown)


bufdev:
Can you file an issue to fix cobra instead? There's actually a bunch more places this needs to be changed, and is a bit more involved, as most of these are auto-created (we'd have to change from `[` to `(` in all of them to be consistent). We can do it depending on Cobra's response, but I would prefer to print out `[`.

bufdev:
This should be fixed in Cobra instead, or if the fix is already present and we are misusing Cobra, we should fix that. See https://github.com/spf13/cobra/issues/1129

PR #116, opened 2020-07-18
Add missing `--` to protoc proxy handler opt flag

odsod:
Fixes issue where the opt flag is interpreted as a file to generate,
causing the compilation to fail.


bufdev:
Whoops! We need add a test for that, not sure how we didn’t catch that.

PR #128, opened 2020-08-16
Change successful command output target to container.Stdout

devholic:
I'm not sure this was intended or not, successful command output is piped to `stderr`, not `stdout`.

I think piping successful command output to `stdout` is more reasonable, so I'd like to suggest changing successful command output target to `container.Stdout`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/bufbuild/buf
 
-go 1.14
+go 1.15
 
 require (
 	github.com/bgentry/go-netrc v0.0.0-20140422174119-9fd32a8b3d3d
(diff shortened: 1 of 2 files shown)


bufdev:
This isn't actually where this is controlled - this is done in each command via `Containers`.

We want this to stay as-is, and existing output is sent to stdout via other mechanisms.

devholic:
Thanks for let me know 🙂

> We want this to stay as-is, and existing output is sent to stdout via other mechanisms.

But in cobra-managed commands, for example:
```sh
buf -v 
```
displays version information in `stderr`, not `stdout` since this command is exited with code 0, so I was confused about this 🤔

bufdev:
We've moved `--version` to stdout for v1, as an FYI!

PR #129, opened 2020-08-20
To allow pre-built binaries use, switch Docker builds to Debian-based

dvdotsenko:
Just dropping `protoc` binary into Alpine-based `bufbuild/buf` docker image did NOT work. 
One of libc dependencies were missing (`ldd /path/to/protoc`) and attempts to run `protoc` resulted in ` 'protoc' not found  ` errors.

This PR:
- adds `protoc` from latest release AND
- switches base image from Alpine to Debian where libc and friends are friendly to having vendor-compiled binaries to be just dropped in.

Same problem is anticipated for multitude of other vendor-pre-compiled binaries. Recompiling all of them just to package them on top of Alpine is unjustified pain.

What's most annoying with current Alpine-based image is that I cannot just use it as base image, just add `protoc` plugins / tools and move on. I literally have to rebuild entire `buf` image from different base.

While rebasing on top of Debian slim adds a few megabytes to the image, it saves so much labor costs and trade off is very much worth it, considering Debian Buster Slim already sits in cache on every Docker host at this point.

Obviously don't need to merge my PR, but contemplate switching like that to a sensible docker base which we can just extend without having to rebuild from scratch

### DIFF
--- a/Dockerfile.buf
+++ b/Dockerfile.buf
@@ -1,19 +1,36 @@
-ARG DOCKER_WORKSPACE_IMAGE
+# syntax=docker/dockerfile:1.0.0-experimental
 
+ARG DOCKER_WORKSPACE_IMAGE
 FROM $DOCKER_WORKSPACE_IMAGE as builder
 
 COPY cmd ./cmd
 COPY internal ./internal
 RUN go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf
 
-FROM alpine:3.12.0
+ENV PROTOC_VERSION=3.13.0
+RUN wget \
+        https://github.com/protocolbuffers/protobuf/releases/download/v${PROTOC_VERSION}/protoc-${PROTOC_VERSION}-linux-x86_64.zip \
+        -O /tmp/protoc.zip \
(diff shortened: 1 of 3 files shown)


bufdev:
The docker image `bufbuild/buf` is not intended to be a base for additional tools to be added, it is intended to be the minimal docker image needed to run the `buf` binary and the required binaries (i.e. `git`, which `buf` shells out to). Moving to debian will significantly increase the image size, which goes against the goals of this image.

For what it's worth, you can actually fix the `protoc` issue via:

```
RUN wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://alpine-pkgs.sgerrand.com/sgerrand.rsa.pub && \
  wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.31-r0/glibc-2.31-r0.apk && \
  apk add --no-cache glibc-2.31-r0.apk && \
  rm -rf /var/cache/apk/*
```

See https://github.com/bufbuild/buf/blob/master/Dockerfile.workspace

For your use case, as you suggest, I'd recommend starting from your own image and building up. Apologies for the inconvenience.

PR #144, opened 2020-09-20
make all test protos valid per protoc

jhump:
Before I can upgrade protoreflect to directly use API v2 and the descriptor interfaces in "google.golang.org/protobuf/reflect/protoreflect", we need these changes in the test protos.

Currently, protoparse fails to perform some validations on the source, which is why the tests currently work. But the API v2 implementations of descriptors _do_ those validations, so the inputs must be valid.

In particular, these are the protoparse issues that allow these test files to compile even though they aren't actually valid:
* https://github.com/jhump/protoreflect/issues/357
* https://github.com/jhump/protoreflect/issues/358

My fix was to remove `option allow_alias` where it was not needed for the first. For the second, I changed the file to proto2, so the uniqueness check is not enforced.


bufdev:
This was done in #145 - closing this one and deleting this branch

PR #145, opened 2020-09-21
Sync from internal

bufdev:
Also includes #144.


bufdev:
I have no idea what is up with Codecov - I think they did an upgrade that changed ignores on the backend. I've checked this locally, and there is no difference (and our local % is considerably higher). Not letting this block us, but we'll want to investigate Codecov further.

Issue #203, opened 2020-11-25
C# output overwrites generated output by proto file name


shaunco:
C# output isn't scoped by package names or directory structure, so if two proto files in different packages have the same file name in the same project, the generated C# output overwrites any previous output:
* `test/first/sample.proto` - with `package test.first` and any set of messages
* `test/second/sample.proto` - with `package test.second` and any set of messages
* `buf.gen.yaml` - set to generate C#

Run `buf generate -o gen` and only a single `gen/csharp/Sample.cs` results. Using a file system monitor, it looks like both first and second were written, but second overwrites first with no warning or error from buf or protoc. `buf check lint` also didn't produce any warnings.

### Sample Project
[buf-csharp-test.zip](https://github.com/bufbuild/buf/files/5595133/buf-csharp-test.zip)

bufdev:
Please do not file duplicate issues with the exact same text but

shaunco:
They are not duplicates. Javascript overwrites by message name and C# by proto filename, which implies totally different places in the internal compiler.

shaunco:
And since you asked in #202, here is the breakdown for #203...

Not sure how this is better than the zip I attached, but here it is:

**buf.yaml**
```yaml
version: v1beta1
build:
  roots:
    - .
lint:
  use:
    - BASIC
    - FILE_LOWER_SNAKE_CASE
  except:
    - ENUM_NO_ALLOW_ALIAS
    - IMPORT_NO_PUBLIC
    - PACKAGE_AFFINITY
    - PACKAGE_DIRECTORY_MATCH
    - PACKAGE_SAME_DIRECTORY
breaking:
  use:
    - WIRE_JSON
```

**buf.gen.yaml**
```yaml
version: v1beta1
plugins:
  - name: csharp
    out: csharp
```

**test/first/sample.proto**
```proto
syntax = "proto3";

package test.first;

option java_multiple_files = true;
option java_package = "org.test.first";
option csharp_namespace = "Test.First";
option go_package = "test.com/first;first";

message Something1 {
    string test_first = 1;
}
```

**test/second/sample.proto**
```proto
syntax = "proto3";

package test.second;

option java_multiple_files = true;
option java_package = "org.test.second";
option csharp_namespace = "Test.Second";
option go_package = "test.com/second;second";

message Something2 {
    string test_second = 1;
}
```

bufdev:
[...]

Issue #202, opened 2020-11-25
js output overwrites generated output by message name


shaunco:
js output isn't scoped by package names or directory structure, so if two proto files in different packages have a message with the same name in the same project, the generated js output overwrites any previous output:
* `test/first/sample1.proto` - with `package test.first` and a message named `something`
* `test/second/sample2.proto` - with `package test.second` and a message named `something`
* `buf.gen.yaml` - set to generate js

Run `buf generate -o gen` and only a single `gen/js/something.js` results. Using a file system monitor, it looks like both first and second were written, but second overwrites first with no warning or error from buf or protoc. `buf check lint` also didn't produce any warnings.

### Sample Project
[buf-js-test.zip](https://github.com/bufbuild/buf/files/5595107/buf-js-test.zip)

bufdev:
Can you please post your code sample copy/pasted in the description? We’re not going to download a zip file.

shaunco:
Not sure how this is better than the zip I attached, but here it is:

**buf.yaml**
```yaml
version: v1beta1
build:
  roots:
    - .
lint:
  use:
    - BASIC
    - FILE_LOWER_SNAKE_CASE
  except:
    - ENUM_NO_ALLOW_ALIAS
    - IMPORT_NO_PUBLIC
    - PACKAGE_AFFINITY
    - PACKAGE_DIRECTORY_MATCH
    - PACKAGE_SAME_DIRECTORY
breaking:
  use:
    - WIRE_JSON
```

**buf.gen.yaml**
```yaml
version: v1beta1
plugins:
  - name: js
    out: js
```

**test/first/sample1.proto**
```proto
syntax = "proto3";

package test.first;

option java_multiple_files = true;
option java_package = "org.test.first";
option csharp_namespace = "Test.First";
option go_package = "test.com/first;first";

message Something {
    string test_first = 1;
}
```

**test/second/sample2.proto**
```proto
syntax = "proto3";

package test.second;

option java_multiple_files = true;
option java_package = "org.test.second";
option csharp_namespace = "Test.Second";
option go_package = "test.com/second;second";

message Something {
    string test_second = 1;
}
```

bufdev:
This isn't actually a bug, this is working as expected *given* the behavior of the `js` builtin plugin.  `buf generate` makes a parallel generation call per directory, roughly the equivalent of:

```bash
for dir in $(find "${ROOT}" -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq); do
[...]

PR #207, opened 2020-12-02
Adding serial option to bufgen config that disables parallel processing per plugin

### LINKED ISSUE #206: Nondeterministic output when using python betterproto with buf generate
I'm not sure if the issue lies with buf or [python betterproto](https://github.com/danielgtaylor/python-betterproto) but I'm posting the issue here since I didn't run into this non determinism when using protoc.

buf version: 0.32.0 (also tried 0.31.1)
protoc version: 3.11.3
python betterproto version: 1.2.3

I am running buf generate using the command `buf generate --template gen_config.json --config build_config.json`.
`gen_config.json` is:
```
{
    "version": "v1beta1",
    "plugins": [
        {
            "name": "python_betterproto",
            "out": OUT_DIR,
        },
    ],
}
```

`build_config.json` is:
```
{
    "version": "v1beta1",
    "build":  {
          "roots": ROOT_DIRS
    }
}
```

The issue is that the python output for a particular proto is non deterministic. The resulting python code changes slightly across multiple runs. Strangely, this doesn't occur when calling betterproto using protoc like so:
```
zcat image.bin.gz | protoc \
  --descriptor_set_in=/dev/stdin \
  --python_betterproto_out="$OUT_DIR" \
  $(buf ls-files --input image.bin.gz)
```

Unfortunately I cannot post the proto in question because it's proprietary. Thank you for taking a look!

shaunco:
Fixes #206, #202, and #203

### DIFF
--- a/internal/buf/bufcore/bufimage/bufimage.go
+++ b/internal/buf/bufcore/bufimage/bufimage.go
@@ -281,10 +281,39 @@ func ImageToCodeGeneratorRequest(image Image, parameter string) *pluginpb.CodeGe
 // ImagesToCodeGeneratorRequests converts the Images to CodeGeneratorRequests.
 //
 // All non-imports are added as files to generate.
-func ImagesToCodeGeneratorRequests(images []Image, parameter string) []*pluginpb.CodeGeneratorRequest {
+func ImagesToCodeGeneratorRequests(images []Image, parameter string, serial bool) []*pluginpb.CodeGeneratorRequest {
 	requests := make([]*pluginpb.CodeGeneratorRequest, len(images))
 	for i, image := range images {
 		requests[i] = ImageToCodeGeneratorRequest(image, parameter)
 	}
+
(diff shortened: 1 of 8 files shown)


bufdev:
Thanks for the effort, but we haven't decided if we want to do this, this isn't how we'd actually implement it in the code either. We'll evaluate this feature when we have time and implement separately.

bufdev:
Closing per https://github.com/bufbuild/buf/blob/master/.github/pull_request_template.md - we're also still evaluating whether we want to do this, there are downstream implications of enabling this feature that we're not sure we want to expose ourselves to, and in the case of #206, this is a bug in the plugin that should be fixed.

bufdev:
Filed https://github.com/bufbuild/buf/issues/208 to track this at a high level. We do appreciate the effort, we just can't commit to this for now.

bufdev:
This is released in v0.33.0. See https://docs.buf.build/generate-usage for details. Let us know if this works!

Issue #288, opened 2021-03-16
Support mulit-architecture Docker image


Juneezee:
Currently [bufbuild/buf](https://hub.docker.com/r/bufbuild/buf/tags) only supports `linux/amd64`. Consider supporting other architectures such as `arm64`, `arm/v7`, `arm/v6` etc.

![image](https://user-images.githubusercontent.com/20135478/111304385-a107fd00-8690-11eb-9416-c6294277d985.png)

This can be done by integrating [docker/setup-qemu-action](https://github.com/docker/setup-qemu-action) and [docker/setup-buildx-action](https://github.com/docker/setup-buildx-action) into the current GitHub Actions workflow. `Dockerfile.buf` will need to be changed to something like the following to support multi-architecture.

```
FROM --platform=${BUILDPLATFORM} golang:1.16.2-alpine3.12 as builder

WORKDIR /workspace

COPY go.mod go.sum /workspace/
RUN go mod download

COPY cmd /workspace/cmd
COPY internal /workspace/internal

ARG TARGETOS
ARG TARGETARCH
RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf

FROM --platform=${BUILDPLATFORM} alpine:3.12.4

RUN apk add --update --no-cache \
    ca-certificates \
    git \
    openssh-client && \
  rm -rf /var/cache/apk/*

COPY --from=builder /go/bin/buf /usr/local/bin/buf

ENTRYPOINT ["/usr/local/bin/buf"]
```

See https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope for Dockerfile reference regarding `BUILDPLATFORM`, `TARGETOS`, `TARGETARCH`

bufdev:
This is definitely something we want to do soon. We will look into it.

PR #289, opened 2021-03-17
Support multi-architecture Docker image

Juneezee:
As per #288, I have made this PR to make GitHub Actions supports building multi-architecture Docker image by using [docker/setup-qemu-action](https://github.com/docker/setup-qemu-action), [docker/setup-buildx-action](https://github.com/docker/setup-buildx-action), and [docker/build-push-action](https://github.com/docker/build-push-action).

I have tested the GitHub Actions workflow in my fork. The `docker` job depends on `build` job, and `docker` job will be run if and only if it is on `master` branch or a tag is pushed.

EDIT: This PR will make `.github/scripts/docker.bash` redundant, the maintainers can consider to remove it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -32,13 +32,46 @@ jobs:
           BUF_BREAKING_PROTO_INPUT: 'https://github.com/bufbuild/buf.git#branch=master,subdir=proto'
           BUF_INPUT_HTTPS_USERNAME: ${{ github.actor }}
           BUF_INPUT_HTTPS_PASSWORD: ${{ github.token }}
-      - name: docker
-        if: success()
-        run: bash .github/scripts/docker.bash
-        env:
-          DOCKER_BUILD_MAKE_TARGET: dockerbuildbuf
-          DOCKER_IMAGE: bufbuild/buf
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
-          DOCKER_LATEST_BRANCH: master
-          DOCKER_VERSION_TAG_PREFIX: v
+  docker:
+    runs-on: ubuntu-20.04
+    needs: build
+    # This job only runs when
+    # 1. When the previous `build` job has completed successfully
+    # 2. When the repository is not a fork, i.e. it will only run on the official bufbuild/buf
+    # 3. When the workflow run is trigged by master branch OR a tag with v prefix
+    # See https://github.com/bufbuild/buf/pull/289/files#r596207623 for the discussion
+    if:  ${{ success() && github.repository == 'bufbuild/buf' && (github.ref == 'refs/heads/master' || contains(github.ref, 'refs/tags/v')) }}
+    steps:
+      - name: setup-qemu
+        uses: docker/setup-qemu-action@v1
+        id: qemu
+        with:
+          # alpine image doesn't support linux/riscv64
+          platforms: linux/386,linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6,linux/ppc64le,linux/s390x
+      - name: setup-docker-buildx
+        uses: docker/setup-buildx-action@v1
+      - name: login-docker
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      - name: set-tag
+        # See the following links for setting job outputs
+        # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs
+        # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter
+        id: tag
+        run: |
+          if echo "${GITHUB_REF}" | grep ^refs/heads/master$ >/dev/null; then
+            echo ::set-output name=tag::latest
+          elif echo "${GITHUB_REF}" | grep ^refs/tags/v >/dev/null; then
+            echo ::set-output name=tag::"$(echo "${GITHUB_REF}" | sed "s/refs\/tags\/v//")"
+          fi
+      - name: docker-build-push
+        uses: docker/build-push-action@v2
+        with:
+          file: Dockerfile.buf
+          platforms: ${{ steps.qemu.outputs.platforms }}
+          push: true
+          tags: |
+            bufbuild/buf:latest
+            bufbuild/buf:${{ steps.tag.outputs.tag }}
--- a/Dockerfile.buf
+++ b/Dockerfile.buf
@@ -1,4 +1,4 @@
-FROM golang:1.16.2-alpine3.12 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.16.2-alpine3.12 as builder
 
 WORKDIR /workspace
 
@@ -7,9 +7,13 @@ RUN go mod download
 
 COPY cmd /workspace/cmd
 COPY internal /workspace/internal
-RUN CGO_ENABLED=0 go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf
 
-FROM alpine:3.12.4
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+  go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf
+
+FROM --platform=${TARGETPLATFORM} alpine:3.12.4
 
 RUN apk add --update --no-cache \
     ca-certificates \


Dockerfile.buf
```diff
@@ -7,14 +7,18 @@ RUN go mod download
...
 COPY internal /workspace/internal
-RUN CGO_ENABLED=0 go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf
 
-FROM alpine:3.12.4
+ARG TARGETOS
+ARG TARGETARCH
+RUN CGO_ENABLED=0 GOOS=${TARGETOS} GOARCH=${TARGETARCH} \
+  go build -ldflags "-s -w" -trimpath -o /go/bin/buf ./cmd/buf
+
+FROM --platform=${BUILDPLATFORM} alpine:3.12.4
```
ZymoticB, line 16: If I'm understanding how the buildx command works correctly, this final stage should be from the target platform

```suggestion
FROM --platform=${TARGETPLATFORM} alpine:3.12.4
```

.github/workflows/ci.yaml
```diff
@@ -8,11 +8,13 @@ jobs:
     steps:
       - name: checkout
         uses: actions/checkout@v2
+
```
bufdev, line 11: no newline

.github/workflows/ci.yaml
```diff
@@ -8,11 +8,13 @@ jobs:
...
       - name: setup-go
         if: success()
         uses: actions/setup-go@v2
         with:
           go-version: 1.16.2
+
```
bufdev, line 17: no newline

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
           key: ${{ runner.os }}-buf-${{ hashFiles('**/go.sum', '/make/**') }}
           restore-keys: |
             ${{ runner.os }}-buf-
+
```
bufdev, line 30: no newline

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
-        if: success()
-        run: bash .github/scripts/docker.bash
-        env:
-          DOCKER_BUILD_MAKE_TARGET: dockerbuildbuf
-          DOCKER_IMAGE: bufbuild/buf
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
-          DOCKER_LATEST_BRANCH: master
-          DOCKER_VERSION_TAG_PREFIX: v
+
```
bufdev, line 38: no newline

Dockerfile.buf
```diff
@@ -7,14 +7,18 @@ RUN go mod download
...
 
 RUN apk add --update --no-cache \
-    ca-certificates \
-    git \
-    openssh-client && \
+  ca-certificates \
```
bufdev, line 19: revert indent

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
+    needs: build
+    if: github.ref == 'refs/heads/master' || contains(github.ref, 'refs/tags/v')
+    steps:
+      - name: setup-qemu
+        uses: docker/setup-qemu-action@v1
+        id: qemu
+        with:
+          # alpine image doesn't support linux/riscv64
+          platforms: linux/386,linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6,linux/ppc64le,linux/s390x
+
```
bufdev, line 50: no newline

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
+      - name: setup-docker-buildx
+        uses: docker/setup-buildx-action@v1
+
```
bufdev, line 53: no newline

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
+      - name: login-docker
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+
```
bufdev, line 59: no newline

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
+      - name: set-tag
+        id: tag
+        run: |
+          if echo "${GITHUB_REF}" | grep ^refs/heads/master$ >/dev/null; then
+            echo ::set-output name=tag::latest
+          elif echo ${GITHUB_REF} | grep ^refs/tags/v >/dev/null; then
+            echo ::set-output name=tag::"$(echo "${GITHUB_REF}" | sed "s/refs\/tags\/v//")"
+          fi
+
```
bufdev, line 68: no newline

bufdev (commented):
Thanks! We'll consider this change and do our own investigation.

Commenting in places that existing style isn't maintained - please match existing style.

This change isn't compatible with the Makefiles, and they will need to be updated as well.

.github/workflows/ci.yaml
```diff
@@ -25,20 +27,51 @@ jobs:
...
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+
+      - name: set-tag
+        id: tag
+        run: |
+          if echo "${GITHUB_REF}" | grep ^refs/heads/master$ >/dev/null; then
+            echo ::set-output name=tag::latest
+          elif echo ${GITHUB_REF} | grep ^refs/tags/v >/dev/null; then
```
bufdev, line 65: quotes

Dockerfile.buf
Juneezee: According to [Docker Buildx | Docker Documentation](https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images)

> Finally, depending on your project, the language that you use may have good support for cross-compilation. In that case, multi-stage builds in Dockerfiles can be effectively used to build binaries for the platform specified with --platform using the native architecture of the build node. A list of build arguments like BUILDPLATFORM and TARGETPLATFORM is available automatically inside your Dockerfile and can be leveraged by the processes running as part of your build.

```
FROM --platform=$BUILDPLATFORM golang:alpine AS build
ARG TARGETPLATFORM
ARG BUILDPLATFORM
RUN echo "I am running on $BUILDPLATFORM, building for $TARGETPLATFORM" > /log
FROM alpine
COPY --from=build /log /log
```

Juneezee: Yes, you are right.

Reference:
1. https://docs.docker.com/engine/reference/builder/#from
2. https://docs.docker.com/buildx/working-with-buildx/#build-multi-platform-images

.github/workflows/ci.yaml
```diff
@@ -32,13 +32,38 @@ jobs:
...
-        if: success()
-        run: bash .github/scripts/docker.bash
-        env:
-          DOCKER_BUILD_MAKE_TARGET: dockerbuildbuf
-          DOCKER_IMAGE: bufbuild/buf
-          DOCKER_USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          DOCKER_TOKEN: ${{ secrets.DOCKER_TOKEN }}
-          DOCKER_LATEST_BRANCH: master
-          DOCKER_VERSION_TAG_PREFIX: v
+  docker:
```
ZymoticB, line 35: I think we've lost the check to ensure this isn't run on forks?

ZymoticB: https://github.com/bufbuild/buf/blob/master/.github/scripts/docker.bash#L20

Juneezee: I will update the `if` conditional to `github.head_ref == "" && (github.ref == 'refs/heads/master' || contains(github.ref, 'refs/tags/v'))`

ZymoticB: I think the existing logic was actually incorrect. https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context

`github.head_ref` is only set during a PR, so a push to master / a version tag in a fork should pass this conditional (which we do not want). I believe the correct check (for skipping forks; the master / tags/v checks should remain) is 
```
github.repository == 'bufbuild/buf'
```

Additionally, can you add a comment with some of the details here so that a reader unexperienced with github actions understands the intent.

ZymoticB (commented):
Last round of comments I think, thanks for the contribution!

.github/workflows/ci.yaml
```diff
@@ -32,13 +32,38 @@ jobs:
...
+        uses: docker/setup-buildx-action@v1
+      - name: login-docker
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      - name: set-tag
+        id: tag
+        run: |
+          if echo "${GITHUB_REF}" | grep ^refs/heads/master$ >/dev/null; then
```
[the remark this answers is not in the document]
Juneezee: Added the comment. Thanks for reviewing this PR :smiling_face_with_three_hearts:

ZymoticB:
@Juneezee it looks like https://github.com/bufbuild/buf/pull/289#discussion_r596237794 was not addressed but marked as resolved.

Juneezee:
> @Juneezee it looks like [#289 (comment)](https://github.com/bufbuild/buf/pull/289#discussion_r596237794) was not addressed but marked as resolved.

Sorry for missing that one. I have updated accordingly and added some comments.

ZymoticB (approved):
Awesome! This looks good to me. I'll coordinate with @bufdev to get it merged.

.github/workflows/ci.yaml
Juneezee: I think you are right. I am not too sure what would be the output of `github.head_ref` actually. I was just using the logic from https://github.com/bufbuild/buf/blob/master/.github/scripts/docker.bash#L20 directly.

bufdev (approved):

PR #318, opened 2021-05-10
Fix double-dot missing flag errors

### LINKED ISSUE #317: Some missing flag errors include two full stops
:wave: I noticed that some missing flag error messages have two full stops, e.g.:

```
$ buf breaking
Usage:
  buf breaking --against against-input <input> [flags]

...snip...

Failed to "breaking": Flag --against is required..
```

This happens in a few commands:

```
$ rg -l "Flag.*is required\."
internal/buf/bufcli/errors.go
internal/buf/cmd/buf/command/beta/mod/modinit/modinit.go
internal/buf/cmd/buf/command/breaking/breaking.go
internal/buf/cmd/buf/command/build/build.go
```

I fixed these in https://github.com/krishicks/buf/commit/2b589676dc70416b39800ee7219cf88760d3fcc2, and am happy to PR the change if this issue is accepted.

krishicks:
Fixes #317

This updates some errors like this:

```
Failed to "breaking": Flag --against is required..
```

to be this:

```
Failed to "breaking": --against is required.
```

This includes some refactors to use bufcli.NewFlagIsRequiredError rather
than appcmd.NewInvalidArgumentErrorf, but only in those instances where
bufcli was already imported. I wasn't sure if adding that import in e.g.
`internal/pkg/spdx/cmd/spdx-go-data` was desired, so I left those
changes out.

### DIFF
--- a/internal/buf/bufcli/errors.go
+++ b/internal/buf/bufcli/errors.go
@@ -110,7 +110,7 @@ func NewTooManyEmptyAnswersError(attempts int) error {
 
 // NewFlagIsRequiredError informs the user that a given flag is required.
 func NewFlagIsRequiredError(flagName string) error {
-	return appcmd.NewInvalidArgumentErrorf("--%s is required.", flagName)
+	return appcmd.NewInvalidArgumentErrorf("--%s is required", flagName)
 }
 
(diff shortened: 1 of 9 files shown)


johanbrandhorst:
Thank you for your contribution!

PR #323, opened 2021-05-20
Change linux binaries generated by release script to use aarch64 suffix instead of arm64

bufdev:


make/buf/scripts/release.bash
```diff
@@ -75,6 +66,9 @@ cd "${RELEASE_DIR}"
 
 for os in Darwin Linux; do
   for arch in x86_64 arm64; do
+    if [ "${os}" == "Linux" ] && [ "${arch}" == "arm64" ]; then
+      arch="aarch64"
+    fi
```
johanbrandhorst, line 71: You want to add a comment here summarizing the discussion?

PR #326, opened 2021-05-22
Make linter not consider directives to be valid comments

### LINKED ISSUE #297: buf:lint:ignore directives treated as comments
Thanks for buf! It is such a great tool that solves so many problems.

I'm migrating some older protobuf files into a directory that also has some new protobuf files. The new protobuf files can be written with best practices and follow buf linting but the older (already deployed in production) protobuf files cannot pass linting rules. I have chosen to disable explicit lint rules via inline comments instead of blanket disable (in buf.yaml) because I want any new additions to the file to pass buf linting. I noticed a couple potential issues:

1. The `// buf:lint:ignore X` directives are copied as comments into the rendered (for example golang) code.  I'm not sure this is a huge problem - it was just a little unexpected. I guess I expected buf to "consume" the directives and not pass them on to the various rendering plugins. I probably would not have even mentioned this but I noticed the next item which is possibly related...
2. The `// buf:lint:ignore X` directives count as comments and hence pass the `COMMENTS` linting. This means that you can effectively skip the need to add a comment:

```
service MyService {
    // buf:lint:ignore RPC_REQUEST_RESPONSE_UNIQUE
    // buf:lint:ignore RPC_REQUEST_STANDARD_NAME
    rpc GetSomething(BadRequestName) returns (BadResponseName) {}
    // ^^ The above will count as "comments" and will pass buf comment related linting - no need to add comments.
}
```

WillAbides:
This makes COMMENT_* checks stop considering buf:lint:ignore directives to be valid comments.

fixes #297

### DIFF
--- a/internal/buf/bufcheck/buflint/buflint_test.go
+++ b/internal/buf/bufcheck/buflint/buflint_test.go
@@ -36,8 +36,10 @@ import (
 )
 
 // Hint on how to get these:
-// 1. cd into the specific diriectory
-// 2. buf lint --error-format=json | jq '[.path, ".", .start_line, .start_column, .end_line, .end_column, .type] | @csv' --raw-output
+// 1. cd into the specific directory
+// 2. buf lint --error-format=json | jq '[.path, .start_line, .start_column, .end_line, .end_column, .type] | @csv' --raw-output
+//      or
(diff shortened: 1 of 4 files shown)


bufdev:
This is great! Thanks so much.

PR #338, opened 2021-06-08
Support git clone for all branches

### LINKED ISSUE #303: buf limits the git commit names that can be use with `breaking --against`
buf seems only to accept local branch to perform breaking change against. 

It would be useful the check against a remote tracking branch for instance: `--against '.git#branch=origin/main'`

Other useful names would be the last commit `HEAD` or the upstream branch `@{upstream}`

In fact, I see no reason why we couldn't pass any name supported by `gitrevision(7)` including specific commit hash.

powerman:
Actually closes #303

### DIFF
--- a/internal/pkg/git/cloner.go
+++ b/internal/pkg/git/cloner.go
@@ -77,7 +77,7 @@ func (c *cloner) CloneToBucket(
 	}
 
 	depthArg := strconv.Itoa(int(depth))
-	args := []string{"clone", "--depth", depthArg}
+	args := []string{"clone", "--no-single-branch", "--depth", depthArg}
 
 	if options.Name != nil {
 		if cloneBranch := options.Name.cloneBranch(); cloneBranch != "" {


bufdev:
We clone a single branch on purpose here for performance reasons - you'll note on line 84 that if you specify a branch name, we will provide it as an argument. You do not need to have your current branch checked out to work (re: your comment on #303) - I'm able to compare against a different branch than the currently checked out branch with `buf` right now.

bufdev:
If you have a specific, reproducible issue, feel free to file it with reproduction steps, and we'll investigate. If we accept it, we'll happily accept a PR per https://github.com/bufbuild/buf/blob/master/.github/pull_request_template.md

powerman:
Yes, it's a reproducible issue: just delete local branch master and try it out.
The use case is running on CI, which doesn't create local "master" branch by default (when testing PR from some other branch), but (in most cases) do have "origin/master" (because of using enough `--depth`).

powerman:
Don't get me wrong, I can open a new issue, but #303 is THE issue. So, if you need accepted open issue then you can just reopen #303 - it have a valid title and a whole story in the comments. Please let me know if you anyway needs a new issue or any other info.

powerman:
Okay, I'll try to prepare a separate reproducible issue, no problem. But in my experience, for most repos:
- current 50 commits is way too much, so you already fetch much more than needed
- there are not so many branches at any moment of time in average to make this a real issue
- it can be easier for repo developers to make this choice: faster clone vs non-local branch support

So, just make this an option, disabled by default. I fully understand trade-offs for adding a lot of extra options, but in this case it sounds like very important buf's feature can't be used without adding extra workarounds on any CI, so to me it looks like adding one more options may worth it here.

PR #360, opened 2021-07-02
Fix windows generation of protoc default plugins

yinzara:
This issue arises for `buf generate ...` in Windows when you're using one of the built in plugins to protoc (i.e. js, csharp, etc).

The protoc_proxy_handler normally uses stdin to pass the descriptor set to the protoc command however on Windows because there is no stdin file, I had to use a temporary file instead.

Since our test cases for Windows don't yet pass, I'm not sure how to show that this PR actually works. I can tell you I've tested on my windows machine and I'm able to run `buf generate ...` now when it includes one of the standard plugins.

### DIFF
--- a/internal/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go
+++ b/internal/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go
@@ -17,7 +17,6 @@ package appprotoexec
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"os/exec"
@@ -69,9 +68,6 @@ func (h *protocProxyHandler) Handle(
 	ctx, span := trace.StartSpan(ctx, "protoc_proxy")
 	span.AddAttributes(trace.StringAttribute("plugin", filepath.Base(h.pluginName)))
 	defer span.End()
-	if app.DevStdinFilePath == "" {
-		return errors.New("app.DevStdinFilePath is empty for this platform")
-	}
 	protocVersion, err := h.getProtocVersion(ctx, container)
 	if err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


bufdev:
Thanks so much for this! We'll look at getting this in tomorrow.

internal/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go
```diff
@@ -86,6 +82,16 @@ func (h *protocProxyHandler) Handle(
 	if err != nil {
 		return err
 	}
+	descriptorFilePath := app.DevStdinFilePath
+	var tmpFile tmp.File
+	if descriptorFilePath == "" {
+		// since we have no stdin file (i.e. Windows), we're going to have to use a temporary file
+		if tmpFile, err = tmp.NewFileWithData(fileDescriptorSetData); err != nil {
+			return err
+		} else {
```
bufdev, line 91: if we have a variable, we don't use `elses`, we would do:

```go
tmpFile, err = tmp.NewFileWithData(fileDescriptorSetData)
if err != nil {
  return err
}
defer func() {
  retErr = multierr.Append(retErr, tmpFile.Close())
}()
descriptorFilePath = tmpFile.AbsPath()
```

internal/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go
```diff
@@ -129,6 +137,9 @@ func (h *protocProxyHandler) Handle(
 	if featureProto3Optional {
 		responseWriter.SetFeatureProto3Optional()
 	}
+	if tmpFile != nil {
```
bufdev, line 140: Remove this in favor of the `defer` in the comment above

internal/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go
```diff
@@ -118,7 +124,9 @@ func (h *protocProxyHandler) Handle(
 	)
 	cmd := exec.CommandContext(ctx, h.protocPath, args...)
 	cmd.Env = app.Environ(container)
-	cmd.Stdin = bytes.NewReader(fileDescriptorSetData)
+	if app.DevStdinFilePath != "" {
```
bufdev, line 127: I would change this to `if descriptorFilePath == app.DevStdinFilePath`

bufdev (commented):
Looks good overall! Just a few comments.

bufdev:
Merging as-is and I will make the fixes on our side - thanks so much!

PR #411, opened 2021-08-11
Update fuzz tests for bufimagebuild

WillAbides:
This makes several changes to fuzz tests:

- Uses protoc as an oracle

It now compares buf and protoc results. It will panic if one errors but the other does not. It also panics if the images they produce differ.

- Changes data format to [txtar](https://pkg.go.dev/golang.org/x/tools/txtar#hdr-Txtar_format)

The existing corpus files were also not working properly because the first line of the copyright headers was being interpreted as a file name. Switching to txtar removed the possibility of comments being interpreted as the start of a new file. txtar also has the advantage of being a more common format with a library we can import.

- Removes copyright headers from corpus files

The txtar format allows for header comments, so the copyright header can be added back if needed. I removed it because its presence makes the mutator less likely to make meaningful mutations.

- Expands the corpus

I grabbed protobufs from various testdata directories around the repo. A bigger corpus means more variety for the mutator to draw from.

- Adds crashers

These are the crashers I got from fuzzing for a few hours.  I don't know if it's a good idea to merge with these crashers in place.

- Adds unit tests for corpus
- Adds unit tests for crashers (currently skipped)
- Updates `make gofuzz`

It now runs `go get github.com/dvyukov/go-fuzz/go-fuzz-dep` before fuzzing then reverts go.mod and go.sum on exit. This made it a little complex for a Makefile, so I moved this to a script.


WillAbides:
For reviewers looking to avoid scrolling through all the corpus files, most of the non-corpus changes are in https://github.com/bufbuild/buf/pull/411/commits/14ee721e87f8ee93fe85c6612a7029bd5a5933a0.

make/buf/scripts/fuzz.sh
```diff
@@ -0,0 +1,23 @@
...
+
+set -e
+
+CDPATH="" cd "$(dirname "${0}")/../../.."
+FUZZ_DIR="${FUZZ_DIR:-$(pwd)/.tmp/gofuzz}"
+GO_FUZZ_VERSION="${GO_FUZZ_VERSION:-master}"
+
+git diff --exit-code --quiet go.mod go.sum || (echo "go.sum and go.mod must be unmodified" && exit 1)
+
+go get github.com/dvyukov/go-fuzz/go-fuzz-dep@"$GO_FUZZ_VERSION"
```
johanbrandhorst, line 11: Why is this version managed in this script rather than in the `make` target?

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,205 @@
...
-		return 0, err
+}
+
+// untxtar extracts txtar data to destDir.
+func untxtar(data []byte, destDir string) (err error) {
+	defer func() {
+		if p := recover(); p != nil {
+			err = fmt.Errorf("panic: %v", p)
+		}
+	}()
```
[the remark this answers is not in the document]
WillAbides: That's right. txtar panics with `slice bounds out of range [3:2]` when parsing a line like `-- --`. I think this it's related to this issue: https://github.com/golang/go/issues/47193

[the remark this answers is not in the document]
johanbrandhorst: Could you add a comment here to explain what you just explained to me 😄?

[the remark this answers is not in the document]
WillAbides: Will do

make/buf/all.mk
```diff
@@ -150,6 +150,4 @@ bufrelease:
 .PHONY: gofuzz
 gofuzz: $(GO_FUZZ)
 	@rm -rf $(TMP)/gofuzz
-	@mkdir -p $(TMP)/gofuzz
-	cd internal/buf/bufbuild/bufbuildtesting; go-fuzz-build -o $(abspath $(TMP))/gofuzz/gofuzz.zip
-	go-fuzz -bin $(TMP)/gofuzz/gofuzz.zip -workdir $(TMP)/gofuzz
+	FUZZ_DIR=$(abspath $(TMP))/gofuzz GO_FUZZ_VERSION=$(GO_FUZZ_VERSION) make/buf/scripts/fuzz.sh
```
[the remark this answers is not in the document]
WillAbides: $(GO_FUZZ) is still needed to install the `go-fuzz` and `go-fuzz-build` binaries.

make/buf/scripts/fuzz.sh
WillAbides: GO_FUZZ_VERSION's value comes from dep_go_fuzz.mk when this script is called from the gofuzz target. [This line](https://github.com/WillAbides/buf/blob/13394c98ffee7b25c9e3aee5e7ad9a954b48eef4/make/buf/scripts/fuzz.sh#L8) may make it look like this script is managing the version, but that is just setting a default value.

I could be more clear by removing the default value and erroring when no value is set.

johanbrandhorst: Sorry, I was being unclear in my initial message - I mean, why isn't this binary just installed together with `go-fuzz` and `go-fuzz-build`?

WillAbides: I see what you mean now. This doesn't install a binary. It just adds "github.com/dvyukov/go-fuzz" to go.mod and go.sum so that go-fuzz-build can build an instrumented binary.

Now that you bring it up, we could have this script install go-fuzz and go-fuzz-build as well so that it's all handled in one place...or is that what you were suggesting before?

johanbrandhorst: I think I'd prefer if we can handle it in the make targets (like `dep_gofuzz.mk`), since that's the pattern we generally follow for binary dependencies. It sounds like this isn't exactly that, but that `go-fuzz` needs `go-fuzz` in its dependency closure when building the binary. I'm not sure I understand this part, could you elaborate a little bit? It's not imported anywhere, so what does adding it to the dependency closure accomplish?

bufdev: dependencies should be exclusively installed via the makefile commands, and there should be a corresponding make target to invoke this script that depends on the installation of the dependency

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
-		buf := bytes.NewBuffer(file)
-		header, err := buf.ReadBytes('\n')
-		if err != nil {
-			return 0, err
-		}
-		content, err := io.ReadAll(buf)
-		if err != nil {
-			return 0, err
+func fuzz(ctx context.Context, data []byte) (_ *fuzzResult, retErr error) {
+	dirPath, err := os.MkdirTemp("", "buffuzz")
```
bufdev, line 52: we have `tmp.NewDir()` from `internal/pkg/tmp`, this should be used

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		err := os.RemoveAll(dirPath)
+		if retErr == nil {
```
bufdev, line 58: we do `retErr = multierr.Append(retErr, err)`

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
+			retErr = err
 		}
-		// Trim comment prefix and newline suffix
-		fileName := strings.TrimSpace(strings.TrimPrefix(string(header), "// "))
-		pathToData[fileName] = content
+	}()
+	err = untxtar(data, dirPath)
```
bufdev, line 62: single errors returned from functions should be scoped to the if statement

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
+	// realistic looking data.
+	if f.protocErr == nil {
+		return 1
+	}
+	return 0
+}
+
+// error returns an error that should cause Fuzz to panic.
+func (f *fuzzResult) error(ctx context.Context) error {
+	if f == nil {
```
bufdev, line 199: this should never be true

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
+	}
+	for _, file := range archive.Files {
+		dir := filepath.Dir(file.Name)
+		if dir != "" {
+			err = os.MkdirAll(filepath.Join(destDir, dir), 0700)
+			if err != nil {
+				return err
+			}
+		}
+		err = os.WriteFile(
```
bufdev, line 163: scope

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,217 @@
...
+	defer func() {
+		if p := recover(); p != nil {
+			retErr = fmt.Errorf("panic from txtar.Parse: %v", p)
+		}
+	}()
+	return txtar.Parse(data), nil
+}
+
+// untxtar extracts txtar data to destDir.
+func untxtar(data []byte, destDir string) error {
```
bufdev, line 147: `destDirPath` or just `dirPath`

make/buf/scripts/fuzz.sh
bufdev: this should be removed

make/buf/scripts/fuzz.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/sh
```
bufdev, line 1: `#!/usr/bin/env bash`, rename to `fuzz.bash`

make/buf/scripts/fuzz.sh
```diff
@@ -0,0 +1,25 @@
...
+
+set -e
+
+DIR="$(CDPATH= cd "$(dirname "${0}")/../../.." && pwd)"
+cd "${DIR}"
+
+[ -n "$FUZZ_DIR" ] || (echo "FUZZ_DIR is required" && exit 1)
+[ -n "$GO_FUZZ_VERSION" ] || (echo "GO_FUZZ_VERSION is required" && exit 1)
+
+git diff --exit-code --quiet go.mod go.sum || (echo "go.sum and go.mod must be unmodified" && exit 1)
```
bufdev, line 11: use https://github.com/bufbuild/buf/blob/main/make/buf/scripts/release.bash#L9

make/buf/scripts/fuzz.sh
```diff
@@ -0,0 +1,25 @@
...
+
+go get github.com/dvyukov/go-fuzz/go-fuzz-dep@"$GO_FUZZ_VERSION"
+trap "git checkout -- go.mod go.sum" EXIT
```
bufdev, line 14: why is this needed? likely it isnt if using the makefile to install the dep

make/buf/scripts/fuzz.sh
WillAbides: go-fuzz-build loads "github.com/dvyukov/go-fuzz/go-fuzz-dep" and errors if it is missing from go.mod.  This is where it happens:

https://github.com/dvyukov/go-fuzz/blob/b1f3d6f4ef4e0fab65fa66f9191e6b115ad34f31/go-fuzz-build/main.go#L297-L313

WillAbides: @bufdev 
> dependencies should be exclusively installed via the makefile commands, and there should be a corresponding make target to invoke this script that depends on the installation of the dependency

This doesn't install a dependency, or at least not a binary dependency. All it does is temporarily add "github.com/dvyukov/go-fuzz/go-fuzz-dep" to the module so that `go-fuzz-build` can use it.

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting_test.go
```diff
@@ -15,18 +15,55 @@
...
+		t.Run(entry.Name(), func(t *testing.T) {
+			data, err := os.ReadFile(filepath.Join("corpus", entry.Name()))
+			require.NoError(t, err)
+			testFuzz(t, data)
+		})
+	}
+}
+
+func TestCrashers(t *testing.T) {
+	t.Skip("skipping known crashers")
```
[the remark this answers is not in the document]
WillAbides: It's temporary one way or the other. I'm unsure whether it's a good idea to merge crashers into main or to fix the issues and add the former crashers to corpus. I lean toward the latter. The main reason I added crashers to this PR was to show the results I got.

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting_test.go
```diff
@@ -15,18 +15,55 @@
...
+	t.Helper()
+	ctx := context.Background()
+	result, err := fuzz(ctx, data)
 	require.NoError(t, err)
-	i, err := fuzz(data)
+	require.NoError(t, result.error(ctx))
+}
+
+func TestCorpus(t *testing.T) {
+	dir, err := os.ReadDir("corpus")
```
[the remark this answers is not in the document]
WillAbides: I switched to filepath.Walk, but I don't think it adds anything here.

WillAbides:
@bufdev @johanbrandhorst Thanks for the reviews. I think I've addressed everything now.

The shell script seemed to be sowing confusion, so I dropped it and put everything in make. It seems simpler this way.

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go
```diff
@@ -15,84 +15,224 @@
...
+		false,
+	)
+
+	image, bufAnnotations, bufErr := fuzzBuild(ctx, dir.AbsPath())
+	return newFuzzResult(
+		bufAnnotations,
+		bufErr,
+		protocErr,
+		actualProtocFileDescriptorSet,
+		image), nil
```
johanbrandhorst, line 84: Style nit: put the closing bracket on a new line 😁 
```suggestion
		image,
	), nil
```

johanbrandhorst (commented):
This looks great. Did you see my meta-comment about adding the corpus to .gitattributes so it's ignored by Github PRs?

make/buf/scripts/fuzz.sh
johanbrandhorst: I see, that sounds like a pretty unique requirement of the fuzz program. I can't wait for Go 1.17 when we can hopefully replace it with `testing.F`!

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/crashers/e55d098c6ca004c39f85be18610ecad5da52b268.output
```diff
@@ -0,0 +1,12 @@
+panic: buf has file annotations but protoc has no error: [/var/folders/gt/cvjl_x695bg3n1p617s0gs6c0000gn/T/buffuzz136699593/000.proto:1:1:syntax error: unexpected $unk]
+
+goroutine 1 [running]:
+github.com/bufbuild/buf/internal/buf/bufimage/bufimagebuild/bufimagebuildtesting.(*fuzzResult).panicOrN(0xc000270f00, 0x193f220, 0xc00002c108, 0x13)
+	/Users/wroden/repos/WillAbides/buf/internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/bufimagebuildtesting.go:205 +0x16c
```
bufdev, line 5: All of these extra files - do we have to check them in? If we do, we for sure can't check in `/Users/wroden`

bufdev (commented):
We didn't want the script removed, I think the new make target has issues, namely no traps, I'd bring it back

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/crashers/e55d098c6ca004c39f85be18610ecad5da52b268.output
johanbrandhorst: The *.output and *.quoted crasher files could probably be `.gitignored` TBH. As long as we store the input for future runs we can reproduce the bugs found.

WillAbides: Good idea with .gitignore. I'll do that.

WillAbides:
@bufdev 
> We didn't want the script removed, I think the new make target has issues, namely no traps, I'd bring it back

I don't think the trap is necessary anymore now that it's using go.*.bak instead of relying on `git checkout` to restore the files.

make/buf/all.mk
```diff
@@ -27,7 +27,9 @@ FILE_IGNORES := $(FILE_IGNORES) \
 	internal/buf/cmd/buf/cache/ \
 	internal/buf/cmd/buf/workspacetests/other/proto/workspacetest/cache/ \
 	internal/buf/internal/buftesting/cache/ \
-	internal/pkg/storage/storageos/tmp/
+	internal/pkg/storage/storageos/tmp/ \
+	**/crashers/*.output \
+    **/crashers/*.quoted
```
johanbrandhorst, line 32: Nit: I think this is mixing tabs and spaces

WillAbides: Good eye. Spaces in a Makefile is a classic blunder.

johanbrandhorst (approved):
LGTM but wait for Peter's thoughts on the script 😄.

make/buf/all.mk
```diff
@@ -27,7 +27,9 @@ FILE_IGNORES := $(FILE_IGNORES) \
 	internal/buf/cmd/buf/cache/ \
 	internal/buf/cmd/buf/workspacetests/other/proto/workspacetest/cache/ \
 	internal/buf/internal/buftesting/cache/ \
-	internal/pkg/storage/storageos/tmp/
+	internal/pkg/storage/storageos/tmp/ \
+	**/crashers/*.output \
```
bufdev, line 31: I'd prefer no `**` in this expression, as I don't believe this works with `.dockerignore`. Even the `*.output` and `*.quoted` might not work, which is why we only use paths otherwise`.

Replace `**` with the actual paths to the directories.

make/buf/all.mk
```diff
@@ -150,6 +152,13 @@ bufrelease:
 .PHONY: gofuzz
 gofuzz: $(GO_FUZZ)
 	@rm -rf $(TMP)/gofuzz
-	@mkdir -p $(TMP)/gofuzz
-	cd internal/buf/bufbuild/bufbuildtesting; go-fuzz-build -o $(abspath $(TMP))/gofuzz/gofuzz.zip
+	@mkdir -p $(TMP)/gofuzz $(TMP)/gofuzz/corpus $(TMP)/gofuzz/crashers
+	# go-fuzz-build requires github.com/dvyukov/go-fuzz be in go.mod, but we don't need that dependency otherwise.
+	# This adds go-fuzz-dep to go.mod, runs go-fuzz-build, then restores go.mod.
+	cp go.mod go.mod.bak; cp go.sum go.sum.bak
```
bufdev, line 158: Without the trap, if the next line fails, we now have leftover files that are not .gitignored. You can copy them to `$(TMP)` to make sure they're ignored, or re-add the script.

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/crashers/115721e379c54d9f8c9055ba554c3c35afce890f
```diff
@@ -0,0 +1,2 @@
+-- .proto --
```
bufdev, line 1: Do we need all these files too?

WillAbides:
When going through the crashers, I think I found a design flaw in the fuzz tests. I am comparing the results of `builder.Build()` with `protoc generate`. 

Several of the crashers are name conflicts that `protoc generate` catches but `builder.Build()` does not. When I try running `buf generate` with the same input, I see that it behaves similar to `protoc generate`

It appears I made an incorrect assumption that all syntax checks happen in `builder.Build()`.

internal/buf/bufimage/bufimagebuild/bufimagebuildtesting/crashers/115721e379c54d9f8c9055ba554c3c35afce890f
WillAbides: I don't think we do. I would rather remove the crashers before merging. I only included them as a way to show the issues the fuzzer found.

WillAbides: I've removed them

bufdev (approved):

Issue #398, opened 2021-08-05
buf does not generate all dart files 


Zhalkhas:
When I generate dart code from schemes, I get files which have relative includes like 
`import '../../../google/protobuf/timestamp.pb.dart' as $1;` , or 
`import '../../../google/type/date.pb.dart' as $4;`
which are not resolved. How can I make buf generate code for external deps like this?

johanbrandhorst:
Hi, thanks for your issue! `buf` generate currently only generates the immediately specified module, and not all dependencies. To also generate code for your dependencies, today you have to explicitly do so in a separate invocation. We've discussed adding something like a flag to allow users to do this easily, but it's not available yet.

However, in your case, it seems that at least one of your unsatisfied import is a _well known type_ (namely `timestamp.pb.dart`). These types should be provided by your runtime library, and not included as a generated dependency in every language I ever used protobuf in (I haven't had the pleasure to try it with Dart yet). The second one (`date.pb.dart`) looks like it came from `google/type/date.proto` which is part of the `googleapis` module. You can generate this type using:

```shell
buf generate buf.build/beta/googleapis
```

Please feel free to [join our public Slack](https://join.slack.com/t/bufbuild/shared_invite/zt-f5k547ki-VDs_iC4TblNCu7ubhRD17w) if you need more assitance.

Zhalkhas:
Thanks for reply! I'm closing this issue then

amitkot:
@johanbrandhorst 
Hi, I just ran into this as well. The problem with your suggestion regarding `timestamp` being a *well-known-type* is that the generated code tries to import it using a relative import which at least in my case means I have to generate this file manually in that position. I opted instead to copy `timestamp.proto` into my project and have **buf** generate the dart file for it as well.

Is it possible to have an option for generating well-known-type files without copying them to the project?

timostamm:
Hey Amit, since this issue was created, we've added two flags to `buf generate`:

```
      --include-imports        Also generate all imports except for Well-Known Types.
      --include-wkt            Also generate Well-Known Types. Cannot be set without --include-imports.
```

Could you have a look whether `include-wkt` solves the issue for you?

amitkot:
Hi @timostamm tried it now and it works great, thanks!

I also see now that this is well documented in the CLI tool. I couldn't find it on the website, though.

PR #416, opened 2021-08-12
Add buf generate --include-imports

bufdev:
This adds `buf generate --include-imports`, which will generate all imports except for the Well-Known Types. For example, if you import `google/type.date.proto` from googleapis, this will generate just that. If there is later a use case to also generate the Well-Known Types, we can add a `--include-wkt` flag.


Goxiaoy:
Please add `--include-wkt` for dart generators

Releated issue 
https://github.com/bufbuild/buf/issues/398
https://github.com/google/protobuf.dart/issues/483

PR #429, opened 2021-08-17
Add regression tests

### LINKED ISSUE #417: Inconsistencies between buf and protoc
These are some inconsistencies between buf and protoc found in #411

### Form Feeds and Vertical Tabs

Protobufs containing the characters "\f" and "\v" fail with `buf generate` but succeeds with protoc. Buf outputs the message "syntax error: unexpected $unk". This seems to be coming from [this line](https://github.com/jhump/protoreflect/blob/fb56a945d0735c55f3bb1853ad1f99766ac126e7/desc/protoparse/proto.y.go#L925) in "github.com/jhump/protoreflect". I'm not familiar with goyacc, so my diagnosis stops there.

### Duplicate message and rpc names

Buf allows an rpc's name to be the same as a message's name in the same protobuf, but protoc does not.

For example:

```protobuf
syntax="proto3";

package a;

option go_package = "foo.bar/foo";

message Foo{}
message Bar{}

service MyService{
  rpc Foo(Foo) returns (Bar);
}
```

Buf generates code for ☝️, but protoc errors with the message: `"Foo" is not a message type.`

### Space after first number in extensions

With this protobuf:

```protobuf
syntax="proto2";

package a;

option go_package = "foo.bar/foo";

message Test{
  extensions 10to max;
}
```

protoc fails with `Need space between number and identifier.` but Buf generates code.

WillAbides:
This adds some regression tests for issues fixed in #430

### DIFF
--- a/internal/buf/bufimage/bufimagebuild/builder_test.go
+++ b/internal/buf/bufimage/bufimagebuild/builder_test.go
@@ -201,12 +201,29 @@ func TestCompareProto3Optional1(t *testing.T) {
 
 func TestCustomOptionsError1(t *testing.T) {
 	t.Parallel()
-	_, fileAnnotations := testBuild(t, false, filepath.Join("testdata", "customoptionserror1"))
-	require.Equal(t, 1, len(fileAnnotations), fileAnnotations)
-	require.Equal(
+	testFileAnnotations(
 		t,
+		"customoptionserror1",
 		"field a.Baz.bat: option (a.foo).bat: field bat of a.Foo does not exist",
-		fileAnnotations[0].Message(),
+	)
+}
+
+func TestNotAMessageType(t *testing.T) {
+	t.Parallel()
+	testFileAnnotations(
+		t,
+		"notamessagetype",
+		"method a.MyService.Foo: invalid request type: a.MyService.Foo is a method, not a message",
+	)
+}
+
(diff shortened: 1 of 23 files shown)


bufdev:
We have this covered internally, will copy over!

bufdev:
This is mostly supplanted by https://github.com/bufbuild/buf/pull/430 - you don't have to, but any chance you could merge that in/revert your common changes here?

internal/buf/bufimage/bufimagebuild/builder_test.go
```diff
@@ -314,3 +331,12 @@ func testGetImageImportPaths(image bufimage.Image) []string {
 	sort.Strings(importNames)
 	return importNames
 }
+
+func testFileAnnotations(t *testing.T, relDirPath string, wantMessages ...string) {
```
WillAbides, line 335: This hit the 3 times threshold so I extracted it.

WillAbides:
I merged main and updated the PR to reflect it's new purpose.

internal/buf/bufimage/bufimagebuild/builder_test.go
bufdev: We don't have a 3 times threshold, but this is fine

bufdev (approved):

PR #456, opened 2021-08-26
Cache: Prevent concurrent module fetches

amckinney:
In short, concurrent processes aren't prevented from entering and executing the "fetch and put" branch, so users will sometime observe duplicative log lines like the following:

```sh
buf: downloading buf.build/bufbuild/buf:9ab0c4681d664e608a2b9509eaae10a3
buf: downloading buf.build/bufbuild/buf:9ab0c4681d664e608a2b9509eaae10a3
```

This proposes another file lock (distinguished by the `.download` suffix) that makes it so that only one process will ever "fetch and put" the module in the cache. Other processes that enter this branch and acquire the lock will immediately check if the module was already fetched so that they can rely on one of the other concurrent processes that finished first.


bufdev:
@amckinney maybe let's close this for now? This isn't actually an issue, and I'd prefer to keep focus on other items

amckinney:
Yea, let's follow-up with another PR when we're ready to close this out.

Issue #465, opened 2021-08-27
Managed Mode configuration


amckinney:
There have been several feature requests for extending the [Managed Mode](https://docs.buf.build/generate/managed-mode) configuration options so that users can configure how each file option behaves (similar to the [`go_package_prefix`](https://docs.buf.build/generate/managed-mode#go_package_prefix) option).

We've discussed this at length before, and consciously decided to keep it strict and opinionated for the time being. However, we can adapt the `buf.gen.yaml` in a backwards-compatible manner to accommodate these use cases if we ever decide to.

For example, we can adopt the same style that the `lint` and `breaking` keys use (i.e. add rule and category IDs for each option):

```yaml
# buf.gen.yaml
version: v1
managed:
  use:
    - DEFAULT
  except:
    - JAVA_MULTIPLE_FILES
  ignore:
    - foo/bar.proto
  ignore_only:
    GO_PACKAGE:
      - baz/qux.proto
  java_package_prefix: "org"
```

> In this new model, the `enabled: true` value could be implicitly set if `managed.use` is non-empty.

There have also been conversations about whether we should move **Managed Mode** configuration to a separate file (e.g. `buf.managed.yaml`) so that it can be more naturally applied to other commands like `buf build`. In this world, we would need to be careful that `buf` remains ergonomic and intuitive to use.

odsod:
Out of everything mentioned in this issue, `java_package_prefix: "org"` would be the most impactful for us, with respect to enabling migration to managed mode without breaking changes to existing generated Java code (and Java artifacts).

amckinney:
We've decided to defer the `use`, `except`, `ignore`, and `ignore_only` approach until it's absolutely necessary (we're not yet convinced it ever will be). **Managed Mode** encourages users to consciously control what file options are set, rather than deferring to their dependencies' source.

Users can `override` file option values for specific files (as described in https://github.com/bufbuild/buf/issues/471). This looks like the following:

```yaml
# buf.gen.yaml
version: v1
managed:
  enabled: true
  override:
    JAVA_OUTER_CLASSNAME:
      google/type/latlng.proto: com.google.type.LatLngProto
```

PR #473, opened 2021-08-29
Add java_package_prefix option

amckinney:
Re: https://github.com/bufbuild/buf/issues/465#issuecomment-907779934

This proposes a `java_package_prefix` option so that users can customize what prefix is used to generate the `java_package` value in **Managed Mode**.

This idea was originally suggested in Slack, with the following context:

```
It's not a super big deal for us if the package path changes as part of adopting Managed Mode,
but it's still a breaking change that affects all our current generated code. It'll make Managed
Mode a little tougher sell.

...

The issue right now is that it's not just the generated code that has a custom TLD as package prefix,
but also our Maven artifacts. E.g:

  <groupId>tech.einride</groupId>
  <artifactId>proto</artifactId>

This is a bit of a hurdle for smooth adoption.
```

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	// defaultJavaPackagePrefix is the default java_package prefix used in the JavaPackage modifier.
-	defaultJavaPackagePrefix = "com."
+	defaultJavaPackagePrefix = "com"
 )
 
 type generator struct {
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	// javaPackagePrefix is the default java_package prefix used in the JavaPackage modifier.
-	javaPackagePrefix = "com."
+	// defaultJavaPackagePrefix is the default java_package prefix used in the JavaPackage modifier.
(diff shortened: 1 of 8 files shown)


johanbrandhorst (approved):

private/buf/bufgen/generator.go
```diff
@@ -29,8 +29,8 @@ import (
 )
 
 const (
-	// javaPackagePrefix is the default java_package prefix used in the JavaPackage modifier.
-	javaPackagePrefix = "com."
+	// defaultJavaPackagePrefix is the default java_package prefix used in the JavaPackage modifier.
+	defaultJavaPackagePrefix = "com."
```
bufdev, line 33: I'd argue we probably want to make this `com`, and define `java_package_prefix` to be used as `PREFIX.PACKAGE`, not `PREFIXPACKAGE`.

amckinney: Yea that's fair. I originally thought we should keep this open-ended so that users could customize their own prefix (e.g. they might not want a `.` after the prefix, etc.). But the gain in configuration cleanliness probably outweighs that. Plus, if there's ever a need for that kind of behavior, we can extend the configuration to support it in a backwards compatible manner.

bufdev (approved):

Issue #454, opened 2021-08-26
Support environment-variable based authentication


bufdev:
Right now when using the BSR with private repositories, we require you to `buf login` to set your `.netrc`. For CI, this isn't a typical workflow, even if we do have `buf login --username USERNAME --token-stdin`. We should support `BUF_USERNAME` and `BUF_TOKEN`.

PR #477, opened 2021-08-30
Support BUF_TOKEN for authentication

bufdev:
Fixes https://github.com/bufbuild/buf/issues/454

We only need to support `BUF_TOKEN`, not `BUF_USERNAME`. The new logic:

- Check for `BUF_TOKEN`. If it exists, use it for all RPCs.
- Otherwise, check the netrc file. If an entry exists for  the address, use it for all RPCs.
- Otherwise, do not authenticate.

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -69,6 +69,8 @@ See https://docs.buf.build/faq for more details.`
 	inputSSHKeyFileEnvKey         = "BUF_INPUT_SSH_KEY_FILE"
 	inputSSHKnownHostsFilesEnvKey = "BUF_INPUT_SSH_KNOWN_HOSTS_FILES"
 
+	tokenEnvKey = "BUF_TOKEN"
+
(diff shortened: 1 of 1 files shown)


private/buf/bufcli/bufcli.go
```diff
@@ -536,22 +538,20 @@ func NewContextModifierProvider(
...
+		token := container.Env(tokenEnvKey)
+		if token == "" {
+			machine, err := netrc.GetMachineForName(container, address)
+			if err != nil {
+				return nil, fmt.Errorf("failed to read server password from netrc: %w", err)
+			}
+			token = machine.Password()
 		}
 		return func(ctx context.Context) context.Context {
-			return rpcauth.WithToken(
```
bufdev, line 548: We do ignore empty values in `WithToken`, but this is making it a bit more explicit

johanbrandhorst (approved):

amckinney (approved):

PR #500, opened 2021-08-31
Client side remote plugin generation

johanbrandhorst:


private/buf/bufgen/config.go
```diff
@@ -157,12 +165,35 @@ func validateExternalConfigV1(externalConfig ExternalConfigV1, id string) error
...
 		}
 		if plugin.Out == "" {
 			return fmt.Errorf("%s: plugin %s out is required", id, plugin.Name)
 		}
+		if plugin.Remote != "" {
+			if _, _, _, err := bufplugin.ParsePluginPath(plugin.Remote); err != nil {
+				return fmt.Errorf("%s: invalid remote plugin name: %w", id, err)
+			}
+			// We limit the amount of validation done on the client side intentionally
+			if plugin.Version == "" {
```
bufdev, line 182: I would probably have us do `buf.build/foo/bar:version` as the value for `remote`, and handle no version as being equivalent to latest

private/buf/bufgen/generator.go
```diff
@@ -76,49 +89,215 @@ func (g *generator) generate(
...
+	}
+	remoteToPlugins := make(map[string][]remotePluginOptions)
+	var localPluginResponses []localPluginResponse
+	// Parallelize local and remote plugin execution with an errgroup.
+	// Note: not using thread.Parallelize because we want to be able to
+	// cancel all execution if one plugin errors. Limit concurrent execution
+	// of local plugins using semaphore channel.
+	semaphoreC := make(chan struct{}, thread.Parallelism())
+	defer close(semaphoreC)
+	eg, ctx := errgroup.WithContext(ctx)
```
bufdev, line 122: We use `thread.Parallelize`. As discussed, if we need to extend this to have a `ctx`, we can

private/buf/bufgen/generator.go
```diff
@@ -76,49 +89,215 @@ func (g *generator) generate(
...
+						}
+					}()
+					if err := handler.Handle(ctx, container, responseWriter, request); err != nil {
+						return fmt.Errorf("failed to generate: %w", err)
+					}
+					return nil
+				})
+			}
+			// Responses are not valid until parallelized jobs have finished
+			localPluginResponses = append(localPluginResponses, localPluginResponse{
```
bufdev, line 183: we always use pointers and constructors

private/buf/bufgen/config.go
johanbrandhorst: I would like to use the format we use for plugins elsewhere and in the UI (`buf.build/owner/plugins/name`), but I like the idea of integrating and parsing the version out of the remote.

private/buf/bufgen/generator.go
johanbrandhorst: Discussed offline, after improvements in #502 I will update this to use `thread.Parallelize`.

private/buf/bufgen/config.go
```diff
@@ -339,3 +370,8 @@ type readConfigOptions struct {
 func newReadConfigOptions() *readConfigOptions {
 	return &readConfigOptions{}
 }
+
+func pluginNameIsRemote(pluginName string) bool {
+	// Local plugins cannot contain "/"
+	return strings.Contains(pluginName, "/")
+}
```
amckinney, line 377: This looks unused (I imagine it was kept around before we started using `bufplugin.ParsePluginPath`).

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -63,21 +63,23 @@ func NewCommand(
...
     # Required.
     out: gen/go
     # Any options to provide to the plugin.
-    # Optional.
     # This can be either a single string or a list of strings.
+    # Optional.
     opt: paths=source_relative
     # The custom path to the plugin binary, if not protoc-gen-NAME on your $PATH.
-    path: custom-gen-go  # optional
+	# Optional, and exclusive with "remote".
```
amckinney, line 81: The spacing is off with this one.

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -241,59 +255,73 @@ func ParseTemplateVersionConfig(config string) (*TemplateVersionConfig, error) {
...
 // MergeInsertionPoints traverses the plugin results and merges any insertion points present in any responses.
-// The order of inserts depends on the order of the input plugins.
-func MergeInsertionPoints(ctx context.Context, pluginResults []PluginResult) (map[string][]byte, error) {
-	files := make(map[string][]byte)
-	for _, pluginResult := range pluginResults {
-		if pluginResult.Response.Error != nil {
-			return nil, fmt.Errorf("unexpected response error: %s", *pluginResult.Response.Error)
+// The order of inserts depends on the order of the input plugins. The returned results are in the same order
+// as the input plugin results.
+func MergeInsertionPoints(responses []*pluginpb.CodeGeneratorResponse) ([]MergedPluginResult, error) {
```
amckinney, line 272: For consistency with pointers elsewhere, this could be a `[]*MergedPluginResult`.

private/buf/cmd/buf/command/generate/generate_test.go
```diff
@@ -222,7 +222,6 @@ func testCompareGeneratedStubsArchive(
 		actualProtocPluginFlags...,
 	)
 	genFlags := []string{
-		"--input",
```
amckinney, line 225: Note that this flag might have been intentionally there to test that the deprecated flag code path continues to work.

private/buf/bufgen/generator.go
```diff
@@ -239,3 +509,53 @@ type generateOptions struct {
...
+func (s semaphore) Release(ctx context.Context) {
+	select {
+	case <-s:
+	case <-ctx.Done():
+	}
+}
+
+func (s semaphore) Discard() {
+	close(s)
+}
```
amckinney, line 561: Curious, why not https://pkg.go.dev/golang.org/x/sync/semaphore? Is the thought that this is simple enough as-is?

amckinney (commented):
Overall the implementation looks good. My main concern is how much code we need to duplicate between `appprotoos` and here (e.g. `generateZip`, `manifestPath`, etc.).

It looks like `buf protoc` still needs to use `appprotoos` so we can't simply move the implementation here, but I wonder if there's something we can do to DRY this up a bit more? What are your thoughts?

private/buf/bufgen/generator.go
johanbrandhorst: Yes, this is simple enough, no need for weights. It's just syntactic sugar around a channel.

private/bufpkg/bufplugin/bufplugin.go
johanbrandhorst: Good shout

johanbrandhorst:
> Overall the implementation looks good. My main concern is how much code we need to duplicate between `appprotoos` and here (e.g. `generateZip`, `manifestPath`, etc.).
> 
> It looks like `buf protoc` still needs to use `appprotoos` so we can't simply move the implementation here, but I wonder if there's something we can do to DRY this up a bit more? What are your thoughts?

Let me see if I can generalize some of the .zip/.jar handling.

private/buf/cmd/buf/command/generate/generate_test.go
johanbrandhorst: The test was not asserting anything around the message, but I can restore this is this was the intention.

johanbrandhorst:
I've undone the joining of the appprotoos changes, turns out it was causing some difficulty with doing insertion points for directories and also apparently we might not support insertion points for .jar outputs? This touches less of the `appprotoos` package now, since it's used for `buf protoc` and we don't want to break any of that behaviour.

amckinney (approved):
Only nits, but this looks good as-is. The _slight_ duplication between `buf protoc` and `buf generate` is less than ideal, but it's probably our best option right now. We can revisit later if need be.

private/buf/bufgen/generator.go
```diff
@@ -239,3 +410,53 @@ type generateOptions struct {
...
+
+func (s semaphore) Acquire(ctx context.Context) error {
+	select {
+	case s <- struct{}{}:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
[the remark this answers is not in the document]
johanbrandhorst: Using a point receiver actually means having to dereference the channel when doing operations on it, which is really ugly.

private/buf/bufgen/generator.go
```diff
@@ -80,48 +91,208 @@ func (g *generator) generate(
...
+			}
+			return nil
+		})
+	}
+	if err := eg.Wait(); err != nil {
+		return nil, fmt.Errorf("failed to execute plugins concurrently: %w", err)
+	}
+	for _, localResponse := range localPluginResponses {
+		pluginResponses[localResponse.index] = localResponse.responseWriter.ToResponse()
+	}
```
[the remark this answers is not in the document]
johanbrandhorst: Overkill defensive programming is in our DNA :salute-canada:

PR #503, opened 2021-09-01
Simplify bufmodule

bufdev:
- Make it so that `newModuleForBucket` and `newModuleForProto` are more independent.
- Do not always read the `documentation` from the `ReadBucket`, instead pass it as a parameter to `newModule`.
- Delete `NewModuleWithDependencyModulePins`.
- Add `DependencyModulePinsForBucket` and expand `PutDependencyModulePinsToBucket`.
- Replace unnecessary creates of `bufmodule.Modules` in `mod prune` and `mod update`.

### DIFF
--- a/private/bufpkg/bufmodule/module.go
+++ b/private/bufpkg/bufmodule/module.go
@@ -17,12 +17,11 @@ package bufmodule
 import (
 	"context"
 	"fmt"
-	"io"
 
 	modulev1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/module/v1alpha1"
+	"github.com/bufbuild/buf/private/pkg/normalpath"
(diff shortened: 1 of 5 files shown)


private/bufpkg/bufmodule/module.go
```diff
@@ -74,49 +72,41 @@ func newModuleForBucket(
...
 	}
-	return newModuleForBucketWithDependencyModulePins(
+	return newModule(
 		ctx,
-		sourceReadBucket,
+		storage.MapReadBucket(sourceReadBucket, storage.MatchPathExt(".proto")),
 		dependencyModulePins,
+		documentation,
 		options...,
 	)
```
amckinney, line 89: Nice 👏

amckinney (approved):
Looks good!

Issue #471, opened 2021-08-29
Java managed mode and google.type.LatLng


odsod:
When turning on managed mode for Java, with a module that depends on buf.build/googleapis/googleapis and a file that imports [google/type/latlng.proto](https://github.com/googleapis/googleapis/blob/master/google/type/latlng.proto), there is a diff in the generated code causing a `cannot find symbol` error for `com.google.type.LatLngProto`.

When inspecting the generated code, the symbol has been generated as `com.google.type.LatlngProto`, most likely since the file is called `latlng.proto` and not `lat_lng.proto`:

```java
    descriptor = com.google.protobuf.Descriptors.FileDescriptor
      .internalBuildGeneratedFileFrom(descriptorData,
        new com.google.protobuf.Descriptors.FileDescriptor[] {
          com.google.api.FieldBehaviorProto.getDescriptor(),
          com.google.type.LatlngProto.getDescriptor(),
          com.google.protobuf.DurationProto.getDescriptor(),
        });
```

There does not seem to be any managed mode options for Java currently to override this to patch this.

amckinney:
This one is unfortunate - as you've found, Google's `google/type/latlng.proto` clearly doesn't conform to the standard mapping that **Managed Mode** enforces. In this case, we won't be able to override exactly _how_ the `java_outer_classname` option transforms the filename into the class name (since it would affect all other files found in the same `buf generate` invocation).

Instead, I think we'll need to expose a configuration similar to the one described in #465. I imagine we'll have something along the lines of the following:

```yaml
# buf.gen.yaml
version: v1
managed:
  use:
    - DEFAULT
  ignore_only:
    JAVA_OUTER_CLASSNAME:
      - google/type/latlng.proto
```

Alternatively, we _might_ consider an `override` model, so that users can explicitly override individual file options on a per-file basis:

```
# buf.gen.yaml
version: v1
managed:
  use:
    - DEFAULT
  override:
    JAVA_OUTER_CLASSNAME:
      google/type/latlng.proto: com.google.type.LatLngProto
```

I'm not yet convinced that we should support the `override` in general though, so I'd prefer to approach this so that it's symmetrical with the `lint` and `breaking` configuration shapes. Regardless, I'll update this issue once we move forward with a solution to your problem.

odsod:
[...]

PR #506, opened 2021-09-03
Add per file overrides for managed mode

doriable:
This change allows users to set per-file overrides for managed mode modifiers. For example, the user can override the `java_outer_classname` modifier for the `latlng.proto` file with the following config in their `buf.gen.yaml`

```
version: v1
managed:
  enabled: true
  override:
    JAVA_OUTER_CLASSNAME:
       google/type/latlng.proto: com.google.type.LatLngProto
```

Overrides are set on a given modifier based on their ID, and the overrides are provided as a map of proto import path to the override value.

Refs: #471


private/bufpkg/bufimage/bufimagemodify/optimize_for.go
```diff
@@ -21,18 +21,26 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+// OptimizeForID is the ID for the optimize_for modifier.
+const OptimizeForID = "OPTMIZE_FOR"
```
johanbrandhorst, line 25:
```suggestion
const OptimizeForID = "OPTIMIZE_FOR"
```

private/bufpkg/bufimage/bufimagemodify/php_metadata_namespace.go
```diff
@@ -22,17 +22,24 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+// PhpMetadataNamespaceID is the ID of the php_namespace_metadata modifier.
+const PhpMetadataNamespaceID = "PHP_NAMESPACE_METADATA"
```
johanbrandhorst, line 26: This should be
```suggestion
const PhpMetadataNamespaceID = "PHP_METADATA_NAMESPACE"
```
Right?

private/bufpkg/bufimage/bufimagemodify/java_string_check_utf8.go
```diff
@@ -22,14 +22,20 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+// JavaStringCheckUtf8ID is the ID of the java_String_check_utf8 modifier.
```
johanbrandhorst, line 25:
```suggestion
// JavaStringCheckUtf8ID is the ID of the java_string_check_utf8 modifier.
```

johanbrandhorst (changes requested):

private/buf/bufgen/bufgen.go
```diff
@@ -169,6 +169,7 @@ type ManagedConfig struct {
 	JavaPackagePrefix     string
 	OptimizeFor           *descriptorpb.FileOptions_OptimizeMode
 	GoPackagePrefixConfig *GoPackagePrefixConfig
+	Overrides             map[string]map[string]string
```
bufdev, line 172: I would call this `override` personally. We say `except`, not `exceptions`, in our config files. Thoughts? cc @amckinney

private/bufpkg/bufimage/bufimagemodify/php_metadata_namespace.go
doriable: Actually, turns out our documentation needs to be updated and the ID is indeed `php_metadata_namespace`. Updating :D

private/buf/bufgen/bufgen.go
```diff
@@ -254,7 +256,8 @@ func (e ExternalManagedConfigV1) IsEmpty() bool {
 		e.JavaStringCheckUtf8 == nil &&
 		e.JavaPackagePrefix == "" &&
 		e.OptimizeFor == "" &&
-		e.GoPackagePrefix.IsEmpty()
+		e.GoPackagePrefix.IsEmpty() &&
+		e.Overrides == nil
```
bufdev, line 260: `len(e.Overrides) == 0`

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
```diff
@@ -169,34 +197,34 @@ func GoPackageImportPathForFile(imageFile bufimage.ImageFile, importPathPrefix s
 //  * If the resulting abbreviation is 1 character, add "XX".
 //  * If the resulting abbreviation is "GPB", change it to "GPX".
 //    "GPB" is reserved by Google for the Protocol Buffers implementation.
-func ObjcClassPrefix(sweeper Sweeper) Modifier {
-	return objcClassPrefix(sweeper)
+func ObjcClassPrefix(sweeper Sweeper, overrides map[string]string) Modifier {
```
bufdev, line 200: Something feels wrong about passing these as required parameters, I can't pin it though

private/buf/bufgen/bufgen.go
amckinney: Agreed, `override` is also symmetrical with `managed.go_package_prefix.override`.

private/buf/bufgen/config.go
```diff
@@ -215,13 +215,26 @@ func newManagedConfigV1(externalManagedConfig ExternalManagedConfigV1) (*Managed
 	if err != nil {
 		return nil, err
 	}
+	overrides := externalManagedConfig.Overrides
+	for overrideID, override := range overrides {
+		for importPath := range override {
+			if importPath != normalpath.Normalize(importPath) {
```
amckinney, line 221: Does this guard against absolute paths and/or paths that jump context (e.g. `../foo/bar.proto`)? We might need to `NormalizeAndValidate`, then check if the normalized value is equivalent to the configured value.

```go
normalizedImportPath, err := normalpath.NormalizeAndValidate(importPath)
if err != nil {
  return nil, fmt.Errorf(...)
}
if normalizedImportPath != importPath {
  return nil, fmt.Errorf(...)
}
```

We can add tests for these cases, too.

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
amckinney: Yea this one's tough. The alternatives are:
1. Add a `ModifierOption` that is shared amongst all `Modifier` constructors. But as soon as one option should belong to a specific `Modifier` and not others, it breaks down.
2. Add a new `<ModifierName>Option` for each `Modifier`. But this is really verbose.

Given that the `nil` value for `overrides` is actually safe in all of these cases, it feels like the least bad option, especially because it keeps the boilerplate to a minimum. There might be other viable alternatives I'm not seeing though.

private/buf/bufgen/config.go
doriable: So in the current situation, based on `normalpath.Normalize`, it wouldn't guard against a path like `../foo/bar.proto`. I can make a change and add an additional test case to `config_test.go`.

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
```diff
@@ -101,46 +107,68 @@ func GoPackage(
...
+func OptimizeFor(
+	sweeper Sweeper,
+	value descriptorpb.FileOptions_OptimizeMode,
+	overrides map[string]string,
+) (Modifier, error) {
+	validatedOverrides := map[string]descriptorpb.FileOptions_OptimizeMode{}
+	for fileImportPath, overrideString := range overrides {
+		overrideOptimizeMode, err := overrideStringToFileOptionOptimizeMode(overrideString)
+		if err != nil {
+			return nil, fmt.Errorf("invalid override for modifier %s and file %s: %w", OptimizeForID, fileImportPath, err)
```
amckinney, line 167: Even though it's only used in one place, it might be worth making a helper function for the multi-`OptimizeMode` override validation so that it matches the error structure of the others (e.g. `stringOverridesToBoolOverrides`).

That way we can do something like:
```
validatedOverrides, err := stringOverridesToOptimizeModeOverrides(overrides)
if err != nil {
  return nil, fmt.Errorf("invalid override for %s: %w", OptimizeForID, err)
}
```

Otherwise we'll have different error messages for different modifier IDs, which isn't great for the CLI UX.

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
```diff
@@ -219,3 +247,27 @@ func int32SliceIsEqual(x []int32, y []int32) bool {
 	}
 	return true
 }
+
+func stringOverridesToBoolOverrides(stringOverrides map[string]string) (map[string]bool, error) {
+	validatedOverrides := map[string]bool{}
```
amckinney, line 252: We know the size of the map - let's use it.
```suggestion
	validatedOverrides := make(map[string]bool{}, len(stringOverrides))
```

This will apply to the other validation helpers when/if they exist (i.e. the `OptimizeFor` logic above).

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
```diff
@@ -219,3 +247,27 @@ func int32SliceIsEqual(x []int32, y []int32) bool {
...
+func overrideStringToFileOptionOptimizeMode(overrideString string) (descriptorpb.FileOptions_OptimizeMode, error) {
+	switch overrideString {
+	case "SPEED":
+		return descriptorpb.FileOptions_SPEED, nil
+	case "CODE_SIZE":
+		return descriptorpb.FileOptions_CODE_SIZE, nil
+	case "LITE_RUNTIME":
+		return descriptorpb.FileOptions_LITE_RUNTIME, nil
+	}
+	return descriptorpb.FileOptions_SPEED, fmt.Errorf("invalid optimize for value: %s", overrideString)
```
amckinney, line 272: We should be able to use the generated enum association for this. It's a neat trick for a quick validation, and it prevents us from having to update this code when/if any other values are added to `descriptorpb`.
```suggestion
	optimizeMode, ok := descriptorpb.FileOptions_OptimizeMode_value[overrideString]
	if !ok {
		return descriptorpb.FileOptions_SPEED, fmt.Errorf("invalid optimize for value: %s", overrideString)
	}
	return descriptorpb.FileOptions_OptimizeMode(optimizeMode), nil
```

private/bufpkg/bufimage/bufimagemodify/php_namespace_test.go
```diff
@@ -49,14 +49,50 @@ func TestPhpNamespaceEmptyOptions(t *testing.T) {
...
 		assert.Equal(t, testGetImage(t, dirPath, false), image)
 	})
+
+	t.Run("with SourceCodeInfo and per-file overrides", func(t *testing.T) {
+		t.Parallel()
+		image := testGetImage(t, dirPath, true)
+		assertFileOptionSourceCodeInfoEmpty(t, image, phpNamespacePath, true)
+
+		sweeper := NewFileOptionSweeper()
+		phpNamespaceModifier := PhpNamespace(sweeper, map[string]string{"a.proto": "override"})
```
amckinney, line 67: Given that we have `a.proto` in _a lot_ of the tests, let's make sure that we're actually testing whether or not this is _only_ affecting `a.proto` and not others. In this test, we're using the `emptyoptions` directory which only contains a single `a.proto`, so we don't actually know whether or not this file is being overridden while others aren't.

Adding another file in each of these relevant `testdata` directories and asserting that it doesn't adopt the `override` value is sufficient.

amckinney (commented):
Well done, this is looking really great. You were _very_ thorough with testing, too.

Now that we have the tests, let's keep them in (more tests the better), but don't feel obligated to test the `override` path in each of these cases - one area of asserting the `override` behavior for each modifier is sufficient.

private/bufpkg/bufimage/bufimagemodify/bufimagemodify.go
doriable: Nice! I was trying to figure out how to do that, haha :D Thanks!

amckinney (approved):
Nicely done 👏

Issue #482, opened 2021-08-31
Module cache misses for identical references


bufdev:
```
buf generate buf.build/bufbuild/buf --template enterprise/data/template/buf.go-server.gen.yaml
buf generate buf.build/bufbuild/buf --template enterprise/data/template/buf.ts-enterprise.gen.yaml
buf generate enterprise/proto --template enterprise/data/template/buf.go-enterprise.gen.yaml
buf generate enterprise/proto --template enterprise/data/template/buf.go-internal-api.gen.yaml
buf: downloading buf.build/bufbuild/buf:9ab0c4681d664e608a2b9509eaae10a3
buf: downloading buf.build/bufbuild/buf:9ab0c4681d664e608a2b9509eaae10a3
buf: downloading buf.build/bufbuild/buf:e050ba116fa04f7d8d6f5c01e2a83b83
buf: downloading buf.build/bufbuild/buf:e050ba116fa04f7d8d6f5c01e2a83b83
```

The same `ModulePin` was resolved multiple times, so we should only expect to see a single `buf: downloading ...` line per `ModulePin`.

amckinney:
Prior investigation in #456 revealed that this is because the "fetch and put" portion of the cacher can be executed simultaneously by concurrent `buf` processes. We should be able to reuse the same file lock if we expose unsynchronized methods from the [moduleCacher](https://github.com/bufbuild/buf/blob/d25fed60dd48b4490112aeee8b505d137bc8f6c8/private/bufpkg/bufmodule/bufmodulecache/module_reader.go#L30).

PR #514, opened 2021-09-07
Fix module caching

bufdev:
When using file locks with the module cache, we were doing the following:

- ReadLock
- CacheGet
- ReadUnlock
- If not exists or invalid sum:
  - DelegateGet
  - WriteLock
  - CachePut
  - WriteUnlock

This was already causing duplicate downloads because we were doing a DelegateGet if the CacheGet failed without having a write lock, see https://github.com/bufbuild/buf/pull/456. However, we didn't think this was an actual issue beyond duplicate downloads.

The problem is that when inside the ReadLock, another caller could be inside the WriteLock, modifying the underlying data in the cache. If this was happening, we could have inconsistent state for a caller that expects the cache to be stable.

Instead, we move to double locking:

- ReadLock
- CacheGet
- ReadUnlock
- If not exists or invalid sum:
  - WriteLock
  - CacheGet
  - If not exists or invalid sum:
    - DelegateGet
    - CachePut
  - WriteUnlock

By proxy, fixes https://github.com/bufbuild/buf/issues/482.

### DIFF
--- a/private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
+++ b/private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
@@ -50,7 +50,7 @@ func TestReaderBasic(t *testing.T) {
 	require.NoError(t, err)
 
 	delegateDataReadWriteBucket, delegateSumReadWriteBucket, delegateFileLocker := newTestDataSumBucketsAndLocker(t)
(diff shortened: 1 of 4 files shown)


doriable (approved):

amckinney (approved):
This is great - much easier to reason about with all the file locking happening in one layer.

bufdev:
To be fair, I had all the file locking in one layer before, I just got way too cute with it

PR #515, opened 2021-09-07
Only include parameters if non-empty

johanbrandhorst:
Previously we were sending an array with a single
empty element if the options were omitted.

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -242,6 +242,11 @@ func (g *generator) generateConcurrently(
 			if err != nil {
 				return nil, fmt.Errorf("invalid plugin path: %w", err)
 			}
+			var parameters []string
+			if len(pluginConfig.Opt) > 0 {
+				// Only include parameters if they're not empty
+				parameters = []string{pluginConfig.Opt}
+			}
 			// Group remote plugins by remote to execute together.
 			remoteToPlugins[remote] = append(
 				remoteToPlugins[remote],
@@ -250,7 +255,7 @@ func (g *generator) generateConcurrently(
 						Owner:      owner,
 						Name:       name,
 						Version:    version,
-						Parameters: []string{pluginConfig.Opt},
+						Parameters: parameters,
 					},
 					// So we know the order this plugins response should slot in
 					i,


doriable (approved):

bufdev (changes requested):
This seems incorrect - `opt` should be split into multiple strings, not put into a single string as a `[]string` - `Opt` from configs is either a single string (in which case it is one option) or a `[]string` (in which case it is multiple options)

johanbrandhorst:
> This seems incorrect - `opt` should be split into multiple strings, not put into a single string as a `[]string` - `Opt` from configs is either a single string (in which case it is one option) or a `[]string` (in which case it is multiple options)

The config parsing has happened at this point, and it's been merged into a single string (https://github.com/bufbuild/buf/blob/main/private/buf/bufgen/config.go#L139-L142).

private/buf/bufgen/generator.go
```diff
@@ -242,6 +242,11 @@ func (g *generator) generateConcurrently(
 			if err != nil {
 				return nil, fmt.Errorf("invalid plugin path: %w", err)
 			}
+			var parameters []string
+			if len(pluginConfig.Opt) > 0 {
```
bufdev, line 246: `Opt` in this context is either a single option `foo=bar`, or multiple `foo=bar,baz=bat` - what you're doing here is `[]string{"foo=bar,baz=bat"}`, which is incorrect, unless im missing something

bufdev:
> The config parsing has happened at this point, and it's been merged into a single string (https://github.com/bufbuild/buf/blob/main/private/buf/bufgen/config.go#L139-L142).

Then `parameter` should be a single string, not `[]string`

private/buf/bufgen/generator.go
johanbrandhorst: It's technically meaningless as they all get comma-merged before being fed to the plugin as a single string.

```
[]string{"foo=bar", "baz=bat", "bob=bap"}
[]string{"foo=bar,baz=bat", "bob=bap"}
"foo=bar,baz=bat,bob=bap"
```

Are all equivalent.

johanbrandhorst:
> > The config parsing has happened at this point, and it's been merged into a single string (https://github.com/bufbuild/buf/blob/main/private/buf/bufgen/config.go#L139-L142).
> 
> Then `parameter` should be a single string, not `[]string`

We could make this change, but note that template configuration currently uses the `[]string` representation, and that technically it is a pointless distinction (I guess except for the case of an empty string, which is not equivalent to `[]string{""}`). I'd sooner just keep this as it is, the patch here is a necessity for converting the CLI data representation to the (more precise?) API representation.

johanbrandhorst:
I'd like to get this fix in before the release, as it's currently causing an error when not specifying `opt`. Obviously once we've made a release we can't change this API signature as suggested, so there is some time pressure on this decision.

bufdev (approved):

Issue #499, opened 2021-08-31
The log level should control the output of certain messages


johanbrandhorst:
Currently a user who uses `buf` commands that require downloading remote modules will sometimes get a `downloading: blabla` message printed on their STDERR. This can be surprising, and should be configurable with the `--log-level` flag. We should design a consistent way for this information to be passed into the CLI execution functions so other log messages like this can also take advantage of it.

bufdev:
A note: we don't have a `--log-level` flag anymore, we replaced it with `--verbose`, which is just the equivalent of the old `--log-level=debug`.

I'd argue we switch the mechanics around to:

- `--verbose`: controls stderr messages such as `downloading:`. This means gating the existing `downloading:` message behind this flag. I'd propose a new `bufverbose.VerboseLogger` that just has `Log(string, ...args)`, which basically prints to stderr for now.
- `--debug`: is the new `--log-level=debug`.

@amckinney we should probably do this pre-v1.0, and it will be a 15 minute task, marking as P1

PR #516, opened 2021-09-07
Add new verbose package and new --debug flag

bufdev:
Fixes https://github.com/bufbuild/buf/issues/499.

This changes the semantics of `--verbose` to mean "print verbose messages" such as "downloading". In support of this, a new `verbose` package is added, and corresponding logic with `appverbose`. This results in a `verbose.Printer` that is either a no-op, or backed by stderr with the app name as a prefix.

This adds a new `--debug` flag that sets the debug logging level.

Some other logic is changed:

- The unsupported `--log-level` flag is removed in favor of an (also unsupported) `--no-warn` flag, used in testing,.
- The options to `bufmodulecache` are now required, as this is more representative of how we operate with this package (i.e. passing no-op types).

### DIFF
--- a/private/pkg/app/appflag/builder.go
+++ b/private/pkg/app/appflag/builder.go
@@ -16,13 +16,13 @@ package appflag
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"time"
 
 	"github.com/bufbuild/buf/private/pkg/app"
 	"github.com/bufbuild/buf/private/pkg/app/applog"
+	"github.com/bufbuild/buf/private/pkg/app/appverbose"
 	"github.com/bufbuild/buf/private/pkg/observability"
 	"github.com/bufbuild/buf/private/pkg/observability/observabilityzap"
 	"github.com/pkg/profile"
@@ -36,7 +36,8 @@ type builder struct {
 	appName string
 
 	verbose   bool
-	logLevel  string
+	debug     bool
+	noWarn    bool
 	logFormat string
 
 	profile           bool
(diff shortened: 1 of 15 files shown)


private/pkg/app/appflag/builder.go
```diff
@@ -63,13 +64,8 @@ func newBuilder(appName string, options ...BuilderOption) *builder {
...
 func (b *builder) BindRoot(flagSet *pflag.FlagSet) {
-	// True means debug, false means info.
-	flagSet.BoolVarP(&b.verbose, "verbose", "v", false, "Enable verbose logging.")
-	flagSet.StringVar(&b.logLevel, "log-level", "", "The log level [debug,info,warn,error].")
-	// Effectively deprecated - we only ever cared about debug vs info, which is now controlled
-	// by verbose, but we keep this around in case it is being used and handle it being set.
-	// Note that both verbose and log-level cannot be set together.
-	_ = flagSet.MarkHidden("log-level")
+	flagSet.BoolVarP(&b.verbose, "verbose", "v", false, "Turn on verbose mode.")
+	flagSet.BoolVar(&b.debug, "debug", false, "Turn on debug logging.")
```
johanbrandhorst, line 68: How do these interact with each other? Does `debug` imply `verbose`? Worth adding to the help I think.

johanbrandhorst (approved):
Nice!!

amckinney (approved):

private/pkg/app/appflag/builder.go
bufdev: They don't, they're completely separate

PR #517, opened 2021-09-07
Allow --path filter to account for imports

amckinney:
## Overview

This adds the ability for the `--path` flag to recognize import paths. For example, consider the following commands:

**ModuleRefs**
```sh
$ buf build buf.build/acme/petapis --path google/type/money.proto -o -#format=json | jq '.file[] | .name'
"google/type/money.proto"
```

**SourceRefs**
```sh
$ buf build petapis --path acme/payment/v2/payment.proto -o -#format=json | jq '.file[] | .name'
"google/type/money.proto"
"acme/payment/v2/payment.proto"
```


private/bufpkg/bufimage/bufimage.go
```diff
@@ -129,6 +129,13 @@ func NewMultiImage(images ...Image) (Image, error) {
 // be treated as non-imports in the returned Image. The first non-import
 // version of a file will be used in the result.
 //
+// Note that the files within each image are expected to be de-duplicated
+// before being passed into this function. In the case that two images
+// consider the same file as a non-import, the first will be used. We purposefully
+// do not validate this condition here as a result of the --path filter, and
+// instead validate it as part of the ParserAccessorHandler when each Image
+// is constructed.
```
amckinney, line 137: I investigated what it would look like to include this validation in each `ModuleFileSet` independently, but realized that we get it for free in the `ParserAccessorHandler` after the fact.

We _could_ add validation that acts upon all of the `ModuleFileSets` together, but it doesn't buy us anything in practice. Having multiple `ModuleFileSets` is only a detail for workspaces, and every `Module` will already be contained in each of these `ModuleFileSets` (as either the primary module, or a dependency). If each `ModuleFileSet` is valid and de-duplicated within itself, then the same holds for its relationship between the rest.

If we were to go this route, we would probably want to merge `ModuleFileSets` into one, but this is something that we should discuss separately since there are other implications.

amckinney:
I'll need to add more tests here before this is ready to go, but this is working in all of my manual test cases so far.

amckinney:
There's one rare edge case that doesn't work yet for workspace environments:

```yaml
version: v1
directories:
  - paymentapis
  - petapis
```

```sh
$ pwd
/Users/me/.../paymentapis/payment/v1alpha1

$ tree
.
├── buf.work.yaml
├── paymentapis
│   ├── buf.yaml
│   └── payment
│       └── v1alpha1
│           └── payment.proto
├── petapis
    ├── buf.yaml
    └── pet
        └── v1
            └── pet.proto


$ buf build ../../../petapis --path payment.proto
Failure: path "payment.proto" has no matching file in the module
```

The path cannot be determined via `PathForExternalPath` since `payment.proto` is outside the context directory of `../../../petapis`.

We should be able to support this by leveraging the `RelativeRootPath` (the path to the `buf.work.yaml`, if one exists), and transform `payment.proto` into its import path equivalent (based on the `buf.work.yaml`'s `directories` entry).

I'll add this in with a test to accompany it.

private/buf/buffetch/source_ref.go
```diff
@@ -24,22 +24,55 @@ import (
 var _ SourceRef = &sourceRef{}
 
 type sourceRef struct {
-	bucketRef internal.BucketRef
-	dirPath   string
+	bucketRef  internal.BucketRef
+	dirPath    string
```
bufdev, line 28: can we add a comment to these saying what they are? even if duplicative?

private/buf/buffetch/source_ref.go
```diff
@@ -24,22 +24,55 @@ import (
...
 func newSourceRef(bucketRef internal.BucketRef) *sourceRef {
 	var dirPath string
 	if dirRef, ok := bucketRef.(internal.DirRef); ok {
 		dirPath = dirRef.Path()
 	}
+	var subDirPath string
+	if archiveRef, ok := bucketRef.(internal.ArchiveRef); ok {
+		subDirPath = archiveRef.SubDirPath()
+	}
+	if subDirPath == "" {
```
bufdev, line 41: if anything I'd want to error if both are true, not silently say that `ArchiveRef` takes precedence

private/buf/buffetch/source_ref.go
```diff
@@ -24,22 +24,55 @@ import (
...
-		dirPath:   dirPath,
+		bucketRef:  bucketRef,
+		dirPath:    dirPath,
+		subDirPath: subDirPath,
 	}
 }
 
 func (r *sourceRef) PathForExternalPath(externalPath string) (string, error) {
+	if r.subDirPath != "" {
+		path, err := normalpath.NormalizeAndValidate(externalPath)
```
bufdev, line 58: what if externalPath is absolute? is it that if subDirPath is set, this will never be true? this all needs more explanation (including in the code)

private/buf/bufwire/bufwire.go
```diff
@@ -84,6 +87,131 @@ type ModuleConfig interface {
 	Module() bufmodule.Module
 	Config() *bufconfig.Config
 	Workspace() bufmodule.Workspace
+	WorkspaceConfig() *bufwork.Config
+}
+
+// BuildModuleFileSetOptionForTargetPaths returns the bufmodulebuild.BuildModuleFileSetOption
+// required for the given ModuleConfigs. This is required to support the --path filter.
+func BuildModuleFileSetOptionForTargetPaths(
```
bufdev, line 95: This doesn't feel like it should be in `bufwire`, but I'm not sure where it is used yet. The exposing of this feels like it is breaking an abstraction.

We need way more explanation here

bufdev (commented):
gonna start with these comments

private/buf/buffetch/source_ref.go
amckinney: Discussed offline - we should just add a new type and not use `newSourceRef` for the `ArchiveRef` and the `GitRef` types.

private/bufpkg/bufimage/bufimage.go
```diff
@@ -129,8 +129,14 @@ func NewMultiImage(images ...Image) (Image, error) {
...
+//
 // Reorders the ImageFiles to be in DAG order.
-// Duplicates can exist across the Images, but only if duplicates are non-imports.
```
bufdev, line 133: add a comment about duplicates being allowed to exist, also say that formerly non-imports couldnt be duplicated, but why this was removed

amckinney:
Discussed offline -

This feature incurs a lot of room for ambiguity and confusing results. For example, if I run the following command:

```sh
$ buf build --path buf/alpha/registry -o -#format=json | jq '.file[] | .name'
```

It's impossible to distinguish between whether or not the user only wanted to filter on the `buf/alpha/registry` paths found in the current directory (i.e. real filesystem paths - we call these _external paths_), or _all_ import paths matched by the given prefix (i.e. another module in the workspace that might share the same hierarchical import structure).

In this case, we have to return _both_ and it's impossible for us to return one vs. the other (unless we support other flags and/or options).

We're going to close this for now, and only revisit it if it's absolutely necessary. We can refer to the implementation provided here since it's largely what we'll need in the long run (if ever).

PR #518, opened 2021-09-07
Add ability to install buf directly in makego

bufdev:
This allows us to use `buf` from `./cmd/buf` instead of from releases. We can turn this off selectively if there is a bug and we need to run `make generate`. This is similar to what we had in the past.

### DIFF
--- a/make/go/dep_buf.mk
+++ b/make/go/dep_buf.mk
@@ -9,7 +9,24 @@ $(call _assert_var,CACHE_BIN)
 # Settable
 # https://github.com/bufbuild/buf/releases
 BUF_VERSION ?= v0.55.0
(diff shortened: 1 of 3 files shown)


make/go/dep_buf.mk
```diff
@@ -18,4 +35,11 @@ $(BUF):
...
 	@touch $(BUF)
 
+# Use this instead of "buf" when using buf.
+BUF_BIN := $(CACHE_BIN)/buf
+
 dockerdeps:: $(BUF)
+endif
+
+
+
```
johanbrandhorst, line 45: Nit: newlines

johanbrandhorst (approved):

Issue #485, opened 2021-08-31
If a generation template has an override that does not match a module, print a warning or error


bufdev:
Basically we had `buf.build/bufbuild/bufapis` and changed the name to `buf.build/bufteam/corepublic`, and the template had a `go_package` override for `buf.build/bufbuild/bufapis`, but we didn't get warned and almost went on a bug chase. We should either print a warning or error if this happens, I don't know which yet.

@amckinney: Given that `buf generate` can still yield correct results with an unused `override`, a warning might be best. Regardless, the warning or error could be included in `private/bufpkg/bufimage/bufimagemodify/go_package.go` we would just need to determine which overrides were unused by any of the `bufimage.ImageFiles`.

amckinney:
This was _technically_ resolved in #520, but we probably want to carry forward a similar logging approach for the `override` key in general, so I'll continue to use this task to track that work.

PR #520, opened 2021-09-08
Add a warning if a go_package_prefix override is unused

amckinney:
Re: https://github.com/bufbuild/buf/issues/485

This adds a `WARN` log for unused `go_package_prefix.override` values.

For example,

```yaml
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/bufbuild/buf/gen/proto/go
    override:
      buf.build/foo/bar: is/not/used
plugins:
  - name: go
    out: gen/proto/go
    opt: paths=source_relative
```

```sh
$ buf generate
WARN	go_package_prefix override for "buf.build/foo/bar" was unused
```

Note that this type of message _might_ be better suited for the `verbose.Printer` or the `--debug` so that users only see the message when the `--verbose` (`-v`) flag or `--debug` flag is set, respectively. However, it feels more appropriate for a general `WARN` log since the user should always know if these overrides aren't taking effect.

Note that we might want to carry forward a similar change for the `managed.override` values in general, but we can do this in a follow-up.

### DIFF
--- a/private/bufpkg/bufimage/bufimagemodify/go_package.go
+++ b/private/bufpkg/bufimage/bufimagemodify/go_package.go
@@ -81,7 +81,7 @@ func goPackage(
 			}
 			for moduleIdentityString := range overrideModuleIdentityStrings {
 				if _, ok := seenModuleIdentityStrings[moduleIdentityString]; !ok {
-					logger.Sugar().Warnf(`go_package_prefix override for "%s" was unused`, moduleIdentityString)
+					logger.Sugar().Warnf("go_package_prefix override for %q was unused", moduleIdentityString)
 				}
 			}
 			return nil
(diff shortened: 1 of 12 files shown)


private/bufpkg/bufimage/bufimagemodify/go_package.go
```diff
@@ -74,6 +79,11 @@ func goPackage(
 					return err
 				}
 			}
+			for moduleIdentityString := range overrideModuleIdentityStrings {
+				if _, ok := seenModuleIdentityStrings[moduleIdentityString]; !ok {
+					logger.Sugar().Warnf(`go_package_prefix override for "%s" was unused`, moduleIdentityString)
```
johanbrandhorst, line 84: How about using %q here?

johanbrandhorst (approved):

bufdev (approved):

private/bufpkg/bufimage/bufimagemodify/go_package.go
bufdev: Done

PR #523, opened 2021-09-08
Fix go module cache on Windows Github Actions

### LINKED ISSUE #522: Fix Windows go cache for GitHub Actions
I don't think the Windows go caching in GitHub Actions is caching downloaded modules:

https://github.com/bufbuild/buf/pull/521/checks?check_run_id=3546190977 shows a cache size of around 37MB (`cache-go-installs`), and I'd expect it to be a lot bigger. More importantly, the `install-buf` step takes around 40s, and this is largely because it is downloading dependencies.

The Windows side is just caching a `go-build` directory, which I believe is not what you want, I think this has to do with intermediate build steps. What we want to cache is the module source files, which are usually at `$GOPATH/pkg/mod`, whatever the equivalent is for Windows.

doriable:
Change the cache to point at the Go module cache on Windows, fixes #522. 

Screen Shot 2021-09-08 at 12 30 44

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -19,7 +19,7 @@ jobs:
         if: success()
         uses: actions/cache@v2
         with:
-          path: $HOME\go\pkg\mod
+          path: C:\Users\runneradmin\go\pkg\mod
           key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
           restore-keys: |
             ${{ runner.os }}-go-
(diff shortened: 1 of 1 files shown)


amckinney:
This looks good to me - I wonder if it's worth re-running the Windows job a couple times in this PR to determine if the results are actually cached as we expect?

doriable:
> This looks good to me - I wonder if it's worth re-running the Windows job a couple times in this PR to determine if the results are actually cached as we expect?

Haha, that's exactly what I did to get that screenshot -- I just re-ran the job a couple of times and made sure the cache size was reasonable, also noteworthy:
- `install-external-plugins`: ~8-11s
- `install-buf`: ~26-34s

bufdev (approved):

Issue #483, opened 2021-08-31
V1 Readiness


bufdev:
* Review commands (potentially review some that are deprecated and/or unused).
* Consider `--include-imports` and `--exclude-imports` in all `buf` commands.
* Allow `--path` filter to account for imports
* Audit API as a whole and verify it's ready for release
* Release V1 versions of all GitHub Actions.
* Re-consider split between buf config and buf mod. They overlap a lot, and we might want to combine these into one, or clarify their respective rules. Note that moving commands around is relatively straightforward, so we can do this in a backwards-compatible way.
* Audit all commands to be properly using workspaces

Maybe outside the scope of this ticket, but seems close enough to mention. 

We should tag all the GitHub actions as `v1` 

>  ...you can recommend that your users specify a major version when using your action, and only direct them to a more specific version if they encounter issues.

From the [docs](https://docs.github.com/en/actions/creating-actions/about-actions#using-release-management-for-actions)

This is a nicer experience and with a stable `buf` v1 for the CLI we can guarantee a stable experience for those actions. Could we capture that as part of this ticket, open a separate issue, or ignore?

UPDATE: this will also reduce the amount of documentation we need to manage throughout minor/patch releases. We just meed to keep them backwards-compatible.

amckinney:
> We should tag all the GitHub actions as v1

In their current state, the GitHub actions use commands and flags that were deprecated in `v1`.

* `buf-push-action` uses `buf push` ([link](https://github.com/bufbuild/buf-push-action/blob/040008e05aecab7de4f5237a341c16bba68eba42/src/run.ts#L91))
* `buf-lint-action` uses `buf --version` ([link](https://github.com/bufbuild/buf-lint-action/blob/cacc586765be84d485141c084b5bd034973a618d/src/run.ts#L72))
* `buf-breaking-action` uses `buf --version` ([link](https://github.com/bufbuild/buf-breaking-action/blob/320295321f9d91c000d401bede59b3b8b524ce0c/src/run.ts#L104))
* `buf-setup-action` uses `buf --version` ([link](https://github.com/bufbuild/buf-setup-action/blob/main/src/run.ts#L72))

[...]

PR #526, opened 2021-09-08
CLI v1: Move around commands

amckinney:
Re: #483

This proposes a new CLI shape for the `v1` release. This PR (in its current state) is meant to serve as a place for discussion before moving forward (there are in-line `TODO`s for additional context), and tests currently fail (they will be fixed and adjusted once we decide a final form).

For a general overview, some commands are deleted and hidden (via `appcmd.NewDeletedCommand`), whereas others are simply deprecated and hidden from view, but continue to work.

A significant portion of the new command layout can be visualized like so:

```
$ buf --help
Available Commands:
  beta            Beta commands. Unstable and will likely change.
  breaking        Check that the input location has no breaking changes compared to the against location.
  build           Build all files from the input location and output an image.
  config          Interact with the configuration of Buf.
  export          Export the files from the input location.
  generate        Generate stubs for protoc plugins using a template.
  help            Help about any command
  lint            Check that the input location passes lint checks.
  ls-files        List all Protobuf files for the input.
  mod             Configure and update buf modules.
  protoc          High-performance protoc replacement.
  registry        Interact with the Buf Schema Registry.
```

```
$ buf registry --help
Available Commands:
  login       Login to the Buf Schema Registry.
  logout      Logout of the Buf Schema Registry.
```

```
$ buf beta registry --help
Available Commands:
  commit       Repository commit commands.
  organization Organization commands.
  plugin       Plugin commands.
  repository   Repository commands.
  tag          Repository tag commands.
  template     Template commands.
```

```
$ buf mod --help
Available Commands:
  clear-cache Clear the module cache.
  prune       Prunes unused dependencies from the buf.lock file.
  push        Push a module to a registry.
  update      Update the modules dependencies. Updates the buf.lock file.
```

```
$ buf config --help
Available Commands:
  init              Initializes and writes a new buf.yaml configuration file.
  ls-breaking-rules List breaking rules.
  ls-lint-rules     List lint rules.
  migrate-v1beta1   Migrate any v1beta1 configuration files in the directory to the latest version.
```


private/buf/cmd/buf/buf.go
```diff
@@ -66,36 +65,40 @@ import (
...
-	checkBreakingDeprecationMessage = `"buf check breaking" has been moved to "buf breaking", use "buf breaking" instead.
+	loginDeprecationMessage = `"buf login" has been moved to "buf registry login".
 We recommend migrating, however this command continues to work.
 See https://docs.buf.build/faq for more details.`
-	checkLintDeprecationMessage = `"buf check lint" has been moved to "buf lint", use "buf lint" instead.
+	logoutDeprecationMessage = `"buf logout" has been moved to "buf registry logout".
 We recommend migrating, however this command continues to work.
 See https://docs.buf.build/faq for more details.`
-	checkLSBreakingCheckersDeprecationMessage = `"buf check ls-breaking-checkers" has been moved to "buf config ls-breaking-rules", use "buf config ls-breaking-rules" instead.
+	pushDeprecationMessage = `"buf push" has been moved to "buf mod push".
```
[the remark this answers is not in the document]
amckinney: There are a few reasons for us to consider for this one.

All other module-related commands are encapsulated in the `mod` or `registry` sub-commands, whereas the `push` command sits awkwardly at the top-level. Pushing it into the `mod` command brings it closer to other similar operations like `buf mod update`.

The push command is _primarily_ meant to be executed in CI systems, and _occasionally_ by hand (i.e. if/when we introduce branches, this will be a more common case for the manual case). Moving it down a single level into `mod` will hardly be more difficult to use in the cases when people are running it manually.

Unlike some of the other commands, this change continues to support `buf push` at the top-level, but suggests `buf mod push` (and hides `buf push` from the `--help` text). I'd argue it's a move worth considering, especially given that all of module and registry commands exist elsewhere. I'm not totally convinced this is the right move just yet, but that's the rationale I have so far.

private/buf/cmd/buf/buf.go
```diff
@@ -186,31 +197,29 @@ func NewRootCommand(name string) *appcmd.Command {
...
+							appcmd.NewDeletedCommand("clear-cache", betaModClearCacheDeprecationMessage, "cc"),
 						},
 					},
 					{
 						Use:   "registry",
 						Short: "Interact with the Buf Schema Registry.",
 						SubCommands: []*appcmd.Command{
+							// TODO(alex): Move these packages into registry/{login,logout}.
+							login.NewCommand("login", builder, "", false),
+							logout.NewCommand("logout", builder, "", false),
```
amckinney, line 222: The justification here is that we should aggressively protect the top-level command space (to prevent ourselves from being a `docker`). It's much easier to hoist a sub-command into the top-level than the reverse.

Here, we can actually move `login` and `logout` to `buf registry {login,logout}` rather than starting them in `buf beta registry {login,logout}`. Then, as we move out the rest of the `buf beta registry ...` commands, they will exist alongside `buf registry {login,logout}`.

`buf registry` should be enough of a top-level command to garner attention.

private/buf/cmd/buf/buf.go
```diff
@@ -338,7 +342,7 @@ See https://docs.buf.build/faq for more details.`,
 					},
 				},
 			},
-			push.NewCommand("push", builder, "", false),
+			push.NewCommand("push", builder, pushDeprecationMessage, true),
```
amckinney, line 345: If we don't delete the commands altogether, then we have to maintain them forever. If we are going to move them now, we need to use `appcmd.NewDeletedCommand`, or don't delete them at all.

private/buf/cmd/buf/buf.go
```diff
@@ -150,23 +148,36 @@ func NewRootCommand(name string) *appcmd.Command {
 				Use:   "mod",
 				Short: "Configure and update buf modules.",
 				SubCommands: []*appcmd.Command{
-					modinit.NewCommand("init", builder, "", false),
+					modinit.NewCommand("init", builder, modInitDeprecationMessage, true), // TODO(alex): Think about this one (see TODO below).
+					export.NewCommand("export", builder, "", false),                      // TODO(alex): Move this command into the modexport package.
```
amckinney, line 152: The `export` command is agnostic to all inputs, not just modules. We should either:
1. Arbitrarily limit this to modules.
2. Keep it in the top-level namespace.

private/buf/cmd/buf/buf.go
```diff
@@ -150,23 +148,36 @@ func NewRootCommand(name string) *appcmd.Command {
 				Use:   "mod",
 				Short: "Configure and update buf modules.",
 				SubCommands: []*appcmd.Command{
-					modinit.NewCommand("init", builder, "", false),
+					modinit.NewCommand("init", builder, modInitDeprecationMessage, true), // TODO(alex): Think about this one (see TODO below).
```
amckinney, line 151: `buf mod init` could be moved to `buf config init` and clearly state itself as "The behavior of this command is interactive, and is subject to change". That way, it can also support `buf.gen.yaml`, `buf.work.yaml`, or any combination of the three (or more later).

private/buf/cmd/buf/buf.go
```diff
@@ -150,23 +148,36 @@ func NewRootCommand(name string) *appcmd.Command {
...
+			//
+			// How do we foresee future configuration commands for `buf.work.yaml` and `buf.gen.yaml` to look? Will these be new
+			// sub-commands under the "config" sub-command? What does that suggest about how we should structure this (i.e. "buf config mod-init")?
 			{
 				Use:   "config",
 				Short: "Interact with the configuration of Buf.",
 				SubCommands: []*appcmd.Command{
+					modinit.NewCommand("init", builder, "", false), // TODO(alex): Move this command into the configinit package.
 					configlslintrules.NewCommand("ls-lint-rules", builder, "", false),
 					configlsbreakingrules.NewCommand("ls-breaking-rules", builder, "", false),
```
amckinney, line 175: We can keep the `ls-{breaking,lint}-rules` commands as-is, and move them into the `buf mod` sub-command later on if it makes more sense. In short, we're fine with the cost of deprecating these commands and hiding them later on, and this doesn't need to be part of the v1 release.

private/buf/cmd/buf/buf.go
```diff
@@ -150,23 +148,36 @@ func NewRootCommand(name string) *appcmd.Command {
...
+			// TODO(alex): We might actually want to keep "buf mod init" in the "mod" sub-command since it's directly tied to
+			// the other "mod" commands. For example, I initialize a module with "buf mod init", then push it immediately after
+			// with "buf mod push". Otherwise, we would need to jump between multiple sub-commands (initialize with "config",
+			// then push with "mod").
+			//
+			// However, keeping it in mod is awkward for "config ls-{breaking,lint}-rules" - listing the content of the `buf.yaml`
+			// should exist alongside how it's initialized.
+			//
+			// How do we foresee future configuration commands for `buf.work.yaml` and `buf.gen.yaml` to look? Will these be new
+			// sub-commands under the "config" sub-command? What does that suggest about how we should structure this (i.e. "buf config mod-init")?
```
amckinney, line 168: This is related to the comment above - we eventually want `buf config init` to be more interactive and support bootstrapping the user's entire `buf` filesystem layout. For example, providing a wizard that asks the user where their `.proto` roots are, and creating a `buf.work.yaml`, `buf.yaml`, and `buf.gen.yaml` in the appropriate places.

private/buf/cmd/buf/buf.go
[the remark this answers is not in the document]
johanbrandhorst: That's reasonable, I think the CI argument is the strongest for me right now

private/buf/cmd/buf/buf.go
amckinney: As I've thought about this one a little bit more, it might actually make sense to leave `login` and `logout` at the top-level. Similar to our discussion around `buf config init` being a general configuration command that will (eventually) apply to `buf.gen.yaml`, `buf.work.yaml` and others, the same could be said about `buf login`.

The `buf login` command already authenticates users into _both_ the BSR and the Go Module Proxy. As we roll out more server-side components that require authentication, the list of things that `buf login` could be responsible for will grow.

These components might have their own authentication mechanisms, so it might make sense to keep each `login` command isolated in their own respective sub-commands (e.g. `buf registry login`, etc.), but I'm not sure how likely that is - it would end up being a confusing user experience to say the least (i.e. "THIS token is for this, and THAT token is for this other thing").

Anyway, this is just some additional food for thought. As my original comment suggests, it's a lot easier to hoist a sub-command into the top-level than the reverse, so we can start with `buf registry {login,logout}` and adapt later if need be.

private/buf/cmd/buf/buf.go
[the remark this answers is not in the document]
doriable: I think it is important to show that `push` is an action that only pertains to modules, and therefore should live with other `mod` commands. It's worth noting that `push` can take an "arbitrary" input, which may be a remote module/state (e.g. a git ref) and is specifically pushing a new state to the BSR (the centralised remote state of the module). I guess the key point here is that the entity that is being referenced by the `push` command will always be a module, so it is important to keep that bundled with `mod` to prevent exploding out the top level commands for `buf`.

private/buf/cmd/buf/buf.go
doriable: I think that option 2 seems the most reasonable. I don't think it makes a lot of sense to go with 1, changing the nature of that functionality seems somewhat unreasonable and unwarranted, and if the entity that is being interacted with is not limited to modules, it should not be a `mod` level command.

private/buf/cmd/buf/buf.go
```diff
@@ -289,56 +274,29 @@ func NewRootCommand(name string) *appcmd.Command {
...
-							{
-								Use:   "token",
-								Short: "Token commands.",
-								SubCommands: []*appcmd.Command{
-									tokencreate.NewCommand("create", builder),
-									tokenget.NewCommand("get", builder),
-									tokenlist.NewCommand("list", builder),
-									tokendelete.NewCommand("delete", builder),
-								},
-							},
```
amckinney, line 336: I figure we may as well clean these up along the way - we don't have any plans for them in the near term, and we can keep each of the command implementations around if we want, but these aren't providing any value as-is. I'm happy to revert this if we feel strongly about keeping them in.

private/buf/cmd/buf/command/config/configinit/configinit.go
```diff
@@ -90,6 +90,11 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		".",
 		`The directory to write the configuration file to.`,
 	)
+	// TODO(alex): Now that 'buf config init' is built to (eventually) be general to all buf
+	// configuration files, should we still support the module name and dependencies flags?
+	//
+	// We might be able to avoid these compatibility concerns if we pitch this command as one
+	// that is subject to change over time (as discussed).
```
amckinney, line 97: Let's discuss whether or not we should remove the `buf config init --name` flag. The only reason I call it out is if/when we extend `buf config init` to other configuration files, then the semantics of `--name` might get confusing (i.e. if multiple configurations had their own concept of `name`).

private/buf/cmd/buf/buf.go
amckinney: Discussed offline - starting with `buf registry {login,logout}` is best for now given all the context above.

private/buf/cmd/buf/command/config/configinit/configinit.go
amckinney: Discussed offline - we'll drop these flags in a follow-up. Some tests will need to be adjusted.

CHANGELOG.md
```diff
@@ -2,7 +2,30 @@
 
 ## [Unreleased]
 
-Complete deletion `protoc-gen-buf-check-breaking` and `protoc-gen-buf-check-lint`, which have been moved to `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
```
bufdev, line 5: I'm gonna want to clean this up even further but that's fine, just as long as everything is noted down.

bufdev:
Commandeering

amckinney:
Merging as-is; we'll follow up with other changes to the `CHANGELOG.md` before release.

PR #533, opened 2021-09-09
Add directory argument for `buf mod update` and `buf mod prune`

doriable:
This adds a directory argument that defaults to `.` to `buf mod update` and `buf mod prune`.
Support for taking a directory already existed, so this primarily exposes that through an argument.

### DIFF
--- a/private/buf/cmd/buf/command/mod/modprune/modprune.go
+++ b/private/buf/cmd/buf/command/mod/modprune/modprune.go
@@ -28,11 +28,6 @@ import (
 	"github.com/bufbuild/buf/private/pkg/rpc"
 	"github.com/bufbuild/buf/private/pkg/storage/storageos"
 	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
-)
-
-const (
-	dirFlagName = "dir"
 )
 
 // NewCommand returns a new prune Command.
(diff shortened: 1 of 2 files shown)


private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -51,8 +50,8 @@ func NewCommand(
 		Long: "Gets the latest digests for the specified references in the config file, " +
 			"and writes them and their transitive dependencies to the " +
 			buflock.ExternalConfigFilePath +
-			" file.",
-		Args:       cobra.NoArgs,
+			` file. The first argument is the directory of the local module to update. If no argument is specified, defaults to "."`,
+		Args:       cobra.MaximumNArgs(1),
```
amckinney, line 54: Don't forget to update the `Use` value above:

```go
Use: name + " <directory>",
```

private/buf/cmd/buf/command/mod/modprune/modprune.go
```diff
@@ -28,61 +28,39 @@ import (
...
 
 // NewCommand returns a new prune Command.
 func NewCommand(
 	name string,
 	builder appflag.Builder,
 ) *appcmd.Command {
-	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name,
+		Use:   name + " <input>",
```
amckinney, line 39: We should probably use `<directory>` so that it's clearly distinguished from the generic `<input>` supported by `buf build` and others.

doriable: Oh yeah, good call out -- you're right, this should only be for local state :D

amckinney (approved):

private/buf/cmd/buf/command/mod/modprune/modprune.go
```diff
@@ -28,61 +28,39 @@ import (
...
-	_ = flagSet.MarkHidden(dirFlagName)
-}
-
 // run tidy to trim the buf.lock file for a specific module.
 func run(
 	ctx context.Context,
 	container appflag.Container,
-	flags *flags,
 ) error {
+	directoryInput, err := bufcli.GetInputValue(container, "", "", "", ".")
```
bufdev, line 57: We actually needed to have `flags.InputHashtag` and then pass it here, basically cobra ignores anything after `-#`. While we only take directories, we probably want to future-proof this. See `push.go` for an example.

Issue #535, opened 2021-09-09
Update `--version` to write to `stdout`


amckinney:
We've received a lot of user complaints about the result of `buf --version` being written to `stderr` instead of `stdout`. Given that this is a breaking change, we should change this behavior before V1.

PR #538, opened 2021-09-10
Move help to stdout and --version to version

bufdev:
Fixes https://github.com/bufbuild/buf/issues/535.

This moves `-h, --help, help` to stdout, and `--version` to `version, while preserving usage errors to stderr.

### DIFF
--- a/private/pkg/app/appcmd/cobra.go
+++ b/private/pkg/app/appcmd/cobra.go
@@ -1,0 +1,41 @@
+package appcmd
+
+import (
+	"fmt"
+	"io"
+	"strings"
+	"text/template"
+	"unicode"
+
+	"github.com/spf13/cobra"
+)
+
+// The functions in this file are mostly copied from github.com/spf13/cobra.
+// https://github.com/spf13/cobra/blob/master/LICENSE.txt
+
+var templateFuncs = template.FuncMap{
+	"trim":                    strings.TrimSpace,
+	"trimRightSpace":          trimRightSpace,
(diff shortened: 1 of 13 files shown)


bufdev:
This is actually entirely bad, because cobra prints a lot of stuff to the writer that you set with `SetOut`, and if we change that to `container.Stdout()`, then we can't have a bunch of error messages from cobra printed to stderr like we want. We'll have to change this around to have `container.Stderr()` remain the value for `SetOut` as it is now, and then specially override `--version` and `--help` to use `container.Stdout()`.

doriable (approved):

private/pkg/app/appcmd/cobra.go
```diff
@@ -0,0 +1,55 @@
...
+func rpad(s string, padding int) string {
+	template := fmt.Sprintf("%%-%ds", padding)
+	return fmt.Sprintf(template, s)
+}
+
+// tmpl executes the given template text on data, writing the result to w.
+func tmpl(w io.Writer, text string, data interface{}) error {
+	t := template.New("top")
+	t.Funcs(templateFuncs)
+	template.Must(t.Parse(text))
```
amckinney, line 53: We can drop `template.Must` since we're already returning the `error` from the `tmpl` signature. That'll prevent us from a potential panic.

amckinney (approved):
Minus the `template.Must` nit, this looks good.

private/pkg/app/appcmd/cobra.go
bufdev: done

PR #544, opened 2021-09-14
Add WorkspaceBuilder to improve performance

amckinney:
This adds a `bufwork.WorkspaceBuilder` that caches modules when building a `bufmodule.Workspace`.

As you can see from the resolved `TODO`, there was a lot of duplicative work happening when constructing a workspace - we would build _each module for every directory_, which has O(n^2) complexity. With the `bufwork.WorkspaceBuilder`, we only build each module once.

With this change, the  previous`bufwork.NewWorkspace` implementation is refactored as `bufwork.WorkspaceBuilder.BuildWorkspace`, and a simpler `bufmodule.NewWorkspace` constructor is introduced so that the constructor exists alongside the `bufmodule.Workspace` definition (like all other types).

There is another optimization we can consider that consolidates `bufmodule.Modules` into a single `bufmodule.ModuleFileSet` before creating a single `bufimage.Image` (instead of multiple `bufimage.Images`). However, there are some other complications in this approach that will need to be investigated, so it's left for a follow-up (if ever).

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -351,15 +351,13 @@ func NewWireImageConfigReader(
 	if err != nil {
 		return nil, err
 	}
-	configProvider := bufconfig.NewProvider(logger)
-	moduleBucketBuilder := bufmodulebuild.NewModuleBucketBuilder(logger)
 	return bufwire.NewImageConfigReader(
 		logger,
 		storageosProvider,
 		newFetchReader(logger, storageosProvider, moduleResolver, moduleReader),
-		configProvider,
+		bufconfig.NewProvider(logger),
 		bufwork.NewProvider(logger),
-		moduleBucketBuilder,
+		bufmodulebuild.NewModuleBucketBuilder(logger),
 		bufmodulebuild.NewModuleFileSetBuilder(logger, moduleReader),
 		bufimagebuild.NewBuilder(logger),
 	), nil
@@ -377,15 +375,13 @@ func NewWireModuleConfigReader(
 	if err != nil {
 		return nil, err
 	}
-	configProvider := bufconfig.NewProvider(logger)
-	moduleBucketBuilder := bufmodulebuild.NewModuleBucketBuilder(logger)
 	return bufwire.NewModuleConfigReader(
 		logger,
 		storageosProvider,
 		newFetchReader(logger, storageosProvider, moduleResolver, moduleReader),
-		configProvider,
+		bufconfig.NewProvider(logger),
 		bufwork.NewProvider(logger),
-		moduleBucketBuilder,
+		bufmodulebuild.NewModuleBucketBuilder(logger),
 	), nil
 }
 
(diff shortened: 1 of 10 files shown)


private/buf/bufwork/bufwork.go
```diff
@@ -94,31 +94,31 @@ var (
...
+	// BuildWorkspace builds a bufmodule.Workspace for the given targetSubDirPath.
+	BuildWorkspace(
+		ctx context.Context,
+		workspaceConfig *Config,
+		readBucket storage.ReadBucket,
+		relativeRootPath string,
+		targetSubDirPath string,
+		configOverride string,
+		externalDirOrFilePaths []string,
+		externalDirOrFilePathsAllowNotExist bool,
```
amckinney, line 108: If we feel strongly about these positional parameters, we could roll out an `BuildWorkspaceOption` for each one (e.g. `BuildWorkspaceWithConfigOverride`, etc.).

private/buf/bufwork/workspace_builder.go
```diff
@@ -23,29 +23,38 @@ import (
...
 )
 
-type workspace struct {
-	// bufmoduleref.ModuleIdentity -> bufmodule.Module
-	namedModules map[string]bufmodule.Module
-	allModules   []bufmodule.Module
+type workspaceBuilder struct {
+	configProvider      bufconfig.Provider
+	moduleBucketBuilder bufmodulebuild.ModuleBucketBuilder
+	moduleCache         map[string]*cachedModule
```
bufdev, line 33: The `WorkspaceBuilder` seems to live around across invocations of `bufwire` functions, which means this cache is long-lived. Given that we never clear this cache, in a service context, this could eventually result in running out of memory. Two options:

1. Make this a LRU cache.
2. (probably more preferred) make `workspaceBuilder` something constructed for a single `GetImageConfig/GetModuleConfig` call.

private/buf/bufwork/bufwork.go
doriable: I think that it makes sense to take an options approach, especially given the nature of these arguments. I am a big +1 on that, but I don't think it blocks this change either.

doriable (approved):

private/buf/bufcli/bufcli.go
```diff
@@ -351,13 +351,15 @@ func NewWireImageConfigReader(
 	if err != nil {
 		return nil, err
 	}
+	configProvider := bufconfig.NewProvider(logger)
```
bufdev, line 354: nit: we should revert these being split out into parameters

private/buf/bufcli/bufcli.go
```diff
@@ -375,13 +377,15 @@ func NewWireModuleConfigReader(
...
```
bufdev, line 380: nit: we should revert these being split out into parameters

private/buf/bufcli/bufcli.go
```diff
@@ -395,13 +399,15 @@ func NewWireModuleConfigReaderForModuleReader(
 ) (bufwire.ModuleConfigReader, error) {
 	logger := container.Logger()
 	moduleResolver := bufapimodule.NewModuleResolver(logger, registryProvider)
+	configProvider := bufconfig.NewProvider(logger)
```
bufdev, line 402: nit: we should revert these being split out into parameters

private/buf/bufwork/bufwork.go
```diff
@@ -94,31 +94,31 @@ var (
 	}
 )
 
-// NewWorkspace returns a new workspace.
-func NewWorkspace(
-	ctx context.Context,
-	config *Config,
-	readBucket storage.ReadBucket,
+// WorkspaceBuilder builds workspaces.
```
bufdev, line 97: Add comment about how this should not be persisted long-term due to the cache

private/buf/bufwork/bufwork.go
bufdev: Eh yea this is a historical relic, but also would be fine if we kept for now and dealt with this in `bufwire` at the same time

private/buf/bufcli/bufcli.go
amckinney: Gah, yea I rushed these. Will update.

private/buf/bufwork/bufwork.go
amckinney: Sounds good, keeping as-is for now.

bufdev (approved):

Issue #490, opened 2021-08-31
Prepend imported module file paths with module path in errors


bufdev:
```
$ buf check breaking --against-input 0.0.0.0:3000/foo/bar/v1:b1-659yVmM4Ut_GZa6Nr-rJW2jT0hmbupCPNFyYr5R28aU=
buf/image/v1/image.proto exists in multiple locations: proto/buf/image/v1/image.proto buf/image/v1/image.proto
```

This actually isn't a bug - this was due to having buf.yaml for bufbuild rely on a pushed module of itself, so the file was in two locations. The error message was just crazy confusing.

This likely gets to my initial intent of external paths: prepend file paths for modules with the module path.

TheDrewscriver:
I would say this is a major issue. This is the problem I am facing:

Even thought the buf.yaml only includes this:
deps:
  - buf.build/beta/googleapis
  - buf.build/grpc-ecosystem/grpc-gateway

When I run 
```
buf mod update -v
DEBUG	get_config	{"duration": "459.9µs"}
DEBUG	grpcclient	dial	{"address": "dns:///api.buf.build"}
DEBUG	buf.alpha.registry.v1alpha1.ResolveService.GetModulePins	{"duration": "624.824732ms", "Client": true, "FailFast": true}
DEBUG	command	{"duration": "730.86471ms"}
```
It adds dependencies for both buf.build/beta and buf.build/googleapis

Contents on buf.lock:
```

# Generated by buf. DO NOT EDIT.
version: v1
deps:
  - remote: buf.build
    owner: beta
    repository: googleapis
    branch: main
    commit: ac0d624de3f746c3b8ef55c74b6f5244
    digest: b1-GubXb4hZ56wTohZYDL-Muute9Jxx8J4TtIYucS30TBM=
    create_time: 2021-08-12T19:53:55.128813Z
  - remote: buf.build
    owner: googleapis
    repository: googleapis
    branch: main
    commit: 6358935a97044c1494fe13d369ac2f3b
    digest: b1-4hxgpEmLuajuxqkZd7NMVnKipt3zqnXaTHsBEKtUOsM=
    create_time: 2021-08-13T15:04:04.653816Z
  - remote: buf.build
    owner: grpc-ecosystem
    repository: grpc-gateway
    branch: main
    commit: d1922677830c448eb65ec590f3abc734
    digest: b1-zmmsl0XlsSF4_BhZe7w9NMdImBCRtMSn68rmn9H1faQ=
[...]

PR #545, opened 2021-09-14
Check for duplicated file paths during module file set construction

doriable:
Based on discussions in #490, when file paths are duplicated across modules, such as in the following example, `pet/v1/pet.proto` exists in both the `petapis` and `paymentapis` modules, an error message without module prefixes is printed out.

```
.
├── buf.gen.yaml
├── buf.work.yaml
├── paymentapis
│   ├── buf.lock
│   ├── buf.yaml
│   ├── payment
│   │   └── v1alpha1
│   │       └── payment.proto
│   └── pet
│       └── v1
│           └── pet.proto
└── petapis
    ├── buf.lock
    ├── buf.md
    ├── buf.yaml
    └── pet
        └── v1
            └── pet.proto
```

We can solve this by catching the duplicated path at module file set build time and returning the same error, but with a module identity string prepended when available:

```
$ buf build
Failure: pet/v1/pet.proto exists in multiple locations: paymentapis/pet/v1/pet.proto buf.build/acme/petapis#petapis/pet/v1/pet.proto.
```

### DIFF
--- a/private/bufpkg/bufmodule/module_file_set.go
+++ b/private/bufpkg/bufmodule/module_file_set.go
@@ -32,7 +32,7 @@ type moduleFileSet struct {
 func newModuleFileSet(
 	module Module,
 	dependencies []Module,
-) *moduleFileSet {
+) (*moduleFileSet, error) {
 	// TODO: We can remove the getModuleRef method on the
 	// Module type if we fetch FileInfos from the Module
 	// and plumb in the ModuleRef here.
@@ -57,10 +57,27 @@ func newModuleFileSet(
 			),
 		)
 	}
+	multiReadBucket := newMultiModuleReadBucket(moduleReadBuckets...)
+	seenPaths := map[string][]string{}
+	if err := multiReadBucket.WalkModuleFiles(context.Background(), "", func(moduleObjectInfo *moduleObjectInfo) error {
+		modulePrefixedPaths, ok := seenPaths[moduleObjectInfo.Path()]
+		modulePrefixedPath := moduleObjectInfo.ExternalPath()
+		if moduleObjectInfo.ModuleIdentity() != nil {
+			modulePrefixedPath = moduleObjectInfo.ModuleIdentity().IdentityString() + "#" + modulePrefixedPath
+		}
+		modulePrefixedPaths = append(modulePrefixedPaths, modulePrefixedPath)
+		if ok {
+			return storage.NewErrExistsMultipleLocations(moduleObjectInfo.Path(), modulePrefixedPaths...)
+		}
+		seenPaths[moduleObjectInfo.Path()] = modulePrefixedPaths
+		return nil
+	}); err != nil {
+		return nil, err
+	}
 	return &moduleFileSet{
 		Module:              module,
-		allModuleReadBucket: newMultiModuleReadBucket(moduleReadBuckets...),
-	}
+		allModuleReadBucket: multiReadBucket,
+	}, nil
 }
 
 func (m *moduleFileSet) AllFileInfos(ctx context.Context) ([]bufmoduleref.FileInfo, error) {
--- a/private/bufpkg/bufmodule/bufmodule.go
+++ b/private/bufpkg/bufmodule/bufmodule.go
@@ -217,7 +217,7 @@ type ModuleFileSet interface {
 func NewModuleFileSet(
 	module Module,
 	dependencies []Module,
-) ModuleFileSet {
+) (ModuleFileSet, error) {
 	return newModuleFileSet(module, dependencies)
 }
 
--- a/private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
+++ b/private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
@@ -85,5 +85,5 @@ func (m *moduleFileSetBuilder) build(
 		}
 		dependencyModules = append(dependencyModules, dependencyModule)
 	}
-	return bufmodule.NewModuleFileSet(module, dependencyModules), nil
+	return bufmodule.NewModuleFileSet(module, dependencyModules)
 }


private/bufpkg/bufmodule/module_file_set.go
```diff
@@ -57,10 +57,27 @@ func newModuleFileSet(
...
+		modulePrefixedPath := moduleObjectInfo.ExternalPath()
+		if moduleObjectInfo.ModuleIdentity() != nil {
+			modulePrefixedPath = moduleObjectInfo.ModuleIdentity().IdentityString() + "#" + modulePrefixedPath
+		}
+		modulePrefixedPaths = append(modulePrefixedPaths, modulePrefixedPath)
+		if ok {
+			return storage.NewErrExistsMultipleLocations(moduleObjectInfo.Path(), modulePrefixedPaths...)
+		}
+		seenPaths[moduleObjectInfo.Path()] = modulePrefixedPaths
+		return nil
```
amckinney, line 73: Whenever possible, it's best to check the `ok` immediately after it's created. We can reorganize this a bit in its current state, but how we do might depend on the comment above.

private/bufpkg/bufmodule/module_file_set.go
```diff
@@ -57,10 +57,27 @@ func newModuleFileSet(
 			),
 		)
 	}
+	multiReadBucket := newMultiModuleReadBucket(moduleReadBuckets...)
+	seenPaths := map[string][]string{}
+	if err := multiReadBucket.WalkModuleFiles(context.Background(), "", func(moduleObjectInfo *moduleObjectInfo) error {
```
bufdev, line 62: This feels *really* expensive to do on construction of a `ModuleFileSet` every time - this could be 1000's of files just to change an error message

private/bufpkg/bufmodule/module_file_set.go
```diff
@@ -57,10 +57,27 @@ func newModuleFileSet(
...
+		modulePrefixedPaths, ok := seenPaths[moduleObjectInfo.Path()]
```
amckinney, line 63: In this case, we only ever return an error that reports two duplicate files. This could get frustrating if there are several duplicates: the user has to fix one at a time, re-run the command, and notice others that need to be fixed.

We can restructure this as collecting all of the `seenPaths`, then report _all_ of the paths that collide with a single path. We want the error message to be deterministic, so we'll need to iterate over the map by either walking over all the files again (in a second pass), or by sorting the keys and walking over them in that order.

We might even want to report _all_ instances of `Path` -> `ExternalPath` duplicates. But the main case I can think of if for when users depend on both `buf.build/beta/googleapis` and `buf.build/googleapis/googleapis` - in those cases the number of duplicates would be huge and the user would see a wall of text (which isn't helpful at all).

With that said, I think reporting all duplicates of a single file in a deterministic manner is sufficient. So we'd have errors like (notice that we can also format this as an `[...]`:

```
$ buf build
Failure: pet/v1/pet.proto exists in multiple locations: [paymentapis/pet/v1/pet.proto, buf.build/acme/petapis/petapis/pet/v1/pet.proto, buf.build/acme/something/else/pet/v1/pet.proto].
```

What are your thoughts?

private/bufpkg/bufmodule/module_file_set.go
```diff
@@ -57,10 +57,27 @@ func newModuleFileSet(
...
+		modulePrefixedPath := moduleObjectInfo.ExternalPath()
+		if moduleObjectInfo.ModuleIdentity() != nil {
+			modulePrefixedPath = moduleObjectInfo.ModuleIdentity().IdentityString() + "#" + modulePrefixedPath
```
amckinney, line 66: Let's discuss whether or not we should use a `#` here. With fully-qualified import paths in mind, we could probably just use `normalpath.Join` here so that the files are reported like they are in the BSR (e.g. https://buf.build/googleapis/googleapis/file/main/google/actions/type/date_range.proto minus the `file/main` components).

private/bufpkg/bufmodule/module_file_set.go
doriable: Yeah, that's very fair... the other option that @amckinney and I explored is extending the `ObjectInfo` interface to take in some metadata and return a "fully-qualified path" (in addition to `Path()` and `ExternalPath()`), so we don't affect the output of the existing implementations. That is probably the other reasonable route.

amckinney: Another way to think of the `ObjectInfo` extension is supporting another `interface` that could be up-casted to, but it's really the same thing e.g.

```go
 // I don't know what to call this right now.
type SomethingObjectInfo interface {
  ObjectInfo

  // FullyQualifiedPath returns a string that contains the fully-qualified path
  // for this object. If set, this will defer from the ExternalPath representation.
  FullyQualifiedPath() string
}
```

Then you could check for its existence:

```go
if v, ok := objectInfo.(SomethingObjectInfo); ok {
  ...
}
```

This isn't great though since it effects the `storage` APIs and it's unclear how this would be used by other `storage` implementations since we already have `ExternalPath` which _should_ represent what `FullyQualifiedPath` does in this specific case.

As for validating this for every `ModuleFileSet`, this operation is already happening: it's just being brought up a layer. We'll end up iterating over each file one more time than we previously did, but it's still a O(n) operation. Given that a ModuleFileSet should really only be in the ~1000s of files, this probably isn't even noticeable in practice - we're just walking over each file. It also feels natural for the `ModuleFileSet` to validate itself this way. With this, we can make other guarantees about the validity of the `ModuleFileSet`, which could be (potentially) leveraged elsewhere.

I could be wrong w.r.t the cost here, but that's my high-level take.

amckinney: With all that said, there _are_ implications around this cost in workspace environments. GIven that we currently build a different `ModuleFileSet` for each workspace directory, there's a lot of duplicative effort here. This ties back into some of the thoughts I have in the summary of #544, however.

private/bufpkg/bufmodule/module_file_set.go
doriable: Hmm... this brings up an interesting point... and you're right, we probably want to avoid the "`beta/googleapis + googleapis/googleapis` wall of text" situation, so +1 on reporting everything deterministically for duplicates. Let me implement that change.

private/bufpkg/bufmodule/module_file_set.go
doriable: Hmm, yeah, that's fair, I suppose `normalpath` is reasonable here since we want to represent the... well normalised path. My only fear with normalpath is that I don't want the path structure to confuse users that it's not necessarily a path on disk, but it should be fine. At least in general this check/change makes duplicated paths easier to debug.

private/bufpkg/bufmodule/module_file_set.go
doriable: Sounds good, I'll keep this in mind as I restructure this based on the comment above :)

doriable:
Had an offline conversation regarding this -- given the performance regression the file check would present, we are currently not pursuing this fix. In future reworks of `bufmodule`, we will keep this issue in mind and continue to improve our error messaging to users.

PR #546, opened 2021-09-14
CLI V1: Remove all deprecated flags

amckinney:
Re: #483

This builds upon #526 and removes all of the previously deprecated flags from `buf`. All of the tests that used deprecated flags are now removed.

Note that a couple `bufcli` helpers are kept so that we can more easily deprecate flags in the future (i.e. `bufcli.GetString[Slice]FlagOrDeprecatedFlag`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@
 
 - The version key in all configuration files (`buf.yaml`, `buf.gen.yaml`, `buf.work.yaml`) is now required.
 - The `--help` flag now writes to stdout instead of stderr.
+- Removed the `--source` flag on `buf build` in favor of the first positional parameter.
+- Removed the `--source-config` flag on `buf build` in favor of the `--config` flag.
+- Removed the `--file` flag on `buf build` in favor of the `--path` flag.
+- Removed the `--input` flag on `buf lint` in favor of the first positional parameter.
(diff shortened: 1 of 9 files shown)


doriable (approved):

bufdev:
I'm wondering if for a while we should have it so that the command errors if the deprecated flags are used, and then we print a custom error message. I think that especially for the input and config flags, this could come up. This would mean practically abandoning this PR and just erroring in `Get.*OrDeprecatedFlag`, thoughts?

amckinney:
Yea that's a good point, it's also symmetrical with the `appcmd.NewDeletedCommand` approach. I'll give that approach a shot in this PR - if it's sufficiently painful I'll just close and reopen a new one.

doriable:
> I'm wondering if for a while we should have it so that the command errors if the deprecated flags are used, and then we print a custom error message. I think that especially for the input and config flags, this could come up. This would mean practically abandoning this PR and just erroring in `Get.*OrDeprecatedFlag`, thoughts?

Yeah, I can see those flags being integral to workflows, so having them error might be a marginally more graceful change while still requiring a migration if error codes are being handled, etc. But I'm also happy just using the "big hammer" approach here since we're going to 1.0

bufdev:
Commandeering temporarily

PR #551, opened 2021-09-15
Move version back to --version and mod push back to push

bufdev:
See https://github.com/bufbuild/buf/issues/483

`--version` now prints to stdout.


doriable (approved):

bufdev:
Will wait for @amckinney to discuss

PR #555, opened 2021-09-15
Add a DeactivateUser method to the UserService

robbertvanginkel:
In the enterprise version of BSR with SSO support we will allow user
deactivation in addition to deletion. Server administrators will be able to
deactivate user accounts, which will revoke all their active credentials and 
prevent new logins while keeping their data intact.


bufdev (approved):

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -127,6 +129,12 @@ message DeleteUserRequest {}
 
 message DeleteUserResponse {}
 
+message DeactivateUserRequest {
+  string id = 1;
```
johanbrandhorst, line 133: Do we not want this to be something easy like a `username`? The user ID is hardly used anywhere.

bufdev: `GetUser` and other not `By*` functions are all id across all our APIs

robbertvanginkel: Discussed offline, keeping to id for now.

PR #556, opened 2021-09-15
Add Buf GitHub Action

bufdev:


.github/workflows/buf.yaml
```diff
@@ -0,0 +1,48 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: bufbuild/buf-setup-action@v0.3.1
+        if: success()
+        with:
+          version: 0.56.0
+      - uses: bufbuild/buf-breaking-action@v0.4.0
+        if: success()
+        with:
+          against: buf.build/bufbuild/core
```
johanbrandhorst, line 30:
```suggestion
          against: buf.build/bufbuild/buf
```

.github/workflows/buf.yaml
```diff
@@ -0,0 +1,48 @@
...
+          buf_token: ${{ secrets.BUF_TOKEN }}
```
johanbrandhorst, line 31: We should allow this to fail, right?

johanbrandhorst (changes requested):

.github/workflows/buf.yaml
bufdev: Not anymore - we ignore unstable packages

johanbrandhorst (approved):

doriable (approved):

amckinney (approved):

Issue #562, opened 2021-09-16
Large workspaces are slow


amckinney:
We've received reports from a user experiencing _very_ slow `buf generate` invocations with a large workspace (~13 directories). We made some performance improvements in https://github.com/bufbuild/buf/pull/544, but the user is still bottlenecked at the `get_module_config` span, which can be found [here](https://github.com/bufbuild/buf/blob/7f3c50953057363be7e7e66705b6c255ba36434d/private/buf/bufwire/module_config_reader.go#L72).

For reference, the user reported the following debug output:
```
get_module_config	{"duration": "1m41.83627575s"} 
```

We'll need to investigate this further and see what else we can do to improve the performance of the `bufwire.ModuleConfigReader`.

bufdev:
I believe this issue was unrelated - we've tested this out and can't find any further issues.

PR #563, opened 2021-09-16
WIP: large workspace test

bufdev:
For https://github.com/bufbuild/buf/issues/562

This builds just fine, so I'm wondering if the issue is something else.


bufdev:
I think maybe we just close this, thoughts @amckinney @doriable? The script isn't that complicated, we can bring it back later if we want.

amckinney:
Agreed, I feel like we can use this as a one-off (if ever) and don't need to check it in.

Issue #541, opened 2021-09-11
Improve remotely generated plugin errors UX


johanbrandhorst:
This is what an error in a remote plugin invocation looks like:
```
$ buf generate
Failure: failed to generate plugins concurrently: failed to execute plugins concurrently: failed to generate files for remote "buf.build": failed to generate files for given image: Plugin exited with non-zero status 1: --ruby_out: bla.proto: proto3 message field bla in file google/protobuf/compiler/plugin.proto has a dependency on a type from proto2 file google/protobuf/compiler/plugin.proto.  Ruby doesn't support proto2 yet, so we must fail.
exit status 1
exit status 1
.
```

bufdev:
The error should just be `could not find protoc plugin for name debug`. https://github.com/bufbuild/buf/pull/500 introduced additional error wrapping, we should remove it

PR #564, opened 2021-09-16
Remove unnecessary error wrapping in 'buf generate'

amckinney:
Fixes https://github.com/bufbuild/buf/issues/541

This removes all of the unnecessary error wrapping in the `buf generate` command, and makes slight improvements to the places where wrapping makes sense.

Now, when users run `buf generate` with duplicate plugins like so:

```yaml
# This template is for go only for OSS types.
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/bufbuild/buf/private/gen/proto/go
plugins:
  - name: go
    out: private/gen/proto/go
    opt: paths=source_relative
  - name: go
    out: private/gen/proto/go
    opt: paths=source_relative
```

They will see:
```sh
$ buf generate
Failure: file "buf/alpha/module/v1alpha1/module.pb.go" was generated multiple times
```

And for plugins that don't exist, they will see:

```sh
$ buf generate
Failure: could not find protoc plugin for name foo
```

### DIFF
--- a/private/bufpkg/bufplugin/bufplugin.go
+++ b/private/bufpkg/bufplugin/bufplugin.go
@@ -311,7 +311,7 @@ func MergeInsertionPoints(responses []*pluginpb.CodeGeneratorResponse) ([]Merged
 				continue
 			}
 			if _, ok := allFiles[fileName]; ok {
-				return nil, fmt.Errorf("file %q defined multiple times", fileName)
+				return nil, fmt.Errorf("file %q was generated multiple times", fileName)
 			}
 			file := &File{
 				Name:    fileName,
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -98,14 +98,14 @@ func (g *generator) generate(
 		includeImports,
 	)
 	if err != nil {
-		return fmt.Errorf("failed to generate plugins concurrently: %w", err)
+		return err
 	}
 	if len(pluginResponses) != len(config.PluginConfigs) {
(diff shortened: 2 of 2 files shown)


doriable (approved):

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -311,7 +311,7 @@ func MergeInsertionPoints(responses []*pluginpb.CodeGeneratorResponse) ([]Merged
 				continue
 			}
 			if _, ok := allFiles[fileName]; ok {
-				return nil, fmt.Errorf("file %q defined multiple times", fileName)
+				return nil, fmt.Errorf("file %q was generated multiple times", fileName)
```
johanbrandhorst, line 314: I think it'd be even better if we could get the names of the plugins that caused the responses. What do you think about changing the parameter to this function to take a wrapper that contains the plugin name as well as the response?

johanbrandhorst (commented):
LGTM, just wondering if we could improve the MergeInsertionPoints error

private/bufpkg/bufplugin/bufplugin.go
amckinney: Yea I was thinking about this and was originally dissuaded since the `PluginConfig` types differed between this usage and the one in `bufbuild/core`. But given that we already have a `bufplugin` type for `MergedPluginResult`, it seems fine to include a symmetrical request type.

I'll check this out in a follow-up and pair it along with the open issue related to this error in particular.

Issue #560, opened 2021-09-16
Regression in v0.55.0: failed to map insertion points: file defined multiple times


bufdev:
See https://github.com/srikrsna/protoc-gen-gotag/issues/26

```
git clone https://github.com/srikrsna/protoc-gen-gotag
cd ./protoc-gen-gotag
rm buf.gen.yaml
cat <<EOF > buf.gen.yaml
version: v1beta1
plugins:
  - name: go
    out: .
    opt: paths=source_relative
  - name: gotag
    out: .
    opt: paths=source_relative,xxx=graphql+"-" bson+"-"
EOF
go install google.golang.org/protobuf/cmd/protoc-gen-go
go install .
buf generate
```

Results in:

```
Failure: failed to map insertion points: file "tagger/tagger.pb.go" defined multiple times
```

This did not happen in v0.54.0.

johanbrandhorst:
The error reported here and that reported in the issue linked is different:
```
[error] [gotag]: open tagger/tagger.pb.go: no such file or directory
```

I think there's a race condition here in the invocation of the go_tag plugin. I'm not sure how it was working before, but it'll either result in the error in this issue (the duplicate generated files), or in the error reported in the original issue (no such file or directory). The README of the plugin itself recommends running the generator _after_ the original generator, so it seems to be completely accidental and arguably unintentional that the old generation supported this pattern. I think it only worked because the old generation was serializing the execution of the plugins, even though each plugin was executed in parallel. The new generation executes plugins and directories in parallel, so dependencies like this are no longer possible.

I think the resolution in the issue is the right one, and that `buf` is handling this correctly. The user should run two `buf generate` invocations to mimic the two `protoc` invocations recommended by the README.

srikrsna:
It is not a race condition. I've updated the repo to include two buf invocations. Basically, it edits the generated Go source code to add the struct tags. One way of fixing it would be to give a flag to run plugins sequentially in buf. But this is a rare use case I wouldn't do it just for this.

amckinney:
I agree that we shouldn't add a flag to support sequential generation for this case. In general, plugins should _always_ assume that they should invoke within a single `protoc` and/or `buf` invocation. If a plugin needs to modify the content of another plugin's output, they really ought to use [insertion points](https://github.com/protocolbuffers/protobuf/blob/2501e9b856c411239e2ae32bcc75518dcf840418/src/google/protobuf/compiler/plugin.proto#L172) (that's what they are for).

[...]

PR #565, opened 2021-09-16
Include plugin name in 'buf generate' errors

amckinney:
Fixes https://github.com/bufbuild/buf/issues/560

> The issue described above is only loosely related to this change - the only action item there was to improve our error messages (hence why this PR fixes the issue).

This builds upon https://github.com/bufbuild/buf/pull/564, and introduces a new `PluginResponse` type that includes the `PluginName` responsible for the creation of a given `CodeGeneratorResponse`.

This was tested with a variety of configurations, including:

#### Two Remote Plugins
```sh
$ buf generate
Failure: file "buf/alpha/image/v1/image.pb.go" was generated multiple times: once by plugin "buf.build/library/plugins/go:v1.27.1-1" and again by plugin "buf.build/library/plugins/go"
```

#### Local + Remote Plugins
```sh
$ buf generate
Failure: file "buf/alpha/image/v1/image.pb.go" was generated multiple times: once by plugin "buf.build/library/plugins/go" and again by plugin "go"
```

#### Two Local Plugins
```sh
$ buf generate
Failure: file "buf/alpha/image/v1/image.pb.go" was generated multiple times: once by plugin "go" and again by plugin "go"
```

We _could_ include the index of each plugin in the error to disambiguate two plugins with the same name, but this error message should be sufficient in telling the user that they've configured something twice. Plus, reporting the index makes the error message a little harder to read and understand at first glance.

If we proceed with this change, we'll need to carry it over to the server-side generator. I have the change complete locally, but I'll need to discuss with the team since there are a couple of minor caveats.

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -37,7 +37,6 @@ import (
 	"go.uber.org/multierr"
 	"go.uber.org/zap"
 	"golang.org/x/sync/errgroup"
-	"google.golang.org/protobuf/types/pluginpb"
 )
 
 type generator struct {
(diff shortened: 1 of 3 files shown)


johanbrandhorst (approved):
:chefs_kiss:

doriable (approved):

PR #566, opened 2021-09-17
Add GetErrorMessage to rpc package

johanbrandhorst:
This is useful for unwrapping just the message from
a wrapped RPC error.

### DIFF
--- a/private/pkg/rpc/errors.go
+++ b/private/pkg/rpc/errors.go
@@ -369,6 +369,18 @@ func GetErrorCode(err error) ErrorCode {
 	return ErrorCodeInternal
 }
 
+// GetErrorMessage gets the rpc error message, if there is an RPC
+// error in the error chain. If not, it returns an empty string.
+func GetErrorMessage(err error) string {
+	if err == nil {
+		return ""
+	}
+	if rpcErr := new(rpcError); errors.As(err, &rpcErr) {
+		return rpcErr.message
+	}
+	return ""
+}
+
 // IsError returns true if err is an error created by this package.
 //
 // If the error is nil, this returns false.
--- a/private/pkg/rpc/errors_test.go
+++ b/private/pkg/rpc/errors_test.go
@@ -1,0 +1,12 @@
+package rpc
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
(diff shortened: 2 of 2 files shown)


private/pkg/rpc/errors.go
```diff
@@ -369,6 +369,18 @@ func GetErrorCode(err error) ErrorCode {
 	return ErrorCodeInternal
 }
 
+// GetErrorMessage gets the rpc error message, if there is an RPC
+// error in the error chain. If not, it returns an empty string.
+func GetErrorMessage(err error) string {
```
amckinney, line 374: The `Error()` implementation below should be sufficient, no?

johanbrandhorst: It isn't used if the error is wrapped. This is used to unwrap the RPC error message only, which is probably what we want to display to users.

amckinney: Ah I see now. The only alternative here would be exporting the `rpcError` type, which isn't what we want. Having a separate getter to access its internal field (like `GetErrorCode`) seems fine.

However, should we just remove all the wrapping in these particular cases so that we can reliably just use `.Error()`? In other words, when is the wrapping actually useful if we just want to extract this bit later on?

johanbrandhorst: Are you suggesting not wrapping errors in the application altogether? They're wrapped to help with pinpointing where the error comes from an add extra context. They're useful for debugging unexpected internal errors, but for RPC errors we don't care for them, since those errors are usually tailored to be seen by the user.

amckinney: Not in general, no - I understand the use case of error wrapping in general. I just figured that it could be another case of https://github.com/bufbuild/buf/pull/564, where the wrapping actually did more harm than good.

johanbrandhorst: We want these richer wrapped errors for unexpected error cases, but we'd still like to have the fine control over what the user sees, so this gives us the best of both worlds, which we don't have yet on the client side.

amckinney: Also discussed offline - it's best to handle this on the server-side rather than the client-side so that every one of our (future) clients get the same, clean error behavior rather than special-casing it for the `buf` CLI.

amckinney (approved):

PR #569, opened 2021-09-19
Upgrade protoc to 3.18.0

bufdev:
Per internal guide upgrading_protoc_version.md


bufdev:
We need to move Windows installation from `choco` to our own script as we need to keep the `protoc` version consistent, otherwise we have a diff in generated files during test, see https://github.com/bufbuild/buf/pull/569/checks?check_run_id=3645036940

bufdev:
See https://github.com/bufbuild/buf/issues/570

bufdev:
Once we figure out how to install `protoc` by Zip in Windows, we should update upgrading_protoc_version.md to reflect that the variable inside `windows/setupDeps.ps1` needs to be updated as well.

amckinney (approved):

doriable (approved):

PR #571, opened 2021-09-19
Cleanup Windows GitHub Action and install protoc from zip

bufdev:
This PR should be merged before https://github.com/bufbuild/buf/pull/569

This does the following:

- Compresses all setup steps into one command `setup-deps` - there's no reason for them to be separate
- Moves the Golang setup right before checkout, similar to `ci.yaml`
- Fixes caching: right now, the Chocolately cache actually is never used, as the key is changed on every git commit, which means that the cache is always overwritten. The cache key should depend on files that say whether or not the cache significantly changed and should be restored or not. This is now controlled by the hash of `go.sum` and the files in `windows`.
- Moves caching all into one step with multiple paths.
- Starts the update to installing `protoc` from zip. I don't know where the `PATH` is on Windows or now to do moves, so this is a TODO inside `windows/setupDeps.ps1`. @doriable can you take this over?

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -1,5 +1,7 @@
 name: windows
 on: push
+# Prevent writing to the repository using the CI token.
+# Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
 permissions: read-all
 jobs:
   test:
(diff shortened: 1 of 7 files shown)


doriable:
> * Starts the update to installing `protoc` from zip. I don't know where the `PATH` is on Windows or now to do moves, so this is a TODO inside `windows/setupDeps.ps1`. @doriable can you take this over?

Yep, I can commandeer this on Monday and check the pathing.

bufdev:
SGTM

bufdev:
LGTM

PR #582, opened 2021-09-22
Add --include-imports to ls-files

bufdev:
Not having this has been annoying me.

We also might want to add a flag such as `--root-relative` or so that strips the external paths.

### DIFF
--- a/private/buf/bufwire/file_lister.go
+++ b/private/buf/bufwire/file_lister.go
@@ -21,11 +21,14 @@ import (
 	"github.com/bufbuild/buf/private/buf/bufconfig"
 	"github.com/bufbuild/buf/private/buf/buffetch"
 	"github.com/bufbuild/buf/private/buf/bufwork"
+	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
+	"github.com/bufbuild/buf/private/bufpkg/bufimage"
 	"github.com/bufbuild/buf/private/bufpkg/bufimage/bufimagebuild"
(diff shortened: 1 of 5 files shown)


private/buf/bufwire/file_lister.go
```diff
@@ -61,6 +77,74 @@ func (e *fileLister) ListFiles(
...
+		return nil, nil, err
+	}
+	imageFiles := image.Files()
+	fileInfos := make([]bufmoduleref.FileInfo, len(imageFiles))
+	for i, imageFile := range imageFiles {
+		fileInfos[i] = imageFile
+	}
+	// The image is ordered in DAG order, which isn't consistent with what we do
+	// in the withoutImports flow and is a little annoying for a user.
+	bufmoduleref.SortFileInfos(fileInfos)
```
bufdev, line 139: Maybe we should do it in DAG order anyways? It is more consistent with other outputs I think, I don't know.

doriable: I think I prefer the alpha-sorted order, seems easier to reason about, from a human readability standpoint. But if we're expecting this to be used as inputs to another process (which seems pretty likely), I think either is fine. But it doesn't seem too costly to sort, so it might be beneficial to have that just for the use-cases where users are inspecting the output.

amckinney: As long as it's deterministic, we should feel empowered to change the ordering later if need be.

At first glance, it seems best to keep this consistent with the `withoutImports` flow.

amckinney (approved):

doriable (approved):

PR #583, opened 2021-09-22
Delete twirp

bufdev:


doriable (approved):

johanbrandhorst (approved):
:salute_canada:

mculp:
I just found this repo, was looking through it, and was just wondering why twirp was deleted? I don't see any context in the PR or commit msg

cep21:
Also curious.  Is buf not a good fit for people using twirp?

bufdev:
This was an implementation detail - for a while, we had code that supported the BSR's API (that `buf` talks to) both over Twirp and gRPC. We never actually exposed Twirp publicly via our API. This PR just deleted that unused code.

Buf is a fine fit for people using Twirp, including `protoc-gen-twirp` - this PR is completely unrelated to `buf`s usability with any plugin. Buf is meant to be used with all Protobuf plugins.

Issue #584, opened 2021-09-23
Started getting `git-submodule cannot be used without a working tree` with `buf breaking`


savardc:
We started getting this error when updating to any version that includes https://github.com/bufbuild/buf/pull/421. We have a single submodule and everything is working fine with v0.50.0

The command looks like: `buf breaking --against '.git#branch=master,recurse_submodules=true'`

Thanks

robbertvanginkel:
Thanks for reporting, I'll take a look.

robbertvanginkel:
This will go out in the next release (should be soon).

savardc:
Thank you!!!

bufdev:
This has been released in v1.0.0-rc2

zxia-wish:
It looks like the problem is not completely solved, on my end it's still producing errors:
```
bash-5.1# buf breaking --against '.git#branch=master,recurse_submodules=true'
Failure: could not clone file:///protos/.git: exit status 1
warning: could not look up configuration 'remote.origin.url'. Assuming this repository is its own authoritative upstream.
Submodule 'third_party/api-common-protos' (/tmp/api-common-protos) registered for path 'third_party/api-common-protos'
fatal: repository '/tmp/api-common-protos' does not exist
fatal: clone of '/tmp/api-common-protos' into submodule path '/third_party/api-common-protos' failed
Failed to clone 'third_party/api-common-protos'. Retry scheduled
fatal: repository '/tmp/api-common-protos' does not exist
fatal: clone of '/tmp/api-common-protos' into submodule path '/third_party/api-common-protos' failed
Failed to clone 'third_party/api-common-protos' a second time, aborting
```

Any comments or suggestions? Is there a way to check what git commands are being used?

robbertvanginkel:
We unfortunately don't have a verbose flag that will log that in mainline, but I put up a quick branch in  https://github.com/bufbuild/buf/tree/robbert-gitexeclog

that you can use like:
```
$ GO111MODULE=on GOBIN=$(pwd) go get \
  github.com/bufbuild/buf/cmd/buf@robbert-gitexeclog
go: downloading github.com/bufbuild/buf v1.0.0-rc2.0.20210928160910-39fcfb299de0
go: downloading golang.org/x/sys v0.0.0-20210923061019-b8560ed6a9b7
go: downloading go.uber.org/atomic v1.9.0
go: downloading google.golang.org/genproto v0.0.0-20210924002016-3dee208752a0
[...]

PR #585, opened 2021-09-23
Fix regression in recurse_submodules

robbertvanginkel:
This should fix https://github.com/bufbuild/buf/issues/584. 

Apparently the submodule subcommand does not support the `--worktree` argument, and needs to be run from a checked out worktree. By changing the first checkout to use `git worktree` (https://git-scm.com/docs/git-worktree, available since git 2.6.7) and the later submodules command to run in the new worktree, git submodules seems to be ok.

Verified this locally with git 2.10 (Sept 2016), similar to https://github.com/bufbuild/buf/pull/421.

### DIFF
--- a/private/pkg/git/cloner.go
+++ b/private/pkg/git/cloner.go
@@ -315,7 +315,10 @@ func getSSHKnownHostsFilePaths(sshKnownHostsFiles string) []string {
 	return filePaths
 }
 
-func getRefspecsForName(gitName Name) (string, string, string) {
+// getRefspecsForName decides the refspecs to use in the subsequent git fetch,
+// git worktree add and git checkout. When checkoutRefspec is empty, Name
+// explicitly refer to a named ref and the checkout isn't a necessary step.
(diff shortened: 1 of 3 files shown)


robbertvanginkel:
Need to dig deeper to understand what is going wrong about this on our windows ci.

private/pkg/git/cloner.go
```diff
@@ -302,9 +315,9 @@ func getSSHKnownHostsFilePaths(sshKnownHostsFiles string) []string {
 	return filePaths
 }
 
-func getRefspecsForName(gitName Name) (string, []string) {
+func getRefspecsForName(gitName Name) (string, string, string) {
```
bufdev, line 318: can you add a comment about what each return argument is?

bufdev (approved):

bufdev:
Approved pending removal of any debugging stuff to fix windows

bufdev:
Also can you add a "Fix ..." line to the CHANGELOG.md?

PR #586, opened 2021-09-23
Update proto for deprecation

WillAbides:
This updates protobuf definitions to support deprecation


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -202,3 +208,13 @@ message DeleteRepositoryByFullNameRequest {
 }
 
 message DeleteRepositoryByFullNameResponse {}
+
+message UpdateRepositoryDeprecationRequest{
+  string id = 1;
+  bool deprecated = 2;
+  string deprecation_message = 3;
+}
```
amckinney, line 216: The semantics here get a little blurred for the two different use cases: undeprecating vs. deprecating.

For deprecating, it makes sense - we require a non-empty `deprecation_message` and can return a `InvalidArgumentError` to the user if it's not set.

For undeprecating though, what does it mean to set `deprecated` to `false`, and give a non-empty `deprecation_message`? Will the `deprecation_message` just be dropped entirely and have no effect? Persisting it to the database wouldn't be useful since it would never be surfaced if `deprecated` is `false`.

With that said, it might be best to either:
* Model this as a `RepositoryState` (similar to the generic `UserState` proposed for [de]activating users).
  * Do we imagine any other states like `retracted` on the repository level that would actually cause an error to the user? If so, a more generic `RepositoryState` could be worthwhile.
* Split this out into two endpoints: `DeprecateRepository` and `UndeprecateRepository`, where each endpoint only exposes the minimal amount of fields necessary to perform the action.

The decision here might depend on how we want to model this in the `Module` definitions. If we ever want to "retract" specific module versions, it might make more sense to model the same ability for Repositories (i.e. the `RepositoryState` option). We could define a separate `ModuleState` enum so that the two states could evolve independently though.

What are your thoughts?

WillAbides: I started out with DeprecateRepository and UndeprecateRepository as you suggested, but switched to UpdateRepositoryDeprecation for consistency after I saw the other UpdateRepository* rpcs. I'm happy to switch back to that.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -63,4 +63,6 @@ message ModulePin {
   string commit = 5;
   string digest = 6;
   google.protobuf.Timestamp create_time = 7;
+  bool deprecated = 8;
+  string deprecation_message = 9;
```
amckinney, line 67: As discussed, including this information on the `ModulePin` is necessary for letting the user know they've updated their dependencies to a deprecated module. And as you've mentioned before, it might be necessary that we also include this information on the `Module` so that the same could be done for other commands like `buf build` (aka anyone that relies on the `DownloadService`)?


If we're worried about including this information on the `Module[Pin]` itself (since it's only used to emit warnings for particular RPCs), we _could_ include it alongside the `ModulePins` returned in the `GetModulePinsResponse`.

In the following examples, I'm referencing `ModuleState` as the `deprecated[_message]` values you have above so that it could eventually support a retracted state, but I'm not even sure that'd be necessary since we might not ever need to return whether or not a module is retracted - we could just return an error on the server-side.

```protobuf
message GetModulePinsResponse {
  repeated buf.alpha.module.v1alpha1.ModulePin module_pins = 1;
  repeated buf.alpha.module.v1alpha1.ModuleState module_states = 2;
}
```

Similarly, we could do this for the `DownloadResponse`, too:

```protobuf
message DownloadResponse {
  buf.alpha.module.v1alpha1.Module module = 1;
  buf.alpha.module.v1alpha1.ModuleState module_state = 2;
}
```

But the only thing that gains us is not tying the idea of a `deprecated[_message]` to the `Module[Pin]` itself, and it actually makes the API less ergonomic and easier to mess up. However, it prevents any changes to the primary module definitions, so it might be worth considering.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -40,6 +40,10 @@ message Repository {
     string organization_id = 6;
   }
   Visibility visibility = 7;
+  // deprecated means this repository is deprecated.
+  bool deprecated = 8;
+  // deprecation_message is the message shown if the repository is deprecated.
+  string deprecation_message = 9;
```
amckinney, line 46: I know you mentioned that you wanted to split these out so that users could deprecate something without including a message, but we might actually want to require a message in all cases (similar to what we do for [API tokens](https://buf.build/settings/user)).

Plus, that gives us the benefit that if `deprecation_message` is set, then we know the entity is deprecated. And it's impossible for `deprecated_message` to be set, but have `deprecated` equal to `false.

proto/buf/alpha/module/v1alpha1/module.proto
amckinney: Thinking about this more - 

It might be that we end up just having a `deprecation_message` field on each of `Module`, `ModulePin` and `Repository` and call it there. We probably don’t need to go overboard and support a generic state like `Retracted` since we can just return server-side errors for those states if we ever need to (e.g. if I try to download a retracted module, we just return an error, not the `Module` type).

Even though the `ModulePin` has a `deprecation_message` field, we don't have to serialize it in the `buf.lock`. And if we ever want to support a `deprecated: <string>` configuration in the `buf.yaml`, we'll already have an avenue to do so since the `Module` will already have it.

proto/buf/alpha/registry/v1alpha1/repository.proto
bufdev: Idk, I might keep them separate anyways, and do backend validation that false + non-empty is never done - I can see it both ways

proto/buf/alpha/module/v1alpha1/module.proto
amckinney: Discussed offline - 

Let's lean into deprecation being part of the `Repository` and start by determining whether or not a `Module[Pin]` is deprecated by interacting with the `RepositoryService`. It will require an additional round-trip for the commands that care about deprecation, but it’s easier to start with this and change it later if need be.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -70,6 +74,20 @@ service RepositoryService {
   rpc DeleteRepository(DeleteRepositoryRequest) returns (DeleteRepositoryResponse);
   // DeleteRepositoryByFullName deletes a repository by full name.
   rpc DeleteRepositoryByFullName(DeleteRepositoryByFullNameRequest) returns (DeleteRepositoryByFullNameResponse);
+  // DeprecateRepositoryByName deprecates the repository.
+  rpc DeprecateRepositoryByName(DeprecateRepositoryByNameRequest) returns(DeprecateRepositoryByNameResponse);
+  // UndeprecateRepositoryByName makes the repository not deprecated and removes any deprecation_message.
+  rpc UndeprecateRepositoryByName(UndeprecateRepositoryByNameRequest) returns(UndeprecateRepositoryByNameResponse);
+  // GetRepositoriesByFullName gets repositories by full name.
+  rpc GetRepositoriesByFullName(GetRepositoriesByFullNameRequest) returns (GetRepositoriesByFullNameResponse);
```
doriable, line 82: Following the previous conventions, should this be `ListRepositoriesByFullName`? I know that it's not a paginated list, so it's not quite the same as the other `List` endpoints, but it's also getting multiple repository entities? Unclear, I leave this up to the jury.

doriable (approved):
Mostly looks good to me, left a comment for discussion, but overall, I agree with the direction, so I'm giving this my stamp. That being said, probably want a stamp from @amckinney and/or @bufdev :)

proto/buf/alpha/registry/v1alpha1/repository.proto
WillAbides: I think of a List* endpoint as one that returns all items that match a criteria vs this which returns a list with a 1:1 correlation of the input list.

bufdev: I think in this context, `Get` is right - I had the same instinct as @doriable but after reviewing other examples in our APIs, this makes sense to me

proto/buf/alpha/module/v1alpha1/module.proto
amckinney: I think we want to remove `deprecated[_message]` from the `ModulePin` type now that we have plans for `GetRepositoriesByFullName``

amckinney (approved):
👏

proto/buf/alpha/registry/v1alpha1/repository.proto
bufdev, line 82: tiny bikeshedding thing: I would name this `GetRepositoriesByFullNames` personally, but I could go either way

bufdev (approved):

proto/buf/alpha/registry/v1alpha1/repository.proto
WillAbides: I considered both and had a slight preference for the singular version because we are getting each one by its full name.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -222,8 +223,10 @@ message DeleteRepositoryByFullNameRequest {
 message DeleteRepositoryByFullNameResponse {}
 
 message DeprecateRepositoryByNameRequest{
-  string owner = 1;
+  string owner_name = 1;
   string repository_name = 2;
+  // A message shown along with the deprecation warning for this repository.
+  // It must be a utf8 encoded string containing 280 or fewer characters.
```
amckinney, line 229: Seems fine, but I'm wondering where `280` comes from.

amckinney (approved):

proto/buf/alpha/registry/v1alpha1/repository.proto
amckinney: Actually, it might be best to standardize on what we use for token notes: 256.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -202,3 +221,24 @@ message DeleteRepositoryByFullNameRequest {
 }
 
 message DeleteRepositoryByFullNameResponse {}
+
+message DeprecateRepositoryByNameRequest{
+  string owner_name = 1;
+  string repository_name = 2;
+  // A message shown along with the deprecation warning for this repository.
+  // It must be a utf8 encoded string containing 256 or fewer characters.
```
amckinney, line 229: Comment is fine as-is, but just so you know the `utf-8` aspect is guaranteed by the `string` definition in Protobuf. You can see the distinction between `string` and `bytes` here: https://developers.google.com/protocol-buffers/docs/proto3#scalar

amckinney (approved):
lgtm!

PR #588, opened 2021-09-23
Add --strip flag to ls-files

bufdev:
Other potential names for this flag:

- `--as-image-paths`: I don't like this as it makes people understand what an Image is.
- `--as-import-paths`: Just seems long, and we don't use the terminology "import paths" anywhere else.

We can always easily change the flag name later.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `--strip` flag to `ls-files` that strips local directory paths and prints file paths as they
+  are imported.
 
 ## [v1.0.0-rc2] - 2021-09-23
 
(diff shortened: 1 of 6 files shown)


amckinney (approved):
Nice, `--strip` is clear and concise as-is.

doriable:
> * `--as-import-paths`: Just seems long, and we don't use the terminology "import paths" anywhere else.

Implementation seems fine, I think I like the name `--as-import-paths`, just because it's explicit about the content displayed. But I'm stamping since the general sentiment of the PR seems reasonable.

doriable (approved):

amckinney:
> Implementation seems fine, I think I like the name `--as-import-paths`, just because it's explicit about the content displayed. But I'm stamping since the general sentiment of the PR seems reasonable.

We do have `--include-imports` and `--exclude-imports`, so `--as-imports` isn't insane. But like you've said we can change it later and hide the old flag if this gives us problems.

PR #589, opened 2021-09-23
Delete tests for --input flags

bufdev:
I think these were made to check for exit code 1 when we deleted these for v1.0.0-rc1. I'd argue they aren't really testing anything anymore, other than commands fail on a non-existent flag. Expecting them causes extra test creation overhead.


amckinney (approved):
👍

Issue #493, opened 2021-08-31
Add --file or equivalent that searches for buf.yaml in IDEs


bufdev:
This is applicable for vim-go, and likely vscode-go.

Right now, vim-go runs `buf lint --path opened/file.proto`. The problem is that this runs in your current directory, which means your config is looked for in your current directory. If you have a config at say `proto/buf.yaml`, this isn't detected, and errors can result.

We should have something like:

```
buf lint --file opened/file.proto
```

That only takes a single file (and errors if `--path` is used), and searches manually for the config file by going up the tree. This flag can even be hidden.

There's other ways to do this, for example we could add `--find-config=true`, and then find all the files using the standard method, then do a search upwards for each file and find any configs, make sure that they all share the same config, and then re-run any file logic (`--path`) given this single config...but this is error-prone, and we really just need this for editors.

We likely do not want to enable config searching universally - this can turn into a nightmare.

bufdev:
I did some investigation here, for future reference, there's a few ways to do this, none of which are great:

### Add a new `--single-file` flag or equivalent

This flag, which can never be used with `--path`, operates as described above. Implementing this is extremely wonky in terms of `buffetch`. The issue is what do you do about `buf lint https://github.com/foo/bar.git#subdir=proto --single-file proto/baz/bat.proto` if you have a `buf.yaml` in `proto/baz`? This is going to be a complete nightmare.

There *may* be a way to use `buffetch/internal.WithGetBucketTerminateFileName` to accomplish this relatively easily (terminate on both `buf.work.yaml` and `buf.yaml`, but this could get ugly. Note that it's only an issue for `Buckets` - `Images` and `Modules` should be relatively straightforward, but still. It could be a lot of nonsense with the `RootRelativePath` and `SubDirPath` on `buffetch/internal.ReadBucketCloser`.

We could instead limit this to say "you can only use `--single file` with directory inputs" or "you can only use `--single-file` with the input `.`", but kind of hacky. But that leads us to...

### Add a new `buffetch/internal.Ref` type

[...]

PR #608, opened 2021-09-27
[WIP] Support a single proto file as a reference for lint in a valid module or workspace

doriable:
The purpose of this is to support use cases, such as editor plugins, where a user will need to provide a specific file as a reference and `buf` will need to resolve the valid module/workspace for linting, and then return a filtered lint result.

Some thoughts/points of discussion:

- Is the name `SingleFileRef` clear enough, or should we make it something like `SingleProtoFileRef`, so it is not conflated with our existing file ref that takes a tar/zipped file
- We may want similar reference for the other commands in the future or for the LSP, since we would want to provide a single proto file reference, but surface build errors as needed
- Currently, we fail when the user provides a single proto file reference as the input but also provides the same proto file as the `--path` flag. This is expected and I believe this behaviour to be correct, but we may want to surface a clearer error
- Currently this is a reference that is only accepted for the `lint` command and is not surfaced in the command line help. We should discuss some of the details of the user experience here

WIP:

- [ ] Add some more testing for workspaces, etc.

Ref: #493

### DIFF
--- a/private/buf/buffetch/buffetch.go
+++ b/private/buf/buffetch/buffetch.go
@@ -157,8 +157,24 @@ type RefParser interface {
 // NewRefParser returns a new RefParser.
 //
 // This defaults to dir or module.
-func NewRefParser(logger *zap.Logger) RefParser {
-	return newRefParser(logger)
+func NewRefParser(logger *zap.Logger, options ...NewRefParserOption) RefParser {
+	return newRefParser(logger, options...)
+}
+
+// NewRefParserOption is an option for NewRefParser. NewRefParserOptions are only accepted
+// for the default RefParser constructor.
+type NewRefParserOption func(*newRefParserOption)
+
+type newRefParserOption struct {
+	allowSingleFileRef bool
+}
+
+// NewRefParserWithSingleFileRefAllowed sets the NewRefParser to return the single file reference
+// where applicable.
+func NewRefParserWithSingleFileRefAllowed() NewRefParserOption {
+	return func(o *newRefParserOption) {
+		o.allowSingleFileRef = true
+	}
 }
 
 // NewImageRefParser returns a new RefParser for images only.
--- a/private/buf/buffetch/format.go
+++ b/private/buf/buffetch/format.go
@@ -35,6 +35,8 @@ const (
 	formatTargz = "targz"
 	// formatZip is the zip format.
 	formatZip = "zip"
+	// formatSingleFile is the single file format
+	formatSingleFile = "file"
 )
 
 var (
@@ -54,6 +56,7 @@ var (
 	sourceFormats = []string{
 		formatDir,
 		formatGit,
+		formatSingleFile,
 		formatTar,
 		formatTargz,
 		formatZip,
@@ -62,6 +65,7 @@ var (
 	sourceFormatsNotDeprecated = []string{
 		formatDir,
 		formatGit,
+		formatSingleFile,
 		formatTar,
 		formatZip,
 	}
@@ -78,6 +82,7 @@ var (
 		formatDir,
 		formatGit,
 		formatMod,
+		formatSingleFile,
 		formatTar,
 		formatTargz,
 		formatZip,
@@ -87,6 +92,7 @@ var (
 		formatDir,
 		formatGit,
 		formatMod,
+		formatSingleFile,
 		formatTar,
 		formatZip,
 	}
@@ -99,6 +105,7 @@ var (
 		formatJSON,
 		formatJSONGZ,
 		formatMod,
+		formatSingleFile,
 		formatTar,
 		formatTargz,
 		formatZip,
@@ -110,6 +117,7 @@ var (
 		formatGit,
 		formatJSON,
 		formatMod,
+		formatSingleFile,
 		formatTar,
 		formatZip,
 	}
--- a/private/buf/buffetch/internal/internal.go
+++ b/private/buf/buffetch/internal/internal.go
@@ -597,6 +597,7 @@ func WithModuleFormat(format string, options ...ModuleFormatOption) RefParserOpt
 }
 
 // WithSingleFileFormat attaches the given format as a single file format.
+// The single file ref format is only allowed if the option for the parser is given. Otherwise this is a nop.
 //
 // It is up to the user to not incorrectly attach a format twice.
 func WithSingleFileFormat(format string, options ...SingleFileFormatOption) RefParserOption {
--- a/private/buf/buffetch/internal/internal.go
+++ b/private/buf/buffetch/internal/internal.go
@@ -161,6 +161,14 @@ func NewDirRef(path string) (DirRef, error) {
 	return newDirRef("", path)
 }
 
+// SingleFileRef is a file reference that incorporates a BucketRef ..
+type SingleFileRef interface {
+	BucketRef
+	// Path is the normalized path to the file reference.
+	Path() string
+	singleFileRef()
+}
+
 // GitRef is a git reference.
 type GitRef interface {
 	// Path is the path to the reference.
@@ -291,6 +299,11 @@ type ParsedDirRef interface {
 // This should only be used for testing.
 func NewDirectParsedDirRef(format string, path string) ParsedDirRef {
 	return newDirectDirRef(format, path)
+}
+
+type ParsedSingleFileRef interface {
+	SingleFileRef
+	HasFormat
 }
 
 // ParsedGitRef is a parsed GitRef.
@@ -583,6 +596,23 @@ func WithModuleFormat(format string, options ...ModuleFormatOption) RefParserOpt
 	}
 }
 
+// WithSingleFileFormat attaches the given format as a single file format.
+//
+// It is up to the user to not incorrectly attach a format twice.
+func WithSingleFileFormat(format string, options ...SingleFileFormatOption) RefParserOption {
+	return func(refParser *refParser) {
+		format = normalizeFormat(format)
+		if format == "" {
+			return
+		}
+		singleFileFormatInfo := newSingleFileFormatInfo()
+		for _, option := range options {
+			option(singleFileFormatInfo)
+		}
+		refParser.singleFileFormatToInfo[format] = singleFileFormatInfo
+	}
+}
+
 // SingleFormatOption is a single format option.
 type SingleFormatOption func(*singleFormatInfo)
 
@@ -614,8 +644,11 @@ type GitFormatOption func(*gitFormatInfo)
 // ModuleFormatOption is a module format option.
 type ModuleFormatOption func(*moduleFormatInfo)
 
-// ReaderOption is an Reader option.
+// ReaderOption is a Reader option.
 type ReaderOption func(*reader)
+
+// SingleFileFormatOption is a single file format option.
+type SingleFileFormatOption func(*singleFileFormatInfo)
 
 // WithReaderHTTP enables HTTP.
 func WithReaderHTTP(httpClient *http.Client, httpAuthenticator httpauth.Authenticator) ReaderOption {
--- a/private/buf/buffetch/internal/reader.go
+++ b/private/buf/buffetch/internal/reader.go
@@ -138,6 +138,13 @@ func (r *reader) GetBucket(
 			t,
 			getBucketOptions.terminateFileNames,
 		)
+	case SingleFileRef:
+		return r.getSingleFileBucket(
+			ctx,
+			container,
+			t,
+			getBucketOptions.terminateFileNames,
+		)
 	default:
 		return nil, fmt.Errorf("unknown BucketRef type: %T", bucketRef)
 	}
@@ -336,6 +343,72 @@ func (r *reader) getDirBucket(
 	}
 	readWriteBucket, err := r.storageosProvider.NewReadWriteBucket(
 		dirRef.Path(),
+		storageos.ReadWriteBucketWithSymlinksIfSupported(),
+	)
+	if err != nil {
+		return nil, err
+	}
+	return newReadWriteBucketCloser(
+		storage.NopReadWriteBucketCloser(readWriteBucket),
+		"",
+		"",
+	)
+}
+
+func (r *reader) getSingleFileBucket(
+	ctx context.Context,
+	container app.EnvStdinContainer,
+	singleFileRef SingleFileRef,
+	terminateFileNames []string,
+) (ReadBucketCloser, error) {
+	if !r.localEnabled {
+		return nil, NewReadLocalDisabledError()
+	}
+	dirPath := filepath.Dir(singleFileRef.Path())
+	terminateFileDirectoryAbsPath, err := findTerminateFileDirectoryPathFromOS(dirPath, terminateFileNames)
+	if err != nil {
+		return nil, err
+	}
+	if terminateFileDirectoryAbsPath != "" {
+		// If the terminate file exists, we need to determine the relative path from the
+		// terminateFileDirectoryAbsPath to the target DirRef.Path.
+		wd, err := osextended.Getwd()
+		if err != nil {
+			return nil, err
+		}
+		terminateFileRelativePath, err := normalpath.Rel(wd, terminateFileDirectoryAbsPath)
+		if err != nil {
+			return nil, err
+		}
+		rootPath := terminateFileRelativePath
+		if filepath.IsAbs(singleFileRef.Path()) {
+			// If the input was provided as an absolute path,
+			// we preserve it by initializing the workspace
+			// bucket with an absolute path.
+			rootPath = terminateFileDirectoryAbsPath
+		}
+		readWriteBucket, err := r.storageosProvider.NewReadWriteBucket(
+			rootPath,
+			storageos.ReadWriteBucketWithSymlinksIfSupported(),
+		)
+		if err != nil {
+			return nil, err
+		}
+		// Verify that the subDirPath exists too.
+		if _, err := r.storageosProvider.NewReadWriteBucket(
+			normalpath.Join(rootPath, dirPath),
+			storageos.ReadWriteBucketWithSymlinksIfSupported(),
+		); err != nil {
+			return nil, err
+		}
+		return newReadWriteBucketCloser(
+			storage.NopReadWriteBucketCloser(readWriteBucket),
+			rootPath,
+			rootPath,
+		)
+	}
+	readWriteBucket, err := r.storageosProvider.NewReadWriteBucket(
+		dirPath,
 		storageos.ReadWriteBucketWithSymlinksIfSupported(),
 	)
 	if err != nil {
--- a/private/buf/buffetch/internal/reader.go
+++ b/private/buf/buffetch/internal/reader.go
@@ -391,8 +391,7 @@ func (r *reader) getSingleFileBucket(
 		}
 	}
 	if terminateFileDirectoryAbsPath == "" {
-		// TODO(doria): determine the correct error to return here
-		return nil, fmt.Errorf("cannot resolve valid workspace for file reference: %s", singleFileRef.Path())
+		return nil, fmt.Errorf(`cannot resolve valid module or workspace for reference: "%s"`, singleFileRef.Path())
 	}
 	// If the terminate file exists, we need to determine the relative path from the
 	// terminateFileDirectoryAbsPath to the target singleFileRef.Path().
--- a/private/buf/buffetch/internal/reader.go
+++ b/private/buf/buffetch/internal/reader.go
@@ -24,6 +24,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/bufbuild/buf/private/buf/bufconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule"
 	"github.com/bufbuild/buf/private/pkg/app"
 	"github.com/bufbuild/buf/private/pkg/git"
@@ -364,59 +365,77 @@ func (r *reader) getSingleFileBucket(
 	if !r.localEnabled {
 		return nil, NewReadLocalDisabledError()
 	}
+	terminateFileNames = append(terminateFileNames, bufconfig.ExternalConfigV1FilePath)
 	dirPath := filepath.Dir(singleFileRef.Path())
+
 	terminateFileDirectoryAbsPath, err := findTerminateFileDirectoryPathFromOS(dirPath, terminateFileNames)
 	if err != nil {
 		return nil, err
 	}
-	if terminateFileDirectoryAbsPath != "" {
-		// If the terminate file exists, we need to determine the relative path from the
-		// terminateFileDirectoryAbsPath to the target DirRef.Path.
-		wd, err := osextended.Getwd()
-		if err != nil {
-			return nil, err
+	if filepath.Base(terminateFileDirectoryAbsPath) == bufconfig.ExternalConfigV1FilePath {
+		// walk again to check if there is a workspace
+		// if workspaces are not enabled here, this is a noop
+		// we always add `buf.yaml` at the end, so we just trim it
+		if len(terminateFileNames) > 1 {
+			terminateFileNames = terminateFileNames[:len(terminateFileNames)-1]
+			workspaceFileDirectoryAbsPath, err := findTerminateFileDirectoryPathFromOS(dirPath, terminateFileNames)
+			if err != nil {
+				return nil, err
+			}
+			if workspaceFileDirectoryAbsPath != "" {
+				// if we find a workspace, then we use the workspace over the module
+				// this avoids the edge case where the working directory is outside of the file referenced
+				// and share a workspace but not a module
+				terminateFileDirectoryAbsPath = workspaceFileDirectoryAbsPath
+			}
 		}
-		terminateFileRelativePath, err := normalpath.Rel(wd, terminateFileDirectoryAbsPath)
-		if err != nil {
-			return nil, err
-		}
-		rootPath := terminateFileRelativePath
-		if filepath.IsAbs(singleFileRef.Path()) {
-			// If the input was provided as an absolute path,
-			// we preserve it by initializing the workspace
-			// bucket with an absolute path.
-			rootPath = terminateFileDirectoryAbsPath
-		}
-		readWriteBucket, err := r.storageosProvider.NewReadWriteBucket(
-			rootPath,
-			storageos.ReadWriteBucketWithSymlinksIfSupported(),
-		)
-		if err != nil {
-			return nil, err
-		}
-		// Verify that the subDirPath exists too.
-		if _, err := r.storageosProvider.NewReadWriteBucket(
-			normalpath.Join(rootPath, dirPath),
-			storageos.ReadWriteBucketWithSymlinksIfSupported(),
-		); err != nil {
-			return nil, err
-		}
-		return newReadWriteBucketCloser(
-			storage.NopReadWriteBucketCloser(readWriteBucket),
-			rootPath,
-			rootPath,
-		)
+	}
+	if terminateFileDirectoryAbsPath == "" {
+		// TODO(doria): determine the correct error to return here
+		return nil, fmt.Errorf("cannot resolve valid workspace for file reference: %s", singleFileRef.Path())
+	}
+	// If the terminate file exists, we need to determine the relative path from the
+	// terminateFileDirectoryAbsPath to the target singleFileRef.Path().
+	wd, err := osextended.Getwd()
+	if err != nil {
+		return nil, err
+	}
+	terminateFileRelativePath, err := normalpath.Rel(wd, terminateFileDirectoryAbsPath)
+	if err != nil {
+		return nil, err
+	}
+	singleFileRefAbsPath, err := normalpath.NormalizeAndAbsolute(dirPath)
+	if err != nil {
+		return nil, err
+	}
+	singleFileRefRelativePath, err := normalpath.Rel(terminateFileDirectoryAbsPath, singleFileRefAbsPath)
+	if err != nil {
+		return nil, err
+	}
+	rootPath := terminateFileRelativePath
+	if filepath.IsAbs(singleFileRef.Path()) {
+		// If the input was provided as an absolute path,
+		// we preserve it by initializing the workspace
+		// bucket with an absolute path.
+		rootPath = terminateFileDirectoryAbsPath
 	}
 	readWriteBucket, err := r.storageosProvider.NewReadWriteBucket(
-		dirPath,
+		rootPath,
 		storageos.ReadWriteBucketWithSymlinksIfSupported(),
 	)
 	if err != nil {
 		return nil, err
 	}
+	// Verify that the file ref path is a sub path of the terminate file root path
+	if _, err := r.storageosProvider.NewReadWriteBucket(
+		normalpath.Join(rootPath, singleFileRefRelativePath),
+		storageos.ReadWriteBucketWithSymlinksIfSupported(),
+	); err != nil {
+		return nil, err
+	}
 	return newReadWriteBucketCloser(
 		storage.NopReadWriteBucketCloser(readWriteBucket),
-		"",
+		rootPath,
 		"",
 	)
 }
@@ -707,7 +726,7 @@ func anyExistsBucket(
 
 // findTerminateFileDirectoryPathFromOS returns the directory that contains
 // the terminateFileName, starting with the subDirPath and ascending until
-// the root of the local filesysem.
+// the root of the local filesystem.
 func findTerminateFileDirectoryPathFromOS(subDirPath string, terminateFileNames []string) (string, error) {
 	if len(terminateFileNames) == 0 {
 		return "", nil
--- a/private/buf/buffetch/ref_parser.go
+++ b/private/buf/buffetch/ref_parser.go
@@ -389,7 +389,8 @@ func newRawRefProcessor(config *rawRefProcessorConfig) func(*internal.RawRef) er
 				compressionType = internal.CompressionTypeGzip
 			case ".git":
 				format = formatGit
-				// TODO(doria): only if the option is turned on...
+				// This only applies if the option accept `SingleFileRef` is passed in, otherwise
+				// it falls through to the `default` case.
 			case ".proto":
 				if config.allowSingleFileRef {
 					format = formatSingleFile
--- a/private/buf/buffetch/ref_parser.go
+++ b/private/buf/buffetch/ref_parser.go
@@ -72,6 +72,7 @@ func newRefParser(logger *zap.Logger) *refParser {
 			internal.WithGitFormat(formatGit),
 			internal.WithDirFormat(formatDir),
 			internal.WithModuleFormat(formatMod),
+			internal.WithSingleFileFormat(formatSingleFile),
 		),
 	}
 }
@@ -191,6 +192,8 @@ func (a *refParser) GetRef(
 		return newSourceRef(t), nil
 	case internal.ParsedModuleRef:
 		return newModuleRef(t), nil
+	case internal.SingleFileRef:
+		return newSourceRef(t), nil
 	default:
 		return nil, fmt.Errorf("unknown ParsedRef type: %T", parsedRef)
 	}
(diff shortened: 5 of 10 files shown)


private/buf/buffetch/buffetch.go
```diff
@@ -157,8 +157,24 @@ type RefParser interface {
 // NewRefParser returns a new RefParser.
 //
 // This defaults to dir or module.
-func NewRefParser(logger *zap.Logger) RefParser {
-	return newRefParser(logger)
+func NewRefParser(logger *zap.Logger, options ...NewRefParserOption) RefParser {
+	return newRefParser(logger, options...)
+}
+
+// NewRefParserOption is an option for NewRefParser. NewRefParserOptions are only accepted
```
bufdev, line 164: `RefParserOption` not `NewRefParserOption` - for constructors, we drop the `New` part

private/buf/buffetch/buffetch.go
```diff
@@ -157,8 +157,24 @@ type RefParser interface {
...
+// for the default RefParser constructor.
+type NewRefParserOption func(*newRefParserOption)
```
bufdev, line 166: For constructors, we generally attach the options directly to the struct type, i.e. `refParser`, so you'll put `allowSingleFileRef` (soon `allowProtoFileRef`) directly on `refParser`.

private/buf/buffetch/buffetch.go
```diff
@@ -157,8 +157,24 @@ type RefParser interface {
...
+
+type newRefParserOption struct {
+	allowSingleFileRef bool
+}
+
+// NewRefParserWithSingleFileRefAllowed sets the NewRefParser to return the single file reference
+// where applicable.
+func NewRefParserWithSingleFileRefAllowed() NewRefParserOption {
```
bufdev, line 174: `RefParserWithProtoFileRefAllowed`

private/buf/buffetch/format.go
```diff
@@ -35,6 +35,8 @@ const (
 	formatTargz = "targz"
 	// formatZip is the zip format.
 	formatZip = "zip"
+	// formatSingleFile is the single file format
+	formatSingleFile = "file"
```
bufdev, line 39: `formatProtoFile = protofile`

private/buf/buffetch/internal/internal.go
```diff
@@ -161,6 +161,14 @@ func NewDirRef(path string) (DirRef, error) {
 	return newDirRef("", path)
 }
 
+// SingleFileRef is a file reference that incorporates a BucketRef ..
```
bufdev, line 164: finish the comment, and `ProtoFileRef`

private/buf/buffetch/internal/internal.go
```diff
@@ -293,6 +301,11 @@ func NewDirectParsedDirRef(format string, path string) ParsedDirRef {
 	return newDirectDirRef(format, path)
 }
 
+type ParsedSingleFileRef interface {
```
bufdev, line 304: godoc

private/buf/buffetch/internal/reader.go
```diff
@@ -348,6 +356,89 @@ func (r *reader) getDirBucket(
...
+func (r *reader) getSingleFileBucket(
+	ctx context.Context,
+	container app.EnvStdinContainer,
+	singleFileRef SingleFileRef,
+	terminateFileNames []string,
+) (ReadBucketCloser, error) {
+	if !r.localEnabled {
+		return nil, NewReadLocalDisabledError()
+	}
+	terminateFileNames = append(terminateFileNames, bufconfig.ExternalConfigV1FilePath)
```
bufdev, line 368: We need to talk about this - `buffetch/internal` was meant to be independent of any buf-specific concepts such as the config file name, this was why we added `terminateFileNames`

private/buf/buffetch/internal/reader.go
```diff
@@ -348,6 +356,89 @@ func (r *reader) getDirBucket(
...
+	dirPath := filepath.Dir(singleFileRef.Path())
+
+	terminateFileDirectoryAbsPath, err := findTerminateFileDirectoryPathFromOS(dirPath, terminateFileNames)
+	if err != nil {
+		return nil, err
+	}
+	if filepath.Base(terminateFileDirectoryAbsPath) == bufconfig.ExternalConfigV1FilePath {
```
bufdev, line 375: same thing - maybe we have to introduce priority order to terminateFileNames?

private/buf/buffetch/internal/reader.go
```diff
@@ -348,6 +356,89 @@ func (r *reader) getDirBucket(
...
+			if workspaceFileDirectoryAbsPath != "" {
+				// if we find a workspace, then we use the workspace over the module
+				// this avoids the edge case where the working directory is outside of the file referenced
+				// and share a workspace but not a module
+				terminateFileDirectoryAbsPath = workspaceFileDirectoryAbsPath
+			}
+		}
+	}
+	if terminateFileDirectoryAbsPath == "" {
+		return nil, fmt.Errorf(`cannot resolve valid module or workspace for reference: "%s"`, singleFileRef.Path())
```
bufdev, line 394: buf-specific error...don't know where to draw the line here, we probably want this error message to eventually be based on `terminateFileNames`

private/buf/buffetch/ref_parser.go
```diff
@@ -309,62 +339,76 @@ func (a *refParser) checkDeprecated(parsedRef internal.ParsedRef) {
...
+				}
+			case ".tgz":
 				format = formatTar
+				compressionType = internal.CompressionTypeGzip
+			case ".git":
+				format = formatGit
+				// This only applies if the option accept `SingleFileRef` is passed in, otherwise
+				// it falls through to the `default` case.
+			case ".proto":
+				if config.allowSingleFileRef {
```
bufdev, line 395: should check if this file is a file or a directory probably via `os.Stat`

private/buf/buffetch/internal/reader.go
doriable: Yeah -- as per the suggestion below, we ended up going with a priority order for terminate files, just added that :)

private/buf/buffetch/internal/reader.go
doriable: We ended up getting rid of this error and returning a bucket that is just the `dirPath`, same behaviour as `dirRef` -- if no terminate files were found, we assume that the input contains a compilable unit. It not, it will simply fail on build. I'm going to QA this behaviour, since I'm not 100% convinced this is the behaviour we want for the `ProtoFileRef`, but yeah, I've replaced this error for now.

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +216,99 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+	// the workspace. If it is, we use the workspace. Otherwise, we use the module.
+	moduleInWorkspace := false
+	if workspaceConfigDirectory != "" && moduleConfigDirectory != "" {
+		workspaceConfig, err := bufwork.GetConfigForBucket(ctx, readBucketCloser, readBucketCloser.RelativeRootPath())
+		if err != nil {
+			return nil, err
+		}
+		for _, dir := range workspaceConfig.Directories {
+			workspaceDirectory := filepath.Join(workspaceConfigDirectory, dir)
+			relativePath, err := filepath.Rel(workspaceDirectory, moduleConfigDirectory)
```
doriable, line 264: Had an offline conversation with @amckinney , going to clean up the use of `normalpath` in the PR.

private/buf/buffetch/internal/internal.go
```diff
@@ -704,18 +766,23 @@ type GetBucketOption func(*getBucketOptions)
 
 // WithGetBucketTerminateFileNames only applies if subdir is specified.
 //
+// The terminate files are organised as a slice of slices of file names.
```
amckinney, line 769:
```suggestion
// The terminate files are organized as a slice of slices of file names.
```

private/buf/buffetch/internal/internal.go
```diff
@@ -704,18 +766,23 @@ type GetBucketOption func(*getBucketOptions)
...
+// Priority will be given to the first slice of terminate file names, and then
+// the next, etc. This allows the `internal` layer to process the terminate file
+// hierarchies without dependencies on the origin of the termiante files.
```
amckinney, line 772:
```suggestion
// hierarchies without dependencies on the origin of the terminate files.
```

amckinney, line 772: We reference the `buf.work.yaml` file in other comments to help make this readable, so feel free to comment with concrete examples rather than being restricted by the generic nature of `buffetch`. See `WithGetBucketTerminateFileNames` for reference.

private/buf/buffetch/internal/internal.go
```diff
@@ -704,18 +766,23 @@ type GetBucketOption func(*getBucketOptions)
...
 // to the terminate file.
 //
-// This is used for workspaces. So if you have i.e. "proto/foo"
+// This is used for workspaces and modules. So if you have i.e. "proto/foo"
 // subdir, and terminate file "proto/buf.work.yaml", the returned bucket will
 // be for "proto", and the SubDirPath will be "foo".
 //
 // The terminateFileNames are expected to be valid and have no slashes.
-func WithGetBucketTerminateFileNames(terminateFileNames ...string) GetBucketOption {
+func WithGetBucketTerminateFileNames(terminateFileNames [][]string) GetBucketOption {
```
amckinney, line 785: Give an example, of what we should expect to see for `terminateFileNames`.

private/buf/buffetch/internal/terminate_file_priority.go
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package internal
+
+var (
+	_ ReadBucketCloserWithTerminateFiles = &readBucketCloserWithTerminateFiles{}
+	_ TerminateFilePriority              = &terminateFilePriority{}
+	_ TerminateFile                      = &terminateFile{}
+)
```
amckinney, line 21: We probably want separate files for these types.

private/buf/buffetch/proto_file_ref.go
```diff
@@ -0,0 +1,70 @@
...
+	return &protoFileRef{
+		protoFileRef: internalProtoFileRef,
+		dirPath:      internalProtoFileRef.Path(),
+	}
+}
+
+func (r *protoFileRef) PathForExternalPath(externalPath string) (string, error) {
+	if r.dirPath == "" {
+		return normalpath.NormalizeAndValidate(externalPath)
+	}
```
amckinney, line 41: Based on the constructor, `dirPath` will always be non-empty, so I don't know if this condition is reachable. Should we just rely on the `protoFileRef` field and drop the `dirPath` field altogether?

private/buf/buffetch/ref_parser.go
```diff
@@ -30,50 +30,62 @@ import (
...
 	}
+	if refParser.allowProtoFileRef {
+		fetchRefParserOptions = append(fetchRefParserOptions, internal.WithProtoFileFormat(formatProtoFile))
+	}
+	refParser.logger = logger.Named("buffetch")
+	refParser.fetchRefParser = internal.NewRefParser(
+		logger,
+		fetchRefParserOptions...,
+	)
+	return refParser
```
amckinney, line 88: Let's rewrite the `RefParserOption` so that it doesn't mutate the `refParser` type. We can define a new un-exported `refParserOptions` type, similar to what you see elsewhere (e.g. [here](https://github.com/bufbuild/buf/blob/60cedc242b5231d92c37841278e223d6308a6637/private/pkg/app/appproto/response_writer.go#L148)).

With this, we don't have to construct the `refParser` up-front, then mutate it here. We can just apply all the options and construct the `refParser` at the end (similar to what we used to have).

And given that this is an un-exported constructor, the options could be applied in the top-level `NewRefParser` and concrete values could be passed here, e.g.

```
newRefParser(logger *zap.Logger, allowProtoFileFormat bool)
```

private/buf/buffetch/ref_parser.go
```diff
@@ -309,62 +323,80 @@ func (a *refParser) checkDeprecated(parsedRef internal.ParsedRef) {
...
-	if rawRef.Path == "-" || app.IsDevNull(rawRef.Path) || app.IsDevStdin(rawRef.Path) || app.IsDevStdout(rawRef.Path) {
-		format = formatBin
-	} else {
-		switch filepath.Ext(rawRef.Path) {
-		case ".bin":
+type rawRefProcessorConfig struct {
+	allowProtoFileRef bool
+}
+
+func newRawRefProcessor(config *rawRefProcessorConfig) func(*internal.RawRef) error {
```
amckinney, line 330: I appreciate the configuration struct, but we can probably get away with a simple `bool` for now (similar to the comment above).

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
 	for {
-		fullTerminateFileNames := make([]string, len(terminateFileNames))
-		for i, terminateFileName := range terminateFileNames {
-			fullTerminateFileNames[i] = normalpath.Join(terminateFileDirectoryPath, terminateFileName)
+		fullTerminateFileNames := make([][]string, len(terminateFileNames))
+		for i := range terminateFileNames {
+			fullTerminateFileNames[i] = make([]string, len(terminateFileNames[i]))
+			for j, terminateFileName := range terminateFileNames[i] {
+				fullTerminateFileNames[i][j] = normalpath.Join(terminateFileDirectoryPath, terminateFileName)
+			}
```
amckinney, line 765: Let's provide some explanation and/or examples of what these pieces of code are doing. In this case, it looks like we're ascending each directory and creating a set of filepaths based on the terminate filenames.

```
// We need to check for the existence of all of the terminate files, so we ascend
// and prepend the current directory to determine the fully-qualified filepaths.
//
// For example,
//   [][]string{
//     ["buf.work.yaml", "buf.work"],
//     ["buf.yaml", "buf.mod"],
//   }
//
//  ==> 
//
//   [][]string{
//     ["/current/path/to/buf.work.yaml", "/current/path/to/buf.work"],
//     ["/current/path/to/buf.yaml", "current/path/to/buf.mod"],
//   }
```

But with all that said, it might be simpler to pass in the current directory to the `terminateFilesInBucket` helper and always prepend the current directory at the call-site:

```go
func terminateFileInBucket(..., currentDirectory string, ...) {
	for i := range paths {
		for _, path := range paths[i] {
			exists, err := storage.Exists(ctx, readBucket, normalpath.Join(currentDirectory, path))
```

If we go with the latter approach, the comment above could be simplified, but we should still describe how this works with some concrete examples.

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
+				// for that hierarchy.
+				if terminateFile != nil {
+					terminateFiles[i] = terminateFile
+					foundFiles++
+				}
+			}
+		}
+		// If we have found a terminate file for each layer, we can return now.
+		if foundFiles == len(terminateFileNames) {
+			return newTerminateFilePriority(terminateFiles), nil
```
amckinney, line 785: Given that we ascend until _all_ of the terminate file hierarchies are found, we should add a test that verifies the single file ref behavior when the nearest `buf.yaml` is a parent to the nearest `buf.work.yaml` (another example when the module is not defined in the workspace).

For this to compile correctly, you might need to leverage `build.excludes` so that the same `.proto` files aren't included twice (once by the `buf.yaml` and again by the `buf.work.yaml`).

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
 	for {
-		fullTerminateFileNames := make([]string, len(terminateFileNames))
-		for i, terminateFileName := range terminateFileNames {
-			fullTerminateFileNames[i] = normalpath.Unnormalize(normalpath.Join(terminateFileDirectoryPath, terminateFileName))
+		fullTerminateFileNames := make([][]string, len(terminateFileNames))
+		for i := range terminateFileNames {
+			fullTerminateFileNames[i] = make([]string, len(terminateFileNames[i]))
+			for j, terminateFileName := range terminateFileNames[i] {
+				fullTerminateFileNames[i][j] = normalpath.Unnormalize(normalpath.Join(terminateFileDirectoryPath, terminateFileName))
+			}
```
amckinney, line 850: Similar to the other path transformation comment, we might be able to just pass in the current directory to `terminateFileOnOS` and un-normalize the path at the `os.Stat` call-site.

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
+		foundTerminateFiles, err := terminateFilesOnOS(fullTerminateFileNames)
 		if err != nil {
-			return "", err
+			return nil, err
 		}
-		if exists {
-			return terminateFileDirectoryPath, nil
+		// More terminate files were found than layers of hierarchy, return an error.
+		if len(foundTerminateFiles) > len(terminateFileNames) {
+			return nil, fmt.Errorf("more than one terminate file found per level of prioritisation: %v", foundTerminateFiles)
```
amckinney, line 858:
```suggestion
			return nil, fmt.Errorf("more than one terminate file found per level of prioritization: %v", foundTerminateFiles)
```

private/buf/buffetch/internal/internal.go
```diff
@@ -356,6 +370,31 @@ func NewRefParser(logger *zap.Logger, options ...RefParserOption) RefParser {
...
+
+// TerminateFile has the config file name and the path found.
+// The Path is always expressed as a normalpath and it is the caller's responsibility to unnormalise
+// it as needed.
+type TerminateFile interface {
+	// ConfigFile returns the string of the config file type of the terminate file (e.g. "buf.work.yaml")
+	ConfigFile() string
+	// Path returns the normalised directory path where the terminate file is found.
+	Path() string
+}
```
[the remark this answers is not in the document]
doriable: Nice, I'm a big fan of this naming change :)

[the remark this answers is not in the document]
doriable: As a sidenote, it is interesting to note that we have multiple `FileInfo`-like interfaces. I don't think we need to consolidate here, but I think it's probably worth thinking about for future features, etc.

private/buf/buffetch/ref_parser.go
doriable: This was based on a previous comment: https://github.com/bufbuild/buf/pull/608#discussion_r716991316.
The original implementation of this used an intermediary options type.

private/buf/buffetch/internal/reader.go
```diff
@@ -379,30 +524,50 @@ func (r *reader) getGitBucket(
...
+	terminateFileProvider, err := findTerminateFileDirectoryPathFromBucket(ctx, readWriteBucket, subDirPath, terminateFileNames)
 	if err != nil {
 		return nil, err
 	}
+	terminateFileDirectoryPath := ""
+	// Get the highest priority file found and use it as the terminate file directory path.
+	terminateFiles := terminateFileProvider.GetTerminateFiles()
+	if len(terminateFiles) != 0 && terminateFiles[0] != nil {
+		terminateFileDirectoryPath = terminateFiles[0].Path()
+	}
```
amckinney, line 536: We're performing this same logic several times - should we expose a `GetHighestPriority` method on the `TerminateFileProvider` so that we can re-use it elsewhere?

doriable: Yeah, I think that is reasonable, since it is encapsulated by the `TerminateFileProvider` and is used for all the different refs :D

amckinney: Now that I think about this more, I guess it would really just be a helper to resolve the first file in the slice.. and that wouldn't be very helpful. Indexing into the first element seems fine as-is.

private/buf/buffetch/internal/reader.go
doriable: Yeah, we can definitely expand on this in terms of comments.
As for expanding the helper... I think that is reasonable, if only to encapsulate the comments in a single place, since the helper isn't really a helper function in that it isn't really being reused (same with the `terminateFileInOS` "helper", lol). If I re-define the helper, I'll do the same with the `terminateFileInOS` helper since it serves the same functionality, except on-disk.

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
+) ([]TerminateFile, error) {
+	foundPaths := make([]TerminateFile, len(paths))
+	for i := range paths {
+		for _, path := range paths[i] {
+			exists, err := storage.Exists(ctx, readBucket, path)
+			if err != nil {
+				return nil, err
+			}
+			if exists {
+				foundPaths[i] = newTerminateFile(filepath.Base(path), normalpath.Normalize(filepath.Dir(path)))
```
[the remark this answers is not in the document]
doriable: Ah yeah, we can.

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +748,156 @@ func findTerminateFileDirectoryPathFromBucket(
...
-		exists, err := anyExistsOS(fullTerminateFileNames)
+		foundTerminateFiles, err := terminateFilesOnOS(fullTerminateFileNames)
 		if err != nil {
-			return "", err
+			return nil, err
 		}
-		if exists {
-			return terminateFileDirectoryPath, nil
+		// More terminate files were found than layers of hierarchy, return an error.
+		if len(foundTerminateFiles) > len(terminateFileNames) {
```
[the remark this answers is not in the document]
doriable: Yes, exactly, this is currently unreachable in the code since the surface area for `terminateFiles` is isolated to the option, and we are passing the option in only one situation, but yeah, I included this check in case there are bad callers in the future. I will add the comment for unreachable in the current code :D

private/buf/buffetch/internal/reader.go
```diff
@@ -338,19 +380,122 @@ func (r *reader) getDirBucket(
...
+		terminateFileProvider,
+	), nil
+}
+
+func (r *reader) getProtoFileBucket(
+	ctx context.Context,
+	container app.EnvStdinContainer,
+	protoFileRef ProtoFileRef,
+	terminateFileNames [][]string,
+) (ReadBucketCloserWithTerminateFileProvider, error) {
```
amckinney, line 402: Is there any way we can consolidate some of this functionality with the OS logic above? There's a fair amount of duplication, so it would be great to DRY this up a bit. If not, we can discuss and comment as needed, but it might be worth checking it out a bit.

private/buf/buffetch/internal/reader.go
doriable: Lol sounds good -- yeah, I implemented the change and it turned out to be the same LOC (+ the extra LOC for defining the helper itself and extending the interface, lol). I can try and take a generally more critical approach around DRY-ing up this file.

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +216,99 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+			}
+			if _, ok := moduleConfigs[terminateFile.Name()]; ok {
+				moduleConfigDirectory = normalpath.Unnormalize(terminateFile.Path())
+			}
+		}
+	}
+	// If a workspace and module are both found, then we need to check of the module is within
+	// the workspace. If it is, we use the workspace. Otherwise, we use the module.
+	moduleInWorkspace := false
+	if workspaceConfigDirectory != "" && moduleConfigDirectory != "" {
```
amckinney, line 257: We should perform this check even if `moduleConfigDirectory == nil` right? What about the case when the workspace contains the directory, but a `buf.yaml` does not exist?

```yaml
# buf.work.yaml
version: v1
directories:
  - foo
```

```
$ buf lint foo/bar/baz.proto
```

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +216,99 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+		workspaceConfig, err := bufwork.GetConfigForBucket(ctx, readBucketCloser, readBucketCloser.RelativeRootPath())
+		if err != nil {
+			return nil, err
+		}
+		for _, dir := range workspaceConfig.Directories {
+			workspaceDirectory := filepath.Join(workspaceConfigDirectory, dir)
+			relativePath, err := filepath.Rel(workspaceDirectory, moduleConfigDirectory)
+			if err != nil || relativePath == "." || filepath.IsAbs(relativePath) {
```
amckinney, line 265: It looks like we deem that the module is contained if the workspace if we _cannot_ determine a relative path between the two (i.e. the `err != nil` case)? Also, why do we care if `filepath.IsAbs`? 

Could we use some form of the `workspaceDirectoryEqualsOrContainsSubDirPath` helper that already exists in this package?

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +216,99 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+	}
+	moduleConfigs := map[string]struct{}{}
+	for _, moduleConfigFile := range bufconfig.AllConfigFilePaths {
+		moduleConfigs[moduleConfigFile] = struct{}{}
+	}
+	terminateFileProvider := readBucketCloser.TerminateFileProvider()
+	workspaceConfigDirectory := ""
+	moduleConfigDirectory := ""
+	for _, terminateFile := range terminateFileProvider.GetTerminateFiles() {
+		if terminateFile != nil {
```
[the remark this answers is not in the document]
doriable: Well... we use `make([]TerminateFile, len(terminateFileNames)` when we instantiate the slice for terminate files, so I believe it would range over the slice, even if the terminate file did not end up being populated: https://play.golang.org/p/hV1Czb77_Pg Thus we need to do a `nil` check against the interface.

private/buf/bufwire/module_config_reader.go
doriable: This check is only specifically looking at if the `moduleConfigDirectory` is a part of any of the directories defined in the workspace. Otherwise, we build the workspace if the workspace config was found, or we build the module otherwise. We could do an additional check here for the ref, but that should probably be done at the `internal/reader` layer.

private/buf/bufwire/module_config_reader.go
doriable: Yeah, I think I should split this out into a helper because this is a bit confusing. The absolute path check is based on the case where the relative path is `../module/is/outside/workspace/dir`, but yeah, that is probably not a great heuristic. I'll find a better way to capture this in a helper.

private/buf/bufwire/module_config_reader.go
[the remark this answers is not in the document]
amckinney: Even if that's the case, we should see how we can refactor things so that we never let this happen. Otherwise users of this abstraction will need to be very careful to not `nil`-dereference the value which will lead to mistakes.

In other words, if I get at least one result from `GetTerminateFiles`, I should be able to access its values without checking if it's `nil`. For example, imagine if I had to check if the `storage.ObjectInfo` was non-`nil` every time I called `readBucket.Walk`?

I realize we aren't initializing it with `make([]TerminateFile, 0, len(terminateFileNames)` since we actually care about indexing into the slice without using `append` (i.e. the position matters here), but there might be something we can do with respect to how its used to initialize the `TerminateFileProvider`. Worst case, we can re-initialize another slice of `[]TerminateFile` without the `nil` elements before it's passed into `newTerminateFileProvider`, but there might be something else we can do to rearrange it a bit.

private/buf/buffetch/internal/reader.go
doriable: Add the comment, but ended up not using a helper, since the pathing is a bit different for bucket vs. os (`normalpath.Unnormalize` is called against OS paths), so I've left them. But I added this comment for both functions.

private/buf/bufwire/module_config_reader.go
doriable: `workspaceDirectoryEqualsOrContainsSubDirPath` does the trick :)

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +216,97 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+		return nil, err
+	}
+	workspaceConfigs := map[string]struct{}{}
+	for _, workspaceConfigFile := range bufwork.AllConfigFilePaths {
+		workspaceConfigs[workspaceConfigFile] = struct{}{}
+	}
+	moduleConfigs := map[string]struct{}{}
+	for _, moduleConfigFile := range bufconfig.AllConfigFilePaths {
+		moduleConfigs[moduleConfigFile] = struct{}{}
+	}
```
amckinney, line 239: We can use the `stringutil.SliceToMap` helper for these.

private/buf/buffetch/internal/internal.go
```diff
@@ -704,18 +766,27 @@ type GetBucketOption func(*getBucketOptions)
 
 // WithGetBucketTerminateFileNames only applies if subdir is specified.
 //
-// This says that if given a subdir, ascend directories until you reach
+// The terminate files are organized as a slice of slices of file names.
+// Priority will be given to the first slice of terminate file names, which will be workspace
+// configuration files. The second layer of priority will be given to modules.
+//
+// This says that for a given subdir, ascend directories until you reach
 // a file one of these names, and if you do, the returned bucket will be
```
amckinney, line 774: nit: typo from the past
```suggestion
// a file with one of these names, and if you do, the returned bucket will be
```

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +701,190 @@ func findTerminateFileDirectoryPathFromBucket(
...
 		parent := normalpath.Dir(terminateFileDirectoryPath)
 		if parent == terminateFileDirectoryPath {
-			return "", nil
+			break
 		}
 		terminateFileDirectoryPath = parent
 	}
+	// The number of terminate files found is less than the number of layers of terminate files
+	// we are accepting, so we must prune the nil values from the initial instantiation of the slice.
+	prunedTerminateFilesList := []TerminateFile{}
```
amckinney, line 848: For default slice instantiation, prefer `var prunedTerminateFiles []TerminateFile`.

private/buf/buffetch/proto_file_ref.go
```diff
@@ -0,0 +1,68 @@
...
+func newProtoFileRef(internalProtoFileRef internal.ProtoFileRef) *protoFileRef {
+	return &protoFileRef{
+		protoFileRef: internalProtoFileRef,
+	}
+}
+
+func (r *protoFileRef) PathForExternalPath(externalPath string) (string, error) {
+	if r.protoFileRef.Path() == "" {
+		return normalpath.NormalizeAndValidate(externalPath)
+	}
```
amckinney, line 39: We should be able to get rid of this case.

private/buf/buffetch/proto_file_ref.go
```diff
@@ -0,0 +1,68 @@
...
+	protoFileRef internal.ProtoFileRef
+}
+
+func newProtoFileRef(internalProtoFileRef internal.ProtoFileRef) *protoFileRef {
+	return &protoFileRef{
+		protoFileRef: internalProtoFileRef,
+	}
+}
+
+func (r *protoFileRef) PathForExternalPath(externalPath string) (string, error) {
```
amckinney, line 36: We should make it clear that `PathForExternalPath` can only return one of two values - `".", nil` in the successful case, or `nil, <$some_error>` in the error case. In short, the `externalPath` must be equivalent to the `protoFileRef` - otherwise we won't be able to create a valid relative path from it. 

We could even adapt the implementation so that it's impossible for an `externalPath` to yield anything else (which _should_ already be impossible with how this is used). Again, either the `externalPath is equal to `protoFileRef` and `".", nil` should be returned, or we should return an error.

 For example, if the `protoFileRef` is equal to `foo/bar/baz.proto` and the `externalPath is `foo/bar/baz.proto/qux.proto`, this function will still work as-is since it would treat `baz.proto` as a directory in the latter example.

private/buf/buffetch/internal/reader.go
```diff
@@ -240,117 +247,208 @@ func (r *reader) getArchiveBucket(
...
 	}
-	terminateFileDirectoryPath, err := findTerminateFileDirectoryPathFromBucket(ctx, readWriteBucket, subDirPath, terminateFileNames)
+	terminateFileProvider, err := findTerminateFileDirectoryPathFromBucket(ctx, readWriteBucket, subDirPath, terminateFileNames)
 	if err != nil {
 		return nil, err
 	}
+	terminateFileDirectoryPath := ""
+	// Get the highest priority file found and use it as the terminate file directory path.
+	terminateFiles := terminateFileProvider.GetTerminateFiles()
+	if len(terminateFiles) != 0 && terminateFiles[0] != nil {
```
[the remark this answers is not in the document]
doriable: Removed from here (and subsequent get bucket functions that also have the same check).

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +701,190 @@ func findTerminateFileDirectoryPathFromBucket(
...
 		parent := normalpath.Dir(terminateFileDirectoryPath)
 		if parent == terminateFileDirectoryPath {
-			return "", nil
+			break
 		}
 		terminateFileDirectoryPath = parent
 	}
+	// The number of terminate files found is less than the number of layers of terminate files
+	// we are accepting, so we must prune the nil values from the initial instantiation of the slice.
+	prunedTerminateFiles := []TerminateFile{}
```
amckinney, line 741: For slices: `var prunedTerminateFiles []TerminateFile`

private/buf/buffetch/internal/reader.go
```diff
@@ -583,101 +701,190 @@ func findTerminateFileDirectoryPathFromBucket(
...
```
amckinney, line 848: Same as above.

private/buf/bufwire/module_config_reader.go
```diff
@@ -205,6 +217,89 @@ func (m *moduleConfigReader) getModuleModuleConfig(
...
+			workspaceConfigDirectory = normalpath.Unnormalize(terminateFile.Path())
+			continue
+		}
+		if _, ok := moduleConfigs[terminateFile.Name()]; ok {
+			moduleConfigDirectory = normalpath.Unnormalize(terminateFile.Path())
+		}
+	}
+	// If a workspace and module are both found, then we need to check of the module is within
+	// the workspace. If it is, we use the workspace. Otherwise, we use the module.
+	moduleOutsideWorkspace := true
```
amckinney, line 249: This variable isn't used outside of the following branch, so we can move it down where it's used.

But the variable doesn't actually add much with respect to control flow. We can invert the condition a bit and drop it altogether.

```go
	if normalpath.EqualsOrContainsPath(normalpath.Join(workspaceConfigDirectory, directory), moduleConfigDirectory, normalpath.Absolute) {
		continue
	}
        // The module is outside of the workspace, so we need to override the SubDirPath to the module's path.
	relativePath, err := filepath.Rel(workspaceConfigDirectory, moduleConfigDirectory)
	if err != nil {
		return nil, err
	}
	readBucketCloser.SetSubDirPath(normalpath.Normalize(relativePath))
        break
}
```

amckinney (commented):
Final nits, but this should otherwise be good to go. Now we just need to discuss https://github.com/bufbuild/buf/issues/493#issuecomment-934878891 and see whether or not this feature makes sense across all of our `buf` commands.

private/buf/bufwire/module_config_reader.go
doriable: In the example you provided, we would call `SetSubDirPath` any time we found workspace directory that didn't encapsulate the module -- I think we want to call `SetSubDirPath` only when _none_ of the directories in the workspace encapsulate the module... right?

I moved the `moduleOutsideWorkspace` variable inside the `if` scope though.

PR #625, opened 2021-10-01
add new endpoint GetModulePackages for module's packages list

cyinma:
add new endpoint `GetModulePackages` for module's packages list

Screen Shot 2021-10-01 at 3 59 54 PM

Screen Shot 2021-10-01 at 3 59 59 PM

Currently when we cannot build the image for the module (e.g. one of the dependency is deleted), we cannot retrieve the packages list, and therefore cannot show the `buf.md` too, as they come in the same endpoint `GetModuleDocumentation`. Therefore, would like to make them separately so that we can show the `buf.md` even when we cannot build the image.

There are multiple options for this:
1. Add a new endpoint `GetModulePackages` to provide the packages list only, remove packages list in the current endpoint `GetModuleDocumentation` (As suggested by @doriable , can do this in two steps)
2. Add a new endpoint `GetModuleMarkdown` to provide the `buf.md` only, keep the current fields and implementation in `GetModuleDocumentation`
3. Include `buf.md` in `GetSourceDirectoryInfo` as they both come from the same source that don't need to build the image

I prefer option (1) as it will makes things more clear, separate different endpoints that provide different informations, and does not include some fields that we don't need in the method.

Option (2) can serve the same purpose, but will keeps field that we no longer needed

Option (3) can also serve the same purpose, but as we don't need source directory info in the `DOCS` tab, and they are not actually the same things, it might be better to not group them together.

But this is just my opinion, I am open to discussion and any opinions.


proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -28,6 +28,9 @@ service DocService {
   // GetSourceFile retrieves the source contents for the given owner, repository,
   // reference, and path.
   rpc GetSourceFile(GetSourceFileRequest) returns (GetSourceFileResponse);
+  // GetModulePackages retrieves the list of packages for module based on the given
```
johanbrandhorst, line 31: I think this sounds better?
```suggestion
  // GetModulePackages retrieves the list of packages for the module based on the given
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -87,12 +108,11 @@ message GetModuleDocumentationResponse {
 // ModuleDocumentation provides the name of the module and associated documentation.
 message ModuleDocumentation {
   string name = 1;
-  repeated string packages = 2;
   // This is the string representation of the contents of the buf.md file for module-level documentation.
   //
   // The buf.md file is a part of the module.
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
-  string documentation = 3;
+  string documentation = 2;
```
bufdev, line 115: Changing tags is a breaking change - you can delete a field, but never change a fields' tag

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -87,12 +108,11 @@ message GetModuleDocumentationResponse {
 // ModuleDocumentation provides the name of the module and associated documentation.
 message ModuleDocumentation {
   string name = 1;
-  repeated string packages = 2;
```
bufdev, line 90: Are we sure we want to delete this, or do we just want to add a new method?

cyinma: we want to remove it from the method, so that even when we cannot build the image and get the packages list, we can still provide the documentation

proto/buf/alpha/registry/v1alpha1/doc.proto
SyrieBianco: Oh yeah, I know we used to be down for breaking changes, but now that our API is public, should we not do that?

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -28,6 +28,9 @@ service DocService {
   // GetSourceFile retrieves the source contents for the given owner, repository,
   // reference, and path.
   rpc GetSourceFile(GetSourceFileRequest) returns (GetSourceFileResponse);
+  // GetModulePackages retrieves the list of packages for module based on the given
+  // owner, repository, and reference.
+  rpc GetModulePackages(GetModulePackagesRequest) returns (GetModulePackagesResponse);
   // GetModuleDocumentation retrieves the documentation for module based on the given
   // owner, repository, and reference.
   rpc GetModuleDocumentation(GetModuleDocumentationRequest) returns (GetModuleDocumentationResponse);
```
doriable, line 36: I like the separation in this way, my only concern is that we need to make sure that we are changing the frontend calls at the same time (and make sure that we deploy the frontend changes at the same time), since we are changing the nature of an existing endpoint and also making a breaking change to the field.

I think I would be more comfortable if this is an incremental change: support the package list for both endpoints first, and then migrate the change to remove the package list... and also, I don't think we need to change the tag number from `3->2` for the documentation... we can simply remove the old `repeated packages = 2;` field... but this is just my opinion.

proto/buf/alpha/registry/v1alpha1/doc.proto
bufdev: It will just make the frontend update harder, but yea if you're willing to take that on, that's fine

cyinma: I think we can just keep the tag and avoid the breaking change

cyinma:
this should be ready for another round of review

PR #629, opened 2021-10-04
remove packages in ModuleDocumentation

cyinma:
This is to remove packages in ModuleDocumentation, so that we will use separated endpoint for `buf.md` and packages list, and will be able to show `buf.md` even if we cannot build the image

previous PR: #625


proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -108,7 +108,6 @@ message GetModuleDocumentationResponse {
 // ModuleDocumentation provides the name of the module and associated documentation.
 message ModuleDocumentation {
   string name = 1;
-  repeated string packages = 2;
```
doriable, line 111: I assume the frontend changes have not been made to remove the dependency on this yet -- I will stamp, but try not to merge until the frontend is ready, since if `core` gets updated, then it would break the frontend ^^"

doriable (approved):

Issue #636, opened 2021-10-06
Too many files are being opened during concurrent generation


amckinney:
Now that we're executing local plugins in parallel, we discovered that we are opening too many files at the same time. By default, Macbooks have a `ulimit` equal to `256`. So regardless of parallel execution, there are situations where users will not be able to generate because they simply have too many files to generate.

We'll need to revisit how many files we can open in parallel and see if we can prevent this from happening in general.

Note that if people have fewer than `256` files to generate, they can get around this by setting `strategy: all`.

PR #637, opened 2021-10-06
Print a special error message if too many files are opened during generation

bufdev:
Fixes https://github.com/bufbuild/buf/issues/636 somewhat.

This prints a special warning if too many files are opened when invoking plugins. We can't control how many files these plugins open directly, although we could later have special logic that limits the amount of parallel execs depending on the file limit. Users should adjust their limits, especially as Mac's limits are so low.

### DIFF
--- a/private/pkg/app/appproto/appprotoexec/util.go
+++ b/private/pkg/app/appproto/appprotoexec/util.go
@@ -1,0 +1,46 @@
+// Copyright 2020-2021 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package appprotoexec
+
+import (
+	"errors"
+	"fmt"
+	"os"
+)
+
+// handlePotentialTooManyFilesError checks if the error is a result of too many files
+// being open, and if so, modifies the output error with a help message.
+//
+// This could potentially go in osextended, but we want to provide a specific help
+// message referencing StrategyAll, so it is simplest to just put this here for now.
+func handlePotentialTooManyFilesError(err error) error {
+	if isTooManyFilesError(err) {
+		return fmt.Errorf("%w: %s", err, tooManyFilesHelpMessage)
+	}
+	return err
+}
+
+func isTooManyFilesError(err error) bool {
+	var syscallError *os.SyscallError
+	if errors.As(err, &syscallError) {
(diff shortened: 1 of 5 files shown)


private/pkg/app/appproto/appprotoexec/util.go
```diff
@@ -0,0 +1,46 @@
...
+	if isTooManyFilesError(err) {
+		return fmt.Errorf("%w: %s", err, tooManyFilesHelpMessage)
+	}
+	return err
+}
+
+func isTooManyFilesError(err error) bool {
+	var syscallError *os.SyscallError
+	if errors.As(err, &syscallError) {
+		// This may not actually be correct on other platforms, however the worst cast
```
doriable, line 38:
```suggestion
		// This may not actually be correct on other platforms, however the worst case
```

private/pkg/app/appproto/appprotoexec/util_darwin.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build darwin
+// +build darwin
+
+package appprotoexec
+
+const tooManyFilesHelpMessage = `This is commonly caused by the maximum file limut being too low. On Mac, the default is 256, which is very low. Run "ulimit -n" to check your file limit. If this happened on generation, setting "strategy: all" for each configured plugin in your buf.gen.yaml can mitigate the issue if you are unable to change your file limit.`
```
doriable, line 20:
```suggestion
const tooManyFilesHelpMessage = `This is commonly caused by the maximum file limit being too low. On Mac, the default is 256, which is very low. Run "ulimit -n" to check your file limit. If this happened on generation, setting "strategy: all" for each configured plugin in your buf.gen.yaml can mitigate the issue if you are unable to change your file limit.`
```

private/pkg/app/appproto/appprotoexec/util_undarwin.go
```diff
@@ -0,0 +1,20 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !darwin
+// +build !darwin
+
+package appprotoexec
+
+const tooManyFilesHelpMessage = `This is commonly caused by the maximum file limut being too low. Run "ulimit -n" to check your file limit. If this happened on generation, setting "strategy: all" for each configured plugin in your buf.gen.yaml can mitigate the issue if you are unable to change your file limit.`
```
doriable, line 20:
```suggestion
const tooManyFilesHelpMessage = `This is commonly caused by the maximum file limit being too low. Run "ulimit -n" to check your file limit. If this happened on generation, setting "strategy: all" for each configured plugin in your buf.gen.yaml can mitigate the issue if you are unable to change your file limit.`
```

doriable (approved):
This seems mostly fine to me -- question on supporting `!darwin` systems: given that the check is based on `syscall.Err.Error()` being matched on `too many open files`, should we instead support `linux, etc.` for the error message excluding `ulimit` and just do a no-op on systems like Windows? I don't feel strongly about it, but just throwing that out there.

PR #658, opened 2021-10-13
Add org member related endpoints, authz endpoints, modify list org user response

cyinma:
Add org member related endpoints, authz endpoint, modify list org user response

This is to add multiple endpoints for new role-based authorization, including:

`OrganizationService`:

1. `AddMember`
3. `UpdateMember`
4. `RemoveMember`

`UserService`

1. `UpdateUserServerRole`

`AuthzService`:

1. `UserCanAddOrganizationMember`
2. `UserCanUpdateOrganizationMember`
3. `UserCanRemoveOrganizationMember`
4. `UserCanDeleteOrganization`
5. `UserCanDeleteRepository`
6. `UserCanDeleteTemplate`
7. `UserCanDeletePlugin`

and modify response of endpoint:

`UserService`:

1. `ListOrganizationUsers`

Things that open to discussion:

1. Where should we place the role-related endpoints? 
2. Better approach / naming of `GroupRole` and `ResourceRole`?

For (1), the previous approach of scope-related endpoints is to place them under `UserService`, however, it seems to be making the `UserService` so huge and unrelated. Therefore, was thinking to put them under `OrganizationService`, `RepositoryService`... for the respective role endpoints. However, roles seems not completely belongs to the organization/repository, therefore, I am more prefer to create a new RoleService that group all role-related endpoints, and separate them from `OrganizationService`, `RepositoryService`.. , but also would like to ask for opinions on this approach.

For (2), I am trying to make the role enum more generic, so that organization role and team role can use `GroupRole`, and repository role, template role and plugin role can use `ResourceRole`, would like to ask for opinions on this approach and naming.

These are just my opinions, I am open to any opinions and discussions.


robbertvanginkel:
On the 2 points: 

1. I think our grpc endpoints shouldn't resemble crud operations on rows our database model, but instead represent actions you would do on the entities. From this perspective, I think I'd prefer something like `OrganizationService.{Add,Remove,Update}Member`, rather than a `RoleService.{Add,Update}...Role`.

2. From 1, I think I would avoid using a generic roles enum across resources. If for any reason in the future we want plugins to have a special role, the current design does not make that easy. I think we should value flexibility of extension over time over conciseness today, this is already something we do in general by requiring every service method to have its own request/response message.

PR #659, opened 2021-10-13
Do not print imports for image file ref for `buf ls-files`

doriable:
Currently, when given an image ref, we are printing out the imports for `buf ls-files`,
even when the `--include-imports`﻿flag is not set. This fixes that behaviour.


amckinney:
Please add a `CHANGELOG.md` entry.

PR #670, opened 2021-10-19
Update changelog in for #608

doriable:


doriable:
> I'd actually probably expand on this even more but this is fine

Sounds good, I read through some of the other `CHANGELOG` entries and modelled it after those.

Issue #224, opened 2020-12-29
Allow excludes for buf generate


zerospiel:
Hi! I have, for sure, pretty common protoc generation flow but when I tried to migrate from generic `buf protoc` or `buf image && protoc`, I've got an unexpected behavior of `buf generate` subcommand.

## Preambula

Let's explore the next simple `file.proto`:

```protobuf
syntax = "proto3";

package some.service.v1;
option go_package = "github.com/user/repo/pkg;some_service";

import "google/api/annotations.proto";

service SomeService {
    rpc Get(Response) returns (Response) {
        option (google.api.http) = {
            get: "/version"
        };
    }
}

message Response {
    string some_name = 1;
}

message Request {}
```

The file structure then looks like this:

```text
├─ protos/
│       ├─ google/api/
│       │         ├─ annotations.proto
│       │         ├─ http.proto
│       ├─ file.proto
```

And to generate files I use the following command, assuming I'm generating for `golang` using vanilla plugins and have module `github.com/user/repo`:

```bash
# here I mitigate some mappings, they do not matter
buf protoc --proto_path ./protos/ \
   --go_out=module=github.com/user/repo:. \
   --go-grpc_out=module=github.com/user/repo:. \
   ./protos/file.proto
```

And then I have perfectly and insanely fast generated files:

```text
├─ pkg/
│    ├─ file.pb.go
│    ├─ file_grpc.pb.go
```

## Issue

Now let's deep dive into the issue! Assume I have the following simple `buf.yaml` contents:

```yaml
version: v1beta1
build:
  roots:
    - protos
lint:
  use:
    - MINIMAL
  except:
    - PACKAGE_DIRECTORY_MATCH
  ignore:
    - google
```

And then I try to reproduce the above generation using the next `buf.gen.yaml` template:

```yaml
version: v1beta1

plugins:
  - name: go
    out: .
    opt: module=github.com/user/repo
    strategy: directory
  - name: go-grpc
    out: .
    opt: module=github.com/user/repo
    strategy: directory
```

And now using `buf generate .` I've started to get the following error:

```bash
plugin go: google.golang.org/genproto/googleapis/api/annotations/http.pb.go: generated file does not match prefix "github.com/user/repo"
```

Of course, this is because under the hood [buf trying to generate](https://docs.buf.build/generate-usage) (see `strategy` description) from all directories within `roots` from `buf.yaml`. But I need that protos only for vendor purposes without any generation for those files.

[...]

PR #689, opened 2021-10-28
Adding an exclude flag to `buf generate`

doriable:
This adds an `--exclude` flag to **only** the `buf generate` command. The flag is passed through the `imageConfigReader` and is handled by the module builder for source and module references, and by the image reader for image references.

For cases where both the `--path` flags and `--exclude`, the more specific case would win. Examples:

```
$ buf generate --path a/foo.proto --exclude a/  # the path a/foo.proto will be generated
$ buf generate --path a --exclude a/foo.proto  # everything in a/ will be generated EXCEPT a/foo.proto
$ buf generate --path a/foo.proto --exclude a/foo.proto # an error is returned, the user cannot set the same path as both a target and an exclude
```

References: #224


private/buf/bufwire/image_reader.go
```diff
@@ -130,20 +131,32 @@ func (i *imageReader) GetImage(
...
-	if len(externalDirOrFilePaths) == 0 {
+	if len(externalDirOrFilePaths) == 0 && len(excludeDirOrFilePaths) == 0 {
 		return image, nil
 	}
 	imagePaths := make([]string, len(externalDirOrFilePaths))
-	for i, externalDirOrFilePath := range externalDirOrFilePaths {
-		imagePath, err := imageRef.PathForExternalPath(externalDirOrFilePath)
-		if err != nil {
-			return nil, err
+	if len(externalDirOrFilePaths) > 0 {
```
amckinney, line 138: Can we drop this condition? We won't `range` over anything if the length is not greater than zero. Same thing several lines below.

private/buf/bufwork/bufwork.go
```diff
@@ -141,13 +143,51 @@ func BuildOptionsForWorkspaceDirectory(
...
 	workspaceID := filepath.Join(normalpath.Unnormalize(relativeRootPath), ExternalConfigV1FilePath)
+	// Note that subDirRelPaths can be empty. If so, this represents
+	// the case where externalDirOrFilePaths were provided, but none
+	// matched.
+	subDirRelPaths, err := externalPathsToSubDirRelPaths(
+		workspaceID,
+		relativeRootPath,
+		subDirPath,
+		externalDirOrFilePaths,
+	)
```
amckinney, line 161: Big fan of this refactor - this is much easier to read with `externalPathsToSubDirRelPaths` now 👏

private/buf/bufwork/bufwork.go
```diff
@@ -141,13 +143,51 @@ func BuildOptionsForWorkspaceDirectory(
...
+	if err != nil {
+		return nil, err
+	}
+	if externalDirOrFilePathsAllowNotExist {
+		buildOptions = append(buildOptions, bufmodulebuild.WithPathsAllowNotExist(subDirRelPaths))
+	} else {
+		buildOptions = append(buildOptions, bufmodulebuild.WithPaths(subDirRelPaths))
+	}
```
amckinney, line 169: We should still gate this (the `subDirRelPaths` option) against `len(externalDirOrFilePaths) > 0` right? I'm thinking about the case where we have at least one `--exclude` path, but no `--path` paths.

If there isn't one already, let's make sure we add a test for this in a workspace environment (since that's where it will be relevant).

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -243,6 +245,13 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		`The config file or data to use.`,
 	)
+	flagSet.StringSliceVar(
+		&f.ExcludePaths,
+		excludePathsFlagName,
+		nil,
+		`Exclude these paths from the generation process. This can be proto files or directories.
+If specified multiple times, the union will be taken.`,
+	)
```
amckinney, line 254: We probably want to take inspiration from `bufcli.BindPathsAndDeprecatedFiles` here:

```
	flagSet.StringSliceVar(
		pathsAddr,
		pathsFlagName,
		nil,
		`Limit to specific files or directories, for example "proto/a/a.proto" or "proto/a".
If specified multiple times, the union will be taken.`,
	)

```

Should we just do something like:
```suggestion
	flagSet.StringSliceVar(
		&f.ExcludePaths,
		excludePathsFlagName,
		nil,
		`Exclude specific files or directories, for example "proto/a/a.proto" or "proto/a".
If specified multiple times, the union will be taken.`,
	)
```

private/buf/cmd/buf/command/generate/generate_test.go
```diff
@@ -154,6 +154,26 @@ func TestProtoFileRefIncludePackageFiles(t *testing.T) {
...
 }
 
+func TestOutputWithPathEqualToExclude(t *testing.T) {
+	tempDirPath := t.TempDir()
+	testRunStdoutStderr(
+		t,
+		nil,
+		1,
+		``,
+		filepath.FromSlash(`Failure: cannot set the same path for both --path and --exclude flags: a/v1/a.proto`),
```
amckinney, line 164: Nice, this error looks great 👍

private/bufpkg/bufimage/util.go
```diff
@@ -71,23 +140,42 @@ func imageWithOnlyPaths(image Image, fileOrDirPaths []string, allowNotExist bool
...
+		// an ImageFile for all fileOrDirPaths, so we can return an Image now.
+		// This means we do not have to do the expensive O(image.Files()) operation
+		// to check to see if each file is within a potential directory path.
+		//
+		// We do not need to check the excluded paths for the allowNotExist flag because all target
+		// paths were image files, therefore the exclude paths would not apply in this case.
+		// TODO: should we be checking exclude paths that would not apply in this case against
+		// the allowNotExist flag? If so, we would need to do an expensive operation across all
+		// image.Files(). We could have early breaks and returns, similar to the check above, but
+		// this may not be worth it for this case.
```
amckinney, line 153: Gah yea this one is a bummer - I think we should still validate that all of the `exclude` paths are valid if `allowNotExist` is not specified. Otherwise users will have silent failures with respect to their `--exclude` flags.

private/bufpkg/bufimage/util.go
```diff
@@ -32,15 +32,81 @@ import (
...
+						break
+					}
+				}
+				if !foundPath {
+					pathsWithNoMatchingFiles = append(pathsWithNoMatchingFiles, excludeFileOrDirPath)
+				}
+			}
+			if len(pathsWithNoMatchingFiles) > 0 {
+				return nil, fmt.Errorf("paths with no matching files in the image: %v", pathsWithNoMatchingFiles)
+			}
```
amckinney, line 89: In the `--path` handling below, it looks like we only report the first offending file (not all of them). For simplicity and symmetry, feel free to do the same here (and below). For reference, this is what we have for `--path`:

```
	if !allowNotExist {
		for potentialDirPath := range potentialDirPathMap {
			if _, ok := matchingPotentialDirPathMap[potentialDirPath]; !ok {
				// no match, this is an error given that allowNotExist is false
				return nil, fmt.Errorf("path %q has no matching file in the image", potentialDirPath)
			}
		}
```

private/bufpkg/bufimage/util.go
```diff
@@ -118,6 +206,26 @@ func imageWithOnlyPaths(image Image, fileOrDirPaths []string, allowNotExist bool
...
+					break
+				}
+			}
+			if !foundPath {
+				pathsWithNoMatchingFiles = append(pathsWithNoMatchingFiles, excludeFileOrDirPath)
+			}
+		}
+		if len(pathsWithNoMatchingFiles) > 0 {
+			return nil, fmt.Errorf("paths with no matching files in the image: %v", pathsWithNoMatchingFiles)
+		}
```
amckinney, line 228: Could we reuse the `matchingPotentialDirPathMap` to validate the `--exclude` flags, too? We just need to verify that the paths specified by `--exclude` were matched, so I feel like we should be able to reuse it, right?

If so, this might simplify the case above, too.

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -158,17 +158,33 @@ func NewModuleForProto(
...
-	return newTargetingModule(module, targetPaths, true)
+func ModuleWithPathsAllowNotExist(
+	module Module,
+	targetPaths []string,
+	excludePaths []string,
+) (Module, error) {
+	return newTargetingModule(module, targetPaths, true, excludePaths, false)
+}
+
+func ModuleWithOnlyExcludePaths(
```
amckinney, line 182: nit: Missing documentation.

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -158,17 +158,33 @@ func NewModuleForProto(
...
+	module Module,
+	fileOrDirPathsAllowNotExist bool,
```
amckinney, line 184: We should have a separate function to handle the `fileOrDirPathsAllowNotExist`, i.e. `ModuleWithOnlyExcludePathsAllowNotExist`, right? That way it's symmetrical with `ModuleWithPaths[AllowNotExist]`.

Plus, the `fileOrDirPathsAllowNotExist bool` parameter should be the last one (after `excludePaths`).

private/bufpkg/bufmodule/bufmodulebuild/util.go
```diff
@@ -29,19 +29,31 @@ func applyModulePaths(
 	roots []string,
 	fileOrDirPaths *[]string,
 	fileOrDirPathsAllowNotExist bool,
+	excludeFileOrDirPaths []string,
```
amckinney, line 32: If we're using the same `fileOrDirPathsAllowNotExist` boolean for both `--path` and `--exclude`, we should move the `excludeFileFileOrDirPaths` parameter _above_ `fileOrDirPathsAllowNotExist `.

```suggestion
	fileOrDirPaths *[]string,
	excludeFileOrDirPaths []string,
	fileOrDirPathsAllowNotExist bool,
```

private/bufpkg/bufmodule/bufmodulebuild/module_include_builder.go
```diff
@@ -112,6 +112,8 @@ func (b *moduleIncludeBuilder) buildForIncludes(
 		absIncludeDirPaths,
 		absFileOrDirPaths,
 		fileOrDirPathsAllowNotExist,
+		// TODO(doria): we need to fix this
+		nil,
```
amckinney, line 116: I'm not sure that we do - `buf protoc` doesn't support a way to specify files to exclude since it's all file-oriented (similar to `protoc`).

I think we can get away with an in-line comment like the following:

```suggestion
		nil, // Excluding files isn't supported for the ModuleIncludeBuilder.
```

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -26,22 +26,28 @@ import (
 
 type targetingModule struct {
 	Module
-	targetPaths                    []string
-	targetPathsAllowNotExistOnWalk bool
+	targetPaths              []string
+	pathsAllowNotExistOnWalk bool
+	excludePaths             []string
+	allowAllTargetPaths      bool
```
amckinney, line 32: The `allowAllTargetPaths` parameter is confusing (as well as how it's used in the implementation below. It looks like we only provide it when `ModuleWithOnlyExcludePaths` is used.

Why is it necessary? Is there anything we can do to consolidate these cases when no `--path` flag is provided since those cases _should_ be equivalent, unless I'm missing something.

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -92,53 +106,116 @@ func (m *targetingModule) TargetFileInfos(ctx context.Context) (fileInfos []bufm
...
+			// where we are accepting all targets, then we can just move on. Otherwise, if no file
+			// matching any of the target paths were found, we continue the walk.
+			if !m.allowAllTargetPaths && len(fileMatchingPathMap) == 0 {
+				return nil
+			}
+			// We have checked all conditions where a matching file was found, so we need to handle
+			// if an exclude was found and we are allowing all targets.
+			if m.allowAllTargetPaths && len(excludeMatchingPathMap) > 0 {
+				return nil
+			}
```
amckinney, line 196: I'd like to see if we can simplify these cases regardless of the `m.allowAllTargetPaths` flag (re: the comment above).

private/buf/bufwork/bufwork.go
doriable: Ah yes, that is correct! This should be gated by the `len` check. Good catch. Also, yeah, I need to add some workspace tests for this, lol.

private/bufpkg/bufimage/util.go
doriable: Ah yeah, basically during the `imageFile` ranging, we could also do a check for the excludes. Yeah, I think this actually works out well, since I'm reworking the logic for the file path "merging"/checking.

private/buf/bufwork/bufwork.go
doriable: Oh, I lied, actually, we're not supposed to gate this behind the check because the `subDirRelPaths` can be empty. This has always been the pattern here ^^"

https://github.com/bufbuild/buf/blob/a29fd091b8d5f63931858ba466bab80783468da3/private/buf/bufwork/bufwork.go#L278-L287

private/bufpkg/bufmodule/targeting_module.go
doriable: Yeah, I think we could rely on the fact that `len(targetPaths) == 0`. I was concerned about that being an expected state, but then it's basically impossible to set a `--path` flag without a value, so that should be fine.

private/bufpkg/bufimage/util.go
doriable: Yeah, I've flip-flopped a few times on this one. This is... annoying because we take a nice shortcut for the cases where we only have specific target files because we can just use `if imageFile := image.GetFile(fileOrDirPath); imageFile != nil` or in the case of the module, we just get the storage object directly with `objectInfo, err := sourceReadBucket.Stat(ctx, targetPath)`.

I'm trying to avoid doing a full walk/iterate through all image files just for this verification... but there probably is no way to avoid that...

amckinney: Yea, I see the problem.. it sucks, but we really need to validate this. Otherwise commands like the following will work (I've just tested this on this branch):

```sh
$ buf generate image.bin --path buf/alpha/registry/v1alpha1/authn.proto --exclude_path bogus --template data/template/buf.go.gen.yaml
```

The `--exclude-path` value of `bogus` obviously doesn't exist, but the command continues to work.

We could carve out a separate function that we can call here (and use elsewhere) that simply validates that all of the `exclude-path` flags are valid with respect to the Image. We already do this in the `if len(fileOrDirPaths) == 0 && len(excludeFileOrDirPaths) > 0 {` case above.

private/bufpkg/bufimage/util.go
```diff
@@ -118,11 +203,48 @@ func imageWithOnlyPaths(image Image, fileOrDirPaths []string, allowNotExist bool
...
+				// no match, this is an error given that allowNotExist is false
+				return nil, fmt.Errorf("path %q has no matching file in the image", excludeFileOrDirPath)
+			}
+		}
 	}
 	// we finally have all files that match fileOrDirPath that we can find, make the image
 	return getImageWithImports(image, nonImportPaths, nonImportImageFiles)
 }
 
+func excludeFile(
```
amckinney, line 231: Please include a comment that describes the algorithm from a high-level. We might also want to name this `excludeFiles` since it's excluding more than one file.

private/bufpkg/bufimage/util.go
```diff
@@ -118,11 +203,48 @@ func imageWithOnlyPaths(image Image, fileOrDirPaths []string, allowNotExist bool
...
+	}
+	// For paths that have never been seen for excludes, we should trim them, since they are
+	// outside of the target paths found.
+	for fileMatchingExcludePath := range fileMatchingExcludePathMap {
+		if _, ok := seenExcludePaths[fileMatchingExcludePath]; ok {
+			continue
+		}
+		delete(fileMatchingExcludePathMap, fileMatchingExcludePath)
+	}
+	return excludeFile(fileMatchingPathMap, fileMatchingExcludePathMap)
```
amckinney, line 259: I think the exclusion process you have above makes sense, but I don't think the recursive call buys us anything. From what I can tell, this function would _always_ conclude in a couple recursive calls:

1. All of the relevant files in `fileMatchingPathMap` are removed according to `fileMatchingExcludePathMap`, then `fileMatchingExcludePathMap` shrinks (to approach the base case) according to which paths weren't matched.
2. All of the files that would have been matched by `fileMatchingExcludePathMap` have already been matched (according to (1)). We don't mark any of the `fileMatchingExcludePathMap` as seen, so _all_ of them get deleted.
3. Now, `len(fileMatchingExcludePathMap) == 0`, so we exit via the base case.

In short, I think we can simplify all of this into (1).

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -148,13 +213,53 @@ func (m *targetingModule) TargetFileInfos(ctx context.Context) (fileInfos []bufm
...
+		}
 	}
 	return fileInfos, nil
 }
+
+func excludeFile(
+	fileMatchingPathMap map[string]struct{},
+	fileMatchingExcludePathMap map[string]struct{},
+) bool {
```
amckinney, line 236: It looks like we're defining the same function in two places - can we DRY this up and define this somewhere so that it can be reused?

It's only interacting with `normalpath`, so `normalpath` might be an appropriate home, but somewhere else might better since it's tailored to such a specific use case.

If it moved there, it would need to be named more generically as a filter or something else along those lines, e.g.

```go
// FilterAllEqualOrContainingPathMap returns the elements in the input that do not equal or contain a path in the filter.
func FilterAllEqualOrContainingPathMap(input map[string]struct{}, filter map[string]struct{}, pathType PathType) map[string]struct{} { ... }
```

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -51,94 +54,156 @@ func (m *targetingModule) TargetFileInfos(ctx context.Context) (fileInfos []bufm
...
+	for _, targetPath := range m.targetPaths {
+		for _, excludePath := range m.excludePaths {
+			if excludePath == targetPath {
+				return nil, fmt.Errorf(
+					"cannot set the same path for both --path and --exclude flags: %s",
+					normalpath.Unnormalize(excludePath),
+				)
+			}
+		}
+	}
```
amckinney, line 67: Similar to the comment about O(n^2) vs two O(n) passes.

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -51,94 +54,156 @@ func (m *targetingModule) TargetFileInfos(ctx context.Context) (fileInfos []bufm
 			bufmoduleref.SortFileInfos(fileInfos)
 		}
 	}()
+	// We start by ensuring that no paths have been duplicated between target and exclude pathes.
+	for _, targetPath := range m.targetPaths {
+		for _, excludePath := range m.excludePaths {
+			if excludePath == targetPath {
+				return nil, fmt.Errorf(
+					"cannot set the same path for both --path and --exclude flags: %s",
```
amckinney, line 62:
```suggestion
					"cannot set the same path for both --path and --exclude-path flags: %s",
```

private/bufpkg/bufimage/util.go
doriable: We had a quick chat during stand-up about this, going to include better context for this in a comment, as mentioned above.

private/bufpkg/bufmodule/targeting_module.go
doriable: Yeah, the only difference here is because we have the case where `fileMatchingPathMap == 0` could be because `targetPaths == nil` (in the case where no `targetPaths` have been provided. Let me see if there is a reasonable way to clean this up and make this more generic.

amckinney (commented):
I made the edits to remove the recursive call from `excludeFile` (and renamed it to `shouldExcludeFile`). All of the tests are still passing, and I've reproduced the case you mentioned in the comment locally and things look OK.

If there is actually a problem with the implementation I proposed, we can follow-up with a test to guard against it. Let me know what you think.

doriable:
> If there is actually a problem with the implementation I proposed, we can follow-up with a test to guard against it. Let me know what you think.

We had a quick discussion about it yesterday, so I'm just going to document a quick summary here :) The implementation looks good since we are already processing the specific files, therefore we should always be able to match on the most specific exclude for a file (which would be excluding the file itself). I think this works great, I should've accounted for the pre-processing of specific target files ^^"

It also covers all the test cases (with workspaces, etc.) so I'm happy with this. I've also addressed the nits.

As an aside, another thing from an offline discussion yesterday, I tested the behaviours of this with `buf build` and `build.excludes` configurations for both a single module and a workspace. The excluded directories are treated as if they don't exist, and I believe this is the intended purpose of the `build.excludes` configuration.

amckinney (commented):
Some final nits worth considering, but none of them are major blockers.

I was about to approve, but I want to verify one last behavior I encountered in manual testing. In this case, I'm generating for a workspace and I've included a single path that contains a single import. When I specify the `--include-imports` flag, it appears to take precedence over the `--exclude-path` flag (the `repository_tag.proto` file is generated). 

I would expect the `--exclude-path` flag to take precedence so that users can say, "I want all my imports, except this one".

```
$ buf generate . --include-imports --exclude-path proto/buf/alpha/registry/v1alpha1/repository_tag.proto --path proto/buf/alpha/registry/v1alpha1/repository_commit.proto -o test --template data/template/buf.go.gen.yaml
```

private/buf/bufwire/module_config_reader.go
```diff
@@ -356,7 +378,39 @@ func (m *moduleConfigReader) getWorkspaceModuleConfigs(
 	// so we construct a separate workspace for each of the configured
 	// directories.
 	var moduleConfigs []ModuleConfig
+	// We need to first get the map of externalDirOrFilePath to subDirRelPath and the map
+	// of excludeDirOrFilePath to subDirRelExcludePath so we can check that all paths that
+	// have been provided at the top level have been accounted for across the workspace.
+	externalToRelPaths := map[string]string{}
+	excludeToRelPaths := map[string]string{}
```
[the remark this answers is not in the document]
doriable: Hmm, I think you're right, including the `external` will help with the understanding of the code, even though it's a bit longer. I think I'm going to go with `externalPathToRelPath` and `externalExcludePathToRelPath` (and other similar edits to the rest of the variables).

private/buf/bufwire/module_config_reader.go
```diff
@@ -356,7 +378,39 @@ func (m *moduleConfigReader) getWorkspaceModuleConfigs(
...
 	for _, directory := range workspaceConfig.Directories {
+		// We are unfortunately doing this work twice, once at the top level here, and when we
+		// build each workspace for the build options. We need to do the work at this level because
+		// we need to check across all workspaces once.
+		// We need to do the work again for each workspace build because a module can span across
+		// several workspaces.
```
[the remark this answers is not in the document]
doriable: Oh yeah, that's fair -- I'll add that to the note :D

private/buf/bufwire/module_config_reader.go
```diff
@@ -505,13 +577,38 @@ func (m *moduleConfigReader) getModuleConfig(
...
 			buildOptions, err = bufwork.BuildOptionsForWorkspaceDirectory(
 				ctx,
 				workspaceConfig,
 				moduleConfig,
-				relativeRootPath,
-				subDirPath,
 				externalDirOrFilePaths,
+				excludeDirOrFilePaths,
+				subDirRelPaths,
+				subDirRelExcludePaths,
```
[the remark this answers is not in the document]
doriable: So, this case is a bit strange: given that we are dealing with workspaces, we want to set an option with 0 `subDir*` paths if `len(externalDirOrFilePaths) > 0` but `len(subDirRelPaths) == 0` vs. `len(externalDirOrFilePaths) == 0`, where we would set no options -- this has to do with the fact that we could have no matching `subDirRelPath` for the given directory. So we would need to pass that information along, either the entire `externalDirOrFilePaths` to do a length check (as we do here), or some kind of flag or make some kind of implicit check based on `nil` for `subDirRelPath`... not sure which one is the best option here...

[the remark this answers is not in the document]
amckinney: Ok if we need it for control flow in the function, then it can remain as-is for now 👍

private/buf/bufwork/bufwork.go
```diff
@@ -211,18 +249,18 @@ func BuildOptionsForWorkspaceDirectory(
...
-		absExternalDirOrFilePaths = append(absExternalDirOrFilePaths, absExternalDirOrFilePath)
+		externalToAbsExternalDirOrFilePaths[externalDirOrFilePath] = absExternalDirOrFilePath
 	}
 	absRelativeRootPath, err := normalpath.NormalizeAndAbsolute(relativeRootPath)
 	if err != nil {
 		return nil, err
 	}
-	relativeRootRelPaths := make([]string, 0, len(absExternalDirOrFilePaths))
-	for i, absExternalDirOrFilePath := range absExternalDirOrFilePaths {
+	externalToRelativeRootRelPaths := map[string]string{}
```
amckinney, line 258: nit:
```suggestion
	externalToRelativeRootRelPaths := make(map[string]string)
```

private/buf/bufwork/bufwork.go
```diff
@@ -253,12 +291,12 @@ func BuildOptionsForWorkspaceDirectory(
 	//
 	// The 'proto' directory will receive the ./proto/buf/... files as ./buf/... whereas the
 	// 'enterprise/proto' directory will have no matching paths.
-	subDirRelPaths := make([]string, 0, len(relativeRootRelPaths))
-	for i, relativeRootRelPath := range relativeRootRelPaths {
+	externalToSubDirRelPaths := map[string]string{}
```
amckinney, line 294: nit:
```suggestion
	externalToSubDirRelPaths := make(map[string]string)
```

private/buf/bufwork/bufwork.go
```diff
@@ -202,7 +236,11 @@ func BuildOptionsForWorkspaceDirectory(
 	//
 	// Alternatively, we could special-case this logic so that we only work with relative paths when we have an ArchiveRef
 	// or GitRef, but this would violate the abstraction boundary for buffetch.
-	absExternalDirOrFilePaths := make([]string, 0, len(externalDirOrFilePaths))
+	externalToAbsExternalDirOrFilePaths := map[string]string{}
```
amckinney, line 239: nit:
```suggestion
	externalToAbsExternalDirOrFilePaths := make(map[string]string)
```

private/bufpkg/bufmodule/targeting_module.go
```diff
@@ -51,94 +54,152 @@ func (m *targetingModule) TargetFileInfos(ctx context.Context) (fileInfos []bufm
...
+								// the walk early if a path is found.
+							}
+							return nil
+						},
+					); walkErr != nil {
+						return nil, walkErr
+					}
+					if !foundPath {
+						return nil, fmt.Errorf("path %q has no matching file in the image", excludePath)
+					}
```
amckinney, line 141: With respect to the `TODO`, ideally yes. You can just return the error that you intend to return and the `Walk` will stop.
```suggestion
					if walkErr := sourceReadBucket.Walk(
						ctx,
						"",
						func(objectInfo storage.ObjectInfo) error {
							if normalpath.EqualsOrContainsPath(excludePath, objectInfo.Path(), normalpath.Relative) {
								return fmt.Errorf("path %q has no matching file in the image", excludePath)
							}
							return nil
						},
					); walkErr != nil {
						return nil, walkErr
					}
```

amckinney (approved):
Looks great! I have a few nits (the most important of which is in `targeting_module.go`), but I trust that you'll address them before merging.

private/bufpkg/bufmodule/targeting_module.go
doriable: Based on offline discussions, we are using a sentinel error for an early return on the walk.

PR #694, opened 2021-10-31
Use method from http package

alexander-melentyev:


bufdev:
We'll consider changing this during a cleanup pass - thanks!

PR #698, opened 2021-11-03
add methods for searching users, templates or plugins

smallsamantha:
adding methods to be implemented for [2857](https://github.com/bufbuild/core/issues/2857)


doriable (approved):

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -87,3 +87,45 @@ message SearchResponse {
   // There are no more pages if this is 0.
   uint32 next_page_token = 2;
 }
+
+message SearchUsersRequest {
```
bufdev, line 91: I would add the RPCs at the same time

smallsamantha: can you look at discussion [here](https://github.com/bufbuild/core/issues/2857) to help us decide if this is the direction we want to go?

bufdev: Commented - I would decide on that direction first though before adding request/response messages here

PR #703, opened 2021-11-05
Support multiple insertion points

### LINKED ISSUE #702: support multiple insertion points per file
buf 1.0.0-rc6 is not handling correctly a plugin that emits more than one insertion point per output file. The protoc compiler is able to handle this case, so this represents a deviation. For example, if we have two plugins `a` and `b`, and the first plugin emits `a.txt` like this:

```
X:
@@protoc_insertion_point(X)
Y:
@@protoc_insertion_point(Y)
```

and a second plugin `b`, attempts to insert at both `X` and `Y`, by emitting the following response:

```
  file: <
    name: "a.txt"
    insertion_point: "X"
    content: "inserted at X\n"
  >
  file: <
    name: "a.txt"
    insertion_point: "Y"
    content: "inserted at Y\n"
  >
```

The expectation is that the the output file `a.txt` will look like this:
```
X:
inserted at X
@@protoc_insertion_point(X)
Y:
inserted at Y
@@protoc_insertion_point(Y)
```

However, buf prints the following warning: 
```
Warning: Duplicate generated file name "a.txt". Buf will continue without error here and drop the second occurrence of this file, but please raise an issue with the maintainer of the plugin.
```

it performs the insertion at `X`, and skips the insertion at `Y`.

I have created a minimal reproducible example in [this repo](https://github.com/thesamet/buf-issue) with trivial plugins that produce the responses above (regardless of the input).

Context for the issue: I am the author and maintainer of ScalaPB. This issue was reported by a user who is using buf to build their scalapb/scalapb-validate project. The scalapb-validate plugin inserts at two locations for files that are created by the ScalaPB plugin.

amckinney:
Fixes #702 

We were guarding against duplicate files, but did not take into account whether or not the file contained an insertion point. The insertion point test was adapted to match the test case demonstrated in #702.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Upgrade to `protoc` 3.19.1 support.
+- Fix issue with `buf generate` where multiple insertion points are defined in the same file.
 
 ## [v1.0.0-rc6] - 2021-10-20
 
--- a/private/buf/cmd/buf/command/generate/testdata/insertion_point/test.txt
+++ b/private/buf/cmd/buf/command/generate/testdata/insertion_point/test.txt
@@ -11,5 +11,13 @@
 				
 		//     @@protoc_insertion_point(example)
 		//
-		// Note that all text should be added above the insertion point.
+		// The 'other' insertion point is also included so that we verify
+		// multiple insertion points can be written in a single invocation.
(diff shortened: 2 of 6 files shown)


bufdev (approved):

bufdev:
Update CHANGELOG.md too

PR #704, opened 2021-11-05
add new endpoints for configurable recommendation

cyinma:
This is to add new endpoints in recommendation service so that it can be configurable.


proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -18,12 +18,42 @@ package buf.alpha.registry.v1alpha1;
...
+    // The resource is a `template`.
+    string template_id = 2;
+  }
+}
+
+message Recommendation {
+  // unique, immutable
+  string id = 1;
+  // unique, immutable
+  Resource resource = 2;
```
johanbrandhorst, line 41: What do you think about inlining this message instead of making a separate message for it?
```suggestion
  oneof resource {
    // The resource is a `repository`.
    string repository_id = 2;
    // The resource is a `template`.
    string template_id = 3;
  }
```

proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -40,6 +70,20 @@ service RecommendationService {
   rpc RecommendedRepositories(RecommendedRepositoriesRequest) returns (RecommendedRepositoriesResponse);
   // RecommendedTemplates returns a list of recommended templates.
   rpc RecommendedTemplates(RecommendedTemplatesRequest) returns (RecommendedTemplatesResponse);
+  // ListRecommendedRepositories returns a list of recommended repositories that user have access to.
+  rpc ListRecommendedRepositories(ListRecommendedRepositoriesRequest) returns (ListRecommendedRepositoriesResponse);
+  // ListRecommendedTemplates returns a list of recommended templates that user have access to.
+  rpc ListRecommendedTemplates(ListRecommendedTemplatesRequest) returns (ListRecommendedTemplatesResponse);
+  // ListRecommendations returns a list of recommendations of a resource type.
+  rpc ListRecommendations(ListRecommendationsRequest) returns (ListRecommendationsResponse);
```
johanbrandhorst, line 78: Do we need this if we have the other two?

proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -53,3 +97,74 @@ message RecommendedTemplatesRequest {}
...
+message UpdateRecommendationsRanksRequest {
+  ResourceType resource_type = 1;
+  repeated string recommendation_ids = 2;
+}
+
+message UpdateRecommendationsRanksResponse {}
+
+message DeleteRecommendationRequest {
+  ResourceType resource_type = 1;
+  string id = 2;
```
johanbrandhorst, line 167: This shouldn't need the type, it should operate on the `id` alone, right?

proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -40,6 +70,20 @@ service RecommendationService {
...
+  // ListRecommendations returns a list of recommendations of a resource type.
+  rpc ListRecommendations(ListRecommendationsRequest) returns (ListRecommendationsResponse);
+  // AddRecommendation add a recommendation to the server.
+  rpc AddRecommendation(AddRecommendationRequest) returns (AddRecommendationResponse);
+  // UpdateRecommendation update a recommendation in the server.
+  rpc UpdateRecommendation(UpdateRecommendationRequest) returns (UpdateRecommendationResponse);
+  // UpdateRecommendationsRanks update the ranking of the recommendations of a resource type in the server.
+  rpc UpdateRecommendationsRanks(UpdateRecommendationsRanksRequest) returns (UpdateRecommendationsRanksResponse);
+  // DeleteRecommendation delete a recommendation in the server.
+  rpc DeleteRecommendation(DeleteRecommendationRequest) returns (DeleteRecommendationResponse);
```
johanbrandhorst, line 86: If we treated recommendations as immutable, we could reduce this to just two endpoints; `AddRecommendations` and `DeleteRecommendations`. If we make it `AddRecommendedRepositories` and `DeleteRecommendedRepositories` and corresponding methods for Templates, we can remove the `Resource` type and enum. I appreciate the value of reducing the number of RPCs, but I really kinda hate the enum-to-determine-type :P. I'm obviously open to change my mind, I think we discussed the explosion of RPCs before.

proto/buf/alpha/registry/v1alpha1/recommendation.proto
cyinma: the other two are for normal user to get the recommendation list, this `ListRecommendations` is for server admin to list and manage the recommendations, so it might be better to have both of them, or we will need to expose more information to normal user so that server admin can use those information to manage the recommendations.

johanbrandhorst: What information do we need to expose to the server_admin that we don't want to expose to the user? I'm confused, can we not use both the List methods for the server admin too?

proto/buf/alpha/registry/v1alpha1/recommendation.proto
cyinma: we can, but we will need to try to delete the record from both recommendation tables as we don't know where this actually lives. I think it might be better to let frontend include this information?

johanbrandhorst: This is an unfortunate consequence of an implementation detail (the two different tables). I don't think we should be exposing that to the API, we should make two deletes if we have to. What do you think?

proto/buf/alpha/registry/v1alpha1/recommendation.proto
cyinma: the `recommendation_id`, I don't think we don't want to expose this to the user, but this is not needed. But I guess it's fine to just include this in the other two endpoints

proto/buf/alpha/registry/v1alpha1/recommendation.proto
cyinma: sounds good for me!

proto/buf/alpha/registry/v1alpha1/recommendation.proto
cyinma: I guess we still want to keep the update rank endpoint, otherwise user will need to delete all recommendations and add them again if they want to add some recommendation in the middle of the list? This also means two/three extra endpoints needed per new recommendation type in the future, though it might be too early to think about it.

johanbrandhorst: Yes, but the user doesn't have to be exposed to the method calls that perform the deletion/addition. We could actually use a `SetRecommendedRepositories` instead that could just overwrite the contents every time, no need to delete. Then the frontend could have a nice drag and drop table with a `Save` button or something to update them all at once. Would be nice to get the input from @timostamm who will most likely implement it!

cyinma: the drag and drop table is actually what I was thinking so that I come up with `UpdateRecommendationsRanks`, but put the add and delete as separated actions. Do you mean no `add`, no `delete`, just `set` for each type of resources?

johanbrandhorst: Perhaps, I thought we might still have a `Delete` but maybe you can just `Set` it to an empty list!

cyinma: sounds good for me, @timostamm what do you think?

timostamm: > Yes, but the user doesn't have to be exposed to the method calls that perform the deletion/addition. We could actually use a `SetRecommendedRepositories` instead that could just overwrite the contents every time, no need to delete. 

I agree that this should be sufficient. Handling in the frontend would be very straight-forward. I expect payload sizes to be so small that it would not make a noticeable difference even with many dozen entries.

cyinma: thats cool, I guess we don't need any `enum`/`oneof` at all too in this case

cyinma: also @timostamm if payload sizes is not a concern for `set`, do we still need to do paging for `ListRecommendedRepositories` and `ListRecommendedTemplates`? my only concern is I don't know how many recommendation items user will set.

timostamm: I don't think it makes sense to recommend more than 10 repositories to the user (without some kind of categorization), and even less for templates (per registry). Where we set the limit is something we should discuss, but from my point of view, pagination is not necessary.

cyinma: cool, then I will also remove the pagination, but will keep the new `ListRecommendedRepositories` and `ListRecommendedTemplates` endpoints so that we can gracefully migrate to the new configurable recommendation.

proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -32,6 +34,19 @@ message RecommendedTemplate {
   string owner = 1;
   string name = 2;
   string description = 3;
+  string template_id = 4;
+}
+
+// RecommendingRepository is the information needed to configure a repository recommendation
+message RecommendingRepository {
```
cyinma, line 41: this name could be better I guess, but I cannot come up with one

johanbrandhorst: Haha I'm trying to think of one but I think you're right I don't know 😂.

johanbrandhorst (approved):
LGTM

proto/buf/alpha/registry/v1alpha1/recommendation.proto
timostamm: It would technically be possible to use the message `RecommendedRepository` here. I'm not sure that the two added messages are much cleaner than documenting that `SetRecommendedRepositories` only requires `repository_id` and `description` to be set. Have we discussed such a case before?

timostamm: > I think we generally use the resources only in return messages, and take whatever the required input to creation/deletion is explicitly.

That seems worth sticking to.

`SetRecommendedRepository` seems like a better fit to me than `RecommendingRepository` - what is your take, @cyinma?

cyinma: it seems good, will update it

timostamm (approved):
👍

PR #705, opened 2021-11-05
Change errors.Is behavior of rpc.Error to allow sentinel error checking

robbertvanginkel:
Our `errors.Is` implementation currently behaves like a type assertion, which is somewhat unexpected compared to the recommended use:

> In the simplest case, the errors.Is function behaves like a comparison to a sentinel error, and the errors.As function behaves like a type assertion. When operating on wrapped errors, however, these functions consider all the errors in a chain.
https://go.dev/blog/go1.13-errors#examining-errors-with-is-and-as

The current implementation makes it so that various combinations of rpc errors match on `errors.Is`:
```
errors.Is(NewCanceledError(""), NewCanceledError("abc")) // true, I'd expect false
errors.Is(NewCanceledError(""), NewAbortedError(""))     // true, I'd expect false
errors.Is(NewCanceledError(""), NewCanceledError(""))    // true, I'd expect false
```

This proposes a change that allows using `Is` again for checking sentinel errors through wrapped error chains. To check if an error is created from this package, code can continue to use `rpc.IsError` which will match all errors, and individual error details can still be extracted with `rpc.GetError{Code,Message}`.


robbertvanginkel:
Putting this up as a draft for discussion, there's some more errors that implement `Is` by just doing a type assertion: https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/bufbuild/buf%24+Is%28err+error%29&patternType=literal. We should discuss and see if we need to do more to stay consistent.

bufdev:
I know there's been a lot of discussion about this, and I won't block, but not only I would expect `errors.Is(NewCanceledError(""), NewCanceledError(""))` to be true, but I'd expect `errors.Is(NewCanceledError(""), NewCanceledError("abc"))` to be true. The `message` component of an error should not have meaning in terms of what the error is, it should be treated as informational.

`errors.Is(NewCanceledError(""), NewAbortedError(""))` is up for debate, but I could see the argument for this being false.

robbertvanginkel:
The `errors.New` behavior from the stdlib also does not match `errors.Is` for similar messages, only if we're comparing the same error:
```
fmt.Println(errors.Is(errors.New("abc"), errors.New("abc"))) // false

x := errors.New("abc")
fmt.Println(errors.Is(x, x)) // true
```

As I understand `Is`, it should be about comparisons for errors across wrapping, which before the existence of wrapping would have been done with `==`. For comparable checks, I don't think we could ignore message or error codes.

amckinney (approved):
Discussed offline - 

This looks good to me, especially because all of our bases are covered with the exported `Is*` functions (e.g. `IsExistsMultipleLocations`) by using `errors.As` under the hood. This also makes things more consistent with the how `errors.Is` is intended to be used (which you've already described well in your summary + comments).

We can revisit this again later if we still feel strongly about our custom `errors.Is` implementation - all that matters for now is compatibility with the previous implementation.

PR #706, opened 2021-11-05
Enable trunk in buf repo

EliSchleifer:
Adds trunk to the bufbuild/buf repo. Enables 9 linters/analyzers/formatters and includes some reasonable default configurations for some of those tools. 

1) Enables trunk in the repo for local runs (recommend running trunk in CI as well once the tool feels properly tuned).
2) trunk is added to the .envrc allowing anyone in repo to run `trunk` without additional installation steps.
3) Enables the following linters:
     actionlint, buf, gitleaks, gofmt, golangci-lint, markdownlint, prettier, shellcheck, shfmt
4) Some changes were made in trunk.yaml to how buf is run. These have landed in trunk mainline but won't release for another week or so. In the meantime the provided trunk.yaml file provides the correct syntax.

To see all existing issues run:
```
trunk check --all -n
```

To apply all formatting changes run:
```
trunk fmt --all
```

Analysis:
shellcheck and markdown lint issues detected in 7 files

Formatting Inconsistencies:
18 files require prettier format
5 files require shfmt


.envrc
```diff
@@ -1 +1,2 @@
 . $(make -s direnv)
+PATH_add .trunk
```
bufdev, line 2: All of `.envrc` is controlled via `make direnv`

.markdownlint.yaml
```diff
@@ -0,0 +1,4 @@
+# Prettier-friendly minimal markdownlint config (turns off rules which prettier handles)
+default: true
```
bufdev, line 2: I don't think we want markdown linting in this repository - the only things that are markdown are the readme and the changelog

.shellcheckrc
```diff
@@ -0,0 +1,3 @@
+# Following source doesn't work in most setups
+disable=SC1090
```
bufdev, line 2: We don't want to enable shellcheck for now

.trunk/.gitignore
```diff
@@ -0,0 +1 @@
+*out
```
bufdev, line 1: We don't want a new top-level directory i.e. `trunk`.

All of our `.gitignores` are controlled via the `make` system.

.trunk/trunk
```diff
@@ -0,0 +1,186 @@
+#!/bin/bash
```
bufdev, line 1: This script doesn't conform to our bash style as a whole, see other bash scripts for equivalents.

All bash scripts live in `make/buf/scripts` and are controlled by the `make` system.

All installation is controlled by the `make` system.

.trunk/trunk
```diff
@@ -0,0 +1,186 @@
...
+    echo -e "\r${x_mark} ${verifying_text} FAILED"
+    echo "Expected sha256: '${sha256}'"
+    echo "Actual sha256:   '${actual}'"
+    exit 1
+  fi
+  echo -e "\r${check_mark} ${verifying_text} done"
+}
+
+add_to_bin() {
+  bin_script="/usr/local/bin/trunk"
```
bufdev, line 61: Nothing in this repository should be installing to `/usr/local/bin` - everything is installed to a cache controlled by the `make` system.

.trunk/trunk
```diff
@@ -0,0 +1,186 @@
...
+  bash_source=${BASH_SOURCE[0]}
+
+  if [[ -f ${bin_script} ]] && [[ ${bin_script} != "${bash_source}" ]]; then
+    read -rp "Trunk Launcher is already installed in /usr/local/bin. Would you like to override it? (Y/n) " yn
```
bufdev, line 65: Having a "launcher" and then a separate binary (?) is worrisome.

This script is interactive, which isn't appropriate for us.

.trunk/trunk.yaml
```diff
@@ -0,0 +1,43 @@
+version: 0.1
+cli:
+  version: 0.1.31-alpha
+  sha256:
+    darwin_x86_64: 15f297b0657b615609adfd1a30b99a6210a6ff7c52dc8bdfeb0053a8aff7a9bc
+    linux_x86_64: 5ed59e39cd2f0f72008834073cec429689c88e480abbe89cc36340b75c648a9f
+lint:
+  enabled:
+    - actionlint@1.6.4
```
bufdev, line 9: All our linters are installed and cached on a per-version basis by our `make` system. We don't want to break this just for trunk.

.trunk/trunk.yaml
```diff
@@ -0,0 +1,43 @@
+version: 0.1
+cli:
+  version: 0.1.31-alpha
+  sha256:
+    darwin_x86_64: 15f297b0657b615609adfd1a30b99a6210a6ff7c52dc8bdfeb0053a8aff7a9bc
```
bufdev, line 5: Most of our developers use M1 Macbooks at this point.

.trunk/trunk.yaml
```diff
@@ -0,0 +1,43 @@
...
+  linters:
+    - name: buf-breaking
+      run_linter_from: root_file
+      run_from_root_file: buf.work.yaml
+      command:
+        [
+          buf,
+          breaking,
+          --against,
+          ".git#branch=${upstream-ref}",
```
bufdev, line 28: The `against` input is different locally than in GitHub actions, and we control this via the `make` system.

bufdev (changes requested):
I added some specific comments. Overall, I don't think we can merge this, two main reasons:

- This is outside of our `make` system that controls all dependency installation on a per-version basis, and all development commands.
- The launcher/binary paradigm is concerning, especially that something is installing to `/usr/local/bin`.

EliSchleifer:
> I added some specific comments. Overall, I don't think we can merge this, two main reasons:
> 
> * This is outside of our `make` system that controls all dependency installation on a per-version basis, and all development commands.
> * The launcher/binary paradigm is concerning, especially that something is installing to `/usr/local/bin`.

Thanks for the thoughtful notes. 
1) trunk is geared towards organizations that don't have the resources or desire to maintain their toolchains through a make or equivalent system. trunk is designed for organizations seeking to get out of the business of maintaining these tools themselves and adds support to allow for incremental rules to be added without engineers having to fix all broken rules at once. You could wrap trunk check in a straigtforward make rule, but feels like you're satisfied with your existing toolset.
2) The launcher/binary allows engineers to start using the tool without fuss. Anyone can hit be running the correct hermetic version of trunk pinned in your repo in seconds, without worrying about which pacakge manager to use to get it, etc... There is no need to install anything to /usr/local/bin (we do offer that as a command in the launcher for users who want to install trunk globally). We will be adding support for brew style installation in the future for those folks who prefer that approach.

.markdownlint.yaml
EliSchleifer: A core value proposition here is to make sure that all your tech is checked, but that's not for everyone. If you don't want to be running tools on all the tech, than no need for a meta-linter :)

.shellcheckrc
EliSchleifer: I would note that you could enable trunk with shellcheck today and continue to land code without tripping any errors - that's the value of our hold-the-line feature. Basically your tech debt is your tech debt and shouldn't get in the path of forward movement. It is a nice tool to have and to understand where those potential bugs/issues are hiding.

bufdev:
Almost all of our tools have a binary we can just `curl` and install - `buf` provides this for example, you just `curl -sSL https://github.com/bufbuild/buf/releases/download/$VERSION/buf-$(uname -s)-$(uname -m)` - I'd expect this to be all that is required for a linting tool.

.trunk/trunk
EliSchleifer: This is an optional command in the launcher script for those people who call it. It is in no way triggered by default or required. Feedback is great - we should note this better in the script itself.

bufdev: I think the fundamental problem is the need for a script to install a linting tool, re: my comment below about not being able to `curl` a `trunk` binary.

.markdownlint.yaml
bufdev: I think it's more that we want to make sure we know what linters we are running, and to understand them. I don't know what markdownlint is or what it does, and don't know if we want to enforce whatever rules they set.

.trunk/trunk.yaml
EliSchleifer: The intention of trunk is to offload responsibility for managing/maintaining these tools. Feels like the wrong fit here.

.shellcheckrc
bufdev: We've gone through shellcheck and don't want to enable all of its features, it's not a matter of not getting existing code in line (if it was, we'd just update the scripts to match what shellcheck expects)

.trunk/trunk.yaml
EliSchleifer: As do we. Love the M1. Most analyzers haven't been recompiled for arm mac, but we plan on compiling a native arm build for trunk in the future.

EliSchleifer:
Love the feedback. Closing this PR for now.

PR #711, opened 2021-11-08
Add `oneof` fields to documentation API

doriable:
Adds the `oneof` fields to the `Message` structure.


proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -260,7 +260,7 @@ message Message {
   // file_path is the normalized path of the file containing the message.
   string file_path = 5;
   bool is_map_entry = 6;
-  repeated Field fields = 7;
+  repeated MessageField fields = 7;
```
amckinney, line 263: Having reviewed the [specification](https://developers.google.com/protocol-buffers/docs/overview#updating) and [encoding](https://developers.google.com/protocol-buffers/docs/encoding#embedded), I think embedding the `oneof` in a nested message actually make this wire incompatible. We're introducing new field numbers for the `MessageField` (those referenced in its `oneof` field), and those will be encoded in the wire format.

From the specification,
> Changing a single optional value into a member of a new oneof is safe and binary compatible. Moving multiple optional fields into a new oneof may be safe if you are sure that no code sets more than one at a time. Moving any fields into an existing oneof is not safe.

Unfortunately, we can't flatten out `repeated` values in a `oneof` at this level since it's not supported, i.e. 
```protobuf
// This isn't valid.
oneof message_field {
  repeated Field fields = 7;
  repeated Oneof oneofs = 10;
}
```

Anyway, this might have already been thought out and I might be missing some context, but I just wanted to call out that this will break compatibility between the BSR and the documentation front-end if they aren't deployed alongside one another.

bufdev: I think we expect breaking changes for v1alpha1, but yes this is not wire compatible.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -303,6 +303,20 @@ message MapEntry {
...
+message OneOf {
+  string name = 1;
+  // the fields that are part of the oneof
+  repeated Field fields = 2;
+}
+
+message MessageField {
+  oneof message_field {
+    Field field = 1;
+    OneOf oneof = 2;
```
bufdev, line 316: Why is this embedded within a MessageField, instead of being directly on the Message? Is this to try to maintain ordering of fields for documentation purposes?

doriable: Yeah, I had this conversation with @SyrieBianco yesterday, and this was the thought behind this. But I believe that is not necessarily a hard requirement for the documentation and we can always handle the sorting on the frontend, so I think I'll be reverting this.

proto/buf/alpha/registry/v1alpha1/doc.proto
doriable: Yeah, that makes sense -- originally I had encoded the `oneofs` as a separate field in the `Message` struct, but as per the comment below, we wanted to try and retain the field ordering for documentation, but this could be handled on the frontend (especially since this isn't wire compatible 😅 ).

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -262,6 +262,7 @@ message Message {
   string file_path = 5;
   bool is_map_entry = 6;
   repeated Field fields = 7;
+  repeated OneOf oneofs = 10;
```
amckinney, line 265: This should be moved after the 9th field number (so that they're in order).

amckinney (approved):

proto/buf/alpha/registry/v1alpha1/doc.proto
doriable: I guess I wanted this logically to be grouped next to `fields`, but also didn't want to swap the field numbers because I didn't want to make an unnecessary breaking change. Thoughts on what the best way to show this is? (Also, thanks for the stamp, haha, I'm just wondering if we have any thoughts on the style for this)

amckinney: Not a huge deal either way - I prefer field numbers to be in order since that's what we do elsewhere, but that's just ne.

doriable: Sounds good, not too worried about it, was just wondering because I've seen different styles for doing this (e.g. descriptor.proto has some random ordering things https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L66-L72).

PR #713, opened 2021-11-09
Enable `exclude-path` on `breaking`, `build`, `export`, and `lint`

doriable:
#689 Added support for `exclude-path` to `buf generate` by enabling our module and image build paths to support exclude paths. This turns that functionality on for the rest of the commands that also supported `--path`.

Reference: #224


amckinney (approved):
Nice 👏

Issue #682, opened 2021-10-25
Encode buf.yaml in Module


amckinney:
We currently encode `dependencies` (i.e. the `buf.lock`) and `documentation` (i.e. the `buf.md`) in the `Module`, but we do not encode anything found in the `buf.yaml`. 

With the `buf.yaml`, the BSR could start to enforce `lint` and `breaking` configurations on the server-side, and we could more easily include license information which we currently plan to include in the `buf.yaml` (re: https://github.com/bufbuild/buf/issues/488).

We'll need to start by modeling the contents of the `buf.yaml` in `.proto`, which will roughly look something like the following:

```diff
diff --git a/proto/buf/alpha/module/v1alpha1/module.proto b/proto/buf/alpha/module/v1alpha1/module.proto
index 5b79ac80..8e4a772a 100644
--- a/proto/buf/alpha/module/v1alpha1/module.proto
+++ b/proto/buf/alpha/module/v1alpha1/module.proto
@@ -34,6 +34,14 @@ message Module {
   //
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
   string documentation = 3;
+  // license is the string representation of the license.
+  string license = 4;
+  // module_name is the name.
+  ModuleName module_name = 5;
+  // lint_config is the lint configuration that can (optionally) be enforced by the BSR.
+  LintConfig lint_config = 6;
+  // breaking_config is the breaking configuration that can (optionally) be enforced by the BSR.
+  BreakingConfig breaking_config = 7;
 }
```

Note that we _don't_ need to include the [build](https://docs.buf.build/configuration/v1/buf-yaml#build) configuration found in the `buf.yaml` since this might be deprecated in favor of `.bufignore` (re: https://github.com/bufbuild/buf/issues/578).

doriable:
Some things we discussed today regarding the encoding the module configuration for use on the server side:

- While it is a "nice side effect" to be able to "reconstruct" the module configuration for the user through the UI, this should be considered a non-goal for this. The primary requirement is to be able to use module configuration for server-side work, such as gating pushes with lint and breaking change detection, etc. We can eventually surface the module configuration to the user, but it won't be the exact representation of their `buf.yaml` (e.g. comments in the `buf.yaml`, whitespace, etc. are not included) since we are not working with the raw representation of the `buf.yaml`
[...]

PR #715, opened 2021-11-10
WIP: Add ModuleDigestB3

amckinney:
As we approach `v1`, we are revisiting some of the decisions we made with respect to the module digest. Currently, we encode the `remote`, `owner`, and `repository` name in the digest, but this is a prone to a couple issues:

1. If a user updates their organization or repository name, they will have a new module digest which invalidates their local module cache entry.
2. If the user pushes a commit with the same content immediately after the repository has been renamed, the user will have two commits with the same content adjacent to each other.
3. A module with the exact same content (and dependencies), but hosted on two different repositories, will have two different digests, which prevents us from leveraging CAS properties that a digest would normally be able to give us in these situations.

Instead, we should only encode information relevant to the module's content, including the following:

1. The module file paths and their content.
2. The commit IDs of the module's dependencies.
3. Module metadata, including its documentation and license.

This change starts by adapting the digest used by the module cacher. The digest function used by the BSR to validate adjacent commits will be updated and deployed in a follow-up - please see the `CHANGELOG.md` for details on the expected outcome for users on different versions of the `buf` CLI.

> Note that we can safely remove the `ModuleDigestB2` function since it was only used by the module cacher in `bufbuild/buf`. The `modulePinCompareToOnlyCommit` function is refactored since it inappropriately accounted for the remote, owner, and repository.


amckinney:
As we move towards https://github.com/bufbuild/buf/issues/682, we'll need to start including the `lint` and `breaking` configurations in the digest. When a user updates either of these configurations, it should create a new commit, but this has a couple of caveats to think through.

I'll describe everything in detail here so that we can reflect on the caveats associated with these types of changes later on.

### Old CLIs

When we deploy this on the BSR, we'll adjust the digest function validation to use the `b3` format. However, any old CLI that doesn't account for the `lint` and `breaking` configurations will _not_ include these fields in their serialized module. This means that two different users with different CLI versions contributing to the same BSR repository would be able to push the same module files and successfully receive a new commit ID (one module will include the `lint` and `breaking` configuration, whereas the other won't).

This isn't ideal since this could invalidate the `lint` or `breaking` enforcement that the repository owner hopes to enable on their repository. For example, suppose that I configure my repository so that it prevents a `buf push` if it breaks the module's `breaking` configuration. If a colleague of mine with write access uses an old CLI, it will omit the `lint` and `breaking` configuration entirely, so they could side-step the validation and overwrite the `main` branch with an incompatible commit. This is something we should actively prevent because the user might think they are preserving the `lint` and `breaking` configuration found in their `buf.yaml`, but their CLI silently omits it.

There are ways around this caveat though. If a user enables this server-side enforcement (which is probably a property on the repository itself, just like deprecation, TBD), the server could require that a non-`nil` `lint` or `breaking` configuration is present in the module. This means that **anyone contributing to the repository will need to upgrade their CLI to support server-side enforcement**, which should be a reasonable ask. In short, enforcing `lint` and `breaking` requires that all consumers use a CLI that is aware of this configuration. This also means that anyone that enables the server-side enforcement feature will not be allowed to push a module that omits `lint` and `breaking` entirely like the one shown below:

```yaml
version: v1
name: buf.build/acme/weather
```

We actually faced a similar issue when we included module documentation in the digest. Anyone using an old CLI would have omitted the documentation, but we didn't face any issues with this in practice. We're going to have a similar issue when we introduce licenses, too. However, the `lint` and `breaking` configuration is slightly different due to the server-side enforcement caveat described above.

### Adjacent Commits

Regardless of the `lint` and `breaking` encoding, whenever we change the digest function on the server-side (i.e. `b1-` -> `b3-`), users will be able to push the same module content and receive a new commit ID because the parent commit's digest will differ from the current commit's digest.

One idea is that we could support multiple digests at the same time and check the parent's digest to determine which algorithm to use, but this would prevent us from switching an old repository's digest algorithm to the new one. This isn't a viable solution.

Instead, **we should just accept that adjacent commits can exist in the midst of a digest update**, but the commits will be treated differently due to their encoded digest. Plus, this is enough of an edge case that it should only affect a fraction of our users with really no effect.

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -117,6 +122,46 @@ func NewConfigV1ForProto(protoConfig *breakingv1.Config) (*Config, error) {
...
+func NewBreakingConfigToBytes(config *Config) []byte {
+	if config == nil {
+		return nil
+	}
+	hash := sha256.New()
+	sort.Slice(config.Rules, func(i, j int) bool { return config.Rules[i].ID() < config.Rules[j].ID() })
+	for _, rule := range config.Rules {
+		str := append([]string{}, rule.ID(), rule.Purpose())
+		str = append(str, rule.Categories()...)
+		hash.Write([]byte(strings.Join(str, "")))
```
doriable, line 135:
```
[]bytes, err := json.Marshal(rule)
if err != nil {
    return nil, nil
}
```

doriable: After sorting.

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -117,6 +122,46 @@ func NewConfigV1ForProto(protoConfig *breakingv1.Config) (*Config, error) {
...
+		var rootPaths []string
+		rootPathsMap := config.IgnoreIDToRootPaths[ignoreID]
+		for rootPath := range rootPathsMap {
+			rootPaths = append(rootPaths, rootPath)
+		}
+		sort.Strings(rootPaths)
+		str := append([]string{}, ignoreID)
+		str = append(str, rootPaths...)
+		hash.Write([]byte(strings.Join(str, "")))
+	}
```
doriable, line 152: Use combined interim structure with sorted values -> json.Marshal.

doriable:
Superseded by: #748

amckinney:
Closing in favor of #748.

PR #716, opened 2021-11-10
Embed `oneofs` into a `MesssageField` structure for documentation ordering

doriable:
After some discussion, in order to retain the field ordering in the frontend for documentation, we should embed `Oneof` with `Field` into a `MessageField` message.


SyrieBianco:
For context, this addresses the ordering issues discussed in [bufbuild/core#2934](https://github.com/bufbuild/core/issues/2934) and the ordering concerns are sufficient enough to override Alex's [wire incompatibility concerns](https://github.com/bufbuild/buf/pull/711/files/e3ac49bc31c3ecfcbdbd6dc3627c9a985c205374#diff-731555e5741291a4f8f2c612decd7a931052a94f1d71d7f09554c2d71b454e10) in the previous version of the API.

SyrieBianco (approved):

PR #717, opened 2021-11-11
new endpoints for update organization base roles

cyinma:
This is to add new endpoints for update organization base roles.
Proposed to update all base roles together in the same endpoint, but still depends on what frontend UX do we want.


proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -66,6 +72,10 @@ service OrganizationService {
   rpc UpdateOrganizationMember(UpdateOrganizationMemberRequest) returns (UpdateOrganizationMemberResponse);
   // RemoveOrganizationMember remove the role of an user in the organization.
   rpc RemoveOrganizationMember(RemoveOrganizationMemberRequest) returns (RemoveOrganizationMemberResponse);
+  // GetOrganizationSettings gets the setting of a organization, including organization base roles.
+  rpc GetOrganizationSettings(GetOrganizationSettingsRequest) returns (GetOrganizationSettingsResponse);
+  // UpdateOrganizationBaseRoles update the organization base repository role, plugin role, and template role.
+  rpc UpdateOrganizationBaseRoles(UpdateOrganizationBaseRolesRequest) returns (UpdateOrganizationBaseRolesResponse);
```
mfridman, line 78: This is part of a larger conversation, should we optimize the API to reduce bloat?

Couldn't we send the desired organization settings instead of building out endpoints for each individual attribute. E.g.,

```
UpdateOrganization
UpdateOrganizationSettings

vs.

UpdateOrganizationBaseRoles
UpdateOrganizationName
UpdateOrganizationLogo
UpdateOrganizationSSOConfig
UpdateOrganizationBanner
UpdateOrganizationDescription
.... etc.
```

It seems like we can have fewer endpoints that do more, instead of endpoint per attribute. I'm pretty sure @robbertvanginkel had this same concern.

johanbrandhorst: One concern is how this works for partial updates - what happens when you just want to update the repository base role?

mfridman: That's an excellent point, re. updates. Potential solutions could be:

1. Send the entire payload. Most of the time, esp. on the frontend, you'll already have the organization context, so update the required fields and send the entire state. Similar to how you'd send k8s the desired state.
2. Do a bit more work to differentiate between null vs empty values. E.g., using `google.protobuf.StringValue` .. I don't think we utilize this pattern, but I've seen this work okay-ish. Build up an `UPDATE` query from the non-null fields.

These come to mind, maybe there are others?

johanbrandhorst: Another option is to use a FieldMask to define a field to update. Unfortunately I think all these suggestions are worse than just having multiple RPCs, IMO.

cyinma: For sending the entire payload, my concern is if there are multiple organization admin editing different parts of setting in the same time, which is very likely to be happened, so with in these 2 suggestions I will prefer to do something like (2) to only update non-null struct in the settings, however I don't think it can fully solve the scenario of updating only the repository base role, we still need to pass all 3 base roles anyway

mfridman: Partial updates aren't awful, they do require a bit more logic in the business layer but the tradeoff is a cleaner and potentially easier API to work with, IMO.

Probably beyond the scope of this PR, but something that worries me is the API surface we'll end up with as we continue to add functionality.

cyinma: I guess it still depends on how small we divide them to multiple RPCs. For me, grouping the update of 3 base roles together is in the balance that acceptable in both API bloat concern and RPC boundary concern. However, if we continue to add functionality, it's true that it will still end up with so many update endpoints.

It's beyond the scope of this PR, but I guess it's a good timing to discuss about this.

timostamm: Considering a GUI in this form:

```
  Organization Base roles
------------------------------

  Repositories: [ Write ▼ ]

  Plugins:      [ Read  ▼ ]

  Templates:    [ Read  ▼ ]

                   [ Save ]
```

Would the client be expected to invoke three RPCs on Save? If the user changes only the repository base role and hits Save, should only the repository base role be transmitted?

A single RPC with partial updates could give us flexibility for different (or changing) use cases. However, I personally think that it might be too early for this abstraction.

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -189,3 +193,33 @@ message RemoveOrganizationMemberRequest {
...
+}
+
+message GetOrganizationSettingsResponse {
+  RepositoryRole repository_role = 1;
+  PluginRole plugin_role = 2;
+  TemplateRole template_role = 3;
+}
+
+message UpdateOrganizationSettingsRequest {
+  message RepositoryRoleMessage {
```
cyinma, line 209: I am struggling with the naming here, also not sure is it better to put these nested, or separately outside the request message

johanbrandhorst: We generally don't nest messages, so I would move this outside this definition. How about `OptionalRepositoryRole`?

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -189,3 +193,33 @@ message RemoveOrganizationMemberRequest {
...
+    PluginRole role = 1;
+  }
+  message TemplateRoleMessage {
+    TemplateRole role = 1;
+  }
+  // The ID of the organization for which to update the base roles.
+  string organization_id = 1;
+  RepositoryRoleMessage repository_role = 2;
+  PluginRoleMessage plugin_role = 3;
+  TemplateRoleMessage template_role = 4;
```
johanbrandhorst, line 222: Could you add a comment to each of these fields explaining that they are optional? If none is set, no update is made.

proto/buf/alpha/registry/v1alpha1/organization.proto
robbertvanginkel: I would follow the wrapper type structure from the well known types (https://buf.build/beta/wellknowntypes/file/main/google/protobuf/wrappers.proto), which has `<xyz>Value` messages wrap something with a single `value` field. I think they can be defined besides the enums in the current `role.proto`.

cyinma: > I would follow the wrapper type structure from the well known types (https://buf.build/beta/wellknowntypes/file/main/google/protobuf/wrappers.proto), which has `<xyz>Value` messages wrap something with a single `value` field. I think they can be defined besides the enums in the current `role.proto`.

this sounds good to me!

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -59,6 +59,10 @@ service OrganizationService {
   rpc UpdateOrganizationMember(UpdateOrganizationMemberRequest) returns (UpdateOrganizationMemberResponse);
   // RemoveOrganizationMember remove the role of an user in the organization.
   rpc RemoveOrganizationMember(RemoveOrganizationMemberRequest) returns (RemoveOrganizationMemberResponse);
+  // GetOrganizationSettings gets the settings of a organization, including organization base roles.
```
johanbrandhorst, line 62: Nit:
```suggestion
  // GetOrganizationSettings gets the settings of an organization, including organization base roles.
```

johanbrandhorst (approved):
LGTM

proto/buf/alpha/registry/v1alpha1/role.proto
```diff
@@ -57,3 +57,15 @@ enum PluginRole {
...
+  RepositoryRole value = 1;
+}
+
+message PluginRoleValue {
+  PluginRole value = 1;
+}
+
+message TemplateRoleValue {
+  TemplateRole value = 1;
+}
```
robbertvanginkel, line 71: @mfridman reminded me that as our enums always have an unspecified 0 value without semantic meaning, we don't need the wrapper values here.

I think we don't actually need these here, my bad for putting you on the wrong track.

mfridman: I liked your suggestion though, and it does keep it consistent with `google.protobuf.StringValue` and co.

I think we should keep it as is.. because it does give the flexibility to differentiate between unset vs unspecified, despite today us not having semantic meaning for the "zero" value.

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/role.proto
cyinma: yes, I like the idea of differentiate between unset vs unspecified, cause in our codebase the unspecified 0 value is usually meaning that it is invalid, while an nil *Value is a valid unset value

robbertvanginkel: https://docs.buf.build/lint/rules#enum_zero_value_suffix in our own docs does suggest that this 0 value should always be something without specified meaning, specifically so you can detect if it was explicitly set or not. 

Unlike some of the others covered by the well known types (bool, string, ints), enums have a specific value which we treat as "unspecified"/unset. If we wouldn't have this value I would agree that the wrapper type makes sense, but given that we already have a special 0 value I think we probably shouldn't also have a wrapper.

mfridman: Ye that's a fair point. The enum value shouldn't convey meaning and its best practice to _always_ have an unspecified value as 0.

In the event we wanted to convey something like "none" then we'd need to introduce a new (valid) enum, but this seems like a reasonable tradeoff.

We should also add this to our company style guide

mfridman (approved):

PR #718, opened 2021-11-11
Remove protoc version information

WillAbides:
This removes protoc version information from the requests sent to plugins except when using `buf protoc`


amckinney:
Do we need the custom transformation function? Why can't we just compare against testdata that encodes `(unknown)` since that's what we'll always expect?

Also, the `protoc` version we encode is not the version found via the `protoc` binary installed on the system - it's the `protoc` version we've used to test against the current version of `buf` [here](https://github.com/bufbuild/buf/blob/297a93c57e4558e89b5050edc9eac32ab3fb8933/make/go/dep_protoc.mk#L14) and [here](https://github.com/bufbuild/buf/blob/10303f87f35ba831e726dbb1f82788991dd6a26c/private/pkg/app/appproto/appprotoexec/appprotoexec.go#L61). The version of `protoc` on users' local machine is independent of the version we've used here.

WillAbides:
> Do we need the custom transformation function? Why can't we just compare against testdata that encodes `(unknown)` since that's what we'll always expect?

That would be the preferred way to go if this test tested against static data, but it runs protoc and compares protoc's generated files with buf's.

> 
> Also, the `protoc` version we encode is not the version found via the `protoc` binary installed on the system - it's the `protoc` version we've used to test against the current version of `buf` [here](https://github.com/bufbuild/buf/blob/297a93c57e4558e89b5050edc9eac32ab3fb8933/make/go/dep_protoc.mk#L14) and [here](https://github.com/bufbuild/buf/blob/10303f87f35ba831e726dbb1f82788991dd6a26c/private/pkg/app/appproto/appprotoexec/appprotoexec.go#L61). The version of `protoc` on users' local machine is independent of the version we've used here.

That's the protoc that `GetProtocVersion` gets the version of.

private/pkg/storage/diff.go
```diff
@@ -278,6 +296,7 @@ type diffOptions struct {
 	externalPaths         bool
 	oneExternalPathPrefix string
 	twoExternalPathPrefix string
+	transforms            []func(side, filename string, content []byte) []byte
```
amckinney, line 299: nit for consistency:
```suggestion
	transforms            []func(side string, filename string, content []byte) []byte
```

private/pkg/storage/diff.go
```diff
@@ -78,6 +78,20 @@ func DiffWithExternalPathPrefixes(
...
 }
 
+// DiffWithTransform returns a DiffOption that adds a transform function. The transform function will be run on each
+// file being compared before it is diffed. transform takes the arguments:
+//  side: one or two whether it is the first or second item in the diff
+//  filename: the filename including path
+//  content: the file content.
+// transform returns a string that is the transformed content of filename.
+func DiffWithTransform(
+	transform func(side, filename string, content []byte) []byte,
```
amckinney, line 88: nit for consistency:
```suggestion
	transform func(side string, filename string, content []byte) []byte,
```

private/pkg/storage/diff.go
```diff
@@ -78,6 +78,20 @@ func DiffWithExternalPathPrefixes(
 	}
 }
 
+// DiffWithTransform returns a DiffOption that adds a transform function. The transform function will be run on each
+// file being compared before it is diffed. transform takes the arguments:
+//  side: one or two whether it is the first or second item in the diff
+//  filename: the filename including path
+//  content: the file content.
+// transform returns a string that is the transformed content of filename.
+func DiffWithTransform(
```
amckinney, line 87: It'd be great if we could implement this functionality in `storagetesting` since it's only relevant for test functions. But given the nature of how you need to apply this option in the `Diff` implementation itself, I'm not sure it's possible.

amckinney (approved):
Ah I see it now, thanks for clarifying the use case. If we can't easily move this functionality into `storagetesting`, then this looks good to me.

private/pkg/storage/diff.go
WillAbides: I poked around at it for a minute, but like you said there isn't a good way to move this to another package without significant changes. We could move all of `diff.go` to `storagetesting`, but it is also used by `./private/pkg/storage/cmd/ddiff`. It doesn't seem right to have a `main` package depend on a package with "testing" in the name.

PR #719, opened 2021-11-11
adding search filter enum

smallsamantha:
adding filter to `Search` endpoint enabling a unified search instead of separate `SearchPlugins`, `SearchTemplates`, etc. for each entity


proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -16,6 +16,16 @@ syntax = "proto3";
 
 package buf.alpha.registry.v1alpha1;
 
+enum SearchFilter {
```
mfridman, line 19: I think others might have different opinions, but I think this is a great use of enums to drive search filtering.

Afaics we don't normally use enums on the request side of the RPC, so worth calling out to see if anyone has major objections?

smallsamantha: yeah for sure going to circulate! still have this in draft so i was going to wait for more comments on my other pr about our approach before i start propagating it into requests 😸

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -86,6 +96,10 @@ message SearchRequest {
   uint32 page_size = 2;
   // The first page is returned if this is 0.
   uint32 page_token = 3;
+  // Empty list means show all. Supplying one or more enums will 
+  // limit the search to the only the requested resources. 
+  // Supplying all possible enums is equivalent to empty list.
+  repeated SearchFilter filters = 4;
```
timostamm, line 102: So multiple enums translate into a logical `AND`? I don't understand how this would be meaningful for the `enum SearchFilter`, because no result can be User and Plugin at the same time.

It could be a logical `OR`, where supplying enum values returns only results that match one or another of the given resource types. Supplying all possible enums would show all results. 

However, I am not sure if we really need that filter capability in the search UI. Instead of going through multiple checkboxes for resource types, the user could just as well go through multiple radio options. For example, see the resource type selection in [GitHub's search](https://github.com/bufbuild/buf/search?q=foo) (sidebar on the left).

The repeated enum with `OR` behavior also adds ambiguity, for example `SEARCH_FILTER_UNSPECIFIED` in a request field, and inconvenient behavior of an empty filter list, so I think it might actually be best to go with a simple version for now, with an unrepeated field, and maybe rename `enum SearchFilter` to something more specific like `SearchResourceType`.

smallsamantha: multiple enums does translate to logical `OR` - when we say "Supplying all possible enums is equivalent to empty list." we mean "Supplying all possible enums is equivalent to empty list of enums." not "Supplying all possible enums is equivalent to empty list of results." Do you think that would be clearer? 

I think @SyrieBianco envisions the filter capability in a future search UI, not necessarily the current searchbar as it exists today. But I understand your point with respect to the Github search, that search implementation would actually be possible with the endpoints that exist today (`SearchPlugins`, `SearchTemplates`, etc)

timostamm: Ah, I misread that as _result_ list. @mfridman gave me some background on the filter enum that sounds very promising, and rebuts my nits.

smallsamantha: Updated the comment for clarity!!

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -16,6 +16,16 @@ syntax = "proto3";
 
 package buf.alpha.registry.v1alpha1;
 
+enum SearchFilter {
+  SEARCH_FILTER_UNSPECIFIED = 0;
+  SEARCH_FILTER_REPOSITORY = 1;
+  SEARCH_FILTER_ORGANIZATION = 2;
+  SEARCH_FILTER_USER = 3;
+  SEARCH_FILTER_TEAM = 4;
```
johanbrandhorst, line 24: ☠️

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -16,6 +16,16 @@ syntax = "proto3";
...
+  SEARCH_FILTER_PLUGIN = 5;
+  SEARCH_FILTER_TEMPLATE = 6;
```
johanbrandhorst, line 26: Nittest of nits: the order of this is a bit curious to me, as we traditionally group things like so:

```
(user, organization)
(repository, plugin, template)
(team)
```

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -86,6 +96,10 @@ message SearchRequest {
   uint32 page_size = 2;
   // The first page is returned if this is 0.
   uint32 page_token = 3;
+  // Empty list means show all. Supplying one or more enums will 
+  // limit the search to the only the requested resources. 
+  // Supplying all possible enums is equivalent to empty list of filters.
+  repeated SearchFilter filters = 4;
```
johanbrandhorst, line 102: With this in place, is the idea to remove `SearchUsers` etc?

johanbrandhorst (approved):
This LGTM, thanks for the clarifying comment!

proto/buf/alpha/registry/v1alpha1/search.proto
smallsamantha: yes that's what i would want to do!

proto/buf/alpha/registry/v1alpha1/search.proto
smallsamantha: i definitely didn't have a good reason to have it in the previous order so i just reordered to match better with the traditional grouping!

PR #720, opened 2021-11-11
Add breaking and lint configs to Module definition

doriable:
Opening up this PR for discussion on encoding `buf.yaml` configurations in our `Module` definition.
Discussion:
- Currently I have only the `v1` config for breaking and lint here. Do we want to support `v1beta1`? If so, the shape of the configuration would be the same, so do we want to define a separate structure specifically suffixed `V1Beta1` or should we encode the configuration version as either a `string` or `enum` within the configuration structure? If we encode it in the structure, in the future, if we change the shape of the configuration, it could get quite confusing. My stance would be, if we decide to support `v1beta1`, we should have a separate message and then set the breaking and lint configs as `oneof` fields, but open to getting some feedback on this.
- There have been a few things omitted, e.g. `name` and `build`, as noted in the issue.

Reference: #682


proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -64,3 +59,40 @@ message ModulePin {
...
+}
+
+// ModuleReference is a module reference.
+message ModuleReference {
+  string remote = 1;
+  string owner = 2;
+  string repository = 3;
+  // either branch, tag, or commit
+  string reference = 4;
+}
```
amckinney, line 98: We might still want this above the `ModulePin` definition so that similar definitions are placed adjacent to each other.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -34,6 +34,10 @@ message Module {
   //
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
   string documentation = 3;
+  // breaking_config is the breaking change detection configuration set for the module.
+  BreakingConfigV1 breaking_config = 5;
+  // lint_config is the lint configuration set for the module.
+  LintConfigV1 lint_config = 6;
```
amckinney, line 40: I'll need to spend some more time on exactly how we should do this, but we might want to lean on a `package` name to distinguish difference versions from each other. So we'd have multiple messages called `BreakingConfig`, but they'd be placed in different packages with different versions, if ever. For example, `buf.module.v1.BreakingConfig` and `buf.module.v2.BreakingConfig`, etc.

Like I said, we probably need to think through this a bit more, but I'm not actually sure we want to include the version suffix in the message name itself.

amckinney: If we split these out, we could model it as `buf.lint.v1.Config` and `buf.breaking.v1.Config` defined in `proto/buf/{lint,breaking}/v1/config.proto`, respectively. We've already cut `v1` for these configurations in the `buf.yaml`, so we should feel free to reserve the same version for the Protobuf-equivalent.

> We _might_ still want to include these definitions in the `alpha` sub-package path, but based on the `buf.yaml` `v1` release point above, we might not need to.

We could reference those types from the `Module` definition, and support a `oneof` later if we ever need to make a `v2` for either of the configurations (i.e. a `oneof` that contains both `buf.lint.{v1,v2}.Config`).

This would mean that the package that defines the `Module` would be independent from the package that defines the configuration it relies on. This could be useful if we ever have another system that needs to consume the `lint` and `breaking` configuration without the `Module` definition, too.

doriable:
Updated the PR based on our discussion today:

- We have split out the lint and breaking configurations into their own proto packages. This allows us to use the configurations in other places as needed without taking a dependency on the module package specifically. And, we are able to version them independently from module.
- We still encode the `rules_version` as a string in the config shape and that allows us to resolve the rule IDs and sets accordingly.
- We will _not_ be encoding the rule IDs and sets as enums, since we may want to support custom rule IDs and sets in the future. This is captured in the `Config` message documentation.

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+// the package. The package version refers to the config shape, the rules_version indicates which
+// rule IDs and catetgories should be used.
+//
+// The rule IDs and categories are not encoded as enums in this package because we may want to support custom rule
+// IDs and categories in the future. Callers will need to resolve the rule IDs and categories strings.
+message Config {
+  // rules_version represents the version of the breaking change rule IDs and categories that should be used with this config.
+  string rules_version = 1;
+  // use lists the rule IDs and/or categories that are included in the breaking change check.
+  repeated string use = 2;
```
bufdev, line 30: `use_ids` etc?

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+  // except lists the rule IDs and/or categories that are excluded from the breaking change check.
+  repeated string except = 3;
```
bufdev, line 32: `except_ids`?

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+  // ignore_unstable_packages ignores packages with a last component that is one of the unstable forms recognised
+  // by the PACKAGE_VERSION_SUFFIX:
+  //   v\d+test.*
+  //   v\d+(alpha|beta)\d+
+  //   v\d+p\d+(alpha|beta)\d+
+  bool ignore_unstable_packages = 6;
+}
+
+// IgnoreOnly represents a rule ID or category and the file and/or directory paths that are ignored for the rule.
+message IgnoreOnly {
```
bufdev, line 48: `IDPaths`? I think that's a bad name, but thinking if we want this message to reflect what it is @amckinney

proto/buf/alpha/breaking/v1/config.proto
doriable: Offline decision: `use_rule_ids` (and etc. for following fields)

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+  // ignore_unstable_packages ignores packages with a last component that is one of the unstable forms recognised
+  // by the PACKAGE_VERSION_SUFFIX:
+  //   v\d+test.*
+  //   v\d+(alpha|beta)\d+
+  //   v\d+p\d+(alpha|beta)\d+
+  bool ignore_unstable_packages = 6;
+}
+
+// RuleIDsPaths represents a rule ID or category and the file and/or directory paths that are ignored for the rule.
+message RuleIDsPaths {
```
amckinney, line 48: Since we only encode a single `rule_id`, we might want this to be `RuleIDPaths`. Plus, we might actually want it to be `IDPaths` like @bufdev suggested since it's technically a RuleID or a CategoryID.

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+  // use_rule_ids lists the rule IDs and/or categories that are included in the breaking change check.
+  repeated string use_rule_ids = 2;
+  // except_rule_ids lists the rule IDs and/or categories that are excluded from the breaking change check.
+  repeated string except_rule_ids = 3;
+  // ignore_paths lists the paths of directories and/or files that should be ignored by the breaking change check.
+  // All paths are relative to the root of the module.
+  repeated string ignore_paths = 4;
+  // ignore_rule_ids_paths is a map of rule IDs and/or categories to directory and/or file paths to exclude from the
+  // breaking change check.
+  repeated RuleIDsPaths ignore_rule_ids_paths = 5;
```
amckinney, line 38: If we rename the `message` name, update the field name too.
```suggestion
  repeated RuleIDsPaths ignore_rule_id_paths = 5;
```

amckinney: We _could_ still name the field according to the configuration name, too.
```suggestion
  repeated RuleIDsPaths ignore_only_paths = 5;
```

proto/buf/alpha/breaking/v1/config.proto
```diff
@@ -0,0 +1,51 @@
...
+// Config represents the breaking change configuration for a module. The rule IDs and categories are defined
+// by the rules_version and apply across the config. The rules_version is independent of the version of
+// the package. The package version refers to the config shape, the rules_version indicates which
+// rule IDs and catetgories should be used.
+//
+// The rule IDs and categories are not encoded as enums in this package because we may want to support custom rule
+// IDs and categories in the future. Callers will need to resolve the rule IDs and categories strings.
+message Config {
+  // rules_version represents the version of the breaking change rule IDs and categories that should be used with this config.
+  string rules_version = 1;
```
amckinney, line 28: I think `version` is sufficient to keep this concise. Similar with `use_rule_ids` -> `use_ids`, and so on. I'm thinking about accessing these fields on the Go types like `.Version` and `.UseIds`, etc.

proto/buf/alpha/breaking/v1/config.proto
amckinney: Yea, I'm fine with `IDPaths`. It doesn't sound that great, but I like the `KeyValue` naming convention for messages like this.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -34,6 +36,10 @@ message Module {
   //
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
   string documentation = 3;
+  // breaking_config is the breaking change detection configuration set for the module.
+  buf.alpha.breaking.v1.Config breaking_config = 5;
+  // lint_config is the lint configuration set for the module.
+  buf.alpha.lint.v1.Config lint_config = 6;
```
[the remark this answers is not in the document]
doriable: Yeah, originally that's what I had, but I think we can just push license to later, lol.

proto/buf/alpha/lint/v1/config.proto
```diff
@@ -0,0 +1,58 @@
...
+//
+// The rule and category IDs are not encoded as enums in this package because we may want to support custom rule
+// and category IDs in the future. Callers will need to resolve the rule and category ID strings.
+message Config {
+  // version represents the version of the lint rule and category IDs that should be used with this config.
+  string version = 1;
+  // use_rule_ids lists the rule and/or category IDs that are included in the lint check.
+  repeated string use_rule_ids = 2;
+  // except_rule_ids lists the rule and/or category IDs that are excluded from the lint check.
+  repeated string except_rule_ids = 3;
```
amckinney, line 32: According to the edits above, don't forget these.

```suggestion
  // use_ids lists the rule and/or category IDs that are included in the lint check.
  repeated string use_ids = 2;
  // except_ids lists the rule and/or category IDs that are excluded from the lint check.
  repeated string except_ids = 3;
```

amckinney (approved):
As long as we address the last comment on consistent naming, this looks great!

proto/buf/alpha/lint/v1/config.proto
doriable: 😳

bufdev (approved):

PR #723, opened 2021-11-12
add user state in ListUsers request and add count users endpoint

cyinma:
This is to add user state filter in the request of `ListUsers` and add count users endpoints.

Although we are likely to need the counting endpoints for other resources, it’s still unclear except the count users one. To avoid creating unused endpoints, I have decided to leave the remaining counting endpoints later when we sure we need them.


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -91,12 +97,15 @@ message ListUsersRequest {
...
   bool reverse = 3;
+  // request with unspecified user state will return users with both user state
+  UserState user_state = 4;
 }
 
 message ListUsersResponse {
   repeated User users = 1;
   // There are no more pages if this is empty.
   string next_page_token = 2;
+  int32 total_size = 3;
```
bufdev, line 108: Unless we are adding `total_size` to every `List*` endpoint, this breaks our convention for `List*` endpoints. If we do want this data, it should likely be a separate RPC.

Also if we do expose this data somehow, it should be a `uint32`.

mfridman:
I'm curious, why is this necessary? 

The returned response is an array of items, and all languages can call `.length` or `len()` to get the total size..

cyinma:
> I'm curious, why is this necessary?
> 
> The returned response is an array of items, and all languages can call `.length` or `len()` to get the total size..

The original thought is to also provide the total size of the elements (e.g. total count of users), while the array of items in the response is limited by the page size.

I think we will need to find out another way to achieve this, probably some new endpoints.

proto/buf/alpha/registry/v1alpha1/user.proto
cyinma: The idea of adding `int32 total_size` in the response of Listing endpoints was came from this [Google Cloud APIs suggested Common design patterns](https://cloud.google.com/apis/design/design_patterns#list_pagination), and I agree that `uint32` is actually a better data type in this case.

For the concern of the convention, I agree that we should keep the convention for the Listing endpoints. In the beginning I was thinking about adding some new endpoints for the information needed, but because of the concern of API bloat, I think adding the `total_size` to all of the Listing endpoints might be a better approach, and I have created a separated PR for it (#734), and this PR will only for the `ListUsers` with user state filter.

cc @robbertvanginkel @mfridman as they also have the concern of API bloat

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -91,6 +99,8 @@ message ListUsersRequest {
   // The first page is returned if this is empty.
   string page_token = 2;
   bool reverse = 3;
+  // request with unspecified user state will return users with both user state.
```
johanbrandhorst, line 102: Nit: this comment will become outdated if we add another state
```suggestion
  // If the user_state is unspecified, all users are included.
```

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -91,6 +99,8 @@ message ListUsersRequest {
...
+  UserState user_state = 4;
```
johanbrandhorst, line 103: Should this be `user_state_filter` instead?

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -131,3 +141,12 @@ message UpdateUserServerRoleRequest {
 }
 
 message UpdateUserServerRoleResponse {}
+
+message CountUsersRequest {
+  // request with unspecified user state will return the number of users with both user state.
```
johanbrandhorst, line 146: Same here

proto/buf/alpha/registry/v1alpha1/user.proto
cyinma: sounds good, I guess we need to apply this to `CountUsersRequest` too?

johanbrandhorst: Yes

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -131,3 +141,12 @@ message UpdateUserServerRoleRequest {
 }
 
 message UpdateUserServerRoleResponse {}
+
+message CountUsersRequest {
+  // If the user_state_filter is unspecified, all users are included.
+  UserState user_state_filter = 4;
```
johanbrandhorst, line 147: Nit: should this be
```suggestion
  UserState user_state_filter = 1;
```

johanbrandhorst (approved):

proto/buf/alpha/registry/v1alpha1/user.proto
cyinma: good catch, thanks!

PR #727, opened 2021-11-15
adding listoptions for list rpcs

smallsamantha:
add sorting enums to all list endpoints

for cli usage, defaulting to `create_time_desc`


proto/buf/alpha/registry/v1alpha1/list.proto
```diff
@@ -0,0 +1,26 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+message ListOptions {
```
bufdev, line 19: This is sort of an anti-pattern in Protobuf - we generally want to expand this for each RPC. Combining this makes the API more difficult for end-users to deal with, and generally only makes sense from an implementation perspective (which is the implementer's problem, not the client's).

proto/buf/alpha/registry/v1alpha1/list.proto
```diff
@@ -0,0 +1,26 @@
...
+  // The number of items to return.
+  uint32 page_size = 1;
+  // The first page is returned if this is empty.
+  string page_token = 2;
+  // string value following SQL syntax specifying order for list results.
+  string order_by = 3;
```
johanbrandhorst, line 25: Following on from Peter's comment, if this is made more explicit, I would argue we could type this field too. This could be an enum, combined with an `bool ascending = 4`. What do you think?

johanbrandhorst: Actually we already have reverse, so no need to do the `ascending` part.

smallsamantha: if we choose to go with an enum do we want it to exhaust every column possibility that we'd want to eventually sort by?

johanbrandhorst: I think we can easily support adding things, so we should just add the ones we need right now

bufdev: I agree with @johanbrandhorst that this should be an enum in Protobuf, you can add to enums without breaking compatibility

smallsamantha: sounds good - enum it is 👍

proto/buf/alpha/registry/v1alpha1/list.proto
mfridman: Interesting, I didn't know this was an anti-pattern. 

I like the implementor vs consumer point, but being able to logically group options potentially makes it easier for the consumer (in addition to the implementor)? Both sides get consistent behaviour across all List endpoints.

bufdev: You can have consistent behavior without having to wrap these fields in another message

proto/buf/alpha/registry/v1alpha1/list.proto
```diff
@@ -0,0 +1,26 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+// OrderedColumn defines the options available for specifying
+// the column by which the results should be ordered.
+enum OrderedColumn {
```
robbertvanginkel, line 21: I don't think we should have a shared enum for any listing endpoints as I imagine the sort options won't be the same across all endpoints. For example, sorting by owner column for ListUsers seems odd.

For a typed API, I think I would prefer the enums to be `{field}_ASC/DESC` options rather than having an additional `bool reversed`. If we ever allow the api to expand (sort by x first, then by y) you can express that.

Other nit: column to me is a database implementation detail and I would avoid putting it in

private/buf/cmd/buf/command/alpha/registry/token/tokenlist/tokenlist.go
```diff
@@ -86,6 +89,12 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		bufprint.FormatText.String(),
 		fmt.Sprintf(`The output format to use. Must be one of %s`, bufprint.AllFormatsString),
 	)
+	flagSet.Uint32Var(
+		&f.OrderedColumn,
+		orderedColumnFlagName,
+		1,
+		`The column by which to sort results.`,
```
johanbrandhorst, line 96: What does this look like on the command line? Just `--ordered-column 1`? Can we make this use a string representation, if we need to add it at all?

proto/buf/alpha/registry/v1alpha1/list.proto
johanbrandhorst: +1 on RPC specific enums, sorry it wasn't made clear in my initial comment. I would argue that `reversed` is already well established in the API, so there's no need to remove the existing one, but the point about being able to order by two or more columns is a neat feature of combining them! I think we could still allow that in the future if we wanted to.

Also  +1 that these should be named based on fields on the resource they operate on, not leaking SQL jargon like column. ListUsers could have `UserSort` and `USER_SORT_USERNAME`, etc.

smallsamantha: thoughts here @johanbrandhorst @bufdev? i don't know if i think it makes sense to have a different sorting type enum for each entity we need to sort (users vs repositories vs plugins etc) - that would make me lean towards the original idea of simply a string for the ordering query instead of a bunch of different enums

and noted the comment about columns - def can make the adjustment to avoid in naming

smallsamantha: dangit my comment was too late - ok noted @johanbrandhorst, will revisit

robbertvanginkel: > I would argue that reversed is already well established in the API, so there's no need to remove the existing on

And similarly the established pattern in the api is that you cannot change the sort order, only reverse it. Now that we're exchanging that to allow specifying what to sort on, I think it makes sense to change into ascending/descending which are standard and have clear meaning, rather than continue to propagate `reversed` (which only has relative meaning compared to some default).

johanbrandhorst: That's totally fair 👍🏻

smallsamantha:
wanted to sum up the revisions i made based on comments.
new changes include the following sort enums:
- `OrganizationSort`
- `PluginSort` (templates and plugins)
- `RepositorySort` (repositories, repositorycommits, repositorytags, repositorybranch which is deprecated if i remember correctly but i included it anyway)
- `TokenSort`
- `UserSort`

for cli usage i did not include sorting as a flag, i just have it automatically fetching in create_time descending order without a way to pick a different option but am interested in hearing peoples' ideas for if we did want a flag option. what would we want people to pass as the strings? let's say for plugins, my vote is that it would be the whole spelled out enum: `PLUGIN_SORT_CREATE_TIME_ASC`. any other ideas?

cc @bufdev @johanbrandhorst @robbertvanginkel @mfridman

johanbrandhorst (commented):
I think generally this might be scoped too widely - we should figure out what the product requirement is exactly, I think. If we want to fix the ordering of templates, we should add ordering to those?

SyrieBianco:
> I think generally this might be scoped too widely - we should figure out what the product requirement is exactly, I think. If we want to fix the ordering of templates, we should add ordering to those?

I think this is a good point — the original scoping here is just to fix the ordering of the templates, but looking farther into the future I saw this being able to sort entity lists. Powering things like the search page and the admin panel, but also even just entity lists on your own page. Something like:
Screen Shot 2021-11-17 at 10 13 05 AM

or

Screen Shot 2021-11-17 at 10 15 42 AM

I haven't had time to dig into the technicals here, but I agree with Johan that I/we should probably weigh in more on not just the API but the product decisions driving it. 

I'm not saying to block this PR on it, but @smallsamantha let's sync offline and we can mind meld on product direction behind this.

amckinney:
Caught up offline and it sounds like we'll revisit this later.

PR #728, opened 2021-11-15
remove unused endpoints

cyinma:
This is to remove the unused endpoints which are unsupported.


mfridman (approved):
Thanks for cleaning this up. Glad we got rid of the lingering scopes references.

PR #730, opened 2021-11-15
Use a separate repository service per remote for resolving dependencies

### LINKED ISSUE #729: For `buf mod update`, we need to use a separate repository service for each dependency
Currently, we are using the remote from the module reference to instantiate a repository service for resolving dependencies:

https://github.com/bufbuild/buf/blob/1275b7dbc179d2d78b0678599a7fa4796d3a45c7/private/buf/cmd/buf/command/mod/modupdate/modupdate.go#L213-L216

However, we should be using a separate repository service for each remote based on the dependency we're trying to resolve.

doriable:
This creates a new repository service per remote across dependencies, and then resolves dependencies based on their remote.

Fixes #729

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 - Change message documentation for fields to be either a single field or a oneof set of fields. This is a breaking API change.
+- Use a separate repository service to for each dependency remote to resolve dependencies for `buf mod update`. Previously, we used a single repository service based on the remote
+  from the module, so it was unable to resolve dependencies from differente remotes.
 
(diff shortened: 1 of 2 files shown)


amckinney (approved):
Nice! Don't forget to include an item in the `CHANGELOG.md`. This isn't really customer facing since we still only advertise `buf.build`, but it'll be nice to track down when this was fixed.

doriable:
> Nice! Don't forget to include an item in the `CHANGELOG.md`. This isn't really customer facing since we still only advertise `buf.build`, but it'll be nice to track down when this was fixed.

Ah, thanks for the reminder, haha <3

PR #734, opened 2021-11-16
add total size in the response of all Listing endpoints

cyinma:
This is to add `total_size` in the response of all Listing endpoints. Some of them will be part of the server admin panel features, and to keep the convention for the Listing endpoints, add the `total_size` to all of them.

We might only implement part of them, and let the remaining return with 0 `total_size` until a follow up PR, depending on the complexity and priority, if this is acceptable.


bufdev (changes requested):
I still don't understand why we are adding this to all `List*` endpoints, it is not a standard pattern for `List*` endpoints across the industry.

PR #735, opened 2021-11-17
initial auditv1 package for audit logging

smallsamantha:


proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.audit.v1;
```
bufdev, line 17: Is there a reason this isn't in `buf.alpha`, and a reason this isn't `v1alpha1`?

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+    RequestInfo request_info = 4;
+    // Represents object(s) per event, e.g. for remote generation,
+    // both repository, template and all the template's
+    // plugins are objects.
+    repeated Object objects = 5;
+    // Details of any error that occurred upon completion of 
+    // the event request.
+    Error error = 6;
+    // Free-form to guarantee stability of fields over time.
+    google.protobuf.Struct request = 7;
```
bufdev, line 37: When did we discuss using this? I would be strongly against this

proto/buf/audit/v1/envelope.proto
smallsamantha: how did you find this draft lol

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+package buf.audit.v1;
+
+import "google/protobuf/struct.proto";
+import "google/protobuf/timestamp.proto";
+
+message EventEnvelope {
+    string event_id = 1;
+    // The actor represents the subject initiating execution
+    // of the event
+    Actor actor = 2;
```
bufdev, line 26: Maybe I'm misremembering, but I thought we were flattening this?

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+    google.protobuf.Timestamp event_time = 3;
+    RequestInfo request_info = 4;
```
bufdev, line 28: Is there a reason we're wrapping this in RequestInfo as opposed to flattening?

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+    // Represents object(s) per event, e.g. for remote generation,
+    // both repository, template and all the template's
+    // plugins are objects.
+    repeated Object objects = 5;
```
bufdev, line 32: Why is this repeated?

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+    Error error = 6;
+    // Free-form to guarantee stability of fields over time.
+    google.protobuf.Struct request = 7;
+    google.protobuf.Struct response = 8;
+  }
+  
+  message UserActor {
+    string id = 1;
+    string username = 2;
+    string source_ip = 3;
```
bufdev, line 44: It feels like there should be an `IP` type that is parsed out to its constituent parts, but not sure

proto/buf/audit/v1/envelope.proto
```diff
@@ -0,0 +1,108 @@
...
+  
+  message TemplateObject {
+    string id = 1;
+    string name = 2;
+    string owner_id = 3;
+    string owner_name = 4;
+    bool public = 5;
+  }
+  
+  message Object {
```
bufdev, line 96: Why do we need a wrapper message for this?

proto/buf/audit/v1/envelope.proto
smallsamantha: there may be a case when a single action affects more than one "object" so we need to include them all in the audit log entry

smallsamantha:
definitions will not live here

PR #736, opened 2021-11-17
add endpoints for explicit resources role

cyinma:
This is to add the explicit resources role control in repositories, plugins, and templates.


proto/buf/alpha/registry/v1alpha1/authz.proto
```diff
@@ -76,6 +76,15 @@ service AuthzService {
...
   rpc UserCanSeeServerAdminPanel(UserCanSeeServerAdminPanelRequest) returns (UserCanSeeServerAdminPanelResponse);
+  // UserCanManageRepositoryContributor returns whether the user is authorized to manage
+  // any contributors to the repository and the list of roles they can manage.
+  rpc UserCanManageRepositoryContributor(UserCanManageRepositoryContributorRequest) returns (UserCanManageRepositoryContributorResponse);
+  // UserCanManagePluginContributor returns whether the user is authorized to manage
+  // any contributors to the plugin and the list of roles they can manage.
+  rpc UserCanManagePluginContributor(UserCanManagePluginContributorRequest) returns (UserCanManagePluginContributorResponse);
+  // UserCanManageTemplateContributor returns whether the user is authorized to manage
+  // any contributors to the template and the list of roles they can manage.
+  rpc UserCanManageTemplateContributor(UserCanManageTemplateContributorRequest) returns (UserCanManageTemplateContributorResponse);
```
cyinma, line 87: from the previous experience of organization roles, it seems like we don't actually need to separate the roles that user can add/update/remove, they should be the same. If this works, we might want to also update the organization role authz

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -136,6 +138,12 @@ service PluginService {
   // a plugin may cause breaking changes for templates using that plugin,
   // and should be done with extreme care.
   rpc DeletePlugin(DeletePluginRequest) returns (DeletePluginResponse);
+  // AddPluginContributor add a role to an user in the plugin.
+  rpc AddPluginContributor(AddPluginContributorRequest) returns (AddPluginContributorResponse);
+  // UpdatePluginContributor update the user's contributorship information in the plugin.
+  rpc UpdatePluginContributor(UpdatePluginContributorRequest) returns (UpdatePluginContributorResponse);
+  // RemovePluginContributor remove the role of an user in the plugin.
+  rpc RemovePluginContributor(RemovePluginContributorRequest) returns (RemovePluginContributorResponse);
```
johanbrandhorst, line 146: We could do a similar things as we discussed for recommended repositories here; `SetUserPluginRole`, which does both add, update and remove (set to UNSPECIFIED). What do you think?

cyinma: my only concern is the logic behind this is probably much more complicated than recommendation, so group them together might increase the complexity of implementation too much, which I guess is also something we want to avoid? there is some trade off and I'm not really sure which one is better

johanbrandhorst: I would think it shouldn't be much harder to use, but would be good to get @timostamm to weigh in.

timostamm: There will be likely be three separate call sites in the frontend, I don't see any complication on this side if the three RPCs were rolled together.

cyinma: cool, will update to group them together then

proto/buf/alpha/registry/v1alpha1/authz.proto
```diff
@@ -76,6 +76,15 @@ service AuthzService {
   // UserCanSeeServerAdminPanel returns whether the user is authorized
   // to see server admin panel.
   rpc UserCanSeeServerAdminPanel(UserCanSeeServerAdminPanelRequest) returns (UserCanSeeServerAdminPanelResponse);
+  // UserCanManageRepositoryContributor returns whether the user is authorized to manage
+  // any contributors to the repository and the list of roles they can manage.
+  rpc UserCanManageRepositoryContributor(UserCanManageRepositoryContributorRequest) returns (UserCanManageRepositoryContributorResponse);
```
johanbrandhorst, line 81: Should this be plural?
```suggestion
  // UserCanManageRepositoryContributors returns whether the user is authorized to manage
  // any contributors to the repository and the list of roles they can manage.
  rpc UserCanManageRepositoryContributors(UserCanManageRepositoryContributorsRequest) returns (UserCanManageRepositoryContributorsResponse);
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -136,6 +138,8 @@ service PluginService {
   // a plugin may cause breaking changes for templates using that plugin,
   // and should be done with extreme care.
   rpc DeletePlugin(DeletePluginRequest) returns (DeletePluginResponse);
+  // SetPluginContributor set the role of an user in the plugin.
```
johanbrandhorst, line 141: Nit
```suggestion
  // SetPluginContributor sets the role of a user in the plugin.
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -156,6 +160,8 @@ service PluginService {
   rpc DeleteTemplate(DeleteTemplateRequest) returns (DeleteTemplateResponse);
   // CreateTemplateVersion creates a new template version.
   rpc CreateTemplateVersion(CreateTemplateVersionRequest) returns (CreateTemplateVersionResponse);
+  // SetTemplateContributor set the role of an user in the template.
+  rpc SetTemplateContributor(SetTemplateContributorRequest) returns (SetTemplateContributorResponse);
```
johanbrandhorst, line 164: Same here

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -257,6 +263,17 @@ message DeletePluginRequest {
 
 message DeletePluginResponse {}
 
+message SetPluginContributorRequest {
+  // The ID of the plugin for which the user will be set a role.
```
johanbrandhorst, line 267:
```suggestion
  // The ID of the plugin for which the user's role will be set .
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -257,6 +263,17 @@ message DeletePluginRequest {
...
+  string plugin_id = 1;
+  // The ID of the user for which to be set a role in the plugin.
```
johanbrandhorst, line 269:
```suggestion
  // The ID of the user whose role will be set.
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -257,6 +263,17 @@ message DeletePluginRequest {
...
+  string user_id = 2;
+  // The new role of the user in the plugin,
+  // unspecified role means to remove the user's role in the plugin.
```
johanbrandhorst, line 272:
```suggestion
  // The role to assign to the user.
  // Setting UNSPECIFIED means removing the user's role.
```

cyinma: applied these suggested changes to both repository, plugin and template endpoints

johanbrandhorst (approved):

PR #738, opened 2021-11-18
update comments in ListUsers and CountUsers for correct expected beha…

cyinma:
update comments in ListUsers and CountUsers for correct expected behaviour


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -99,7 +99,7 @@ message ListUsersRequest {
   // The first page is returned if this is empty.
   string page_token = 2;
   bool reverse = 3;
-  // If the user_state_filter is unspecified, all users are included.
+  // The user_state_filter must be specified, otherwise an error will be returned.
```
johanbrandhorst, line 102: I'm confused, I thought the old functionality was nice, I think we should be allowed to not specify this, even if we don't need it right now. My comment about allowing it in the application code was specific to that code, not the RPC in general. We can leave this as is if you think it makes more sense though 👍🏻.

PR #739, opened 2021-11-18
Encode breaking and lint config in module

doriable:
This adds the breaking and lint configurations to the module build.

Fixes: #682


proto/buf/alpha/lint/v1/config.proto
```diff
@@ -41,6 +41,8 @@ message Config {
   // this value with the given string.
   string enum_zero_value_suffix = 6;
   // rpc_allow_same_request_response allows the same message type for both the request and response of an RPC.
+  bool rpc_allow_same_request_response = 11;
```
amckinney, line 44: We haven't actually used this over the wire yet, so I'd prefer to reorder the field numbers so that this actually fits into the mold (i.e. this would be `7`, then the following fields would all be incremented by 1).

private/bufpkg/bufmodule/module.go
```diff
@@ -56,11 +61,45 @@ func newModuleForProto(
...
+			return nil, err
+		}
+	}
+	if protoBreakingConfig.GetVersion() == bufconfig.V1Beta1Version {
+		var err error
+		breakingConfig, err = bufbreakingconfig.NewConfigV1Beta1ForProto(protoBreakingConfig)
+		if err != nil {
+			return nil, err
+		}
+	}
```
amckinney, line 79: It's impossible for both of these `if` conditions to be true, so we can better convey this with either a `switch protoBreakingConfig.GetVersion()`, or by encapsulating all of this in a simple helper function that returns the `*bufbreakingconfig.Config` type like so:

```go
func breakingConfigForProto(protoBreakingConfig *breakingv1.Config) (*bufbreakingconfig.Config, error) {
  if protoBreakingConfig == nil {
    // Not all modules will have a *breakingv1.Config, so we
    // don't validate against it if it doesn't exist.
    return nil, nil
  }
  switch version := protoBreakingConfig.GetVersion(); version {
    case bufconfig.V1Beta1Version:
      return bufbreakingconfig.NewConfigV1Beta1ForProto(protoBreakingConfig)
    case bufconfig.V1Version:
      return bufbreakingconfig.NewConfigV1ForProto(protoBreakingConfig)
    default:
      return nil, fmt.Errorf("invalid version %q found in proto breaking configuration", version)
  }
}
```

That way you can simplify the error handling here to the following:
```go
breakingConfig, err := breakingConfigForProto(protoModule.GetBreakingConfig())
if err != nil {
  return nil, err
}
lintConfig, err := lintConfigForProto(protoModule.GetLintConfig())
if err != nil {
  return nil, err
}
return newModule(
  ...
)
```

You could also argue that the `breakingConfigForProto` function could be in the `bufbreakingconfig` package, which would let you un-export `bufbreakingconfig.NewConfig{V1Beta1,V1}ForProto`.

amckinney (commented):
Moving all of the `bufcheck` packages out and into `bufpkg` looks good to me - we'll need to invoke the `buf{breaking,lint}.Handler` types from the BSR later so we'll need those types in `bufpkg` anyway. Otherwise, just a couple nits but looks good.

We should add a sufficient number of tests to once this is ported to the BSR to ensure that the changes interoperate well with previously persisted modules.

Wire compatibility should be fine since we're just adding fields, but we should make sure we don't validate against modules that don't contain these values (which is what you're already doing here).

proto/buf/alpha/lint/v1/config.proto
doriable: I would, but I ran into the `buf breaking` check in our lint. It appears we have started to enforce the breaking change detection? This was something I wanted to discuss, haha.

amckinney: It's because we have a `v1` package name here - we only protect against `v1alpha1`. But this is a new package and it's unreleased, so feel free to break it this one time.

private/bufpkg/bufmodule/module.go
doriable: +1 on the switch. As for rolling them up into a single `NewConfigForProto` function, I think I prefer it to follow the current model for the external config (the separate functions for v1 and beta1) so that it's consistent (even though it's slightly repetitive code, the pattern already exists).

amckinney (approved):
Assuming the field numbers are reordered and the other nits are addressed, this looks good to me - feel free to merge once those are good to go.

proto/buf/alpha/lint/v1/config.proto
doriable: Ah yeah -- urgh, lol. Okay, sounds good, since I have your written approval here, I will push up that change :)

bufdev:
Looks sane to me, will let Alex take this for approval

PR #740, opened 2021-11-19
reorder GetTemplateVersionRequest/response

smallsamantha:
reordering requests/response to conform with style


johanbrandhorst (approved):
🙏🏻

PR #745, opened 2021-11-29
new list resource contributors endpoints

cyinma:
This is to add new endpoints for BSR to list the resource contributors.


proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -118,6 +119,26 @@ message PluginVersionMapping {
...
 }
 
+message PluginContributor {
+  User user = 1;
+  // The ID of the plugin for which the role belongs to.
+  string plugin_id = 2;
+  // The role that the user explicitly has in the plugin above.
+  PluginRole explicit_role = 3;
+  // The role that the user implicitly has from the organization that owns the plugin.
+  PluginRole implicit_role = 4;
```
johanbrandhorst, line 129: What do you think about
```suggestion
  // Optionally defines the role that the user has implicitly against the plugin through the owning organization.
  // If the plugin does not belong to an organization, this is unset.
  PluginRole implicit_role = 4;
```

Also is this set for the owner if the plugin belongs to the user, or is that considered an explicit role? I guess it should be explicit.

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -118,6 +119,26 @@ message PluginVersionMapping {
   bool deleted = 4;
 }
 
+message PluginContributor {
+  User user = 1;
+  // The ID of the plugin for which the role belongs to.
```
johanbrandhorst, line 124: Nit:
```suggestion
  // The ID of the plugin which the role belongs to.
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -118,6 +119,26 @@ message PluginVersionMapping {
...
+  string plugin_id = 2;
+  // The role that the user explicitly has in the plugin above.
```
johanbrandhorst, line 126:
```suggestion
  // The role that the user has been explicitly assigned against the plugin.
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -140,6 +161,8 @@ service PluginService {
   rpc DeletePlugin(DeletePluginRequest) returns (DeletePluginResponse);
   // SetPluginContributor sets the role of a user in the plugin.
   rpc SetPluginContributor(SetPluginContributorRequest) returns (SetPluginContributorResponse);
+  // ListPluginContributors returns the list of contributors that has an explicit role in the plugin.
```
johanbrandhorst, line 164:
```suggestion
  // ListPluginContributors returns the list of contributors that has an explicit role against the plugin.
  // This does not include users who have implicit roles against the plugin, unless they have also been
  // assigned a role explicitly.
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
cyinma: the first line is good to me, the second line seems a little bit confusing, as it can also be unset if that is an outside contributor for the organization-owning plugin, how about:
`If the plugin does not belong to an organization or the user is not part of the owning organization, this is unset.` ?

johanbrandhorst: Sounds good!

johanbrandhorst (approved):
LGTM

PR #746, opened 2021-11-29
Add protobufs needed to implement repository tracks

WillAbides:
This adds the protobuf definitions for RepositoryTrackService and updates PushRequest to include tracks.

This doesn't change buf.

https://github.com/bufbuild/buf/commit/2876d813756dfcbdb3927e07c2645b6e82eb957e shows how these protobufs will be used.


proto/buf/alpha/registry/v1alpha1/repository_track.proto
```diff
@@ -0,0 +1,72 @@
...
+  // There are no more pages if this is empty.
+  string next_page_token = 2;
+}
+
+message AppendRepositoryTrackRequest {
+  string repository_track_id = 1;
+  repeated string commit_id = 2;
+}
+
+message AppendRepositoryTrackResponse {}
```
amckinney, line 72: Let's discuss this one - I'm not sure what's best to do here just yet -

Do we actually need these endpoints to start? I wonder if we should just require that the `track` is an optional parameter to the `PushRequest` like you have above. The `PushService` already references branch resources without explicitly calling the `RepositoryBranchService`, so I imagine we would be able to do something similar here.

I know this could be useful for adding a commit that already exists on another track, but what do we imagine this user experience to be (e.g. the CLI command and/or the front-end experience)?

Even if we have a `buf registry track append <$track_name> <$commit_id>` users will still be able to `buf push --track <$track_name>` and `buf push --track <$another_track_name>` and create two separate commits (even though they have the same content).

amckinney (approved):
Nice, this looks great 👏

PR #747, opened 2021-11-30
Refactor `buflintconfig` and `bufbreakingconfig` to match external config fields

doriable:
This changes the implementations of `buflintconfig.Config` and `bufbreakingconfig.Config` to closer match their external representations. This is in preparation for storing the `buflintconfig.Config` and `bufbreakingconfig.Config` structures on the module and ensuring that we have an external representation to work with for the digest (#682).

This change also moves the point at which we compile the config into the `bufcheck.internal.Config` representation when we run the check.

A couple of points of discussion:
- We are unfortunately repeating the constants for `v1` and `v1beta1` in the respective `buflintconfig` and `bufbreakingconfig` subpackages. This is because we cannot take a dependency on `bufconfig` (since `bufconfig` takes a dependency on both `buflintconfig` and `bufbreakingconfig`.
- We are now only compiling the config when we run the check, and we will need to compile the config each time we run the check. This should be fine, since we are not currently caching the compiled configs and this should not change the user experience of the CLI, since we need to compile the module config and module each time we run the checks.


private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -21,95 +21,92 @@ import (
...
 type Config struct {
-	// Rules are the rules to run.
-	//
-	// Rules will be sorted by first categories, then id when Configs are
-	// created from this package, i.e. created wth ConfigBuilder.NewConfig.
-	Rules                  []Rule
-	IgnoreIDToRootPaths    map[string]map[string]struct{}
-	IgnoreRootPaths        map[string]struct{}
+	Use    []string
+	Except []string
```
amckinney, line 32: Note that this means we'll generate a new digest when a user changes their configured rules/categories even if it won't have an effect on the resulting `[]bufcheck.Rule`.

I actually think this is good since it means we're capturing changes in the users' `buf.yaml` configuration. Plus, it will prevent a hard-to-debug edge case when a user thinks they've made a change to their `lint` configuration, but their push gets rejected by the server since it looks like the same content (because the rules get unrolled to the same set as the previous commit).

amckinney (commented):
This is looking great so far! Bunch of comments about structure and naming, but this looks like it's moving in the right direction!

private/buf/cmd/buf/command/config/configlslintrules/configlslintrules.go
```diff
@@ -121,7 +121,10 @@ func run(
 			}
 		}
 	} else {
-		rules = config.Lint.GetRules()
+		rules, err = config.Lint.GetRules()
+		if err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
doriable: Yeah, I think I like this idea -- moving it up to `buflint` and `bufbreaking` respectively is also good, because the `internal` package is actually internal to `bufcheck`, and not the config packages. Going to update these + the naming :D

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
doriable: I believe this is very reasonable. I mean, if we really wanted to capture the rules as part of the digest, we can always build the rules during the digest and use that, but I think encoding the config structure in this manner is also very reasonable.

amckinney (approved):
Awesome!

PR #748, opened 2021-11-30
Add `breakingConfig` and `lintConfig` to Module and upgrade to B3 digest

doriable:
This PR adds the `breaking` and `lint` config structures to the module build.
It also incorporates the changes from https://github.com/bufbuild/buf/pull/715, which includes the module identity and metadata (e.g. configuration and documentation) as a part of the digest.

Discussion - module identity as a part of `b3` digest:
In order to support the module identity as part of the `b3` digest, we need to ensure the module identity is available in both the module that is put into the module cache and also the module that is being pulled out of the cache. To do this, 2 changes were made to the module interface:

- `getModuleIdentity() bufmoduleref.ModuleIdentity` -> `ModuleIdentity() bufmoduleref.ModuleIdentity`: We need to export the module identity function, since this will need to be referenced in the `bufmodulecache` in order to check for the case where a module without a name is being put into the cache. This _should_ never be the case in practice/main code flows, since when we put a module into the cache, we use the module pin, but we need this check insulates against modules that were built without a module identity (e.g. test cases) or if any of the flows built directly from the bucket.
- `SetModuleIdentity(moduleIdentity bufmoduleref.ModuleIdentity)`: This allows us to set the module identity on the module if the module identity is `nil` when checked in the `bufmodulecache`.

In the current case, we are already adding the module identity to modules that we `Get` from the cache, so it is just a matter of ensuring that modules we `Put` in the cache has the same behaviour.

With these changes, we are able to use the `b3` digest for the cache. The push service is still using the `b1` digest, so the changes are primarily centered around the cache.

As a test case, an old version of the CLI was used to test the push and then a new version of the CLI was used to test the push, and we were able to successfully resolve the fact that the commits were the same, so no push (through the `b1` digest) but the cache was not used.


private/bufpkg/bufmodule/bufmodulecache/module_cacher.go
```diff
@@ -120,7 +120,10 @@ func (m *moduleCacher) PutModule(
 	module bufmodule.Module,
 ) error {
 	modulePath := newCacheKey(modulePin)
-	digest, err := bufmodule.ModuleDigestB2(ctx, module)
+	if module.ModuleIdentity() == nil {
+		module.SetModuleIdentity(modulePin)
+	}
```
amckinney, line 125: Can we ensure that the `Module` passed into this function already has its `ModuleIdentity` set? I noticed that this is only called after we resolve the module from the delegate `ModuleReader`, i.e.

```go
module, err = m.delegate.GetModule(ctx, modulePin)
if err != nil {
  return nil, err
}
if err := m.cache.PutModule(
  ctx,
  modulePin,
  module,
); err != nil {
  return nil, err
}
```

The `delegate` you see above should actually be backed by the following implementation (which always sets the `ModuleIdentity`:

```go
func (m *moduleReader) GetModule(ctx context.Context, modulePin bufmoduleref.ModulePin) (bufmodule.Module, error) {
	downloadService, err := m.downloadServiceProvider.NewDownloadService(ctx, modulePin.Remote())
	if err != nil {
		return nil, err
	}
	module, err := downloadService.Download(
		ctx,
		modulePin.Owner(),
		modulePin.Repository(),
		modulePin.Commit(),
	)
	if err != nil {
		if rpc.GetErrorCode(err) == rpc.ErrorCodeNotFound {
			// Required by ModuleReader interface spec
			return nil, storage.NewErrNotExist(modulePin.String())
		}
		return nil, err
	}
	moduleIdentity, err := bufmoduleref.NewModuleIdentity(
		modulePin.Remote(),
		modulePin.Owner(),
		modulePin.Repository(),
	)
	if err != nil {
		return nil, err
	}
	return bufmodule.NewModuleForProto(
		ctx, module,
		bufmodule.ModuleWithModuleIdentityAndCommit(moduleIdentity, modulePin.Commit()),
	)
}
```

So it looks like this might be relevant for tests - is there something we can do in each of those test cases to prevent ourselves from supporting the `SetModuleIdentity` method altogether?

private/bufpkg/bufmodule/module.go
```diff
@@ -173,14 +190,26 @@ func (m *module) Documentation() string {
...
+func (m *module) BreakingConfig() *bufbreakingconfig.Config {
+	return m.breakingConfig
+}
+
+func (m *module) LintConfig() *buflintconfig.Config {
+	return m.lintConfig
 }
 
-func (m *module) getModuleIdentity() bufmoduleref.ModuleIdentity {
+func (m *module) ModuleIdentity() bufmoduleref.ModuleIdentity {
```
amckinney, line 201: From what I can tell, all of the call-sites to this method happen within the same `bufmodule` package. Can this remain un-exported?

I know you mentioned exported this in your summary, but perhaps that's not relevant anymore?

doriable: Hmm, that's a good point, let me see if there is some way to force the module identity in the test cases :)

private/bufpkg/bufmodule/bufmodulecache/module_cacher.go
doriable: Hmm, this is a good point... let me take a look to see if I can refactor the test such that the module identity is properly constructed in the test cases.

private/bufpkg/bufmodule/module.go
doriable: Given the point above, if we ensure that the test cases are properly covered, we won't have to check outside of the `bufmodule` package (e.g. no longer need to check in the cache code), so we should be fine :)

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -88,3 +119,71 @@ type ExternalConfigV1 struct {
...
+		sort.Strings(rootPaths)
+		ignoreIDPathsJSON = append(ignoreIDPathsJSON, idPathsJSON{
+			ID:    ignoreID,
+			Paths: rootPaths,
+		})
+	}
+	sort.Slice(ignoreIDPathsJSON, func(i, j int) bool { return ignoreIDPathsJSON[i].ID < ignoreIDPathsJSON[j].ID })
+	sort.Strings(config.Use)
+	sort.Strings(config.Except)
+	sort.Strings(config.IgnoreRootPaths)
```
amckinney, line 141: nit/optional: Consider putting all of this into an un-exported `configToJSON` helper function that could be called here before we call `json.Marshal`. The same applies in the `buflintconfig` package below.

amckinney (approved):
Nice, this looks great! Let's make sure that we test thoroughly test this through the BSR side before we merge.

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -88,3 +119,75 @@ type ExternalConfigV1 struct {
 	IgnoreOnly             map[string][]string `json:"ignore_only,omitempty" yaml:"ignore_only,omitempty"`
 	IgnoreUnstablePackages bool                `json:"ignore_unstable_packages,omitempty" yaml:"ignore_unstable_packages,omitempty"`
 }
+
+// BytesForConfig takes a *Config and returns the deterministic []byte representation.
+// We use an unexported intermediary JSON form and sort all fields to ensure that the bytes
+// associated with the *Config are deterministic.
+func BytesForConfig(config *Config) ([]byte, error) {
```
bufdev, line 126: This should be done by implementing `json.Marshaler` on `Config`

private/bufpkg/bufcheck/buflint/buflintconfig/buflintconfig.go
```diff
@@ -127,6 +163,63 @@ type ExternalConfigV1 struct {
 	AllowCommentIgnores                  bool                `json:"allow_comment_ignores,omitempty" yaml:"allow_comment_ignores,omitempty"`
 }
 
+// BytesForConfig takes a *Config and returns the deterministic []byte representation.
+// We use an unexported intermediary JSON form and sort all fields to ensure that the bytes
+// associated with the *Config are deterministic.
+func BytesForConfig(config *Config) ([]byte, error) {
```
bufdev, line 169: `json.Marshaler`

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -93,6 +95,14 @@ type Module interface {
 	// Documentation gets the contents of the module documentation file, buf.md and returns the string representation.
 	// This may return an empty string if the documentation file does not exist.
 	Documentation() string
+	// BreakingConfig returns the breaking change check configuration set for the module.
+	//
+	// This may be nil, since older versions of the module would not have this stored.
```
bufdev, line 100: At the least, this should be documented to have a consistent version with `LintConfig`

private/bufpkg/bufmodule/module.go
```diff
@@ -105,6 +120,8 @@ func newModule(
 		sourceReadBucket:     sourceReadBucket,
 		dependencyModulePins: dependencyModulePins,
 		documentation:        documentation,
+		breakingConfig:       breakingConfig,
```
bufdev, line 123: Validation of versions should be done in this constructor

bufdev: Versions should also never be empty

PR #750, opened 2021-12-01
add uuidutil FromStringSlice

smallsamantha:


private/pkg/uuidutil/uuidutil.go
```diff
@@ -78,3 +78,21 @@ func ValidateDashless(dashless string) error {
...
+func FromStringSlice(s []string) ([]uuid.UUID, error) {
+	var parsedUUIDS []uuid.UUID
+	for _, stringID := range s {
+		if len(stringID) != 36 {
+			return nil, fmt.Errorf("expected uuid to be of length 36 but was %d: %s", len(s), s)
+		}
+		parsed, err := uuid.FromString(stringID)
+		if err != nil {
+			return nil, err
+		}
```
johanbrandhorst, line 94: Should we not just use `FromString` here?

smallsamantha: yes i was testing u 🙃

private/pkg/uuidutil/uuidutil_test.go
```diff
@@ -64,3 +64,28 @@ func TestValidateDashlessFailsWithUUID(t *testing.T) {
...
+
+func TestFromStringSlice(t *testing.T) {
+	id1, err := New()
+	require.NoError(t, err)
+	id2, err := New()
+	require.NoError(t, err)
+	ids := []string{id1.String(), id2.String()}
+	uuids, err := FromStringSlice(ids)
+	require.NoError(t, err)
+	require.Equal(t, 2, len(uuids))
```
johanbrandhorst, line 90: Do we not want to test that this is also equivalent to `[]uuid.UUID{id1, id2}`?

johanbrandhorst (approved):

PR #751, opened 2021-12-01
update GetRepositoryCommitBySequenceID rpc name

smallsamantha:
updating GetRepositoryCommitBySequenceID to GetRepositoryCommitBySequenceId for autogeneration purposes


johanbrandhorst (approved):
This is a breaking change, but the only user of this API is `bufgoproxyd`, so it should be fine.

PR #754, opened 2021-12-02
registry login: suppress bug report message on EOF

lrewega:
A user entering EOF (^D) when prompted does not warrant filing a bug
report.

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -19,6 +19,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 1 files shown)


private/buf/bufcli/bufcli.go
```diff
@@ -684,13 +685,20 @@ func promptUser(container app.Container, prompt string, isPassword bool) (string
...
+					return "", err
+				}
 				return "", NewInternalError(err)
 			}
 			value = string(data)
 		} else {
 			scanner := bufio.NewScanner(container.Stdin())
 			if !scanner.Scan() {
-				return "", NewInternalError(scanner.Err())
+				// scanner.Err() returns nil on EOF.
```
bufdev, line 697: Does it *only* return nil on `io.EOF`?

private/buf/bufcli/bufcli.go
```diff
@@ -684,13 +685,20 @@ func promptUser(container app.Container, prompt string, isPassword bool) (string
 		if isPassword {
 			data, err := term.ReadPassword(int(file.Fd()))
 			if err != nil {
+				if errors.Is(err, io.EOF) {
```
bufdev, line 688: comment as to why would be helpful

private/buf/bufcli/bufcli.go
lrewega: Good question!

Yes, I believe nil is only possible on `io.EOF` or if the scanner's split func returns `bufio.ErrFinalToken`, which the default split func (`bufio.ScanLines`, which we're using here) never returns.

From https://pkg.go.dev/bufio#Scanner.Scan
> After Scan returns false, the Err method will return any error that occurred during scanning, except that if it was io.EOF, Err will return nil.

bufdev (approved):

PR #755, opened 2021-12-02
registry login: handle context cancel (SIGINT)

lrewega:
If a user sends a SIGINT to `buf`, the top-level application context is
cancelled and signal masks are reset. However, during an interactive
login the context is not respected; for example, it takes _two_ SIGINTs
to interrupt the process.

Ideally we could just trigger an I/O timeout by setting the deadline on
stdin, but when stdin is connected to a terminal the underlying fd is in
blocking mode making it ineligible. As changing the mode of stdin is
dangerous, this change takes an alternate approach of simply returning
early.

Note that this does not gracefully handle the case where the terminal is
in no-echo mode, as is the case when prompting for a password
interactively.

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -86,13 +86,25 @@ func run(
 	container appflag.Container,
 	flags *flags,
 ) error {
-	errc := make(chan error)
+	// If a user sends a SIGINT to buf, the top-level application context is
+	// cancelled and signal masks are reset. However, during an interactive
+	// login the context is not respected; for example, it takes two SIGINTs
+	// to interrupt the process.
 
+	// Ideally we could just trigger an I/O timeout by setting the deadline on
+	// stdin, but when stdin is connected to a terminal the underlying fd is in
+	// blocking mode making it ineligible. As changing the mode of stdin is
+	// dangerous, this change takes an alternate approach of simply returning
+	// early.
+
+	// Note that this does not gracefully handle the case where the terminal is
+	// in no-echo mode, as is the case when prompting for a password
+	// interactively.
+	errc := make(chan error, 1)
 	go func() {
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -85,6 +85,25 @@ func run(
 	ctx context.Context,
 	container appflag.Container,
 	flags *flags,
+) error {
+	errc := make(chan error)
+	defer close(errc)
```
lrewega, line 90: Ah this is problematic as the goroutine could still try to write.

lrewega:
```suggestion
```

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -85,6 +85,25 @@ func run(
...
+
+	go func() {
+		errc <- inner(container, flags)
```
lrewega, line 93:
```suggestion
		errc <- inner(container, flags)
		close(errc)
```

lrewega:
This is still valid, I'll mark it for review.

bufdev:
@lrewega let's sync up about this on Monday

bufdev (approved):
I think this is fine

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -85,6 +85,25 @@ func run(
 	ctx context.Context,
 	container appflag.Container,
 	flags *flags,
+) error {
+	errc := make(chan error)
```
bufdev, line 89: Should this be of size 1?

bufdev:
@lrewega added some comments and did some cleanup, I think this is good to go now

lrewega:
Thanks @bufdev!

PR #756, opened 2021-12-02
plugin config inaccessible

cyinma:
This is to add a bool for `inaccessible` plugin config, and this should replace the use of `deleted` later.


proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -85,6 +85,9 @@ message PluginConfig {
   // True if the source plugin has been deleted. This dooms
   // the template.
   bool deleted = 4;
+  // True if the source plugin is inaccessible by the user.
+  // This dooms the template.
```
robbertvanginkel, line 89:
```suggestion
```
I think this wording is a little weird as the template itself isn't doomed, its just the user who won't be able to use it (but can still look at it). If we ever allow people to switch out plugins (to for example a similar template of their own with the same functionality), this would also not be very forward looking.

cyinma: yes you are right 👍

robbertvanginkel (approved):

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -85,6 +85,8 @@ message PluginConfig {
   // True if the source plugin has been deleted. This dooms
   // the template.
   bool deleted = 4;
```
johanbrandhorst, line 87: Do we want to mark this as `deprecated`?

cyinma: I guess we should do this after we implemented the `inaccessible`? but by that time I think we can just remove this field

johanbrandhorst: I think we can mark it as deprecated here, and then remove it later.
```suggestion
  bool deleted = 4 [deprecated=true];
```

timostamm: Marking as deprecated would bring some certainty when implementing the frontend change, and when coming back here and deleting the field. Certainly not strictly necessary for this small change (and maybe technically missing a step) but good practice for coming larger changes, in my mind.

timostamm (approved):

PR #759, opened 2021-12-03
Add enums for new RPCs

johanbrandhorst:
Missed when adding the RPCs. None of these need extra details included.


smallsamantha (approved):
🚀

PR #766, opened 2021-12-07
Refactor the `WriteConfig` function for bufconfig for a generic use case

doriable:
Add `WriteConfigWithBreakingConfig`, `WriteConfigWithLintConfig`, and `WriteConfigWithVersion` options for the `bufconfig.WriteConfig` function to support more generic use cases of the `WriteConfig`.
Before this change, we are only using `bufconfig.WriteConfig` in the `buf config init` command, but we now need it to write configuration structures into our module cache so that we are persisting `breaking` and `lint` config in our cache for our module structure.

### DIFF
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -1159,11 +1159,11 @@ func TestConfigInitBasic(t *testing.T) {
 		t,
 		`version: v1
 breaking:
-  use:
-  - FILE
+    use:
+        - FILE
 lint:
-  use:
-  - DEFAULT
+    use:
+        - DEFAULT
 `,
 		false,
 		"",
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -1158,12 +1158,12 @@ func TestConfigInitBasic(t *testing.T) {
 	testConfigInit(
 		t,
 		`version: v1
-lint:
-  use:
-    - DEFAULT
 breaking:
   use:
-    - FILE
+  - FILE
+lint:
+  use:
+  - DEFAULT
 `,
 		false,
 		"",
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -1159,11 +1159,11 @@ func TestConfigInitBasic(t *testing.T) {
 		t,
 		`version: v1
 breaking:
-    use:
-        - FILE
+  use:
+    - FILE
 lint:
-    use:
-        - DEFAULT
+  use:
+    - DEFAULT
 `,
 		false,
 		"",
--- a/private/buf/cmd/buf/command/config/configinit/configinit.go
+++ b/private/buf/cmd/buf/command/config/configinit/configinit.go
@@ -19,6 +19,8 @@ import (
 	"fmt"
 
 	"github.com/bufbuild/buf/private/buf/bufcli"
+	"github.com/bufbuild/buf/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig"
+	"github.com/bufbuild/buf/private/bufpkg/bufcheck/buflint/buflintconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
@@ -126,6 +128,23 @@ func run(
 			bufconfig.WriteConfigWithUncomment(),
 		)
 	}
+	// Need to include the default version (v1), lint config, and breaking config.
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithVersion(bufconfig.V1Version),
+	)
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithBreakingConfig(&bufbreakingconfig.Config{
+			Use: []string{"FILE"},
+		}),
+	)
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithLintConfig(&buflintconfig.Config{
+			Use: []string{"DEFAULT"},
+		}),
+	)
 	return bufconfig.WriteConfig(
 		ctx,
 		readWriteBucket,
--- a/private/buf/cmd/buf/command/config/configinit/configinit.go
+++ b/private/buf/cmd/buf/command/config/configinit/configinit.go
@@ -129,20 +129,23 @@ func run(
 		)
 	}
 	// Need to include the default version (v1), lint config, and breaking config.
+	version := bufconfig.V1Version
 	writeConfigOptions = append(
 		writeConfigOptions,
-		bufconfig.WriteConfigWithVersion(bufconfig.V1Version),
+		bufconfig.WriteConfigWithVersion(version),
 	)
 	writeConfigOptions = append(
 		writeConfigOptions,
 		bufconfig.WriteConfigWithBreakingConfig(&bufbreakingconfig.Config{
-			Use: []string{"FILE"},
+			Version: version,
+			Use:     []string{"FILE"},
 		}),
 	)
 	writeConfigOptions = append(
 		writeConfigOptions,
 		bufconfig.WriteConfigWithLintConfig(&buflintconfig.Config{
-			Use: []string{"DEFAULT"},
+			Version: version,
+			Use:     []string{"DEFAULT"},
 		}),
 	)
 	return bufconfig.WriteConfig(
--- a/private/bufpkg/bufconfig/bufconfig.go
+++ b/private/bufpkg/bufconfig/bufconfig.go
@@ -143,6 +143,35 @@ func WriteConfigWithDependencyModuleReferences(dependencyModuleReferences ...buf
 	}
 }
 
+// WriteConfigWithBreakingConfig returns a new WriteConfigOption that sets the breaking change
+// config for the module.
+//
+// If this option is used and the version used will be consistent with the rest of the config.
+func WriteConfigWithBreakingConfig(breakingConfig *bufbreakingconfig.Config) WriteConfigOption {
+	return func(writeConfigOptions *writeConfigOptions) {
+		writeConfigOptions.breakingConfig = breakingConfig
+	}
+}
+
+// WriteConfigWithLintConfig returns a new WriteConfigOption that sets the lint config for // the module.
+//
+// If this option is used and the version used will be consistent with the rest of the config.
+func WriteConfigWithLintConfig(lintConfig *buflintconfig.Config) WriteConfigOption {
+	return func(writeConfigOptions *writeConfigOptions) {
+		writeConfigOptions.lintConfig = lintConfig
+	}
+}
+
+// WriteConfigWithVersion returns a new WriteConfigOption that sets the version of the config
+// being written.
+//
+// If this is not set, the default is v1.
+func WriteConfigWithVersion(version string) WriteConfigOption {
+	return func(writeConfigOptions *writeConfigOptions) {
+		writeConfigOptions.version = version
+	}
+}
+
 // ReadConfigOS reads the configuration from the OS or an override, if any.
 //
 // ONLY USE IN CLI TOOLS.
--- a/private/bufpkg/bufconfig/bufconfig.go
+++ b/private/bufpkg/bufconfig/bufconfig.go
@@ -165,7 +165,7 @@ func WriteConfigWithLintConfig(lintConfig *buflintconfig.Config) WriteConfigOpti
 // WriteConfigWithVersion returns a new WriteConfigOption that sets the version of the config
 // being written.
 //
-// If this is not set, the default is v1.
+// If this is not set, the default is v1beta1.
 func WriteConfigWithVersion(version string) WriteConfigOption {
 	return func(writeConfigOptions *writeConfigOptions) {
 		writeConfigOptions.version = version
--- a/private/bufpkg/bufconfig/bufconfig.go
+++ b/private/bufpkg/bufconfig/bufconfig.go
@@ -146,16 +146,16 @@ func WriteConfigWithDependencyModuleReferences(dependencyModuleReferences ...buf
 // WriteConfigWithBreakingConfig returns a new WriteConfigOption that sets the breaking change
 // config for the module.
 //
-// If this option is used and the version used will be consistent with the rest of the config.
+// If this option is used then the version used must be consistent with the rest of the config.
 func WriteConfigWithBreakingConfig(breakingConfig *bufbreakingconfig.Config) WriteConfigOption {
 	return func(writeConfigOptions *writeConfigOptions) {
 		writeConfigOptions.breakingConfig = breakingConfig
 	}
 }
 
-// WriteConfigWithLintConfig returns a new WriteConfigOption that sets the lint config for // the module.
+// WriteConfigWithLintConfig returns a new WriteConfigOption that sets the lint config for the module.
 //
-// If this option is used and the version used will be consistent with the rest of the config.
+// If this option is used then the version used must be consistent with the rest of the config.
 func WriteConfigWithLintConfig(lintConfig *buflintconfig.Config) WriteConfigOption {
 	return func(writeConfigOptions *writeConfigOptions) {
 		writeConfigOptions.lintConfig = lintConfig
--- a/private/bufpkg/bufconfig/write.go
+++ b/private/bufpkg/bufconfig/write.go
@@ -18,27 +18,21 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"text/template"
 
+	"github.com/bufbuild/buf/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig"
+	"github.com/bufbuild/buf/private/bufpkg/bufcheck/buflint/buflintconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
 	"github.com/bufbuild/buf/private/pkg/storage"
+	"gopkg.in/yaml.v2"
 )
 
 // If this is updated, make sure to update docs.buf.build TODO automate this
 
 const (
-	exampleName          = "buf.build/acme/weather"
-	tmplUndocumentedData = `{{$top := .}}version: v1
-{{if not .NameUnset}}name: {{.Name}}
-{{end}}{{if not .DepsUnset}}deps:
-{{range $dep := .Deps}}  - {{$dep}}
-{{end}}{{end}}lint:
-  use:
-{{range $lint_id := .LintIDs}}    - {{$lint_id}}
-{{end}}breaking:
-  use:
-{{range $breaking_id := .BreakingIDs}}    - {{$breaking_id}}
-{{end}}`
+	exampleName = "buf.build/acme/weather"
+	// This is only used for `buf config init`.
 	tmplDocumentationCommentsData = `{{$top := .}}# This specifies the configuration file version.
 #
 # This controls the configuration file layout, defaults, and lint/breaking
@@ -47,7 +41,7 @@ const (
 #
 # The only valid versions are "v1beta1", "v1".
 # This key is required.
-version: v1
+version: {{.Version}}
 
 # name is the module name.
 {{if .NameUnset}}#{{end}}name: {{.Name}}
@@ -89,7 +83,7 @@ lint:
   #
   # The default is [DEFAULT].
   use:
-{{range $lint_id := .LintIDs}}    - {{$lint_id}}
+{{range $lint_id := .LintConfig.Use}}    - {{$lint_id}}
 {{end}}
   # except is the list of rule ids or categories to remove from the use
   # list.
@@ -202,7 +196,7 @@ breaking:
   #
   # The default is [FILE], as done below.
   use:
-{{range $breaking_id := .BreakingIDs}}    - {{$breaking_id}}
+{{range $breaking_id := .BreakingConfig.Use}}    - {{$breaking_id}}
 {{end}}
   # except is the list of rule ids or categories to remove from the use
   # list.
@@ -254,16 +248,6 @@ breaking:
   {{if not .Uncomment}}#{{end}}ignore_unstable_packages: false`
 )
 
-var (
-	defaultLintIDs     = []string{"DEFAULT"}
-	defaultBreakingIDs = []string{"FILE"}
-	exampleDeps        = []string{
-		"buf.build/acme/petapis",
-		"buf.build/acme/pkg:alpha",
-		"buf.build/acme/paymentapis:7e8b594e68324329a7aefc6e750d18b9",
-	}
-)
-
 func writeConfig(
 	ctx context.Context,
 	writeBucket storage.WriteBucket,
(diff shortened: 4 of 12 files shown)


private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,162 @@ func writeConfig(
...
-	if writeConfigOptions.moduleIdentity == nil && len(writeConfigOptions.dependencyModuleReferences) > 0 {
-		return errors.New("cannot set deps without a name for WriteConfig")
+	// This is the same default as the bufconfig getters.
+	version := V1Beta1Version
+	if writeConfigOptions.version != "" {
+		if err := validateVersion(writeConfigOptions.version); err != nil {
+			return err
+		}
+		version = writeConfigOptions.version
 	}
```
amckinney, line 270: Is this right? The default version is `v1` when we run `buf config init`, and the `WriteConfigWithVersion` option says `v1` is still the default, but this looks like `v1beta1` is.

I've tested that `buf config init` continues to write a configuration with `v1` on this branch, so I guess it's working as expected, but I wonder if it's just because we're always providing the options at each call-site.

private/buf/cmd/buf/buf_test.go
```diff
@@ -1158,12 +1158,12 @@ func TestConfigInitBasic(t *testing.T) {
...
-  use:
-    - DEFAULT
 breaking:
-  use:
-    - FILE
+    use:
+        - FILE
+lint:
+    use:
+        - DEFAULT
```
amckinney, line 1166: Gah yea this isn't great, especially because it effects how users interact with `buf config init`. Unless there's a `yaml.Marshal` option we can take advantage of to control indentation, this might require that the `buf config init` path takes on a different path than the configuration written to the module cache.

Unfortunately, the aesthetic of the configuration we write with `buf config init` matters here.

Check out https://pkg.go.dev/gopkg.in/yaml.v3#Encoder.SetIndent and see if we can get that to work.

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,162 @@ func writeConfig(
...
-	for _, dependencyModuleReference := range writeConfigOptions.dependencyModuleReferences {
-		externalConfigV1.Deps = append(
-			externalConfigV1.Deps,
-			dependencyModuleReference.String(),
-		)
+	if writeConfigOptions.lintConfig != nil {
+		config.Lint = writeConfigOptions.lintConfig
+		// Always force the provided version.
+		config.Lint.Version = version
```
amckinney, line 283: We should validate that the provided `lintConfig.Version` is equivalent to the `version` (same with `breakingConfig` above).

Forcing the override might lead to invalid results if the configured rules changed between `v1beta` and `v1`.

private/bufpkg/bufconfig/bufconfig.go
```diff
@@ -143,6 +143,35 @@ func WriteConfigWithDependencyModuleReferences(dependencyModuleReferences ...buf
 	}
 }
 
+// WriteConfigWithBreakingConfig returns a new WriteConfigOption that sets the breaking change
+// config for the module.
+//
+// If this option is used and the version used will be consistent with the rest of the config.
```
amckinney, line 149: nit: If we update the behavior to validate, let's update the wording to match it.
```suggestion
// If this option is used then the version used must be consistent with the rest of the config.
```

private/bufpkg/bufconfig/bufconfig.go
```diff
@@ -143,6 +143,35 @@ func WriteConfigWithDependencyModuleReferences(dependencyModuleReferences ...buf
...
+func WriteConfigWithBreakingConfig(breakingConfig *bufbreakingconfig.Config) WriteConfigOption {
+	return func(writeConfigOptions *writeConfigOptions) {
+		writeConfigOptions.breakingConfig = breakingConfig
+	}
+}
+
+// WriteConfigWithLintConfig returns a new WriteConfigOption that sets the lint config for // the module.
+//
+// If this option is used and the version used will be consistent with the rest of the config.
```
amckinney, line 158: nit: Similar to above
```suggestion
// If this option is used then the version used must be consistent with the rest of the config.
```

private/bufpkg/bufconfig/bufconfig.go
```diff
@@ -143,6 +143,35 @@ func WriteConfigWithDependencyModuleReferences(dependencyModuleReferences ...buf
...
+func WriteConfigWithLintConfig(lintConfig *buflintconfig.Config) WriteConfigOption {
+	return func(writeConfigOptions *writeConfigOptions) {
+		writeConfigOptions.lintConfig = lintConfig
+	}
+}
+
+// WriteConfigWithVersion returns a new WriteConfigOption that sets the version of the config
+// being written.
+//
+// If this is not set, the default is v1.
```
amckinney, line 168: Depending on the conversation below, the default might be `v1beta1` to preserve the default value of `buf.yaml`, but I'm not totally sure yet if this option will match that behavior from a `WriteConfig` standpoint.

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,162 @@ func writeConfig(
...
+	config *Config,
+	dependencies []string,
+) error {
+	var name string
+	if config.ModuleIdentity != nil {
+		name = config.ModuleIdentity.IdentityString()
+	}
+	breakingConfig := config.Breaking
+	lintConfig := config.Lint
+	if version == V1Beta1Version {
```
amckinney, line 313: We should still write this so that the `version` must be either `v1beta1` or `v1` (instead of defaulting to `v1` when version is not `v1beta1`). I know we already validate it via `validateVersion` earlier, but this implicit default behavior is a little less obvious.

For consistency, we can adopt the `switch` approach that you see in the other configuration functions, like [getConfigForDataInternal](https://github.com/bufbuild/buf/blob/e4a9b7b053022398f70922a9d7241b822154fe5d/private/bufpkg/bufconfig/get.go#L88).

Similarly, we could write separate helper functions (e.g. `externalConfig{V1,V1Beta1}ForConfig`) so that the implementation reads similar to the `case` found in `getConfigForDataInternal`:

```go
case V1Beta1Version:
  var externalConfigV1Beta1 ExternalConfigV1Beta1
  if err := unmarshalStrict(data, &externalConfigV1Beta1); err != nil {
    return nil, err
  }
  return newConfigV1Beta1(externalConfigV1Beta1)
```

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -472,7 +473,37 @@ func ModuleToBucket(
...
+	if breakingConfigVersion != "" && lintConfigVersion != "" {
+		if breakingConfigVersion != lintConfigVersion {
+			return fmt.Errorf("breaking config version %q does not match lint config version %q", breakingConfigVersion, lintConfigVersion)
+		}
+		version = breakingConfigVersion
+	} else if breakingConfigVersion != "" {
+		version = breakingConfigVersion
+	} else if lintConfigVersion != "" {
+		version = lintConfigVersion
+	}
```
amckinney, line 499: We should return an error in these cases. We should force the caller of this function to provide consistent versions for the module's `LintConfig` and `BreakingConfig`, otherwise it's an internal error on our end since that should always be in our control.

We should also be able to guarantee that the `{Breaking,Lint}Config` version is _always_ non-empty if the `{Breaking,Lint}Config` is non-`nil`.

So in short:
* A version _must_ be set on both of the `LintConfig` and `BreakingConfig` types if they are non-`nil`.
* The version _must_ be equivalent between them if they are both non-`nil`.
* If one of the `{Breaking,Lint}Config` types is `nil`, the versions don't necessarily need to match, but this isn't actually possible in practice because there's always a default value for each one. If one of the configs is provided, both will be non-`nil`.

private/bufpkg/bufconfig/write.go
amckinney: I see now, it's probably right that we should default to `v1beta1` since that's consistent with the configuration getters - we just need to update the documentation on `WriteConfigWithVersion` to say that the default is `v1beta`.

private/bufpkg/bufconfig/write.go
doriable: Yeah -- that's fair. The validation happens in the module build also well, but maybe it makes sense to actually validate the versions here and then move the version inferences out of `bufmodule.go`?

private/buf/cmd/buf/buf_test.go
doriable: Nice -- this works, so I'm happy about this :D

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
+}
+
+func validateVersion(version string) error {
+	switch version {
+	case V1Version, V1Beta1Version:
+		return nil
+	default:
+		return fmt.Errorf("invalid config version %q provided", version)
+	}
+}
```
amckinney, line 375: nit: Move this function down so that it's adjacent to the other validation helper (at the bottom of the file),

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
-			dependencyModuleReference.String(),
-		)
+	var lintConfigVersion string
+	if writeConfigOptions.lintConfig != nil {
+		config.Lint = writeConfigOptions.lintConfig
+		lintConfigVersion = version
+	}
+	if breakingConfigVersion != lintConfigVersion {
+		return fmt.Errorf("breaking config version %q does not match lint config version %q", breakingConfigVersion, lintConfigVersion)
+	}
```
amckinney, line 287: I still don't know if this is right.

We should validate that both `breakingConfigVersion` and `lintConfigVersion` are non-empty and set to the exact same value as the `version` provided by `writeConfigOptions.version`, if any.

Otherwise you'll end up in that situation where the `writeConfigOptions.version` would incorrectly override the `{Breaking,Lint}`Config version as discussed in [this](https://github.com/bufbuild/buf/pull/766#discussion_r764439767) comment.

private/bufpkg/bufmodule/module.go
```diff
@@ -89,6 +89,10 @@ func newModuleForBucket(
 	if err != nil {
 		return nil, err
 	}
+	// if the module config has an identity, set the module identity
+	if moduleConfig.ModuleIdentity != nil {
+		options = append(options, ModuleWithModuleIdentity(moduleConfig.ModuleIdentity))
+	}
```
amckinney, line 95: This will override the `ModuleWithModuleIdentity[AndCommit]` option provided by the caller of `NewModuleForBucket`, if one was provided:

```go
// NewModuleForBucket returns a new Module. It attempts reads dependencies
// from a lock file in the read bucket.
func NewModuleForBucket(
	ctx context.Context,
	readBucket storage.ReadBucket,
	options ...ModuleOption,
) (Module, error) {
	return newModuleForBucket(ctx, readBucket, options...)
}
```

In practice, this might be OK since the `ModuleIdentity` found in the bucket _should_ always be equivalent to the option provided (i.e. the option is set with the module pin in the module cacher).

However, this is prone to error - if there's ever a reason that the caller of `NewModuleForBucket` would need to _override_ the `ModuleIdentity` that would otherwise be determined by the `buf.yaml` (this condition), then the caller's option will actually be overwritten with what's found in the bucket.

--

One solution out of this is to _prepend_ this option, which allows the other call-site provided option to take effect _after_ this one, hence taking precedence. This is also nuanced behavior though and feels kinda hacky, so we would need to aggressively include comments here if that's our only option. For example,

```go
if moduleConfig.ModuleIdentity != nil {
	// Explain why we're doing this.
	options = append([]ModuleOption{ModuleWithModuleIdentity(moduleConfig.ModuleIdentity)}, options...)
}
```

Alternatively, we could add an argument to the `newModule` constructor below so that it takes a positional `bufmodule.ModuleIdentity` parameter (which may be `nil`). This would be the default value, but it could be naturally overwritten by the option.

```go
// this should only be called by other newModule constructors
func newModule(
	ctx context.Context,
	// must only contain .proto files
	sourceReadBucket storage.ReadBucket,
	moduleIdentity ModuleIdentity,
	dependencyModulePins []bufmoduleref.ModulePin,
	documentation string,
	breakingConfig *bufbreakingconfig.Config,
	lintConfig *buflintconfig.Config,
	options ...ModuleOption,
) (_ *module, retErr error) {
	if err := bufmoduleref.ValidateModulePinsUniqueByIdentity(dependencyModulePins); err != nil {
		return nil, err
	}
	// we rely on this being sorted here
	bufmoduleref.SortModulePins(dependencyModulePins)
	module := &module{
		sourceReadBucket:     sourceReadBucket,
		moduleIdentity:       moduleIdentity,
		dependencyModulePins: dependencyModulePins,
		documentation:        documentation,
		breakingConfig:       breakingConfig,
		lintConfig:           lintConfig,
	}
	for _, option := range options {
		option(module)
	}
	return module, nil
}
```

Both aren't ideal, but I think I'd lean towards the latter option (updating the `newModule` signature) since it's a lot easier to follow. It will require passing in a `nil` argument in the `newModule` referenced in `newModuleForProto` which is a code smell, but we don't have the `ModuleName` in the Protobuf representation of `Module` - it's explicitly passed in via the [ModuleReader](https://github.com/bufbuild/buf/blob/0786cea8690ed81cdd1f5b31b9b5c0529ffdab74/private/bufpkg/bufapimodule/module_reader.go#L67). If we go this route, we should add a comment in-line explaining this at that call-site.

What are your thoughts?

private/bufpkg/bufconfig/write.go
doriable: Ah yes... hmm... I thought I changed this -- maybe I only changed it in the `bufmodule.go`. Let me fix this, my bad!

private/bufpkg/bufmodule/module.go
doriable: After some offline discussion (and some flipflopping, lol), the option we're going to go with is the second, adding a new parameter to `newModule`.

The crux of this discussion is around the intended behaviour for the build. The original intention was that any `ModuleIdentity` set by the call site _should not_ deviate from the `ModuleIdentity` from the config getter. And in practice, this is the case, since there could be a lot of side effects and breakages caused by overriding with a malformed module identity (including to the digest, etc.). That being said, there could be some use cases, such as testing, etc. where the module identity will need to be overridden, and thus we should always respect the value provided by the call site.

Thus, by adding a new parameter, we can explicitly set the module identity when building via the bucket (and keep it `nil` via other build paths), and then still respect the option override.

private/bufpkg/bufmodule/module.go
```diff
@@ -65,6 +65,7 @@ func newModuleForProto(
 		ctx,
 		readWriteBucket,
 		dependencyModulePins,
+		nil,
```
amckinney, line 68: Let's include that comment here - that the Module Identity isn't part of the Protobuf definition, but it's expected to be provided by the layer above (e.g. the `ModuleReader`).

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,170 @@ func writeConfig(
...
+			return fmt.Errorf("invalid version found for breaking config: %q", breakingConfig.Version)
+		}
+		config.Breaking = breakingConfig
 	}
-	for _, dependencyModuleReference := range writeConfigOptions.dependencyModuleReferences {
-		externalConfigV1.Deps = append(
-			externalConfigV1.Deps,
-			dependencyModuleReference.String(),
-		)
+	var lintConfigVersion string
```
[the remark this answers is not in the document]
doriable: 🤦🏼‍♀️

private/bufpkg/bufconfig/write_test.go
```diff
@@ -0,0 +1,96 @@
...
+		writeConfigOptions := []WriteConfigOption{
+			WriteConfigWithVersion(V1Beta1Version),
+			WriteConfigWithLintConfig(&buflintconfig.Config{
+				Version: V1Version,
+				Use:     []string{"DEFAULT"},
+			}),
+		}
+		err = WriteConfig(context.Background(), readWriteBucket, writeConfigOptions...)
+		require.Error(t, err)
+		require.Equal(t, err.Error(), `invalid version found for lint config: "v1"`)
```
amckinney, line 94: nit: we should say what we expect the version to be - otherwise the error message is confusing (`v1` is a valid version).

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
+			Deps:     dependencies,
+			Breaking: externalBreakingConfig,
+			Lint:     externalLintConfig,
+		}
+		var buffer bytes.Buffer
+		encoder := yaml.NewEncoder(&buffer)
+		encoder.SetIndent(2)
+		if err := encoder.Encode(externalConfig); err != nil {
+			return err
+		}
```
amckinney, line 347: Discussed offline - 

This will produce non-deterministic results because the YAML structure contains a `map[string][]string` nested within it.

Right now, this isn't actually user-facing - the only time we actually write these potentially non-deterministic values is in the users' local module cache. It's definitely not ideal, but we can follow-up with a fix after this PR is merged (i.e. using the `yaml.Node` API in `yaml.v3`).

What matters most is that the digest algorithm is deterministic and the `buf.yaml` can be successfully round-tripped without losing any information (just sometimes in a different order, such as with `ignore_only`).

We will need a fix in these representations before encoding the `buf.yaml` is formally complete.

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -269,6 +267,42 @@ func TestModuleReaderCacherWithDocumentation(t *testing.T) {
...
+	err = moduleCacher.PutModule(
+		context.Background(),
+		modulePin,
+		module,
+	)
+	require.NoError(t, err)
+	module, err = moduleCacher.GetModule(ctx, modulePin)
+	require.NoError(t, err)
+	readWriteBucket := storagemem.NewReadWriteBucket()
+	require.NoError(t, bufmodule.ModuleToBucket(ctx, module, readWriteBucket))
```
amckinney, line 299: This test is valuable, but is there anything else we can add to it with respect to verifying the configuration is round-trippable?

Can we verify that the module we put is exactly the same as the module we get out? Right now it looks like we're only doing so for the documentation below.

private/bufpkg/bufconfig/write_test.go
doriable: Ah yeah, that's a good call.

private/bufpkg/bufconfig/write.go
doriable: +1, thanks for comment :) I am currently working on summarizing the requirements for having the config be round-trippable in a stand-alone issue, does it makes sense to bundle that as a strict requirement for the encoding `buf.yaml` issue or should it be its own thing?

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
doriable: Yep, just added to the test case to check against the `bufmoduletesting.TestModuleConfiguration` config.

amckinney (approved):
Looks good!

private/buf/cmd/buf/command/config/configinit/configinit.go
```diff
@@ -126,6 +128,26 @@ func run(
...
 	}
+	// Need to include the default version (v1), lint config, and breaking config.
+	version := bufconfig.V1Version
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithVersion(version),
+	)
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithBreakingConfig(&bufbreakingconfig.Config{
```
bufdev, line 139: Newline after ( and after final }

private/buf/cmd/buf/command/config/configinit/configinit.go
```diff
@@ -126,6 +128,26 @@ func run(
...
+			Version: version,
+			Use:     []string{"FILE"},
+		}),
+	)
+	writeConfigOptions = append(
+		writeConfigOptions,
+		bufconfig.WriteConfigWithLintConfig(&buflintconfig.Config{
```
bufdev, line 146: Newline after ( and after final }

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
 	}
-	if !writeConfigOptions.documentationComments && writeConfigOptions.uncomment {
-		return errors.New("cannot set uncomment without documentationComments for WriteConfig")
+	if err := validateWriteConfigOptions(writeConfigOptions); err != nil {
+		return err
 	}
-	if writeConfigOptions.moduleIdentity == nil && len(writeConfigOptions.dependencyModuleReferences) > 0 {
-		return errors.New("cannot set deps without a name for WriteConfig")
+	// This is the same default as the bufconfig getters.
+	version := V1Beta1Version
```
bufdev, line 264: This seems wrong for the default value for write, are we sure about this?

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
-	externalConfigV1.Lint.Use = defaultLintIDs
-	externalConfigV1.Breaking.Use = defaultBreakingIDs
-	if writeConfigOptions.moduleIdentity != nil {
-		externalConfigV1.Name = writeConfigOptions.moduleIdentity.IdentityString()
+	var breakingConfigVersion string
+	breakingConfig := writeConfigOptions.breakingConfig
+	if breakingConfig != nil {
+		breakingConfigVersion = breakingConfig.Version
+		if breakingConfigVersion != version {
+			return fmt.Errorf("version %q found for breaking config, does not match top level config version: %q", breakingConfigVersion, version)
```
bufdev, line 280: This worries me - I'm not sure what to do about it but it seems like an API flaw that we can have mismatched versioms

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
+			externalLintConfig = buflintconfig.ExternalConfigV1Beta1ForConfig(lintConfig)
+		}
+		externalConfig := ExternalConfigV1Beta1{
+			Name:     name,
+			Version:  config.Version,
+			Deps:     dependencies,
+			Breaking: externalBreakingConfig,
+			Lint:     externalLintConfig,
+		}
+		var buffer bytes.Buffer
```
bufdev, line 342: We do `buffer := bytes.NewBuffer(nil)` in the rest of the codebase

private/bufpkg/bufconfig/write.go
```diff
@@ -273,30 +257,173 @@ func writeConfig(
...
+		name = config.ModuleIdentity.IdentityString()
+	}
+	if name == "" {
+		name = exampleName
+		nameUnset = true
 	}
+	var dependenciesUnset bool
+	if len(dependencies) == 0 {
+		dependencies = []string{
+			"buf.build/acme/petapis",
```
bufdev, line 401: Why were these removed from being consts?

private/bufpkg/bufconfig/write_test.go
```diff
@@ -0,0 +1,96 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestWriteConfigSuccess(t *testing.T) {
+	storageosProvider := storageos.NewProvider(storageos.ProviderWithSymlinks())
+	readWriteBucket, err := storageosProvider.NewReadWriteBucket(t.TempDir())
+	require.NoError(t, err)
+	writeConfigOptions := []WriteConfigOption{
+		WriteConfigWithVersion(V1Version),
+		WriteConfigWithBreakingConfig(&bufbreakingconfig.Config{
```
bufdev, line 34: newline

private/bufpkg/bufconfig/write_test.go
```diff
@@ -0,0 +1,96 @@
...
+			Version: V1Version,
+			Use:     []string{"FILE"},
+		}),
+		WriteConfigWithLintConfig(&buflintconfig.Config{
```
bufdev, line 38: newline

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -472,7 +473,36 @@ func ModuleToBucket(
 			return err
 		}
 	}
-	return bufmoduleref.PutDependencyModulePinsToBucket(ctx, writeBucket, module.DependencyModulePins())
+	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, writeBucket, module.DependencyModulePins()); err != nil {
+		return err
+	}
+	// This is the default version created by bufconfig getters. The versions should be the
+	// same across lint and breaking configs.
+	version := bufconfig.V1Beta1Version
```
bufdev, line 481: Same question as before

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -472,7 +473,36 @@ func ModuleToBucket(
...
+	var breakingConfigVersion string
+	if module.BreakingConfig() != nil {
+		breakingConfigVersion = module.BreakingConfig().Version
```
bufdev, line 484: Why are we exposing `Version` at all? This feels like an external concern, whereas this should be abstracted away for internal types.

private/bufpkg/bufconfig/write.go
doriable: Yeah, this should not be a code path that we ever hit and would be an API level problem, but I do think it's important to surface this error if we are attempting to write invalid config.

private/bufpkg/bufconfig/write.go
doriable: I moved them here since they are only being used for the template rendering, which is only used by `configinit` and I figured we should reduce the number globals. But I can move them back if that is preferred.

private/bufpkg/bufconfig/write.go
doriable: Yeah, this is consistent with the behaviour for when we go to get configs, if none are found, we default to `v1beta1` right now: https://github.com/bufbuild/buf/blob/f4fd40a47101848d722041ee6bfec433621bf1fe/private/bufpkg/bufconfig/get.go#L48-L50

There is a note to change this when we move to `v1`.

private/bufpkg/bufmodule/bufmodule.go
doriable: Addressed above.

private/bufpkg/bufmodule/bufmodule.go
doriable: This is based because the `Config` structure we store needs to reflect the external type for the digest. The config we store alongside the module needs to reflect changes in the configuration being sent by the user so it can be updated accordingly.

PR #768, opened 2021-12-08
Fix multiple request issue listing repositories

WillAbides:
This fixes an n+1 issue where `buf beta registry repository list` was doing a separate API request for each repository in the result set.

### DIFF
--- a/private/buf/bufprint/bufprint.go
+++ b/private/buf/bufprint/bufprint.go
@@ -20,7 +20,6 @@ import (
 	"io"
 	"strconv"
 
-	"github.com/bufbuild/buf/private/gen/proto/apiclient/buf/alpha/registry/v1alpha1/registryv1alpha1apiclient"
 	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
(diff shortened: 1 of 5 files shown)


johanbrandhorst (commented):
~This doesn't actually set the `owner_name` anywhere, does it? We'd still have to do this extra API request anywhere we call `newRepositoryPrinter`, no? I'm not sure I love this duplication in the `Repository` API type, maybe we should just use a new type defined inside `bufprint`?~

EDIT: wrong repository 😅

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +46,7 @@ message Repository {
   bool deprecated = 8;
   // deprecation_message is the message shown if the repository is deprecated.
   string deprecation_message = 9;
+  string owner_name = 10;
```
johanbrandhorst, line 49: I am a little concerned about the duplication here, but I recognize that this probably makes things easier both in the CLI and the frontend 👍🏻.

bufdev:
I don't think this is what we want to do, see my comment on https://github.com/bufbuild/buf/issues/769 - we probably want to just be able to query owners in batches, and then we can have the printing only make two calls, and then bring the information together within the printing.

PR #771, opened 2021-12-08
Proto changes for #768

WillAbides:
Proto changes for #768


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +46,7 @@ message Repository {
   bool deprecated = 8;
   // deprecation_message is the message shown if the repository is deprecated.
   string deprecation_message = 9;
+  string owner_name = 10;
```
amckinney, line 49: This one is tough.. it’s clearly more convenient to include the `owner_name` alongside the `owner_id` for this particular interaction, but it’s a slippery slope. We might be inclined to include other owner-specific information in the repository itself, which would continue to violate that boundary. There aren’t any other fields we would want to copy over right now, but it’s still something to be wary of.

Alternatively, we could reduce the number of RPCs if we send by collecting all of the ownerIDs associated with all of the repositories, and fire off a single `GetOwnerNamesForIDs` and assign them to each repository on the client side.

That means we would have 2 RPCs every time instead of n+1, but I’ll admit that the interaction doesn’t feel clean, especially given that we want to revisit the BSR API as a whole.

In the current state of the API, I feel like the 2 RPC approach is best, but I wonder if we can justify an exception for including the `owner_name` in the `Repository` since it’s always going to be included in the repository’s full name (e.g. `buf.build/<owner>/<repository>`).

bufdev:
I don't think this is what we want to do, see other comments

PR #773, opened 2021-12-08
Add new ulidutil package

johanbrandhorst:
This new function lets us creates ULIDs. This will be useful
for the audit logging work as ULIDs should index much more
efficiently for its usage patterns.

### DIFF
--- a/private/pkg/uuidutil/uuidutil.go
+++ b/private/pkg/uuidutil/uuidutil.go
@@ -39,7 +39,7 @@ func NewULID(timestamp time.Time) (uuid.UUID, error) {
 	if err != nil {
 		return uuid.Nil, err
 	}
-	return uuid.FromBytes(id[:])
+	return uuid.UUID(id), nil
 }
 
 // ToDashless returns the uuid without dashes.
--- a/private/pkg/uuidutil/uuidutil.go
+++ b/private/pkg/uuidutil/uuidutil.go
@@ -15,12 +15,9 @@
 package uuidutil
 
 import (
-	"crypto/rand"
 	"fmt"
-	"time"
 
 	"github.com/gofrs/uuid"
-	"github.com/oklog/ulid/v2"
 )
 
 // New returns a new random UUIDv4.
@@ -30,16 +27,6 @@ func New() (uuid.UUID, error) {
 		return uuid.Nil, err
 	}
 	return id, nil
-}
-
-// NewULID creates a new ULID for the given timestamp,
-// and returns it in its UUID form.
-func NewULID(timestamp time.Time) (uuid.UUID, error) {
-	id, err := ulid.New(ulid.Timestamp(timestamp), rand.Reader)
-	if err != nil {
-		return uuid.Nil, err
-	}
-	return uuid.UUID(id), nil
 }
 
 // ToDashless returns the uuid without dashes.
--- a/private/pkg/uuidutil/uuidutil.go
+++ b/private/pkg/uuidutil/uuidutil.go
@@ -15,9 +15,12 @@
 package uuidutil
 
 import (
+	"crypto/rand"
 	"fmt"
+	"time"
 
 	"github.com/gofrs/uuid"
+	"github.com/oklog/ulid/v2"
 )
 
 // New returns a new random UUIDv4.
@@ -27,6 +30,16 @@ func New() (uuid.UUID, error) {
 		return uuid.Nil, err
 	}
 	return id, nil
+}
+
+// NewULID creates a new ULID for the given timestamp,
+// and returns it in its UUID form.
+func NewULID(timestamp time.Time) (uuid.UUID, error) {
+	id, err := ulid.New(ulid.Timestamp(timestamp), rand.Reader)
+	if err != nil {
+		return uuid.Nil, err
+	}
+	return uuid.FromBytes(id[:])
 }
 
 // ToDashless returns the uuid without dashes.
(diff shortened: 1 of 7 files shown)


smallsamantha (approved):
:aw_thanks:

cyinma:
https://github.com/oklog/ulid/issues/33
Do we care about this?

johanbrandhorst:
> [oklog/ulid#33](https://github.com/oklog/ulid/issues/33) Do we care about this?

Great question! The docs mention that `math/rand` is not concurrency safe, which is why we're using `crypto/rand` in this implementation. You've reminded me that we should have a test that does some concurrent generations just to be safe though! I'll also do a bit of digging into whether `crypto/rand` is appropriate here, since we don't _necessarily_ need cryptographic randomness and it might be more expensive to generate than necessary.

johanbrandhorst:
It appears we're already using `crypto/rand` for our existing UUIDs, so we can probably keep using it:

https://github.com/gofrs/uuid/blob/6fe74fff266df57f91ca3f49a202057c27bf3f4b/generator.go#L164
https://github.com/gofrs/uuid/blob/6fe74fff266df57f91ca3f49a202057c27bf3f4b/generator.go#L26

private/pkg/uuidutil/uuidutil.go
```diff
@@ -29,6 +32,16 @@ func New() (uuid.UUID, error) {
 	return id, nil
 }
 
+// NewULID creates a new ULID for the given timestamp,
+// and returns it in its UUID form.
+func NewULID(timestamp time.Time) (uuid.UUID, error) {
```
mfridman, line 37: Why does the caller need to pass a timestamp? Shouldn't this be an implementation detail of the package?

johanbrandhorst: It could be, though we generally make it an anti-pattern to call `time.Now()` explicitly in the code, and I think this is an OK interface that makes it easy to test.

bufdev:
I do not know on face value if I'm comfortable saying we have two types of unique identifiers/I don't know if efficiency gains are enough to justify this.

johanbrandhorst:
> I do not know on face value if I'm comfortable saying we have two types of unique identifiers/I don't know if efficiency gains are enough to justify this.

Very fair. The theory is that ULIDs are much more economic for Postgres to index (which is of course an implementation detail of a specific implementation of `bufdaudit.Publisher`, but may also apply to other systems which we would be interested in using for this sort of time series data). We haven't done any hard measurements on whether this is true, but the cost here is pretty minimal IMO, as the return type is the same as the rest of the package. We could elaborate on when it is appropriate to use this type of UUID in the comment, or we could even not add it to this package and just implement it in the `bufdrpcaudit` package where it will be used. What do you think?

private/pkg/uuidutil/uuidutil.go
cyinma: I am not sure I understand what you mean by anti-pattern, as if we are not calling `time.Now()` here in `uuidutil`, we still need to call it when we use the `NewULID`?

mfridman: I think its worth making exceptions where it makes sense, if a caller needs a random UUID why do they need to bother passing a timestamp. Which opens questions on what _kind_ of time do I need to pass.

Also, can you clarify when we would use `NewULID` vs `NewUUID` ?

johanbrandhorst:
The good thing about ULID's in this case is that they are really interchangeable with UUIDv4 UUIDs - they're the same length (as you say), they're just generated in a fashion that makes then lexicographically (and numerically) sortable. This is what 10 ULIDs (in UUID format) generated 1 millisecond apart look like, together with their timestamps and base10 representations:
```
017d9ffd-e650-6e78-ad94-96ec471705eb, 2021-12-09T11:19:16.688-05:00, 1981510122232654061853666534466455019
017d9ffd-e651-9c1b-dd69-010a1f722145, 2021-12-09T11:19:16.689-05:00, 1981510122234078504430790603594408261
017d9ffd-e652-53a3-a04d-d2dbb3abdf54, 2021-12-09T11:19:16.690-05:00, 1981510122234945201851185935571214164
017d9ffd-e654-df91-bcbf-59b99add15dc, 2021-12-09T11:19:16.692-05:00, 1981510122238023854806191210533557724
017d9ffd-e655-bc58-a512-a99b25d2b176, 2021-12-09T11:19:16.693-05:00, 1981510122239066444628561196955644278
017d9ffd-e656-6ebd-6b32-f818d4f0bceb, 2021-12-09T11:19:16.694-05:00, 1981510122239908884813416485239176427
017d9ffd-e657-9bdb-7290-eca1b19691b5, 2021-12-09T11:19:16.695-05:00, 1981510122241330871057931659489218997
017d9ffd-e658-62f7-66a2-497186693baa, 2021-12-09T11:19:16.696-05:00, 1981510122242271137637052776601697194
017d9ffd-e659-ddc8-b9fd-6d65b5eefeb8, 2021-12-09T11:19:16.697-05:00, 1981510122244060053543522967493738168
017d9ffd-e65b-13fa-7d40-dfdc0426a647, 2021-12-09T11:19:16.699-05:00, 1981510122245524905113887682557552199
```

As can be tested (or checked manually) these are ordered by their creation time, which is a convenient property for creating b-tree indices on their value. This should in theory provide use with faster insertions into the audit logging table, since the index will be easy to update. But developers don't need to care about where the UUID was generated unless they really want to use a lexicographically sortable version.

lrewega:
We talked offline and I agree that the distinction between UUIDv4 and ULID/UUIDv6 is muddy. I think there is consensus on having a distinct ULID package and treating ULIDs as different types throughout the code.

johanbrandhorst:
I've updated the PR to add a new package rather than changing `uuidutil`. Let me know what you think!

lrewega (approved):
🚢

bufdev (approved):

PR #775, opened 2021-12-09
Fix validation for duplicate plugins

### LINKED ISSUE #774: Cannot create separate client/server implementations of packages in C#
Real quick, the relevant `buf.gen.yaml`:

```
plugins:
  - name: csharp
    out: csharp-abstractions
  - name: grpc-csharp
    out: csharp-client
    opt:
      - no_server
  - name: grpc-csharp
    out: csharp-server
    opt:
      - no_client
```

What I'm trying to do here is create three separate NuGet packages for internal use. The first, output to the csharp-abstractions directory, contains the generated contracts. The second, output to the csharp-client directory, contains the generated client. The third, output to the csharp-server directory, contains the generated server stub. The NuGet packages are defined such that the csharp-client & csharp-server packages depend on the csharp-abstractions package. Doing this directly with the C# tooling works, and doing the equivalent with protoc from the command line works.

However, when I attempt to use buf with this .yaml, I get this error message: `Failure: file "SaladBowlServiceGrpc.cs" was generated multiple times: once by plugin "grpc-csharp" and again by plugin "grpc-csharp"`

Right now, I can do a workaround either by using the CLI or by commenting out either the serverside or clientside portion of the .yaml, running buf, and then swapping the commented-out portion of the file to generate the other file.

We need both of these files to refer to the same service and exist in the same C# namespace. I do not need them to have the same filename or classname.

Is there a workaround for this issue? I'm pretty sure this is a reasonable use case for C#. I saw what looked like a relevant issue [here](https://github.com/bufbuild/buf/issues/760), but it's not the same issue.

amckinney:
Fixes #774

Note that this does _not_ fix the equivalent behavior for `buf protoc` yet. The solution is more nuanced there because the implementation strongly assumes that there can only one `_out` per plugin. At an initial glance, `buf protoc` will require more significant refactors in order to get this working. For reference, check it out [here](https://github.com/bufbuild/buf/blob/0786cea8690ed81cdd1f5b31b9b5c0529ffdab74/private/buf/cmd/buf/command/protoc/flags.go#L356).

We will follow-up with larger refactors to `buf protoc`, but it's left for a follow-up so that we focus on the user issue at hand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 - Use a separate repository service to for each dependency remote to resolve dependencies for `buf mod update`. Previously, we used a single repository service based on the remote
   from the module, so it was unable to resolve dependencies from differente remotes.
 - Display the user-provided Buf Schema Registry remote, if specified, instead of the default within the `buf login` message.
+- Fix issue where `buf generate` fails when the same plugin was specified more than once in a single invocation.
 - Update the digest algorithm so that it encodes the `name`, `lint`, and `breaking` configuration encoded in the `buf.yaml`.
   When this change is deployed, users will observe the following:
   - Users on `v0.43.0` or before will notice mismatched digest errors similar to the one described in https://github.com/bufbuild/buf/issues/661.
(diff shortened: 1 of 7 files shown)


doriable (approved):

doriable:
Interesting that the code generator response does not take into account the out destination, I guess this is just handled at the plugin invocation layer above. Nice fix :)

PR #781, opened 2021-12-13
Add `nil` check for `ConfigForProto` for `breaking` and `lint` configs

doriable:
As a part of #682, we are encoding the modules, but we need to have a `nil` check for mappers for the `Config` structures from the proto structures. Remote references stored in the BSR will not have configs stored and this needs to handle the `nil` case seamlessly.

As a follow-up, tests in `core` will need to be added with unmarshalling references from the `DownloadService`.

### DIFF
--- a/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
+++ b/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
@@ -76,6 +76,9 @@ func NewConfigV1(externalConfig ExternalConfigV1) *Config {
 
 // ConfigForProto returns the Config given the proto.
 func ConfigForProto(protoConfig *breakingv1.Config) *Config {
+	if protoConfig == nil {
+		return nil
+	}
 	return &Config{
 		Use:                           protoConfig.GetUseIds(),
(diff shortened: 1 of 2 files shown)


amckinney (approved):

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -76,6 +76,9 @@ func NewConfigV1(externalConfig ExternalConfigV1) *Config {
 
 // ConfigForProto returns the Config given the proto.
 func ConfigForProto(protoConfig *breakingv1.Config) *Config {
+	if protoConfig == nil {
```
bufdev, line 79: The codebase generally expects callers to validate this, ie if you have a pointer you are passing, it is expected to exist...I would think this should be fixed at a higher level

doriable: Let me open an issue to address this at a higher level of the code base -- I think this is a valid concern. Right now, since we're migrating and dealing with older module structures alongside new module structures that will contain the config, the nil checks have been put in at the constructor level.

PR #783, opened 2021-12-13
new display endpoints for explicit resource role

cyinma:


proto/buf/alpha/registry/v1alpha1/authz.proto
```diff
@@ -79,12 +82,21 @@ service AuthzService {
   // UserCanManageRepositoryContributors returns whether the user is authorized to manage
   // any contributors to the repository and the list of roles they can manage.
   rpc UserCanManageRepositoryContributors(UserCanManageRepositoryContributorsRequest) returns (UserCanManageRepositoryContributorsResponse);
+  // UserCanManageRepositoryContributor returns whether the user is authorized to manage
+  // a specific user to the repository and the list of roles they can manage.
```
johanbrandhorst, line 86: Nit: I think you say _on_ the repository?
```suggestion
  // UserCanManageRepositoryContributor returns whether the user is authorized to manage
  // a specific user on the repository and the list of roles they can manage.
```

proto/buf/alpha/registry/v1alpha1/authz.proto
```diff
@@ -293,6 +315,21 @@ message UserCanManageRepositoryContributorsResponse {
   repeated RepositoryRole authorized_roles = 1;
 }
 
+message UserCanManageRepositoryContributorRequest {
+  // The ID of the repository for which to check
+  // whether the user is authorized.
+  string repository_id = 1;
+  // The ID of the target user for which to check
+  // whether the user is authorized.
```
johanbrandhorst, line 323:
```suggestion
  // whether the user is authorized to manage.
```

proto/buf/alpha/registry/v1alpha1/authz.proto
```diff
@@ -293,6 +315,21 @@ message UserCanManageRepositoryContributorsResponse {
...
+  string user_id = 2;
+}
+
+message UserCanManageRepositoryContributorResponse {
+  // The list of roles that the user is authorized to manage against target user,
```
johanbrandhorst, line 328:
```suggestion
  // The list of roles that the user is authorized to manage against the target user,
```

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+import "buf/alpha/registry/v1alpha1/role.proto";
+
+// DisplayService is the display service that
+// provides information regarding to UI element display.
```
johanbrandhorst, line 22:
```suggestion
// DisplayService provides information regarding UI element displaying
// based on the users role against a resource.
```

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+service DisplayService {
+  // DisplayRepositoryRolesToManage returns which roles should be displayed
+  // to the user when they are managing contributors on the repository.
+  rpc DisplayRepositoryRolesToManage(DisplayRepositoryRolesToManageRequest) returns (DisplayRepositoryRolesToManageResponse);
```
johanbrandhorst, line 26: I don't know if we need to include `Display` in the RPC name since it's in the service. What do you think about
```suggestion
  // ListManagableRepositoryRoles returns which roles should be displayed
  // to the user when they are managing contributors on the repository.
  rpc ListManagableRepositoryRoles(ListManagableRepositoryRolesRequest) returns (ListManagableRepositoryRolesResponse);
```
If this endpoint is only used for adding users to a repository (in contrast to `DisplayRepositoryRolesToManageByUser`), maybe we can use a different word than "manage" too, maybe `ListAssignableRepositoryRoles`?

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+  // DisplayRepositoryRolesToManageByUser returns which roles should be displayed
+  // to the user when they are managing a specific contributor on the repository.
+  rpc DisplayRepositoryRolesToManageByUser(DisplayRepositoryRolesToManageByUserRequest) returns (DisplayRepositoryRolesToManageByUserResponse);
```
johanbrandhorst, line 29: How about 
```suggestion
  // ListManagableRepositoryRoles returns which roles should be displayed
  // to the user when they are managing a specific contributor on the repository.
  rpc ListManagableRepositoryRoles(ListManagableRepositoryRolesRequest) returns (ListManagableRepositoryRolesResponse);
```

Maybe `Manageable` is enough to indicate that this is per-user, as opposed to `Assignable`, which is for adding a new role?

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+  // to the user when they are managing contributors on the template.
+  rpc DisplayTemplateRolesToManage(DisplayTemplateRolesToManageRequest) returns (DisplayTemplateRolesToManageResponse);
+  // DisplayTemplateRolesToManageByUser returns which roles should be displayed
+  // to the user when they are managing a specific contributor on the template.
+  rpc DisplayTemplateRolesToManageByUser(DisplayTemplateRolesToManageByUserRequest) returns (DisplayTemplateRolesToManageByUserResponse);
+}
+
+message DisplayRepositoryRolesToManageRequest {
+  // The ID of the repository for which to check
+  // which roles should be displayed to manage.
```
johanbrandhorst, line 46: (If we go with `Assignable`):
```suggestion
  // The ID of the repository for which to check
  // which roles should be displayed as assignable.
```

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+  string repository_id = 1;
+}
+
+message DisplayRepositoryRolesToManageResponse {
+  // The list of roles that should be displayed to the user to manage.
```
johanbrandhorst, line 51:
```suggestion
  // The list of roles that should be displayed to the user as assignable.
```

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+  repeated RepositoryRole roles = 1;
+}
+
+message DisplayRepositoryRolesToManageByUserRequest {
+  // The ID of the repository for which to check
+  // which roles should be displayed to manage.
+  string repository_id = 1;
+  // The ID of the target user for which to check
+  // which roles should be displayed to manage.
+  string user_id = 2;
```
johanbrandhorst, line 61:
```suggestion
  // The ID of the repository for which to check
  // which roles should be displayed as manageable.
  string repository_id = 1;
  // The ID of the target user for which to check
  // which roles are manageable.
  string user_id = 2;
```

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+}
+
+message DisplayRepositoryRolesToManageByUserResponse {
+  // The list of roles that should be displayed to the user to manage.
```
johanbrandhorst, line 65:
```suggestion
  // The list of roles that should be displayed to the user as manageable.
```

johanbrandhorst (commented):
Just bike shedding on naming and comments

proto/buf/alpha/registry/v1alpha1/display.proto
cyinma: `based on the users role against a resource`
is this necessary true? server and organization are probably not considered as a resource, but there will be display endpoints related to them later. Also I am trying to avoid using words related to authorization/role/permission to avoid confusing with other services, but it seems unavoidable.
how about just `based on the users roles`?

proto/buf/alpha/registry/v1alpha1/display.proto
cyinma: I think our current practice is the RPC name should be self explainable without the service name, e.g. `RepositoryService.DeleteRepository`, `OrganizationService.DeleteOrganization`. 
I think `ListAssignableRepositoryRoles` is good and self explainable without service name, but for some endpoints later that only return a bool we may still want to put the `Display` in the RPC name, what do you think?
besides, this endpoint should apply to all add/update/remove action (cc @timostamm @seankimdev correct me if I am wrong), so I think `ListManageableRepositoryRoles` is good enough.

proto/buf/alpha/registry/v1alpha1/display.proto
cyinma: I think we will still want a Manageable endpoints that is not per-user (I assume this is one of the use case for `UserCanManageRepositoryContributors` in the contributors list to determine show setting icon or not, again  @timostamm @seankimdev correct me if I am wrong)

proto/buf/alpha/registry/v1alpha1/display.proto
timostamm: > this endpoint should apply to all add/update/remove action

Yes, this should be the one that applies to the list of contributors, and does _not_ take into account that individual users may already have an implicit role on the repository.

proto/buf/alpha/registry/v1alpha1/display.proto
johanbrandhorst: Ah right, so we have 3 use cases for these two RPCs:

1. What roles to display when user presses "Add User To Organization". Is this covered by `ListManageableRepositoryRoles`? I could see it being user-dependent, so maybe it should be covered by `ListManageableUserRepositoryRoles` instead?
2. On which members to display the cog wheel icon when viewing the list of members. This is covered by `ListManageableRepositoryRoles` I assume, since it's simple to do a check for the each listed users role against such a list.
3. What roles to display when _changing_ an existing users role. This is presumably covered by `ListManageableUserRepositoryRoles`, since it's user specific.

Am I missing any use cases? Maybe we should have 3 RPCs that map cleanly to each use case instead? Or is there some other way we can unify these? CC @timostamm.

proto/buf/alpha/registry/v1alpha1/display.proto
johanbrandhorst: I think I agree that we can use `Display` as a prefix for `DisplayService.DisplayRepositorySettings` for example 👍🏻.

proto/buf/alpha/registry/v1alpha1/display.proto
cyinma: I will let Timo answer if any use case missing / incorrect as he knows better about that, but considering these 3 use cases, I think 1 and 3 share exactly the same request/response/behaviour, and in fact representing the same thing? I don't think we will need 3 RPCs if that is the case

timostamm: I concur with @cyinma - `ListManageableRepositoryRoles` for case 2, and `ListManageableUserRepositoryRoles` for cases 1 and 3.

johanbrandhorst: Sounds good to me 👍🏻

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -0,0 +1,117 @@
...
+  // to the user when they are managing contributors on the template.
+  rpc ListManageableTemplateRoles(ListManageableTemplateRolesRequest) returns (ListManageableTemplateRolesResponse);
+  // ListManageableUserTemplateRoles returns which roles should be displayed
+  // to the user when they are managing a specific contributor on the template.
+  rpc ListManageableUserTemplateRoles(ListManageableUserTemplateRolesRequest) returns (ListManageableUserTemplateRolesResponse);
+}
+
+message ListManageableRepositoryRolesRequest {
+  // The ID of the repository for which to check
+  // which roles should be displayed as manageable.
```
robbertvanginkel, line 46: Nittest of nits: the width of the comments is a bit inconsistent between here and the response message, looks a bit odd. Same for the rest of the file.

robbertvanginkel (approved):

PR #784, opened 2021-12-13
Move `nil` checks out of `breaking` and `lint` mappers

doriable:
A small clean-up related to #682 re: https://github.com/bufbuild/buf/pull/781#discussion_r767933162
This moves the `nil` checks to the call sites and keeps the concerns of the constructors/mappers in the `bufbreakingconfig` and `buflintconfig` packages clean.


WillAbides (approved):
LGTM - Let's Get This Merged

PR #786, opened 2021-12-14
Fix private/usage for Golang 1.18

bufdev:
In 1.18, https://pkg.go.dev/runtime/debug#ReadBuildInfo returns true for tests, when it returned false in 1.17. The resulting `*debug.BuildInfo` has an empty `Main.Path`, which results in a panic.

### DIFF
--- a/private/usage/usage.go
+++ b/private/usage/usage.go
@@ -32,13 +32,23 @@ func init() {
 func check() error {
 	buildInfo, ok := debug.ReadBuildInfo()
 	if !ok {
-		if !strings.HasSuffix(os.Args[0], testSuffix) && filepath.Base(os.Args[0]) != debugBin {
-			return errors.New("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects")
+		if shouldSkip() {
+			return nil
 		}
-		return nil
+		return errors.New("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects")
+	}
+	if buildInfo.Main.Path == "" {
+		if shouldSkip() {
+			return nil
+		}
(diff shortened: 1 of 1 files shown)


lrewega (approved):

private/usage/usage.go
```diff
@@ -32,13 +32,23 @@ func init() {
...
 	if !ok {
-		if !strings.HasSuffix(os.Args[0], testSuffix) && filepath.Base(os.Args[0]) != debugBin {
-			return errors.New("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects")
+		if shouldSkip() {
+			return nil
 		}
-		return nil
+		return errors.New("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects")
+	}
+	if buildInfo.Main.Path == "" {
```
lrewega, line 40: `if !ok || buildInfo.Main.Path == "" {`

private/usage/usage.go
```diff
@@ -32,13 +32,23 @@ func init() {
...
+		return errors.New("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects")
 	}
 	if !strings.HasPrefix(buildInfo.Main.Path, "github.com/bufbuild") {
 		return fmt.Errorf("github.com/bufbuild/buf/private code must only be imported by github.com/bufbuild projects but was used in %s", buildInfo.Main.Path)
 	}
 	return nil
 }
+
+func shouldSkip() bool {
+	return strings.HasSuffix(os.Args[0], testSuffix) || filepath.Base(os.Args[0]) == debugBin
```
lrewega: Can expand this inline if there is only once branch.

lrewega (approved):

Issue #790, opened 2021-12-16
[1.0.0-rc9] does not match stored digest


GaoJianAllen:
I update the buf to 1.0.0-rc9 and meet warning as below:
**_Module "buf.build/googleapis/googleapis:5860854adf6a48c39b19d54342b68385" has invalid cache state: calculated digest "b3-xi5p-ipG4lG8iOzD7LBmNPs015OkSjXFmgTuGfQ4URQ=" does not match stored digest "b2-WayFxGJKhSLbpMCQ-VZ5-3R5Gj9iRFYsMG7o57lgqog=". The cache will attempt to self-correct._**

The rc9 version can work well but the wierd warning.

doriable:
Hi @GaoJianAllen, thanks for filing the issue. In `v1.0.0-rc9`, we are expecting a cache invalidation since we changed the digest that we are using for the cache ([release notes](https://github.com/bufbuild/buf/releases/tag/v1.0.0-rc9)).

> Users on v0.44.0 or after will have their module cache invalidated, but it will repair itself automatically.

This will result in the warning that you see above. Just for reference, this seems to be happening when you run `buf mod update`, is that correct? Once the cache has self-corrected, you should not continue to see the warnings.

doriable:
It looks like this is happening with remote targets during the generation process, I was able to reproduce locally. Looking into this now!

doriable:
The fix has been merged and a release as made https://github.com/bufbuild/buf/releases/tag/v1.0.0-rc10.

PR #791, opened 2021-12-16
Default to the `v1beta1` configuration in the module constructor for proto structures

doriable:
Now that we have migrated to the `b3` digest, we are using the `breaking` and `lint` configs for the digest calculation. For remote references, we do not have any `breaking` and `lint` configs stored. Thus, when we construct the module based on the response we are receiving from the proto, we need to create the same default configurations that are created for modules from buckets without a `buf.yaml`.

Fixes: #790

### DIFF
--- a/private/bufpkg/bufconfig/write.go
+++ b/private/bufpkg/bufconfig/write.go
@@ -280,6 +280,11 @@ func writeConfig(
 			return fmt.Errorf("version %q found for breaking config, does not match top level config version: %q", breakingConfigVersion, version)
 		}
 		config.Breaking = breakingConfig
+	} else {
+		// Otherwise, we set a breaking config with the given version.
+		config.Breaking = &bufbreakingconfig.Config{
+			Version: version,
+		}
 	}
 	var lintConfigVersion string
 	lintConfig := writeConfigOptions.lintConfig
@@ -289,6 +294,11 @@ func writeConfig(
 			return fmt.Errorf("version %q found for lint config, does not match top level config version: %q", lintConfigVersion, version)
 		}
 		config.Lint = lintConfig
+	} else {
+		// Otherwise, we set a breaking config with the given version.
+		config.Lint = &buflintconfig.Config{
+			Version: version,
+		}
 	}
 	// We should never hit this condition since we are already validating against `version`.
 	if breakingConfigVersion != lintConfigVersion {
--- a/private/bufpkg/bufconfig/write.go
+++ b/private/bufpkg/bufconfig/write.go
@@ -295,7 +295,7 @@ func writeConfig(
 		}
 		config.Lint = lintConfig
 	} else {
-		// Otherwise, we set a breaking config with the given version.
+		// Otherwise, we set a lint config with the given version.
 		config.Lint = &buflintconfig.Config{
 			Version: version,
 		}
--- a/private/bufpkg/bufconfig/write.go
+++ b/private/bufpkg/bufconfig/write.go
@@ -263,7 +263,7 @@ func writeConfig(
 	// This is the same default as the bufconfig getters.
 	version := V1Beta1Version
 	if writeConfigOptions.version != "" {
-		if err := validateVersion(writeConfigOptions.version); err != nil {
+		if err := ValidateVersion(writeConfigOptions.version); err != nil {
 			return err
 		}
 		version = writeConfigOptions.version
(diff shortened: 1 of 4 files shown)


amckinney:
Let's better understand #790, too. This behavior will only occur when `buf` on `v1.0.0-rc9` or later reads a module that was produced by `buf` on version `v1.0.0-rc8` or earlier. Otherwise, the module will have the default `breaking` and `lint` configuration value and the digest will include those values.

However, this still doesn't cover the case for eventual users of the BSR API - the `lint` and `breaking` configurations aren't required so anyone pushing modules to the BSR by interacting the API directly (instead of via `buf`) will be able to trigger these cases (i.e. the new `buf` CLI could download a module with empty `breaking` and `lint` configurations).

private/bufpkg/bufmodule/bufmodulecache/module_cacher.go
```diff
@@ -142,6 +139,18 @@ func (m *moduleCacher) PutModule(
...
+		dataReadWriteBucket,
+		bufmodule.ModuleWithModuleIdentityAndCommit(modulePin, modulePin.Commit()),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to write module to bucket: %w", err)
+	}
+	digest, err := bufmodule.ModuleDigestB3(ctx, bucketModule)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
doriable: Yeah, I think that's a reasonable pushback. Let me take a look at populating the default config for the returned proto structure. In that case, we likely need to file an issue to clean-up some of the configuration package so that the defaulted behaviour is less spread out and more clear across the package.

private/bufpkg/bufmodule/module.go
```diff
@@ -61,11 +61,16 @@ func newModuleForProto(
...
+	breakingConfig := &bufbreakingconfig.Config{
+		Version: bufconfig.V1Beta1Version,
+	}
 	if protoModule.GetBreakingConfig() != nil {
 		breakingConfig = bufbreakingconfig.ConfigForProto(protoModule.GetBreakingConfig())
 	}
-	var lintConfig *buflintconfig.Config
+	lintConfig := &buflintconfig.Config{
+		Version: bufconfig.V1Beta1Version,
+	}
```
amckinney, line 73: I think we still need to verify that those versions are always the same right? Imagine the case when a `v1` lint configuration was found, but we default to `v1beta1` for the breaking configuration.

Thinking through our options, it might be best to attempt to default to `v1beta1` if the configuration is not present, but return an error if lint and breaking are mismatched. So if the Protobuf message contains a `v1` version for one of the configurations but not the other, we would have an error.

What are your thoughts?

private/bufpkg/bufconfig/write.go
```diff
@@ -280,6 +280,11 @@ func writeConfig(
 			return fmt.Errorf("version %q found for breaking config, does not match top level config version: %q", breakingConfigVersion, version)
 		}
 		config.Breaking = breakingConfig
+	} else {
+		// Otherwise, we set a breaking config with the given version.
+		config.Breaking = &bufbreakingconfig.Config{
+			Version: version,
+		}
```
amckinney, line 287: Are these still needed if we go with the `newModule` approach that you have below?

private/bufpkg/bufmodule/module.go
doriable: Oh hmm, that is... probably reasonable, tbh. We _shouldn't_ hit the case where if neither is `nil`, they're mismatched, but that's probably good to check here as a safety thing, similar to what we do in `WriteConfig`.

private/bufpkg/bufconfig/write.go
doriable: My conjecture is that if we are defaulting everywhere, we should probably default on the `WriteConfig` also... but maybe that is just... too much. But yes, the code path is not _required_, it's mostly for consistency in behaviour. I can remove it for now, since it'll be one less place we'll need to deal with for the `v1` migration.

amckinney (approved):
Please confirm that the latest changes continue to fix the relevant issue. Otherwise lgtm.

doriable:
> Please confirm that the latest changes continue to fix the relevant issue. Otherwise lgtm.

I ran some `buf generate` directives against both a local cluster as well as the public production cluster.

PR #792, opened 2021-12-16
remove fields deleted in PluginConfig and PluginVersionMapping

cyinma:
These fields were deprecated and are now safe to be removed.


proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -83,9 +83,6 @@ message PluginConfig {
   // Parameters that should be provided to the plugin. These are
   // joined with a "," before being provided to the plugin at runtime.
   repeated string parameters = 3;
-  // True if the source plugin has been deleted. This dooms
-  // the template.
-  bool deleted = 4 [deprecated=true];
```
mfridman, line 88: As a best practice, we should either use `reserved` or decrement the field numbers. Leaving the `4` omitted will cause confusing to the next person.

mfridman: @cyinma When you get a chance, can you update the Buf style guide ?

Framing the question: "How do I remove a field from our public Protobuf definitions?"

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -76,16 +76,14 @@ message Template {
 
 // PluginConfig defines a runtime configuration for a plugin.
 message PluginConfig {
+  reserved 4;
```
johanbrandhorst, line 79: We need to reserve the name too
```suggestion
  reserved 4;
  reserved "deleted";
```

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -110,15 +108,13 @@ message TemplateVersion {
 
 // PluginVersionMapping maps a plugin_id to a version.
 message PluginVersionMapping {
+  reserved 4;
```
johanbrandhorst, line 111: Same here

Issue #795, opened 2021-12-16
Add buf generate --include-wkt


chathaway-codes:
I'm trying to use buf to generate files for a Dart/Flutter project, and I'm having trouble with the error:

```
Target of URI hasn't been generated: '../../google/protobuf/timestamp.pb.dart'.
Try running the generator that will generate the file referenced by the URI.
```

Based on [this discussion](https://gitanswer.com/buf-buf-does-not-generate-all-dart-files-go-961661179), this is intentional for other languages.

From what I can tell in the [Dart sample projects](https://github.com/grpc/grpc-dart/tree/master/example/googleapis), it is expected that these files will be generated, and they are not provided by the runtime as they are in other languages.

Could an option be added to really, really say "please generate everything"; or at least generate everything that is needed?

amckinney:
Thanks for the issue - this is one we've debated for quite a while.

We could roll out a `--include-wkt` flag to support generating the well-known types explicitly, but it has some unfortunate consequences.

The real solution here is that we should provide a Protobuf standard library so that users always explicitly define their dependencies (i.e. in the `deps:` key in their `buf.yaml` configuration), and the well-known types wouldn't need to exist at all - the standard library would just be treated as an ordinary dependency.

We're wary to support for a flag that could have long-term negative side effects for the Buf community as a whole. `dart` is the only culprit implementation that does _not_ include the types in the runtime library itself, so it's hard to justify on its own.

The work-around today is to have the well-known type you need included alongside your module sources. In this case, it looks like you need the `google/protobuf/timestamp.proto` file. I know it's far from ideal, but it'll at least get you moving while we get an answer here in the meantime.

amckinney:
For some more information on this issue from the `protoc-gen-dart` side, check out [this](https://github.com/google/protobuf.dart/issues/496) issue.

chathaway-codes:
Cool; thanks for the explanation.

I started work on a `--include-wkt` flag at https://github.com/bufbuild/buf/pull/797, if others want to use that as a temporary work around.

I look forward to whichever 'real solution' you folks come up with :).

bufdev:
@chathaway-codes this feature has now been merged in #892 and will go out in the next release.

pablojimpas:
> @chathaway-codes this feature has now been merged in #892 and will go out in the next release.

Is there a way to enable this flag on a plugin by plugin basis? So that we can enable it only for Dart.

bufdev:
[...]

PR #797, opened 2021-12-17
add: flag to generate files for well-known types

chathaway-codes:
DO NOT SUBMIT

Tests are failing, but this works to generate code

Related discussion in https://github.com/bufbuild/buf/issues/795; it sounds like this is not likely to be merged, so this PR is just for record keeping.

### DIFF
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -145,6 +145,13 @@ func GenerateWithIncludeImports() GenerateOption {
 	}
 }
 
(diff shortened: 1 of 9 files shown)


chathaway-codes:
Usage for anyone who needs this temporary work around:

```
go install github.com/chathaway-codes/buf/cmd/buf@latest

# Ensure you are using the install ~/go/bin/buf; my system used one in /usr/local/shared instead
buf generate --include-imports --include-well-known-types
```

PR #806, opened 2021-12-20
Proto definition for deleting a track

WillAbides:
This adds DeleteRepositoryTrackByName


proto/buf/alpha/registry/v1alpha1/repository_track.proto
```diff
@@ -61,3 +63,12 @@ message ListRepositoryTracksResponse {
   // There are no more pages if this is empty.
   string next_page_token = 2;
 }
+
+message DeleteRepositoryTrackByNameRequest {
+  // The ID of the repository this track should be deleted from.
+  string repository_id = 1;
+  // The name of the repository track.
+  string name = 2;
+}
```
amckinney, line 72: Mixing an ID with a name might be painful to use from the CLI. Should we do this by the split name representation used elsewhere (i.e. `owner`, `repository`, etc.)?

amckinney (approved):

PR #807, opened 2021-12-20
Update protos for plugin service

mfridman:


mfridman:
Do we also need to add a oneof `action_details` ? Not entirely sure when that's used.

johanbrandhorst:
> Do we also need to add a oneof `action_details` ? Not entirely sure when that's used.

Adding an entry to the action details is mostly useful when there is information we want to store that isn't captured by one of the "objects" of the operation. I think in this case it might actually be interesting to know what the runtime dependencies are of an image that's being uploaded, so we could add an action detail for this operation.

johanbrandhorst (approved):
LGTM!

PR #808, opened 2021-12-20
Add config validation as a part of validating proto module

doriable:
This adds validation for the `breaking` and `lint` configs as a part of `ValidateProtoModule`.
This mostly looks at valid versions, we do not validate the rules -- this gets validated as a part of the rule builder.
`ValidateProtoModule` is also used by the push service validation layer, so once this is merged, the BSR server-side logic can be upgraded to use this same validation.

### DIFF
--- a/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
+++ b/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
@@ -16,6 +16,7 @@ package bufbreakingconfig
 
 import (
 	"encoding/json"
+	"fmt"
 	"sort"
 
 	breakingv1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/breaking/v1"
@@ -95,6 +96,21 @@ func ProtoForConfig(config *Config) *breakingv1.Config {
 		IgnoreIdPaths:          protoForIgnoreIDOrCategoryToRootPaths(config.IgnoreIDOrCategoryToRootPaths),
 		IgnoreUnstablePackages: config.IgnoreUnstablePackages,
 		Version:                config.Version,
+	}
+}
+
+// ValidateProtoConfig validates that the proto config is valid.
+func ValidateProtoConfig(protoConfig *breakingv1.Config) error {
+	// An empty config is valid.
+	if protoConfig == nil {
+		return nil
+	}
+	version := protoConfig.GetVersion()
+	switch version {
+	case v1Version, v1Beta1Version:
+		return nil
+	default:
+		return fmt.Errorf("invalid breaking config version %q provided", version)
 	}
 }
 
--- a/private/bufpkg/bufcheck/buflint/buflintconfig/buflintconfig.go
+++ b/private/bufpkg/bufcheck/buflint/buflintconfig/buflintconfig.go
@@ -17,6 +17,7 @@ package buflintconfig
 import (
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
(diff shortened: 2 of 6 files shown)


private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -98,6 +99,21 @@ func ProtoForConfig(config *Config) *breakingv1.Config {
...
+	}
+	version := protoConfig.GetVersion()
+	switch version {
+	case v1Version, v1Beta1Version:
+		return nil
+	default:
+		return fmt.Errorf("invalid breaking config version %q provided", version)
+	}
+}
+
```
amckinney, line 116: I don't know if we want to validate on the client-side. This is how we ended up in that issue about requiring fields in older versions of the CLI, but changing that requirement on the server-side later (the `v0.43.0` -> `v0.44.0` change). We need to think through this.

private/bufpkg/bufcheck/buflint/buflintconfig/buflintconfig.go
```diff
@@ -131,6 +132,22 @@ func ProtoForConfig(config *Config) *lintv1.Config {
...
+	if protoConfig == nil {
+		return nil
+	}
+	version := protoConfig.GetVersion()
+	switch version {
+	case v1Version, v1Beta1Version:
+		return nil
+	default:
+		return fmt.Errorf("invalid lint config version %q provided", version)
+	}
```
amckinney, line 147: We need to validate a lot more than this. The Protobuf definition uses free-form strings for its IDs and `ignore_only` paths. At the very least, we should validate that each ID is real with respect to the configured version _and_ that every path is normalized.

Plus, I think this should exist on the server side in the module validator before anything else. We can export functions like this for the BSR to use, but this placement is susceptible to being used by the CLI later on (i.e. accidentally by another engineer) even if we don't mean it to.

amckinney:
A good counterargument to client-side validation is if we ever add rules and/or categories to our `breaking` or `lint` configuration.

Suppose that we add a `FOO` category to our `v1` lint rules in a backward compatible manner. If a new client supports this rule and pushes it to the BSR, then an old client could pull it down. If the old client validates against all of the rules/categories, it will complain that the `FOO` rule does not exist.

Instead, to use Protobuf's compatibility strategy, those unrecognized rules should just be dropped. This would require some changes to Protobuf conversion though, so we still need to think through whether or not that's a good strategy in general. 

Note that this would be behaviorally different than reading the rules from a local `buf.yaml` - we would still validate that all of those rules exist in the current version of the CLI. This only pertains to our compatibility story between. the client and the server.

doriable:
Our primary goal is to make sure we don't store modules with "malformed configurations", which are defined as:

- Breaking and lint configurations that are not empty but do not contain a valid version, where valid version is defined as one of the supported versions (`v1` and `v1beta1`) and that the versions match each other and the broader configuration

I guess the question here is "are invalid rules and/or paths considered malformed configurations"? I think at the very least, they would be unusable configurations, and if we want to make sure that we are always pushing _usable_ configurations, that validation should probably be isolated to the server side.

As for the workflow where an older client version is validating configurations pushed to the server by a newer client, we currently do not have any workflows that use the remote configurations. Given that this would need a new workflow/code path, I think it's valid to say that is separate from reading local `buf.yaml` configurations and that we would be able to drop unrecognised rules that come in from the server-side, since we can always assume the server is returning a validated configuration.

Given that's the case, I'm going to close this PR in favour of a more rigourous configuration check from the server-side.

PR #814, opened 2021-12-22
add grouped display service endpoints

cyinma:
add grouped display service endpoints, some of the usage of existing authz service endpoints should be migrated to these new endpoints after implemented.


proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -41,6 +54,86 @@ service DisplayService {
...
+  // Display create organization repository element if true.
+  bool create_repository = 1;
+  // Display create organization plugin element if true.
+  bool create_plugin = 2;
+  // Display create organization template element if true.
+  bool create_template = 3;
+  // Display organization settings element if true.
+  bool settings = 4;
+  // Display update organization settings element if true.
+  bool update_settings = 5;
```
johanbrandhorst, line 73: As I understand it, there is no difference between these two right now, right? Admins can see and change org settings but can't delete the org, right?

proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -41,6 +54,86 @@ service DisplayService {
...
+
+message DisplayPluginElementsRequest {
+  // The ID of the plugin for which to check
+  // which elements should be displayed.
+  string plugin_id = 1;
+}
+
+message DisplayPluginElementsResponse {
+  // Display create plugin version element if true.
+  bool create_version = 1;
```
johanbrandhorst, line 99: Is this even something you can do from the UI today? I think we do this by pushing to the registry.

johanbrandhorst (commented):
Don't forget to add new auditv1alpha1 actions for the new RPCs!

proto/buf/alpha/registry/v1alpha1/display.proto
timostamm: Plugins are pushed to the registry, not uploaded in a form on the frontend, but we _do_ show a slightly different UI for users who are permitted to do so, with some instructions to get started.

johanbrandhorst (approved):
LGTM

PR #816, opened 2021-12-23
remove deleteuser audit log details

smallsamantha:


johanbrandhorst (approved):
Sneaky reordering of the Actions 🤔

smallsamantha:
> Sneaky reordering of the Actions 🤔

we're letting this slide since the api isn't in use yet

PR #817, opened 2021-12-23
Make tags repeated in commits

johanbrandhorst:
This was missed in the initial implementation


johanbrandhorst:
Compare: https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/repository_commit.proto#L42

smallsamantha (approved):

PR #819, opened 2021-12-23
Speed up the windows workflow

WillAbides:
This speeds up the windows workflow by:
- moving caches to D:\ where restoring cache is much faster
- not using chocolatey to install curl and zip (or anything else)

With these changes the windows workflow typically takes between 2 and 4 minutes. Previously it took around 9 minutes.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -1,5 +1,8 @@
 name: windows
-on: push
+on:
+  push:
+  workflow_dispatch:
+
 # Prevent writing to the repository using the CI token.
 # Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
 permissions: read-all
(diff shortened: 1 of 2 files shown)


doriable (approved):

.github/workflows/windows.yaml
```diff
@@ -19,9 +22,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: |
-            C:\Users\runneradmin\AppData\Local\Temp\chocolatey
-            C:\Users\runneradmin\go\pkg\mod
-          key: ${{ runner.os }}-buf-windows-${{ hashFiles('**/go.sum', '/windows/**') }}
+            ${{ env.GOCACHE }}
+            ${{ env.GOMODCACHE }}
+          key: ${{ runner.os }}-buf-windows-${{ hashFiles('**/go.sum', '.github/workflows/windows.yaml') }}
```
bufdev, line 27: I think the key needs to include the `windows` directory

PR #820, opened 2021-12-23
Add windows/** to the cache key

WillAbides:
Follow up to #819

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -24,7 +24,7 @@ jobs:
           path: |
             ${{ env.GOCACHE }}
             ${{ env.GOMODCACHE }}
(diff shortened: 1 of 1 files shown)


doriable (approved):

.github/workflows/windows.yaml
```diff
@@ -24,7 +24,7 @@ jobs:
           path: |
             ${{ env.GOCACHE }}
             ${{ env.GOMODCACHE }}
-          key: ${{ runner.os }}-buf-windows-${{ hashFiles('**/go.sum', '.github/workflows/windows.yaml') }}
+          key: ${{ runner.os }}-buf-windows-${{ hashFiles('**/go.sum', 'windows/**') }}
```
bufdev, line 27: Our other GitHub Actions workflow files have a leading `/`, so we should be consistent.

PR #822, opened 2021-12-29
CI caching improvements

WillAbides:
Follow up to #819 and #820

This makes a few caching changes speed up ci.

- Removes `/` prefix from paths in `hashFiles` calls so that caches are invalidated on Makefile changes.
- Gives lint and test their own caches. This speeds up lint by letting it cache the dependencies it has separate from test.
- Caches `protoc.exe` on Windows.

In addition it swaps `windows/test.ps1` for `windows/test.bash`. I did this because it was quicker than learning enough PowerShell to make these changes, and I think there are more people working on this repo who are be able to maintain a bash script.


windows/test.bash
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+PROTOC_VERSION="3.19.1"
+PROTOC_GEN_GO_VERSION="v1.27.1"
+PROTOC_GEN_GO_GRPC_VERSION="30dfb4b933a50fd366d7ed36ed4f71dbba2d382e"
```
doriable, line 7: Moving to bash seems reasonable given that this seems to running well in the Windows CI environment.
Just a small note here to update the documentation for upgrading protoc for this, but otherwise I'm on board with this approach.

doriable (approved):
Nice clean-up 👏 tyty!

windows/test.bash
WillAbides: Thanks for the reminder. https://github.com/bufbuild/private-docs/pull/63 will update the doc

windows/test.bash
```diff
@@ -0,0 +1,35 @@
...
+  echo "Using cached protoc"
+fi
+
+PATH="${DOWNLOAD_CACHE}/protoc/bin:${PATH}"
+
+go install google.golang.org/protobuf/cmd/protoc-gen-go@${PROTOC_GEN_GO_VERSION}
+go install google.golang.org/grpc/cmd/protoc-gen-go-grpc@${PROTOC_GEN_GO_GRPC_VERSION}
+go install ./private/buf/cmd/buf/command/protoc/internal/protoc-gen-insertion-point-writer
+go install ./private/buf/cmd/buf/command/protoc/internal/protoc-gen-insertion-point-receiver
+go install ./cmd/buf
```
amckinney, line 34: Should this be one `go install` command with each parameter separated by `\` on newlines?

WillAbides: Thanks, I didn't know `go install` worked with multiple packages.

I combined the three local packages into one command, but that trick apparently doesn't work when you specify different versions for the modules.

```shell
$ go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.27.1 \
  google.golang.org/grpc/cmd/protoc-gen-go-grpc@30dfb4b933a50fd366d7ed36ed4f71dbba2d382e
go install: google.golang.org/grpc/cmd/protoc-gen-go-grpc@30dfb4b933a50fd366d7ed36ed4f71dbba2d382e: all arguments must have the same version (@v1.27.1)
```

amckinney: Ah got it, thanks for sharing that.

amckinney (approved):
Looks good to me, thanks for taking a closer look here!

PR #827, opened 2022-01-04
plugin and template deprecation endpoints

cyinma:
this is to add new fields and endpoints for plugin and template deprecation.


proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -171,6 +179,10 @@ service PluginService {
   // This does not include users who have implicit roles against the plugin, unless they have also been
   // assigned a role explicitly.
   rpc ListPluginContributors(ListPluginContributorsRequest) returns (ListPluginContributorsResponse);
+  // DeprecatePluginByName deprecates the plugin, if found.
+  rpc DeprecatePluginByName(DeprecatePluginByNameRequest) returns (DeprecatePluginByNameResponse);
```
johanbrandhorst, line 183: I don't think we need the `ByName` on this service, everything uses the name as a lookup key. I appreciate it's consistent with repositories, but in that service we have both ID and Name functions, which it's why it makes more sense there. What do you think?

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -324,6 +340,27 @@ message ListPluginContributorsResponse {
   string next_page_token = 2;
 }
 
+message DeprecatePluginByNameRequest {
+  // The owner of the plugin.
+  string owner = 1;
+  // The name of the plugin.
+  string name = 2;
+  // A message shown along with the deprecation warning for this plugin.
+  // It must be a utf8 encoded string containing 256 or fewer characters.
```
johanbrandhorst, line 349: We don't include the validation constraints in any other comments in this file, so we should probably remove them from here?

proto/buf/alpha/registry/v1alpha1/plugin.proto
cyinma: make sense to me 👍

johanbrandhorst (approved):

Issue #798, opened 2021-12-17
trailing comment parsing different from protoc


eyalpost:
It seems that the definition of trailing comment is different between buf and protoc when it comes to elements that are defined with `{}` (curly braces)

For example, for the following enum definition:

```
enum Enum { // Comment 1

  VALUE = 0;

} // Comment 2
```

protoc will treat `Comment 1` as the trailing comment for `Enum`
while buf will treat `Comment 2` as the trailing comment

amckinney:
Thanks for the issue, we'll take a closer look and see if we can get a fix out soon.

bufdev:
This is fixed and will go out in the next release.

PR #828, opened 2022-01-05
Test that trailing comments match between protoc and buf

bufdev:
Fixes https://github.com/bufbuild/buf/issues/798.

https://github.com/jhump/protoreflect/pull/475 must be merged first and then the Makefile updated.


WillAbides (approved):
LGTM

PR #830, opened 2022-01-06
testingextended: helper to get test timeout value

lrewega:
It's useful to set go test's -timeout flag in different contexts, such
as during integration tests or when using verbose debugging or running
in a resource-constrained environment. Keeping such scenarios in mind,
it's additionally often useful to avoid internal timeouts during certain
tests by specifying a larger timeout duration. Unfortunately, this is
prone to error when the `go test` timeout exceeds the internal timeout,
which may not have been the intent of the test author, resulting in a
new, distinct and surprising failure.

Let's teach testingextended to expose `testing`'s global `timeout` flag
value so tests can refer to a duration that is guaranteed to exceed the
deadline of the test.

### DIFF
--- a/private/pkg/testingextended/testingextended.go
+++ b/private/pkg/testingextended/testingextended.go
@@ -34,5 +34,5 @@ func GetTestTimeout(t *testing.T) time.Duration {
 	}
 	// It's fine if this panics. We expect to be in a test, and this should
 	// be covered by the Go 1 compatability promise.
-	return flag.Lookup("timeout").Value.(testing.Getter).Get().(time.Duration)
+	return flag.Lookup("timeout").Value.(flag.Getter).Get().(time.Duration)
 }
--- a/private/pkg/testingextended/testingextended.go
+++ b/private/pkg/testingextended/testingextended.go
@@ -27,12 +27,12 @@ func SkipIfShort(t *testing.T) {
 	}
 }
 
-//GetTestTimeout returns the value of the go test -timeout flag.
+// GetTestTimeout returns the value of the go test -timeout flag.
 func GetTestTimeout(t *testing.T) time.Duration {
 	if !flag.Parsed() {
 		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
 	}
 	// It's fine if this panics. We expect to be in a test, and this should
-	// be covered by the Go 1 compatability promise.
+	// be covered by the Go 1 compatibility promise.
 	return flag.Lookup("test.timeout").Value.(flag.Getter).Get().(time.Duration)
 }
--- a/private/pkg/testingextended/testingextended.go
+++ b/private/pkg/testingextended/testingextended.go
@@ -14,11 +14,25 @@
 
 package testingextended
 
-import "testing"
+import (
+	"flag"
+	"testing"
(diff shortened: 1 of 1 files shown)


private/pkg/testingextended/testingextended.go
```diff
@@ -14,11 +14,25 @@
...
 }
+
+//GetTestTimeout returns the value of the go test -timeout flag.
+func GetTestTimeout(t *testing.T) time.Duration {
+	if !flag.Parsed() {
+		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
+	}
+	// It's fine if this panics. We expect to be in a test, and this should
+	// be covered by the Go 1 compatability promise.
+	return flag.Lookup("timeout").Value.(testing.Getter).Get().(time.Duration)
```
lrewega, line 37:
```suggestion
	return flag.Lookup("timeout").Value.(flag.Getter).Get().(time.Duration)
```

private/pkg/testingextended/testingextended.go
```diff
@@ -14,11 +14,25 @@
...
 }
+
+//GetTestTimeout returns the value of the go test -timeout flag.
+func GetTestTimeout(t *testing.T) time.Duration {
+	if !flag.Parsed() {
+		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
+	}
+	// It's fine if this panics. We expect to be in a test, and this should
+	// be covered by the Go 1 compatability promise.
+	return flag.Lookup("timeout").Value.(flag.Getter).Get().(time.Duration)
```
lrewega, line 37:
```suggestion
	return flag.Lookup("test.timeout").Value.(flag.Getter).Get().(time.Duration)
```

lrewega:
See also https://github.com/golang/go/issues/34707#issuecomment-539019076

johanbrandhorst (approved):
I like it

private/pkg/testingextended/testingextended.go
```diff
@@ -14,11 +14,25 @@
...
+)
 
 // SkipIfShort skips the test if testing.short is set.
 func SkipIfShort(t *testing.T) {
 	if testing.Short() {
 		t.Skip("skipping test in short mode")
 	}
 }
+
+//GetTestTimeout returns the value of the go test -timeout flag.
```
bufdev, line 30: space between // and GetTestTimeout

private/pkg/testingextended/testingextended.go
```diff
@@ -14,11 +14,25 @@
...
+func GetTestTimeout(t *testing.T) time.Duration {
+	if !flag.Parsed() {
+		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
+	}
+	// It's fine if this panics. We expect to be in a test, and this should
+	// be covered by the Go 1 compatability promise.
```
lrewega, line 36:
```suggestion
	// be covered by the Go 1 compatibility promise.
```

private/pkg/testingextended/testingextended.go
lrewega, line 30:
```suggestion
// GetTestTimeout returns the value of the go test -timeout flag.
```

PR #834, opened 2022-01-06
Add option to omit tags when listing repository commits

WillAbides:
This adds an `omit_tags` field to `ListRepositoryCommitsByReferenceRequest` and `ListRepositoryCommitsByBranchRequest`. When set the server won't populate `RepositoryCommit.tags` in the response, which will make the response quite a bit speedier.


private/buf/cmd/buf/command/beta/registry/commit/commitlist/commitlist.go
```diff
@@ -124,6 +124,7 @@ func run(
 		flags.PageSize,
 		flags.PageToken,
 		flags.Reverse,
+		false,
```
WillAbides, line 127: We should set this to true some time before `v1.0.0`, but I don't want to add that decision to the scope of this PR.

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -68,6 +68,8 @@ message ListRepositoryCommitsByBranchRequest {
   uint32 page_size = 4;
   string page_token = 5;
   bool reverse = 6;
+  // Allow the server to omit tags from the response.
```
bufdev, line 71: It feels like we'd want a general-purpose solution to this across our API calls, as opposed to a one-off. Is there a performance issue we're having that necessitates this?

PR #836, opened 2022-01-07
CLI support for plugin and template deprecation

cyinma:
will also include warning message when user is using deprecated plugin/template in this PR

### Plugin

To deprecate a plugin: 
`buf beta registry plugin deprecate buf.build/owner/plugins/plugin [--message "please use buf.build/owner/plugins/newplugin"]`

To undeprecate a plugin:
`buf beta registry plugin undeprecate buf.build/owner/plugins/plugin`

When you run `buf generate` with a deprecated remote plugin in your `buf.gen.yaml`, you will get a warning message:
`WARN    Plugin "buf.build/owner/plugins/plugin" is deprecated: please use buf.build/owner/plugins/newplugin`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Upgrade to `protoc` 3.19.2 support.
+- Add `buf beta registry plugin deprecate`, `buf beta registry plugin undeprecate`, `buf beta registry template deprecate` and `buf beta registry template undeprecate`.
 
 ## [v1.0.0-rc10] - 2021-12-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 - Upgrade to `protoc` 3.19.2 support.
-- Add `buf beta registry plugin deprecate`, `buf beta registry plugin undeprecate`, `buf beta registry template deprecate` and `buf beta registry template undeprecate`.
+- Add `buf beta registry {plugin,template} {deprecate,undeprecate}`
 
 ## [v1.0.0-rc10] - 2021-12-16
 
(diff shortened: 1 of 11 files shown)


private/buf/bufgen/generator.go
```diff
@@ -294,6 +294,21 @@ func (g *generator) execRemotePlugin(
...
+	if err != nil {
+		return nil, err
+	}
+	if plugin.Deprecated {
+		warnMsg := fmt.Sprintf(`Plugin "%s/%s/%s" is deprecated`, remote, owner, name)
+		if plugin.DeprecationMessage != "" {
+			warnMsg = fmt.Sprintf("%s: %s", warnMsg, plugin.DeprecationMessage)
+		}
+		g.logger.Sugar().Warn(warnMsg)
+	}
```
doriable, line 311: This is happening after the plugin has been used for the generate service... that is probably fine since we still want the result of the generation, it's just that we want to check whether or not the plugin is deprecated? The warning message looks good though, but @amckinney can confirm :)

doriable:
Oh, small note -- since this is adding functionality to the CLI, we should add a line item on this in the `CHANGELOG`.

private/buf/bufgen/generator.go
amckinney: Yea, we'll definitely still want to use the generation result. This looks good, but it's worth noting that it adds an additional RPC for _every_ remote plugin the user has configured in their `buf.gen.yaml`.

Fortunately all of these requests are concurrently run in the same goroutine that executes the plugin in question. It's OK to start with this - if we ever have too many RPCs, we can batch them together (just like we used to have with each plugin execution, too).

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Upgrade to `protoc` 3.19.2 support.
+- Add `buf beta registry plugin deprecate`, `buf beta registry plugin undeprecate`, `buf beta registry template deprecate` and `buf beta registry template undeprecate`.
```
amckinney, line 6: Fine as-is so don't feel obligated to change it, but we can make this message a little shorter if we wanted.
```suggestion
- Add `buf beta registry {plugin,template} {deprecate,undeprecate}`
```

amckinney (approved):
Well done, this looks great 👏

PR #839, opened 2022-01-07
Check dependencies when inferring default remote

robbertvanginkel:
When users of the enterprise BSR have a reference to a private dependency but don't specify a `name:` in their module:
```
version: v1
deps:
    - customer.buf.team/owner/repo
```
, `buf` will use the default remote (`buf.build`) to attempt to resolve their dependencies. This will lead to an error like:
```
$ buf mod update
Failure: invalid remote "customer.buf.team", did you mean "buf.build" or "buf.build"?
```

This PR attempts to print out a more useful error message

```
$ buf mod update
WARN	"buf.yaml" does not specify a "name", buf will default to using remote "buf.build" for dependency resolution. This remote may not be able to resolve "customer.buf.team/owner/repo" if it is a enterprise BSR module, did you mean to specify a "name: customer.buf.team/..." on this module?
Failure: invalid remote "customer.buf.team", did you mean "buf.build" or "buf.build"?
```


johanbrandhorst (approved):
Nice one 👍🏻

private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -113,6 +113,20 @@ func run(
 	remote := bufrpc.DefaultRemote
 	if moduleConfig.ModuleIdentity != nil && moduleConfig.ModuleIdentity.Remote() != "" {
 		remote = moduleConfig.ModuleIdentity.Remote()
+	} else {
+		for _, moduleReference := range moduleConfig.Build.DependencyModuleReferences {
+			if moduleReference.Remote() != bufrpc.DefaultRemote {
+				warnMsg := fmt.Sprintf(
```
robbertvanginkel, line 119: Perhaps we should restrict this warning to `buf.teams` for now? Otherwise this'll also show if people add a `github.com/...` dep and suggest to add github to `name:`.

robbertvanginkel (commented):
@doriable added you as reviewer too mainly because I think you suggested a different location to put the warning on https://github.com/bufbuild/core/issues/3412#issuecomment-1004969999. Also lmk if this warning text seems style coherent with the rest of the cli.

private/buf/cmd/buf/command/mod/modupdate/modupdate.go
doriable: Yeah, I think that might makes sense... I can see how this could be confusing to a new user of `buf` thinking that we can support VCS remotes...

doriable:
> @doriable added you as reviewer too mainly because I think you suggested a different location to put the warning

Yeah, so I guess it depends on how we want to handle the warning. If we feel that this is scoped solely to `mod update`, this location seems fine. Otherwise, yeah, I suggested putting it at module compile time so that we can catch that before any specific commands are run, but it would come up on _any_ command run against a module.

The messaging looks good/consistent :D

amckinney (approved):
Cool, this warning should help. The summary shows the following error message -

```
Failure: invalid remote "customer.buf.team", did you mean "buf.build" or "buf.build"?
```

It doesn't look like that stutter is being addressed here though - is that intentional?

amckinney:
It might be good to add a `CHANGELOG.md` entry, too.

robbertvanginkel:
> It doesn't look like that stutter is being addressed here though - is that intentional?

This string comes from the server, and has already been addressed internally (pr 3455)

robbertvanginkel:
> If we feel that this is scoped solely to mod update, this location seems fine. Otherwise, yeah, I suggested putting it at module compile time so that we can catch that before any specific commands are run, but it would come up on any command run against a module.

I think it should be fine to put this only on mod update, as this is the only command where it makes sense to do something about it. If you deps are out of sync with the lock file (or you dont have one) other commands like buf build will tell you that you should run mod update already.

Issue #777, opened 2021-12-10
Simplify `buf.lock`


amckinney:
# buf.lock

The `buf.lock` currently contains some information that isn't really needed. For example, the `branch` is being replaced with the new concept of `track`, and a commit can actually be associated with >=0 tracks at any given time.

The current structure looks like the following:

```yaml
# Generated by buf. DO NOT EDIT.
version: v1
deps:
  - remote: buf.build
    owner: googleapis
    repository: googleapis
    branch: main
    commit: 84c3cad756d2435982d9e3b72680fa96
    digest: b1-GVZocI28gwtZCKNAO4Jd83jaE_XMjcAuEBtjnWvybO0=
    create_time: 2021-08-17T15:06:58.03881Z
```

We can simplify this all the way down to just the `remote`, `owner`, `repository`, and `commit`:

```yaml
# Generated by buf. DO NOT EDIT.
version: v1
deps:
  - remote: buf.build
    owner: googleapis
    repository: googleapis
    commit: 84c3cad756d2435982d9e3b72680fa96
```

We _could_ consider consolidating the `remote`, `owner`, and `repository` keys into a `name` key for terseness, but that might not buy us much - it's just slightly more visually appealing.

```yaml
# Generated by buf. DO NOT EDIT.
version: v1
deps:
  - name: buf.build/googleapis/googleapis
    commit: 84c3cad756d2435982d9e3b72680fa96
```

Regardless of what representation we go with, we should be able to keep using `v1` because we'd be using a new key (in the latter case) or simply omitting other ones that were previously used.

# Module Pins

## Client-side

In their current state, the `bufmodule.ModulePin` has a 1:1 association with the entries rendered in the `buf.lock`. If we make the changes to the `buf.lock` above, we might have an opportunity to simplify the `bufmodule.ModulePin`, too.

If we remove fields from the [buf.alpha.module.v1alpha1.ModulePin](https://github.com/bufbuild/buf/blob/7e66668a878c4fcd4b71beac727b7a344e8aa1bb/proto/buf/alpha/module/v1alpha1/module.proto#L64) (the Protobuf representation), then this is clearly a breaking change.

Unfortunately, even on the latest version of `buf`, it looks like we would have a validation error if _any_ of the values are missing in [ValidateProtoModulePin](https://github.com/bufbuild/buf/blob/68d587215467ad43db41e4440c19d1f4c811e8de/private/bufpkg/bufmodule/bufmoduleref/validate.go#L49).

[...]

PR #843, opened 2022-01-11
Remove extraneous validation

doriable:
In preparation for #777, this removes extraneous validation methods that we no longer require on the client and server-side. We no longer want to store this information (and some of it will be stale soon, e.g. `branch`), so we need to make sure we are not validating against values we will not be storing in the future.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleref/validate.go
+++ b/private/bufpkg/bufmodule/bufmoduleref/validate.go
@@ -17,11 +17,9 @@ package bufmoduleref
 import (
 	"errors"
 	"fmt"
-	"strings"
 
 	modulev1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/module/v1alpha1"
 	"github.com/bufbuild/buf/private/pkg/netextended"
-	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 // ValidateProtoModuleReference verifies the given module reference is well-formed.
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmoduleref/validate.go
```diff
@@ -59,18 +57,9 @@ func ValidateProtoModulePin(protoModulePin *modulev1alpha1.ModulePin) error {
 	if err := ValidateRepository(protoModulePin.Repository); err != nil {
 		return err
 	}
-	if err := ValidateBranch(protoModulePin.Branch); err != nil {
```
doriable, line 62: I didn't remove the function because there are currently some server-side use-cases I am currently unsure of and wanted to discuss with @WillAbides in relation to his work on tracks.

amckinney: Sounds good.

For the sake of completeness here, is it possible to remove _all_ client-side validation? It looks like we're still including `ValidateProtoModulePin` as a whole, so we would still be validating against remotes, owners, repositories, etc. 

If we ever decided to drop any of these fields from the ModulePin, then old CLIs would break. I know it's unlikely since these fields feel the most minimal, but I'd challenge us to see how much validation we can remove (without comprising basic functionality) to prevent these types of failures later.

amckinney (approved):
This is a good start (we want to remove this validation no matter what), but let's make sure to take a closer look to see if we can remove even more in a follow-up.

PR #847, opened 2022-01-12
Remove `digest`, and `created_at` fields from the `buf.lock`

doriable:
This PR removes the writing of `branch`, `digest`, and `created_at` fields into the `buf.lock`.
This also means that we will not be reading those values out of future `buf.lock`, so those values will be the `nil`/zero values for the module pin sent to the server-side. This means, that if the server-side is still on the `b1` digest (which it currently is), a new commit will be written with a client that includes this change. If the server-side digest is migrated to `b3`, then this change does not present any issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
 - Add `buf beta registry {plugin,template} {deprecate,undeprecate}`.
 - Add warning when using enterprise dependencies without specifying a enterprise
   remote in the module's identity.
+- Remove `branch`, `digest`, and `created_at` fields from the `buf.lock`. This will create
+  when pushing the same contents to an existing repository, since the `ModulePin` has been reduced down.
 
 ## [v1.0.0-rc10] - 2021-12-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - Add `buf beta registry {plugin,template} {deprecate,undeprecate}`.
 - Add warning when using enterprise dependencies without specifying a enterprise
   remote in the module's identity.
-- Remove `branch`, `digest`, and `created_at` fields from the `buf.lock`. This will create
+- Remove `branch`, `digest`, and `created_at` fields from the `buf.lock`. This will temporarily create a new commit
   when pushing the same contents to an existing repository, since the `ModulePin` has been reduced down.
 
 ## [v1.0.0-rc10] - 2021-12-16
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
 - Add `buf beta registry {plugin,template} {deprecate,undeprecate}`.
 - Add warning when using enterprise dependencies without specifying a enterprise
   remote in the module's identity.
+- Remove `branch`, `digest`, and `created_at` fields from the `buf.lock`. This will create
```
amckinney, line 10:
```suggestion
- Remove `branch`, `digest`, and `created_at` fields from the `buf.lock`. This will temporarily create a new commit
```

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -416,10 +416,10 @@ func DependencyModulePinsForBucket(
 			dep.Remote,
 			dep.Owner,
 			dep.Repository,
-			dep.Branch,
+			"",
```
amckinney, line 419: Does the `NewModulePin` constructor still need to expose these parameters for other call-sites? Is this because we still have these fields represented in the Protobuf representation of `ModulePin`? Will they be removed later, or kept around forever?

doriable: Yes, it's because they are still being represented in the Proto definition. And yeah, I think we will remove them when we are confident that all clients are upgraded (which is reasonable, since we'll require everyone to move to `v1` when we cut over the digest, since old clients cannot read new commits that have been pushed with the `b3` digest). So as a final clean-up after all the upgrades, I plan to cull down the Proto representation and handle it for both client and server.

amckinney: Make sense, especially because the constructor is used on the server-side.

We use the `CreateTime` attached to the `ModulePin` in our dependency resolution, so we'll need to make sure that's still possible with these changes.

amckinney (approved):

doriable:
Add the `branch` field back to the `buf.lock` for 2 reasons:
1. We want to be able to show users which track/branch they are depending on through their `buf.lock`. We also want to store `branch` for that reason for the module pin resolution -- if we change branch/tracks, we need to be able to resolve that change.
2. Old clients are also unable to read new module pins without the branch value, and so keeping the structure in tact, especially combine with reason #1 makes sense.

amckinney (approved):

doriable:
We plan on still removing the `branch` field once the resolve service has been adjusted accordingly to support tracks (or at least default to returning the `main` branch). The issue is that old clients are still validating against `branch`, so we'll need to return something for this in the `ModulePin`. The server-side will need to be reworked to return a valid `branch` value in the interim, so `branch` will be adjusted in a follow-up PR.

PR #848, opened 2022-01-12
Add GetReferenceAncestorSequenceIdOnRepositoryTrack

WillAbides:
This will be needed by go module proxy to generate pseudo-versions for commits on non-main tracks.


proto/buf/alpha/registry/v1alpha1/repository_track.proto
```diff
@@ -72,3 +75,14 @@ message DeleteRepositoryTrackByNameRequest {
 }
 
 message DeleteRepositoryTrackByNameResponse {}
+
+message GetReferenceAncestorSequenceIdOnRepositoryTrackRequest {
+  string owner_name = 1;
+  string repository_name = 2;
+  string reference = 3;
+  string repository_track_name = 4;
```
amckinney, line 83: The `repository_track_name` is a `reference`, no? Why do we need both a `repository_track_name` and a `reference`? Is `reference` supposed to be `commit`?

WillAbides: This finds the first ancestor of `reference` that is on the track `repository_track_name`.

bufdev:
Is this stale?

WillAbides:
This probably won't be needed. It's easy enough to recreate if we need it.

PR #849, opened 2022-01-12
Add buf push --track

WillAbides:
This adds a `--track` option to `buf push`. When set, the module will be appended to the track(s) specified. Otherwise the module will be appended to the main track.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
+++ b/private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
@@ -28,9 +28,13 @@ import (
 )
 
 const (
+	// MainTrack is the name of the track created for every repository.
+	// This is the default track used if no track is specified.
+	MainTrack = "main"
+
 	// MainBranch is the name of the branch created for every repository.
 	// This is the default branch used if no branch or commit is specified.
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -28,9 +28,13 @@ import (
...
 
 const (
+	// MainTrack is the name of the track created for every repository.
+	// This is the default track used if no track is specified.
+	MainTrack = "main"
+
 	// MainBranch is the name of the branch created for every repository.
 	// This is the default branch used if no branch or commit is specified.
-	MainBranch = "main"
+	MainBranch = MainTrack
```
bufdev, line 37: Separate from this PR, but why do we still need `MainBranch`?

bufdev (approved):

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
WillAbides: There is some dead branch code that uses it both in this repository and core. MainBranch won't be around for much longer.

PR #852, opened 2022-01-13
Keep pushing to main branch

WillAbides:
Push still needs to set branch to "main" for now because of server-side validations.


doriable:
Hmm, I thought we removed the validation on the server-side... is it because of this?

https://github.com/bufbuild/core/blob/e17f5926eac72bc038f8a7fbf5463a0a7c063979/internal/bufd/bufdapiserver/internal/bufdvalidate/push_service.go#L44

and if so, should we be removing this and/or validating a track instead?

WillAbides:
nevermind

PR #853, opened 2022-01-13
Add more support for tracks

WillAbides:
- Update `buf beta registry commit list` to allow a track to be specified.
- Add `buf beta registry track {list,delete}` commands.

Also add `buf push --track` to CHANGELOG.md because I forgot to earlier.

### DIFF
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -49,6 +49,7 @@ import (
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/beta/registry/template/templateundeprecate"
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/beta/registry/template/templateversion/templateversioncreate"
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/beta/registry/template/templateversion/templateversionlist"
+	"github.com/bufbuild/buf/private/buf/cmd/buf/command/beta/registry/track/tracklist"
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/breaking"
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/build"
 	"github.com/bufbuild/buf/private/buf/cmd/buf/command/config/configinit"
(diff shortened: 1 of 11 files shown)


private/buf/cmd/buf/buf.go
```diff
@@ -180,14 +182,14 @@ func NewRootCommand(name string) *appcmd.Command {
...
-							//		branchcreate.NewCommand("create", builder),
-							//		branchlist.NewCommand("list", builder),
-							//	},
-							//},
+							{
+								Use:   "track",
+								Short: "Repository track commands.",
+								SubCommands: []*appcmd.Command{
+									tracklist.NewCommand("list", builder),
+									trackdelete.NewCommand("delete", builder),
```
amckinney, line 190: Will we ever need a `create` command (would be in a follow-up, if ever)? Is it possible to create a track without a commit (i.e. instead of using the `buf push --track` command)? Or is that not something we plan to support?

private/buf/cmd/buf/command/beta/registry/commit/commitlist/commitlist.go
```diff
@@ -42,8 +42,7 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		//Use:   name + " <buf.build/owner/repo:branch>",
-		Use:   name + " <buf.build/owner/repo>",
+		Use:   name + " <buf.build/owner/repo[:track]>",
```
amckinney, line 45: Does this technically support tags? I know it would only ever list a single commit, but we might want this to be `[:reference]` if tags are supported by the `ListRepositoryCommitsByReference` endoint.

amckinney (approved):
Looks good, minus a nit in the `--help` text.

private/buf/cmd/buf/buf.go
WillAbides: When we met about this @bufdev said there should be just one way to create a track, and I agree. It shouldn't be hard to add later if we find a compelling use case for it. As you pointed out, we will need to think through how to handle tracks with zero commits.

private/buf/cmd/buf/command/beta/registry/commit/commitlist/commitlist.go
WillAbides: Yes it does technically support tags (also commit names). I thought about making this [:ref] to match `... commit get`, but went with track because that's the only one where it makes sense to use list instead of get. On second thought I think I do like [:ref] better. Let users decide what makes more sense to use.

PR #855, opened 2022-01-14
Adds image digest to plugin version

mfridman:
This PR will enable us to print the docker image digest of the plugin version.

Example output (sha is made up here):

```
Name       Plugin Name  Plugin Owner  Digest
v1.41.0-1  java         grpc          sha256:1d3fa9292790d8ec2e199abcc510c86c48948172433df0905260986c098e5f00
```

An argument could be made the digest should be opt-in with a flag (similar to the `docker images --digests` command).


proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -48,6 +48,8 @@ message PluginVersion {
   string plugin_name = 3;
   // The owner of the plugin to which this version relates.
   string plugin_owner = 4;
+  // The full docker image digest of this plugin version.
+  string digest = 5;
```
bufdev, line 52: I have a minor level of concern that just calling this `digest` overlaps with our concept of digest for modules, perhaps we should clarify this is a docker digest, I'm not sure.

mfridman: Would either of `ImageDigest` or ManifestDigest` overlap less with our existing nomenclature?

I think I prefer `ManifestDigest` because it's not a term we've used (afaics)

mfridman: Updated to `image_digest`. I think most folks are used to the term "image" when it comes to docker images so it's slightly more familiar as compared to `manifest_digest`.

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -48,6 +48,8 @@ message PluginVersion {
   string plugin_name = 3;
   // The owner of the plugin to which this version relates.
   string plugin_owner = 4;
+  // The full docker image digest of this plugin version.
+  string image_digest = 5;
```
bufdev, line 52: `image` has a specific meaning in Buf, this is more confusing for me - I think it's worth talking about with the team

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -48,6 +48,8 @@ message PluginVersion {
   string plugin_name = 3;
   // The owner of the plugin to which this version relates.
   string plugin_owner = 4;
+  // The full docker image digest associated with this plugin version.
```
johanbrandhorst, line 51: Can you include a reference to the format of this string? It's probably somewhere in the Docker docs or OCI image spec.

mfridman: Added a link to https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests and clarified the comment.

johanbrandhorst (approved):

PR #857, opened 2022-01-14
Set a default timeout in GetTestTimeout

WillAbides:
This updates `GetTestTimeout` to return the time remaining until `t.Deadline()`. When there is no deadline it return 10 minutes.

This is needed because when running tests with goland, no `-test.timeout` is set by default.

### DIFF
--- a/private/pkg/testingextended/testingextended.go
+++ b/private/pkg/testingextended/testingextended.go
@@ -15,7 +15,6 @@
 package testingextended
 
 import (
-	"flag"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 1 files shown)


private/pkg/testingextended/testingextended.go
```diff
@@ -27,12 +26,10 @@ func SkipIfShort(t *testing.T) {
 	}
 }
 
-// GetTestTimeout returns the value of the go test -timeout flag.
+// GetTestTimeout returns the time remaining until the test times out or 10m if the test is not set to timeout.
 func GetTestTimeout(t *testing.T) time.Duration {
-	if !flag.Parsed() {
-		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
+	if deadline, ok := t.Deadline(); ok {
```
mfridman, line 31: Should we also `!deadline.IsZero()` ... is that even possible?

WillAbides: I don't think it's possible, but it doesn't hurt anything to add

lrewega: `The ok result is false if the -timeout flag indicates “no timeout” (0).`

mfridman (approved):

private/pkg/testingextended/testingextended.go
```diff
@@ -27,12 +26,10 @@ func SkipIfShort(t *testing.T) {
...
 func GetTestTimeout(t *testing.T) time.Duration {
-	if !flag.Parsed() {
-		t.Fatal("unable to read testing timeout flag as flags have not been parsed")
+	if deadline, ok := t.Deadline(); ok && !deadline.IsZero() {
+		return time.Until(deadline)
 	}
-	// It's fine if this panics. We expect to be in a test, and this should
-	// be covered by the Go 1 compatibility promise.
-	return flag.Lookup("test.timeout").Value.(flag.Getter).Get().(time.Duration)
+	return 10 * time.Minute
```
lrewega, line 34: This makes me a *bit* nervous, but I can't think of anything obvious that could go wrong.

PR #871, opened 2022-01-20
add force delete user endpoint

cyinma:
This is to add a new endpoint to force delete a user including the organization and resources they have.


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -62,6 +62,9 @@ service UserService {
   rpc ListOrganizationUsers(ListOrganizationUsersRequest) returns (ListOrganizationUsersResponse);
   // DeleteUser deletes a user.
   rpc DeleteUser(DeleteUserRequest) returns (DeleteUserResponse);
+  // ForceDeleteUser forces to delete a user. Resources and organizations that is
```
johanbrandhorst, line 65: Nit: shouldn't this be
```suggestion
  // ForceDeleteUser forces to delete a user. Resources and organizations that are
```

johanbrandhorst (approved):

Issue #876, opened 2022-01-21
Add `except` and `override` to `java_package_prefix`


amckinney:
Re: https://github.com/bufbuild/buf/issues/831#issuecomment-1016301391

For anyone depending on the [GoogleAPIs Maven repository](https://mvnrepository.com/artifact/com.google.api), it's important that the `java_package` value is set to `com.google.api` (as shown in the module [here](https://buf.build/googleapis/googleapis/file/main/google/api/annotations.proto#L25)). This is similar to the `go_package` exception we have too, which is described in further detail in [the tour](https://docs.buf.build/tour/use-managed-mode#remove-modules-from-managed-mode).

We should add an `override` and `except` option to the `java_package_prefix` configuration (similar to what we expose for `go_package_prefix`). We'll probably need to use a custom `UnmarshalYAML` method (i.e. `yaml.Unmarshaler` implementation) so that we can preserve backwards compatibility with the old `java_package_prefix: <string>` configuration.

amckinney:
For some valuable notes on customizing the `yaml.Unmarshaler`, see [this](https://abhinavg.net/posts/flexible-yaml/) blog post.

amckinney:
@JacobSMoller This will go out in the next release (`v1.0.0-rc12`).

JacobSMoller:
@amckinney  thank you for the update

And thanks a lot for an awesome tool.

PR #879, opened 2022-01-21
add except and override to java_package_prefix

srikrsna-buf:
#876

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `except` and `override` to `java_package_prefix`.
 
 ## [v1.0.0-rc11] - 2022-01-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Add `except` and `override` to `java_package_prefix`.
+- Add `default`, `except` and `override` to `java_package_prefix`.
 
 ## [v1.0.0-rc11] - 2022-01-18
 
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -306,8 +306,23 @@ func (e ExternalJavaPackagePrefixConfigV1) IsEmpty() bool {
 }
 
 // UnmarshalYAML satisfies the yaml.Unmarshaler interface. This is done to maintain backward compatibility
-// of accepting a plain string value for java_package_prefix
+// of accepting a plain string value for java_package_prefix.
 func (e *ExternalJavaPackagePrefixConfigV1) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	return e.unmarshalWith(unmarshal)
+}
+
+// UnmarshalJSON satisfies the json.Unmarshaler interface. This is done to maintain backward compatibility
+// of accepting a plain string value for java_package_prefix.
+func (e *ExternalJavaPackagePrefixConfigV1) UnmarshalJSON(data []byte) error {
+	unmarshal := func(v interface{}) error {
+		return json.Unmarshal(data, v)
+	}
+
+	return e.unmarshalWith(unmarshal)
+}
+
+// unmarshalWith is used to unmarshal into json/yaml. See https://abhinavg.net/posts/flexible-yaml for details.
+func (e *ExternalJavaPackagePrefixConfigV1) unmarshalWith(unmarshal func(interface{}) error) error {
 	var prefix string
 	if err := unmarshal(&prefix); err == nil {
 		e.Default = prefix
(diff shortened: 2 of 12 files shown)


doriable (commented):
I've given this an initial once-over, tests look good and nice job on the unmarshalling for maintaining backwards compatibility. I'm just QA-ing and then I'll do a second sweep :) Also, I want a better understanding of the `make generate` changes and ensure there isn't a bug that is causing that.

doriable:
Just ran it in my a playground module, one of my files contains the following:

```
🐱~/src/playground/a @ $ cat foo.proto
syntax = "proto3";

package a;
import "google/type/datetime.proto";

// Change the types of these fields (breaking change)
message HelloWorld {
  string hello = 1;
  string world = 2;
  google.type.DateTime create_time = 3;
  string blah = 4;
}
```

```
🐱~/src/playground/a @ $ cat buf.gen.yaml
version: v1
managed:
  enabled: true
  java_package_prefix: com.dkeung.playground
plugins:
  - name: java
    out: pkg/server/grpc
```

Output:

```
...
        new java.lang.String[] { "Hello", "World", "CreateTime", "Blah", });
    com.dkeung.playground.google.type.DatetimeProto.getDescriptor();
  }
...
```

With `except` set in managed mode:

```
🐱~/src/playground/a @ $ cat buf.gen.yaml
version: v1
managed:
  enabled: true
  java_package_prefix:
    default: com.dkeung.playground
    except:
plugins:
  - name: java
    out: pkg/server/grpc
```

Output:

```
...
internal_static_a_HelloWorld_descriptor,
        new java.lang.String[] { "Hello", "World", "CreateTime", "Blah", });
    com.google.type.DatetimeProto.getDescriptor();
  }
...
```

Nice!

private/buf/bufgen/generator.go
```diff
@@ -346,20 +346,23 @@ func newModifier(
...
-	}
-	javaPackageModifier, err := bufimagemodify.JavaPackage(
-		logger,
-		sweeper,
-		javaPackagePrefix,
-		managedConfig.Override[bufimagemodify.JavaPackageID],
-	)
-	if err != nil {
-		return nil, err
+	if managedConfig.JavaPackagePrefix != nil {
```
doriable, line 349: In order to maintain full backwards compatibility, I think we still need to set `managedConfig.JavaPackagePrefix.Default` to `DefaultJavaPackagePrefix` value (`com`). Currently, in this implementation, that value is not used at all, so if there is no managed mode config set for `java_package_prefix`, we are now using an empty default instead of the original default, `com`.

srikrsna-buf: Nice catch! This fixed the `make generate` issue

doriable (approved):

private/buf/bufgen/bufgen.go
```diff
@@ -261,27 +270,75 @@ type ExternalPluginConfigV1 struct {
...
+
+// IsEmpty returns true if the config is empty.
+func (e ExternalJavaPackagePrefixConfigV1) IsEmpty() bool {
+	return e.Default == "" &&
+		len(e.Except) == 0 &&
+		len(e.Override) == 0
+}
+
+// UnmarshalYAML satisfies the yaml.Unmarshaler interface. This is done to maintain backward compatibility
+// of accepting a plain string value for java_package_prefix
```
amckinney, line 309:
```suggestion
// of accepting a plain string value for java_package_prefix.
```

private/buf/bufgen/bufgen.go
```diff
@@ -261,27 +270,75 @@ type ExternalPluginConfigV1 struct {
...
+	type rawExternalJavaPackagePrefixConfigV1 ExternalJavaPackagePrefixConfigV1
+	if err := unmarshal((*rawExternalJavaPackagePrefixConfigV1)(e)); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// UnmarshalJSON satisfies the json.Unmarshaler interface. This is done to maintain backward compatibility
+// of accepting a plain string value for java_package_prefix
```
amckinney, line 326:
```suggestion
// of accepting a plain string value for java_package_prefix.
```

private/buf/bufgen/bufgen.go
```diff
@@ -261,27 +270,75 @@ type ExternalPluginConfigV1 struct {
...
+}
+
+// IsEmpty returns true if the config is empty.
+func (e ExternalJavaPackagePrefixConfigV1) IsEmpty() bool {
+	return e.Default == "" &&
+		len(e.Except) == 0 &&
+		len(e.Override) == 0
+}
+
+// UnmarshalYAML satisfies the yaml.Unmarshaler interface. This is done to maintain backward compatibility
```
amckinney, line 308: It'd also be nice to include a link to the blog post that references why the `raw*` type is critical. Something as simple as "See https://abhinavg.net/posts/flexible-yaml for details." The same comment could be included in the `UnmarshalJSON` method, too.

amckinney (approved):
Nice job, this is great! A few things to improve comments, but this otherwise looks good to go. Please add an entry to the `CHANGELOG.md` describing the addition before mergin (using previous `CHANGELOG.md` entries as inspiration).

private/buf/bufgen/bufgen.go
```diff
@@ -261,27 +270,75 @@ type ExternalPluginConfigV1 struct {
...
+		return nil
+	}
+
+	type rawExternalJavaPackagePrefixConfigV1 ExternalJavaPackagePrefixConfigV1
+	if err := unmarshal((*rawExternalJavaPackagePrefixConfigV1)(e)); err != nil {
+		return err
+	}
+
+	return nil
+}
```
doriable, line 323: nit: The `UnmarshalYAML` implementation is currently being duplicated in the `UnmarshalJSON` implementation. We can instead re-use the logic by passing in `func(target interface{}) error { return json.Unmarshal(data, target) }` as the `unmarshal` function.

doriable: Reference: https://abhinavg.net/posts/flexible-yaml/#share-logic-with-yaml

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `except` and `override` to `java_package_prefix`.
```
amckinney, line 5:
```suggestion
- Add `default`, `except` and `override` to `java_package_prefix`.
```

amckinney (approved):
Nice!

PR #880, opened 2022-01-22
fix: git Cloner not running in isolation

kobim:
# What's the issue?

When running `buf breaking --against git://repo` inside [a git hook](https://git-scm.com/docs/githooks), the `Cloner` messes the current git repository because the environment contains new `GIT_` environment variables relating to the commit.

## Reproduction

Clone https://github.com/kobim/buf-example-breaking, and follow the instructions.

## What is the solution?

Following the same method [`pre-commit` uses](https://github.com/pre-commit/pre-commit/blob/84372e0edc7f5ac08361a10bcd312bc80dd20960/pre_commit/git.py#L27-L47), the `Cloner` will clean any `GIT_` environment variable that might interfere with the execution of `git` during the cloning.

### DIFF
--- a/private/pkg/git/cloner.go
+++ b/private/pkg/git/cloner.go
@@ -39,6 +39,17 @@ import (
 // will fail to fetch so we need to pick something.
 const bufCloneOrigin = "bufCloneOrigin"
 
+var (
+	validGitEnvVariables = []string{
+		"GIT_EXEC_PATH",
+		"GIT_SSH",
+		"GIT_SSH_COMMAND",
+		"GIT_SSL_CAINFO",
+		"GIT_SSL_NO_VERIFY",
+		"GIT_CONFIG_COUNT",
+	}
+)
+
 type cloner struct {
 	logger            *zap.Logger
 	storageosProvider storageos.Provider
@@ -70,6 +81,8 @@ func (c *cloner) CloneToBucket(
 ) (retErr error) {
 	ctx, span := trace.StartSpan(ctx, "git_clone_to_bucket")
 	defer span.End()
+
+	envContainer = c.getEnvContainerWithoutGitVariables(envContainer)
 
 	var err error
 	switch {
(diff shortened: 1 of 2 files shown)


lrewega:
I'm not sure `buf breaking` is intended to be run within the context of a hook. For one, it makes no attempt to distinguish between staged or unstaged files, so unless you're doing something fancy in your hook, `buf breaking --against .git#main` for example translates into `buf breaking "$PWD" --against .git#main`. If you are doing something fancy, then I think it's reasonable to solve for this in the context of the hook, which I presume is a shell script, simply by un-exporting `GIT_INDEX_FILE`:

```shell
#!/usr/bin/env bash
export -n GIT_INDEX_FILE
exec buf breaking ...
```

That being said, I think there are a few things `buf` could be doing better to ensure it doesn't make any destructive changes, like masking out `GIT_INDEX_FILE` from the environment when running `git --git-dir=...` commands, using `-C` or even perhaps setting `GIT_CEILING_DIRECTORIES`.

If you're not already using something for CI, I recommend checking out our buf-breaking action https://github.com/marketplace/actions/buf-breaking

lrewega:
Note also that you can use `git archive --format=tar "$(git write-tree)" | buf breaking -#format=tar --against .git#branch=main` to write the staged tree to `buf` directly.

Here's an example hook that seems to work correctly for your example:
```bash
#!/usr/bin/env bash
set -eo pipefail
git archive --format=tar "$(git write-tree)" | GIT_INDEX_FILE= buf breaking -#format=tar --against .git#branch=main
```

Or for something more `exec`-oriented
```
#!/usr/bin/env bash
GIT_INDEX_FILE= exec buf breaking -#format=tar --against .git#branch=main < <(git archive --format=tar "$(git write-tree)")
```

bufdev:
Closing this for now, I'm not sure we want to merge this.

PR #884, opened 2022-01-24
add force delete user endpoint

cyinma:
This is to add a new endpoint under new admin service to force delete a user including the organization and resources they have.


johanbrandhorst (approved):
LGTM, but wait for others to weigh in before merging

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -0,0 +1,46 @@
...
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+import "buf/alpha/registry/v1alpha1/organization.proto";
+import "buf/alpha/registry/v1alpha1/plugin.proto";
+import "buf/alpha/registry/v1alpha1/repository.proto";
+import "buf/alpha/registry/v1alpha1/user.proto";
+
+// UserService is the User service.
```
robbertvanginkel, line 24: This comment seems wrong.

robbertvanginkel:
I think this is the most reasonable way forward today given the circular import problem. I suppose because we're on `FILE` breaking we can't just move things around, even though for go it doesn't matter as much.

What I'm not 100% convinced of is saying only admins can force delete. As a user if I want to leave buf, I think forcing me to go through all of my resources individually and manually delete them is a bad experience. Maybe we don't care, because the user is leaving anyway, but I do feel that this direction makes it harder to revisit that experience.

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -0,0 +1,46 @@
...
+service AdminService {
+  // ForceDeleteUser forces to delete a user. Resources and organizations that are
+  // solely owned by the user will also be deleted.
+  rpc ForceDeleteUser(ForceDeleteUserRequest) returns (ForceDeleteUserResponse);
```
robbertvanginkel, line 28: I would rather see a generic "DeleteUser" with a parameter around deleting. More flexible if for whatever reason we want to let admins to "soft"(?) deletes of other users.

robbertvanginkel: Resolved offline, keeping as is.

robbertvanginkel (approved):

amckinney (approved):

PR #885, opened 2022-01-24
Draft requestbuilder service proto

robbertvanginkel:
What the apis for requestbuilder would look like with a dedicated service.


proto/buf/alpha/registry/v1alpha1/request_builder.proto
```diff
@@ -0,0 +1,100 @@
...
+  string method = 4;
+}
+
+message GetMethodDetailsResponse {
+  // A partial image containing only the files in which the request and
+  // response type for this method are declared, as well as all the files
+  // in the the transitive closure of their imports.
+  buf.alpha.image.v1.Image partial_image = 1;
+
+  // A jsonschema representing what the json encoded request message of
```
bufdev, line 90: On face value, big no-no here - we fundamentally say we do not ship JSON blobs within Protobuf, it's adding untyped data to a typed structure. If there's issues with using Protobuf, we need to solve those. There may be more to this though.

proto/buf/alpha/registry/v1alpha1/request_builder.proto
```diff
@@ -0,0 +1,100 @@
...
+  // the method should conform to. This schema is an approximation to be
+  // used by editors for validation and autocompletion, not a lossless
+  // representation of the type's descriptor (for that use partial_image).
+  bytes request_json_schema = 2;
+
+  // The commit the request reference resolved to. May be used on the
+  // client side to merge multiple partial_image's into one for repeated
+  // GetMethodDetails requests on the same module.
+  string resolved_commit = 3;
```
bufdev, line 99: What is the difference between a commit and a resolved commit?

robbertvanginkel:
The changes in e4d762e are an alternative to the dedicated service, based on the conversation I had with Alex about a more generic way to download only a part of a image. I'll put some inline comments in clarifying its intended use.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -34,3 +37,53 @@ message GetImageRequest {
...
+  // - "pkg.foo.*" for all messages in package 'pkg.foo'
+  // - "pkg.foo.*.**" for all messages under 'pkg.foo' and subpackages
+  // - "*.**" for all messages in all packages.
+  repeated string messages = 6;
+  // Type names or patterns describing which enums should be included in the
+  // partial image. For examples see `messages`.
+  repeated string enums = 7;
+  // Type names or patterns describing which services should be included in the
+  // partial image. For examples see `messages`.
+  repeated string services = 8;
```
robbertvanginkel, line 79: Talking to @bufdev, should these all be collapsed into a single `types`? If we implement this in bufmodule in general and allow partial images on types from the cli too, a single type may be a better flag experience.

I had originally separated them so it would be possible to get an image containing only the service definitions (without the messages+imports for their request/response etc) as this satisfies the GetServices needs of what is now in the RequestBuilder service. There may be other options there, like using something like fieldmask.

proto/buf/alpha/registry/v1alpha1/request_builder.proto
robbertvanginkel: Talked offline: this is probably closer to what's happening in the DecodeService for the runtime, where the expected out product is json. As the lsp for json in the editor is driven by an existing json config format, its more like shipping over a yaml or toml for use by another system: re-specifying that format in proto here to encode and decode will mostly be toil.

This functionality could be implemented on the client side in js/ts instead, avoiding this alltogether. However considering our resources its probably better here for now. With leaning towards allowing filters in image for partial images, this should go into its own separate service.

proto/buf/alpha/registry/v1alpha1/request_builder.proto
robbertvanginkel: Talked offline: I was unaware that Images contain this already. Resolving with same comment as https://github.com/bufbuild/buf/pull/885/files#r791240764.

robbertvanginkel:
@timostamm / @amckinney please see the changes in https://github.com/bufbuild/buf/commit/e4d762e94dd65cc6fb58c5df3cd9966f8854ead7 and the commits as separate. Not suggesting we do both - only including them in the same pr so we can compare and contrast easier. I chatted with @bufdev and added comments of main points, we're leaning towards the filtered getimage approach - does that make sense to you too?

proto/buf/alpha/registry/v1alpha1/image.proto
amckinney: I agree with consolidating in a single `types` for the symmetry you calling out. Having something like `buf build --type pet.v1.Pet --type payment.v2.Order` would be great.

I wanted to open an issue to discuss a tree-shaking solution like `buf build --only pet.v1.Pet`, but `--type` is just as clear, if not more so. It's also symmetrical with other potential commands like `buf decode`.

amckinney:
> I chatted with @bufdev and added comments of main points, we're leaning towards the filtered getimage approach - does that make sense to you too?

I agree with pushing more of these features into a single `GetImage` endpoint. It sounds like we're pushing to make this more symmetrical with the `buf build` command, which is generally something we should strive to do (i.e. symmetry between a subset of CLI commands and RPCs).

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -34,3 +37,53 @@ message GetImageRequest {
...
+
+message GetPartialImageRequest {
+  string owner = 1;
+  string repository = 2;
+  string reference = 3;
+
+  // If the image should exclude depdendencies. When set to true, a incomplete
+  // image without descriptors for dependencies of the requested types is
+  // returned.
+  bool exclude_dependencies = 4;
```
amckinney, line 61: Do we need to start with this option? I understand the distinction and potential use case, but when would we need to return an `Image` containing the `pet.v1.Pet` type without the `google.type.DateTime` definition?

For request builder and other use cases, I'd imagine that we still need to return the tight transitive closure around the type, so that all of the message's field could be referenced and included in an auto-complete placeholder.

To be clear, I'm referring to [this](https://buf.build/acme/petapis/file/main/pet/v1/pet.proto#L22) where the `google.type.DateTime` fields are unclear from the `Pet` type on its own:

```protobuf
// Pet represents a pet in the pet store.
message Pet {
  PetType pet_type = 1;
  string pet_id = 2;
  string name = 3;
  google.type.DateTime created_at = 4;
}
```

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -34,3 +37,53 @@ message GetImageRequest {
...
+
+  // Determines if the patterns specified in {messages,enums,services} should
+  // consider types from imported modules.
+  bool match_dependencies = 5;
```
amckinney, line 65: Also a cool idea, but I wonder if we could drop this option for now. Based on my comment above, if we include the transitive closure of the target type, we would only need a single RPC and I don't think we'd need this option at all. 

Was the intention that a separate RPC would be issued for the `google.type.DateTime` definition against the same module reference (e.g. `buf.build/acme/petapis`)? Then both of the results would be used by request builder?

I could see the value of an option like this in general since it's not always clear where the `google.type.DateTime` type is defined, but I'm not sure it's necessary for this use case. I could be missing something though.

proto/buf/alpha/registry/v1alpha1/image.proto
robbertvanginkel: Separate point that I now notice isn't clear, with dependencies here I mean to refer to imported files (`repeated string dependencies` in the FileDescriptor definitions. Below with `match_dependencies` I had meant to refer to imports at the buf module level, not at the file level.

------

The use case I was eyeing for this option is that if I just want a list of all services (and their methods), I don't want the request's dependencies yet at this point. Say this is part of a large module:
```
message DoRequest {
    google.type.DateTime created_at = 4;
}
service PetAPI {
  rpc Do(DoRequest) returns (...);
}
```
And I would like filedescriptors that contain requests, but don't want things they refer to. For a large module like googleapis, querying for only the services while not receiving all the enums and messages that are required for making calls with them seems useful (maybe I should dig up some numbers)?

I don't think this is the common case though, which is why I made exclude (and false by default).

proto/buf/alpha/registry/v1alpha1/image.proto
robbertvanginkel: Related to the point above, with dependencies here I mean to refer to dependencies as in buf modules (from buf.lock's).

---

The intention here is that when I query for or "services in buf module x" I don't want to match all services in its dependencies (say we import googleapis, I don't want to show googleapi's services). However, at the point that I would want to query for a partial image with method `y`'s input type there is no guarantee in which bsr module that type is. In most cases (especially when people follow our lint rules) it will be, but it isn't required. 

For this API to work for request builder, I'm looking for a way to express that when I query for services I only want matches in this buf module 'shallowly' (without file imports) whereas if I'm querying for the request/response I want types from any bsr module in the image 'deeply' (with file imports needed).

proto/buf/alpha/registry/v1alpha1/request_builder.proto
```diff
@@ -0,0 +1,100 @@
...
+
+  // Fully qualified method in the module, like "pkg.baz.BarService/Foo"
+  string method = 4;
+}
+
+message GetMethodDetailsResponse {
+  // A partial image containing only the files in which the request and
+  // response type for this method are declared, as well as all the files
+  // in the the transitive closure of their imports.
+  buf.alpha.image.v1.Image partial_image = 1;
```
timostamm, line 88: I assume that with `rpc GetPartialImageRequest`, this is no longer necessary? As input / output types could be retrieved via `repeated string types` discussed above?

proto/buf/alpha/registry/v1alpha1/request_builder.proto
```diff
@@ -0,0 +1,100 @@
...
+
+// RequestBuilderService serves endpoints used by requestbuilder.
+service RequestBuilderService {
+  // GetServices returns an overview of all services declared in a module
+  // and their methods.
+  rpc GetServices(GetServicesRequest) returns (GetServicesResponse);
+
+  // GetMethodDetails returns data needed to make dynamic requests for a method.
+  rpc GetMethodDetails(GetMethodDetailsRequest) returns (GetMethodDetailsResponse);
+}
```
timostamm, line 61: It looks to me like this service would provide all necessary information to allow users to quickly browse trough a list of services and methods (or search by name), and show the most relevant information (name, method type, deprecation) at the same time. This should be quite efficient too, and allow for sharing of a request as well.

Just a thought: If it weren't for the jsonschema, I _think_ we could get away with just `rpc GetServices`, and simply get a (partial) image for the service type. Obviously there would be a bit of overhead of retrieving all methods of the service where we only need one, but it might be a worthy trade-off. We do however have jsonschema, and I'm glad we do :sweat_smile:.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -34,3 +37,53 @@ message GetImageRequest {
...
+  PARTIAL_IMAGE_LEVEL_UNSPECIFIED = 0;
+
+  // Returns complete files in which the requested types are contained.
+  PARTIAL_IMAGE_LEVEL_FILES = 1;
+
+  // Returns trimmed files in which the requested types are contained. In a
+  // trimmed file, all declarations that are not embedded (directly or
+  // indirectly) in requested types are removed.
+  PARTIAL_IMAGE_LEVEL_TYPES = 2;
+}
```
timostamm, line 51: As far as I understand, this is a remnant of an earlier version? I think the most robust and expected behavior would be to keep types in their original files.

timostamm (commented):
Moving some functionality into the `GetImage` endpoint for symmetry with `buf build` seems  like a great idea to me. But from what I can see we still need a separate endpoint to list services only, and think we should try to avoid filters in `GetImage` for concerns that can be solved by the separate endpoint.

proto/buf/alpha/registry/v1alpha1/image.proto
robbertvanginkel: I forgot to include this as a field, it is something that Alex and I discussed to get more minimal filedescriptors that only contain the types you're interested in without returning types just because they were declared in the same file.

I agree with you that this is much harder to get right. Especially rewriting `SourceCodeInfo` and removeing unused imports as if those types were never there will be tricky. Additionally, merging multiple of these received over multiple client calls on the client side is painful, whereas if you only have to combine complete filedescriptors its much more straightforward.

I don't think it needs to be included and agreed on today. Perhaps if we want this very granular type information we should look for other avenues than the Images/FileDescriptor to communicate that.

proto/buf/alpha/registry/v1alpha1/image.proto
amckinney: Yea, I prefer dropping the leading `.` prefix since it's generally just simpler and cleaner to read. This is also what the [protoreflect.FullName](https://github.com/protocolbuffers/protobuf-go/blob/e5db2960ed1380681b571cdf4648230beefaf58b/reflect/protoreflect/proto.go#L438) representation does.

Wildcards sound cool, but I doubt we'll need to support them at first.

robbertvanginkel:
Here's a small example with some expectations that should demonstrate why the current fields are here:

```
--- `pkg/foo/foo.proto` in buf.build/project/a ---
package pgk.foo;

import "pkg/foo/bar.proto"
import "pkg/bar/bar.proto"

message Foo { 
	....
	OtherFileMessage thing = 1;
}

service Service {
	rpc Foo(Foo) returns (pkg.bar.OtherModuleMessage);
}
service Service2 { ... }

--- `pkg/foo/baz.proto` buf.build/project/a ---
package pkg.foo;

message OtherFileMessage { ... }

--- `pkg/bar/bar.proto` buf.build/project/b ---
package pkg.bar;

message OtherModuleMessage { ... }

service OtherModuleService { ... }
```


Now I have the following needs:

1. A way of listing all services+method 'metadata' in a specific buf module. Must be somewhat cheap, don't want services in dependent buf modules, nor a the transitive closure of files/types needed for constructing the input/output types.
2. A to get the transitive closure of files/types needed for constructing a particular type, not knowing in which buf module this type resides.

With the parameters as they are today, I could do the following requests:
1.
```
GetPartialImageRequest {
	exclude_imports = true,
	match_dependencies = false,
	services = ['*.**']
}
which would return an image with only `pkg/foo/foo.proto`.

2. and once a user selects a method, say `pkg.foo.Service/Foo`
```
GetPartialImageRequest {
	exclude_imports = false,
	match_dependencies = true,
	messages = ['pkg.foo.Foo', 'pkg.bar.OtherModuleMessage']
}
```
which would return an image with all 3 files.

Maybe the specific subset of the image that I want for 1 should not be part of getting partial image. If so, wildcards and `match_dependencies` aren't necessary features but we'll need a requestbuilder specific endpoint for the reasons in https://github.com/bufbuild/buf/pull/885/files#r791589673. However discussing yesterday we were looking if we could rely on partial images for all of the features (except the jsonschema serving), with the hopes of avoiding a requestbuilder specific listing endpoint.

File: proto/buf/alpha/registry/v1alpha1/image.proto (hunk 3 above)
Comment (bufdev) [reply proto/buf/alpha/registry/v1alpha1/image.proto:79]: we drop the `.` for full names across the board

File: proto/buf/alpha/registry/v1alpha1/jsonschema.proto [hunk 9]
```diff
@@ -0,0 +1,49 @@
...
+}
+
+
+message GetJSONSchemaRequest {
+  string owner = 1;
+  string repository = 2;
+  string reference = 3;
+
+  // A fully qualified name of the type to generate a JSONSchema for, e.g.
+  // "pkg.foo.Bar".
```
robbertvanginkel, line 34: Given https://github.com/bufbuild/buf/pull/885#discussion_r792040774, should we call out that the leading `.` is dropped or is this good enough with the example?

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -29,6 +30,17 @@ message GetImageRequest {
   string owner = 1;
   string repository = 2;
   string reference = 3;
+
+  // Exclude files from imported buf modules in this image.
+  bool exclude_imports = 4;
+
```
robbertvanginkel, line 36: I did not end up including a "exclude_source_info" bool, as users can use a field mask to omit them. As the field masks are include only in this proposal, its not very ergonomic though as you have to specify everything else. Open to suggestion around the ux.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -29,6 +30,17 @@ message GetImageRequest {
...
+  // When specified the returned image will only contain the necessary files to
+  // describe these types.
+  repeated string types = 5;
+
+  // When specified, applies as projection mask to the resulting Image, i.e.
+  // only fields specified in this mask are returned.
+  google.protobuf.FieldMask image_mask = 6;
```
robbertvanginkel, line 43: The only thing that's somewhat concerning about this is that according to fieldmask's documentation:

> ```
> // A repeated field is not allowed except at the last position of a
> // paths string.
> ```

https://github.com/protocolbuffers/protobuf/blob/b696347f80d324c624e65a0b7e3ea5dac7ec5a41/src/google/protobuf/field_mask.proto#L85-L86

Whereas images are repeated ImageFiles at the toplevel, so we wouldn't conform to that. Go doesnt really have a well known fieldmask library so we can handle it however we want, but I'm looking on a bit if this makes sense in any of the other languages.

johanbrandhorst: This documentation has confused me in the past in the grpc-gateway, it's surprisingly subtle: https://github.com/grpc-ecosystem/grpc-gateway/issues/1766. I'm still not sure I understand entirely what's up.

proto/buf/alpha/registry/v1alpha1/jsonschema.proto
```diff
@@ -0,0 +1,49 @@
...
+message GetJSONSchemaResponse {
+  // A json schema representing what the json encoded request message of
+  // the method should conform to. This schema is an approximation to be
+  // used by editors for validation and autocompletion, not a lossless
+  // representation of the type's descriptor.
+  //
+  // If the requested type resolves to a message or enum, the json schema can be
+  // used as is. If the type resolves to a service, the json schema will only
+  // contain definitions for each request type for that service and the client
+  // will have to manaully set the toplevel "$ref" property.
```
robbertvanginkel, line 47: @amckinney @timostamm this comes out a little ugly, as json schema without a toplevel type doesnt really mean anything (I'll link a draft pr in a bit to clarify).

Similarly we could include the response types as well, but as these won't be be used (at least in rb) this still feels a little ugly. Perhaps allowing types from dependencies is the least worse compromise in the end?

proto/buf/alpha/registry/v1alpha1/image.proto
amckinney: Got it. An `exclude_source_info` parameter sounds nice for the symmetry with `buf build`, but it can be added later (if ever).

As for the field mask problem, I wonder if we should expose a `include_mask` and an `exclude_mask` that could be combined for both whitelisting and blacklisting behavior?

proto/buf/alpha/registry/v1alpha1/jsonschema.proto
timostamm: Looks like there are quite a few downsides:
1. There doesn't seem to be a meaningful representation for services in json schema, because services are not data
2. We ask the client to manipulate the schema to make it usable (relying on implementation details)
3. We only return request types (I can imagine other clients might want response types too)

> Perhaps allowing types from dependencies is the least worse compromise in the end?

I tend to think so. But if we decide otherwise, we can obviously make it work in the rb frontend.

proto/buf/alpha/registry/v1alpha1/image.proto
robbertvanginkel: I dove into fieldmask and its documented restrictions a bit more, here's the summary:

First, highlights from the documented restrictions from https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/field_mask.proto
```
// `FieldMask` represents a set of symbolic field paths, for example:
//
// ...
//
// # Field Masks in Projections
//
// ...
//
// A repeated field is not allowed except at the last position of a
// paths string.
//
// ...
//
// # Field Masks in Update Operations
//
// ...
// 
// If a repeated field is specified for an update operation, new values will
// be appended to the existing repeated field in the target resource. Note that
// a repeated field is only allowed in the last position of a `paths` string.
//
// ...
//
//
// ## Field Mask Verification
//
// The implementation of any API method which has a FieldMask type field in the
// request should verify the included field paths, and return an
// `INVALID_ARGUMENT` error if any path is unmappable.
```

Looking through the fieldmask helper implementations in the protobuf repository, it seems like these are all real limitations in their `Merge` methods:

- c#: https://cs.opensource.google/protobuf/protobuf/+/master:csharp/src/Google.Protobuf/FieldMaskTree.cs;l=273-278;drc=e479410564727d8954e0704254f4345f97e3d844
- c++: https://cs.opensource.google/protobuf/protobuf/+/master:src/google/protobuf/util/field_mask_util.cc;l=468-476;drc=d6699989ab5164cc7778ea4802e49aae51a4f417
- java: https://cs.opensource.google/protobuf/protobuf/+/master:java/util/src/main/java/com/google/protobuf/util/FieldMaskTree.java;l=294-301;drc=de371235c9f50eadad7a3562572b79a9422cd754

Note that the `Merge` method isn't present in go's protobuf implementation. The signature is usually something like `Merge(fullMessage, fieldmask, emptyMessage)`, after which `emptyMessage` will have `fullMessage`'s fields if they were present in field mask.

However, it seems like google itself doesn't really stick to this limitation:

https://google.aip.dev/157

> An API may allow read masks with non-terminal repeated fields (unlike update masks), but is not obligated to do so.

https://google.aip.dev/161

> ```
>   // Reviews for the back cover. The key is the author of the review,
>   // and the value is the text of the review.
>   //
>   // Valid field masks: reviews, reviews.smith, reviews.`John Smith`
>   map<string, string> reviews = 2;
> 
> ```
> 
> > Field masks may permit the use of the * character on a repeated field or map to indicate the specification of particular sub-fields in the collection:
> 
> ```
>   // Valid field masks: authors, authors.*.given_name, authors.*.family_name
>   // Invalid field masks: authors.0, authors.0.given_name
>   repeated Author authors = 2;
> ```

---

I think in short the wkt fieldmask has a:
1. documented limitations on non terminal repeated fields
2. protobuf implementation helpers that conform to 1
3. recommendations that apis accept field masks not conforming to the spec as long as the server implements it.

Point 3 seems like the one we can work with, there's nothing that prevents us from accepting a field mask like `image.file.*.services` and doing the thing we want on the server side. As golang has no fieldmask utils, we're on our own to parse and interpret them. Unfortunately, we would need to prevent the usage of [fieldmaskpb#FieldMask.IsValid](https://pkg.go.dev/google.golang.org/protobuf/types/known/fieldmaskpb#FieldMask.IsValid) as that does only allow repeateds in the last part of the fieldmask: https://github.com/protocolbuffers/protobuf-go/blob/v1.27.1/types/known/fieldmaskpb/field_mask.pb.go#L397-L400.

Still, we could absorb that with our own fieldmask parsing/handling code. Since clients just send field masks, as long as we document what we accept on the server side that could technically work. Unfortunately, languages that have utils for creating fieldmasks like java's `FieldMaskUtil.fromFieldNumbers(Production.class, Production.TITLE_FIELD_NUMBER)` [won't be able to](https://cs.opensource.google/protobuf/protobuf/+/master:java/util/src/main/java/com/google/protobuf/util/FieldMaskUtil.java;l=149-171;drc=3afc8283099dd1747260e088cdf27240a010161e) create the correct `.*.` path parts programatically.

Still, if we accept that on the go side we can make it work. There's a go utility library for fieldmasks that works through repeated fields: https://github.com/mennanov/fmutils/blob/af9ce0f96ea1983d7a3350cb5edebd499f405b0d/fmutils.go#L80-L84. Purely illustrative to what we can do on the server side, we probably don't want to use that for other reasons (it's documented to panic if the mask is invalid). 

If we look at all of the above and decide its acceptable, I would need to look more at if the `.*` (and maybe backtick case) messes with fieldmask has documented special proto json handling.

Overall, this feels like a serious limitation and a mess. I think I'm at the point where a buf fieldmask or a enum with possible view options (similar to described in https://google.aip.dev/157) seems more sane.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -25,10 +25,90 @@ service ImageService {
   rpc GetImage(GetImageRequest) returns (GetImageResponse);
 }
 
+// ImageMask is used in GetImageRequest to specify which parts of an image are
+// requested.
+enum ImageMask {
```
robbertvanginkel, line 30: This is now the proposal to replace the fieldmask, a specific enum over the repeateds we may want to include.

Alternatively, a `bool services_only` would be good enough for now, until we do the work for a better field mask. Both the this enum and a services only bool should be easy to convert in whatever new fieldmask we come up with, so I'm not strongly tied to either.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -25,10 +25,90 @@ service ImageService {
...
+//      extensions 2 to 3;
+//    }
+//    extend Foo {
+//      optional Baz baz = 2;
+//    }
+//    extend google.protobuf.FieldOptions {
+//      optional Qux my_option = 51234;
+//    }
+//
+enum TypeDependencyStrategy {
```
robbertvanginkel, line 66: This is based on a comment @amckinney and I had the other day on how to calculate the transitive closure of types. The minimal strategy is good enough for what I need at this time, but for a generic API it may be nice. Unless of course we decide that the 3rd strategy / "options" is always the way to go.

proto/buf/alpha/registry/v1alpha1/image.proto
robbertvanginkel: Given the field mask issues described in https://github.com/bufbuild/buf/pull/885#discussion_r793136489, I don't want to use those so resolving.

proto/buf/alpha/registry/v1alpha1/jsonschema.proto
robbertvanginkel: This is now updated to say that in this API, it is ok if the type you ask for is declared in a dependent module.

proto/buf/alpha/registry/v1alpha1/image.proto
amckinney: I like what you're going for here, but I agree that we should probably punt on more granular control here for now.

The only thing I'd add is that we might want this enum to be _additive_ such that the user could request, `EXTENSIONS` _and_ `OPTIONS` instead of having the implicit relationship that `OPTIONS` includes `EXTENSIONS`. The `MINIMAL` enum value is always relevant, so we could model this as something like this:

```
message GetImageRequest {
  ...
  include_extensions = 7;
  include_options = 8;
  ...
}
```

Encapsulating all of this in an `enum` is nice, but I'm not sure that we'd ever need to add any other strategies here, so the explicit parameters might be simplest (since there's only two). But to be clear, we could only include the options and extensions right? Can you imagine anything else?

robbertvanginkel: Yes I think you're right. I think originally I was considering that options are always extensions on a {message,field,file,service}. Those extensions have a special meaning though and and should probably get some custom handling. Consider someone importing `descriptor.proto`, should requesting extensions cause all options in a module to be sent over?

I can't currently think of any other way to follow dependencies, 2 bools will probably do but they need to be `exclude_` for backwards compatibility. The only worry I have is the amount of bools that end up in this request. `exclude_imports` & `exclude_source_info` seem fine and are similar to the cli, does an extra `exclude_extensions`, `exclude_option` and `type` work for you as cli args too? I don't think we need a cli arg for the equivalent of `include_mask` no?

amckinney: Good call out on the extensions defined in `descriptor.proto`. I'd prefer if there wasn't any custom handling around `descriptor.proto`, but I know that we'll need to consider future proofing here. What are your thoughts? I feel like it's simplest for us to say, "You've asked for extensions and included `descriptor.proto`, so that's what you get."

I agree that shoving more `bool` parameters into the request doesn't feel great - it might be a slippery slope. But I think we can justify a `--include-mask`, `--exclude-extensions`, and `--exclude-options` flags in `buf build` for the same reasons we're justifying it here. I don't think we need to expose them on the CLI to begin with, but anything we have on `GetImage` needs to be available on `buf build` so that people can perform this operation locally rather than via an RPC.

robbertvanginkel: I think that when you include `descriptor.proto`, this is because you aim to write/send (parts of) filedescriptors. In this case I don't see what the use is of returning you the custom options of the FileDescriptorSet used to compile the application that handles the descriptor protos would be used for.

Example, consider:
```
import "google/protobuf/descriptor.proto"
message Foo {
	google.protobuf.FileDescriptorSet fds = 1;
}
// other file same module
message Bar {
	string a = 1 [(my_option) = {}];
}
extends google.protobuf.FieldOption {
	optional Baz my_option = 51234;
}
```

If you request the types needed for `Foo`, should that include `Baz`? Without special handling, I think you'd recurse  `Foo -> FileDescriptorSet -> ... -> FieldOption -> find all extensions -> Baz`.

The worry I have about a generic `--exclude-extensions/--exclude-options` is that they don't make much sense without specifying a type to filter on.

robbertvanginkel: Resolved offline: we will treat the `google.protobuf.*Options` specially, similar to how the compiler treats them specially and by allowing the language to be extended based on these options. Looking at the example above, we can't come up with a reason as to why `Baz` should be included in a minimal FileDescriptorSet that describes Foo. If we would make a Foo message that contained a FileDescriptorSet that Foo was made with, the extensions would be declared inside the `fds` field as a concrete instance of an extension in `FieldDescriptorProto extensions = 7`. There is no need to include its definition in the FileDescriptor declaring `Foo`.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -25,10 +25,79 @@ service ImageService {
...
+  //    extend Foo {
+  //      optional Baz baz = 2;
+  //    }
+  //    --- qux.proto ---
+  //    extend google.protobuf.FieldOptions {
+  //      optional Qux my_option = 51234;
+  //    }
+  //
+  //  When `exclude_extensions` is true and `types` is "Foo", the resulting image
+  //  will contains types for "Foo", "Bar" and "Qux", but not "Baz".
```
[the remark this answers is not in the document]
bufdev: If compiled with it, extension of a message is considered part of the message - you can't arbitrarily exclude it, that changes the meaning of Protobuf. Unless I'm missing something, this option shouldn't exist.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -25,10 +25,79 @@ service ImageService {
...
+  bool exclude_extensions = 7;
+
+  // When `exclude_extensions` is true, images may omit extends declarations for
```
bufdev, line 87: Any option applied is a part of that definition, i.e. a field option is a part of that definition. Just as you wouldn't say "exclude any `deprecated` options", you can't then say that custom field options are ignored. Again unless I'm missing something, this option shouldn't exist.

bufdev:
A general point here that I'm worried is missed: you can't arbitrarily say "exclude extensions of this message" or "exclude custom options of this field/rpc/message/etc". Excluding extensions is like saying "give me the message definition, but exclude any fields with tags greater than 10" - an extension, if present, is a part of the message as far as protobuf is concerned. Similarly, you wouldn't say "give me the message definition, but don't tell me what fields are marked as `deprecated`" - if a custom option is present, it is no different in importance to builtin options.

I've missed a lot of the conversation around this so assumedly there is more context, but just in case: if this is driven by an implementation detail of implementing type filtering, (1) the API shouldn't be driven by this implementation detail :-) (2) I think I solved it over the weekend in principle, can explain the workflow tomorrow.

robbertvanginkel:
Having parameters for excluding custom options and extensions is a continuation of the idea of returning more granular Images, and wether it makes sense or not depends on the context that image will be used in. Similar to how omitting files from a Image changes its contents/meaning of that image, (using partial images for breaking change detection wouldn't work) omitting extensions or custom options does change the messages but if this is desirable or problematic depends on the use.

Custom options and built in options differ because the default proto runtime libraries only interpret built in options. If you need to take a binary protobuf payload and convert it into a json encoded message, custom options are irrelevant. In another case someone might maintain custom options on their internal protobuf's used for code generation, but they publish a public set of proto's with those custom options omitted: that does change the message definition but not in a way that would be problematic to any caller.

Extensions don't have to be declared in the same module as their extendee, there's already no guarantees we can give about returning all extensions for a message. In that sense this is different from omitting fields above field number `x` as a message's regular fields are all declared in one place. As for the use of omitting extensions there's parts of the protobuf ecosystem that don't work with extensions (like the json encoding), as well as use cases that don't require knowing about them (like a infrastructure level middleware that forwards based on a envelope message's declared fields but treats extensions as unknown fields).

For our immediate use cases in rb/runtime encoding we don't have a use for custom options due to just mapping between encodings which is only influenced by messages and their built in options, similar for extensions we will eventually use the recommended but not mandatory proto2 json encoding but since this isn't supported in all languages it is lower on the priority list.

I think we agree that omitting extensions or options changes a message. When discussing with @amckinney, it seemed to us that omitting details of messages is not significantly different than omitting types in images: wether it makes sense depends on how you use the resulting image. If the above cases justify building this now is an open question, but in the context of what does a partial image endpoint look like and how could it be used I think they are defensible options.

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -25,10 +25,79 @@ service ImageService {
...
 }
 
+// ImageMask is used in GetImageRequest to specify which parts of an image are
+// requested.
+enum ImageMask {
+  IMAGE_MASK_UNSPECIFIED = 0;
+
+  // IMAGE_MASK_MESSAGES refers to ImageFile's `google.protobuf.DescriptorProto
+  // message_type` field.
+  IMAGE_MASK_MESSAGES = 1;
```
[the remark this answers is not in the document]
robbertvanginkel: Yes the ImageMask enum is meant as a fieldmask like "don't send these top level fields in responses" with no regard for what possible dependencies or links exist across those fields.

bufdev:
A Module is a logical buildable unit. It describes the Protobuf world a given builder sees, so if you are getting an Image for a given Module, any options or extensions declared as part of it are considered to be first-class citizens. A Module contains it's definitions and all of it's dependency's definitions, so we can understand when I say to build `buf.build/foo/bar`, what extensions and options I expect are part of this Module.

To remove extensions or options from this would change the meaning of the extended messages, and change the meaning of any of the elements with applied options. While we of course can't know about extensions for Modules that itself depend on `buf.build/foo/bar`, we can know about all extensions in the context of `buf.build/foo/bar`, which is what we are asking to build.

Partial images make sense in the context of:

- Filtering source code info: This is metadata and does not change the meaning of a given message/service/enum. Extensions and custom options are not metadata, even if options seem like metadata sometimes.
- Filtering imports: The meaning of a given message/service/enum is not changed, we are just not providing all the relevant data to have a fully-encapsulated FileDescriptorSet that describes a given message/service/enum (if such a type depends on an import for a field/request/response type).
- Filtering by type: This is the primary goal we're trying to enable here - we want to say "don't give me all the definitions in this Module, just give me the definitions for types X, Y, Z". Just giving us the definitions for X, Y, Z doesn't change the meaning of X, Y, Z, as filtering extensions or custom options would.

> as well as use cases that don't require knowing about them (like a infrastructure level middleware that forwards based on a envelope message's declared fields but treats extensions as unknown fields)

`--exclude-extensions` is a 100% no go, but on face value, you could make an argument for `--exclude-options` in the long term based on saying "all I care about is the structure of the underlying message, ever" since marshal/unmarshal should not, in theory, be affected by custom options. However, there are options that affect marshal/unmarshal, specifically `packed` https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto#L554 - this means that as part of Protobuf, it's already an accepted fact that the wire protocol can be affected by options, and if someone were to implement a custom extension to the protobuf wire format in the future, they could reasonably have custom options that affect it. You could make the argument that THAT is changing the meaning of Protobuf, which is valid, but I think the key issue here is that the endpoint (and `buf build`) should assume the purpose of any given custom option.

Even if the above wasn't the case, we have no need for these options right now, so there's no reason to spend time to implement them. Extensions will only be present in the 0.01% case and add very little in the way of payload, and options add very little in the way of payload, so there's no reason to optimize here.

proto/buf/alpha/registry/v1alpha1/image.proto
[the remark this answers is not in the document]
bufdev: Spelling this out a bit more in the comments would be nice too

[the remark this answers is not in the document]
robbertvanginkel: Updated the comment on the enum and on `include_mask`.

bufdev (approved):

amckinney (approved):

PR #889, opened 2022-01-25
add members count in organization setting and add repository setting

cyinma:
This is to add members count field for organization setting which returns the number of members of an organization.

This also added a new `GetRepositorySettings` endpoints which returns the number of contributors of a repository.


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -244,3 +246,13 @@ message ListRepositoryContributorsResponse {
...
   string next_page_token = 2;
 }
+
+message GetRepositorySettingsRequest {
+  // The ID of the repository for which to get the settings.
+  string repository_id = 1;
+}
+
+message GetRepositorySettingsResponse {
+  // The number of outside contributors in the repository.
```
johanbrandhorst, line 256: Does this include implicit members with an explicit role?

cyinma: no, it's not, will add this to the comment

johanbrandhorst (approved):

Issue #567, opened 2021-09-18
`buf breaking` errors with workspaces containing different directories


amckinney:
Several users have been bitten by the "Failure: input contained 2 images, whereas against contained 1 images" error when running `buf breaking` with workspaces. Let's revisit this and see if we can do anything about it.

amckinney:
I did a bit more investigation into this one. At first glance, it seems like we can just merge the images and compare the current snapshot with the `--against` snapshot, but this breaks down with respect to the [breaking configuration configured for each image](https://github.com/bufbuild/buf/blob/624a8ed4621d654e167bb6c01ce752f39a6961c6/private/buf/cmd/buf/command/breaking/breaking.go#L376).

If we were to do this, we need a way to assign each configuration to different portions of the image, and each file contained in the image would only have a single configuration. This gets complicated in a variety of cases, such as files that are present in the current snapshot, but not in the `--against` snapshot. In general, these files should only introduce _new_ types, but it's possible that they redeclare and/or break types that are already found in other files in the same package. We'd need to compare these files against the merged `--against` image (at least for files that belong to the same package).

I'll continue to look into this and see what we can do.

andrewparmet:
I was just bitten by this in my Buf Gradle plugin integration.

The way the plugin works is that it uses Buf to create a JSON image and then breakage checks run against images from previous versions. I recently modified the plugin to use workspaces with symlinks to protobuf source directories instead of copying protobuf source into a single directory and now breakage checks fail with this error.

Given that I want to check compatibility against a published artifact, is there a better way to achieve this workflow than publishing a JSON image with Buf itself? Should I just export the protobuf directories as an archive? I feel like this is a use case that images are intended for.

Here's [a branch with a failing test](https://github.com/andrewparmet/buf-gradle-plugin/tree/multi-workspace-breaking-failure). The test can be executed by running `./gradlew test --tests 'BreakingTest.schema with multi-directory workspace'`.

The test runs against the sample project [defined here](https://github.com/andrewparmet/buf-gradle-plugin/tree/multi-workspace-breaking-failure/src/test/resources/BreakingTest/schema%20with%20dependency) and creates a temporary `buf.work.yaml` that looks like this:

```yaml
version: v1
directories:
  - src-main-proto
  - build-extracted--include--protos-main
```

[...]

PR #893, opened 2022-01-28
match `buf breaking` behaviour with `buf build` for workspaces

srikrsna-buf:
## Overview
One solution for #567 and also for workspace directories being altered is to treat the entire workspace as one module. This will emulate,
```shell
buf breaking --against against.bin $(buf build .)
```

## Reasoning
- `buf build` already treats a workspace as one module producing only one image. 
- This is similar to running `buf build` on two different revisions of code and testing their output for breaking change detection. 
- At a workspace level it does ensure nothing breaks even if files are moved around.

## Problem
- We will not be able to detect if files are moved amongst the directories of a workspace breaking the modules. 
- This may mislead people to believe the modules themselves didn't break whereas such a check never happens. Similar to how one might assume `buf build` when targeting a workspace will have information regarding the workspace itself.

## Implementation
We can merge all the images for both `input` and `against` respectively and use their respective configs if available. Otherwise fallback to defaults.

### DIFF
--- a/private/buf/bufwire/image_config.go
+++ b/private/buf/bufwire/image_config.go
@@ -38,3 +38,10 @@ func (i *imageConfig) Image() bufimage.Image {
 func (i *imageConfig) Config() *bufconfig.Config {
 	return i.config
 }
+
+func NewImageConfig(image bufimage.Image, config *bufconfig.Config) *imageConfig {
+	return &imageConfig{
+		image:  image,
+		config: config,
+	}
+}
--- a/private/buf/cmd/buf/command/breaking/breaking.go
+++ b/private/buf/cmd/buf/command/breaking/breaking.go
@@ -24,10 +24,12 @@ import (
 	"github.com/bufbuild/buf/private/buf/bufwire"
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
 	"github.com/bufbuild/buf/private/bufpkg/bufcheck/bufbreaking"
+	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/command"
+	"github.com/bufbuild/buf/private/pkg/storage"
 	"github.com/bufbuild/buf/private/pkg/storage/storageos"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


private/buf/bufwire/image_config.go
```diff
@@ -38,3 +38,10 @@ func (i *imageConfig) Image() bufimage.Image {
 func (i *imageConfig) Config() *bufconfig.Config {
 	return i.config
 }
+
+func NewImageConfig(image bufimage.Image, config *bufconfig.Config) *imageConfig {
```
bufdev, line 42: See the internal go style guide regarding file layout and where public types go, and see other places in this codebase for the patterns to follow here.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -400,3 +434,30 @@ func getExternalPathsForImages(imageConfigs []bufwire.ImageConfig, excludeImport
 	}
 	return stringutil.MapToSlice(externalPaths), nil
 }
+
+func mergeImageConfigs(
```
bufdev, line 438: Logic like this should not be in a specific command's package.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -400,3 +434,30 @@ func getExternalPathsForImages(imageConfigs []bufwire.ImageConfig, excludeImport
...
+	ctx context.Context,
+	imageConfigs []bufwire.ImageConfig,
+	readBucket storage.ReadBucket,
+	inputConfig string,
+) (bufwire.ImageConfig, error) {
+	config, err := bufconfig.ReadConfigOS(
```
bufdev, line 444: This all needs more explanation as to what is going on, I'm a bit confused here

amckinney: Yea I don't think this is right. This looks like we're reading a single configuration file from the OS and using it for the  entire workspace - this would violate the property we need to maintain that's described [here](https://github.com/bufbuild/buf/issues/567#issuecomment-936156567).

private/buf/cmd/buf/command/breaking/breaking.go
srikrsna-buf: Agreed. I'll refactor once we finalise this

private/buf/cmd/buf/command/breaking/breaking.go
srikrsna-buf: This is the first proposed approach where we try to emulate
```bash
buf breaking --against against.bin $(buf build .) 
```
This is similar to the behaviour explained [here](https://github.com/bufbuild/buf/issues/567#issuecomment-998149291).

Here we are reading from `--against-config` and `--config` if not found will revert to the OS.

bufdev:
What do we want to do with this?

PR #896, opened 2022-01-30
Add RepositoryTrackCommitService

WillAbides:
This adds a RepositoryTrackCommitService with one rpc named GetRepositoryTrackCommitForCommitReference. GetRepositoryTrackCommitForCommitReference will be used by the bsr front end to get a commit's sequence id on a track.


proto/buf/alpha/registry/v1alpha1/repository_track_commit.proto
```diff
@@ -0,0 +1,54 @@
...
+
+  // immutable
+  google.protobuf.Timestamp create_time = 2;
+
+  // We reserve field number '3' for the update_time.
+  // google.protobuf.Timestamp update_time = 3;
+
+  // unique for repository_track, immutable
+  int64 sequence_id = 4;
+}
```
amckinney, line 35: I suppose we don't need it to start, but I imagine we might need this object to include the ID for the commit it's associated with. You could say the same for the track's ID.

In short, we effectively want to expose the public-equivalent of our internal representation. 

```
message RepositoryTrackCommit {
  // immutable
  bytes repository_track_entity_id = 1;
  // immutable
  google.protobuf.Timestamp create_time = 2;
  // mutable
  google.protobuf.Timestamp update_time = 3;
  // unique for repository_track_entity_id, immutable
  bytes repository_commit_entity_id = 4;
  // unique for repository_track_entity_id, immutable
  int64 sequence_id = 5;
}
```

So in this case, the entity IDs, would be the public IDs.

proto/buf/alpha/registry/v1alpha1/repository_track_commit.proto
```diff
@@ -0,0 +1,78 @@
...
+  string repository_track_name = 2;
+  // The commit reference to resolve.
+  string commit = 3;
+}
+
+message GetRepositoryTrackCommitByCommitReferenceResponse {
+  RepositoryTrackCommit repository_track_commit = 1;
+}
+
+message ListRepositoryTrackCommitsByRepositoryTrackRequest {
```
amckinney, line 65: I think we normally keep it consistent for IDs with other IDs, and names with other names. Should we make this act upon `repository_id` and `repository_track_id`? Then we could reserve another `ListRepositoryTrackCommitsByRepositoryTrackNameRequest` endpoint that works upon `repository_name` and `repository_track_name` later (if ever).

What are your thoughts? This isn't necessarily hard blocking, but something we should consider before we move forward with the API.

amckinney: I guess you could say the same about the `Get` endpoint above (with names mixed with IDs). Maybe this isn't so bad. Let's look at the other APIs to make sure.

amckinney (approved):
Much simpler, well done 👏

PR #904, opened 2022-02-01
Remove unused action enum

mfridman:
This PR removes (and reserves) an action enum. This was deprecated in favor of a single endpoint in:

https://github.com/bufbuild/buf/pull/855/files#diff-928df8dd916324f95c5e6588f973004810b6e7af9c787a58cea921a46173a758R52-R57


johanbrandhorst (commented):
I don't think we want to remove it from here. This is a permanent record of methods, and may appear in persisted messages, so we should leave it in, IMO.

mfridman:
Sounds good, thanks for pointing that out. Closing.

Issue #861, opened 2022-01-17
Validate token during buf registry login


robbertvanginkel:
When using `buf registry login`, we currently only display a message that the credentials are saved:
```
$ buf registry login
Login with your Buf Schema Registry username. If you don't have a username, head over to https://buf.build to create one.

Username: user
Token:
Credentials saved to /Users/robbert/.netrc.
```

We had a user over the weekend be confused about a seemingly successful `buf registry login` not giving them access to their modules (they received access denieds on their queries). It turned out that their token was expired, they suggested it would have been helpful if `buf registry login` had warned them about the login failing.

mfridman:
What are your thoughts if we kept the existing behaviour as is--no registry calls and continue writing out to netrc file. But offered a flag such as `--verify` or `--validate` to trigger an extra token verification step.

robbertvanginkel:
I think that would be appropriate for a command named `credential-write` or something, but a successful "login" to me convey's that its validated the credentials I've provided.

mfridman:
Great point, agreed.

johanbrandhorst:
I think we could use the existing `GetCurrentUser` endpoint for this.

PR #905, opened 2022-02-01
Check user provided token during `buf registry login`

doriable:
Check the token and username provided when calling `buf registry login`.
If the token is invalid, we return an error and if the user resolved from the token does not match the username provided, we return an error.

Fixes: #861

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -25,6 +25,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/netrc"
+	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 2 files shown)


private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -128,6 +129,26 @@ func run(
...
+	if err != nil {
+		// We don't want to use the default error from wrapError here if the error
+		// an unauthenticated error.
+		return errors.New("invalid token provided")
+	}
+	if user == nil {
+		return errors.New("no user found for provided token")
+	}
+	if user.Username != username {
+		return fmt.Errorf("token user does not match username provided: %s", username)
```
amckinney, line 150: nit:
```suggestion
		return fmt.Errorf("the username associated with that token (%s) does not match the username provided (%s)", user.Username, username)
```

We _might_ want to let the user know what the resolved `user.Username` was so that it's clear what the difference is. At first I thought that might be leaking information, but the user can discover who's token it is by other means so this seems relatively harmless (but helpful).

amckinney (approved):
Looks good to me - small nit on the error message. We might want to update the `CHANGELOG.md` for this one, too.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
doriable: Yeah, that's a good point -- I went back and forth on this also, for the exact reason that you stated around leaking information. But yeah, you're right, ultimately this is something that can already be discovered (by basically make this exactly call, lol), so it's probably fine. Will update this and the `CHANGELOG` :D

Issue #829, opened 2022-01-05
Investigate adding JUnit format for errors


bufdev:

hborham:
Many CI tools have Junit Parsers available to parse output and present it nicely for a build result:  TeamCity, Jenkins, and Bitbucket  to name a few.

Buf Lint reporting would be easily consumable then by CI's and users if we had an output option for junit xml.

srikrsna-buf:
This is a draft of the mapping of annotation to `jUnit` xml can be.

- Starts with a `testsuites`. All output for a command like `lint` will be under one `testsuites` tag.
- Each file will map to a `testsuite` with `name`, `file` attributes mapping to the file path.
- `tests` and `failures` will always be equal and map to number of annotations in each file.
- Each type will map to a `testcase` tag with `name` attribute filled with `type`
- Each `testcase` exactly contains one `failure` tag containing the error message.

Example:
```xml
<testsuites>
    <testsuite name="foo/foo.proto" tests="3" failures="3" errors="0" file="foo/foo.proto">        
        <testcase name="PACKAGE_VERSION_SUFFIX">
            <failure>Package name "foo" should be suffixed with a correctly formed version, such as "foo.v1"</failure>
        </testcase>
        <testcase name="FIELD_LOWER_SNAKE_CASE">
            <failure>Message name "snake_case_message" should be PascalCase, such as "SnakeCaseMessage"</failure>
        </testcase>
        <testcase name="FIELD_LOWER_SNAKE_CASE">
            <failure>Field name "CamelCaseField" should be lower_snake_case, such as "camel_case_field"</failure>
        </testcase>
    </testsuite>
</testsuites>
```

Alternative simpler approach without `file` attribute:
```xml
<testsuite name="buf-lint" tests="3" failures="3" errors="0">
    <testcase name="foo/foo.proto">
        <failure type="PACKAGE_VERSION_SUFFIX">Package name "foo" should be suffixed with a correctly formed version, such as "foo.v1"</failure>
        <failure type="MESSAGE_PASCAL_CASE">Message name "snake_case_message" should be PascalCase, such as "SnakeCaseMessage"</failure>
        <failure type="FIELD_LOWER_SNAKE_CASE">Field name "CamelCaseField" should be lower_snake_case, such as "camel_case_field"</failure>
    </testcase>
</testsuite>
```

Both of these lack line and column numbers. Can be added to `testsuite` tag using `properties` tag. Not sure if that will be useful.

[...]

PR #907, opened 2022-02-03
Support error formats that require all annotations to print

srikrsna-buf:
Related to #829 

The current approach formats file annotations one after the other joining them with a new line. This works for the current set of formats (text, msvs, and json). However, it falls short when trying to print in a format such as `jUnit`. Here is a sample `jUnit` xml:

```xml
<testsuite name="buf-lint" tests="3" failures="3" errors="0">
    <testcase name="foo/foo.proto">
        <failure type="PACKAGE_VERSION_SUFFIX">Package name "foo" should be suffixed with a correctly formed version, such as "foo.v1"</failure>
        <failure type="MESSAGE_PASCAL_CASE">Message name "snake_case_message" should be PascalCase, such as "SnakeCaseMessage"</failure>
        <failure type="FIELD_LOWER_SNAKE_CASE">Field name "CamelCaseField" should be lower_snake_case, such as "camel_case_field"</failure>
    </testcase>
</testsuite>
```
To support such formats, this PR refactors `bufanalysis` package to use a new `printer` interface. Each `Format` gets a printer. This removes methods used for formatting from `FileAnnotation` interface. All other public api is unchanged.


private/bufpkg/bufanalysis/bufanalysis.go
```diff
@@ -64,6 +64,11 @@ var (
 		FormatJSON: "json",
 		FormatMSVS: "msvs",
 	}
+	formatToPrinter = map[Format]printer{
+		FormatText: printerFunc(textPrinter),
+		FormatJSON: printerFunc(jsonPrinter),
+		FormatMSVS: printerFunc(msvsPrinter),
+	}
```
doriable, line 71: Instead of having this be a global, we could rework `FormatFileAnnotation` with the `switch` statement to instead operate on the entire set of file annotations, `[]FileAnnotation`?

The concern is for the `junit` implementation, where we'll need all the `FileAnnotations` to print, so the print logic would need to be one level up from the single `FileAnnotation` unit, so by reworking the logic such that we handle the entire slice of `[]FileAnnotation` given the `formatString` provided to `PrintAllFileAnnotations` should be sufficient.

private/bufpkg/bufanalysis/bufanalysis.go
```diff
@@ -200,34 +196,28 @@ func PrintFileAnnotations(writer io.Writer, fileAnnotations []FileAnnotation, fo
...
-		s, err := FormatFileAnnotation(fileAnnotation, format)
-		if err != nil {
-			return err
-		}
-		if _, err := writer.Write([]byte(s + "\n")); err != nil {
-			return err
-		}
-	}
-	return nil
+	return formatToPrinter[format].PrintFileAnnotations(
```
bufdev, line 199: Need to check to make sure printer from map is not nil, and if it is, error

private/bufpkg/bufanalysis/bufanalysis.go
```diff
@@ -200,34 +196,28 @@ func PrintFileAnnotations(writer io.Writer, fileAnnotations []FileAnnotation, fo
...
+	)
 }
 
 // FormatFileAnnotation formats the FileAnnotation.
 func FormatFileAnnotation(fileAnnotation FileAnnotation, format Format) (string, error) {
+	buffer := bytes.NewBuffer(nil)
 	switch format {
 	case FormatText:
-		return fileAnnotation.String(), nil
+		_ = printFileAnnotationAsText(buffer, fileAnnotation)
```
bufdev, line 210: You can't ignore this error, you need to propagate it

private/bufpkg/bufimage/bufimagebuild/builder_test.go
```diff
@@ -374,7 +374,7 @@ func testFileAnnotations(t *testing.T, relDirPath string, want ...string) {
 	_, fileAnnotations := testBuild(t, false, filepath.Join("testdata", filepath.FromSlash(relDirPath)))
 	got := make([]string, len(fileAnnotations))
 	for i, annotation := range fileAnnotations {
-		got[i] = annotation.String()
+		got[i], _ = bufanalysis.FormatFileAnnotation(annotation, bufanalysis.FormatText)
```
bufdev, line 377: check the error with `assert.NoError(t, err)`

private/bufpkg/bufanalysis/bufanalysis.go
srikrsna-buf: Even if `printFileAnnotationAsText` always returns `nil` similar to `bytes.Buffer`?

private/bufpkg/bufanalysis/bufanalysis.go
srikrsna-buf: I think we can remove `FormatFileAnnotation` altogether, it is only being used in tests. 

IMO separating the printing part from `FileAnnotation` is useful as the formats can grow. Instead of modifying the `FileAnnotation` interface we can just add a new `Format` and `printer`. 

The `FileAnnotation` can have a default format in the form of `Stringer` interface (used in tests of `bufimagebuild`).

Not that it is major concern here, but this is slightly faster than using `FormatFileAnnotation` to return a `string` as we can reuse the `bytes.Buffer`

doriable: So I think that we probably want to prevent `Format` from becoming more than just the enum for managing format types (which this current design adheres to), however, we should also avoid this global here if it can be handled by a `switch`. We should also avoid attaching the `printer` interface to the `Format`, even though they are 1:1 since that logic can be encapsulated into a single `Print` function that takes in the format and file annotations (without defining a new structure). I agree with the decision to probably get rid of `FormatFileAnnotation`, since for the `junit` output in particular, we will need to print all `FileAnnotation`s together, so formatting each file annotation separately doesn't really make much sense anymore.

doriable (approved):
Looks good to me, and I also pulled it and did a couple of tests against this :)

private/bufpkg/bufanalysis/bufanalysistesting/bufanalysistesting_test.go
```diff
@@ -24,43 +25,52 @@ import (
...
+	assert.Equal(
 		t,
-		"path/to/file.proto",
-		1,
-		0,
-		1,
-		0,
-		"FOO",
-		"Hello.",
+		`path/to/file.proto:1:1:Hello.`+"\n"+
```
bufdev, line 55: You don't need to do this, you can just do:

```
`This
is how you do multiline strings
and more`
```

private/bufpkg/bufanalysis/bufanalysis.go
```diff
@@ -103,12 +102,6 @@ type FileInfo interface {
 type FileAnnotation interface {
 	// Stringer returns the string representation in text format.
 	fmt.Stringer
```
bufdev, line 104: Should we kill this as well?

private/bufpkg/bufanalysis/bufanalysistesting/bufanalysistesting_test.go
srikrsna-buf: Tried that, gofmt is formatting it like this,
```go
	assert.Equal(
		t,
		`path/to/file.proto:1:1:Hello.
path/to/file.proto:2:1:Hello.
`,
		sb.String(),
	)
```

private/bufpkg/bufanalysis/bufanalysis.go
srikrsna-buf: Stringer is useful (in one test so far) and generic enough (not an encoding format) to be part of the interface. I think we should let this be.

private/bufpkg/bufanalysis/bufanalysistesting/bufanalysistesting_test.go
bufdev: Yep thats right

private/bufpkg/bufanalysis/bufanalysis.go
bufdev: It's saying that this is how you do the text format though - we should at least update the comment to not reference the text format and not tie it to the text format

srikrsna-buf: Updated the comment

private/bufpkg/bufanalysis/bufanalysis.go
bufdev: If it will always return nil, then the function should not have an error signature. I think what you mean is that writes to the `buffer` will never result in error, so this function, due to how we happen to be calling it, will always return nil. That's a big usage assumption IMO - if we later change from `bytes.Buffer` to something else that can error, we are then ignoring an error. We should always check errors such as these.

bufdev:
Do we still want to pursue this?

doriable:
> Do we still want to pursue this?

From what I can see, after merging this PR, the bulk of the remaining work would be implementing a junit printer, with the above proposed layout (which @srikrsna-buf has solicited the opinions of other team members and the general consensus seems like this is a sane test output for junit).

I think this should be pretty quick to close out, but I know there's also a lot on @srikrsna-buf's plate right now. If he'd like to close it out, I'll let him comment, otherwise, I'm happy to add it to my to-do list.

srikrsna-buf:
@doriable Thank you for the response and the offer. We can merge this without any changes. 

I'll implement the `juit` after this. Although it will not be my priority for at least 2 weeks. Feel free to take it up @doriable

srikrsna-buf:
@doriable Thank you for the response and the offer. We can merge this without any changes. 

I'll implement the `juit` after this. Although it will not be my priority for at least 2 weeks. Feel free to take it up @doriable

PR #912, opened 2022-02-03
pkg/thread: use GOMAXPROCS instead of NumCPU

akshayjshah:
In modern production environments, NumCPU is a poor estimate of the
actual parallelism available: it does consult the population count of
the cpuset mask (if any), but doesn't account for cgroups v1 or v2. Both
are commonly used in containerized environments.

The intention of this code is (I think?) to prevent us from burning
memory on initial stacks for zillions of goroutines, so we should
just use GOMAXPROCS: by definition, it's the maximum number of
goroutines that can run in parallel.

### DIFF
--- a/private/pkg/thread/thread.go
+++ b/private/pkg/thread/thread.go
@@ -23,7 +23,7 @@ import (
 )
 
 var (
-	globalParallelism = runtime.NumCPU()
+	globalParallelism = runtime.GOMAXPROCS(0)
 	globalLock        sync.RWMutex
 )
 


amckinney (approved):
The way you've lined it out sounds reasonable to me, but we'll need @bufdev's review on this one.

akshayjshah:
Clarifying somewhat: GOMAXPROCS is also often wrong in these environments, but at least it's directly tied to the number of goroutines that the runtime will allow to run in parallel. When it's wrong, it's fixable by operators.

bufdev:
Maybe should connect quickly - give me an example, so say on my MBP with 8 cores hyperthreaded, will GOMAXPROCS be 16, or 3000? I'm reading the above both ways heh.

bufdev:
I think this answers the question: https://pkg.go.dev/runtime#GOMAXPROCS

Tldr, GOMAXPROCS defaults to runtime.NumCPU. Assuming this is true, this is likely all OK.

The main thing to point out is that the code (specifically the image building code) is currently optimized to assume it's actually only going to run a max of one thread per CPU, and doing more than that causes a significant performance hit. Might be worth grepping for other usages of runtime.NumCPU in this codebase.

doriable (approved):

akshayjshah:
> Tldr, GOMAXPROCS defaults to runtime.NumCPU. Assuming this is true, this is likely all OK.

Yep, this is correct.

> The main thing to point out is that the code (specifically the image building code) is currently optimized to assume it's actually only going to run a max of one thread per CPU, and doing more than that causes a significant performance hit. Might be worth grepping for other usages of runtime.NumCPU in this codebase.

Definitely makes sense to avoid pointless over-parallelization.

PR #913, opened 2022-02-03
pkg/thread: fix select precedence

akshayjshah:
We want context cancellation or timeout to always take precedence over
an unblocked semaphore, so we never start goroutines that can't
successfully complete their work. We're very close to that goal right
now, but there's a timing bug in our logic.

Select chooses pseudorandomly from the unblocked, non-default cases. No
matter what checks we do beforehand, code like this isn't quite correct:

```go
select {
case <-ctx.Done():
  // abort
case <-semaphoreC:
  // work
}
```

Under load, it's likely that a fair bit of time has elapsed between the
previous checks and this select; since the goroutine is blocked on the
channel reads, it gets parked and some other goroutine has a chance to
run. Once this goroutine is unblocked and runs, it's entirely possible
that _both_ the `ctx.Done()` and `semaphoreC` reads will succeed. In
that circumstance, we don't want to choose randomly! Instead, we want
this:

```go
select {
case <-ctx.Done():
  // abort
case <-semaphoreC:
  select {
  case <-ctx.Done():
    // abort
  default:
    // work
  }
}
```

Now, we're guaranteed to do work only if the semaphore is unblocked and
the context is blocked. This is similar to how RWMutexes are used: once
you acquire the write lock, you typically need to re-check that the
write is valid before mutating data.

### DIFF
--- a/private/pkg/thread/thread.go
+++ b/private/pkg/thread/thread.go
@@ -63,41 +63,35 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
 		return nil
 	case 1:
 		return jobs[0](ctx)
-	default:
-		multiplier := parallelizeOptions.multiplier
-		if multiplier < 1 {
-			multiplier = 1
+	}
+	multiplier := parallelizeOptions.multiplier
+	if multiplier < 1 {
+		multiplier = 1
+	}
+	semaphoreC := make(chan struct{}, Parallelism()*multiplier)
+	var retErr error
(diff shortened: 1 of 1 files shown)


bufdev (approved):
Whoops - this was assumedly my fault.

akshayjshah:
> Whoops - this was assumedly my fault.

No idea, but it makes me feel better about the style guide's recommendation to use this package. This stuff gets complicated fast, and it's easy to make mistakes.

PR #914, opened 2022-02-03
pkg/thread: add a simple test

akshayjshah:
It's difficult to test all the timing conditions in Parallelize, but we
can easily test the most basic use cases.

### DIFF
--- a/private/pkg/thread/thread_test.go
+++ b/private/pkg/thread/thread_test.go
@@ -1,0 +1,41 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


amckinney (approved):

johanbrandhorst (approved):

private/pkg/thread/thread_test.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.uber.org/atomic"
+)
+
+func TestParallelizeWithImmediateCancellation(t *testing.T) {
+	// The bulk of the code relies on subtle timing that's difficult to
```
bufdev, line 26: Needs a t.Parallel(), im surprised lint didnt catch this

akshayjshah: Sure, can do.

akshayjshah: I'm tempted to make one of those Xzibit memes about parallel testing for parallel, but even reading the template confuses me.

PR #917, opened 2022-02-04
Append commits to tracks

WillAbides:
This adds `buf beta registry track append` which lets you add an existing commit to a track.


bufdev (changes requested):
I don't think we should allow arbitrary appends of commits to tracks

amckinney:
I agree - we might eventually want a way to copy one track's contents to another one at initialization time (like we discussed), but we don't need to support appending an already-created commit to another track from the CLI.

amckinney:
Does this tie into the CI/CD strategy we have planned? If so, it'd be great to see how we plan to use it. That way we can see if there are alternatives, etc.

WillAbides:
This is part of the plan for tracks. It is needed for the case where the same git commit appears on multiple branches that are shadowed by tracks. When it is pushed to the first git branch, CI will push it to BSR and tag it with the git commit sha. When it is pushed to the second git branch, CI won't be able to tag a second BSR commit with the same git commit sha. Putting the same BSR commit on multiple tracks was our way around that.

bufdev:
You should be able to derive what branches a git commits should be pushed to in a single `buf push` call, without the need for another command that we don't want exposed to the user

WillAbides:
I'll try to make that work

PR #925, opened 2022-02-08
add runtime decoding with source and type provided

cyinma:
This add `buf beta decode` command to decode message with provided image source and message type.

This only accept stdin as the input.
```
cat pet.plain.bin | buf beta decode buf.build/acme/petapis --type pet.v1.Pet
```

The output can be stdout or a file
```
# default is stdout
cat pet.plain.bin | buf beta decode buf.build/acme/petapis --type pet.v1.Pet
# stdout with a dash as output flag
cat pet.plain.bin | buf beta decode buf.build/acme/petapis --type pet.v1.Pet -o -
# file output
cat pet.plain.bin | buf beta decode buf.build/acme/petapis --type pet.v1.Pet -o result.json
```

This support using module reference, image bin/json, directory, git, or archive as source
```
# module reference
cat pet.plain.bin | buf beta decode buf.build/acme/petapis --type pet.v1.Pet
# image bin
cat pet.plain.bin | buf beta decode image.bin --type pet.v1.Pet
# image json
cat pet.plain.bin | buf beta decode image.json --type pet.v1.Pet
# directory
cat pet.plain.bin | buf beta decode proto/petapis --type pet.v1.Pet
# git
cat pet.plain.bin | buf beta decode https://github.com/bufbuild/buf-tour.git#subdir=start/petapis --type pet.v1.Pet
# archive
cat pet.plain.bin | buf beta decode https://github.com/bufbuild/buf-tour/archive/main.zip#strip_components=1,subdir=start/petapis --type pet.v1.Pet
```

This also support fully-qualified path and reference as type without providing source
```
# default main track
cat pet.plain.bin | buf beta decode --type buf.build/tommyma/petapis#pet.v1.Pet | jq
{
  "petId": "123",
  "name": "Ekans"
}

# specified main track 
cat pet.plain.bin | buf beta decode --type buf.build/tommyma/petapis:main#pet.v1.Pet | jq
{
  "petId": "123",
  "name": "Ekans"
}

# specified dev track
cat pet.plain.bin | buf beta decode --type buf.build/tommyma/petapis:dev#pet.v1.Pet | jq
{
  "petId": "123",
  "petName": "Ekans"
}

# specified commit
cat pet.plain.bin | buf beta decode --type buf.build/tommyma/petapis:d318898ca58f4c5ca9f8a113d3da7d91#pet.v1.Pet | jq
{
  "petId": "123",
  "petName": "Ekans"
}

```


private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,61 @@
...
+	case protoreflect.MessageDescriptor:
+		return dynamicpb.NewMessage(typedDescriptor), nil
+	default:
+		return nil, fmt.Errorf("%q must be a message but is a %T", typeName, typedDescriptor)
+	}
+}
+
+// ValidateTypeName validates that the typeName is well-formed, such that it has one or more
+// '.'-delimited package components and no '/' elements.
+func ValidateTypeName(typeName string) error {
```
amckinney, line 56: We can un-export this now that we don't use this anywhere else than `NewMessage` above.

private/buf/bufcli/bufcli.go
```diff
@@ -725,6 +726,103 @@ func ReadModuleWithWorkspacesDisabled(
...
+	images := make([]bufimage.Image, 0, len(imageConfigs))
+	for _, imageConfig := range imageConfigs {
+		images = append(images, imageConfig.Image())
+	}
+	return bufimage.MergeImages(images...)
+}
+
+// NewDescriptorBytesForInput returns the descriptor bytes for an input, which can be
+// from stdin, argument, or a file.
+func NewDescriptorBytesForInput(
```
amckinney, line 786: Nicely done, this is great.

amckinney (commented):
This is looking great so far!

private/buf/bufcli/bufcli.go
```diff
@@ -725,6 +738,103 @@ func ReadModuleWithWorkspacesDisabled(
...
+	case buffetch.SourceRef:
+		return []byte(input), nil
+	case buffetch.ImageRef:
+		path, err := ref.PathForExternalPath(input)
+		if err != nil {
+			return nil, err
+		}
+		if path == "-" {
+			return io.ReadAll(container.Stdin())
+		}
```
amckinney, line 818: I don't actually think you need to use `ref.PathForExternalPath` for this one. We're only checking whether or not the user specified an explicit `-` for their input right? `PathForExternalPath`  is only used to map an external path (e.g. `proto/acme/weather/v1/weather.proto` to the path relative to the module root, such as `acme/weather/v1/weather.proto` if the root was `proto`).

I'd just remove the `ref.PathForExternalPath` call and check `if input == "-" {`

private/buf/bufwire/bufwire.go
```diff
@@ -202,3 +203,27 @@ func NewImageWriter(
 		fetchWriter,
 	)
 }
+
+// ProtoEncodingWriter is a proto encoding writer.
+type ProtoEncodingWriter interface {
+	// PutMessage writes the message to the value.
```
amckinney, line 209:
```suggestion
	// PutMessage writes the message to the path.
```

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,167 @@
...
+) (protoSource string, protoType string, _ error) {
+	if flagSource != "" && flagType != "" {
+		return flagSource, flagType, nil
+	}
+	if flagType == "" {
+		return "", "", appcmd.NewInvalidArgumentError("type is required")
+	}
+	moduleName, typeName, err := bufreflect.ParseFullyQualifiedPath(flagType)
+	if err != nil {
+		return "", "", err
```
amckinney, line 164: We might need to include a better error message here. If the user specifies a `--type` without a `--source`, we're making the assumption that the `--type` is a fully-qualified type name. If it's not, let's describe this implicit behavior in the error message.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,167 @@
...
+		container.Logger(),
+	).PutMessage(
+		ctx,
+		container,
+		message,
+		flags.Output,
+	)
+}
+
+func parseSourceAndType(
```
amckinney, line 151: It'd be good to leave a lightweight comment about what's happening here (similar to the error message comment below).

It'd also be good to outline that case we discussed earlier, i.e. if someone redundantly specifies `--source buf.build/acme/weather --type buf.build/acme/weather/weather.v1.Units`.

private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	switch typedDescriptor := descriptor.(type) {
+	case protoreflect.MessageDescriptor:
+		return dynamicpb.NewMessage(typedDescriptor), nil
+	default:
+		return nil, fmt.Errorf("%q must be a message but is a %T", typeName, typedDescriptor)
+	}
+}
+
+func ParseFullyQualifiedPath(
```
amckinney, line 56: Don't forget to leave a comment for this one.

private/bufpkg/bufreflect/bufreflect_test.go
```diff
@@ -0,0 +1,40 @@
...
+package bufreflect_test
+
+import (
+	"testing"
+
+	"github.com/bufbuild/buf/private/bufpkg/bufreflect"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseFullyQualifiedModuleTypeName(t *testing.T) {
```
amckinney, line 24: This test name should be updated to match the function name.
```suggestion
func TestParseFullyQualifiedPath(t *testing.T) {
```

private/buf/bufwire/bufwire.go
```diff
@@ -202,3 +203,27 @@ func NewImageWriter(
...
+	//
+	// Currently, this only support json format.
```
johanbrandhorst, line 211: What is "value"? Where is the format specified?

private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,79 @@
...
+		return "", "", appcmd.NewInvalidArgumentError("you must specify a module type name")
+	}
+	components := strings.Split(moduleTypeName, "/")
+	if len(components) != 4 {
+		return "", "", appcmd.NewInvalidArgumentErrorf("%q is not a valid fully qualified path", moduleTypeName)
+	}
+	if err := validateTypeName(components[3]); err != nil {
+		return "", "", err
+	}
+	return moduleTypeName[:strings.LastIndex(moduleTypeName, "/")], components[3], nil
```
amckinney, line 69: I think we should use `moduleref.ModuleIdentity` or `moduleref.ModuleIdentityForString` to get some extra validation before returning the `moduleName` here. That would be more effective than just returning `moduleTypeName[:strings.LastIndex(moduleTypeName, "/")]`.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,167 @@
...
+		"",
+		"The fully-qualified type name of the serialized descriptor (e.g. acme.weather.v1.Units)",
+	)
+	flagSet.StringVarP(
+		&f.Output,
+		outputFlagName,
+		outputFlagShortName,
+		"-",
+		// TODO: If we ever support other formats (e.g. prototext), we will need
+		// to build a buffetch.DecodeOutputRefParser.
```
amckinney, line 101: Now that we're modeling this for both `encode` and `decode`, might be better to say something like this:
```suggestion
		// to build a buffetch.ProtoEncodingRefParser.
```

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	image, err := bufcli.NewImageForSource(ctx, container, registryProvider, protoSource, flags.ErrorFormat)
+	if err != nil {
+		return err
+	}
+	message, err := bufreflect.NewMessage(ctx, image, protoType)
+	if err != nil {
+		return err
+	}
+	if err := proto.Unmarshal(descriptorBytes, message); err != nil {
```
johanbrandhorst, line 138: Should this use our `protoencoding` marshaler?

private/buf/cmd/buf/testdata/decode/descriptor.plain.bin
```diff
@@ -0,0 +1 @@
+7
```
johanbrandhorst, line 1: 😂 love this

private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,79 @@
...
+	descriptor, err := files.FindDescriptorByName(protoreflect.FullName(typeName))
+	if err != nil {
+		return nil, err
+	}
+	switch typedDescriptor := descriptor.(type) {
+	case protoreflect.MessageDescriptor:
+		return dynamicpb.NewMessage(typedDescriptor), nil
+	default:
+		return nil, fmt.Errorf("%q must be a message but is a %T", typeName, typedDescriptor)
+	}
```
johanbrandhorst, line 53: It's a bit unusual to use a type switch for a single type like this, how about:
```suggestion
	typedDescriptor, ok := descriptor.(protoreflect.MessageDescriptor) {
	if !ok {
		return nil, fmt.Errorf("%q must be a message but is a %T", typeName, typedDescriptor)
	}
	return dynamicpb.NewMessage(typedDescriptor), nil
```

johanbrandhorst (commented):
> ```
> # argument input
> buf beta decode "$(cat pet.plain.bin)" --source buf.build/acme/petapis --type pet.v1.Pet
> # file input
> buf beta decode pet.plain.bin --source buf.build/acme/petapis --type pet.v1.Pet
> ```

Supporting both of these requires some heuristic to tell the difference between a binary literal and a file argument. Do we really need to support
```
 buf beta decode "$(cat pet.plain.bin)" --source buf.build/acme/petapis --type pet.v1.Pet
```
when we have
```
cat pet.plain.bin | buf beta decode --source buf.build/acme/petapis --type pet.v1.Pet
```
?

> This also support fully-qualified path as type
> 
> ```
> buf beta decode pet.plain.bin --type buf.build/acme/petapis/pet.v1.Pet
> ```

Should this include the track as well?

```
buf beta decode pet.plain.bin --type buf.build/acme/petapis/main/pet.v1.Pet
```

Our URL paths include `file/track`: https://buf.build/bufbuild/buf/file/main/buf/alpha/registry/v1alpha1/user.proto

amckinney:
> Should this include the track as well?

We should default to `main` when one isn't specified, but supporting a track element is also necessary.

So we should be able to do all of the following

```
$ buf decode --type buf.build/acme/weather      // main track
$ buf decode --type buf.build/acme/weather/main // main track
$ buf decode --type buf.build/acme/weather/dev  // dev track
```

You should be able to use `moduleref.ModuleReference` for this stuff.

private/buf/cmd/buf/command/decode/decode.go
amckinney: Yes, and we should use `protoencoding.NewJSONMarshalerIndent` for the `proto.Marshal` end.

private/buf/bufwire/bufwire.go
amckinney: Commented on this above - the comment should be updated to be `path`, not `value`.

private/bufpkg/bufreflect/bufreflect.go
amckinney: From the comment below, see `moduleref.ModuleReference` to support tracks, tags, etc.

private/buf/bufcli/bufcli.go
```diff
@@ -725,6 +738,103 @@ func ReadModuleWithWorkspacesDisabled(
...
+	registryProvider registryv1alpha1apiclient.Provider,
+	input string,
+) ([]byte, error) {
+	ref, err := buffetch.NewRefParser(container.Logger(), buffetch.RefParserWithProtoFileRefAllowed()).GetRef(ctx, input)
+	if err != nil {
+		return nil, err
+	}
+	switch t := ref.(type) {
+	case buffetch.SourceRef:
+		return []byte(input), nil
```
amckinney, line 810: This is related to Johan's point with respect to supporting `buf decode "$(cat ...)"` instead of forcing the user to do `cat ... | buf decode`. We're kind of abusing the `buffetch.SourceRef` as the default behavior when in reality we wouldn't actually read the ref if it was a regular file.

For example, with `buf decode foo.txt` we would interpret the descriptor bytes as `foo.txt` instead of reading the file.

Overall, I think Johan is right - it's probably easiest to only support reading on `stdin` and forcing the user to pipe in the information themselves. It's already an input mode we must support, so we can start with less.

private/buf/bufcli/bufcli.go
```diff
@@ -725,6 +738,103 @@ func ReadModuleWithWorkspacesDisabled(
...
+	case buffetch.ImageRef:
```
amckinney, line 811: Similar to the above, this is actually abusing the `.bin` extension a bit since that's reserved for Image files. But this isn't actually an Image (which is a `FileDescriptorSet`) - this is a binary-encoded Protobuf message.

All of this works because we're able to use `GetImageFile` and that just reads the content directly. The rest of the `moduleResolver`, etc. aren't actually ever used.

Again, it's probably easiest just to require that the user provides the bytes on `stdin`, and we leave it at that for now.

amckinney:
Ugh, you're right about the ambiguity. I just hate that `main` would need to be an explicit part of the path. But now that I'm looking at it more, we might require that the track is delimited by a `:`. It doesn't map to the URL you pasted earlier, but it _does_ map to the representation in the user's `buf.yaml` (i.e. `deps`).

What are your thoughts?

amckinney:
But in this case, we're not actually adding directory elements - we're only referring to the package which is defined in the module's top-level scope.

So you'd never have

```
$ buf decode --type buf.build/acme/weather/dev/v1/dev.proto
```.

You'd only have

```
$ buf decode --type buf.build/acme/weather/dev/dev.v1.Something
```

The last element is always a single component (no `/` elements).

**Comment by johanbrandhorst** [comment]
> You'd only have
> 
> ```
> $ buf decode --type buf.build/acme/weather/dev/dev.v1.Something
> ```
> 
> The last element is always a single component (no `/` elements).

Ah, that might work!

**Comment by amckinney** [comment]
@cyinma try to use `bufmoduleref.ModuleReference` and see if we can support just a module identity that defaults to the `main` track (e.g. `buf.build/acme/weather`) without any ambiguity. To be clear, let's see if we can get all of these different forms working with zero chance of confusion:

```
$ buf decode --type buf.build/acme/weather/weather.v1.Units      // main track
$ buf decode --type buf.build/acme/weather/main/weather.v1.Units // main track
$ buf decode --type buf.build/acme/weather/dev/weather.v1.Units  // dev track
```

**Comment by cyinma** [comment]
after the discussion with @amckinney , it seems like it would be clearest to use `:` delimiter for reference and `#` delimiter for type, and it will be:

```
$ buf decode --type buf.build/acme/weather#weather.v1.Units      // main track
$ buf decode --type buf.build/acme/weather:main#weather.v1.Units // main track
$ buf decode --type buf.build/acme/weather:dev#weather.v1.Units  // dev track
```
@johanbrandhorst how do you think?

File: private/buf/cmd/buf/command/decode/decode.go [hunk 17]
```diff
@@ -0,0 +1,171 @@
...
+// When source and type are redundantly specified (e.g. buf.build/acme/weather and buf.build/acme/weather#weather.v1.Units),
+// we are not handling it and just let it fail.
+func parseSourceAndType(
+	ctx context.Context,
+	flagSource string,
+	flagType string,
+) (moduleReference string, typeName string, _ error) {
+	if flagSource != "" && flagType != "" {
+		return flagSource, flagType, nil
+	}
```
amckinney, line 162: We should still be validating the `flagType` here according to the `ValidateTypeName` helper. We validate it in `bufreflect.ParseFullyQualifiedPath`, so we should make the same validation guarantee for all cases in this function.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,171 @@
...
+		container,
+		message,
+		flags.Output,
+	)
+}
+
+// parseSourceAndType returns the moduleReference and typeName from the source and type provided by the user.
+// When source is not provided, we assume the type is a fully-qualified path to the type and try to parse it.
+// When source and type are redundantly specified (e.g. buf.build/acme/weather and buf.build/acme/weather#weather.v1.Units),
+// we are not handling it and just let it fail.
```
amckinney, line 154: Reviewing this comment some more - it's impossible to have `/` elements in the `--type` component (it wouldn't be a valid full type name), so we might be able to just get away with some extra validation around this (re: the comment below).

I think we can rephrase this to:
```suggestion
// parseSourceAndType returns the moduleReference and typeName from the source and type provided by the user.
// When source is not provided, we assume the type is a fully-qualified path to the type and try to parse it.
// Otherwise, if both source and type are provided, the type must be a valid Protobuf identifier (e.g. weather.v1.Units).
```

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,171 @@
...
+		&f.Source,
+		sourceFlagName,
+		"",
+		"The source that defines the serialized descriptor (e.g. buf.build/acme/weather)",
+	)
+	flagSet.StringVar(
+		&f.Type,
+		typeFlagName,
+		"",
+		"The fully-qualified type name of the serialized descriptor (e.g. acme.weather.v1.Units)",
```
amckinney, line 94: We might need to add some more information to this flag so that it's clear people can specify the fully-qualified path (including the module reference prefix).

We can work with @lucperkins on the `--help` text later on (in a separate PR) to see if we can make any other improvements.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,171 @@
...
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <descriptor>",
+		Short: "Decode binary descriptors with a source reference.",
+		Long: `The first argument is the serialized descriptor to decode.
+If no argument is specified, defaults to stdin.`,
```
amckinney, line 51: We also need to update this now that we require the message on stdin.

We should rephrase it to refer to a serialized _message_ instead of a descriptor (my fault on that earlier).

private/buf/cmd/buf/command/decode/decode.go
cyinma: sounds good to me! I am thinking about putting `parseSourceAndType`, `ParseFullyQualifiedPath` and `validateTypeName` together under `buf decode` for now, and move them to `bufreflect` when we need to use them in `buf encode`, so that `parseSourceAndType` can use `validateTypeName`

private/buf/cmd/buf/command/decode/decode.go
cyinma: I added an extra line to mention the fully-qualified path, but we probably still want to make more improvements on that

private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// ParseSourceAndType returns the moduleReference and typeName from the source and type provided by the user.
+// When source is not provided, we assume the type is a fully-qualified path to the type and try to parse it.
+// Otherwise, if both source and type are provided, the type must be a valid Protobuf identifier (e.g. weather.v1.Units).
+func ParseSourceAndType(
+	ctx context.Context,
+	flagSource string,
+	flagType string,
+) (moduleReference string, typeName string, _ error) {
```
amckinney, line 63: Given that this function is particularly tailored to flag parameters, it probably belongs in the `bufcli` package. We might be able to justify that for `parseFullyQualifiedPath` too, but we could export `validateTypeName` from here so it can be used by both `bufreflect.NewMessage` and the `bufcli` functions.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <descriptor>",
```
amckinney, line 48: Removed the `<descriptor>` positional parameter since we're reading the message from `stdin`.

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,149 @@
...
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name,
+		Short: "Decode binary serialized message with a source reference.",
```
bufdev, line 49: @amckinney and I are thinking about the UI here and wondering if there's one command we can have that can take an arbitrary input of either json/bin format and produce an output in another specified format. So i.e. `buf encode --input - --output message.json`. We can't figure out a good UX for this yet though. Let's get this into `buf beta` and then iterate on this. Bug @amckinney and I about this

private/buf/cmd/buf/command/decode/decode.go
```diff
@@ -0,0 +1,149 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
```
bufdev, line 1: Move to `private/buf/cmd/buf/command/beta/decode/decode.go`

private/bufpkg/bufreflect/bufreflect.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protodesc"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/dynamicpb"
+)
+
+// NewMessage returns a new dynamic proto.Message for the fully qualified typeName
+// in the bufimage.Image.
+func NewMessage(
```
bufdev, line 30: @amckinney can you have you know who take a quick glance at this?

bufdev (commented):
@amckinney and I did a joint video review - looks really good, some comments

private/buf/cmd/buf/buf_test.go
```diff
@@ -1723,6 +1724,181 @@ Successfully migrated your buf.yaml and buf.gen.yaml to v1.`,
...
+		defer stdin.Close()
+		stdout := bytes.NewBuffer(nil)
+		testRun(
+			t,
+			0,
+			stdin,
+			stdout,
+			"beta",
+			"decode",
+			"--source",
```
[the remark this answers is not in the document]
cyinma: I am confused for this, this cmd currently only take stdin as input and doesn't accept arguments for now, is that what you mean by "first parameter" ?

private/buf/cmd/buf/command/decode/decode.go
cyinma: I am not sure about the use case for this as I think the user know what they want to do (encode/decode) in most of the time? but yes let's discuss this 🙌

private/buf/cmd/buf/buf_test.go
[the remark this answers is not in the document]
bufdev: Instead of using `--source SOURCE`, just have `SOURCE`

private/buf/bufcli/bufcli.go
```diff
@@ -725,6 +739,93 @@ func ReadModuleWithWorkspacesDisabled(
...
+}
+
+// ParseSourceAndType returns the moduleReference and typeName from the source and type provided by the user.
+// When source is not provided, we assume the type is a fully-qualified path to the type and try to parse it.
+// Otherwise, if both source and type are provided, the type must be a valid Protobuf identifier (e.g. weather.v1.Units).
+func ParseSourceAndType(
+	ctx context.Context,
+	flagSource string,
+	flagType string,
+) (moduleReference string, typeName string, _ error) {
```
amckinney, line 812: We might not want to use name arguments here because we're not always returning a `moduleReference`. The `flagSource` could be a local directory, and it could be returned as-is if both `--source` and `--type` are provided.

I'd just drop the return argument names. Also, I'd drop the `flag` prefix from `flagSource` and `flagType` and potentially relabel `type` to `typeName` to avoid the keyword conflict.

private/buf/cmd/buf/command/beta/decode/decode.go
```diff
@@ -0,0 +1,154 @@
...
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name,
+		Short: "Decode binary serialized message with a source reference.",
+		Long: `The first argument is the source that defines the serialized message (e.g. buf.build/acme/weather).
+If no argument is specified, the type provided need to be a fully-qualified path to the type (e.g. buf.build/acme/weather#acme.weather.v1.Units).`,
```
amckinney, line 50:
```suggestion
If no argument is specified, the type provided must be a fully-qualified path to the type (e.g. buf.build/acme/weather#acme.weather.v1.Units).`,
```

private/buf/cmd/buf/command/beta/decode/decode.go
```diff
@@ -0,0 +1,154 @@
...
+		fmt.Sprintf(
+			"The format for build errors, printed to stderr. Must be one of %s.",
+			stringutil.SliceToString(bufanalysis.AllFormatStrings),
+		),
+	)
+	flagSet.StringVar(
+		&f.Type,
+		typeFlagName,
+		"",
+		`The fully-qualified type name of the serialized message (e.g. acme.weather.v1.Units)
```
amckinney, line 86:
```suggestion
		`The full type name of the serialized message (e.g. acme.weather.v1.Units)
```

bufdev (approved):
This mostly looks fine to me, but @amckinney make sure you have no complaints. We can work on the UX since this is in `buf beta` post-commit.

amckinney (approved):
This looks great now that those remaining comments have been addressed - nice job on consolidating this logic with `buf build`.

We can decide whether to bake more of the `buf encode` functionality in the same command in a follow-up. If we do, we can rename `buf decode` to `buf encode` and continue to iterate from there.

TL;DR @bufdev I think this is in a great starting place for us to continue designing the rest.

PR #938, opened 2022-02-14
DISCUSS: Do not override existing token in the bufcli Context Modifier

### LINKED ISSUE #934: `buf registry login` token validation should override existing .netrc token
If there is an existing token in the `.netrc` or `BUF_TOKEN` env var, this is being used by the context modifier set for the container of the `buf` CLI instead of the token set for verification.

Currently, the workaround for this is to ensure there is no `BUF_TOKEN` env var and no `.netrc` (basically by running `buf registry logout` first) to ensure that the verification is run on the new token.

doriable:
Fixes #934

For `buf registry login`, we make a call out to `GetCurrentUser` with the token provided in order to check the validity of the token. However, the `bufcli` has a context modifier that will use an existing token in the `.netrc` or set to `BUF_TOKEN` in the request unless those are both unset. This change will prevent the context modifier from overriding a token that has already been set on the context.

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -547,7 +547,9 @@ func NewContextModifierProvider(
 		}
 		return func(ctx context.Context) context.Context {
 			ctx = bufrpc.WithOutgoingCLIVersionHeader(ctx, Version)
-			if token != "" {
+			// Only set the token if there is no existing token on the context.
+			// This allows requests to override the token at the call site as needed.
+			if token != "" && !rpcauth.ExistingToken(ctx) {
 				return rpcauth.WithToken(ctx, token)
 			}
 			return ctx
--- a/private/pkg/rpc/rpcauth/rpcauth.go
+++ b/private/pkg/rpc/rpcauth/rpcauth.go
@@ -60,6 +60,14 @@ func WithToken(ctx context.Context, token string) context.Context {
 	return ctx
 }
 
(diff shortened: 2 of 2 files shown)


private/pkg/rpc/rpcauth/rpcauth.go
```diff
@@ -60,6 +60,12 @@ func WithToken(ctx context.Context, token string) context.Context {
 	return ctx
 }
 
+// ExistingTokenOutgoingHeader returns true if a token is already set on the outgoing
+// context, otherwise it returns false.
+func ExistingTokenOutgoingHeader(ctx context.Context) bool {
```
bufdev, line 65: `TokenOutgoingHeaderExists`

private/buf/bufcli/bufcli.go
```diff
@@ -547,10 +547,7 @@ func NewContextModifierProvider(
 		}
 		return func(ctx context.Context) context.Context {
 			ctx = bufrpc.WithOutgoingCLIVersionHeader(ctx, Version)
-			if token != "" {
-				return rpcauth.WithToken(ctx, token)
-			}
-			return ctx
+			return rpcauth.WithTokenIfNoneSet(ctx, token)
```
doriable, line 550: I guess I could in-line the `bufrpc.WithOutgoingCLIVersionHeader` call here also, but I'm not sure that's necessary...

PR #940, opened 2022-02-14
Add buf push --vcs-check

WillAbides:
This adds a `--vcs-check` flag to `buf push` with two values "github" and "none". Of course "none" does nothing.

`--vcs-check=github` is meant to be used by `buf-push-action` to prevent out-of-order runs from overwriting the latest BSR commit with something older, however it can also be used outside of `buf-push-action`.

`--vcs-check=github` requires:
  - Exactly one `--tag` is a git commit. There can be more tag values set, but only one can be a 40 character hex string.
  - Environment variables `GITHUB_TOKEN` and `GITHUB_REPOSITORY` are set.
  - If using GitHub Enterprise Server `GITHUB_API_URL` must be set.

`--vcs-check=github` will block the push if all the following conditions are met for the head of any track being pushed:
  - It has a different digest from the module being pushed.
  - It is not already tagged with the same git commit tag specified by `--tag`.
  - It is not tagged with any git commit hashes that are behind the git commit tag specified by `--tag`.
  - It is tagged with one or more git commit hashes that are ahead of the git commit tag specified by `--tag`.

Another option is adding `--vcs-check=git` which would do the same checks using the local git repository. I opted to start with the github api because it is more straight-forward to run from a github action where you have to deal with shallow clones that will prevent the checks from working.

### DIFF
--- a/private/buf/cmd/buf/command/push/vcscheck.go
+++ b/private/buf/cmd/buf/command/push/vcscheck.go
@@ -1,0 +1,192 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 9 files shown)


private/buf/cmd/buf/command/push/vcscheck.go
```diff
@@ -0,0 +1,192 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
```
WillAbides, line 1: This probably belongs in another package. I just wasn't quite sure where.

PR #944, opened 2022-02-15
Disable config override for workspaces

doriable:
The behaviour of the `--config` flag for workspaces is not immediately apparent to users. If a config contains [build configurations](https://docs.buf.build/configuration/v1/buf-yaml#build), for example, then that does not apply seamlessly to all modules across the workspace, and the build will error. Since this behaviour is very brittle, we have made the decision to disable the `--config` flag altogether for workspaces.

### DIFF
--- a/private/buf/bufwork/workspace_builder.go
+++ b/private/buf/bufwork/workspace_builder.go
@@ -57,7 +57,7 @@ func (w *workspaceBuilder) BuildWorkspace(
 		return nil, errors.New("received a nil workspace config")
 	}
 	if configOverride != "" {
-		return nil, errors.New("the --config flag has is not compatible with workspaces")
+		return nil, errors.New("the --config flag is not compatible with workspaces")
 	}
 	// We know that if the file is actually buf.work for legacy reasons, this will be wrong,
 	// but we accept that as this shouldn't happen often anymore and this is just
--- a/private/buf/bufwork/workspace_builder.go
+++ b/private/buf/bufwork/workspace_builder.go
@@ -57,7 +57,7 @@ func (w *workspaceBuilder) BuildWorkspace(
 		return nil, errors.New("received a nil workspace config")
 	}
 	if configOverride != "" {
-		return nil, errors.New("the --config flag has been disabled for workspaces")
+		return nil, errors.New("the --config flag has is not compatible with workspaces")
 	}
 	// We know that if the file is actually buf.work for legacy reasons, this will be wrong,
 	// but we accept that as this shouldn't happen often anymore and this is just
--- a/private/buf/bufwork/workspace_builder.go
+++ b/private/buf/bufwork/workspace_builder.go
@@ -56,6 +56,9 @@ func (w *workspaceBuilder) BuildWorkspace(
 	if workspaceConfig == nil {
 		return nil, errors.New("received a nil workspace config")
 	}
+	if configOverride != "" {
+		return nil, errors.New("the --config flag has been disabled for workspaces")
+	}
 	// We know that if the file is actually buf.work for legacy reasons, this will be wrong,
 	// but we accept that as this shouldn't happen often anymore and this is just
 	// used for error messages.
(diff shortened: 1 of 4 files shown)


private/buf/bufwork/workspace_builder.go
```diff
@@ -56,6 +56,9 @@ func (w *workspaceBuilder) BuildWorkspace(
 	if workspaceConfig == nil {
 		return nil, errors.New("received a nil workspace config")
 	}
+	if configOverride != "" {
+		return nil, errors.New("the --config flag has been disabled for workspaces")
```
amckinney, line 60: I'd update the message a bit here - we shouldn't refer to it as something that was previously enabled.
```suggestion
		return nil, errors.New("the --config flag has is not compatible with workspaces")
```

amckinney (approved):
We might not need to keep around _all_ the test cases from before, but it doesn't hurt to leave them around either. Looks good to me.

private/buf/bufwork/workspace_builder.go
doriable: Ah yeah, that is a lot cleaner, I am a fan of this :) (Committing this suggestion, I know it'll break tests, lol, I'll fix that, haha)

doriable:
> We might not need to keep around _all_ the test cases from before, but it doesn't hurt to leave them around either. Looks good to me.

Yeah, I wasn't sure about that, since on one hand, they all end up testing the same code path (lol), but on the other hand, they're tests I don't want to miss if we bring back compatibility with the flag. But thinking through that a bit, I think it might still make sense to cull down on the tests (as you say, maybe not _entirely_), since if we were to re-implement this, it would be rewritten anyway, and we'd need to go through and carefully reassess the test cases.

amckinney:
Yea it's no big deal, feel free to leave them in if it's easier. As long as different arguments are being used (e.g. targeting a workspace, targeting a module within a workspace, etc), we're still testing that the `configOverride` takes effect for different workspace scenarios.

andrewparmet:
How can we specify a config for a workspace if we use a symlinks for our directories? I don't have control of the modules; I reference external protobuf and want to specify my own config.

andrewparmet:
I got around this by using an image created from the workspace instead of comparing against the workspace. 

However I was required to provide a `buf.yaml` config file; Buf rejected the config option even though I was only using the command from a directory that included a `buf.work.yaml` and not actually using it. Not sure if that’s desired behavior.

amckinney:
> How can we specify a config for a workspace if we use symlinks for our directories? I don't have control of the modules, so I cannot write buf.yaml files in them; I reference external protobuf and want to specify my own config.

Could you go into a little more detail? Do you need different `buf.yaml` configurations in each of your symlink'ed workspace directories, or did you just want to use a single `buf.yaml` uniformly across each of the directories specified in your `buf.work.yaml`? The more detail you can give to your use case, the better.

> However I was required to provide a buf.yaml config file; Buf rejected the config option even though I was only using the command from a directory that included a buf.work.yaml and not actually using it. Not sure if that’s desired behavior.

This is a bug - I was able to reproduce this behavior on my end, too. I'll open another issue to track this and we'll get a fix out soon.

amckinney:
If you want to ignore linting the downloaded proto altogether, you should just just target the sub-directory you care to lint. In your case, it sounds like the symlink'ed directory specified in the `buf.work.yaml`.

So if you have the following:

```yaml
# buf.work.yaml
version: v1
directories:
  - downloaded-directory
  - user-directory-symlink
```

You could just run the following:

```sh
$ buf lint path/to/user-directory-symlink
```

If the user has a `buf.yaml` defined in that symlink'ed directory, the configuration defined there will be used and you don't need to use the `--config` flag at all.

Let me know if that helps - I might be missing something.

andrewparmet:
That's handy, thanks. I didn't know you could target directories. I wrote this over a year ago - maybe it wasn't a feature yet! 

In general the user may specify an arbitrary config file, so I may have to use the config option anyways.

For now my solution of linting from the image is probably good enough, but next time I revisit this implementation I'll see about linting only the user directory protos. That'll probably be more correct in case a non-ignored package makes its way into the downloaded schema - right now they're forced to ignore the whole package that has violating definitions across all modules if any one module has a violation, so they could accidentally ignore a lint violation in their own code if they have more definitions in the same package.

PR #945, opened 2022-02-15
Update golangci-lint to v1.44.0

amckinney:
This updates `golangic-lint` to `v1.44.0` and addresses some of the issues with the `forcetypeassert` plugin that were previously uncaught (they were caught in the upgrade).

There have been several new linters added in this release:
- [containdctx](https://github.com/sivchari/containedctx)
- [decorder](https://gitlab.com/bosi/decorder)
- [errchkjson](https://github.com/breml/errchkjson)
- [grouper](https://github.com/leonklingele/grouper)
- [maintidx](https://github.com/yagipy/maintidx)

I don't think we want to add any of them - the `errchkjson` one is interesting, but we actually encourage that we check all of our errors so it's not really useful in our case.

### DIFF
--- a/make/go/dep_golangci_lint.mk
+++ b/make/go/dep_golangci_lint.mk
@@ -7,7 +7,7 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/golangci/golangci-lint/releases 20211103 checked 20211112
+# https://github.com/golangci/golangci-lint/releases 20220125 checked 20220215
 # Check for new linters and add to .golangci.yml (even if commented out) when upgrading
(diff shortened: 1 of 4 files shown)


bufdev:
I'd disagree: `decorder` in particular seems to enforce our specific style guidelines. `containdctx` looks nice, I actually don't like `errchkjson` for the reason you described, and `grouper` looks nice too. The only thing I wouldn't want is `maintidx` probably.

make/go/dep_golangci_lint.mk
```diff
@@ -9,7 +9,7 @@ $(call _assert_var,CACHE_BIN)
 # Settable
 # https://github.com/golangci/golangci-lint/releases 20211103 checked 20211112
```
bufdev, line 10: Update this

PR #946, opened 2022-02-15
Add support of JSON input and binary output in buf beta decode

cyinma:
This add support of JSON input and binary output in buf beta decode (i.e. run time encoding)

The default input is in binary format and the default output is in JSON format. To override it, use a file with `.json` or `.bin` extension as input / output, or specify the format in option:

use a file with `.json` or `.bin` extension as input / output
```
buf beta convert buf.build/acme/petapis --type pet.v1.Pet --input pet.plain.json --output result.bin
```

specify the format in option for file
```
buf beta convert buf.build/acme/petapis --type pet.v1.Pet --input pet.plain.txt#format=json --output result.txt#format=bin
```

specify the format in option for stdin (this will make the default output format to be bin)
```
cat pet.plain.json | buf beta convert buf.build/acme/petapis --type pet.v1.Pet --input -#format=json
```

a round trip from binary to JSON back to binary
```
cat pet.plain.bin | buf beta convert --type buf.build/acme/petapis#pet.v1.Pet | buf beta convert --type buf.build/acme/petapis#pet.v1.Pet --input -#format=json
```


private/buf/buffetch/buffetch.go
```diff
@@ -34,6 +34,10 @@ const (
 	ImageEncodingBin ImageEncoding = iota + 1
 	// ImageEncodingJSON is the JSON image encoding.
 	ImageEncodingJSON
+	// MessageEncodingBin is the binary image encoding.
+	MessageEncodingBin MessageEncoding = iota + 1
```
bufdev, line 38: I'm worried about this - `buffetch` is meant to be specific to things that can represent modules, namely images, sources, and modules themselves. It seems that `buffetch` is being re-used to do some work with message encoding/decoding. `buffetch/internal` was originally designed to be a library that was independent of what you were trying to fetch, buf `buffetch` is not.

cyinma: do you think it will be better to put these logic to `bufreflect`? we might also need a ref parser to handle the output format, which seems should also live in `bufreflect`

amckinney: I don't think `bufreflect` is the right home either - it's meant to be a lightweight package that interacts with the `protoreflect` APIs. Ideally we just keep it scoped to things like `bufreflect.NewMessage`.

But I agree with @bufdev - this doesn't really make sense in `buffetch` because it should remain specific to modules. We just want similar logic handling the inputs we support for encoding/decoding (e.g. binary protobuf, JSON, etc).

Given that we only have one CLI command, we could start by putting all of this logic in the same package as the command implementation for now. If we ever need it elsewhere, we could create a new home for it and go from there. Let's keep it scoped to just supporting `.bin` and `.json` for now. This wouldn't work with the `bufwire` abstractions in the same way (since it won't be a `buffetch.Ref`), but we might not need all of that to get something tailored to what we want in these commands.

TL;DR Try simplifying some of this logic in the command implementation and see if we can come up with something simpler that doesn't affect `buffetch`.

amckinney:
Haven't reviewed the code yet, but there's only one other thing I'd add to the UX (discussed with @bufdev) - 

Let's determine the default output based on the input. So in the case that the user provides JSON input, we should default to write binary. You already have the case covered when a user provides binary input (we default to writing JSON).

Finally, and I know we've gone back and forth on this a ton, but let's name the command `buf beta convert` so that it's clear we're converting between formats. "Decoding" JSON to binary just doesn't feel right.

TL:DR:
1. Change up the default input/output a bit.
2. Move this to `buf beta convert` and we can keep bike-shedding on what to name this later. Starting in `beta` means we can make incompatible changes later.

private/buf/bufconvert/bufconvert.go
```diff
@@ -0,0 +1,151 @@
...
+		return MessageEncodingBin, nil
+	case formatJSON:
+		return MessageEncodingJSON, nil
+	default:
+		return 0, fmt.Errorf("invalid format for message: %q", format)
+	}
+}
+
+// rawPath will be non-empty
+func getRawPathAndOptions(value string) (string, map[string]string, error) {
```
amckinney, line 114: It'd be great if we could share this logic between `buffetch` and this - copying complex code like this can cause problems.

In both cases we're working with some sort of reference - maybe we could move this to `bufref.GetRawPathAndOptions` and have both `bufconvert` and `buffetch` import from it? I'm not sure what other package names to use here (cc @bufdev).

private/buf/bufconvert/bufconvert.go
```diff
@@ -0,0 +1,151 @@
...
+
+func getPathAndMessageEncoding(
+	ctx context.Context,
+	value string,
+	defaultEncoding MessageEncoding,
+) (string, MessageEncoding, error) {
+	path, options, err := getRawPathAndOptions(value)
+	if err != nil {
+		return "", 0, err
+	}
```
amckinney, line 80: In `buffetch` we validate that all of the options are valid. In this case we're only checking for the existence of `format` and acting upon it if it exists. I'd recommend using ` switch` like `getRawRef` in `buffetch` does and return an error if an option (anything but `format`) is specified.

private/buf/bufwire/proto_encoding_reader.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	var unmarshaler protoencoding.Unmarshaler
+	switch messageRef.MessageEncoding() {
+	case bufconvert.MessageEncodingBin:
+		unmarshaler = protoencoding.NewWireUnmarshaler(resolver)
+	case bufconvert.MessageEncodingJSON:
+		unmarshaler = protoencoding.NewJSONUnmarshaler(resolver)
+	default:
+		return nil, fmt.Errorf("unknown message encoding type")
+	}
```
amckinney, line 97: If we can return an error from the `messageRef.MessageEncoding()`, we should validate it as soon as possible.  Otherwise we'll read the input, creating a resolver, and a dynamic message before we inevitably fail.

In short, let's move this block closer to the top of the function to prevent ourselves from doing any unnecessary work.

private/buf/bufwire/proto_encoding_writer.go
```diff
@@ -60,14 +59,26 @@ func (i *protoEncodingWriter) PutMessage(
...
-	if err != nil {
-		return err
+	var marshaler protoencoding.Marshaler
+	switch messageRef.MessageEncoding() {
+	case bufconvert.MessageEncodingBin:
+		marshaler = protoencoding.NewWireMarshaler()
+	case bufconvert.MessageEncodingJSON:
+		marshaler = protoencoding.NewJSONMarshalerIndent(resolver)
+	default:
+		return fmt.Errorf("unknown message encoding type")
```
amckinney, line 69: Similar to above w.r.t. earlier validation, but this one is less of a big deal because there's only one other thing we're doing before this potential error.

private/buf/cmd/buf/buf_test.go
```diff
@@ -1873,13 +1873,126 @@ func TestDecodeInvalidTypeName(t *testing.T) {
...
 		"beta",
-		"decode",
+		"convert",
 		filepath.Join(tempDir, "image.bin"),
 		"--type",
 		".foo",
 	)
 }
 
+func TestConvertEncodeDecodeRoundTrip(t *testing.T) {
```
amckinney, line 1883:
```suggestion
func TestConvertRoundTrip(t *testing.T) {
```

private/buf/cmd/buf/buf_test.go
```diff
@@ -1851,7 +1851,7 @@ func TestDecodeOutput(t *testing.T) {
 	})
 }
 
-func TestDecodeInvalidTypeName(t *testing.T) {
+func TestConvertDecodeInvalidTypeName(t *testing.T) {
```
amckinney, line 1854:
```suggestion
func TestConvertInvalidTypeName(t *testing.T) {
```

private/buf/cmd/buf/buf_test.go
```diff
@@ -1749,17 +1749,17 @@ func TestDecodeWithImage(t *testing.T) {
...
+			"convert",
 			filepath.Join(tempDir, "image.bin"),
 			"--type",
 			"buf.Foo",
 		)
 	})
 }
 
-func TestDecodeOutput(t *testing.T) {
+func TestConvertDecodeOutput(t *testing.T) {
```
amckinney, line 1762:
```suggestion
func TestConvertOutput(t *testing.T) {
```

private/buf/cmd/buf/buf_test.go
```diff
@@ -1711,7 +1711,7 @@ Successfully migrated your buf.yaml and buf.gen.yaml to v1.`,
 	})
 }
 
-func TestDecodeWithImage(t *testing.T) {
+func TestConvertDecodeWithImage(t *testing.T) {
```
amckinney, line 1714:
```suggestion
func TestConvertWithImage(t *testing.T) {
```

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -45,7 +44,7 @@ func NewCommand(
 	flags := newFlags()
 	return &appcmd.Command{
 		Use:   name + " <source>",
-		Short: "Use a source reference to decode a binary serialized message supplied through stdin.",
+		Short: "Use a source reference to encode or decode a binary serialized message supplied through stdin.",
```
amckinney, line 47: This one is kinda hard to write. We should make it clear that the input and output can both be binary or JSON though.

```suggestion
		Short: "Use a source reference to convert a binary or JSON serialized message supplied through stdin.",
```

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -105,50 +115,71 @@ func run(
...
+func counterEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEncoding, error) {
+	switch encoding {
+	case bufconvert.MessageEncodingBin:
+		return bufconvert.MessageEncodingJSON, nil
+	case bufconvert.MessageEncodingJSON:
+		return bufconvert.MessageEncodingBin, nil
+	default:
+		return 0, fmt.Errorf("unknown message encoding %v", encoding)
+	}
+}
```
amckinney, line 185: Let's add a comment to this. I see what you're going for with the word "counter", but it might not be immediately obvious to future readers.

amckinney (commented):
This is looking great, nice work! Mostly structural things at this point - the UX is looking solid. Check out my comments, then we'll have @bufdev and I do a final review.

private/buf/bufwire/proto_encoding_writer.go
cyinma: for this one, the resolver is actually need to be existed before this switch block, as we need it to create the marshaler, so the only way we can return the error earlier is to do an additional checking in the beginning to check the value of `messageRef.MessageEncoding()`, which doesn't seems to be worth to do as this should only happen in an implementation error, how do you think?

private/buf/cmd/buf/command/beta/convert/convert.go
cyinma: added a comment and also renamed it to `inverseEncoding` to see will it be more obvious

private/buf/bufconvert/bufconvert.go
```diff
@@ -0,0 +1,121 @@
...
+
+// MessageEncoding is the encoding of the message
+type MessageEncoding int
+
+type MessageEncodingRef interface {
+	Path() string
+	MessageEncoding() MessageEncoding
+}
+
+func NewMessageEncodingRef(
```
amckinney, line 60: All exported functions and identifiers need a comment.

private/buf/bufref/bufref.go
```diff
@@ -0,0 +1,61 @@
...
+
+package bufref
+
+import (
+	"strings"
+)
+
+// GetRawPathAndOptions returns the raw path and options from the value provided,
+// the rawPath will be non-empty when returning without error here.
+func GetRawPathAndOptions(value string) (string, map[string]string, error) {
```
amckinney, line 23: Flagging @bufdev to check this one out. This stemmed from [this](https://github.com/bufbuild/buf/pull/946#discussion_r810413228) comment. Let me know if you have any other ideas.

private/buf/bufref/errors.go
```diff
@@ -0,0 +1,44 @@
...
+	return fmt.Errorf("%q ends with # which is invalid", value)
+}
+
+func newOptionsInvalidError(s string) error {
+	return fmt.Errorf("invalid options: %q", s)
+}
+
+func newOptionsDuplicateKeyError(key string) error {
+	return fmt.Errorf("duplicate options key: %q", key)
+}
```
amckinney, line 44: Great to see that we can un-export all these now.

private/buf/bufwire/proto_encoding_writer.go
```diff
@@ -60,14 +59,26 @@ func (i *protoEncodingWriter) PutMessage(
...
 	}
-	writeCloser, err := i.fetchWriter.PutSingleFile(ctx, container, path)
+	data, err := marshaler.Marshal(message)
 	if err != nil {
 		return err
 	}
+	writeCloser := ioextended.NopWriteCloser(container.Stdout())
+	if messageRef.Path() != "-" {
+		writeCloser, err = os.Create(messageRef.Path())
```
amckinney, line 77: This caught my eye at first, but it looks like we do the same thing in our internal `buffetch` writer.

amckinney (approved):
Nicely done, a few things to consider, but otherwise this looks good to ship in `beta`. We can continue to iterate from there.

private/buf/bufconvert/bufconvert.go
bufdev: Idk I'd just choose a package name, doesn't need to make too much sense

private/buf/bufref/bufref.go
bufdev: sgtm

private/buf/bufwire/proto_encoding_writer.go
cyinma: yes the underlying logic in `singleRef` is something similar to this

PR #949, opened 2022-02-16
Filtering images by type

robbertvanginkel:
This adds a package and functionality to filter an image by a specific set of types (messages, enums, services).

To use this from the cli, there's an added `--type` flag (that can be specified repeatedly):
```
$ buf build -o -#format=json --type buf.alpha.registry.v1alpha1.DownloadService | jq '.file[] | {package:.package, messages: [.messageType[] | .name]}'
{
  "package": "buf.alpha.lint.v1",
  "messages": [
    "Config",
    "IDPaths"
  ]
}
{
  "package": "buf.alpha.registry.v1alpha1",
  "messages": [
    "DownloadRequest",
    "DownloadResponse"
  ]
}
...
 ```

### CODE REVIEW

File: private/buf/cmd/buf/command/build/build.go [hunk 1]
```diff
@@ -114,6 +116,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		`The file or data to use to use for configuration.`,
 	)
+	flagSet.StringSliceVar(&f.Types, "type", nil, "create filtered images")
```
robbertvanginkel, line 119:
```suggestion
	flagSet.StringSliceVar(&f.Types, "type", nil, "The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include types required to describe the requested types.")
```

private/buf/cmd/buf/command/build/build.go
```diff
@@ -150,6 +153,12 @@ func run(
 	if err != nil {
 		return fmt.Errorf("--%s: %v", outputFlagName, err)
 	}
+	if len(flags.Types) > 0 {
```
robbertvanginkel, line 156: I don't know if this is the right place to add this or if it needs to go somewhere else (like bufcli.NewImageForSource). Added here for now so its usable from the cli.

private/buf/cmd/buf/command/build/build.go
bufdev: Seems right

private/buf/cmd/buf/command/build/build.go
bufdev: I think optimally this would be in `bufcli.NewImageForSource`, but there's a wider UI discussion about whether we want this flag somehow exposed in other commands, for example `buf breaking`. If we do, then we probably want to make this logic common. If we don't, then here is fine.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -60,6 +61,470 @@ func FreeMessageRangeStrings(
 	return s, nil
 }
 
+// ImageFilteredByTypes returns a minimal image containing only the descriptors
+// required to interpret types.
+//
+// Although this returns a new bufimage.Image, it mutates the original image's
+// underlying file's `descriptorpb.FileDescriptorProto` and the old image should
+// not continue to be used.
+func ImageFilteredByTypes(image bufimage.Image, types []string) (bufimage.Image, error) {
```
bufdev, line 70: Nit: would probably make this variadic, ie `types ...string`

private/bufpkg/bufimage/bufimageutil/image_index.go
```diff
@@ -0,0 +1,107 @@
...
+package bufimageutil
+
+import (
+	"fmt"
+
+	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"github.com/bufbuild/buf/private/pkg/protosource"
+)
+
+type imageIndex struct {
```
bufdev, line 24: docs would be good for all of this

bufdev (commented):
I would love a doc on the exact mechanics of the filter, i.e. what happens with dependencies, extensions, options, etc, in different situations. The best case is this is Godoc on the function for now, but we'll probably need to add this to docs.buf.build as part of the build documentation. Overall, this is mostly going to come down to QA for me - seems reasonably laid out.

bufdev:
Also I'm mostly approved on this, I think regardless of if we add this to other commands, this PR makes sense in isolation, but would love to see the docs. Thanks for doing this.

private/bufpkg/bufimage/bufimageutil/image_index.go
```diff
@@ -0,0 +1,107 @@
...
+	NameToDescriptor map[string]protosource.NamedDescriptor
+	NameToExtensions map[string][]protosource.Field
+	NameToOptions    map[string]map[int32]protosource.Field
+}
+
+func newImageIndexForImage(image bufimage.Image) (*imageIndex, error) {
```
amckinney, line 30: There's a lot of repetition in this function. We could build a couple functions for adding a type and adding a field to DRY this up and make it a little more readable.

Something along the lines of:

```go
func (i *imageIndex) AddType(namedDescriptor protosource.NamedDescriptor) error {
	if storedDescriptor, ok := i.NameToDescriptor[namedDescriptor.FullName()]; ok && storedDescriptor != namedDescriptor {
		return fmt.Errorf("duplicate for %q: %#v != %#v", namedDescriptor.FullName(), storedDescriptor, namedDescriptor)
	}
	i.NameToDescriptor[namedDescriptor.FullName()] = namedDescriptor
	return nil
}

func (i *imageIndex) AddField(field protosource.Field) error {
	i.NameToDescriptor[field.FullName()] = field
	extendeeName := field.Extendee()
	if isOptionsTypeName(extendeeName) {
		if _, ok := i.NameToOptions[extendeeName]; !ok {
			i.NameToOptions[extendeeName] = make(map[int32]protosource.Field)
		}
		i.NameToOptions[extendeeName][int32(field.Number())] = field
	} else {
		i.NameToExtensions[extendeeName] = append(i.NameToExtensions[extendeeName], field)
	}
	return nil
}

func newImageIndexForImage(image bufimage.Image) (*imageIndex, error) {
	index := &imageIndex{
		NameToDescriptor: make(map[string]protosource.NamedDescriptor),
		NameToExtensions: make(map[string][]protosource.Field),
		NameToOptions:    make(map[string]map[int32]protosource.Field),
	}
	for _, file := range image.Files() {
		protosourceFile, err := protosource.NewFile(newInputFile(file))
		if err != nil {
			return nil, err
		}
		if err := protosource.ForEachMessage(func(message protosource.Message) error {
			index.AddType(message)
			for _, field := range message.Extensions() {
				index.AddField(field)
			}
			return nil
		}, protosourceFile); err != nil {
			return nil, err
		}
		if err = protosource.ForEachEnum(func(enum protosource.Enum) error {
			index.AddType(enum)
			return nil
		}, protosourceFile); err != nil {
			return nil, err
		}
		for _, service := range protosourceFile.Services() {
			index.AddType(service)
		}
		for _, field := range protosourceFile.Extensions() {
			index.AddField(field)
		}
	}
	return index, nil
}
```

There might be problems with the code I included above - it's mostly meant to demonstrate an alternative approach. Take it or leave it.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -60,6 +61,470 @@ func FreeMessageRangeStrings(
...
+	// Check types exist
+	startingDescriptors := make([]protosource.NamedDescriptor, 0, len(types))
+	for _, typeName := range types {
+		descriptor, ok := imageIndex.NameToDescriptor[typeName]
+		if !ok {
+			return nil, fmt.Errorf("not found: %q", typeName)
+		}
+		if image.GetFile(descriptor.File().Path()).IsImport() {
+			return nil, fmt.Errorf("type %q is in an import", typeName)
+		}
```
amckinney, line 84: This error makes sense for `buf build`, but it might be confusing for other commands if we ever support the`--type` flag elsewhere.

For example, imagine getting this error after invoking the following:
```sh
$ buf generate --type google.actions.type.DateTimeRange --include-imports
type "google.actions.type.DateTimeRange" is an import
```

With that said, `buf build` includes imports by default, but the default behavior wouldn't make sense either. For example, it'd be weird for this to work:

```sh
$ buf build buf.build/acme/petapis --type google.actions.type.DateTimeRange 
```

Nothing actionable here, but it might be worth flagging this as something we might need to revisit if we ever support `--type` in the commands that act upon imports. It's easier to start with an error anyway.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -60,6 +61,470 @@ func FreeMessageRangeStrings(
...
+			// with public import only inserted in the middle).
+			// We should check if all keys in importmap get looked up.
+			if _, ok := importsRequired[importPath]; ok {
+				indexFromTo[int32(indexFrom)] = int32(indexTo)
+				imageFileDescriptor.Dependency[indexTo] = importPath
+				indexTo++
+			}
+		}
+		imageFileDescriptor.Dependency = imageFileDescriptor.Dependency[:indexTo]
+		i := 0
```
amckinney, line 145: Use `var i int` for default value initialization whenever possible. Applies elsewhere.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -60,6 +61,470 @@ func FreeMessageRangeStrings(
...
+// descriptorAndDirects holds a protsource.NamedDescriptor and a list of all
+// named descriptors it directly references. A directly referenced dependency is
+// any type that if defined in a different file from the principal descriptor,
+// an import statement would be required for the proto to compile.
+type descriptorAndDirects struct {
+	Descriptor protosource.NamedDescriptor
+	Directs    []protosource.NamedDescriptor
+}
+
+func descriptorTransitiveClosure(namedDescriptor protosource.NamedDescriptor, imageIndex *imageIndex, seen map[string]struct{}) ([]descriptorAndDirects, error) {
```
amckinney, line 261: Overall the idea is sane. I trust that this covers all the bases and is sufficiently DRY'd up.

amckinney (commented):
Overall looks good. Just minor details.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
robbertvanginkel: Yes perhaps this should be configurable, especially if we want to extend beyond `buf build`. When building I had the `GetImage` api case in mind where we decided we wouldn't want it. Lets keep it like this and make it configurable when we reason through if it makes sense for other CLI commands.

private/buf/cmd/buf/command/build/build.go
robbertvanginkel: I can see how this might be useful for other commands, generation seems like the most natural. Perhaps we should make the flag hidden till this is thought through a little bit more - my main need is in the GetImage api for now and this was primarily nice last minute addition so we could more easily play with it. My vote would be to keep it here (perhaps hidden) and follow up with adding it to all places in the CLI where it makes sense with proper changelog in a follow up.

robbertvanginkel:
Added a round of comments on the types and public godoc, hope that clarifies things a bit.

private/bufpkg/bufimage/bufimageutil/image_index.go
amckinney: Fair points, I agree that a "builder" would be the right way to go about this. But it's fine as-is - you don't need to roll one out if you don't see a need.

private/pkg/protosource/option_extension_descriptor.go
```diff
@@ -35,3 +36,23 @@ func (o *optionExtensionDescriptor) OptionExtension(extensionType protoreflect.E
...
+	extensionRanges := o.message.ProtoReflect().Descriptor().ExtensionRanges()
+	for b := o.message.ProtoReflect().GetUnknown(); len(b) > 0; {
+		fieldNo, _, n := protowire.ConsumeField(b)
+		if extensionRanges.Has(fieldNo) {
+			fieldNumbers = append(fieldNumbers, int32(fieldNo))
+		}
+		b = b[n:]
+	}
+	// Need to think if this is actually necessary, or if extensions will
+	// always be unknown...
```
amckinney, line 51: Is there anything that still needs to be investigated here?

amckinney (approved):
Looks good from an implementation perspective! I haven't played around with the CLI yet, but the UX looks solid.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -60,6 +61,534 @@ func FreeMessageRangeStrings(
...
+		indexFromTo := make(map[int32]int32)
+		indexTo := 0
+		for indexFrom, importPath := range imageFileDescriptor.GetDependency() {
+			// TODO: this only filters the existing imports down to
+			// the ones requested, if there was a type we picked up
+			// through a public import in a dependent file may
+			// filter out that file here, a type not to be found. We
+			// may need to add the file directly (or have a file
+			// with public import only inserted in the middle).
+			// We should check if all keys in importmap get looked up.
```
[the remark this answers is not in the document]
robbertvanginkel: I'll add a testcase to demonstrate. As for follow up, Given the following protos
```
"a.proto": []byte(`syntax = "proto3";package a;message Foo{}`),
"b.proto": []byte(`syntax = "proto3";package b;import public "a.proto";message Bar {}`),
"c.proto": []byte(`syntax = "proto3";package c;import "b.proto";message Baz{ a.Foo foo = 1; }`),
```
I think there's two ways to resolve this:
1. generate a b.proto with only a public import for a.proto and no other types
2. import a.proto directly in c.proto

I'm not sure what's more reasonable, so given that we lint against using `import public` I'm leaning towards documenting/testing the deficiency and leaving a decision till this comes up for someone wanting to use it.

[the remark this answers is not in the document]
amckinney: Got it, thanks for writing this up. I agree with the approach on "let's solve this when it's a problem". For what it's worth, (1) sounds like the correct answer, but I could be wrong.

bufdev (approved):

bufdev:
Need to update `CHANGELOG.md`

PR #955, opened 2022-02-17
golangci-lint: remove duplicated linter

akshayjshah:
We have `govet` listed twice. This commit pipes the list through `sort
-u`, so we should have a nice clean list now.


amckinney (approved):

amckinney:
Let's do the same for our other repositories if need be.

PR #957, opened 2022-02-18
Bufstyle ignores

bufdev:
Mostly notes for myself: We should reconcile the style of `buf, bandeps, bufstyle` optimally, and realistically `bandeps, bufstyle`. I went back and forth with having a `bufstyleapp` package. This is closer to `bandeps` now, but there's some nuance between application config and run-specific config (`buf` lint/breaking config and `bandeps` config is run-specific config, whereas config here is created as application config). Also us using `multichecker.Main` results in our `app` package flow not being usable. But preferably there would be more consistency for people to pattern off of. Also the option for `NewAnalyzerProvider` is a little wonky.

This actually does work, there's a POC in here.


private/bufpkg/bufstyle/analyzers.go
```diff
@@ -0,0 +1,81 @@
...
+					for expr, typeAndValue := range pass.TypesInfo.Types {
+						if t := typeAndValue.Type; t != nil {
+							if t.String() == "sync.Pool" {
+								pass.Reportf(expr.Pos(), "sync.Pool cannot be used")
+							}
+						}
+					}
+				}
+				return nil, nil
+			},
```
amckinney, line 77: I imagine some of these `Run` functions will get long and we'll want a separate named function for them. At that point, would we want every `Run` function to be named, or are we OK with a mixed bag of anonymous functions alongside named ones?

private/bufpkg/bufstyle/bufstyle.go
```diff
@@ -15,73 +15,39 @@
...
+//
+// relFilePath should be relative to rootDirPath.
+func WithIgnore(relFilePath string, analyzerName string) AnalyzerProviderOption {
+	return func(analyzerProvider *analyzerProvider) {
+		analyzerNames := analyzerProvider.ignoreRelFilePathToAnalyzerNames[relFilePath]
+		if analyzerNames == nil {
+			analyzerNames = make(map[string]struct{})
+			analyzerProvider.ignoreRelFilePathToAnalyzerNames[relFilePath] = analyzerNames
 		}
+		analyzerNames[analyzerName] = struct{}{}
```
amckinney, line 51: For idiomatic Go, I'd refactor the `nil` check to use the following:

```go
analyzerNames, ok := ...
if !ok {
  ...
}
```

We get the same functionality though, so 🤷

.bufstyle.yaml
```diff
@@ -0,0 +1,4 @@
+ignore:
+  - path: private/bufpkg/bufstyle/cmd/bufstyle/main.go
+    analyzers:
+      - NO_SYNC_POOL
```
amckinney, line 4: I think we talked about this before, but what are your thoughts on inverting the control here? So we'd ignore by `analyzer` name, and include one or more paths to ignore.

```
ignore:
  - analyzer: NO_SYNC_POOL
     paths:
       - private/bufpkg/bufstyle/cmd/bufstyle/main.go
```

You could even reduce it down further to look like the `ignore_only` `lint` configuration:

```
ignore:
  NO_SYNC_POOL:
     - private/bufpkg/bufstyle/cmd/bufstyle/main.go
```

And if we ever wanted to make this more widely available, we could reserve `ignore` for excluding certain linters (like `lint.ignore`), and rename the configuration above to `ignore_only`, but we shouldn't optimize for that yet.

amckinney (approved):
This is awesome! Handling all of the absolute/relative paths looks great.

Just a couple things to consider but otherwise ready to go imo.

private/bufpkg/bufstyle/analyzers.go
bufdev: Let's figure that out as these grow, happy to move them out to separate functions (or packages) later

private/bufpkg/bufstyle/bufstyle.go
bufdev: whoops, saturday morning brain fart

.bufstyle.yaml
bufdev: I updated to this style for now:

```
ignore:
  NO_SYNC_POOL:
     - private/bufpkg/bufstyle/cmd/bufstyle/main.go
 ```

PR #963, opened 2022-02-22
add convert service

cyinma:
This is to add a new convert service with convert endpoint, which serve for runtime encoding and decoding in backend


amckinney (commented):
Looks great - good call on renaming this to the `ConvertService`. Just a couple smaller details.

proto/buf/alpha/registry/v1alpha1/convert.proto
```diff
@@ -0,0 +1,57 @@
...
+  rpc Convert(ConvertRequest) returns (ConvertResponse);
+}
+
+message ConvertRequest {
+  // type_name is the full type name of the serialized message (like acme.weather.v1.Units).
+  string type_name = 1;
+  // image is the image source that defines the serialized message.
+  //
+  // Cannot be provided if module_info is provided.
+  buf.alpha.image.v1.Image image = 2;
```
bufdev, line 41: It seems to me that `image` and `module_info` are either/or - do we really need to support both? It seems that we'd only support one, likely `image`, and leave building the `image` as an exercise for the caller. If we really need both, then we need to use `oneof`, but I don't like it

proto/buf/alpha/registry/v1alpha1/convert.proto
```diff
@@ -0,0 +1,57 @@
...
+  // module_info is the module source that defines the serialized message.
+  //
+  // Cannot be provided if image is provided.
+  buf.alpha.convert.v1alpha1.ModuleInfo module_info = 3;
+  // message_bytes is the serialized Protobuf message.
+  bytes message_bytes = 4;
```
bufdev, line 47: `payload` perhaps, I don't love `message_bytes` - we don't use `_bytes` anywhere else, and we want naming here to be symmetrical with `ConvertResponse.result` (which we might want to rename to `payload`)

proto/buf/alpha/registry/v1alpha1/convert.proto
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+import "buf/alpha/convert/v1alpha1/convert.proto";
+import "buf/alpha/image/v1/image.proto";
+
+enum ConvertFormat {
```
bufdev, line 22: needs comment

proto/buf/alpha/convert/v1alpha1/convert.proto
```diff
@@ -0,0 +1,41 @@
...
+
+syntax = "proto3";
+
+package buf.alpha.convert.v1alpha1;
+
+// ModuleInfo contains information about a module that a serialized message belongs to.
+//
+// This definition is equivalent to buf.alpha.image.v1.ModuleInfo, but it's
+// purposefully redefined here to prevent import cycles in the generated code.
+message ModuleInfo {
```
bufdev, line 23: Why is this needed? This shouldn't be a separate type in a `convert` package, we should use the existing types

proto/buf/alpha/convert/v1alpha1/convert.proto
```diff
@@ -0,0 +1,41 @@
...
+  // commit is the repository commit.
+  //
+  // This will always be set.
+  string commit = 2;
+}
+
+// ModuleName is a module name.
+//
+// All fields will always be set.
+message ModuleName {
```
bufdev, line 37: same here

proto/buf/alpha/registry/v1alpha1/convert.proto
```diff
@@ -0,0 +1,57 @@
...
+  // module_info is the module source that defines the serialized message.
+  //
+  // Cannot be provided if image is provided.
+  buf.alpha.convert.v1alpha1.ModuleInfo module_info = 3;
+  // message_bytes is the serialized Protobuf message.
+  bytes message_bytes = 4;
+  // input_format is the format of the message_bytes.
+  ConvertFormat input_format = 5;
+  // output_format is the desired format of the output result.
+  ConvertFormat output_format = 6;
```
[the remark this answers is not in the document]
cyinma: I think the default behaviour will be more suitable to be handled by the CLI, and we require them to be set in the endpoint, otherwise the same behaviour will live in two places, which might be difficult to manage, but we can leave it as-is and discuss this latter.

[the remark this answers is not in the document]
cyinma: sounds good to me!

proto/buf/alpha/registry/v1alpha1/convert.proto
cyinma: we can leave building the image as an exercise for the caller, but this will add some overhead to the caller as they need to get the module, build it, and pass it in the request, instead of directly pass the module info and let us handle it in our side. And this might potentially makes using BSR module reference as source not a better option than other source (e.g. local image). How do you think?
We cannot use `oneof` in request field though, as it is disallowed by `go-apiclientgrpc` as Alex mentioned previously ([code](https://github.com/bufbuild/buf/blob/60cedc242b5231d92c37841278e223d6308a6637/private/pkg/protogenutil/protogenutil.go#L361))

proto/buf/alpha/registry/v1alpha1/convert.proto
cyinma: we cannot symmetrically name both fields in request and response as `payload`, as the interface generated has named return and this will make `payload` duplicated, so I am keeping the field in `ConvertResponse` `result`

proto/buf/alpha/registry/v1alpha1/convert.proto
bufdev: I'd prefer to keep it simple to start - a RPC endpoint should do one thing and do it well. Having an option between images and modules is a combination of build and convert, not just convert.

`go-apiclientgrpc` is something we wrote, we could always adjust that. But yea, keep just as `image` for now.

proto/buf/alpha/registry/v1alpha1/convert.proto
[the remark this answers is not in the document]
bufdev: I would be +1 to require these to be set, I agree with @cyinma

proto/buf/alpha/registry/v1alpha1/convert.proto
bufdev: We should adjust the implementation that name these then, we shouldn't be designing an API based around our internally-implemented Protobuf plugins.

proto/buf/alpha/registry/v1alpha1/convert.proto
cyinma: sounds reasonable 👍

proto/buf/alpha/registry/v1alpha1/convert.proto
cyinma: makes sense, I will have a look into the plugin and name both of the fields `payload`

proto/buf/alpha/registry/v1alpha1/convert.proto
amckinney: I agree that a `oneof` is the right way to do this - we'll change `go-apiclientgrpc` later if we ever need to.

Requiring the client to compile the `Image` is fine for now, but we might eventually need to revisit this if we want other systems (not the `buf` CLI) to call this API. Otherwise they'd be forced to write their own `--type` -> `Image` conversion logic in whatever programming language their client is written is, which is impractical and error prone.

Regardless, we definitely need the `Image` parameter option so it makes sense to just start there and solve this later (if ever).

proto/buf/alpha/registry/v1alpha1/convert.proto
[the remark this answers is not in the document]
amckinney: Sounds great, I also agree with @cyinma

proto/buf/alpha/registry/v1alpha1/convert.proto
amckinney: We could fix this by handling name collisions in the plugin - I imagine that'd be fairly simple.

Regardless, I'd be OK with leaving it as `ConvertRequest.payload` and `ConvertResponse.result` to unblock development and we can revisit this in a polishing follow-up before we consider the API released.

proto/buf/alpha/registry/v1alpha1/convert.proto
bufdev: I think if we want to revisit, we add a new endpoint, not add a `oneof` here.

bufdev (approved):

PR #964, opened 2022-02-22
handle name collision of request and response fields in generation

cyinma:
This is to handle name collision of request and response fields in `protoc-gen-go-api` and `protoc-gen-go-apiclientgrpc`,
for example, if both request and response have fields `payload`, the response field is now named as `payloadResponse`. 

Alternatively, we can just name the field to `_`, as this will be safer to prevent other collision (e.g. there might be another field called `payloadResponse` already) but the interface generated will have unclear argument names.

### DIFF
--- a/private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-api/main.go
+++ b/private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-api/main.go
@@ -48,11 +48,7 @@ func handle(helper protogenutil.NamedHelper, plugin *protogen.Plugin, file *prot
 			if err := protogenutil.ValidateMethodUnary(method); err != nil {
 				return err
 			}
-			requestParameterStrings, err := protogenutil.GetParameterStrings(g, method.Input.Fields)
-			if err != nil {
-				return err
-			}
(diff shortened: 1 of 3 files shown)


amckinney:
The `Response` suffix seems fine. Like you've already said, it's still prone to collisions if a `*Response` variable already exists. You can easily handle such cases by repeatedly adding a suffix or prefix to a string until you find one that's available.

This is a common strategy in code generators that need to generate a unique `import` alias. For example,

```go
import (
  json "encoding/json"
  _json "github.com/bufbuild/buf/pkg/json"
  __json "github.com/some/other/json"
)
```

In general, I like using a character that is impossible to be seen in the output set so you reduce such collisions. `_` is a good example because we are splitting on `_` and use `camelCase` in the generated names (thus it's impossible to see a natural _` in the result).

But given that all of this is just generated in a shim that is meant to be implemented by a human, I'm less concerned about the naming - the `Response` suffix you've selected is probably fine and is a little prettier than a `_` prefix.

---

Let's add in that check to prevent collisions by repeatedly suffixing. It's very unlikely to happen in practice, but we may as well implement it since it can be done with a simple loop.

bufdev (approved):

amckinney (approved):

PR #966, opened 2022-02-23
new get repository contributor endpoint

cyinma:
This add a new endpoint to get the contributor information in a repository


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -247,6 +249,18 @@ message ListRepositoryContributorsResponse {
   string next_page_token = 2;
 }
 
+message GetRepositoryContributorRequest {
+  // The ID of the repository for which to get the contributor information.
+  string repository_id = 1;
+  // The ID of the user for which to get the contributor information.
+  string user_id = 2;
```
robbertvanginkel, line 256: Can you add a few more details on the goal of this endpoint? Is this something that we need to get individually per users, something that should be batch requested from the frontend etc.

robbertvanginkel: Resolved offline. When adding a new direct contributer to a repository, they may already have access to this repo through their org membership. We want that UI to display their inherited access level too so we need a per user api here that's aimed at serving individual details.

robbertvanginkel (approved):

smallsamantha (approved):

PR #967, opened 2022-02-24
Add buf format

amckinney:
## Overview

This adds the `buf format` command, which formats Protobuf files (very similar to `gofmt`). Like `buf export`, this command works with source or module references (i.e. workspaces, directories, single `.proto` files, and module references like `buf.build/acme/petapis`).

## UX

The `buf format` command includes a couple important flags: one for diffing and another for rewriting the files in-place. They can be used separately or in combination with a variety of inputs. The flags work exactly like they do with `gofmt`.

For example,

```sh
# Prints the diff to stdout.
$ buf format -d

# Rewrite the file in-place with nothing written to stdout.
$ buf format proto/buf/alpha/image/v1/image.proto -w
 
# Prints the diff to stdout and rewrites all of the files in-place.
$ buf format -d -w
```

You can also write the formatted content to another directory, which works with module references, too:

```sh
$ buf format buf.build/acme/petapis -o formatted
```

## Formatting Opinions

The formatter is opinionated - it sorts imports, separates options from types, and more. Formatters are special such that their output does not need to be reliable - the formatted result doesn't need to conform to the same compatibility guarantee that the flags do. The current implementation's opinions are outlined below:

Files are written in the following structure:
* Syntax
* Package
* Imports (Sorted)
* Options (Sorted)
* Types, i.e. Messages, Extends, Enums, Services (Order is preserved)

For example,

```protobuf
enum Foo {
  option deprecated = true;

  FOO_UNSPECIFIED = 0;
  FOO_ONE = 1;
}

message Bar {
  option deprecated = true;
  reserved 4;
  extensions 10 to max;

  Foo foo = 1;
}
```

Newlines are preserved from the original source, but if multiple contiguous newlines exist, they are consolidated into one:

```protobuf
message Bar {
  option deprecated = true;
  reserved 4;
  extensions 10 to max;

  message Baz {
    Foo foo = 10;
  }

  Foo foo = 1;
}
```

See the diff here for the formatted result for all of our `proto` files (accomplished by running `buf format -w` at the root of the repository).

## Future Work

We can easily change these opinions now or over time. Plus, we could let users configure these formatting opinions and potentially even encode it in a user's `buf.yaml`. There's a lot that needs to go into the configuration design - we can handle that in a follow up if it's ever needed.


private/pkg/storage/copy.go
```diff
@@ -94,11 +96,22 @@ func CopyWithExternalPaths() CopyOption {
...
 
+// CopyWithWriteToExternalPaths returns a new CopyOption that says to copy and write
+// to external paths.
+//
+// The to WriteBucket must support setting external paths.
+func CopyWithWriteToExternalPaths() CopyOption {
+	return func(copyOptions *copyOptions) {
+		copyOptions.writeToExternalPaths = true
+	}
+}
```
amckinney, line 107: Flagging this one for discussion @bufdev - it's currently used to write the files in-place based on the files' external paths with `storage.Copy` (see `private/buf/cmd/buf/command/format/format.go`).

proto/buf/alpha/registry/v1alpha1/repository_track_commit.proto
```diff
@@ -21,21 +21,14 @@ import "google/protobuf/timestamp.proto";
 // RepositoryTrackCommit is the existance of a RepositoryCommit on a RepositoryTrack. Currently its only purpose is
 // for querying whether a RepositoryCommit is on a RepositoryTrack and determining it's sequence id.
 message RepositoryTrackCommit {
-  // We reserve field number '3' for id, but RepositoryTrackCommits have no id.
-  // string id = 1;
```
amckinney, line 25: I investigated this - it looks like a bug with `protocompile`. The leading detached comment for the first field in a message isn't attached to any of the nodes on the `*ast.FileNode`. I'll see if I can fix this.

private/buf/bufformat/formatter.go
```diff
@@ -0,0 +1,921 @@
...
+// Run runs the formatter and writes the file's content to the formatter's writer.
+func (f *formatter) Run() error {
+	return f.writeFile(f.fileNode)
+}
+
+// P prints a line to the generated output.
+func (f *formatter) P(elements ...interface{}) {
+	if len(elements) > 0 {
+		// Don't use an indent if we're just writing a newline.
+		fmt.Fprint(f.writer, strings.Repeat("  ", f.indent))
```
bufdev, line 51: I'm surprised this isn't caught by `errcheck`, i.e. I think you have to do `_, _ =`

private/buf/bufformat/formatter.go
```diff
@@ -0,0 +1,921 @@
...
+	"io"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/jhump/protocompile/ast"
+)
+
+// formatter writes an *ast.FileNode as a .proto file.
+type formatter struct {
```
bufdev, line 28: I might record a top-level `err` parameter, and add to it as you go using `multierr.Append`, and return it from `Run`. We might want to record errors in places such as `Out` if you go negative

private/pkg/storage/copy.go
bufdev: I don't quite understand it at first glance, can talk offline

proto/buf/alpha/registry/v1alpha1/image.proto
```diff
@@ -29,15 +29,12 @@ service ImageService {
 // should be masked in responses.
 enum ImageMask {
   IMAGE_MASK_UNSPECIFIED = 0;
-
```
bufdev, line 32: I am not sure if this should be deleting newlines

bufdev (commented):
Overall looks pretty sane to me. Obviously would be best if we 100% migrated to protocompile, but that requires the protoreflect fixes all being ported in

proto/buf/alpha/registry/v1alpha1/repository_track_commit.proto
jhump: I think this might be parsed as the _trailing_ comment for message `RepositoryTrackCommit`, not a leading or detached comment.

I just did a quick test with this file, to round-trip the file -- parse to AST and then print the AST back out. And this comment is in there. So it's in the AST somewhere. Perhaps it's just not where your formatter expects it?

amckinney: Thanks for the heads up, you're probably right. I tested this earlier by using the `ast.SimpleVisitor` and printed the leading/trailing comments attached to every node. I might have made a mistake somewhere tho - I'll figure it out!

private/buf/bufformat/testdata/proto3/all/v1/all.golden.proto
```diff
@@ -0,0 +1,117 @@
...
+
+    int64 foo1 = 8;
+    string foo2 = 9;
+    // oneof-field comment 2
+    int64 woot3 = 10 [(custom.field_option) = false, (custom.field_thing_option) = {foo:8}]; // oneof-option-field-inline comment 3
+    // oneof-field comment 2
+    custom.Thing custom_field = 11; // oneof-option-field-inline comment 4
+  }
+  int64 integer_field = 12 [(custom.repeated_field_option) = 1, (custom.repeated_field_option) = 2];
+  int64 integer_field_two = 13 [(custom.field_thing_option) = {foo:9 repeated_foo:[1, 2]}, (custom.repeated_field_thing_option) = {foo:1 recursive:{foo:2 repeated_foo:[3, 4]}}, (custom.repeated_field_thing_option) = {foo:5 recursive:{foo:6 repeated_foo:[7, 8]}}, (custom.repeated_field_thing_option) = {foo:1 repeated_thing:[{foo:1 repeated_foo:[1, 2]}, {foo:3 repeated_foo:[3, 4]}]}];
```
amckinney, line 66: Formatting large compact options in one line probably isn't reasonable. I'll see what I can do to spread this across multiple lines when the options are long enough.

proto/buf/alpha/registry/v1alpha1/repository_track_commit.proto
amckinney: Discussed offline - it turns out that it's because the comment is attached to the `{` node - not the message itself. This is also true for other composite nodes like `enum`, `extend`, etc. I'll revisit this and make sure to handle these cases.

private/buf/bufformat/testdata/proto3/all/v1/all.golden.proto
```diff
@@ -0,0 +1,117 @@
...
+  // body comment 3
+  // nested-message comment
+  message Three {
+    int64 three_one = 1;
+  }
+
+  /*
+  c-style comment
+  */
+  int64 one_one = 1 [(custom.field_option) = true];
```
kindermoumoute, line 49: @amckinney for alignment we found the following [clang-format rules](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) quite convenient:

```
AlignConsecutiveDeclarations: true
AlignConsecutiveAssignments: true
```

For this few lines formatting would look like this:

```proto
  int64              one_one   = 1 [(custom.field_option) = true];
  Two                one_two   = 2 [(custom.field_option) = false, (custom.field_thing_option) = {foo:6}]; // field-option-inline comment
  Three              one_three = 3; /* field-inline c-style comment */
  map<string,uint32> map_field = 4;
```

<details><summary>Click to show longer example</summary>
<p>

#### From one of our public proto

```proto
message CreateClusterRequest {
  message OpenIDConnectConfig {
    string                      issuer_url      = 1;
    string                      client_id       = 2 [(validate.rules).string.min_len = 1];
    google.protobuf.StringValue username_claim  = 3 [(validate.rules).string.min_len = 1];
    google.protobuf.StringValue username_prefix = 4 [(validate.rules).string.min_len = 1];
    scaleway.std.StringsValue   groups_claim    = 5;
    google.protobuf.StringValue groups_prefix   = 6 [(validate.rules).string.min_len = 1];
    scaleway.std.StringsValue   required_claim  = 7;
  }

  message PoolConfig {
    message UpgradePolicy {
      google.protobuf.UInt32Value max_unavailable = 1;
      google.protobuf.UInt32Value max_surge       = 2;
    }

    string                      name               = 1 [(validate.rules).string.min_len = 1];
    string                      node_type          = 2;
    google.protobuf.StringValue placement_group_id = 3 [(validate.rules).string.uuid = true];
    bool                        autoscaling        = 4;
    uint32                      size               = 5;
    google.protobuf.UInt32Value min_size           = 6;
    google.protobuf.UInt32Value max_size           = 7;
    Runtime                     container_runtime  = 8 [(validate.rules).enum.defined_only = true];
    bool                        autohealing        = 9;
    repeated string             tags               = 10;
    map<string, string>         kubelet_args       = 11;
    UpgradePolicy               upgrade_policy     = 12;
    string                      zone               = 13;
  }

  message AutoscalerConfig {
    google.protobuf.BoolValue   scale_down_disabled              = 1;                                             // default to false
    google.protobuf.StringValue scale_down_delay_after_add       = 2;                                             // default to 10m
    AutoscalerEstimator         estimator                        = 3 [(validate.rules).enum.defined_only = true]; // default to binpacking
    AutoscalerExpander          expander                         = 4 [(validate.rules).enum.defined_only = true]; // default to random
    google.protobuf.BoolValue   ignore_daemonsets_utilization    = 5;                                             // default to false
    google.protobuf.BoolValue   balance_similar_node_groups      = 6;                                             // default to false
    google.protobuf.Int32Value  expendable_pods_priority_cutoff  = 7;                                             // default to -10
    google.protobuf.StringValue scale_down_unneeded_time         = 8;                                             // default to 10m
    google.protobuf.FloatValue  scale_down_utilization_threshold = 9 [(validate.rules).float = { gt: 0, lt: 1 }]; // default to 0.5
    google.protobuf.UInt32Value max_graceful_termination_sec     = 10;                                            // default to 600
  }

  message AutoUpgrade {
    bool              enable             = 1;
    MaintenanceWindow maintenance_window = 2;
  }

  oneof ProjectIdentifier {
    string organization_id = 1 [(validate.rules).string.uuid = true, deprecated = true];
    string project_id      = 14 [(validate.rules).string.uuid = true];
  }

  string              type                   = 17;
  string              name                   = 2 [(validate.rules).string.min_len = 1];
  string              description            = 3;
  string              version                = 4;
  repeated string     tags                   = 5;
  CNI                 cni                    = 6 [(validate.rules).enum.defined_only = true];
  bool                enable_dashboard       = 7 [deprecated = true];
  Ingress             ingress                = 8 [(validate.rules).enum.defined_only = true];
  repeated PoolConfig pools                  = 9;
  AutoscalerConfig    autoscaler_config      = 10;
  AutoUpgrade         auto_upgrade           = 11;
  repeated string     feature_gates          = 12;
  repeated string     admission_plugins      = 13;
  OpenIDConnectConfig open_id_connect_config = 15;
  repeated string     apiserver_cert_sans    = 16;
}

```

</p>
</details>

Do you think it's interesting to add this kind of formatting option in the buf formater?

amckinney: Nice, I think this is totally reasonable. I'd like to start with with a configuration-less formatter implementation, but I imagine that we'll follow up with shortly after with a `format` configuration (see summary above for details).

I'll make note of this for future changes - thanks for the suggestion!

amckinney:
It turns out that there is a signifiant amount of complexity involved in preserving all of the possible comments. Every token (e.g. `'{'`, `';'`, etc.) can have comments attributed to it, so the formatter needs to handle each node separately, and stitch them all together in unique ways for each composite node. For example, fields need a space between each token, whereas map fields do not.

There are still some bugs in the current state, but a lot of the issues have been addressed. I'll need to revisit some of the implementation to see what we can DRY up (many node writing functions are verbose and share a similar shape between each other), but I'll need to make sure any changes don't compromise the readability of each individual node writing function.

I'll come back to this shortly!

private/buf/bufformat/bufformat.go
```diff
@@ -0,0 +1,78 @@
...
+			return nil, err
+		}
+		defer func() {
+			retErr = multierr.Append(retErr, moduleFile.Close())
+		}()
+		moduleFiles = append(moduleFiles, moduleFile)
+	}
+	fileNodes := make([]*ast.FileNode, 0, len(moduleFiles))
+	for _, moduleFile := range moduleFiles {
+		fileNode, err := parser.Parse(moduleFile.ExternalPath(), moduleFile, reporter.NewHandler(nil))
```
bufdev, line 49: This should be done in parallel right? This is one of the goals of protocompile

amckinney:
Nearly everything is good to go here. The only remaining details have to do with our formatting opinions.

In particular, we need to decide whether or not we want to preserve the newline between elements that are potentially intentional from the user. For example, consider the following examples:

```protobuf
message Object {
  string key = 1;

  bytes value = 2;
  map<string, string> meta = 3;
}
```

Should the newline be preserved between `key` and `value`, or should they all be consolidated into a single block? Does anything change if there are leading comments attached to them?

```protobuf
message Object {
  string key = 1;

  // value is an arbitrary object value.
  bytes value = 2;

  map<string, string> meta = 3;
}
```

For now, the formatter is simple: these lines are consolidated into a single block, regardless of the original whitespace and leading comments:

```protobuf
message Object {
  string key = 1;
  // value is an arbitrary object value.
  bytes value = 2;
  map<string, string> meta = 3;
}
```

We can choose to change these opinions over time, but we should still spend time on deciding what we should do first. The most important thing (for backward compatibility's sake) is that the command's flags and semantics are stable. In other words, the formatted result is technically flexible, but I think we should be very careful with how often we decide to change our opinion (if ever).

bufdev:
Newlines should likely be preserved

amckinney:
I've made the change to preserve the newlines from the original source. If there are contiguous newlines, they are consolidated into one.

There's a known bug with respect to formatting comments attached to compact option elements, but it should only affect a very small number of users. I've included a test here (skipped for now), and opened an [issue](https://github.com/jhump/protocompile/pull/3) to track.

I'll follow up with some additional testing to verify some other edge cases, but this is otherwise good to go.

private/buf/cmd/buf/command/format/format.go
```diff
@@ -0,0 +1,388 @@
...
+		return storage.WalkReadObjects(
+			ctx,
+			originalReadWriteBucket,
+			"",
+			func(readObject storage.ReadObject) error {
+				formattedReadObject, err := formattedReadBucket.Get(ctx, readObject.Path())
+				if err != nil {
+					return err
+				}
+				file, err := os.OpenFile(readObject.ExternalPath(), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
```
bufdev, line 344: We have `storage.Copy` to a storage OS bucket, why can't we use that?

bufdev:
Make sure we have tasks to update vim-buf and vscode-buf after this

private/buf/cmd/buf/command/format/format.go
```diff
@@ -0,0 +1,388 @@
...
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <input>",
+		Short: "Format all Protobuf files from the specified input and output the result.",
+		Long:  bufcli.GetInputLong(`the source or module to format`),
```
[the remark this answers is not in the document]
amckinney: Done - let me know what you think.

private/buf/bufformat/testdata/proto3/file/v1/empty.golden.proto
```diff
@@ -0,0 +1 @@
+
```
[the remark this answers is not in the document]
amckinney: Good call, fixed.

private/buf/cmd/buf/command/format/format.go
```diff
@@ -0,0 +1,388 @@
...
+		}
+		readWriteBucket, err = storageosProvider.NewReadWriteBucket(
+			flags.Output,
+			storageos.ReadWriteBucketWithSymlinksIfSupported(),
+		)
+		if err != nil {
+			return err
+		}
+	}
+	if protoFileRef, ok := sourceOrModuleRef.(buffetch.ProtoFileRef); ok {
```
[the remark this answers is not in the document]
amckinney: Good catch, that's fixed now.

private/buf/cmd/buf/command/format/format.go
amckinney: In short, it's a lot simpler to use this approach since it can be applied to all of our output types in the same way. I left a comment to justify the current implementation - let me know what you think.

private/buf/cmd/buf/command/format/format.go
```diff
@@ -0,0 +1,536 @@
...
+package simple;
+
+message Object {
+  string key = 1;
+  bytes value = 2;
+}
+
+Write the result to a specified output file or directory with -o. For example,
+
+# Write the formatted file to another file
```
bufdev, line 92: Can the first example be what happens be `buf format` and then `buf format -w`, explaining that this will use `.`, ie whatever module or workspace is in the current directory? It might also just be a note that "most people will want to do `buf format -w`".

bufdev (approved):
Approving assuming documentation comment is fixed

private/buf/cmd/buf/command/format/format.go
```diff
@@ -0,0 +1,388 @@
...
+		flags.Paths,
+		flags.ExcludePaths,
+		false,
+	)
+	if err != nil {
+		return err
+	}
+	var readWriteBucket storage.ReadWriteBucket
+	if flags.Output != "-" {
+		if err := os.MkdirAll(flags.Output, 0755); err != nil {
```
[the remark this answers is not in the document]
amckinney: Discussed offline - 

We eventually want to deprecate `bufwire` since it's kind of devolved into a catch-all place for functionality. Given that this functionality is specific to the format command, we'll leave this as-is.

The single file input issue was addressed.

private/buf/cmd/buf/command/format/format.go
amckinney: Yea I agree - I'll rearrange this a bit.

private/buf/bufformat/testdata/proto3/all/v1/all.golden.proto
kindermoumoute: Opened a PR to keep track of this https://github.com/bufbuild/buf/issues/1312

PR #968, opened 2022-02-24
Remove space and new line from token

mfridman:
This PR adds logic to trim new lines and remove spaces from the user-supplied token.

In our docs for [CI authentication](https://docs.buf.build/bsr/authentication#ci-authentication) we have the following:

```
echo ${BUF_API_TOKEN} | buf registry --username ${BUF_USER} --token-stdin
```

Some users might be using `echo -n` and others might not. Before we added logic for validating tokens in https://github.com/bufbuild/buf/issues/861 the new line didn't matter, worst case you got an extra line in your .netrc file following the `password` field.

But now we're sending this token to the backend for validation, and `token\n` will result in an error:

```
Failure: invalid token provided
```

Again, we could tell users to `echo -n`, but to improve the user experience let's just trim the lines and spaces altogether.

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/bufbuild/buf/private/buf/bufcli"
 	"github.com/bufbuild/buf/private/bufpkg/bufrpc"
@@ -26,7 +27,6 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
-	"github.com/bufbuild/buf/private/pkg/stringutil"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 )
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -137,6 +138,7 @@ func run(
 	if err != nil {
 		return err
 	}
+	token = stringutil.TrimLines(token)
```
amckinney, line 141: I know you mentioned this in the summary, but we could return an error here and include an error message with something like `"token contains a newline - did you mean to use 'echo -n'?"`

Trimming the newlines is probably harmless, I'm just always weary of silently adapting the user's input if we can avoid it. Let me know what you think - I'm not strongly opposed.

lrewega: Ah, good to know about `stringutil.TrimLines`. I definitely think this is one of those few Postel's law situations where liberally accepting is not likely to cause problems and makes for a much better user experience. Perhaps a comment to that effect is warranted? Otherwise this LGTM.

lrewega: I suppose we also have the same problem with username, potentially?

lrewega: > our stringutil.TrimLines(..) also calls strings.TrimSpace(..). Maybe strings.TrimSpace(strings.TrimSuffix(s, "\n")) would be sufficient, but iirc we tend to prefer our utils.

FWIW `strings.TrimSpace` also strips newlines, as they are considered white space per `unicode.IsSpace`.

lrewega (approved):

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
mfridman: Ah yes, forgot about that. Switched to `strings.TrimSpace`, great suggestion! Also added a comment and an extra check for empty tokens.

PR #971, opened 2022-02-25
Add track and tag count RPCs

WillAbides:
This will enable BSR to display counts of a repository's tracks and tags without making more costly `List*` calls.


robbertvanginkel:
Can you add details on how these are intended to be used and what the tradeoff is for adding extra endpoints rather than fields on a `GetRepository` response?

WillAbides:
@robbertvanginkel I opted to make these separate RPCs because `Repository` appears to be an intentionally simple message that mostly mirrors the repositories table. I assumed that is intentional.

mfridman:
What about re-using the existing `GetRepositorySettings`, which already does some counting? At most the FE will have to make 2 calls, one for repo and one for any metadata (or settings, as we called it).

WillAbides:
@mfridman I like that idea. I hadn't looked at `GetRepositorySettingsResponse` before and assumed it was about configuration settings.

I was about to suggest adding `total_count` to `ListRepositoryTracks` and `ListRepositoryTags`.  I think I like the `GetRepositorySettings` idea better though.

cyinma:
> What about re-using the existing `GetRepositorySettings`, which already does some counting? At most the FE will have to make 2 calls, one for repo and one for any metadata (or settings, as we called it).

`GetRepositorySettings` is currently limited to repository owner or admin only, do we plan to display the track and tag count to repository owner or admin only too?

seankimdev:
It would be used to display the numbers shown below in the repository page. I think it should be visible to all users with access to the repository.
image

WillAbides:
It sounds like `GetRepositorySettings` won't work because of permissions. Maybe we could add a `GetRepositoryDetails` RPC that would contain these counts and any other occasionally used details that come up later.

robbertvanginkel:
> opted to make these separate RPCs because `Repository` appears to be an intentionally simple message that mostly mirrors the repositories table. I assumed that is intentional

I don't think our public api interface should be influenced by our internal table layout. If this is information that is generally useful to anyone who asks for a repository details, putting it in there seems fine to me. If we don't want to return this by default, I think I would advocate for a flag/enum for the level of details on `GetRepository` rather than introducing a new endpoint.

mfridman:
Ooops, I didn't realize that Settings endpoint was admin only. In which case I think we should stick that metadata on the GetRepositoryResponse object. E.g.,

```proto
message GetRepositoryResponse {
  Repository repository = 1;
  RepositoryCounts repository_counts = 2;
}

message RepositoryCounts {
  uint32 tags_count = 1;
  uint32 tracks_count = 2;
}
```

And then the counts could always be available (by default), or omitted by default and opt-in via an enum or flag on the request. I think we should just always include them and the rest is an implementation detail.

WillAbides:
@mfridman I'll give that a try

mfridman (approved):
I don't love the multi returns, but that's an implementation detail of our interface generator.

lgtm

robbertvanginkel (approved):

PR #989, opened 2022-03-09
add machineactor to audit log envelope

smallsamantha:
Today all requests will look they were issued by an anonymous user, but in some cases the actor is really a service/machine. This PR adds `MachineActor` to represent these cases


proto/buf/alpha/audit/v1alpha1/envelope.proto
```diff
@@ -614,6 +615,10 @@ message UserActor {
   string subject = 5;
 }
 
+message MachineActor {
```
mfridman, line 618: I think my only worry is we use the term "machine" for things like "machine token" and those belong to users. Maybe `ServiceActor` would be more accurate? No strong opinion though..

proto/buf/alpha/audit/v1alpha1/envelope.proto
```diff
@@ -614,6 +615,10 @@ message UserActor {
...
+  string service_name = 1;
```
robbertvanginkel, line 619: Can you list the cases in which we'd start using service_name here? If a user triggers something on goproxyd and that reaches out to bufd, that would still be a user actor right?

smallsamantha: anything user triggered would still be user actor but for something like auto provisioning calling `AddUserToOrganization` this would allow us to use `envoyauthd` as the `service_name` (it's recorded as `anonymous` right now)

mfridman (approved):
I think I'd prefer the term ServiceActor, which may be an internal service today, or an external service in the future. But no strong opinion.

proto/buf/alpha/audit/v1alpha1/envelope.proto
```diff
@@ -531,6 +531,7 @@ message Event {
   // of the event.
   oneof actor {
     UserActor user = 2;
+    ServiceActor service_actor = 59;
```
smallsamantha, line 534: @mfridman i updated to ServiceActor as suggested, however, i'm not going to change field 5 to `service_name` and the new field as `service` because even though a field rename is not typically a breaking change it would be in this case because these entries are stored as JSONB fields in postgres (table structure example below). the inconsistency in the actor field naming grosses me out a little bit, what do you think?
```
CREATE TABLE audit_logs (
  ...
  event JSONB NOT NULL,
  ...
);
```

mfridman (approved):
Sounds good, let's just go with machine as initially proposed. Feel free to merge.

PR #1002, opened 2022-03-14
Add ListRepositoryCommitsOnTrack call

rubensf:
This is different from ListRepositoryCommitsByReference
as in that it returns _all_ commits starting at reference.


rubensf:
Added `repository_track_id` as parameter

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -79,12 +82,34 @@ message ListRepositoryCommitsByBranchResponse {
...
 }
 
+message ListRepositoryCommitsOnTrackRequest {
+  // The owner of the repository which the repository reference belongs to.
+  string repository_owner = 1;
+  // The name of the repository which the repository reference belongs to.
+  string repository_name = 2;
+  // The id of the track which to list commits from.
+  string repository_track_id = 3;
+  // The reference used to resolve track commits. Can be a track, tag or commit.
```
WillAbides, line 92:
```suggestion
  // The reference used to resolve track commits. Can be a track, tag or commit name.
```

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -79,12 +82,34 @@ message ListRepositoryCommitsByBranchResponse {
...
+  // Commits will be returned starting at reference.
+  // Using a track as a reference will return results if the reference track head
+  // is in the queried track's history.
```
WillAbides, line 95: Can you specify what happens when reference is empty?

WillAbides (approved):
Looks good. Just a couple of comment suggestions.

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
rubensf: Done

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
rubensf: Done.

PR #1004, opened 2022-03-15
Upgrade to Go 1.18

bufdev:
Docker images should be released first.


bufdev:
When https://hub.docker.com/_/golang has 1.18 tags (not 1.18rc1 tags), this should be good to go.

bufdev:
Docker has been updated.

cyinma (approved):

PR #1009, opened 2022-03-16
Change track_id to track_name

rubensf:
So that the feature can be more user friendly


WillAbides (approved):

bufdev:
This was a breaking change - is this a new API?

WillAbides:
@bufdev Yes it isn't implemented yet.

bufdev:
OK great, just making sure

PR #1011, opened 2022-03-18
add latest commit info to repository track

cyinma:
This added the latest commit info including create time and author to the `RepositoryTrack`.

There are other options like embed the `RepositoryCommit` into `RepositoryTrack`, or a separated endpoint that return a list of new message that includes a `RepositoryTrack` and a `RepositoryCommit`, but I think we tend to avoid nested message structure, so it might be a better approach to expand it in `RepositoryTrack` like this.


proto/buf/alpha/registry/v1alpha1/repository_track.proto
```diff
@@ -27,6 +27,10 @@ message RepositoryTrack {
   // google.protobuf.Timestamp update_time = 3;
   // The name of the repository track, i.e. "v1".
   string name = 4;
+  // The create time of the latest commit to the track.
+  google.protobuf.Timestamp latest_commit_time = 5;
+  // The username of the author of the latest commit to the track.
+  string latest_commit_author = 6;
```
robbertvanginkel, line 33: Adding a RepositoryCommit message here seems more extensible than pulling in just the fields we need. Tracks can exist without commits on them and having to check each individual field would get messy, especially with the empty string.

cyinma: updated 👍 this was also my initial thought, but I was not sure do we want to avoid embedding message, however I think it is a good point that it will be easier to represent a track that without commits on them

robbertvanginkel (approved):

Issue #1013, opened 2022-03-20
Temporarily remove private/usage restrictions and do 1.1.1 release


bufdev:
See https://github.com/bufbuild/homebrew-buf/issues/8
See https://github.com/golang/go/issues/51831

There's no immediately apparent workaround here - unless this is fixed, we can't use this technique anymore. We need to make sure homebrew builds aren't broken, so we should temporarily remove the `check()` call in `init()` and do a micro release.

PR #1016, opened 2022-03-21
Temporarily remove `check` from `init` in private usage

doriable:
This is currently breaking our Homebrew build: https://github.com/bufbuild/buf/issues/1013
Once this is merged, we will make a `v1.1.1` release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Temporarily remove private usage check on private packages to unbreak Homebrew releases.
+- Remove check for proper usage of private packages due to a breaking change made in the Golang standard library in 1.18.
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Temporarily remove private usage check on private packages to unbreak Homebrew releases.
```
bufdev, line 5: ` - Remove check for proper usage of private packages due to a breaking change made in the Golang standard library in 1.18.`

private/usage/usage.go
```diff
@@ -26,9 +26,10 @@ import (
 const debugBinPrefix = "__debug_bin"
 
 func init() {
```
bufdev, line 28: Just comment out the whole init function

PR #1018, opened 2022-03-21
Fix build scripts and re-enable private usage check

lrewega:
A change introduced in go 1.18, [CL 339170][0], altered
`runtime/debug.BuildInfo` details for the `command-line-arguments`
pseudo-package, which represents file arguments provided to tools such
as `go build` or `go run`. As such, it's no longer possible to determine
the package name associated with the `main` function when the file it's
defined in is passed as a file argument. Let's avoid
`command-line-arguments` altogether by changing how we build binaries,
instead referring to `main` packages using package path syntax.

Now that we can determine the package path of the `main` function again,
re-introduce invoking `private/usage.check` in `private`'s `init()`.

See also #1013 and https://github.com/bufbuild/homebrew-buf/issues/8

[0]: https://go-review.googlesource.com/c/go/+/339170/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Fix build scripts to avoid using the `command-line-arguments` pseudo-package
+  when building binaries and re-introduce checking for proper usage of private
+  packages.
 
 ## [v1.1.1] - 2022-03-21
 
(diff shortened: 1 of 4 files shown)


doriable (approved):
Chatted with @lrewega, this seems like a very sane fix given that building with the target file is not module-aware.

bufdev:
We've confirmed this works with I assume right? Heh

doriable:
> We've confirmed this works with I assume right? Heh

There's a better way to do this, but I pulled the branch and ran `make bufrelease` and it's all good (the brew script seems to use the same invocation as the release script, other than the `-a` flag, so it seems reasonable?).

bufdev (approved):

PR #1024, opened 2022-03-23
update ListUsers with sort/filter [WIP]

smallsamantha:
Putting up for discussion based on: 
https://www.notion.so/bufbuild/Search-and-filter-bce8caf78d53413a9c71260a79f96e12


bufdev:
Do we still want to pursue this?

Issue #1030, opened 2022-03-25
Make `buf format --diff` exit with non-zero code if a difference is found


christophermaier:
We're using `buf format --diff` in our CI pipelines to check that our protobuf files are properly formatted. However, since `buf format --diff` uses an exit code of `0` whether there are differences or not, we have to jump through some hoops to make our CI job fail if the files aren't formatted properly:

```sh
output="$(buf format --diff)"
if [ -n "${output}" ]; then
    echo "${output}"
    echo
    echo "Protobuf files are not properly formatted!"
    exit 1
fi
```
If `buf format --diff` could exit with something other than `0` when differences are found (like the `diff` command itself does), this would make it easier to incorporate into various automation scenarios.

Thanks!

rcrowe:
We run our CI with...

```
buf format -w
git diff --exit-code
```

So perhaps we could get a `--exit-code` flag to match git?

bufdev:
Exiting with 0 on diff is pretty standard for formatters, see i.e. `gofmt`. As we see it, producing a diff is the goal of the command, and the production of this diff means you didn't have an error, whereas if your files weren't readable for example, this is an error. While I understand there are arguments both ways (and understand the pain of above), we're going to stick with this pre-established convention.

bufdev:
This is a close call. On one hand, this sticks with existing conventions, which we're inclined to do (i.e. `gofmt`). In our view, returning a diff is a valid thing to do, whereas a file not being readable is an error. On the other, I get the `diff` returning error argument. We'll think about it briefly, but are unlikely to change.

wimax-grapl:
One option would be to expose this as a `--exit-code` flag. The most obvious precedent is `git`! 
```
# Make sure our tree is clean.
git diff HEAD --exit-code > /dev/null 2>&1 || (echo "Your git tree is dirty; bailing!" && exit 1)
```

bufdev:
Good idea.

bufdev:
This is fixed in #1033 and will go out in a release we are doing today.

wimax-grapl:
Thanks!!

christophermaier:
Thanks for such a quick turnaround!

doriable:
https://github.com/bufbuild/buf/releases/tag/v1.3.0 :)

PR #1033, opened 2022-03-25
Add --exit-code to buf format

bufdev:
Fixes https://github.com/bufbuild/buf/issues/1030.

This makes it so that `buf format` exits with a non-zero exit code if files were not already formatted. Common invocations:

```
# Print the diff and exit with non-zero exit code if there is a diff
$ buf format -d --exit-code
# Print nothing and exit with non-zero exit code if there is a diff
$ buf format --exit-code >/dev/null
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `--exit code` flag to `buf format` to exit with a non-zero exit code if
+  the files were not already formatted.
 
 ## [v1.2.1] - 2022-03-24
 
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -1944,19 +1944,6 @@ func TestFormatDiff(t *testing.T) {
 @@ -1,13 +1,7 @@
 -
  syntax = "proto3";
- 
--
--
- package diff;
- 
--
--message   Diff  {
--
-+message Diff {
-   string content = 1;
--
--  }
-+}
 `,
 	)
 	testRunStdout(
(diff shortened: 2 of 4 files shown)


private/buf/cmd/buf/buf_test.go
```diff
@@ -1944,19 +1944,6 @@ func TestFormatDiff(t *testing.T) {
 @@ -1,13 +1,7 @@
 -
  syntax = "proto3";
-
```
bufdev, line 1947: This test is affected if there is whitespace or not in the diff - this is kind of a mess across IDEs, just simplifying it to test what we need to

private/pkg/app/appcmd/appcmdtesting/appcmdtesting.go
```diff
@@ -115,8 +115,18 @@ func RunCommandExitCode(
 	// make the use something different than the actual command
 	// to make sure that all code is binary-name-agnostic.
 	use := "test"
-	stderrCopy := bytes.NewBuffer(nil)
 	stdoutCopy := bytes.NewBuffer(nil)
+	if stdout == nil {
```
bufdev, line 119: This came up as a panic in the tests when I had `stdout` as nil - this makes it so that `stdout/stderr` can be nil

doriable (approved):

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `--exit code` flag to `buf format` to exit with a non-zero exit code if
+  the files were not already formatted.
```
wimax-grapl, line 6: (you may mean --exit-code not --exit code)

doriable: lol, yes -- we were a bit excited about this. Thanks for catching this and we'll update the changelog :)

PR #1042, opened 2022-03-28
Add UpdateRepositorySettingsByName

rubensf:


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -101,6 +101,8 @@ service RepositoryService {
   rpc GetRepositoryContributor(GetRepositoryContributorRequest) returns (GetRepositoryContributorResponse);
   // GetRepositorySettings gets the settings of a repository.
   rpc GetRepositorySettings(GetRepositorySettingsRequest) returns (GetRepositorySettingsResponse);
+  // UpdateRepositorySettingsByName updates the settings of a repository. Unspecified values mean no change is made.
```
cyinma, line 104: I think the unspecified values handling comment should be put in the request message instead, as it is something by field (e.g. you must still provide `owner_name` and `repository_name`)

rubensf: Done

cyinma (approved):
LGTM!

robbertvanginkel (approved):

PR #1045, opened 2022-03-28
Format compound string literals across multiple lines

### LINKED ISSUE #1036: `buf format`: consider preserving line breaks between multi-line strings
See here for an example of how `buf format` differs from `clang-format`:

https://github.com/einride/iam-go/pull/251/files#diff-5707dfb52cbebbde2fb0e443d7a63565abf47c36c0adb67ee91f98ad0e623120R253

It might be preferable to preserve any line breaks between adjacent strings, to improve formatting and enable wrapping of long option strings onto multiple lines.

amckinney:
Fixes #1036

This updates `buf format` so that adjacent string literals are written across multiple lines. For example,

```protobuf
syntax "proto3";

package example;

import "custom.proto";

option (custom) =
  "one,"
  "two,"
  "three"
```

Note that this takes a different stance than `clang-format` - the strings are **_not_** aligned according to the first element like so:

```protobuf
syntax "proto3";

package example;

import "custom.proto";

option (custom) = "one,"
                  "two,"
                  "three"
```

The former approach is more consistent with the rest of the formatter's style, and it's actually the same style seen in GoogleAPIs (shown below).

```protobuf
option (google.api.oauth_scopes) =
  "https://www.googleapis.com/auth/analytics,"
  "https://www.googleapis.com/auth/analytics.readonly";
```

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -322,7 +322,6 @@ func (f *formatter) writeFileOption(optionNode *ast.OptionNode) {
 //
 func (f *formatter) writeOption(optionNode *ast.OptionNode) {
 	f.writeOptionPrefix(optionNode)
-	f.Space()
 	if optionNode.Semicolon != nil {
 		f.writeInline(optionNode.Val)
 		f.writeLineEnd(optionNode.Semicolon)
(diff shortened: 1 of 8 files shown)


amckinney:
@odsod seeing as you opened #1036, I'd like to hear your feedback on the adjusted formatting style - see the summary for details.

bufdev (approved):

odsod:
@amckinney I tried it out on https://github.com/einride/iam-go/pull/251 and I think it works great! :tada:

PR #1051, opened 2022-04-03
linux install

randomizedcoder:


bufdev:
This is actually accurate - go install is a universal way to install on any platform with go, but this documentation says to build from the head of main, not from one of the releases, which is also not the recommended install path for most users.

bufdev:
Thanks for this, but as I don't think this is what we want in our README for linux - we can re-evaluate the README going forward though.

PR #1057, opened 2022-04-07
Update ErrNoConfigFile error message

domdoescode:
Command was moved from "buf config init" to "buf mod init", [docs were updated](https://github.com/bufbuild/buf/issues/958) but this error message wasn't updated to reflect that change.


amckinney (approved):
Thanks! Please update the `CHANGELOG.md` with the following line under the `[Unreleased]` section:

```
- Fixed an error message to recommend 'buf mod init' instead of 'buf config init'.
```

PR #1078, opened 2022-04-20
Adds a Github Service definition

smaye81:
This adds a protobuf definition for a Github Service to interact with Github integration information.


proto/buf/alpha/registry/v1alpha1/github.proto
```diff
@@ -0,0 +1,36 @@
...
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+message GithubAppConfig {
+    string app_id = 1;
+}
+
+message GetGithubAppConfigRequest{
+    string id = 1;
```
smaye81, line 24: I'm guessing `id` is not a good name, but not sure what this ID represents.

smaye81:
@SyrieBianco @srikrsna-buf Adding you two for now just to make sure I'm on the right track.  I'm sure there's some conventions I'm not following, so wanted to let you see it before I went to far in code.

proto/buf/alpha/registry/v1alpha1/github.proto
srikrsna-buf: For now we can leave it blank as it should only have one entry per deployment.

bufdev:
Should this all be in the private APIs, or the public ones?

srikrsna-buf:
It is for the admin panel integration

proto/buf/alpha/registry/v1alpha1/github.proto
smaye81: OK, meaning just an empty request object?

srikrsna-buf (approved):
Checked with @robbertvanginkel. These APIs go here. Looks good to me.

PR #1081, opened 2022-04-21
Add stubs for Module Service

rubensf:
We're in the process of renaming

- repository -> module
- module -> snapshot

This includes renaming all our gRPC services. In this PR
we add the Module counterparts of the Repository Service.


bufdev:
Just to make sure it's clear - if any non-`buf beta` command depends on any service we intend to remove from v1alpha1, we can't remove that service - `buf` can't break because of this rename other than `buf beta` commands

rubensf:
Yes - this isn't breaking anything (or really effectively changing anything so far)

robbertvanginkel:
If we rename a set of concepts, I think to me that indicates a newer version of the api that does not belong side by side to the older concepts in a `buf.alpha.registry.v1alpha1`. Having two sets of services with different messages representing the same resources feels like it will get messy pretty fast.

Should this be in some `beta` or `v2alpha1` or `v1alpha2` package?

rubensf:
After some slack conversation - yup, that actually seems a pretty good idea. Just all new protos in v1alpha2.

proto/buf/alpha/registry/v1alpha2/module.proto
```diff
@@ -0,0 +1,291 @@
...
+import "buf/alpha/registry/v1alpha2/role.proto";
+import "google/protobuf/timestamp.proto";
+
+enum Visibility {
+  VISIBILITY_UNSPECIFIED = 0;
+  VISIBILITY_PUBLIC = 1;
+  VISIBILITY_PRIVATE = 2;
+}
+
+message Module {
```
bufdev, line 29: We already have something named `Module`, this should be something different, but I don't know what

rubensf: Isn't this what we want to change? Repository -> Module, and Module -> Snapshot?

Because we already have both a `module/v1alpha1/module.proto` and a `repository/v1alpha1/module.proto` (and the "repository" version has `Local` varieties of the "module" version, which I personally find confusing at a first glance :).

We could go with `RegistryModuleService` or something of the sorts, I guess.

rubensf: though `RegistryModuleService` would probably be a bad naming practice/redundant since we would arledy be referencing this with something like `registry.v1alpha2.RegistryModuleService`

rubensf: I went with `RemoteModule` (thought about `HostedModule` which sounded less explicit, and `RemoteModuleRepository` which is kinda too verbose). Lmk what you think

amckinney:
A new package identifier (e.g. `v1lalpha2`) makes sense, but I don't like it would introduce an incomplete `buf.alpha.regsitry.v1alpha2` package. In the current proposal, it looks like we intend the `buf.alpha.registry.v1alpha{1,2}` packages to coexist - we only refer to the `v1alpha2` packages for specific services, but we continue to use `buf.alpha.registry.v1alpha1` for the rest.

 From an API perspective, this is confusing - it's not obvious what set of versions the `buf` CLI and the BSR is interacting with. As things continue to evolve, the BSR might support a larger permutation like `buf.alpha.registry.v1alpha{1,2,3,...}` for different services.

I think I'd actually prefer this to be a new service in the same `buf.alpha.regsitry.v1alpha1` package with a `V2` identifier (e.g. `ModuleV2Service`). Then, once we've made more changes to some of the other services, we can do a more complete upgrade and move over all of the relevant services to `buf.alpha.registry.v1alpha2` as a complete set, and deprecate `buf.alpha.regsitry.v1alpha1` all at once. 
- At that point, we can also debate whether or not we should be more granular with these packages to prevent these issues from happening again later (e.g. `buf.alpha.registry.module.v1alpha1`, etc).

> Having two sets of services with different messages representing the same resources feels like it will get messy pretty fast.

I agree that it could get messy fast, but I think it's the lesser of the two evils while we have this package structure.

What are your thoughts?

amckinney:
I'm not a huge fan of the `RemoteModule` naming, but I think it's the best we can come up with - it's obvious what it represents, and it doesn't conflict with the `buf.alpha.module.v1alpha1.Module` naming.

The definitions themselves look good - I would just place this in the same `buf.alpha.registry.v1alpha1` package for the reasons I described above. I wanna make sure that @robbertvanginkel agrees though.

rubensf:
I'm not entirely convinced that `v1alpha1.ModuleV2Service` (or alternatives like `NewModuleService`, `RemoteModuleV2`) are more semantic - it sounds just slightly less confusing than leaving things in `v1alpha1`. There's also some argument that Buf consists of sub services, which can independently be updated to new versions.

After some other discussions, this is starting to seem like a general API versioning strategy:
- When would we pivot the a v2?
- Is this the general strategy we'd take for any future changes?
- How are we planning to maintain backwards compatibility?

rubensf:
Additionally - it seems we should at least wait until Tracks have a path forward, since there's a non-trivial (even if low) chance that it may affect how we do naming :)

amckinney:
> I'm not entirely convinced that v1alpha1.ModuleV2Service (or alternatives like NewModuleService, RemoteModuleV2) are more semantic - it sounds just slightly less confusing than leaving things in v1alpha1.

Now that I see the `RemoteModule` renaming that you have here, I agree. The `V2` component suggestion above is primarily meant to argue that we should preserve the same package. For example, if we had to make a new `UserService`, I'd want it to be `buf.alpha.registry.v1alpha1.UserV2Service` rather than intermixing `buf.alpha.registry.v1alpha2.UserService` with the rest of `buf.alpha.registry.v1alpha1`.

So to be clear, the only thing I'd change about what you have here is to move the `RemoteModule` API in `buf.alpha.registry.v1alpha1` and drop `v1alpha2` entirely.

> Additionally - it seems we should at least wait until Tracks have a path forward, since there's a non-trivial (even if low) chance that it may affect how we do naming :)

Sounds good to me 👍

bufdev:
We can probably close this now right?

rubensf:
Yeah, closing this.

For future context: TL;DR renaming Repository/Module/Commit to Module/Snapshot is messy (expected), but also very annoying to maintain backwards compatibility, particularly because of `buf.lock` in which `repository` and `commit` are keywords and in which we already committed to no breaking changes.

Issue #1086, opened 2022-04-26
buf lint stat error 


lmolas:
If you have an import proto path like this: /license/api.proto and you have a license file in another module, a stat error "not a directory" appears:
Failure: stat {omitted path}/license/api.proto: not a directory

In the code you only continue if you encounter "not exists" errors.
But "not a directory" error breaks.

bufdev:
For us to look into this, you must provide exact reproduction steps.

lmolas:
[buf-issue.tar.gz](https://github.com/bufbuild/buf/files/8571022/buf-issue.tar.gz)

As explained in the README, just run buf lint.

bufdev:
@amckinney 

The issue is here: https://github.com/bufbuild/buf/blob/14ed91784fcd5cd06763c420ce6d765092571430/private/buf/buffetch/internal/reader.go#L807 this is getting triggered if you have:

```
├── buf.work.yaml
├── module1
│   └── license
│       └── v1
│           ├── api.proto
│           └── core.proto
└── module2
    ├── dummy
    │   └── v1
    │       └── dummy.proto
    └── license
```

`module1/license` is a directory, but `module2/license` is not. If you rename `module2/license` to `module2/license2`, there is no issue. You might know right away what is going on here.

amckinney:
Thanks for the issue @lmolas! This will go out in the next release.

lmolas:
Thanks for the quick fix

PR #1092, opened 2022-04-27
Fix file type collision across multiple roots

amckinney:
Fixes https://github.com/bufbuild/buf/issues/1086

Multiple `storage.ReadBucket`s can be consolidated into one with `storage.MultiReadBucket`, which most frequently happens for multiple `build.roots` or `buf.work.yaml` directories. In this case, it's possible that one of the roots/directories defines a file with the same name, but with different modes (e.g. one is a regular file and the other is a directory).

For example, consider the following, where `license` is both a directory (for `module1`) and a regular file (for `module2`):

```
├── buf.work.yaml
├── module1
│   └── license
│       └── v1
│           ├── api.proto
│           └── core.proto
└── module2
    ├── dummy
    │   └── v1
    │       └── dummy.proto
    └── license
```

In this case, we'll successfully recognize that all of the files contained in the union are the following:

```
license/v1/api.proto
license/v1/core.proto
dummy/v1/dummy.proto
license
```

When we query the multi `storage.Bucket` for each of these paths, we must iterate through each of its delegates to determine which path belongs where. However, when we ask the `module2` directory for the `license/v1/api.proto` file, it'll recognize it as not a directory (because the `module2/license` file is a regular file).

If one of the delegates actually contains the path, then everything is fine. So we need to recognize if the path is a regular file, or one of its path components is a regular file (as illustrated above).

### DIFF
--- a/private/pkg/storage/errors.go
+++ b/private/pkg/storage/errors.go
@@ -16,7 +16,6 @@ package storage
 
 import (
 	"errors"
-	"fmt"
 	"strings"
 
 	"github.com/bufbuild/buf/private/pkg/normalpath"
@@ -30,8 +29,6 @@ var (
 
 	// errNotExist is the error returned if a path does not exist.
 	errNotExist = errors.New("does not exist")
-	// errInvalidPath is the error returned if a path is invalid.
-	errInvalidPath = errors.New("invalid path")
 )
 
 // NewErrNotExist returns a new error for a path not existing.
@@ -42,21 +39,6 @@ func NewErrNotExist(path string) error {
 // IsNotExist returns true for a error that is for a path not existing.
 func IsNotExist(err error) bool {
 	return errors.Is(err, errNotExist)
-}
-
-// NewErrInvalidPath returns a new error for an invalid path.
-func NewErrInvalidPath(path string) error {
-	return normalpath.NewError(path, errInvalidPath)
-}
-
-// NewErrInvalidPathf returns a new error for an invalid path.
-func NewErrInvalidPathf(path string, template string, args ...interface{}) error {
-	return normalpath.NewError(path, fmt.Errorf("%w: %s", errInvalidPath, fmt.Sprintf(template, args...)))
-}
-
-// IsInvalidPath returns true for a error that is for an invalid path.
-func IsInvalidPath(err error) bool {
-	return errors.Is(err, errInvalidPath)
 }
 
 // NewErrExistsMultipleLocations returns a new error if a path exists in multiple locations.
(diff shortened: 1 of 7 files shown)


private/pkg/storage/storageos/bucket.go
```diff
@@ -256,12 +256,43 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
...
+		// It's important that we detect these cases so that
+		// multi buckets don't unnecessarily fail when one of
+		// its delegates actually defines the path.
+		elements := strings.Split(normalpath.Normalize(externalPath), "/")
+		if len(elements) == 1 {
+			// The path is a single element, so there aren't
+			// any other files to check.
+			return err
+		}
+		for i := len(elements) - 1; i >= 0; i-- {
```
bufdev, line 278: There's no way we can do this I think - this is introducing an O(file depth) operation to basically every `storageos` Bucket operation.

private/pkg/storage/errors.go
```diff
@@ -41,6 +44,21 @@ func IsNotExist(err error) bool {
 	return errors.Is(err, errNotExist)
 }
 
+// NewErrInvalidPath returns a new error for an invalid path.
```
bufdev, line 47: What is an invalid path? I don't understand this from an API perspective, ie this doesn't make sense in the context of paths and buckets within the storage API.

bufdev (commented):
I don't understand at a conceptual level what an invalid path is, and the O(file depth) operation seems like a performance concern waiting to happen.

private/pkg/storage/storageos/bucket.go
```diff
@@ -256,12 +256,43 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
...
+				return storage.NewErrInvalidPath(path)
+			}
+		}
 		return err
 	}
 	if !fileInfo.Mode().IsRegular() {
 		// making this a user error as any access means this was generally requested
 		// by the user, since we only call the function for Walk on regular files
-		return fmt.Errorf("%q is not a regular file", path)
+		return storage.NewErrInvalidPathf(path, "not a regular file")
```
bufdev, line 295: It seems like this might result in a not exist error, not a new "invalid" error

private/pkg/storage/storageos/bucket.go
bufdev: Talked offline - this is just in the error case where we have a non-is-not-exist error, so not so bad

private/pkg/storage/errors.go
bufdev: Commented - just removing this and returning is-not-exist in the case of non-regular file. storageos Buckets don't iterate over non-regular files in Walk as it is, it seems consistent to treat this as a not-exist case.

bufdev (approved):

PR #1097, opened 2022-04-29
Adds the http2client and connectclient plugin

smaye81:
This adds the `http2client` and `protoc-gen-go-connectclient` plugin to the main Buf repo.  

Note that this is just a straight copy from `buf/core`.  Once this is committed to main, a subsequent PR will remove these from `buf/core` and instead reference them from this repository.


mfridman (approved):

akshayjshah (approved):
Looks good!

bufdev:
This needed a file ignore similar to other binaries, see `make/buf/all.mk`

PR #1101, opened 2022-05-02
Generate connect clients

smaye81:
`make generate` was run with the connectclient plugin, generating connect clients for eventual implementation in the Buf CLI.


smaye81:
I noticed there's a lint issue but I'm not entirely sure how to fix this one.  The `lint` target runs fine locally.  I'm sure I'm missing something, but just not sure what.

make/buf/all.mk
```diff
@@ -112,13 +114,20 @@ bufgenerateclean:: bufgeneratecleango
 bufgenerateprotogo:
 	$(BUF_BIN) generate proto --template data/template/buf.go.gen.yaml
 
+.PHONY: bufgenerateprotoconnectgo
+bufgenerateprotoconnectgo: $(PROTOC_GEN_CONNECT_GO)
+	$(BUF_BIN) generate $(BUF_PROTO_SOURCE) --template data/template/buf.connect-go.gen.yaml
+	$(BUF_BIN) generate proto --template data/template/buf.connect-go.gen.yaml
+	$(BUF_BIN) generate --template data/template/buf.connect-go.gen.yaml --path envoy/service/auth/v3 buf.build/envoyproxy/envoy
```
mfridman, line 121: This shouldn't be required for the CLI, its specific to the schema registry. We might need to update the template files a bit.

mfridman: Specifically referring to the envoy auth service.

make/go/dep_protoc_gen_connect_go.mk
```diff
@@ -0,0 +1,20 @@
+# Must be set
```
mfridman, line 1: I assume this can live in `make/go` without being removed/overwritten/etc. by makego .. but I'm not 100% sure.

data/template/buf.connect-go.gen.yaml
```diff
@@ -0,0 +1,13 @@
+# This template is to generate connect go.
+version: v1
+managed:
+  enabled: true
+  go_package_prefix:
+    default: github.com/bufbuild/buf/private/gen/proto/go
+    override:
```
mfridman, line 7: I don't think we need this `override` since we're in the repo with the generated code?

smaye81: Ah yeah good call.  Removed.

make/go/dep_protoc_gen_connect_go.mk
smaye81: Maybe I'll leave it here for now and then create a task / issue in Linear to track it?

mfridman (commented):
LGTM, now the fun begins.

mfridman (approved):

data/template/buf.connect-go.gen.yaml
```diff
@@ -0,0 +1,11 @@
+# This template is to generate connect go.
```
bufdev, line 1: Why did we need a separate file for this? This seems incorrect

make/go/dep_protoc_gen_connect_go.mk
bufdev: As with all other files in `make/go`, this should not be managed in repositories other than makego, as denoted by the file headers. This should be created in makego first, then copied over.

data/template/buf.connect-go.gen.yaml
smaye81: Should I move this to `data/template/buf.go-client.gen.yaml`?

Issue #1102, opened 2022-05-02
PACKAGE_SERVICE_NO_DELETE breaking change rule does not work


sarahegler:
When deleting the only service from a package and running breaking change detection using the `PACKAGE` rules, no breaking changes are detected as expected. This should result in: `[filename:line]:Previously present service [service_name] was deleted from package [package_name].`

Breaking changes are detected when _renaming_ the service with `PACKAGE` rules, but missed when the service is deleted altogether. This is not an issue under `FILE` rules; both deleting and renaming the service are detected with `FILE` rules.

Took a quick look and this appears to be because the `packageToNameToService` map made [here](https://github.com/bufbuild/buf/blob/847d45959ff3d442e99bcc25f02042cdf6a52de2/private/pkg/protosource/protosource.go#L827) is empty when there is no service in a particular package. So on the branch being tested this map is empty, which means the check [here](https://github.com/bufbuild/buf/blob/v0.53.0/internal/buf/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go#L920) in checkPackageServiceNoDelete is false and so we don't end up checking whether the service exists in the package.

doriable:
Thank you for reporting this issue and the very clear diagnosis! I was definitely able to reproduce this and I've put out a fix for review.

sarahegler:
thanks for jumping on this so quickly @doriable!

bufdev:
I think this is fixed now right?

doriable:
Yes, in that it's merged -- I originally intended on closing this on release, but closing now also works :)

PR #1105, opened 2022-05-04
Fix `PACKAGE_SERVICE_NO_DELETE` check

doriable:
This addresses #1102 by ensuring that we are creating an entry in the `PackageToNameToService` for every package, even if there are no services present, so we can check if a service is deleted out of a package that now no longer contains any services.

Test:

```
# BarService is the only service in package a

$ cat bar.proto
syntax = "proto3";

package a;

service BarService {
  rpc BarRPC(BarRequest) returns (BarResponse);
}

message BarRequest {
  string key = 1;
  int32 value = 2;
}

message BarResponse {
  string key = 1;
  int32 value = 2;
}
```

After deleting `BarService`, previously we returned no error. Now we are returning:

```
$ buf breaking --against first.bin
bar.proto:1:1:Previously present service "BarService" was deleted from package "a".
```


private/buf/cmd/buf/buf_test.go
```diff
@@ -482,6 +482,20 @@ func TestFailCheckBreaking5(t *testing.T) {
 	)
 }
 
+func TestFailCheckBreaking6(t *testing.T) {
```
bufdev, line 485: The test for this should be in https://github.com/bufbuild/buf/blob/main/private/bufpkg/bufcheck/bufbreaking/bufbreaking_test.go

bufdev (approved):

PR #1111, opened 2022-05-06
Generate new apiclientconnect stubs with new plugin

smaye81:
This makes a few changes to the existing `connectclient` plugin in preparation for integrating Connect into the Buf CLI.  Namely:

- renames plugin to `apiclientconnect` to be consistent with the grpc plugin `apiclientgrpc`
- adds additional generated code to the plugin output to allow for the creation of a `NewProvider` backed by Connect
- adds a function for creating a `NewConnectRegistryProvider` in the `bufapiclient`.  This is currently unused, but will be integrated when the CLI is switched to Connect as part of a followup PR.
- regenerated all stubs with new plugin functionality.

Once this lands, the next step is to add a function to create a `newConnectProvider`, which will replace the `newGRPCProvider` currently in place.  In addition, we will fix the error code inspections to ensure they are acting on Connect error code logic and not gRPC errors.

Note this does not affect any existing CLI transport quite yet.  This is just a first PR focused just on getting the generated clients and providers correct. 

Also note that once landed, this `Provider` approach will be reused inside `buf/core` also for consistency.


private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclientconnect/main.go
```diff
@@ -0,0 +1,444 @@
...
+
+	// Iterate over the services and create a constructor function for each
+	for _, service := range goPackageFileSet.Services() {
+		interfaceName := service.GoName
+		interfaceGoIdent := apiGoImportPath.Ident(interfaceName)
+		interfaceGoIdentString := g.QualifiedGoIdent(interfaceGoIdent)
+		interfaceConnectClientGoIdent := connectClientGoImportPath.Ident("New" + interfaceName + "Client")
+		interfaceConnectClientGoIdentString := g.QualifiedGoIdent(interfaceConnectClientGoIdent)
+
+		g.P(`func (p *provider) New`, interfaceName, `(ctx `, contextGoIdentString, `, baseURL string) (`, interfaceGoIdentString, `, error) {`)
```
smaye81, line 185: It seems that `context` isn't even needed here, but removing this doesn't make this quite a drop-in replacement as all the callsites will need changes to remove the context they are passing to `NewXXXXService`.  If this overall solution seems OK, though, I can remove it when the PR is submitted for final review.

data/template/buf.go-client.gen.yaml
```diff
@@ -40,5 +40,16 @@ plugins:
...
+      - named_go_package=connect=github.com/bufbuild/buf/private/gen/proto/connect
       - named_go_package=connectclient=github.com/bufbuild/buf/private/gen/proto/connectclient
+    strategy: all
+  - name: go-apiclientconnect
+    out: private/gen/proto/apiclientconnect
+    opt:
+      - paths=source_relative
+      - named_go_package=api=github.com/bufbuild/buf/private/gen/proto/api
       - named_go_package=connect=github.com/bufbuild/buf/private/gen/proto/connect
+      - named_go_package=connectclient=github.com/bufbuild/buf/private/gen/proto/connectclient
```
mfridman, line 52: Why does this generator need `connectclient` and `connect` ?

Going to come back to this PR and dive a bit more into the generated code, but figured I'd drop this comment to follow up.

smaye81:
> One outstanding issue with the connect provider is it doesn't attach the context items to the outgoing request.
> 
> Example, with gRPC this was done in `private/pkg/rpc/rpcgrpc/rpcgrpc.go` via `NewUnaryClientInterceptor` by calling [`metadata .NewOutgoingContext`](https://pkg.go.dev/google.golang.org/grpc/metadata#NewOutgoingContext).
> 
> Connect exposes the raw headers, so I think we should be able to do something similar with a client-side interceptor?
> 
> See example here:
> 
> https://github.com/bufbuild/connect-go/blob/9161c3d3927ce9f62c136b5f22e5eef18220a8e8/interceptor_example_test.go#L44-L49
> 
> The 2 main items we attached to the context in `NewContextModifierProvider` is the token and the cli version.

If I'm reading the code correctly, I _think_ this is being done already in the http2client?  

If you take a look at the `NewClient` function here:  https://github.com/bufbuild/buf/blob/main/private/pkg/transport/http2client/http2client.go#L45

it is using a new client interceptor, which is defined here:  

https://github.com/bufbuild/buf/blob/a955cdd7e9cb0173bf21694a7030cfc603d7721c/private/pkg/rpc/rpchttp/rpchttp.go#L49

The `RoundTrip` function in that file seems to be doing the same process with context, which is unwrapping the headers and adding them to the HTTP headers in the outgoing request.

https://github.com/bufbuild/buf/blob/a955cdd7e9cb0173bf21694a7030cfc603d7721c/private/pkg/rpc/rpchttp/rpchttp.go#L63

Let me know if I'm following this correctly.

private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclient/main.go
```diff
@@ -95,6 +95,27 @@ func handleGlobal(helper protogenutil.NamedHelper, plugin *protogen.Plugin, goPa
 	if err != nil {
 		return err
 	}
+	g.P(`const (`)
```
smaye81, line 98: I put the `ProviderScheme` code here in the `apiclient` plugin since I figured this was more global.  I can move it though if this isn't the best place.

private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclientconnect/main.go
```diff
@@ -0,0 +1,424 @@
...
+		newRequestGoIdentString := g.QualifiedGoIdent(newRequestGoIdent)
+
+		g.P(`type `, structName, `Client struct {`)
+		g.P(`logger *`, loggerGoIdentString)
+		g.P(`client `, clientGoIdentString)
+		g.P(`contextModifier func (`, contextGoIdentString, `) `, contextGoIdentString)
+		g.P(`}`)
+		g.P()
+
+		g.P(`func New`, interfaceName, `Client(`)
```
smaye81, line 239: By exposing this constructor function, we now have the `contextModifier` in the arg list.  This isn't used by the callers in `buf/core`, but they should be able to pass `nil` for this value since the generated code for the service files checks whether the `contextModifier` is nil before applying it to the context.

mfridman: I don't love it. But I think this is an acceptable trade-off that allows us to instantiate new connect clients that implement the `registryv1alpha1api` set of interfaces generated by protoc-gen-go-api.

This enables us to use single clients (which we do elsewhere) without the provider abstraction. 

The alternative is to unexport the constructors and be explicit about the clients we pass around. Although this does mean always having to instantiate a provider via `NewProvider`.

I'll leave this to your discretion.

bufdev: Just a drive-by comment, but throughout buf, constructor arguments are required, optional arguments are options...not sure if relevant here yet

private/gen/proto/connectclient/buf/alpha/registry/v1alpha1/registryv1alpha1connectclient/registryv1alpha1connectclient.pb.go
```diff
@@ -1,246 +0,0 @@
...
-// Code generated by protoc-gen-go-connectclient. DO NOT EDIT.
-
-package registryv1alpha1connectclient
-
-import (
-	registryv1alpha1api "github.com/bufbuild/buf/private/gen/proto/api/buf/alpha/registry/v1alpha1/registryv1alpha1api"
-	connect_go "github.com/bufbuild/connect-go"
-)
-
-func NewAdminServiceClient(
```
bufdev, line 24: Any reason we killed this file and moved the constructors?

private/gen/proto/apiclient/apiclient.pb.go
```diff
@@ -20,6 +20,27 @@ import (
...
 
+const (
+	// ProviderSchemeHTTP represents a provider scheme of http
+	ProviderSchemeHTTP ProviderScheme = iota + 1
+	// ProviderSchemeHTTPS represents a provider scheme of https
+	ProviderSchemeHTTPS
+)
+
+// ProviderScheme represents a scheme that will be used by all clients returned from a given provider
+type ProviderScheme int
```
bufdev, line 31: Why are we generating this?

I also wonder why we need it, probably worth hopping on a Zoom

private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclientconnect/main.go
bufdev: I think @mfridman and I figured out the point of confusion, tldr we should choose one of `Provider.New*` or `New*`

private/gen/proto/apiclient/apiclient.pb.go
bufdev: Talked through this, we should definitely kill this heh (or move it to a non-generated place), I think `WithAddressMapper` takes care of the concern though

private/buf/bufcli/bufcli.go
```diff
@@ -549,8 +549,7 @@ func NewConfig(container appflag.Container) (*bufapp.Config, error) {
 	return bufapp.NewConfig(container, externalConfig)
 }
 
-// NewRegistryProvider creates a new registryv1alpha1apiclient.Provider.
-func NewRegistryProvider(ctx context.Context, container appflag.Container) (registryv1alpha1apiclient.Provider, error) {
+func newGRPCRegistryProvider(ctx context.Context, container appflag.Container) (registryv1alpha1apiclient.Provider, error) {
```
smaye81, line 552: I removed it in this PR bc it was unused and wouldn't pass lint.  I didn't want to make it exported to pass so I will paste here instead:

When the Connect switch is officially turned on in bufcli, this function will be replaced by `newConnectRegistryProvider` like so:

```
func newConnectRegistryProvider(ctx context.Context, container appflag.Container) (registryv1alpha1apiclient.Provider, error) {
	client := http2client.NewClient(
		http2client.WithObservability(),
	)
	options := []bufapiclient.RegistryProviderOption{
		bufapiclient.RegistryProviderWithContextModifierProvider(NewContextModifierProvider(container)),
		bufapiclient.RegistryProviderWithAddressMapper(func(address string) string {
			if buftransport.IsAPISubdomainEnabled(container) {
				address = buftransport.PrependAPISubdomain(address)
			}
			return buftransport.PrependHTTPS(address)
		}),
	}
	return bufapiclient.NewConnectClientProvider(container.Logger(), client, options...)
}
```

private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclientconnect/main.go
```diff
@@ -0,0 +1,400 @@
...
+	// Import path for the connect named_go_package
+	connectGoImportPath, err := helper.NewPackageGoImportPath(
+		goPackageFileSet,
+		"connect",
+	)
+	if err != nil {
+		return err
+	}
+
+	// Iterate over the services and create a constructor function for each
```
mfridman, line 164: Nit, remove comment. We're not creating constructor functions anymore.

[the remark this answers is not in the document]
mfridman: Ignore this comment.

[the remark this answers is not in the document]
smaye81: Talked offline.  This comment should be ok.

mfridman (commented):
Let's update to the latest version of [connect-go](https://github.com/bufbuild/connect-go/commit/c35ee255c4f0265262da15a5f431d761a7498b93) and the plugin in:

make/go/dep_protoc_gen_connect_go.mk

I believe there were some changes upstream, and after this lands we're going to have to update core anyways.

smaye81:
> Let's update to the latest version of [connect-go@c35ee25](https://github.com/bufbuild/connect-go/commit/c35ee255c4f0265262da15a5f431d761a7498b93) and the plugin in:
> 
> make/go/dep_protoc_gen_connect_go.mk
> 
> I believe there were some changes upstream, and after this lands we're going to have to update core anyways.

Added this file to `makego` (with updated SHA) via https://github.com/bufbuild/makego/pull/23 and ran `make copyfrommakego` to pull it in here.

mfridman (approved):
lgtm.

data/template/buf.go-client.gen.yaml
```diff
@@ -35,10 +35,13 @@ plugins:
...
+    out: private/gen/proto/apiclientconnect
     opt:
       - paths=source_relative
       - named_go_package=api=github.com/bufbuild/buf/private/gen/proto/api
-      - named_go_package=connectclient=github.com/bufbuild/buf/private/gen/proto/connectclient
       - named_go_package=connect=github.com/bufbuild/buf/private/gen/proto/connect
+      - named_go_package=connectclient=github.com/bufbuild/buf/private/gen/proto/connectclient
+      - named_go_package=apiclient=github.com/bufbuild/buf/private/gen/proto/apiclient
+      - named_go_package=apiclientconnect=github.com/bufbuild/buf/private/gen/proto/apiclientconnect
+    strategy: all
```
bufdev, line 47: do we have `strategy: all` set for `protoc-gen-apiclientgrpc`? I think so but just checking

private/buf/bufcli/bufcli.go
bufdev, line 552: Public functions always go above private, ie move this down the file to below the public functions

private/bufpkg/bufapiclient/bufapiclient.go
```diff
@@ -67,6 +87,14 @@ func RegistryProviderWithAddressMapper(addressMapper func(string) string) Regist
 	}
 }
 
+// RegistryProviderWithScheme returns a new RegistryProviderOption that adds the given scheme to the transport address
+// for all clients returned by a provider
+func RegistryProviderWithScheme(scheme string) RegistryProviderOption {
```
bufdev, line 92: Is this still needed now that you have `PrependHTTPS`?

data/template/buf.go-client.gen.yaml
smaye81: Yep, we do.

private/buf/bufcli/bufcli.go
smaye81: Got it.  Moved the private ones down in the file.

private/bufpkg/bufapiclient/bufapiclient.go
smaye81: Ugh no.  Removed along with the other scheme options.

bufdev (approved):

Issue #1123, opened 2022-05-11
Provide a way to determine if generated files are up to date


stabai:
Since keeping generated code up to date is very important, it's great to have CI checks that ensure this. If there were a command like `buf generate --check` that would return an error if any generated code is stale, that would be incredibly helpful.

Ideally, this would run exactly the same way that `buf generate` does. But every time a generated file would be written, a check is performed instead:

- If the output file does not exist, return an error
- If the output file's contents do not match the generated content, return an error
- Otherwise, succeed

In none of these cases would any files be written. Showing the stale files or the differences between them could be useful, but more likely all that matters is that there are stale files at all.

bufdev:
It's a fair ask, but this is something we're unlikely to do:

- It's a gray area, but it's not very unix-y. If you want to determine if something changed, you can do something along the lines of https://github.com/bufbuild/buf/blob/main/make/go/scripts/checknodiffgenerated.bash via `bash checknodiffgenerated.bash buf generate`. It's not perfect (and you might note that this script isn't perfect either, it only goes down to the granularity of if a file changed or not), but it is something you can do.
- Regardless, and candidly, this would be a lot of work that we just can't budget for, and we're about to have a push where we're going to close issues we won't get to in the foreseeable future (~1yr at least).

stabai:
Makes sense, thanks for the details and candor!

I didn't know about that script, which is a major help. In the meantime, I wrote a script for this that I think is mildly better because it does the work in a temp dir and compares that to the working directory (so there's no reliance on git or pollution of the workspace).
https://gist.github.com/stabai/316dc7c2d097b39ad66dc92cc31f6ff9

It would need a bit of rework to be fully generic, but I'm happy to contribute to the scripts dir if helpful.

PR #1124, opened 2022-05-11
Cache ReadWriteBuckets based on absolute path in appprotoos.ResponseWriter

bufdev:
While investigating the feasibility of https://github.com/bufbuild/buf/issues/1123, I realized that our caching would not work for absolute vs relative paths. This should fix it, but there is a test failure.

```
--- FAIL: TestInsertionPointMixedPathsFail (0.29s)
    appcmdtesting.go:147:
        	Error Trace:	appcmdtesting.go:147
        	           				protoc_test.go:249
        	           				protoc_test.go:221
        	Error:      	Not equal:
        	           	expected: 1
        	           	actual  : 0
        	Test:       	TestInsertionPointMixedPathsFail
        	Messages:   	
```

From the comments on the test, I can't figure out what is going on and why this should be failing. @amckinney can you take a quick look?


amckinney:
So that failure is actually valid from a `protoc` perspective - I discovered this back when we were refactoring all of the insertion point stuff in [this](https://github.com/bufbuild/buf/pull/621) PR. I left a comment on the test helper:

```go
// testInsertionPointMixedPathsFail demonstrates that insertion points are only
// able to generate to the same output directory, even if the absolute path points
// to the same place.
func testInsertionPointMixedPathsFail(t *testing.T, runner command.Runner, receiverOut string, writerOut string) {
  ...
}
```

So the test above SHOULD be failing because this doesn’t actually work:

```sh
$ protoc -I proto proto/buf/alpha/image/v1/image.proto --insertion-point-receiver_out=insertion --insertion-point-writer_out=$(pwd)/insertion
```

With that said, we could just remove this test so that `buf alpha protoc` actually supports absolute paths (even though `protoc` doesn't). In that case, we'd actually support _more_ features than `protoc`, so everything else would still be backwards compatible.

I'll follow-up here by just refactoring the test verifying that mixing the path format (absolute vs. relative) works, then leave another comment justifying it.

bufdev:
LGTM

PR #1129, opened 2022-05-16
Replace transport of CLI from gRPC to Connect

smaye81:


private/buf/bufcli/errors.go
```diff
@@ -164,21 +164,30 @@ func NewTemplateNotFoundError(owner string, name string) error {
 // Note that this function will wrap the error so that the underlying error
 // can be recovered via 'errors.Is'.
 func wrapError(err error) error {
-	if err == nil || (err.Error() == "" && !rpc.IsError(err)) {
-		// If the error is nil or empty and not an rpc error, we return it as-is.
+	if err == nil || (err.Error() == "" && !isConnectError(err)) {
```
mfridman, line 167: Something to note about Connect error behavior is `.Error()` prints the code followed by the message.

https://github.com/bufbuild/connect-go/blob/main/error.go#L72-L78

Throughout this PR we want to make sure we're not composing errors with a stutter. I don't think that'll be the case because iirc the errors are hand-crafted.

Also, if you need to get at the underlying error can use `.Message()`:

https://github.com/bufbuild/connect-go/blob/main/error.go#L84-L89

mfridman:
To confirm, we'll land this PR swapping out the underlying clients and then we'll do a final sweep to rid of the gRPC-specific code? E.g.,

private/pkg/transport/grpc/grpcclient
private/pkg/rpc/rpcgrpc
private/pkg/rpc (not all, but a lot can be removed here)
etc.

private/bufpkg/bufapiclient/bufapiclient.go
```diff
@@ -12,45 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Package bufapiclient provides client-side gRPC constructs.
+// Package bufapiclient provides client-side Connect constructs.
```
mfridman, line 15: Random thought, maybe we could combine `bufapiclient` and `buftransport` so it's easier for others to discover the mapper functions.

bufdev: How are they used in core?

private/buf/bufcli/errors.go
smaye81: In testing, it turns out there were cases where the displayed error would have stuttered such as `Failure: FailedPrecondition: track main has no commits`.

I modified the `wrapError` function here that hopefully fixes this issue.  Now it attempts to convert the given error to a `connect.Error` and then reads the code and message from that.

I think the `if` statement should still catch the same conditions that were previously being caught.  If `err == nil` then the `asConnectError` function will return `nil, false`.  The empty condition should be getting caught via the `asConnectError` function also.

Let me know if I'm missing something though.

private/buf/bufcli/errors.go
```diff
@@ -164,19 +169,29 @@ func NewTemplateNotFoundError(owner string, name string) error {
...
 	switch {
-	case rpcCode == rpc.ErrorCodeUnauthenticated, isEmptyUnknownError(err):
+	case connectCode == connect.CodeUnauthenticated, isEmptyUnknownError(err):
 		return errors.New(`Failure: you are not authenticated. Create a new entry in your netrc, using a Buf API Key as the password. For details, visit https://docs.buf.build/bsr/authentication`)
-	case rpcCode == rpc.ErrorCodeUnavailable:
+	case connectCode == connect.CodeUnavailable:
 		return fmt.Errorf(`Failure: the server hosted at that remote is unavailable: %w`, err)
 	}
-	return fmt.Errorf("Failure: %w", err)
+	return fmt.Errorf("Failure: %s", connectMessage)
```
mfridman, line 187: Although this could have happened before if `err.Error() == "" && rpc.IsError(err)`, should we consider an explicit check for `connectMessage == ""` and output `Failure: unknown error` instead of `Failure: ` .. technically this should never happen.

There is a `isEmptyUnknownError` function, but it only catches empty strings on "Unknown" errors.

private/buf/bufcli/errors.go
```diff
@@ -164,19 +169,29 @@ func NewTemplateNotFoundError(owner string, name string) error {
...
+}
+
+// asConnectError uses errors.As to unwrap any error and look for a connect *Error.
+func asConnectError(err error) (*connect.Error, bool) {
```
mfridman, line 191: I'd love to see this helper as part of the core connect library, something we've floated in the past:

```
CheckAs(err) (*connect.Error, bool)
```

It's easy enough to write (as you did), and in core we inline it:

```
if connectErr := (&connect.Error{}); errors.As(err, &connectErr) {
    // handle connect error
}
```

Having one way to do this as defined by the connect library would be ideal. Nbd, but a nicer developer experience I think.

smaye81: There actually already is a function like this in the library, it's just private.  I'm definitely in favor of making it exported:

https://github.com/bufbuild/connect-go/blob/main/error.go#L134

private/buf/bufcli/errors.go
smaye81: Actually this can happen if you do something like `buf beta registry repository list buf.build/sayers` (i.e. try to list repositories using an address that already has a repository in it).

In this case, Connect seems to return `Unimplemented` with no message, so your `Failure: ` scenario will occur.  I'll think of a better way to handle that.

private/buf/bufcli/errors.go
akshayjshah: It used to be exported! I have no objection to exporting it again if it's useful.

private/bufpkg/bufapiclient/bufapiclient.go
mfridman: Oops, ignore this comment. `apiclient` is generated so it doesn't make sense.

private/buf/bufcli/errors.go
smaye81: For posterity, this is now being handled in the CLI itself by validating the subset of commands that do not expect subdirectories.  Connect is behaving as it should with its returned codes.

private/buf/cmd/buf/command/alpha/registry/token/tokendelete/tokendelete.go
```diff
@@ -85,6 +86,10 @@ func run(
 	if remote == "" {
 		return appcmd.NewInvalidArgumentError("you must specify a remote module")
 	}
+	parts := strings.Split(remote, "/")
+	if len(parts) > 1 {
+		return appcmd.NewInvalidArgumentError("remote address cannot contain subdirectories")
```
smaye81, line 91: Let me know if there is better wording for this error.

private/buf/bufcli/errors.go
```diff
@@ -164,21 +166,42 @@ func NewTemplateNotFoundError(owner string, name string) error {
...
+	// Attempt to convert err to a Connect error.  If it can be converted, then interpret it and return an intuitive error
+	if connectError := (&connect.Error{}); errors.As(err, &connectError) {
+		connectCode := connectError.Code()
+		connectMessage := connectError.Message()
+		switch {
+		case connectCode == connect.CodeUnauthenticated, isEmptyUnknownError(err):
+			return errors.New(`Failure: you are not authenticated. Create a new entry in your netrc, using a Buf API Key as the password. For details, visit https://docs.buf.build/bsr/authentication`)
+		case connectCode == connect.CodeUnavailable:
+			msg := `Failure: the server hosted at that remote is unavailable.`
+
```
mfridman, line 183: Nit.

1. Tighten up the whitespace.
2. Use `connectError.Message()` directly, no need to assign it at the beginning of the switch

private/buf/bufcli/errors.go
```diff
@@ -164,21 +166,42 @@ func NewTemplateNotFoundError(owner string, name string) error {
...
+		}
+		return fmt.Errorf("Failure: %s", connectMessage)
 	}
+
+	// Error was not empty, but was not a Connect error
 	return fmt.Errorf("Failure: %w", err)
 }
 
+// isConnectError returns true if the given error can be interpreted as a Connect error
+func isConnectError(err error) bool {
```
mfridman, line 200: Can we get rid of this in favor of `connect.AsError` 😄  
Added in https://github.com/bufbuild/connect-go/pull/215

private/buf/cmd/buf/command/alpha/registry/token/tokenget/tokenget.go
```diff
@@ -87,6 +88,10 @@ func run(
 	if remote == "" {
 		return appcmd.NewInvalidArgumentError("you must specify a remote module")
 	}
+	parts := strings.Split(remote, "/")
+	if len(parts) > 1 {
+		return appcmd.NewInvalidArgumentError("remote address cannot contain subdirectories")
```
mfridman, line 93: Maybe just me, but I prefer the term path over subdirectory.

`invalid remote address, must not contain any paths`

If you wanted to get fancy can parse the path and do something like:

`invalid remote address, must not contain any paths. Try removing "/foo/bar" from the address.`

private/buf/bufcli/errors.go
```diff
@@ -164,21 +166,42 @@ func NewTemplateNotFoundError(owner string, name string) error {
...
+			if dnsError := (&net.DNSError{}); errors.As(err, &dnsError) && dnsError.IsNotFound {
+				return fmt.Errorf(`%s Are you sure "%s" is a valid remote address?`, msg, buftransport.TrimAPISubdomain(dnsError.Name))
+			}
+
+			return fmt.Errorf(msg)
+		}
+		return fmt.Errorf("Failure: %s", connectMessage)
 	}
+
+	// Error was not empty, but was not a Connect error
```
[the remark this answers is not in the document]
smaye81: I don't think that would happen.  Because the first if statement will return if the error is empty, but not a Connect error

```
if err == nil || (err.Error() == "" && !isConnectError(err)) {
```

If it makes it past that and the error is empty, we know it must be a Connect error then, which means it will enter the second if statement:

```
if connectError := (&connect.Error{}); errors.As(err, &connectError) {
```

and return from there.

private/buf/cmd/buf/command/alpha/registry/token/tokenget/tokenget.go
smaye81: Challenge accepted.  Added a validation file to hold this too.

private/pkg/app/appcmd/validate.go
```diff
@@ -0,0 +1,40 @@
...
+package appcmd
+
+import (
+	"fmt"
+	"strings"
+)
+
+func ValidateRemoteNotEmpty(remote string) error {
+	if remote == "" {
+		return NewInvalidArgumentError("you must specify a remote module")
```
mfridman, line 24: I think this is referring to a remote address, not a remote module?

smaye81: It seems that way yeah, but the previous code all used this wording so I went with it.  Not sure which is correct.

private/pkg/app/appcmd/validate.go
```diff
@@ -0,0 +1,37 @@
...
+
+package appcmd
+
+import (
+	"fmt"
+	"strings"
+)
+
+// ValidateRemoteNotEmpty validates that the given remote address is not an empty string
+func ValidateRemoteNotEmpty(remote string) error {
```
bufdev, line 23: This should not be in appcmd - anything in `pkg` is non-buf-specific, this is something specific to buf. This should be next to other validation functions in `buf` or `bufpkg`.

private/bufpkg/buftransport/buftransport.go
```diff
@@ -39,6 +39,11 @@ func SetDisableAPISubdomain(env map[string]string) {
 	env[disableAPISubdomainEnvKey] = "disable"
 }
 
+// TrimAPISubdomain removes the API subdomain from the given address (if present).
+func TrimAPISubdomain(address string) string {
```
bufdev, line 43: We just use this in an error message, do we want to have a helper function that has this behavior? This made me wonder why we want to trim a subdomain, which worried me.

bufdev (changes requested):

private/bufpkg/buftransport/buftransport.go
smaye81: Yeah the only reason it's there is because this file has the subdomain.  I could just move this to the errors.go file and export the `apiSubdomain`

smaye81: Ok done.  Let me know if this seems better.

private/pkg/app/appcmd/validate.go
smaye81: Moved to `private/bufpkg/bufmodule/bufmoduleref/validate.go`

private/bufpkg/buftransport/buftransport.go
```diff
@@ -21,7 +21,7 @@ import (
 )
 
 const (
-	apiSubdomain = "api"
+	APISubdomain = "api"
```
bufdev, line 24: needs godoc

bufdev (approved):

mfridman (approved):
Great job @smaye81.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 - Fix check for `PACKAGE_SERVICE_NO_DELETE` to detect deleted services.
 - Remove `buf beta registry track`.
 - Remove `buf beta registry branch`.
+- Switch RPC transport
```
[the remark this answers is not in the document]
smaye81: OK I will remove it

bufdev:
Last thing (and probably not part of this PR) - moving to connect should result in `pkg/rpc` being deleted, we still have these references though, let's make sure we clean them up:

```
$ agg pkg\\\/rpc
private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go:29:	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
private/buf/bufcli/bufcli.go:56:	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
private/bufpkg/bufrpc/bufrpc.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
private/pkg/transport/grpc/grpcclient/client_conn_provider.go:24:	"github.com/bufbuild/buf/private/pkg/rpc/rpcgrpc"
private/pkg/transport/http2client/http2client.go:23:	"github.com/bufbuild/buf/private/pkg/rpc/rpchttp"
private/pkg/transport/http/httpclient/client.go:23:	"github.com/bufbuild/buf/private/pkg/rpc/rpchttp"
private/pkg/rpc/rpchttp/rpchttp.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
private/pkg/rpc/rpchttp/rpchttp.go:22:	"github.com/bufbuild/buf/private/pkg/rpc/rpcheader"
private/pkg/rpc/rpcgrpc/rpcgrpc.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
private/pkg/rpc/rpcgrpc/rpcgrpc.go:22:	"github.com/bufbuild/buf/private/pkg/rpc/rpcheader"
private/pkg/rpc/rpcauth/rpcauth.go:23:	"github.com/bufbuild/buf/private/pkg/rpc"
```

akshayjshah:
Yep. Plan is to clean that package up in a post-migration pass.

smaye81:
> Last thing (and probably not part of this PR) - moving to connect should result in `pkg/rpc` being deleted, we still have these references though, let's make sure we clean them up:
> 
> ```
> $ agg pkg\\\/rpc
> private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go:29:	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
> private/buf/bufcli/bufcli.go:56:	"github.com/bufbuild/buf/private/pkg/rpc/rpcauth"
> private/bufpkg/bufrpc/bufrpc.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
> private/pkg/transport/grpc/grpcclient/client_conn_provider.go:24:	"github.com/bufbuild/buf/private/pkg/rpc/rpcgrpc"
> private/pkg/transport/http2client/http2client.go:23:	"github.com/bufbuild/buf/private/pkg/rpc/rpchttp"
> private/pkg/transport/http/httpclient/client.go:23:	"github.com/bufbuild/buf/private/pkg/rpc/rpchttp"
> private/pkg/rpc/rpchttp/rpchttp.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
> private/pkg/rpc/rpchttp/rpchttp.go:22:	"github.com/bufbuild/buf/private/pkg/rpc/rpcheader"
> private/pkg/rpc/rpcgrpc/rpcgrpc.go:21:	"github.com/bufbuild/buf/private/pkg/rpc"
> private/pkg/rpc/rpcgrpc/rpcgrpc.go:22:	"github.com/bufbuild/buf/private/pkg/rpc/rpcheader"
> private/pkg/rpc/rpcauth/rpcauth.go:23:	"github.com/bufbuild/buf/private/pkg/rpc"
> ```

There's still a bit of code in `pkg/rpc` that is being used even after the migration to Connect.  Granted, a good bit can be deleted, but a few that I see that are still in use:

`private/pkg/rpc/rpcauth` - Used in many places in both `buf` and `core`.
`private/pkg/rpc/rpchttp` - Client interceptors used in `buf` and the Server interceptor used in `core`.
`private/pkg/rpc/headers.go` - The header functions here are used in both `buf` and `core`.

I'm sure the above can mostly eventually go away, but I think it will take more than just cleanup to remove them.  Probably going to have to refactor a few things to obviate the need for them.

bufdev:
- The interceptors in https://github.com/bufbuild/buf/blob/main/private/pkg/rpc/rpchttp/rpchttp.go were meant to convert from HTTP headers to `rpc` package headers, which were put on context. Since we no longer use `rpc` package header mechanisms (since we have connect), this is duplicating the headers across the connect location and the `rpc` location. Therefore these interceptors should not be needed anymore (if they are, that's a bug).
- https://github.com/bufbuild/buf/blob/main/private/pkg/rpc/rpcauth/rpcauth.go deals with putting headers onto the context using the `rpc` package header mechanism. Similarly, this is overlapping with connect, and we should be using the connect method for headers. I think that this continues to work is a weird bug actually - we happen to be putting headers onto the HTTP protocol in the same manner that connect does (I think).
- The above two comments effectively address https://github.com/bufbuild/buf/blob/main/private/pkg/rpc/headers.go

mfridman:
Agreed, we definitely need to streamline the header/context propagation you mentioned.

What are your thoughts we tackle this immediately after the chunk of work to swap grpc protocol with connect? The rationale is we want to give ourselves as much time with the connect protocol in production.

Issue #1134, opened 2022-05-17
segfault in `buf export`


lukebond:
I'm trying to run `buf export -o path` for a simple use case and hitting a segfault. I'm new to `buf` so perhaps I'm not using it right (in which case this should highlight a missing `nil` check and a nice error message).

Steps to reproduce:
1. Create the following `buf.yaml`:
  ```
  version: v1
  breaking:
    use:
      - FILE
  lint:
    use:
      - DEFAULT
  deps:
    - buf.build/grpc/grpc
  ```
3. Run `buf mod update`
4. Run `buf export -o vendor_protobuf/`

You should see the following segfault:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x55fc54a440c4]

goroutine 1 [running]:
github.com/bufbuild/buf/private/buf/cmd/buf/command/export.run({0x55fc55060290, 0xc0006e4c90}, {0x55fc55066b28?, 0xc00034d7c0}, 0xc0002462a0)
	github.com/bufbuild/buf/private/buf/cmd/buf/command/export/export.go:329 +0xd44
github.com/bufbuild/buf/private/buf/cmd/buf/command/export.NewCommand.func1({0x55fc55060290?, 0xc0006e4c90?}, {0x55fc55066b28?, 0xc00034d7c0?})
	github.com/bufbuild/buf/private/buf/cmd/buf/command/export/export.go:63 +0x31
github.com/bufbuild/buf/private/buf/bufcli.NewErrorInterceptor.func1.1({0x55fc55060290?, 0xc0006e4c90?}, {0x55fc55066b28?, 0xc00034d7c0?})
	github.com/bufbuild/buf/private/buf/bufcli/errors.go:89 +0x31
github.com/bufbuild/buf/private/pkg/app/appflag.(*builder).run(0xc000420680, {0x55fc550601e8, 0xc00034c000}, {0x55fc550643c0, 0xc0000b20f0}, 0xc000233110)
	github.com/bufbuild/buf/private/pkg/app/appflag/builder.go:142 +0x524
github.com/bufbuild/buf/private/pkg/app/appflag.(*builder).NewRunFunc.func1({0x55fc550601e8, 0xc00034c000}, {0x55fc550643c0, 0xc0000b20f0})
	github.com/bufbuild/buf/private/pkg/app/appflag/builder.go:97 +0x7f
[...]

PR #1135, opened 2022-05-17
Check if mergedImage is nil

bufdev:
Should fix https://github.com/bufbuild/buf/issues/1134.

This needs review as I'm not sure on the code.


doriable (approved):

bufdev:
Is this the correct fix? Ie we should continue if `mergedImage == nil`, not `mergedImage != nil`?

doriable:
> Is this the correct fix? Ie we should continue if `mergedImage == nil`, not `mergedImage != nil`?

We would continue if `mergedImage == nil` or if there is an image but the path doesn't exist in said image.

PR #1136, opened 2022-05-17
Rename bufplugin package to bufremoteplugin

bufdev:
This is just code reorganization - we may want to use `bufplugin` for something else soon, and this more accurately conveys what this package is covering.


bufdev:
I'll follow up with the core change right after.

SyrieBianco (approved):
:blindstamp: 🙈

PR #1137, opened 2022-05-18
add buf beta studio-agent command with request logger

cyinma:
fixes TCN-81 TCN-82

an example usage to test the agent:

execute
```
buf beta studio-agent "buf.build" --port=9999
```
to run the agent server,

then execute
```
curl -d "$(echo '{
    "target":"https://api.buf.build:443/buf.alpha.registry.v1alpha1.SearchService/Search",
    "headers":[
        {
            "key":"Content-Type",
            "value":["application/grpc"]
        }
    ],
    "body":"'$(echo '{
        "query": "buf",
        "pageSize": 10,
        "pageToken": 0,
        "filters": []
    }' | buf beta convert --type buf.build/bufbuild/buf#buf.alpha.registry.v1alpha1.SearchRequest --input -#format=json | base64)'"
}' | buf beta convert --type buf.build/tommyma/buf#buf.alpha.studio.v1alpha1.InvokeRequest --input -#format=json | base64)" \
 -H "Content-Type: text/plain" -s localhost:9999 | base64 -d | \
 buf beta convert --type buf.build/tommyma/buf#buf.alpha.studio.v1alpha1.InvokeResponse | jq -r .body | base64 -d | \
 buf beta convert --type buf.build/bufbuild/buf#buf.alpha.registry.v1alpha1.SearchResponse | jq
```
to call the search endpoint via the agent


private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+func newFlags() *flags {
+	return &flags{}
+}
+
+func (f *flags) Bind(flagSet *pflag.FlagSet) {
+	flagSet.StringVar(
+		&f.Port,
+		portFlagName,
+		"8080",
+		"The port to be exposed to accept HTTP request.",
```
bufdev, line 77: `s/request/requests/`

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+		&f.DisallowedHeaders,
+		disallowedHeadersFlagName,
+		nil,
+		`The headers to be disallowed via the agent to the target server. Multiple headers are appended if specified multiple times.`,
+	)
+	flagSet.StringSliceVar(
+		&f.ForwardHeaders,
+		forwardHeadersFlagName,
+		nil,
+		`The headers to be forwarded via the agent to the target server. Must be a colon-separated key-value pair (like --forward-header=fromHeader1:toHeader1). Multiple header pairs are appended if specified multiple times.`,
```
bufdev, line 89: key/values are typically separated by `=` instead of `:`, and there's also a function to use to do this https://pkg.go.dev/github.com/spf13/pflag#StringToStringVar

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+	ctx context.Context,
+	container appflag.Container,
+	flags *flags,
+) error {
+	logger := container.Logger().Named("studio-agent")
+	// convert the disallowedHeaders from a list of string to a map
+	disallowedHeaders := make(map[string]struct{}, len(flags.DisallowedHeaders))
+	for _, header := range flags.DisallowedHeaders {
+		disallowedHeaders[header] = struct{}{}
+	}
```
robbertvanginkel, line 102: https://github.com/bufbuild/buf/blob/638c089e1e61efa25fe0fdf8be86d08da97c7594/private/pkg/stringutil/stringutil.go#L75

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+		`The headers to be forwarded via the agent to the target server. Must be a colon-separated key-value pair (like --forward-header=fromHeader1:toHeader1). Multiple header pairs are appended if specified multiple times.`,
+	)
+}
+
+func run(
+	ctx context.Context,
+	container appflag.Container,
+	flags *flags,
+) error {
+	logger := container.Logger().Named("studio-agent")
```
robbertvanginkel, line 97: Can you add some context on how this logger works in a comment here? Does this logger only show things when buf is called with `--debug`, is it line oriented, can it do json logging?

Normally the cli logger is meant for human output, but the output of this command should also support a more machine readable log sink.

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+		forwardHeaders[s[0]] = s[1]
+	}
+	config, err := bufcli.NewConfig(container)
+	if err != nil {
+		return err
+	}
+	mux := bufstudioagent.NewHandler(
+		logger,
+		container.Arg(0), // the origin from command argument
+		config.TLS,
```
robbertvanginkel, line 110: Where does this TLS come from, is this some global config option you can set over the whole binary?

I imagine this was something to configure the TLS for outgoing calls made by the cli towards a bsr, which we could re-use here if we document it well. Besides the outgoing tls config, I think we also need to let users set a server TLS config to be used by the `listenAndServe{TLS}` used in this subcommand.

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+		disallowedHeaders,
+		forwardHeaders,
+	)
+	server := http.Server{
```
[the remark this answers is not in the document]
robbertvanginkel: We could move the runner from core to here, but it will pull in more dependencies that aren't used in the cli like chi/`x/sync`/opencensus. We can probably factor out the opencensus part quite easily, but chi is pervasive in the default middlewares added by the package.

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,165 @@
...
+	if err := logger.Sync(); err != nil {
+		return err
+	}
+	return server.Shutdown(ctx)
+}
+
+func requestLoggingHandler(mux http.Handler, logger *zap.Logger) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
+		bodyBytes, err := ioutil.ReadAll(http.MaxBytesReader(w, r.Body, bufstudioagent.MaxMessageSizeBytesDefault))
```
robbertvanginkel, line 139: We shouldn't retain+log the whole body here, up to 5mb log lines wouldn't exactly be a summary. The bodies are also base64 encoded binary proto so not exactly user friendly. Depending on https://github.com/bufbuild/buf/pull/1137/files#r881775141, we may get some basic functionality to logger embedded there. 

However, logging at this level only gives us insight into the envelope whereas I think the log would be more interesting if the target url, upstream response (grpc) status would be in there somewhere. I'm not sure if there's an easy way to attach details to logs made in a wrapping handler, lets chat about this offline.

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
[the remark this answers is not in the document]
cyinma: there is no reason for that, do you think we should move that `httpserver` package to here (`bufbuild/buf`) and use it in both side?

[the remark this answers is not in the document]
cyinma: refactored to use a similar way to build the server, and will have a follow up PR to move part of the `httpserver` package to `bufbuild/buf` to use it in this command

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
cyinma: resolve as requesting logging is removed from the PR, however the logger do support json logging with a global `--log-format` flag

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
cyinma: added flags for ca cert, client cert&key and server cert&key file, instead of the global TLS config

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
cyinma: resolve as requesting logging is removed from the PR

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,241 @@
...
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <origin>",
+		Short: "Run an HTTP(S) server as the studio agent with the origin be set as the allowed origin for CORS options.",
```
doriable, line 67: I think we should capitalise `Studio` (both here and in our documentation).

nit: "Run an HTTP(S) server as the Studio agent with the origin set as the allowed origin for CORS options."

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,241 @@
...
+		&f.Port,
+		portFlagName,
+		"8080",
+		"The port to be exposed to accept HTTP requests.",
+	)
+	flagSet.StringSliceVar(
+		&f.DisallowedHeaders,
+		disallowedHeadersFlagName,
+		nil,
+		`The headers to be disallowed via the agent to the target server. Multiple headers are appended if specified multiple times.`,
```
doriable, line 105: nit: "The headers disallowed via the agent to the target server. Multiple headers are appended if specified multiple times."

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,241 @@
...
+	)
+	flagSet.StringToStringVar(
+		&f.ForwardHeaders,
+		forwardHeadersFlagName,
+		nil,
+		`The headers to be forwarded via the agent to the target server. Must be an equals sign separated key-value pair (like --forward-header=fromHeader1=toHeader1). Multiple header pairs are appended if specified multiple times.`,
```
doriable, line 111: "(e.g. --forward-header=fromHeader1=toHeader1)."

doriable (approved):
Some small nits, but overall looks good to me :)

doriable (commented):
Oh, also, this needs a changelog entry!

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
cyinma: discussed offline, will keep this as is

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,241 @@
...
+			return httpServer.Shutdown(ctx)
+		},
+	}
+	if err := thread.Parallelize(ctx, jobs); err != http.ErrServerClosed {
+		return err
+	}
+	return nil
+}
+
+func newCACertPool(caCertFile string) (*x509.CertPool, error) {
```
robbertvanginkel, line 214: Lets refactor this a bit to use our `certclient` package, export:

https://github.com/bufbuild/buf/blob/6092bfe2dfa0052644676c4397943133bdb94909/private/pkg/cert/certclient/util.go#L24-L25

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,233 @@
...
+			IdleTimeout: defaultIdleTimeout,
+		})
+	}
+	httpServer := &http.Server{
+		Handler:           mux,
+		ReadHeaderTimeout: defaultReadHeaderTimeout,
+		TLSConfig:         serverTLSConfig,
+	}
+	var httpListenConfig net.ListenConfig
+	httpListener, err := httpListenConfig.Listen(ctx, "tcp", fmt.Sprintf("0.0.0.0:%s", flags.Port))
```
robbertvanginkel, line 192: I'm thinking we may want an option to let users specify `127.0.0.1` here, rather than hardcode 
`0.0.0.0`. The latter causes the listener to bind to the public interfaces (and makes it reachable outside the machine) whereas 127 will bind on the loopback interface and only be exposed locally. For the case of running the agent in a network the `0.0.0.0` interface should be used, but during local dev `127.0.0.1` is more appropriate.

Let's flag that defaults to `127.0.0.1` that can optionally be set to `0.0.0.0`. For the name I think we can use `bind`, which seems to be used by grpcui, a few k8s tools and a few hashicorp tools.

Lets also add a log statementhere when the http server starts so the user is aware of what we bound to.

private/buf/cmd/buf/command/beta/studioagent/studioagent.go
```diff
@@ -0,0 +1,242 @@
...
+	var serverTLSConfig *tls.Config
+	if flags.ServerCert != "" || flags.ServerKey != "" {
+		serverTLSConfig, err = newTLSConfig(rootCAConfig, flags.ServerCert, flags.ServerKey)
+		if err != nil {
+			return fmt.Errorf("cannot create new server TLS config: %w", err)
+		}
+	}
+	mux := bufstudioagent.NewHandler(
+		container.Logger(),
+		container.Arg(0), // the origin from command argument
```
robbertvanginkel, line 184: Last thing: I think we should specify origin through a flag param that defaults to `studio.buf.build`. That'll be a bit more user friendly for the default case.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf beta studio-agent` command which supports running an HTTP(S) server as the Studio agent.
```
robbertvanginkel, line 5:
```suggestion
- Add `buf beta studio-agent` command to support the upcoming Buf Studio.
```

Given that "the Studio agent" doesn't say much about what its for, I think its better to note like this.

PR #1142, opened 2022-05-19
add studio service endpoints for preset agents

cyinma:
as the list of preset agents should be relatively short, and is only editable by server admin, so I am using the approach of setting the whole list instead of separated `add`, `update`, `delete` endpoints, aligning with the configurable recommendation endpoints in recommendation service


proto/buf/alpha/registry/v1alpha1/studio.proto
```diff
@@ -0,0 +1,45 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+// PresetAgent is the information about an agent in the studio.
+message PresetAgent {
```
robbertvanginkel, line 20: How about `StudioAgentPreset`? Considerning that `buf.alpha.registry.v1alpha1` is the package, I feel like something that starts with Studio would make sense.

cyinma: sounds good, updated!

proto/buf/alpha/registry/v1alpha1/studio.proto
```diff
@@ -0,0 +1,45 @@
...
+  string url = 1;
+  // The optional alias of the agent URL.
+  string alias = 2;
+}
+
+// StudioService is the Studio service.
+service StudioService {
+  // ListStudioAgentPresets returns a list of agent presets in the server.
+  rpc ListStudioAgentPresets(ListStudioAgentPresetsRequest) returns (ListStudioAgentPresetsResponse);
+  // SetStudioAgentPresets set the list of agent presets in the server.
```
robbertvanginkel, line 31:
```suggestion
  // SetStudioAgentPresets sets the list of agent presets in the server.
```

PR #1143, opened 2022-05-19
Remove tracks related flags and subcommands

robbertvanginkel:
Fixes BSR-130.


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   a file with the same name, but with different file types (i.e. a
   regular file vs. a directory).
 - Fix check for `PACKAGE_SERVICE_NO_DELETE` to detect deleted services.
+- Remove the `--track` flag from `buf push`.
```
robbertvanginkel, line 10: Although technically a breaking change in a non `beta` subcommand, the BSR will error on any value set for it so this seems ok to me.

Alternatively if we have some way of deprecating flags in a way that adds an explanation as to why the flag isn't available anymore I can do that as well.

bufdev (approved):

CHANGELOG.md
bufdev: I updated this

PR #1148, opened 2022-05-20
Update protoc version handling

bufdev:
The github.com/protocolbuffers/protobuf library versioning is changing:

- https://github.com/protocolbuffers/protobuf/releases/tag/v21.0-rc1
- https://developers.google.com/protocol-buffers/docs/news/2022-05-06

As such, our current version parsing scheme is broken. This updates the version parsing scheme to account for this, as well as fixes a few bugs:

- We needed to split the handling of whether to set the `--experimental_allow_proto3_optional` flag from whether or not a response should say it handles proto3 `optionals`. This splits that handling.
- Suffixes with a `-` in them, such as `rc-1`, were not properly handled.

### DIFF
--- a/private/pkg/app/appproto/appprotoexec/version.go
+++ b/private/pkg/app/appproto/appprotoexec/version.go
@@ -46,44 +46,58 @@ func parseVersionForCLIVersion(value string) (_ *pluginpb.Version, retErr error)
 	// protoc always starts with "libprotoc "
 	value = strings.TrimPrefix(value, "libprotoc ")
 	split := strings.Split(value, ".")
-	if len(split) != 3 {
-		return nil, errors.New("more than three components split by '.'")
+	if len(split) != 2 && len(split) != 3 {
+		return nil, fmt.Errorf("%d components split by '.'", len(split))
 	}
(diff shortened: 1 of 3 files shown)


private/pkg/app/appproto/appprotoexec/version.go
```diff
@@ -46,57 +46,90 @@ func parseVersionForCLIVersion(value string) (_ *pluginpb.Version, retErr error)
 	// protoc always starts with "libprotoc "
 	value = strings.TrimPrefix(value, "libprotoc ")
 	split := strings.Split(value, ".")
-	if len(split) != 3 {
-		return nil, errors.New("more than three components split by '.'")
+	if len(split) != 2 && len(split) != 3 {
+		return nil, fmt.Errorf("%d components split by '.'", len(split))
```
lrewega, line 50:
```suggestion
	if n := len(split); n != 2 && n != 3 {
		return nil, fmt.Errorf("%d components split by '.'", n)
```

wdyt?

lrewega (approved):

doriable (approved):

PR #1162, opened 2022-05-31
add httpserver package and provide request logging in studio-agent

cyinma:
fixes TCN-82

this is a followup of a [discussion in a previous PR](https://github.com/bufbuild/buf/pull/1137#discussion_r883907506), and this will also provide request logging for the studio-agent


private/pkg/transport/http/httpserver/httpserver.go
```diff
@@ -0,0 +1,180 @@
...
+
+// RunnerWithObservability returns a new RunnerOption that turns on
+// OpenCensus tracing and metrics.
+//
+// The default is to not turn on observability.
+func RunnerWithObservability(middleware func(http.Handler) http.Handler) RunnerOption {
+	return func(runner *runner) {
+		runner.observability = middleware
+	}
+}
```
robbertvanginkel, line 161: This looks like it has changed from the current internal version of this package? While this does drop the direct dependency on the observability libraries that it pulled in before, due to the way `RunnerWithObservability` is used it is now essentially an option to add arbitrary middleware here.

I think @mfridman had some ideas around how mappers/runners should take middleware, lets sync on this offline.

private/pkg/transport/http/httpserver/runner.go
```diff
@@ -0,0 +1,150 @@
...
+			s.logger.Info("finished", zap.Duration("duration", time.Since(start)))
+		}
+	}()
+
+	mux := newLoggingRouter(s.logger, zap.InfoLevel, chi.NewMux(), "")
+	mux.Use(middleware.Recoverer)
+	mux.Use(middleware.StripSlashes)
+	mux.Use(newZapMiddleware(s.logger))
+	// This sets a limit on the max body size. Ideally this could be configured
+	// in Connect itself: https://github.com/bufbuild/core/issues/4506
```
robbertvanginkel, line 74: Lets remove this comment.

private/pkg/transport/http/httpserver/runner.go
```diff
@@ -0,0 +1,150 @@
...
+	start := time.Now()
+	defer func() {
+		if retErr != nil {
+			s.logger.Error("finished", zap.Duration("duration", time.Since(start)), zap.Error(retErr))
+		} else {
+			s.logger.Info("finished", zap.Duration("duration", time.Since(start)))
+		}
+	}()
+
+	mux := newLoggingRouter(s.logger, zap.InfoLevel, chi.NewMux(), "")
```
robbertvanginkel, line 69: I think loggingrouter makes sense for us internally, but I don't think we want this in the studio agent.

private/pkg/transport/http/httpserver/httpserver.go
```diff
@@ -0,0 +1,180 @@
...
+// A router should route all requests with the path prefix to this handler.
+type PrefixedHTTPHandler interface {
+	http.Handler
+	PathPrefix() string
+}
+
+// NewPrefixedHTTPHandler creates a new PrefixedHTTPHandler using a handler and prefix.
+func NewPrefixedHTTPHandler(handler http.Handler, prefix string) PrefixedHTTPHandler {
+	return newPrefixedHTTPHandler(handler, prefix)
+}
```
robbertvanginkel, line 100: Some of these are unused in core, lets make sure we don't port that over here and just delete them instead.

private/pkg/transport/http/httpserver/runner.go
cyinma: that sounds reasonable, how do you think if I put the `mux` as a runner field, and use a runner option to put the logging router into it internally?

doriable (approved):
This generally looks good to me, however it seems like there is some offline discussion that needs to occur. Stamping from my perspective.

private/pkg/transport/http/httpserver/httpserver.go
```diff
@@ -0,0 +1,150 @@
...
+func RunnerWithTLSConfig(tlsConfig *tls.Config) RunnerOption {
+	return func(runner *runner) {
+		runner.tlsConfig = tlsConfig
+	}
+}
+
+// RunnerWithMiddlewares returns a new RunnerOption that use middlewares when the Runner Run.
+func RunnerWithMiddlewares(middlewares ...func(http.Handler) http.Handler) RunnerOption {
+	return func(runner *runner) {
+		runner.middlewares = append(runner.middlewares, middlewares...)
```
mfridman, line 122: Should this be an assignment, e.g., `runner.middlewares = middlewares` or are we appending middleware?

private/pkg/transport/http/httpserver/runner.go
```diff
@@ -0,0 +1,152 @@
...
+	mux := chi.NewMux()
+	mux.Use(middleware.Recoverer)
+	mux.Use(middleware.StripSlashes)
+	mux.Use(newZapMiddleware(s.logger))
+	if s.maxBodySize > 0 {
+		mux.Use(func(next http.Handler) http.Handler {
+			return http.MaxBytesHandler(next, s.maxBodySize)
+		})
+	}
+	mux.Use(s.middlewares...)
```
mfridman, line 79: Should user-defined middleware _always_ come last, maybe move this one line below?

private/pkg/transport/http/httpserver/httpserver.go
cyinma: assignment works for me, I'm just not sure how do we want to handle when this runner option is called multiple time? assignment will replace the existing middlewares. how do you think?

mfridman (approved):
Minor nits you may want to address, but after that feel free to merge.

PR #1163, opened 2022-05-31
Switches CLI to use the Connect protocol

smaye81:
This switches all clients in the CLI to use Connect as their underlying protocol.  In doing so, this removes any usage of `WithGRPC` as well as the logic for adding `WithGRPC` to clients in the generated code.


smaye81:
@robbertvanginkel I noticed that `private/bufpkg/bufstudioagent/plain_post_handler.go` is still using `WithGRPC`, but I left it alone in this PR.  I wasn't sure if you intended to keep this around, plus didn't want to break anything you were working on.

mfridman:
Mark it as Ready for review 💪

robbertvanginkel:
> @robbertvanginkel I noticed that `private/bufpkg/bufstudioagent/plain_post_handler.go` is still using `WithGRPC`, but I left it alone in this PR. I wasn't sure if you intended to keep this around, plus didn't want to break anything you were working on.

Thats fine for now, thanks for checking! In the future we should make that use whatever protocol is specified in the enveloped content type, similar to how we get the codec:

https://github.com/bufbuild/buf/blob/e8de962c9051bdcafa35e1d80811f8d68ce66af5/private/bufpkg/bufstudioagent/plain_post_handler.go#L146-L150

That way it'll be up to the request caller what the agent should be using.

mfridman (approved):
Easy stamp 🏷️ . 

Nothing like using a production-ready gRPC alternative.

PR #1165, opened 2022-05-31
Remove additional extraneous gRPC references

smaye81:
This removes some gRPC references in non-critical areas such as strings in unit tests and some test data.  It also removes unused code in the `private/pkg/rpc/` package.  Note that a future PR will remove this further, but this was low-hanging fruit that is no longer used and can be immediately removed.

In addition, this removes gRPC installation testing from `windows/test.bash`


private/buf/bufgen/config_test.go
```diff
@@ -36,7 +36,7 @@ func TestReadConfigV1Beta1(t *testing.T) {
 			{
 				Name:     "go",
 				Out:      "gen/go",
-				Opt:      "plugins=grpc",
+				Opt:      "plugins=connect",
```
mfridman, line 39: This probably doesn't impacts the test itself, but iirc `plugins=grpc` was a way to pass a plugin to `protoc-gen-go` via an option. This functionality was later removed in favor of distinct plugins.

Probably fine because we're testing the shape of the config and not actual behaviour, but might want to get a second opinion whether we want to change this or not for correctness.

mfridman (approved):
lgtm removing cruft. Note the comment about the tests.

private/buf/bufgen/config_test.go
smaye81: Makes sense.  @doriable any thoughts?

doriable (approved):
lgtm!

PR #1166, opened 2022-06-01
make upgrade and upgrade connect-go to v0.1.0

bufdev:


rubensf:
Retroactively approved

PR #1168, opened 2022-06-01
Add empty protos for webhooks

buildbreaker:
Intentionally empty for now. I'm looking to finalize the API a bit later but this will unblock prototyping work on my end and avoids a backwards incompatible change needed later

Following https://github.com/bufbuild/buf/pull/841


proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -0,0 +1,47 @@
...
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+// WebhookService exposes the functionality for a caller to
+// subscribe to or unsubscribe to a Webhook for a given repository.
+service WebhookService {
+  rpc SubscribeToRepository(SubscribeToRepositoryRequest) returns (SubscribeToRepositoryResponse) {}
+  rpc UnsubscribeToRepository(UnsubscribeToRepositoryRequest) returns (UnsubscribeToRepositoryResponse) {}
+  rpc TestRepository(TestRepositoryRequest) returns (TestRepositoryResponse) {}
```
bufdev, line 24: What is this for?

proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -0,0 +1,47 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+// WebhookService exposes the functionality for a caller to
+// subscribe to or unsubscribe to a Webhook for a given repository.
+service WebhookService {
+  rpc SubscribeToRepository(SubscribeToRepositoryRequest) returns (SubscribeToRepositoryResponse) {}
```
bufdev, line 22: Add doc to the rpcs (I think the doc is on the request/responses instead)

proto/buf/alpha/registry/v1alpha1/webhook.proto
buildbreaker: The intention here is to provide a way to initiate an event via curl to test out a registered webhook end to end

We can provide this as a way for the user to test out their webhook without doing a bunch of pushes but if we're not convinced we need this and if we want to avoid a backwards incompatible change, I can remove it!

buildbreaker: Let's remove it now then!

buildbreaker: I can hack a local endpoint to do it instead of using this generated method. We can add something like this on short order if we think it'll be useful

bufdev (approved):

PR #1173, opened 2022-06-06
add lang key to buf.plugin.yaml go config

smallsamantha:
add lang key for go version in buf.plugin.yaml to allow for go minimum version selection

[BSR-226](https://linear.app/bufbuild/issue/BSR-226/connect-go-plugin-and-requires-go118-go-mod)


mfridman (approved):
Just realized with `runtime.go.min_lang_version` .. we are already in the nested "go" block so `lang` might be redundant?

`runtime.go.min_version` or `runtime.go.minimum_version` might read a bit better, but I'll leave it to you. Lgtm either way.

PR #1177, opened 2022-06-07
add plugincurationservice for buf curated plugins

smallsamantha:
Adding initial PluginCurationService to create and interact with Buf-curated first party plugins.


bufdev:
@amckinney @smallsamantha @mfridman we need to talk about this

PR #1181, opened 2022-06-08
Do not sort in place when converting breaking and lint configs to JSON

doriable:
Currently, we create a deterministic JSON structure for breaking and
lint configs for the `b3` digest. However, we should not be sorting
in-place when creating these structures, since it mutates the underlying
configs. Instead, we should be creating a copy.

### DIFF
--- a/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
+++ b/private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
@@ -169,10 +169,12 @@ type idPathsJSON struct {
 func configToJSON(config *Config) *configJSON {
 	ignoreIDPathsJSON := make([]idPathsJSON, 0, len(config.IgnoreIDOrCategoryToRootPaths))
 	for ignoreID, rootPaths := range config.IgnoreIDOrCategoryToRootPaths {
-		sort.Strings(rootPaths)
+		rootPathsCopy := make([]string, len(rootPaths))
+		copy(rootPathsCopy, rootPaths)
+		sort.Strings(rootPathsCopy)
 		ignoreIDPathsJSON = append(ignoreIDPathsJSON, idPathsJSON{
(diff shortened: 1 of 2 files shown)


rubensf (approved):

private/bufpkg/bufcheck/bufbreaking/bufbreakingconfig/bufbreakingconfig.go
```diff
@@ -176,13 +176,18 @@ func configToJSON(config *Config) *configJSON {
 		})
 	}
 	sort.Slice(ignoreIDPathsJSON, func(i, j int) bool { return ignoreIDPathsJSON[i].ID < ignoreIDPathsJSON[j].ID })
-	sort.Strings(config.Use)
-	sort.Strings(config.Except)
-	sort.Strings(config.IgnoreRootPaths)
+	// We should not be sorting in place for the config structure, since it will mutate the
+	// underlying config ordering.
+	use := config.Use
```
bufdev, line 181: Unless I'm missing something, this doesn't change anything - slices are just pointers effectively, so this will continue to mutate the underlying slice

bufdev: Also, if we're doing this, it should be universal - we have the same problem with `rootPaths` above for example, if you want to say that `configToJSON` doesn't affect the underlying `Config` then we should be consistent about that (and add a comment to this effect)

doriable: Ah, blergh, you're right -- I should be using `copy` here. Fixing this up.

bufdev:
You need to do this consistently across the entire function - there are still places that are sorting that are mutating the underlying slice.

doriable:
> You need to do this consistently across the entire function - there are still places that are sorting that are mutating the underlying slice.

Was just pushing up this change, sorry ^^"

Issue #1077, opened 2022-04-20
include pre-commit config


ryanrhee:
same as https://github.com/bufbuild/buf/issues/340 ;

[pre-commit](https://github.com/pre-commit/pre-commit) has almost double the number of stars as buf (7.9k vs 4.1k at the time of writing),  so i'm a bit surprised it was written off as having not enough usage.

I'm asking the buf team to please reconsider setting up an officially supported pre-commit hook in the repo.

Thank you for your consideration.

bufdev:
Feel free to put up a PR for it, we can consider it - I'd like to see exactly what we're proposing, as we're not familiar with pre-commit:

- Are there separate hooks for lint, breaking, other actions?
- How to do configuration?
- What do we have to do in terms of maintenance?
- Does this have to live in bufbuild/buf?

asottile:
I was asked by ryan to throw some thoughts here -- hope this is helpful!  (@-mention me directly if you need clarification since I'll have muted)

>     * Are there separate hooks for lint, breaking, other actions?

it could be as many or as few makes sense for consumers -- though it's also flexible (consumers can override the arguments / options directly via `args` or `entry`)

>     * How to do configuration?

in ~theory configuration should work the same as it does for other users -- though there would be a tiny bit of config that a consumer would use to reference your repository (and any `args` they would want to customize there) -- it'd look something like:

```yaml
-   repo: https://github.com/bufbuild/buf
    rev: ...
    hooks:
    -   id: buf-validate  # or whatever the hook entry id is called!
```

>     * What do we have to do in terms of maintenance?

most integrations require zero maintenance after the initial commit -- though I think either ryan or I would be happy to field questions / support for this

>     * Does this have to live in bufbuild/buf?

it's usually easiest for consumers if the small yaml file ends up here (they know where to go for documentation, etc. and there won't need to be additional moving parts for releases / etc.).  a "mirror repo" can be set up, but it's a significantly worse experience for consumers

hope this clarifies things!

anthony

bufdev:
We're happy to consider it, but it's not something we can prioritize. We're happy to consider a PR though. Adding to wishlist for our own tracking.

ryanrhee:
Understood. I will put up a PR.

Is there a way to have invocations of `buf` use filenames as opposed to directories containing `buf.yaml` or `buf.work.yaml` files? i didn't see anything when i skimmed the docs, but i wanted to make sure.

amckinney:
[...]

PR #1182, opened 2022-06-08
Define pre-commit hooks

ryanrhee:
Addresses #1077 .

---

Define official pre-commit hooks so users can reference the buf hooks directly.

For `buf lint` and `buf format`, given a standard configuration (i.e. a `buf.yaml` or `buf.work.yaml` in the repo root), zero-arg invocations should function.

```yaml
repos:
  - repo: https://github.com/bufbuild/buf
    hooks:
      - id: buf-lint
      - id: buf-format
```

Breaking change detection requires a previous version to compare against, which typically is not included in the source repository. Thus, `buf breaking` needs additional configuration to function properly.

Users with `buf.yaml` or `buf.work.yaml` not in the repo root need to specify the directory in the `args` list. Similarly, any custom arguments should be specified in the `args` list. Multiple instances can be specified if there are multiple buf yaml files that need to be run- and in this case, the `files` filter should be updated to only contain proto files contained by each yaml file.

For example[^1]:

[^1]: example courtesy of @lrewega

```yaml
repos:
  - repo: https://github.com/bufbuild/buf
    hooks:
      - id: lint
        files: ^path/to/workspace/.*\.proto$
        args: [path/to/workspace/buf.work.yaml, --error-format=json]
      - id: lint
        files: ^other/path/.*\.proto$
        args: [other/path/buf.work.yaml, --error-format=json]
```

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,17 +2,17 @@
   name: buf breaking
   language: golang
   entry: buf breaking
-  files: '\.proto$'
+  types: [proto]
   pass_filenames: false
 - id: buf-lint
   name: buf lint
   language: golang
   entry: buf lint
-  files: '\.proto$'
+  types: [proto]
   pass_filenames: false
 - id: buf-format
   name: buf format
   language: golang
   entry: buf format -w --exit-code
-  files: '\.proto$'
+  types: [proto]
   pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,18 +2,17 @@
   name: buf breaking
   language: golang
   entry: buf breaking
-  # match files ending in .proto,
-  files: &protofiles '[^/]+\.proto$'
+  files: '\.proto$'
   pass_filenames: false
 - id: buf-lint
   name: buf lint
   language: golang
   entry: buf lint
-  files: *protofiles
+  files: '\.proto$'
   pass_filenames: false
 - id: buf-format
   name: buf format
   language: golang
   entry: buf format -w --exit-code
-  files: *protofiles
+  files: '\.proto$'
   pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,19 @@
+- id: buf-breaking
+  name: buf breaking
+  language: golang
+  entry: buf breaking proto
+  # match files ending in .proto, .protobuf, or .pb
+  files: &protofiles '\.(?:proto(?:buf)?|pb)$'
+  pass_filenames: false
+- id: buf-lint
+  name: buf lint
+  language: golang
+  entry: buf lint
+  files: *protofiles
+  pass_filenames: false
+- id: buf-format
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,19 @@
+- id: buf-breaking
+  name: buf breaking
+  language: golang
+  entry: buf breaking proto
+  # match files ending in .proto, .protobuf, or .pb
```
bufdev, line 5: what are `.protobuf` and `.pb` for? these aren't protobuf file extensions

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,19 @@
+- id: buf-breaking
+  name: buf breaking
+  language: golang
+  entry: buf breaking proto
```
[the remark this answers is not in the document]
ryanrhee: that shouldn't be there, it's a remnant from me testing the hooks. removed.

.pre-commit-hooks.yaml
ryanrhee: I've seen those extensions used for proto files in previous companies. Sounds like buf prefers to be opinionated about the extension here, so I updated it to only include the `.proto` extension.

this is overridable by end users so it will not preclude repos with other extensions.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,19 @@
+- id: buf-breaking
+  name: buf breaking
+  language: golang
+  entry: buf breaking
+  # match files ending in .proto,
+  files: &protofiles '[^/]+\.proto$'
```
bufdev, line 6: As opposed to using anchors, I'd prefer all three `files` keys are just `files: '[^/]+\.proto$'`.

Also, shouldn't this be `[^\/]+\.proto$`, and even moreso, is this regex correct? This won't work on Windows, as an example - we could probably simplify this down to just `\.proto$`.

.pre-commit-hooks.yaml
bufdev: `.proto` is the canonical Protobuf source file extensions, and is the only extension I've seen in the wild, and the only one recognized by `buf`.

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,19 @@
+- id: buf-breaking
+  name: buf breaking
+  language: golang
```
bufdev, line 3: Can you explain what the language key is for?

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,19 @@
...
+  entry: buf breaking
+  # match files ending in .proto,
```
bufdev, line 5: No need for this comment, this is self-explanatory

.pre-commit-hooks.yaml
ryanrhee: this is what instructs pre-commit to use `go install` to set up the environment that executes the `buf` executable.

bufdev: Got it, thanks

bufdev (approved):

PR #1190, opened 2022-06-10
Support protoc on arm64

pkwarren:
There are now binaries of protoc compiled for arm64 (which avoid the
requirement of using Rosetta).


doriable (approved):
Oh cool, lgtm!

Issue #1107, opened 2022-05-04
COMMENT_MESSAGE Lint rule exceptions for request/response types


sarahegler:
Prototool has a lint rule `MESSAGES_HAVE_SENTENCE_COMMENTS_EXCEPT_REQUEST_RESPONSE_TYPES`.

Is there a way to achieve this same level of linting in buf? It seems that `COMMENT_MESSAGE` requires a comment for every message type, including RPC request/response message types. 

If not, is this something you would consider adding?

amckinney:
Thanks for issue @sarahegler - we'd definitely consider adding this one.

We'd probably just name it `COMMENT_MESSAGE_EXCEPT_REQUEST_RESPONSE_TYPES` in our case (for consistency with the rest of our `lint` rules). You'll need to configure it explicitly though - we won't add it to any of the `lint` categories that already exist (e.g. `DEFAULT`) since that'd be a backwards incompatible change.

We might not get to this immediately due to some other priorities, but stay tuned and we'll try to get this change in soon!

sarahegler:
thanks so much @amckinney!

bufdev:
I started working on this issue in https://github.com/bufbuild/buf/pull/1191 - it's non-trivial, but consider this accepted as an issue we should work on eventually.

bufdev:
I'd like to extend the work here to really rethink how we enforce proper documentation in the style we'd suggest universally. I think in addition to the main issue, there's a few other points:

- No need to comment `_UNSPECIFIED` zero values of enums.
- Optimally, complete sentences in comments.
- Potentially, some type of enforcement around how to do links. This may be unnecessary for the built-in rules, however.

PR #1191, opened 2022-06-10
WIP: add comments_exclude_request_responses lint configuration

bufdev:
Starts work on https://github.com/bufbuild/buf/issues/1107.

This option needs to be fed through to `buflintbuild` into a common "comment adapter" that takes this option into account. If this option is set, any type being linted should not produce a lint error if it is a request or response type. This can be determined by first calling `protosource.RequestResponseMessageFullNames` at the top level. See `buflintbuild.EnumZeroValueSuffix` as an example of a lint rule that takes a configuration option.

This can be picked up as an onboarding task for someone who will be working on the CLI.


bufdev:
Moved to https://github.com/bufbuild/buf/tree/do-not-delete-this-branch-comments-exclude-request-responses - will close this PR for now.

Issue #1183, opened 2022-06-08
Filtering images by type is nondeterministic


mfridman:
In https://github.com/bufbuild/buf/pull/949 we added the ability to create filtered images containing only the specified types and their required dependencies.

The issue is some consumers require deterministic output. Example from [public slack](https://bufbuild.slack.com/archives/CRZ680FUH/p1654656935483599):

> Hello there. I’m using buf to build binary proto descriptors to inject into
envoy’s grpc-json transcoding filter. When it inspects the descriptors it seems
to want things to be ordered by dependencies so it doesn’t have to read the
next thing to validate the current thing which feels fair.

> If I filter my output using --type the order of the protos in the descriptor
set changes and fails envoy validation.

Based on this comment we knew this was going to be the case. Let's discuss if we want to fix this.

https://github.com/bufbuild/buf/blob/102100e61f95173ce323f47a64ef34f1d706eb98/private/bufpkg/bufimage/bufimageutil/bufimageutil.go#L169-L171

bufdev:
I think we definitely want to fix this if it's feasible in a short amount of time - there is almost for sure some kind of sort we can do right? Often it just means tldr keeping a `[]string` alongside your `map[string]whatever`.

robbertvanginkel:
When I was writing this code thinking about just the types, DAG ordering isn't always possible on a type level since types can have circular references. However, since the types are still outputted in their original (but trimmed down) files we should still be able to sort the files in dependency order.

Using the `reorder` in https://github.com/bufbuild/buf/blob/852c2ed5c1f410686e668695423376466a753bc5/private/bufpkg/bufimage/image.go#L29 should fix this.

PR #1201, opened 2022-06-15
Return DAG ordered images from ImageFilteredByTypes

robbertvanginkel:
Fixes https://github.com/bufbuild/buf/issues/1183. By using the same order (although with some filtered files if they are not required) as the provided input, we should maintain the DAG order like this.


private/bufpkg/bufimage/bufimageutil/bufimageutil_test.go
```diff
@@ -221,3 +222,16 @@ func runDiffTest(t *testing.T, testdataDir string, typenames []string, expectedF
 		require.NoError(t, writer.Close())
 	}
 }
+
+func imageIsDependencyOrdered(image bufimage.Image) bool {
```
bufdev, line 226: cute

bufdev (approved):

PR #1202, opened 2022-06-15
HTML escape some error output in studio-agent

robbertvanginkel:
To address:

- https://github.com/bufbuild/buf/security/code-scanning/3: this one is techinically a false positive, as the url.Parse would never succeed if there was a character that needed to be escaped in the scheme. Still encode the result of url.Parse insead.
- https://github.com/bufbuild/buf/security/code-scanning/2: this is pretty low impact, as only header names matching a user specified value would be returned (and it makes little sense for the operator to specify a value that could lead to xss issues).


lrewega (approved):

bufdev (approved):

bufdev:
Can we merge?

timostamm (approved):
I don't want to block. Error messages in the client will show <, >, &, ' and " escaped, but we can unescape them there.

PR #1205, opened 2022-06-16
Remove the unused connectclient plugin

smaye81:
This plugin is being removed as it is no longer used by any generation steps.


mfridman (approved):
lgtm. To confirm, we'll have a follow up in core to swap out the connectclient in favor of the provider abstraction?

PR #1208, opened 2022-06-16
Reorganize and cleanup rpc package 

smaye81:
To facilitate the eventual removal of the `private/pkg/rpc` package from `buf`, this begins the reorganization of code from that package.  Note that this should _not_ affect any functionality in the CLI and is effectively just a rearrangement and a deletion of code that is no longer needed.

The `private/pkg/rpc/headers.go` file will need to stay around for the foreseeable future because of the `incomingHeadersContextKey` and `outgoingHeadersContextKey` structs.  These are used for keys in the context and moving these to a different package would break the lookup for the token.

Since the above file is staying for a bit, all header-specific logic that was in `rpcauth` was moved to this file since it makes more sense to be located there.  In addition, the header constants were exported so that downstream consumers can all use the same values.

Most of `rpcauth` was also deleted as this code is not used in `buf` at all but is instead imported by other consumers.  If the intent is delete `rpcauth`, then this code will be moved to where it is used instead.

In addition, renamed the header functions in `headers.go` to specify more accurately that they are setting headers in the context.  This should make things more clear as the refactoring progresses.


private/pkg/rpc/rpcauth/rpcauth.go
```diff
@@ -16,106 +16,23 @@ package rpcauth
...
-	user, ok := userValue.(*User)
-	return user, ok
-}
-
-// WithToken adds the token to the context via a header.
+// WithToken adds the token to the outgoing context via a header.
 func WithToken(ctx context.Context, token string) context.Context {
 	if token != "" {
-		return rpc.WithOutgoingHeader(ctx, authenticationHeader, authenticationTokenPrefix+token)
+		return rpc.WithOutgoingContextHeader(ctx, rpc.AuthenticationHeader, rpc.AuthenticationTokenPrefix+token)
```
smaye81, line 26: The `WithToken` and `WithTokenIfNoneSet` functions will be replaced by Connect interceptors in a future PR.  Leaving here now because the ethos of this PR is to not change any functionality in the CLI.

bufdev:
At a high-level, I'd want to see the PR after this (perhaps chained on this) to understand why we're doing this re-org a bit better, because it seems like this is just combining two packages that will soon be deleted, and renaming some functions.

> The private/pkg/rpc/headers.go file will need to stay around for the foreseeable future because of the incomingHeadersContextKey and outgoingHeadersContextKey structs. These are used for keys in the context and moving these to a different package would break the lookup for the token.

Neither of these keys should be used - we should be using connect-go constructs and not using anything from `rpc`, we should be able to `rm -rf `private/pkg/rpc`. On the backend, we may need to handle `rpc-* auth headers, but I thought we were going to do that in envoy (or already are)?

smaye81:
> At a high-level, I'd want to see the PR after this (perhaps chained on this) to understand why we're doing this re-org a bit better, because it seems like this is just combining two packages that will soon be deleted, and renaming some functions.
> 
> > The private/pkg/rpc/headers.go file will need to stay around for the foreseeable future because of the incomingHeadersContextKey and outgoingHeadersContextKey structs. These are used for keys in the context and moving these to a different package would break the lookup for the token.
> 
> Neither of these keys should be used - we should be using connect-go constructs and not using anything from `rpc`, we should be able to `rm -rf `private/pkg/rpc`. On the backend, we may need to handle `rpc-* auth headers, but I thought we were going to do that in envoy (or already are)?

Yep, I can provide a glimpse at the follow-up PRs.

Agreed on the `contextKey` structs.  The follow-up PR in the CLI will remove the usage of those.  However, for the older CLIs in the wild, the backend will still need to look up these context keys so we can't remove or move them until older CLIs are no longer using them (unless I'm misunderstanding how the context lookup is done with that struct).

bufdev:
> However, for the older CLIs in the wild, the backend will still need to look up these context keys so we can't remove or move them until older CLIs are no longer using them (unless I'm misunderstanding how the context lookup is done with that struct).

The backend just needs to basically do this at our entrypoint:

```go
// these are the only two headers we sent
if httpHeader.Get("Authorization") == "" && httpHeader.Get("rpc-Authorization") != "" {
  httpHeaer.Set("Authorization, httpHeader.Get("rpc-Authorization")
}
if httpHeader.Get("buf-version") == "" && httpHeader.Get("rpc-buf-version") != "" {
  httpHeaer.Set("buf-versioh, httpHeader.Get("rpc-buf-version")
}
```

If that is set up, everything else should just work, so we don't need anything in `rpc`.

smaye81:
> > However, for the older CLIs in the wild, the backend will still need to look up these context keys so we can't remove or move them until older CLIs are no longer using them (unless I'm misunderstanding how the context lookup is done with that struct).
> 
> The backend just needs to basically do this at our entrypoint:
> 
> ```go
> // these are the only two headers we sent
> if httpHeader.Get("Authorization") == "" && httpHeader.Get("rpc-Authorization") != "" {
>   httpHeaer.Set("Authorization", httpHeader.Get("rpc-Authorization"))
> }
> if httpHeader.Get("buf-version") == "" && httpHeader.Get("rpc-buf-version") != "" {
>   httpHeaer.Set("buf-version", httpHeader.Get("rpc-buf-version"))
> }
> ```
> 
> If that is set up, everything else should just work, so we don't need anything in `rpc`.

OK, I got you.  I forgot about the custom RoundTripper putting context headers into the request headers.  OK, we can remove those contextKeys also

smaye81:
@bufdev closing this in favor of https://github.com/bufbuild/buf/pull/1211, which combines everything (removing the `rpc` package, switching to Connect interceptors, etc.)

PR #1210, opened 2022-06-16
Add user-facing buf.plugin.yaml

amckinney:
This adds a simple version of the `buf.plugin.yaml`. Some fields (e.g. `deps`, better `opts`, etc) are not included for now, but they will be added later on.

```yaml
version: v1
name: buf.build/grpc/go
opts:
  - paths=source_relative
runtime:
  go:
    min_version: v1.18
    deps:
      - "google.golang.org/grpc:v1.32.0"
```

This takes some inspiration from the [bufpluginconfig](https://github.com/bufbuild/buf/blob/main/private/bufpkg/bufpluginconfig/bufpluginconfig.go) package that already exists, but adapts some of the keys and internal implementation so that we separate the `ExternalConfig` from the internal `Config` representation. There are a couple things to consider with this one:

**Structured options**

Not all plugin options are expressed as key, value pairs. For example, `protoc-gen-java` has [an option](https://github.com/protocolbuffers/protobuf/issues/42) that are simple strings, just like we see in command line flags (e.g. `--dry-run`).

For now, we handle these cases by mapping the key (the option name) to an empty string value. This is susceptible to break down if there's ever a case that we need to explicitly specify the empty string as a value though. I imagine that we might change this later on (options are just free-form strings in the `buf.gen.yaml`), but this gives us something to work on top of for the time being - comments are left in-line.

**Runtime dependencies**

Runtime dependencies aren’t easily expressed in the same format across the board. Some output formats produce dependency names with ‘:’ (e.g. Java for Gradle/Maven) and ‘@‘ (e.g. JS and TS). It's also unclear how archive runtime dependencies can be reliably used in a generic fashion - the dependency format will be directly influenced by the target that's being produced.

For now, we can always treat the final ‘:’ element as the version, and trust that the user expressed their dependency in the format that they need. Then, it’s up to the registry (e.g. Go proxy, NPM registry, etc) to recognize this runtime dependency and act accordingly. All of this should be validated in the planned server-side unit test / verification (if possible).

### DIFF
--- a/private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
+++ b/private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
@@ -1,0 +1,205 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package bufpluginconfig defines the buf.plugin.yaml file.
+package bufpluginconfig
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginref"
+	"github.com/bufbuild/buf/private/pkg/encoding"
+	"github.com/bufbuild/buf/private/pkg/storage"
+)
+
+const (
+	// ExternalConfigFilePath is the default configuration file path for v1.
+	ExternalConfigFilePath = "buf.plugin.yaml"
+	// V1Version is the version string used to indicate the v1 version of the buf.plugin.yaml file.
+	V1Version = "v1"
+)
+
+var (
+	// AllConfigFilePaths are all acceptable config file paths without overrides.
+	//
+	// These are in the order we should check.
+	AllConfigFilePaths = []string{
+		ExternalConfigFilePath,
+	}
+)
+
+// Config is the plugin config.
+type Config struct {
+	// Name is the name of the plugin (e.g. 'buf.build/protocolbuffers/go').
+	Name bufpluginref.PluginIdentity
+	// Options is the set of options available to the plugin.
+	//
+	// For now, all options are string values. This could eventually
+	// support other types (like JSON Schema and Terraform variables),
+	// where strings are the default value unless otherwise specified.
+	//
+	// Note that some legacy plugins don't always express their options
+	// as key value pairs. For example, protoc-gen-java has an option
+	// that can be passed like so:
+	//
+	//  java_opt=annotate_code
+	//
+	// In those cases, the option value in this map will be set to
+	// the empty string, and the option will be propagated to the
+	// compiler without the '=' delimiter.
+	Options map[string]string
+	// Runtime is the runtime configuration, which lets the user specify
+	// runtime dependencies, and other metadata that applies to a specific
+	// remote generation registry (e.g. the Go module proxy, NPM registry,
+	// etc).
+	Runtime *RuntimeConfig
+}
+
+// RuntimeConfig is the configuration for the runtime of a plugin.
+//
+// Only one field will be set.
+type RuntimeConfig struct {
+	Go      *GoRuntimeConfig
+	NPM     *NPMRuntimeConfig
+	Archive *ArchiveRuntimeConfig
+}
+
+// GoRuntimeConfig is the runtime configuration for a Go plugin.
+type GoRuntimeConfig struct {
+	MinVersion string
+	Deps       []string
+}
+
+// NPMRuntimeConfig is the runtime configuration for a JavaScript NPM plugin.
+type NPMRuntimeConfig struct {
+	Deps []string
+}
+
+// ArchiveRuntimeConfig is the runtime configuration for a plugin that can be downloaded
+// as an archive instead of a language-specific registry.
+type ArchiveRuntimeConfig struct {
+	Deps []string
+}
+
+// GetConfigForBucket gets the Config for the YAML data at ConfigFilePath.
+//
+// If the data is of length 0, returns the default config.
+func GetConfigForBucket(ctx context.Context, readBucket storage.ReadBucket) (*Config, error) {
+	return getConfigForBucket(ctx, readBucket)
+}
+
+// GetConfigForData gets the Config for the given JSON or YAML data.
+//
+// If the data is of length 0, returns the default config.
+func GetConfigForData(ctx context.Context, data []byte) (*Config, error) {
+	return getConfigForData(ctx, data)
+}
+
+// ExistingConfigFilePath checks if a configuration file exists, and if so, returns the path
+// within the ReadBucket of this configuration file.
+//
+// Returns empty string and no error if no configuration file exists.
+func ExistingConfigFilePath(ctx context.Context, readBucket storage.ReadBucket) (string, error) {
+	for _, configFilePath := range AllConfigFilePaths {
+		exists, err := storage.Exists(ctx, readBucket, configFilePath)
+		if err != nil {
+			return "", err
+		}
+		if exists {
+			return configFilePath, nil
+		}
+	}
+	return "", nil
+}
+
+// ParseConfig parses the file at the given path as a Config.
+func ParseConfig(config string) (*Config, error) {
+	var data []byte
+	var err error
+	switch filepath.Ext(config) {
+	case ".json", ".yaml", ".yml":
+		data, err = os.ReadFile(config)
+		if err != nil {
+			return nil, fmt.Errorf("could not read file: %w", err)
+		}
+	default:
+		return nil, fmt.Errorf("invalid extension %s, must be .json, .yaml or .yml", filepath.Ext(config))
+	}
+	var externalConfig ExternalConfig
+	if err := encoding.UnmarshalJSONOrYAMLStrict(data, &externalConfig); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal plugin config: %w", err)
+	}
+	switch externalConfig.Version {
+	case V1Version:
+		return newConfig(externalConfig)
+	}
+	return nil, fmt.Errorf("invalid plugin configuration version: must be one of %v", AllConfigFilePaths)
+}
+
+// ExternalConfig represents the on-disk representation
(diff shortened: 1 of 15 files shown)


private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// NPMRuntimeConfig is the runtime configuration for a JavaScript NPM plugin.
+type NPMRuntimeConfig struct {
+	Deps []string
+}
+
+// ArchiveRuntimeConfig is the runtime configuration for a plugin that can be downloaded
+// as an archive instead of a language-specific registry.
+type ArchiveRuntimeConfig struct {
```
bufdev, line 94: I don't think this should exist, what's a use case of it?

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -0,0 +1,205 @@
...
+	// support other types (like JSON Schema and Terraform variables),
+	// where strings are the default value unless otherwise specified.
+	//
+	// Note that some legacy plugins don't always express their options
+	// as key value pairs. For example, protoc-gen-java has an option
+	// that can be passed like so:
+	//
+	//  java_opt=annotate_code
+	//
+	// In those cases, the option value in this map will be set to
```
bufdev, line 61: Questionable, but I understand it

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -0,0 +1,205 @@
...
+// If the data is of length 0, returns the default config.
+func GetConfigForData(ctx context.Context, data []byte) (*Config, error) {
+	return getConfigForData(ctx, data)
+}
+
+// ExistingConfigFilePath checks if a configuration file exists, and if so, returns the path
+// within the ReadBucket of this configuration file.
+//
+// Returns empty string and no error if no configuration file exists.
+func ExistingConfigFilePath(ctx context.Context, readBucket storage.ReadBucket) (string, error) {
```
bufdev, line 116: Do we need this?

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	return nil, fmt.Errorf("invalid plugin configuration version: must be one of %v", AllConfigFilePaths)
+}
+
+// ExternalConfig represents the on-disk representation
+// of the plugin configuration at version v1.
+type ExternalConfig struct {
+	Version string                `json:"version,omitempty" yaml:"version,omitempty"`
+	Name    string                `json:"name,omitempty" yaml:"name,omitempty"`
+	Opts    []string              `json:"opts,omitempty" yaml:"opts,omitempty"`
```
bufdev, line 158: Maybe just `options`? I don't know

bufdev: I guess `opts` is consistent with `deps, min_version`

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -0,0 +1,191 @@
...
+	}, nil
+}
+
+func newRuntimeConfig(externalRuntimeConfig ExternalRuntimeConfig) (*RuntimeConfig, error) {
+	var (
+		isArchiveEmpty = externalRuntimeConfig.Archive.IsEmpty()
+		isGoEmpty      = externalRuntimeConfig.Go.IsEmpty()
+		isNPMEmpty     = externalRuntimeConfig.NPM.IsEmpty()
+	)
+	if isArchiveEmpty && isGoEmpty && isNPMEmpty {
```
bufdev, line 90: Id probably do this as:

```go
if (!isArchiveEmpty && !isGoEmpty) || (!isArchiveEmpty && !isNPMEmpty) || (!isGoEmpty && !isNPMEmpty) {
  return error
}
// nil vs error, idk if we should distinguish, right?
return &RuntimeConfig {
  NPM: newNPM,
  Go: newGo,
  Archive: newArchive,
}
```

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -0,0 +1,191 @@
...
+	// This will not actually verify that the go version is
+	// in the valid set. It's impossible to capture the
+	// real set of valid identifiers at any given time (for
+	// an old version of the buf CLI) without reaching out to
+	// some external source at runtime.
+	//
+	// Note that this ensures the user's configuration specifies
+	// a 'v' prefix in the version (e.g. v1.18) even though the
+	// minimum version is rendered without it in the go.mod.
+	if externalGoRuntimeConfig.MinVersion != "" && !semver.IsValid(externalGoRuntimeConfig.MinVersion) {
```
srikrsna-buf, line 156: I think we should loose the `v` and use `golang.org/x/mod/modfile.GoVersionRE.MatchString()` to check the go version.

private/bufpkg/bufplugin/bufpluginconfig/config.go
bufdev: Or, if you want to make this scale to 10 more runtimes:

```go
emptyCount := 0
for _, isEmpty := range []bool {
  externalRuntimeConfig.Archive.IsEmpty(),
  externalRuntimeConfig.Go.IsEmpty(),
  ...
} {
  if isEmpty {
    emptyCount++
  }
}
if emptyCount > 1 {
  return error
}
```

private/bufpkg/bufplugin/bufpluginconfig/get.go
```diff
@@ -0,0 +1,125 @@
...
+	"io"
+
+	"github.com/bufbuild/buf/private/pkg/encoding"
+	"github.com/bufbuild/buf/private/pkg/storage"
+	"github.com/bufbuild/buf/private/pkg/stringutil"
+	"go.opencensus.io/trace"
+	"go.uber.org/multierr"
+)
+
+func getConfigForBucket(ctx context.Context, readBucket storage.ReadBucket) (_ *Config, retErr error) {
```
bufdev, line 29: Will `getConfigForBucket` work the same as `buf.yaml`, i.e. are we expecting to pass a bucket where `buf.plugin.yaml` should be at the root? I guess so yea

private/bufpkg/bufplugin/bufpluginconfig/config.go
bufdev: Or, add a `pkg/boolutil` with `Count([]bool) int`

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -0,0 +1,191 @@
...
+func newArchiveRuntimeConfig(externalArchiveRuntimeConfig ExternalArchiveRuntimeConfig) (*ArchiveRuntimeConfig, error) {
+	if err := validateRuntimeDeps(externalArchiveRuntimeConfig.Deps); err != nil {
+		return nil, err
+	}
+	return &ArchiveRuntimeConfig{
+		Deps: externalArchiveRuntimeConfig.Deps,
+	}, nil
+}
+
+func validateRuntimeDeps(dependencies []string) error {
```
srikrsna-buf, line 174: Instead of defining a format, since we already have language specific config, I think we should have language specific dependency definitions. It gives us,
- Idiomatic names (modules for go, packages for npm etc...)
- Language specific constraints (simple semver for go, semver with constraints for npm)
- Won't have to worry about archive registy as they can include anything that is at least human readable (package/version)

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -0,0 +1,205 @@
...
+	AllConfigFilePaths = []string{
+		ExternalConfigFilePath,
+	}
+)
+
+// Config is the plugin config.
+type Config struct {
+	// Name is the name of the plugin (e.g. 'buf.build/protocolbuffers/go').
+	Name bufpluginref.PluginIdentity
+	// Options is the set of options available to the plugin.
```
srikrsna-buf, line 49: What does it mean by options available to the plugin. Isn't this the default options that we pass to the plugin?

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
amckinney: For now, no - I'll remove it and we can add it back later if we ever do.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
amckinney: Agreed on it being questionable - left a note in the summary that mentions we might change this behavior later. Feels fine to start like this for now, and we can iterate on it as we go.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
amckinney: Yea, I'd like it to be consistent with the `buf.gen.yaml`, too.

private/bufpkg/bufplugin/bufpluginconfig/config.go
amckinney: Noted, I'll find a better way to do it 👍

srikrsna-buf: I would do it with `UnmarshalYAML` and `UnmarshalJSON`

private/bufpkg/bufplugin/bufpluginconfig/config.go
amckinney: I was hoping that _all_ versions would conform to the same style, as in "all of the runtime dependencies require a `v` prefix, so this should too for consistency". But I felt uneasy about this as I was going about it - I'll adjust this to use the function you just shared and we can go from there.

private/bufpkg/bufplugin/bufpluginconfig/config.go
amckinney: In general I agree, but I don't like that it puts more cognitive overhead on the user to understand how to specify dependencies for the specific runtime in question - it'd be great if we had standard way of specifying `deps` for the user-facing configuration (what this attempts to propose). This was something I discussed with @bufdev.

With that said, I admit that it feels like this is susceptible to break down at some point - we just need to be mindful of the ergonomics here.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
amckinney: Yea I can rephrase this comment a bit - they are the default options passed into the plugin.

pkwarren (approved):

amckinney:
Landing this one as-is - we have a couple branches stacked on top of this, so this will make it easier to continue iterating. I'll follow-up with all of the remaining concerns in a separate PR soon after this is merged.

PR #1218, opened 2022-06-20
Remove tracks/branches related proto

robbertvanginkel:
Fixes BSR-128. Putting this on draft while discussing frontend impact in FE-154.


proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -16,29 +16,35 @@ syntax = "proto3";
...
-    // The requested reference is a branch.
-    RepositoryBranch branch = 1;
     // The requested reference is a tag.
     RepositoryTag tag = 2;
     // The requested reference is a commit.
     RepositoryCommit commit = 3;
-    // The requested reference is a track.
-    RepositoryTrack track = 4;
+    // The requested reference is the default reference.
+    RepositoryMainReference main = 5;
```
robbertvanginkel, line 32: @seankimdev @timostamm related to FE-154, my understanding is that we'd need something like this if the `ReferenceService` continues to exist. The follow up pr in https://github.com/bufbuild/buf/pull/1219/files#diff-9c96176c5e8386f22fe9a9f8f8e32bd0dff996b855ec337245229311eda5508eR34 shows how drafts could be stacked on top of this.

seankimdev: I think that makes sense. @timostamm, what do you think?

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -49,8 +49,10 @@ message Repository {
 }
 
 message RepositoryCounts {
+  reserved 2;
+  reserved "tracks_count";
+
   uint32 tags_count = 1;
-  uint32 tracks_count = 2;
```
seankimdev, line 53: Is it okay to just remove this instead of deprecating it?

proto/buf/alpha/registry/v1alpha1/push.proto
```diff
@@ -28,15 +28,15 @@ service PushService {
 message PushRequest {
   string owner = 1;
   string repository = 2;
-  string branch = 3;
+  string branch = 3 [deprecated = true];
```
amckinney, line 31: Will we eventually mark these as `reserved`, too?

amckinney (approved):
Nice 👏

proto/buf/alpha/registry/v1alpha1/push.proto
robbertvanginkel: Yes! However throughout the cli version sometimes it'll send `main` as branch or as `track`, or none. We still want the server side to be able to read these to return informative error messages rather than push to the main track when a user did `buf push --track foo` because we can't read the track argument.

PR #1219, opened 2022-06-20
Add proto for draft commits

robbertvanginkel:
Fixes BSR-131. Depends on https://github.com/bufbuild/buf/pull/1218.


proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -40,6 +40,8 @@ message RepositoryCommit {
   string author = 7;
   // The tags associated with this commit
   repeated buf.alpha.registry.v1alpha1.RepositoryTag tags = 8;
+  // The commit's draft name, if it is a draft commit.
+  optional string draft_name = 9;
```
robbertvanginkel, line 44: We're not avid proto3 optional users, but this seemed most appropriate. As draft names are always non empty, we could remove the `optional` here and interpret empty as not a draft commit.

srikrsna-buf: I would just remove the optional and add a comment that empty is no draft.

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -61,6 +63,10 @@ service RepositoryCommitService {
   // GetRepositoryCommitBySequenceId returns the repository commit matching
   // the provided sequence ID and branch, if it exists.
   rpc GetRepositoryCommitBySequenceId(GetRepositoryCommitBySequenceIdRequest) returns (GetRepositoryCommitBySequenceIdResponse);
+  // ListRepositoryDraftCommits lists draft commits in a repository.
```
srikrsna-buf, line 66: I am guessing that we will use `GetRepositoryCommitByReference` as opposed to a dedicated rpc for draft (`GetRepositoryCommitByDraft`)?

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
timostamm: For languages ​​with nullable types, proto3 optionals usually provide a better experience than comparing against a zero value. I welcome exposure to this feature in our code base, but am okay with both ways.

proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -41,6 +43,13 @@ message RepositoryMainReference {
   RepositoryCommit commit = 2;
 }
 
+message RepositoryDraft {
+  // The name of the draft
+  string name = 1;
+  // The commit this draft points to.
+  RepositoryCommit commit = 2;
```
timostamm, line 50: Just for my understanding - there can be any number of drafts in a repository, but a draft always corresponds to exactly one commit?

robbertvanginkel: Correct. And when a user would push a new commit with the same draft name, it'll first remove any existing commit with that name if one exists.

timostamm (approved):

srikrsna-buf (approved):

seankimdev (approved):

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
bufdev: I really dislike proto3 optionals - is there any reason to disambiguate between unset and empty here? Is it possible to have a valid, empty draft name? If not, there is likely no use for this.

robbertvanginkel: Removed `optional` to match the rest of the api.

PR #1220, opened 2022-06-20
`make upgrade` and fix lint for deprecated `strings.Title`

doriable:
`make upgrade` and fix new lint issue, `strings.Title` is deprecated
in favour of `cases.Title` to better handle Unicode punctuation.


doriable:
Hmm, looks like I had a stale branch locally `make upgrade`. Going to open this in a new, clean branch :S

PR #1221, opened 2022-06-20
`make upgrade` and update `.gitignore`

doriable:


make/buf/all.mk
```diff
@@ -26,7 +26,8 @@ FILE_IGNORES := $(FILE_IGNORES) \
 	.vscode/ \
 	private/buf/cmd/buf/workspacetests/other/proto/workspacetest/cache/ \
 	private/bufpkg/buftesting/cache/ \
-	private/pkg/storage/storageos/tmp/
+	private/pkg/storage/storageos/tmp/ \
+	private/buf/cmd/buf/command/alpha/protoc/test.txt
```
cyinma, line 30: is it alphabetical order originally?

cyinma (approved):

make/buf/all.mk
doriable: Hmm, might be a nicer way to structure this -- I'll clean that up.

PR #1222, opened 2022-06-21
[BSR-279] Receive single event when creating a webhook subscription

unmultimedio:
For webhook events creation, we should receive a single event to subscribe, as we're returning just one event in the response.


buildbreaker (approved):
Thanks @unmultimedio !

PR #1226, opened 2022-06-23
Refactor and simplify buf.plugin.yaml configuration

amckinney:
This revises the `buf.plugin.yaml` according to the feedback in #1210. Specifically, this covers the following:

* Remove `archive` runtime config
* Simplify runtime configuration handling
* Use `modfile.GoVersionRE.MatchString`
* Cleanup `bufpluginconfig.Config` comment
* Use separate runtime configuration shapes (e.g. separate `go` and `npm` runtime keys).

Note that there's one `TODO` left in the implementation surrounding `npm` runtime dependency validation. For now, we permit any non-empty string for the version. There might be some good solutions for this to capture exactly what `npm` does [not] support, but that isn't handled for now - we'll need to follow-up here.

### DIFF
--- a/private/bufpkg/bufplugin/bufpluginconfig/config.go
+++ b/private/bufpkg/bufplugin/bufpluginconfig/config.go
@@ -149,7 +149,7 @@ func newNPMRuntimeConfig(externalNPMRuntimeConfig ExternalNPMRuntimeConfig) (*NP
 
 func newGoRuntimeConfig(externalGoRuntimeConfig ExternalGoRuntimeConfig) (*GoRuntimeConfig, error) {
 	if externalGoRuntimeConfig.MinVersion != "" && !modfile.GoVersionRE.MatchString(externalGoRuntimeConfig.MinVersion) {
-		return nil, fmt.Errorf("the go minimum version %q must be a valid semantic version", externalGoRuntimeConfig.MinVersion)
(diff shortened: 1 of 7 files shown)


private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -100,92 +105,73 @@ func newRuntimeConfig(externalRuntimeConfig ExternalRuntimeConfig) (*RuntimeConf
...
-	// real set of valid identifiers at any given time (for
-	// an old version of the buf CLI) without reaching out to
-	// some external source at runtime.
-	//
-	// Note that this ensures the user's configuration specifies
-	// a 'v' prefix in the version (e.g. v1.18) even though the
-	// minimum version is rendered without it in the go.mod.
-	if externalGoRuntimeConfig.MinVersion != "" && !semver.IsValid(externalGoRuntimeConfig.MinVersion) {
+	if externalGoRuntimeConfig.MinVersion != "" && !modfile.GoVersionRE.MatchString(externalGoRuntimeConfig.MinVersion) {
 		return nil, fmt.Errorf("the go minimum version %q must be a valid semantic version", externalGoRuntimeConfig.MinVersion)
```
pkwarren, line 152: Do we want to be more specific here - this will allow only `<major>.<minor>` syntax: https://pkg.go.dev/golang.org/x/mod/modfile#pkg-variables

pkwarren (approved):

PR #1228, opened 2022-06-23
initial pass at buf plugin image creation

pkwarren:
Update the `buf plugin push` command to support building a Docker image
from `Dockerfile.plugin` and pushing it to the BSR's registry. This will
create a docker image with the name `<remote>/<owner>/<name>` and a
unique ID for the tag locally, push it to the registry, and then clean
up the local image when complete (so we don't clutter up user's Docker
images).

This initially supports Docker with BuildKit and simple Docker images
where the context only contains the Dockerfile and no other files (no
support for ADD/COPY instructions).


private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -109,38 +113,44 @@ func run(
...
+
+	buildResponse, err := client.Build(ctx, dockerfile, pluginConfig, bufplugindocker.BuildParams{
+		ConfigDirPath: container.ConfigDirPath(),
+	})
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if _, err := client.Delete(ctx, buildResponse.Image); err != nil {
```
pkwarren, line 143: Our Docker image has a unique tag for each built image. We perform a best effort clean up here after a successful build to not clutter up developer's machines with our intermediate images. Buildkit caching doesn't require these (so subsequent builds still use a build cache for speedy builds).

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/moby/buildkit/session"
+	"go.uber.org/zap"
+)
+
+const pathBuildkitNodeID = ".buildkit_node_id"
```
pkwarren, line 31: Filename of persistent node id file stored in ~/.config/buf.

private/bufpkg/bufplugin/bufplugindocker/context.go
```diff
@@ -0,0 +1,91 @@
...
+// This is used to create images that don't rely on any other filesystem state to copy into the context.
+func createDockerContext(dockerfile io.Reader) (io.Reader, error) {
+	dockerfileBytes, err := io.ReadAll(dockerfile)
+	if err != nil {
+		return nil, err
+	}
+	var buffer bytes.Buffer
+	writer := tar.NewWriter(&buffer)
+	dockerfileInfo := &fileInfo{
+		name: "Dockerfile",
```
pkwarren, line 35: We store the file as `Dockerfile` even if it was named `Dockerfile.plugin` - this just removes one more non-default setting we have to pass to the Docker build.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+
+		response, err := d.cli.ImageBuild(ctx, dockerContext, types.ImageBuildOptions{
+			Tags:       []string{imageName},
+			Dockerfile: "Dockerfile",
+			Platform:   "linux/amd64",
+			Labels: map[string]string{
+				"build.buf.plugins.config.remote": pluginConfig.Name.Remote(),
+				"build.buf.plugins.config.owner":  pluginConfig.Name.Owner(),
+				"build.buf.plugins.config.name":   pluginConfig.Name.Plugin(),
+			},
```
pkwarren, line 102: These were default labels that I though might make it easier on the BSR side of things to determine more information about published images (since they'll be published with a unique tag).

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,79 @@
...
+const nodeIDLength = 32
+
+func createSession(contextDir, configDirPath string, logger *zap.Logger) (*session.Session, error) {
+	sharedKey := getBuildSharedKey(contextDir, configDirPath, logger)
+	s, err := session.NewSession(context.Background(), filepath.Base(contextDir), sharedKey)
```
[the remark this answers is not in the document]
amckinney: Regardless, we should explicitly pass in the `context.Context` we have from the layer interacting with this. We'll always have a `context.Context` from the CLI command, so we'll need to make sure it gets propagated as needed.

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,79 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/moby/buildkit/session"
+	"go.uber.org/zap"
+)
+
+const pathBuildkitNodeID = ".buildkit_node_id"
+const nodeIDLength = 32
```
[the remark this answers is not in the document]
amckinney: Let's leave this comment in the implementation, too.

Also, group the constants together in a single declaration.

```go
const (
  pathBuildkitNodeID = ".buildkit_node_id"
  // <YOUR_COMMENT>
  nodeIDLength = 32
)
```

private/bufpkg/bufplugin/bufplugindocker/context.go
amckinney: I wonder if we should just make the default filename we look for `Dockerfile`. As long as it's in a single directory alongside the `buf.plugin.yaml`, it would be obvious what it represents. With that said, I like the `.plugin` suffix so that it's clear what `Dockerfile` will be used as the primary entrypoint (if there's ever a reason to have multiple Dockerfiles exists alongside each other).

What are your thoughts?

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type Client interface {
+	// Build creates a Docker image for the plugin using the Dockerfile.plugin and plugin config.
+	Build(ctx context.Context, dockerfile io.Reader, config *bufpluginconfig.Config, params BuildParams) (*BuildResponse, error)
+	// Push the Docker image to the remote registry.
+	Push(ctx context.Context, image string, auth RegistryAuth) (*PushResponse, error)
```
amckinney, line 37: Now that I see how the `RegistryAuth` is used for now, we might want to make this an optional parameter with a `PushOption` interface.

```go
// PushOption is an option for Push.
type PushOption interface func(*pushOptions)

// WithRegistryAuthorizer returns a PushOption that configures registry authentication.
func WithRegistryAuthorizer(registryAuthorizer RegistryAuthorizer) PushOption {
	return func(pushOptions *pushOptions) {
		pushOptions.registryAuthorizer = registryAuthorizer
	}
}
```

It also looks like `FromHeader` isn't used - do we need to expose that on the `RegistryAuth` type that you have here? If not, we could start with a simple interface like this:

```go
// RegistryAuthorizer implements the Docker API authorization semantics.
type RegistryAuthorizer interface {
  // Header returns the authorization information as a base64 encoded JSON object.
  // This is suitable for passing to the Docker API as the X-Registry-Auth header.
  Header() string
}
```

This also means we could construct the `Header` value upfront rather than handling it all in the `ToHeader` call. If there's ever a reason why we need to call `ToHeader` multiple times, the value will already be computed for us.

```go
// bufplugindocker.go
func NewRegistryAuthorizer(...) (RegistryAuthorizer, error) {
  return newRegistryAuthorizer(...)
}
```

```go
// registry_authorizer.go
type registryAuthorizer struct {
  header string
}

func newRegistryAuthorizer(...) (*registryAuthorizer, error) {
  ...
  return &registryAuthorizer{
    header: header,
  }
}

func (r *registryAuthorizer) Header() string {
  return r.header
}
```

But I'll admit that a lot of this is splitting hairs - the single `Header` method doesn't feel all that great either (especially with Go's interface naming guidelines). My primary goal is to outline the general approach we take with thin interface definitions that are implemented by a single un-exported struct.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	// Close releases any resources used by the underlying Docker client.
+	Close() error
+}
+
+type BuildParams struct {
+	// ConfigDirPath specifies the location where the .buildkit_node_id should be persisted.
+	// This is used to establish a session when building images using buildkit.
+	// If empty, the node id will be generated randomly and not persisted to disk.
+	ConfigDirPath string
+}
```
amckinney, line 49: Rather than `*Params` structs, let's define functional options. This is similar to the `PushOption` recommendation above - let's make a `BuildOption` for this so that the default behavior doesn't require an explicit parameter at the call-site.

amckinney (commented):
Looking great so far - a few comments for now, but I'll continue to review.

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -109,38 +113,44 @@ func run(
...
-	//  // Build and push to the OCI registry using the Docker library ...
-	//
-	// ---
-	//
+	defer func() {
+		retErr = multierr.Append(retErr, client.Close())
+	}()
+
+	buildResponse, err := client.Build(ctx, dockerfile, pluginConfig, bufplugindocker.BuildParams{
+		ConfigDirPath: container.ConfigDirPath(),
```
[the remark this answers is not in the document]
pkwarren: Docker creates `~/.docker/buildNodeID` (this doesn't go to `~/Library/Caches/com.docker.docker` or other types of cache locations). To me it felt like a file which we'd want to remain pretty consistent as an identifier for the node doing the builds but I don't know details on how it is used. Fly.io persists it to their config file w/ Viper.

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
pkwarren: Good call - will update!

private/bufpkg/bufplugin/bufplugindocker/context.go
pkwarren: I like the plugin suffix. I'm only storing it as Dockerfile in the context since we don't have to set it in two places (here and in ImageBuildOptions.Dockerfile).

private/bufpkg/bufplugin/bufplugindocker/docker.go
pkwarren: > Now that I see how the `RegistryAuth` is used for now, we might want to make this an optional parameter with a `PushOption` interface.

It is actually required - https://docs.docker.com/engine/api/v1.41/#operation/ImagePush. Even if you don't populate the auth struct, you have to at least send across something otherwise the client will fail to push.

> It also looks like `FromHeader` isn't used - do we need to expose that on the `RegistryAuth` type that you have here?

I added FromHeader really just to verify marshaling to/from headers. We won't write a Docker engine implementation any time soon though so I can un-export that method or move it to test code if you'd like.

> This also means we could construct the `Header` value upfront rather than handling it all in the `ToHeader` call. If there's ever a reason why we need to call `ToHeader` multiple times, the value will already be computed for us.

I think we could add caching to this if needed. Happy to refactor as proposed but I think this won't really need to have alternative implementations so felt simpler this way.

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -109,38 +113,44 @@ func run(
...
+		}
+		return err
 	}
-	if !pluginSourceFilePath {
-		return fmt.Errorf("please define a %s plugin source file in the target directory", bufpluginsource.DockerSourceFilePath)
+	defer func() {
+		retErr = multierr.Append(retErr, dockerfile.Close())
+	}()
+
+	client, err := bufplugindocker.NewClient(bufplugindocker.WithLogger(container.Logger()))
```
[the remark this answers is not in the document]
amckinney: We often expose the `*zap.Logger` as a positional parameter -  no options needed here since it's always available.

Feel free to adapt this so that you can just call:

```go
client, err := bufplugindocker.NewClient(container.Logger())
```

private/bufpkg/bufplugin/bufplugindocker/auth.go
```diff
@@ -0,0 +1,52 @@
...
+
+package bufplugindocker
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"strings"
+)
+
+type RegistryAuth struct {
```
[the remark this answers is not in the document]
amckinney: Got it, if we end up keeping this concrete struct because it matches the Docker API, let's add a reference and annotate this as such.

```suggestion
// RegistryAuth represents the fields required to authenticate with the Docker Engine API.
// Ref: https://docs.docker.com/engine/api/v1.41/#section/Authentication
type RegistryAuth struct {
```

Also, if this isn't moved into the `bufplugindocker.go` file, let's rename the file to match the `type` name (i.e. `registry_auth.go`).

private/bufpkg/bufplugin/bufplugindocker/context.go
```diff
@@ -0,0 +1,91 @@
...
+		size: int64(len(dockerfileBytes)),
+		mode: 0644,
+	}
+	header, err := tar.FileInfoHeader(dockerfileInfo, "")
+	if err != nil {
+		_ = writer.Close()
+		return nil, err
+	}
+	if err := writer.WriteHeader(header); err != nil {
+		_ = writer.Close()
```
amckinney, line 45: Let's defer the `writer.Close` call as soon as the `writer` is constructed instead of manually calling it in all these places.

We'll also want to capture the error and we can do so with a named return value (you'll see this elsewhere too).

```go
func createDockerContext(dockerfile io.Reader) (_ io.Reader, retErr error) {
  ...
  writer := tar.NewWriter(&buffer)
  defer func() {
    retErr = multierr.Append(retErr, writer.Close()
  }()
  ...
}
```

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginconfig"
+	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/client"
+	"github.com/docker/docker/pkg/stringid"
+	"go.uber.org/multierr"
+	"go.uber.org/zap"
+	"golang.org/x/sync/errgroup"
+)
+
+type Client interface {
```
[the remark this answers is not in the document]
amckinney: Make sure to add a comment to all of the exported types.

private/bufpkg/bufplugin/bufplugindocker/auth.go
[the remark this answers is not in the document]
amckinney: We might also want to rename this to `RegsitryAuthConfig` since it really is just a configuration that can map to a different representation. More details in a comment on the call-site below.

private/bufpkg/bufplugin/bufplugindocker/docker.go
amckinney: Sounds good - if we keep it in its current state, let's update this to a pointer `*RegistryAuth`. We do this elsewhere, and the methods you defined on the type specify pointer receivers, so this actually matches what you've implemented there.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,203 @@
...
+	Build(ctx context.Context, dockerfile io.Reader, config *bufpluginconfig.Config, options ...BuildOption) (*BuildResponse, error)
+	// Push the Docker image to the remote registry.
+	Push(ctx context.Context, image string, auth RegistryAuth) (*PushResponse, error)
+	// Delete removes the Docker image from local Docker Engine.
+	Delete(ctx context.Context, image string) (*DeleteResponse, error)
+	// Close releases any resources used by the underlying Docker client.
+	Close() error
+}
+
+type BuildOption func(*buildParams)
```
amckinney, line 44: The un-exported struct should match the exported name.
```suggestion
type BuildOption func(*buildOptions)
```

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,203 @@
...
+
+func WithConfigDirPath(path string) BuildOption {
```
amckinney, line 46: Don't forget comments on all these exported types/functions.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	// It is created from the bufpluginconfig.Config's Name.IdentityString() and a unique id.
+	Image string
+	// Digest specifies the Docker image digest in the format <hash_algorithm>:<hash>.
+	// Example: sha256:65001659f150f085e0b37b697a465a95cbfd885d9315b61960883b9ac588744e
+	Digest string
+}
+
+type PushResponse struct{}
+
+type DeleteResponse struct{}
```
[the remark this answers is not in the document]
amckinney: Sounds good - let's just leave a comment explaining that.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+
+	dockerContext, err := createDockerContext(dockerfile)
+	if err != nil {
+		return nil, err
+	}
+
+	eg, errGroupCtx := errgroup.WithContext(ctx)
+	eg.Go(func() error {
+		return buildkitSession.Run(errGroupCtx, func(ctx context.Context, proto string, meta map[string][]string) (net.Conn, error) {
+			return d.cli.DialHijack(ctx, "/session", proto, meta)
```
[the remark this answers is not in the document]
amckinney: Might be another candidate for a comment for future contributors. I'll leave it to you.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,203 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	dockerContext, err := createDockerContext(dockerfile)
+	if err != nil {
+		return nil, err
+	}
+
+	eg, errGroupCtx := errgroup.WithContext(ctx)
```
amckinney, line 88: If we really _must_ use `errgroup` here, we should explain why. Otherwise, we should use `pkg/thread`. I know you had a good reason for this, but let's make sure it's loud and clear.

private/bufpkg/bufplugin/bufplugindocker/docker.go
amckinney: What happens if the same image digest is actually produced by multiple plugins? Does it actually make sense to assign this to a single remote/owner/name combination? For example, the `buf.build/alex/go` and `buf.build/philip/go` plugins might use the same `Dockerfile.plugin`.

Or are you saying that we'll always only have a unique representation here via the unique tag, regardless of the image digest?

amckinney: We might need to be mindful of the labels that we define here - if it can ever be potentially part of our API, we need to make sure not to break them. That could put a little more weight on this decision, but that might not actually be a real concern.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,203 @@
...
+				"build.buf.plugins.config.owner":  pluginConfig.Name.Owner(),
+				"build.buf.plugins.config.name":   pluginConfig.Name.Plugin(),
+			},
+			Version:   types.BuilderBuildKit, // DOCKER_BUILDKIT=1
+			SessionID: buildkitSession.ID(),
+		})
+		if err != nil {
+			return err
+		}
+		defer response.Body.Close()
```
amckinney, line 114: We should capture these `Close` errors in a deferred function (the same applies to the other deferred function in this `eg.Go` call above):
```suggestion
		defer func() {
		    retErr = multierr.Append(retErr, response.Body.Close())
		}()
```

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+	cli, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, err
+	}
+	dockerClient := &dockerAPIClient{cli: cli}
+	for _, option := range options {
+		option(dockerClient)
+	}
+	if dockerClient.logger == nil {
+		dockerClient.logger = zap.L()
```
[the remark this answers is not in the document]
amckinney: If we require a `*zap.Logger` in the constructor so that we can always instrument it rather than using `zap.L`. This was mentioned in another comment above, but just mentioning it here for clarity.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -0,0 +1,114 @@
...
+	_, err := buildDockerPlugin(t, dockerClient, "testdata/success/Dockerfile")
+	require.Nil(t, err)
+}
+
+func TestMain(m *testing.M) {
+	if cli, err := client.NewClientWithOpts(client.FromEnv); err == nil {
+		ctx, cancel := context.WithTimeout(context.Background(), time.Second)
+		defer cancel()
+		if _, err := cli.Ping(ctx); err == nil {
+			dockerEnabled = true
```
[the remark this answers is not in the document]
amckinney: I wonder if we should require that these tests are run in the CI environments we care about though. Otherwise if there's ever a runner that can't successfully ping, we'll have a bunch of silent failures and the build will be green because the tests will have been skipped.

amckinney (commented):
Overall this is awesome! Some final style comments and organizational recommendations, but I think this is in a pretty good spot.

private/bufpkg/bufplugin/bufplugindocker/context.go
pkwarren: Reimplementing os.FileInfo is required here to create a tar entry from an io.Reader. This code is written where we don't make assumptions on where the Dockerfile exists (is it loaded from a tar file, a zip file, a directory, etc.).

What we're basically doing here is the following:
* Take Dockerfile.plugin from a storage.ReadBucket (this could be any of the supported types).
* Create a tar file which writes a single file - including the entry (this is what the FileInfo is needed for) and the contents of the file.
* Pass that as input to the Docker image build command. If we specify the file in the tar file as Dockerfile, then we don't have to override the default value here: https://docs.docker.com/engine/api/v1.41/#operation/ImageBuild.

We could change this to ensure that we have a real directory to build from (essentially extracting the Dockerfile.plugin to a temporary directory if it is loaded from an archive file), but that felt like more edge cases to cover with the initial pass.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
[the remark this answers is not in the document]
pkwarren: Since Docker isn't documented as a requirement to run tests, I was reluctant to make these required for others building/testing Buf. I do think we'd want to catch in CI though if these were for some reason disabled (i.e. we switched to a different CI system that didn't provide Docker).

What do you think about having an environment variable `TEST_DOCKER_FORCE=true` which could be set in our Github actions CI workflows to ensure we run these tests each time? If that was set, we'd skip the ping check and just enable these tests.

mfridman (commented):
Fantastic job, can't wait to try this. Alex covered most of it, just minor questions from me.

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,87 @@
...
+	var nodeIDPath string
+	if len(configDirPath) > 0 {
+		nodeIDPath = filepath.Join(configDirPath, pathBuildkitNodeID)
+		if nodeID := loadNodeID(logger, nodeIDPath); len(nodeID) > 0 {
+			return nodeID
+		}
+	}
+	b := make([]byte, nodeIDLength)
+	if _, err := rand.Read(b); err != nil {
+		return configDirPath
```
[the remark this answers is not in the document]
pkwarren: It is strange but that's what the other API consumers I found that created a session did as a fallback: https://github.com/docker/docker-ce/blob/523cf7e71252013fbb6a590be67a54b4a88c1dae/components/cli/cli/command/image/build_session.go#L48

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -109,38 +112,46 @@ func run(
...
+	// buildkit maintains a separate build cache so removing the image doesn't appear to impact future rebuilds.
+	defer func() {
+		if _, err := client.Delete(ctx, buildResponse.Image); err != nil {
+			// Not fatal - we did our best to clean up after ourselves.
+			container.Logger().Warn("failed to delete image", zap.String("image", buildResponse.Image), zap.Error(err))
+		}
+	}()
+
+	// TODO: Implement authentication to BSR registry using ~/.netrc
+	_, err = client.Push(ctx, buildResponse.Image, &bufplugindocker.RegistryAuthConfig{})
```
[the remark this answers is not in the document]
pkwarren: Pushing the same image doesn't trigger an error - you do see 'layer already exists' logged from response status.

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,87 @@
...
+	"go.uber.org/zap"
+)
+
+const (
+	pathBuildkitNodeID = ".buildkit_node_id"
+	// Matches length of node ID in Docker CLI
+	nodeIDLength = 32
+)
+
+func createSession(ctx context.Context, logger *zap.Logger, contextDir, configDirPath string) (*session.Session, error) {
```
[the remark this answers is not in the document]
pkwarren: To be honest I need to do some more research to understand how this is used in the Docker engine.

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -109,38 +112,46 @@ func run(
...
+		return err
+	}
+
+	// We build a Docker image using a unique ID label each time.
+	// After we're done publishing the image, we delete it to not leave a lot of images left behind.
+	// buildkit maintains a separate build cache so removing the image doesn't appear to impact future rebuilds.
+	defer func() {
+		if _, err := client.Delete(ctx, buildResponse.Image); err != nil {
+			// Not fatal - we did our best to clean up after ourselves.
+			container.Logger().Warn("failed to delete image", zap.String("image", buildResponse.Image), zap.Error(err))
```
amckinney, line 145: We might just want to use `multierr` here and let the command return the error (similar to what we do for `client.Close` above). I know that there's a good reason to emit a warning (the command was technically successful - we just didn't clean up some of the resources), but returning the error in these cases is consistent with our other commands.
```suggestion
			retErr = mutlierr.Append(retErr, fmt.Errorf("failed to delete image %q", buildResponse.Image))
```

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,87 @@
...
+	sharedKey := getBuildSharedKey(logger, contextDir, configDirPath)
+	s, err := session.NewSession(ctx, filepath.Base(contextDir), sharedKey)
+	if err != nil {
+		return nil, err
+	}
+	return s, nil
+}
+
+func getBuildSharedKey(logger *zap.Logger, dir, configDirPath string) string {
```
amckinney, line 46: nit: We usually explicitly type the type of the parameter in every position, regardless if the same types are adjacent. Also, we should name parameters consistently across the functions that interact with it.
```suggestion
func getBuildSharedKey(logger *zap.Logger, contextDir string, configDirPath string) string {
```

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
```diff
@@ -0,0 +1,87 @@
...
+
+func loadNodeID(logger *zap.Logger, nodeIDPath string) string {
+	var nodeID string
+	if nodeIDBytes, err := os.ReadFile(nodeIDPath); err == nil {
+		nodeID = strings.TrimSpace(string(nodeIDBytes))
+		decoded, err := hex.DecodeString(nodeID)
+		if err != nil || len(decoded) != nodeIDLength {
+			// Ignore node id - not in expected format
+			logger.Debug("invalid buildkit node id - ignoring", zap.String("path", nodeIDPath))
+			nodeID = ""
```
amckinney, line 83: I wonder if we should just return an error in this case - something is clearly wrong with the user's system, but we ignore it and continue. I know we _can_ because we can just generate a new nodeID, but I'm not sure if we should.

private/bufpkg/bufplugin/bufplugindocker/context.go
amckinney: Got it, this seems reasonable for now then. Please write up some of the justification on the comment above the `fileInfo` type declaration though - this context is important for future readers.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,191 @@
...
+			Version:   types.BuilderBuildKit, // DOCKER_BUILDKIT=1
+			SessionID: buildkitSession.ID(),
+		})
+		if err != nil {
+			return err
+		}
+		defer response.Body.Close()
+		scanner := bufio.NewScanner(response.Body)
+		for scanner.Scan() {
+			d.logger.Debug(scanner.Text())
```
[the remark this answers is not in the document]
amckinney: Ok we can come back to this - would you mind writing a comment here, too? I notice a similar case in the `Push` method below, so we may as well include it there too.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -0,0 +1,213 @@
...
+func NewClient(logger *zap.Logger) (Client, error) {
+	if logger == nil {
+		return nil, errors.New("logger required")
+	}
+	cli, err := client.NewClientWithOpts(client.FromEnv)
+	if err != nil {
+		return nil, err
+	}
+	dockerClient := &dockerAPIClient{cli: cli, logger: logger}
+	return dockerClient, nil
```
amckinney, line 208: nit: I would just return the instance without a variable declaration.
```suggestion
	return &dockerAPIClient{
	    cli: cli,
	    logger: logger,
	}, nil
```

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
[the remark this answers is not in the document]
amckinney: An environment variable would work - we can handle this in a follow-up though. As long as this doesn't yield flaky results in our CI right now, I think it's fine to mark it with a `TODO` and revisit. If this is useful for testing locally in the interim, that's fine with me.

amckinney (approved):
Awesome work here, excited to see this in practice! There's a few things we'll need to follow-up on (testing in CI and a variety of other `TODO`s). Otherwise, I only have a few nits - I'll leave it to you to address them before merging.

private/bufpkg/bufplugin/bufplugindocker/buildkit.go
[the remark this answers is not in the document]
mfridman: Ye that's fair, everything about this PR made sense to me except this chunk. Would be nice to document the behavior. Maybe we can add a TODO in the upcoming "Create Plugin" PR in this chunk of code if we haven't documented it.

PR #1232, opened 2022-06-24
Change ENUM_ZERO_VALUE_SUFFIX to match the full value name

russellhaering:
Change ENUM_ZERO_VALUE_SUFFIX to match the full value name, instead of only checking the suffix.

Previously no lint rule would catch a zero value with "stuff in the middle", eg `TYPE_NAME_OTHER_STUFF_UNSPECIFIED`.

### DIFF
--- a/private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
+++ b/private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
@@ -245,7 +245,8 @@ func checkEnumZeroValueSuffix(add addFunc, enumValue protosource.EnumValue, suff
 		return nil
 	}
 	name := enumValue.Name()
-	if !strings.HasSuffix(name, suffix) {
+	expectedName := fieldToUpperSnakeCase(enumValue.Enum().Name()) + suffix
+	if name != expectedName {
 		add(
 			enumValue,
 			enumValue.NameLocation(),
@@ -254,9 +255,9 @@ func checkEnumZeroValueSuffix(add addFunc, enumValue protosource.EnumValue, suff
 			[]protosource.Location{
 				enumValue.Enum().Location(),
 			},
-			"Enum zero value name %q should be suffixed with %q.",
+			"Enum zero value name %q should be %q.",
 			name,
-			suffix,
+			expectedName,
 		)
 	}
 	return nil


amckinney:
Nice, looks good! We'll need to add a `CHANGELOG.md` entry for this, then we can release this as a patch (as long as no other minor features have snuck in).

amckinney:
It'd also be good to add a test that explicitly tests this - we can do that though, no need on your end. Just a note for us before we merge/release.

bufdev:
This PR should actually be rolled back - the behavior specified was intended. Lint rules should be independent, and this was meant to just enforce the suffix, and not also enforce other enum value naming behaviors - that is the responsibility of other rules.

bufdev:
Specifically, if you want to enforce that `enum FooBar` has zero value `FOO_BAR_UNSPECIFIED`, that is the combined job of `ENUM_ZERO_VALUE_SUFFIX` (which checks the suffix of the zero value as `UNSPECIFIED`) and `ENUM_VALUE_PREFIX` (which checks that it starts with `FOO_BAR`). You should be able to opt into checking that the zero value ends with `UNSPECIFIED` without enforcing that it has a prefix named after the enum.

russellhaering:
Yeah, I wasn't too sure about this for exactly that reason.

The problem is that independently checking the prefix and the suffix doesn't achieve the overall style objective of making sure that the zero value is named exactly `${prefix}_{suffix}` - which is done to avoid inadvertently assigning semantic meaning to the zero value. I'm not sure how to check for this without also looking at the prefix. Perhaps there is a way to introduce a dependency between the two rules so this won't won't run unless there is an exact prefix match?

To give an example of how having two independent rules will cause problems for users, I stumbled across this because I have an enum:

```proto
enum Operator {
  EXISTS = 0;
  NOT_EXISTS = 1;
}
```

The first time I linted it, I got two errors:

```
Enum value name "EXISTS" should be prefixed with "OPERATOR_".
Enum zero value name "EXISTS" should be suffixed with "_UNSPECIFIED".
```

I corrected those errors in the order they were printed, resulting in the following enum:

```proto
enum Operator {
  OPERATOR_EXISTS_UNDEFINED = 0;
  OPERATOR_NOT_EXISTS = 1;
}
```

This looked slightly odd, but linted just fine, so I figured maybe it was just trying to signal that the zero value was "exists _or_ unset". I trusted that guess enough that I committed the code before having second thoughts and looking up a style guide.

Maybe a different approach would be just to introduce a new rule (or rename this one) that uses this same logic but is named / framed in terms of only using the zero value to signify an unset / invalid input?

PR #1241, opened 2022-06-27
Revert "Change ENUM_ZERO_VALUE_SUFFIX to match the full value name"

bufdev:
Reverts bufbuild/buf#1232


amckinney (approved):
Re: https://github.com/bufbuild/buf/pull/1232#issuecomment-1167671373

PR #1242, opened 2022-06-27
Revert "Add tests and changelog for ENUM_ZERO_VALUE_SUFFIX"

bufdev:
Reverts bufbuild/buf#1239


bufdev:
Merging this into #1241 as they should be together

PR #1243, opened 2022-06-27
[WIP / do not merge] Better support for remote plugins in buf.gen.yaml

amckinney:
## Background

Now that the `buf.plugin.yaml` is defined and (soon) able to be pushed to the BSR, this extends the `buf.gen.yaml` syntax so that it can recognize and interact with these plugins. The new plugin path representation omit the `/plugins/` path element so that fully-qualified plugin names look like any other ordinary BSR resource:

```yaml 
version: v1
plugins:
  - buf.build/library/java:v21.1.0-1
```

This syntax leverages a custom `{json,yaml}.Unmarshaler` implementation so that it implicitly assigns the above configuration to the following:

```yaml
version: v1
plugins:
  - name: buf.build/library/java:v21.1.0-1
    out: buf.build/library/java
    strategy: all
```

This is supported for both YAML and JSON so you can even run the following:

```sh
$ buf generate --template '{"version": "v1", "plugins": ["buf.build/library/java:v21.1.0-1"]}'
```

## Open Questions

The current proposal is up for debate - we need to consider a few things before moving forward here.

1. What should the default `out` value be? The fully-qualified representation makes it so that two plugins with the same base name don't collide (e.g. `buf.build/protocolbuffers/java` vs `buf.build/grpc/java`), but it makes the generated code path far more verbose:

```sh
.
└── buf.build
    └── protocolbuffers
        └── java
            └── buf
                └── alpha
                    └── breaking
                       └── v1
                           └── ConfigOuterClass.java
```

To be clear, the alternative is to map the shorthand representation to the following:

```yaml
version: v1
plugins:
  - name: buf.build/library/java:v21.1.0-1
    out: java
    strategy: all
```

No matter what we decide, the user will be able to override the default behavior by explicitly specifying the `out` key (as demonstrated above). **This might suggest that we use the shorthand name by default, and only instruct users to override it if they have conflicts.**

2. Should we instead use the `remote` key? In the current proposal, we now overload the `name` key so that it can either be a local binary (discoverable on the user's `PATH`), or a fully-qualified `bufpluginref.PluginReference`. We previously used the separate `remote` key to distinguish the two, but we can easily do this by adapting the `name` like so. In this case, we are effectively deprecating the `remote` key and pushing users towards using the `name` key in general, but we might actually want to continue leaning into the `remote` key for clarity.

3. This proposes that we continue to use the `buf.alpha.registrypublic.v1alpha1.PluginService/GeneratePlugins` endpoint to generate code for the plugins defined by a `buf.plugin.yaml`. Looking at the interface we have today, we should be able to do so - it will just require changes on the server-side to account for the `buf.plugin.yaml`, but we have all the parameters we need in its current state. See [this](https://github.com/bufbuild/buf/blob/8d2cea2f9d911a7e5355a3187e598384af78e860/proto/buf/alpha/registry/v1alpha1/generate.proto#L71) for details.

TL;DR Putting this up early so that we have an interface to experiment with for plugins produced with a `buf.plugin.yaml`. There's a few open questions, but this will unlock other development/experimentation for now.

Note that we **should not** merge this until we're ready for it to be in our users' hands. In other words, we need to be mindful of compatibility here.

### DIFF
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -25,13 +25,16 @@ import (
 
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
+	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginref"
 	"github.com/bufbuild/buf/private/gen/proto/apiclient/buf/alpha/registry/v1alpha1/registryv1alpha1apiclient"
 	"github.com/bufbuild/buf/private/pkg/app"
 	"github.com/bufbuild/buf/private/pkg/command"
+	"github.com/bufbuild/buf/private/pkg/encoding"
 	"github.com/bufbuild/buf/private/pkg/storage"
 	"github.com/bufbuild/buf/private/pkg/storage/storageos"
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/types/descriptorpb"
+	"gopkg.in/yaml.v3"
 )
 
 const (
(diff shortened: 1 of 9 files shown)


amckinney:
Closing this for now - we'll want to start with a simpler way to invoke plugins defined by a `buf.plugin.yaml` (likely just allowing the `remote` key to recognize plugins as `<remote>/<owner>/<plugin>` (omitting the `/plugins/` path element).

PR #1245, opened 2022-06-28
add buf alpha plugin push

pkwarren:
Add back the plugin push command under 'alpha'. Pass .netrc
authentication when publishing Docker images. Add additional test
coverage to Docker client simulating issues seen while testing locally
(in particular, errors returned via response.Body instead of via an
error in ImagePush).


private/buf/cmd/buf/buf.go
```diff
@@ -238,6 +239,13 @@ func NewRootCommand(name string) *appcmd.Command {
 							},
 						},
 					},
+					{
+						Use:   "plugin",
```
pkwarren, line 243: Re-adds the `plugin push` command under alpha.

private/buf/cmd/buf/command/beta/plugin/pluginpush/pluginpush.go
```diff
@@ -144,8 +145,17 @@ func run(
 		}
 	}()
 
-	// TODO: Implement authentication to BSR registry using ~/.netrc
-	_, err = client.Push(ctx, buildResponse.Image, &bufplugindocker.RegistryAuthConfig{})
+	machine, err := netrc.GetMachineForName(container, pluginConfig.Name.Remote())
```
pkwarren, line 148: Tested locally by adding a plugins entry to ~/.netrc.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -135,12 +137,13 @@ func (d *dockerAPIClient) Build(ctx context.Context, dockerfile io.Reader, plugi
...
 		}()
-		// TODO: We're just logging the messages from the image build.
-		// In a future PR, we might want to add support for parsing well known messages (i.e. the image id).
-		// This would allow us to potentially skip the ImageInspectWithRaw client call below.
 		scanner := bufio.NewScanner(response.Body)
 		for scanner.Scan() {
 			d.logger.Debug(scanner.Text())
+			var message jsonmessage.JSONMessage
+			if err := json.Unmarshal([]byte(scanner.Text()), &message); err == nil && message.Error != nil {
+				return message.Error
```
pkwarren, line 145: While testing, I found that both the build and push commands might return a `nil` error but then communicate an error via the streaming response body. This adds handling to return an error.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -215,6 +224,20 @@ func NewClient(logger *zap.Logger) (Client, error) {
 	}, nil
 }
 
+type ClientOption func(options []client.Opt) []client.Opt
+
+func WithHost(host string) ClientOption {
```
pkwarren, line 229: In order to test against a httptest.Server Docker engine, needed to expose this option for overriding the default host. Alternatively, I could use `DOCKER_HOST` env var but I didn't think that would work reliably with parallel tests.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -65,10 +77,83 @@ func TestBuildFailure(t *testing.T) {
...
+	if !dockerEnabled {
+		t.Skip("docker disabled")
+	}
+	if testing.Short() {
+		t.Skip("docker tests disabled in short mode")
+	}
+	dockerVersion := "1.41"
+	server := newDockerServer(t, dockerVersion)
+	// Send back an error on ImagePush (still return 200 OK).
+	server.pushErr = errors.New("failed to push image")
```
pkwarren, line 119: Here we simulate an error during push which is sent back on the response body.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -65,10 +77,83 @@ func TestBuildFailure(t *testing.T) {
...
+	if !dockerEnabled {
+		t.Skip("docker disabled")
+	}
+	if testing.Short() {
+		t.Skip("docker tests disabled in short mode")
+	}
+	dockerVersion := "1.41"
+	server := newDockerServer(t, dockerVersion)
+	// Send back an error on ImagePush (still return 200 OK).
+	server.buildErr = errors.New("failed to build image")
```
pkwarren, line 145: Here we simulate an error during build which is sent back on the response body.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +204,136 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+func (d *dockerServer) sessionHandler(w http.ResponseWriter, r *http.Request) {
+	if _, err := io.Copy(io.Discard, r.Body); err != nil {
+		d.t.Error("failed to discard body:", err)
+	}
+	if r.Method != http.MethodPost {
+		http.NotFound(w, r)
+		return
+	}
+	if strings.EqualFold(r.Header.Get("Connection"), "upgrade") && r.ProtoMajor < 2 {
+		w.WriteHeader(http.StatusSwitchingProtocols)
```
pkwarren, line 244: This is needed to allow h2c upgrades to proceed.

private/bufpkg/bufplugin/bufplugindocker/docker.go
amckinney: Sounds good, let's leave a comment explaining that. It's OK to keep this around for testing as long as it's clear.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -215,6 +224,20 @@ func NewClient(logger *zap.Logger) (Client, error) {
 	}, nil
 }
 
+type ClientOption func(options []client.Opt) []client.Opt
```
amckinney, line 227: In general, I like what you're doing here - the `ClientOption` interface it's a thin abstraction around the `client.Opt` type.

However, we often explicitly map these options into each other and assign them wherever it's necessary. In this case, I'd expect to see an un-exported `clientOptions` struct that has the same values you have here (i.e. `clientOptions.host` and `clientOptions.version`). Then in the `NewClient` constructor, you would manually assign each of these values to the `client.Opt`:

```go
opts := &clientOptions{}
for _, option := range options {
	opts = option(opts)
}
dockerClientOpts := []client.Opt{client.FromEnv}
if opts.host != "" {
  dockerClientOpts = append(dockerClientOpts, client.WithHost(opts.host))
}
if opts.version != "" {
  dockerClientOpts = append(dockerClientOpts, client.WithVersion(opts.version))
}
```

Although more verbose, this means the `ClientOption` is able to modify more than just the underlying docker client - it could modify other aspects of the `dockerAPIClient` because it's not tied to the `client.Opt` type.

Your current solution is probably fine as-is, but I'd like us to stay consistent with other places for now and decouple this a bit.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -215,6 +224,20 @@ func NewClient(logger *zap.Logger) (Client, error) {
...
+
+func WithHost(host string) ClientOption {
+	return func(options []client.Opt) []client.Opt {
+		return append(options, client.WithHost(host))
+	}
+}
+
+func WithVersion(version string) ClientOption {
```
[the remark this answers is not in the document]
amckinney: Sweet, let's write it down! :D

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
amckinney: Nice, let's write that down too.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +186,136 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+	d.builtImages[imageID] = &builtImage{tags: r.URL.Query()["t"]}
+}
+
+func (d *dockerServer) imagesHandler(w http.ResponseWriter, r *http.Request) {
+	if _, err := io.Copy(io.Discard, r.Body); err != nil {
+		d.t.Error("failed to discard body:", err)
+	}
+	d.t.Logf("%s %s", r.Method, r.URL.Path)
+	pathSuffix := strings.TrimPrefix(r.URL.Path, "/v1.41/images/")
+	imagePattern := regexp.MustCompile("^(?P<image>[^/]+/[^/]+/[^/:]+)(?::(?P<tag>[^/]+))?(?:/(?P<op>[^/]+))?$")
```
amckinney, line 253: We might want to move this out into a top-level `var` so that it doesn't need to be recompiled every time `imagesHandler` is run.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +186,136 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+	imageID := stringid.GenerateRandomID()
+	d.builtImages[imageID] = &builtImage{tags: r.URL.Query()["t"]}
+}
+
+func (d *dockerServer) imagesHandler(w http.ResponseWriter, r *http.Request) {
+	if _, err := io.Copy(io.Discard, r.Body); err != nil {
+		d.t.Error("failed to discard body:", err)
+	}
+	d.t.Logf("%s %s", r.Method, r.URL.Path)
+	pathSuffix := strings.TrimPrefix(r.URL.Path, "/v1.41/images/")
```
amckinney, line 252: This `/v1.41/images/` might be something we declare as a `const` too - that way it has some more self-documenting semantic meaning.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +186,136 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+	w.WriteHeader(http.StatusOK)
+	if d.pushErr != nil {
+		d.writeError(w, d.pushErr)
+		return
+	}
+}
+
+func (d *dockerServer) findImageIDFromName(name string) string {
+	for imageID, builtImageInfo := range d.builtImages {
+		for _, imageTag := range builtImageInfo.tags {
```
amckinney, line 305: Should we make the `builtImageInfo.tags` a `map[string]struct{}` so that we can check for existence in O(1) instead of O(n)?

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +186,136 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+		d.mu.RLock()
+		defer d.mu.RUnlock()
+		foundImageID := d.findImageIDFromName(image + ":" + tag)
+		if len(foundImageID) == 0 {
+			http.NotFound(w, r)
+			return
+		}
+		if err := json.NewEncoder(w).Encode(&types.ImageInspect{
+			ID:       "sha256:" + foundImageID,
+			RepoTags: d.builtImages[foundImageID].tags,
```
amckinney, line 271: If we decide to use a `map[string]struct{}` instead, we can easily transform it into a slice with our `stringutil` package.

amckinney (commented):
This looks great, just a few minor things.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
pkwarren: This is only used for testing and will contain a max of one image / one tag (using existing test cases). I didn't want to over complicate the design if we could simplify the in-memory storage.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -119,3 +186,142 @@ func buildDockerPlugin(t testing.TB, dockerClient bufplugindocker.Client, docker
...
+		d.writeError(w, d.buildErr)
+		return
+	}
+	d.mu.Lock()
+	defer d.mu.Unlock()
+	imageID := stringid.GenerateRandomID()
+	d.builtImages[imageID] = &builtImage{tags: r.URL.Query()["t"]}
+}
+
+var imagePattern = regexp.MustCompile("^(?P<image>[^/]+/[^/]+/[^/:]+)(?::(?P<tag>[^/]+))?(?:/(?P<op>[^/]+))?$")
```
amckinney, line 253: Let's declare this at the top of the file - all global, un-exported `const` and `var` declarations should be at the top.

amckinney (approved):
Nice, just a few remaining nits. Please address them, but feel free to merge after!

PR #1249, opened 2022-06-29
[BSR-304] Webhook payload definition

unmultimedio:
Payload definition for the webhook events to be sent from the BSR to the customer's backend.


proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
```diff
@@ -0,0 +1,52 @@
...
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+import "buf/alpha/registry/v1alpha1/webhook.proto";
+import "google/protobuf/timestamp.proto";
+
+// WebhookListenerService represents a service implemented by a customer, to
+// receive webhook events invocations from the BSR.
+service WebhookListenerService {
```
buildbreaker, line 24: WebhookReceiverService?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
```diff
@@ -0,0 +1,52 @@
...
+  // The timestamp of when the event happened.
+  google.protobuf.Timestamp event_time = 1;
+  // The event that happened.
+  buf.alpha.registry.v1alpha1.WebhookEvent event = 2;
+  // The repository id tied to the event.
+  string repository_id = 3;
+  // Te repository commit id tied to the event.
+  string repository_commit_id = 4;
+  // The user id of the author who triggered the event.
+  string author_user_id = 5;
```
buildbreaker, line 51: I'm thinking maybe the names would be more helpful here than the ids since the user wouldn't have a way to look up these ids themselves

unmultimedio: In the context of an UI-oriented use case, sure, human-readable attributes are much more helpful (so they're rendered and seen by a human). But webhooks will be consumed by their backend service, and most likely they already have these IDs stored. Repo names and owner names can change, so that also leave us with possible race conditions. In general terms, it's much easier to deal with IDs for relational objects, those will never change.

unmultimedio: Another option is to inject the complete commit object, something like:

```proto
message WebhookPayload {
  google.protobuf.Timestamp event_time = 1;
  buf.alpha.registry.v1alpha1.WebhookEvent event = 2;
  RepositoryCommit repository_commit = 3; // which would include already the repository and author ids
  Repository repository = 4; // an overkill in my opinion, the event relates to the commit, not to the repository
}
```

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: Sure why not, although I don't see a big difference vs "Listen". What's the upside?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
```diff
@@ -0,0 +1,52 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
```
robbertvanginkel, line 17: This doesn't belong in registry, registry so far is the services that make up a buf schema registry. This is something a customer would implement, it should probably go into its own package?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
robbertvanginkel: What is the expected use of the service definition? Is it so that users can codegen a handler that they give to connect?

In that case, am I expected to provide a `callback_url` that is `https://acme.corp.com/webhooks` and we'll append the `buf.alpha.registry.v1alpha1.WebhookListenerService/ListenWebhook` when making the outgoing request, or should the `callback_url` be `https://acme.corp.com/webhooks/buf.alpha.registry.v1alpha1.WebhookListenerService/ListenWebhook`?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
```diff
@@ -0,0 +1,52 @@
...
+message ListenWebhookResponse {}
+
+// WebhookPayload defines the structure of an actual webhook invocation.
+message WebhookPayload {
+  // The timestamp of when the event happened.
+  google.protobuf.Timestamp event_time = 1;
+  // The event that happened.
+  buf.alpha.registry.v1alpha1.WebhookEvent event = 2;
+  // The repository id tied to the event.
+  string repository_id = 3;
```
bufdev, line 47: What if a webhook event isn't tied to a repository?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
```diff
@@ -0,0 +1,52 @@
...
+  // Te repository commit id tied to the event.
+  string repository_commit_id = 4;
```
bufdev, line 49: Same question, what if not related to a commit?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: So far it's designed to "belong" to a repository, so it'll be always tied to one. When you create a webhook, the owner/repo names are required params.

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
robbertvanginkel: Including more information seems like the right direction. Looking at something like https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads, the response payloads seem quite detailed. The two things that I can come up with why that seems desirable:

- If we emit webhooks for an event like "repo rename", you'd probably want to include details like old/new name.
- If in response to receiving a webhook users would need to send `n` more requests to have a picture of what happened, besides the ux being bad there is some potential for load amplification when pushing out webhooks (not sure if that will really be a concern, the ux argument is probably stronger).

How detailed are the payloads for other popular webhooks (slack, github, others)?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
buildbreaker: The current intention is for it to be `https://acme.corp.com/webhooks`

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: True! and that makes me think, we should have a different RPC (with different payloads), per event type, something like `ListenRepositoryPushEvent`. So we can customize how/which data we want to send to the customer, regardless of the event.

Downside of this proposal, is that if the customer wants to subscribe to just 1 event, but we support 20, they'll endup with 20 autogenerated functions that do nothing. Doesn't seem too bad imo, depending on their language they can just delete them.

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
buildbreaker: Let's make them WKTs!

buildbreaker: Spoke with @unmultimedio about how Github handles things -- they special case their push events. A direction that we can go in is having a separate endpoint and payload for non-push events in Buf and cross that decision when we have it

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
robbertvanginkel: I think that webhook receiving should be possible in languages other than go, given that we have no connect in those other languages it does feel weird that we would enforce `pkg.Service/Method` be part of the url if I'm just writing some python/java/js handler that does the equivalent of `proto.Unmarshal(io.Readall(request.Body()), m)`.

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
robbertvanginkel: `buf.alpha.registry.v1alpha1.webhook`? Package level separation seems nice here so that if we ever have other interfaces for the user to implement that we call out to, they won't be forced to drag in the webhook types if they don't need to.

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
buildbreaker: They can still implement the gRPC service handler if they're not using connect!

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
robbertvanginkel: Having a toplevel `oneof` in the webhook payload to separate out diffent events as compared to having different service names feels closer to what I'm reading in the github docs on webhooks. Related to the pkg in the url discussion, I don't think it'd be great if we required users to expose a new handler/path for each event type they'd subscribe on.

We had something similar to that in the now defunct audit log structure:

https://github.com/bufbuild/buf/blob/14ed91784fcd5cd06763c420ce6d765092571430/proto/buf/alpha/audit/v1alpha1/envelope.proto#L579-L583

And I've seen it before in bazel as well: https://github.com/bazelbuild/bazel/blob/2594ea48f65008e81d29bbf77b5798f4c43fc454/src/main/java/com/google/devtools/build/lib/buildeventstream/proto/build_event_stream.proto#L1119-L1122

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
cyinma: The last component of a package should be a version as per [the style guide](https://docs.buf.build/best-practices/style-guide#files-and-packages), might be better to just separate it from `registry` and be `buf.alpha.webhook.v1alpha1`?

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: Done

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: Resolving, since the payload is now specific per event.

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: Resolving, since the payload is now specific per event.

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
```diff
@@ -0,0 +1,45 @@
...
+
+// EventRequest is the request payload that will be sent to the customer that is
+// subscribed to webhook events in the BSR. This request message serves both as
+// (1) documentation of the payload that will be sent in a JSON blob, or (2)
+// using this in an rpc signature in your own proto file.
+message EventRequest {
+  // The event that was triggered.
+  buf.alpha.registry.v1alpha1.WebhookEvent event = 1;
+  // The event payload depending on which event was triggered.
+  oneof payload {
```
buildbreaker, line 32:
```suggestion
  oneof event {
```?

File: proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto [hunk 7]
```diff
@@ -0,0 +1,45 @@
...
+import "buf/alpha/registry/v1alpha1/webhook.proto";
+import "google/protobuf/timestamp.proto";
+
+// EventRequest is the request payload that will be sent to the customer that is
+// subscribed to webhook events in the BSR. This request message serves both as
+// (1) documentation of the payload that will be sent in a JSON blob, or (2)
+// using this in an rpc signature in your own proto file.
+message EventRequest {
+  // The event that was triggered.
+  buf.alpha.registry.v1alpha1.WebhookEvent event = 1;
```
buildbreaker, line 30: I am thinking that this maybe having this would be redundant now since we have the oneof

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
unmultimedio: It collides with the previous property name

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
unmultimedio: I was doubting here as well, but trying to follow [this comment](https://github.com/bufbuild/buf/pull/1249#discussion_r911323627), it includes both, the action and the action payload. It was either including it here, including it on every payload of every event type, or not including it at all. It felt right to do it here.

Also, depending on the code generation language, I assume it'll be easier for the customer to switch on this value before reading the payload, but that's just an assumption.

buildbreaker: Ah, that makes sense. Fair enough!

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
buildbreaker: 😢

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
```diff
@@ -0,0 +1,45 @@
...
+import "buf/alpha/registry/v1alpha1/repository.proto";
+import "buf/alpha/registry/v1alpha1/repository_commit.proto";
+import "buf/alpha/registry/v1alpha1/webhook.proto";
+import "google/protobuf/timestamp.proto";
+
+// EventRequest is the request payload that will be sent to the customer that is
+// subscribed to webhook events in the BSR. This request message serves both as
+// (1) documentation of the payload that will be sent in a JSON blob, or (2)
+// using this in an rpc signature in your own proto file.
+message EventRequest {
```
bufdev, line 28: Is this a request for an `rpc` named `Event`? If not, it probably should not be named `*Request`, as this is reserved for requests and responses for `rpcs`.

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
bufdev, line 30: Why is this not in the `registry.webhook.v1alpha1` package, ie why is `WebhookEvent` in a different package than `EventRequest`?

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
unmultimedio: Yes, this is the result as I understood it from the meeting with @buildbreaker @akshayjshah @lrewega 

It's intended to be used by us with the connect-go client library as a request object to invoke customer's callback URL. See the proof of concept here: https://github.com/bufbuild/core/pull/5045/files#diff-4f20eea6e0f53b4e36eb4c80d45f876f4bac9c2660d15c358217bbe0b21eb27bR23

But we didn't strictly define the `rpc` since we're not forcing the service definition to customers. If they are already in protobuf, then they can define their own `rpc` using this message as request, if not, this same structure will be sent [as JSON](https://github.com/bufbuild/core/pull/5045/files#diff-4f20eea6e0f53b4e36eb4c80d45f876f4bac9c2660d15c358217bbe0b21eb27bR27).

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
unmultimedio: The event types were defined for internal usage [when registering a webhook tied to an event](https://github.com/bufbuild/buf/blob/5211c6467af3ee6c01b3dc20e2c13e8b415c0f3e/proto/buf/alpha/registry/v1alpha1/webhook.proto#L36).

This PR originally included the external RPC that would be implemented by customers, so I got [the suggestion](https://github.com/bufbuild/buf/pull/1249#discussion_r911236908) to change it to another package because this is something for the customer to implement, not something that belongs to the BSR. Since still it's something related to the BSR usage, I proposed making it a `registry` subpackage.

But now, since we're not defining any `rpc` here, I don't see it wrong to make it live within the original `registry` package. @robbertvanginkel wdyt?

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
```diff
@@ -0,0 +1,45 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.webhook.v1alpha1;
```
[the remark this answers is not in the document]
bufdev: See above

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
bufdev: Following those comments, I don't think they explain why `WebhookEvent` (which is not a request or response for anything in `registry`) would be in a different package than `EventRequest`.

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
bufdev: I'd like to understand any design decision here, if there was one, it should be recorded in the Webhook design doc. From my perspective, I'm not sure why we wouldn't just tell users to implement a Protobuf service - providing flexibility beyond that is out of scope for this project

buildbreaker: Yeah, @unmultimedio and I originally proposed that the user implements the Protobuf service. After chatting with @akshayjshah and @lrewega it felt that our appetite for forcing a user to implement a Protobuf service was a bit too high

proto/buf/alpha/registry/webhook/v1alpha1/event_payloads.proto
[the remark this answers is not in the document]
unmultimedio: Moved back to parent pkg.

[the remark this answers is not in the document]
bufdev: I wasn't necessarily implying parent package, I meant that we don't have parent/child packages, ie the current standard would say that if you are creating a new package, it would be `buf.alpha.webhook.v1alpha1`.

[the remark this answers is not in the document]
unmultimedio: But if we do that, then `WebhookEvent` and `WebhookRequest` would live in 2 different packages, since `WebhookEvent` is used when subscribing to an event.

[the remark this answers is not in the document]
bufdev: Or they would both live in a `buf.alpha.webhook.v1alpha1` package, alongside the service users are expected to implement, which is what I was hinting at (but could have been more explicit), also re: https://github.com/bufbuild/buf/pull/1249#discussion_r911236908=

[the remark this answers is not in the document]
unmultimedio: We could move everything webhook-related to a separate `webhook` package, but I think it belongs more to `registry`:

- Create/Delete/List webhooks is something that's implemented by the BSR UI
- The BSR implements the client stub of the webhook event service (customer would be implementing the server stub)
- The payload of the event contains (and will contain as we keep adding more events) several nested messages that already live in the registry package (repository, commits, users...)

If we separate to a different package, we could still use/import those messages from the registry, but I think it'd be better to declare a new "pubic" repository message, to be embedded in the webhook payload message.

proto/buf/alpha/webhook/v1alpha1/event.proto
```diff
@@ -0,0 +1,59 @@
...
+
+package buf.alpha.webhook.v1alpha1;
+
+import "buf/alpha/registry/v1alpha1/repository.proto";
+import "buf/alpha/registry/v1alpha1/repository_commit.proto";
+import "buf/alpha/registry/v1alpha1/webhook.proto";
+import "google/protobuf/timestamp.proto";
+
+// EventService is the service which receives webhook events from Buf.
+service EventService {
```
unmultimedio, line 25: I went for [no-stutter](https://sben.dev/stutter/) when naming things in this proto, so we wouldn't have `webhook.WebhookEventService`, but `webhook.EventService` in the generated go code.

But, this isn't go code but proto code. Is it fine like this? or should I include the `Webhook` prefix everywhere?

buildbreaker: I am a fan of removing the redundancies in naming

buildbreaker (commented):
Looks good @unmultimedio. @bufdev, could you take a final pass here when you get a chance?

proto/buf/alpha/webhook/v1alpha1/event.proto
```diff
@@ -0,0 +1,59 @@
...
+  // Event is the rpc which receives webhook events.
+  rpc Event(EventRequest) returns (EventResponse);
```
robbertvanginkel, line 27: Now that the service is back https://github.com/bufbuild/buf/pull/1249#discussion_r911239446 is still somewhat of an open question to me. We should document how `callback_url` will work, but that probably doesn't go into this service definition but in the registry request where you set it.

proto/buf/alpha/webhook/v1alpha1/event.proto
robbertvanginkel: The registry services also do not start with `...registry.v1alpha1.Registry...Service/`, so this seems good to me too!

proto/buf/alpha/webhook/v1alpha1/event.proto
buildbreaker: Sounds good. We'll do it here: https://github.com/bufbuild/buf/pull/1264

proto/buf/alpha/registry/v1alpha1/webhook_payload.proto
unmultimedio: Done, new `webhook` package.

bufdev (approved):

robbertvanginkel (approved):

PR #1250, opened 2022-06-29
fix connection upgrade for Docker tests

pkwarren:
Currently, the Docker tests are occassionally hitting a 10s delay on
shutdown when closing the buildkit session. This appears to be due to
incorrect logic when upgrading the connection from HTTP/1.1 to HTTP/2.
The Docker client API doesn't send the required HTTP2-Settings header
and so the h2c handler doesn't upgrade the connection properly.
    
Update the Docker engine implementation for tests to correctly upgrade
to a H2C connection, fixing the delay on close.
    
Update some additional tests in this package to support parallel
execution.

### DIFF
--- a/private/bufpkg/bufplugin/bufplugindocker/docker_test.go
+++ b/private/bufpkg/bufplugin/bufplugindocker/docker_test.go
@@ -15,10 +15,13 @@
 package bufplugindocker
 
 import (
+	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"io"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -216,27 +222,93 @@ func newDockerServer(t testing.TB, version string) *dockerServer {
...
 
 func (d *dockerServer) sessionHandler(w http.ResponseWriter, r *http.Request) {
-	if _, err := io.Copy(io.Discard, r.Body); err != nil {
-		d.t.Error("failed to discard body:", err)
-	}
 	if r.Method != http.MethodPost {
 		http.NotFound(w, r)
 		return
 	}
 	if strings.EqualFold(r.Header.Get("Connection"), "upgrade") && r.ProtoMajor < 2 {
```
pkwarren, line 238: The Docker client isn't sending a required HTTP2-Settings header so the h2c implementation isn't working out of the box: https://cs.opensource.google/go/x/net/+/1bab6f36:http2/h2c/h2c.go;l=177;drc=1bab6f366d9ec8c152f8735b2e359822ef75e3bd

mfridman (approved):

PR #1253, opened 2022-06-30
Update CuratedPlugin revision to be a signed integer

smallsamantha:


pkwarren (approved):

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -88,7 +88,7 @@ message CuratedPlugin {
   // The configuration for the runtime of the plugin.
   RuntimeConfig runtime_config = 11;
   // The revision for this plugin version.
-  uint32 revision = 12;
+  int32 revision = 12;
```
bufdev, line 91: why?

smallsamantha: to minimize conversions in our business logic. postgres doesnt support unsigned int type.

bufdev: This is changing our API based on an implementation detail - our API shape should not be dictated by how we happen to implement it. A revision can never be less than 0, so this should be unsigned.

PR #1254, opened 2022-06-30
update plugin push to create curated plugin

pkwarren:
Update the `buf alpha plugin push` command to call CreateCuratedPlugin
to add the plugin to the BSR.

### DIFF
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -29,7 +29,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
-	connect_go "github.com/bufbuild/connect-go"
+	"github.com/bufbuild/connect-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/multierr"
@@ -179,9 +179,10 @@ func run(
 		return err
 	}
 	var nextRevision int32
+	// TODO: Revisit if we decide to make revision part of plugin_version
 	currentRevision, err := service.GetLatestCuratedPlugin(ctx, pluginConfig.Name.Owner(), pluginConfig.Name.Plugin(), pluginConfig.PluginVersion)
 	if err != nil {
-		if connect_go.CodeOf(err) != connect_go.CodeNotFound {
+		if connect.CodeOf(err) != connect.CodeNotFound {
 			return err
 		}
 		nextRevision = 0
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -22,13 +22,14 @@ import (
 
 	"github.com/bufbuild/buf/private/buf/bufcli"
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
+	"github.com/bufbuild/buf/private/bufpkg/bufplugin"
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufplugindocker"
-	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
+	connect_go "github.com/bufbuild/connect-go"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/multierr"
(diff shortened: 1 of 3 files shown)


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -160,57 +161,52 @@ func run(
...
 		return err
 	}
-	// TODO: Refactor this to be a new endpoint that uses the protoPlugin
-	// object. This is simply left as a placeholder.
-	if _, err := service.CreatePlugin(
+	var nextRevision int32
+	currentRevision, err := service.GetLatestCuratedPlugin(ctx, pluginConfig.Name.Owner(), pluginConfig.Name.Plugin(), pluginConfig.PluginVersion)
+	if err != nil {
+		if connect_go.CodeOf(err) != connect_go.CodeNotFound {
+			return err
```
pkwarren, line 185: This is currently unimplemented, so local testing is returning an error here.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -160,57 +161,52 @@ func run(
...
+		}
+		nextRevision = 0
+	} else {
+		nextRevision = currentRevision.Revision + 1
+		if nextRevision < 0 {
+			return fmt.Errorf("next plugin revision out of range")
```
pkwarren, line 191: Probably overthinking this (and we can also validate we don't wrap over on the server side (no negative revisions).

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -160,57 +161,52 @@ func run(
...
 		pluginConfig.Name.Owner(),
 		pluginConfig.Name.Plugin(),
-		registryv1alpha1.PluginVisibility_PLUGIN_VISIBILITY_PUBLIC,
+		bufplugin.PluginToProtoPluginLanguage(plugin),
+		plugin.Version(),
+		plugin.ContainerImageDigest(),
+		bufplugin.PluginOptionsToOptionsSlice(plugin.Options()),
+		nil, // dependencies
+		"",  // sourceUrl
+		"",  // description
```
pkwarren, line 204: These don't currently have a representation in the yaml configuration.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -160,57 +161,52 @@ func run(
...
+		bufplugin.PluginRuntimeToProtoRuntimeConfig(plugin.Runtime()),
+		nextRevision,
 	); err != nil {
 		return err
 	}
-	// TODO: Print out the plugin that was just created.
+	// TODO: Determine how to print curated plugin - see bufprint
```
pkwarren, line 210: Need to work through which fields to print from the returned plugin. I'm assuming we'll have either text/JSON output to stdout similar to how some of the existing CLI commands work.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -22,13 +22,14 @@ import (
...
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
+	"github.com/bufbuild/buf/private/bufpkg/bufplugin"
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufplugindocker"
-	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
+	connect_go "github.com/bufbuild/connect-go"
```
amckinney, line 32: I would just alias this as `connect` so that it matches the real package name.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
amckinney: As discussed offline, we might even want to require that the revision is part of the `plugin_version` so that the client doesn't need to perform the `GetLatestCuratedPlugin` call.

This is fine for now though, especially because its already written in this state - we can start with this and change it later if need be.

Maybe just add in a `TODO` here that outlines some of this comment so that we remember to revisit it.

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -60,3 +63,122 @@ func NewPlugin(
 ) (Plugin, error) {
 	return newPlugin(version, options, runtimeConfig, imageDigest)
 }
+
+func PluginToProtoPluginLanguage(plugin Plugin) registryv1alpha1.PluginLanguage {
```
amckinney, line 67: Don't forget the comments on all these exported functions!

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -60,3 +63,122 @@ func NewPlugin(
...
+}
+
+func ProtoRuntimeConfigToPluginRuntime(config *registryv1alpha1.RuntimeConfig) *bufpluginconfig.RuntimeConfig {
+	if config == nil {
+		return nil
+	}
+	runtimeConfig := &bufpluginconfig.RuntimeConfig{}
+	if config.GetGoConfig() != nil {
+		goConfig := &bufpluginconfig.GoRuntimeConfig{}
+		goConfig.MinVersion = config.GetGoConfig().MinimumVersion
```
amckinney, line 110: Although (mostly) equivalent, we might want to use the `Get` helper on each of these secondary accessors (this applies to the `.RuntimeLibraries` calls below, too.
```suggestion
		goConfig.MinVersion = config.GetGoConfig().GetMinimumVersion()
```

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -60,3 +63,122 @@ func NewPlugin(
...
+}
+
+func NPMRuntimeDependencyToProtoNPMRuntimeLibrary(config *bufpluginconfig.NPMRuntimeDependencyConfig) *registryv1alpha1.NPMConfig_RuntimeLibrary {
+	return &registryv1alpha1.NPMConfig_RuntimeLibrary{
+		Package: config.Package,
+		Version: config.Version,
+	}
+}
+
+func ProtoNPMRuntimeLibraryToNPMRuntimeDependency(config *registryv1alpha1.NPMConfig_RuntimeLibrary) *bufpluginconfig.NPMRuntimeDependencyConfig {
```
amckinney, line 148: It looks like we're only using this in this package - let's make sure to un-export all the functions that don't need to be exported. This will apply to a variety of other functions in this file.

amckinney (approved):
Nice, this looks great! Mostly nits, so I'm gonna stamp as-is and trust that you'll address them before merging.

PR #1261, opened 2022-07-04
Support protocol and encoding client options based on content-type in Studio Agent

doriable:
Currently we are defaulting to the gRPC protocol in Studio Agent,
however, we are able to support `application/json` and `application/proto`
with the Connect protocol. This just refactors the client options so that
we are setting protocol and codec options based on the content-type.

### DIFF
--- a/private/bufpkg/bufstudioagent/bufstudioagent_test.go
+++ b/private/bufpkg/bufstudioagent/bufstudioagent_test.go
@@ -68,34 +68,65 @@ func testPlainPostHandler(t *testing.T, upstreamServer *httptest.Server) {
 	)
 	defer agentServer.Close()
 
-	requestProto := &studiov1alpha1.InvokeRequest{
-		Target: upstreamServer.URL + echoPath,
-		Headers: goHeadersToProtoHeaders(http.Header{
-			"Content-Type": []string{"application/grpc+proto"},
-		}),
-		Body: []byte("echothis"),
-	}
-	requestBytes := protoMarshalBase64(t, requestProto)
-	request, err := http.NewRequest(http.MethodPost, agentServer.URL, bytes.NewReader(requestBytes))
-	require.NoError(t, err)
-	request.Header.Set("Content-Type", "text/plain")
-	request.Header.Set("Origin", "https://example.buf.build")
-	request.Header.Set("Foo", "foo-value")
-	response, err := agentServer.Client().Do(request)
-	require.NoError(t, err)
+	t.Run("content_type_grpc_proto", func(t *testing.T) {
+		requestProto := &studiov1alpha1.InvokeRequest{
+			Target: upstreamServer.URL + echoPath,
+			Headers: goHeadersToProtoHeaders(http.Header{
+				"Content-Type": []string{"application/grpc+proto"},
+			}),
+			Body: []byte("echothis"),
+		}
+		requestBytes := protoMarshalBase64(t, requestProto)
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufstudioagent/plain_post_handler.go
```diff
@@ -222,6 +203,43 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func connectClientOptionsFromContentType(contentType string, requestMsgSize int) ([]connect.ClientOption, error) {
+	switch contentType {
+	case "application/grpc", "application/grpc+proto":
+		return []connect.ClientOption{
+			connect.WithGRPC(),
+			connect.WithCodec(&bufferCodec{name: "proto"}),
+		}, nil
+	case "application/grpc-json":
```
bufdev, line 213: I've never seen this before, shouldn't this be `application/grpc+json`?

doriable: Uh, yes, it should be 😬

doriable:
Removed the empty content-type case with an empty message, since defaulting to gRPC there seems like it would be hard to debug. This way, we should always expect client to provide content-type, which is reasonable since we control the client-side interaction via Studio.

private/bufpkg/bufstudioagent/plain_post_handler.go
```diff
@@ -222,6 +203,31 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	})
 }
 
+func connectClientOptionsFromContentType(contentType string, requestMsgSize int) ([]connect.ClientOption, error) {
```
robbertvanginkel, line 206: No need for `requestMsgSize int` anymore.

private/bufpkg/bufstudioagent/bufstudioagent_test.go
```diff
@@ -68,34 +68,65 @@ func testPlainPostHandler(t *testing.T, upstreamServer *httptest.Server) {
...
+		assert.Equal(t, http.StatusOK, response.StatusCode)
+		assert.Equal(t, "https://example.buf.build", response.Header.Get("Access-Control-Allow-Origin"))
+		responseBytes, err := io.ReadAll(response.Body)
+		assert.NoError(t, err)
+		invokeResponse := &studiov1alpha1.InvokeResponse{}
+		protoUnmarshalBase64(t, responseBytes, invokeResponse)
+		upstreamResponseHeaders := make(http.Header)
+		addProtoHeadersToGoHeader(invokeResponse.Headers, upstreamResponseHeaders)
+		addProtoHeadersToGoHeader(invokeResponse.Trailers, upstreamResponseHeaders)
+		assert.Equal(t, []byte("echo: echothis"), invokeResponse.Body)
```
robbertvanginkel, line 128: Could add something like `assert.Equal(t, "", upstreamResponseHeaders.Get("grpc-status"))` to validate the response wasnt a grpc response?

private/bufpkg/bufstudioagent/plain_post_handler.go
doriable: 😬 🤦🏼‍♀️

private/bufpkg/bufstudioagent/bufstudioagent_test.go
doriable: Ah yeh, this is a good idea -- added some new assertions.

robbertvanginkel (approved):

PR #1262, opened 2022-07-05
update http2client to respect proxy from environment

### LINKED ISSUE #1184: Buf CLI does not respect HTTPS_PROXY variable
Before buf v1.5.0, `buf` respected the HTTPS_PROXY environment variable for outgoing requests. After the switch to Connect transport protocol (https://github.com/bufbuild/buf/pull/1129) it does not use the configured HTTP proxy.

rhbuf:
I didn't see any easy solutions beyond copying what grpc-go has done (minus grpc bits), the only other idea I saw was to use `http2.ConfigureTransports` to configure a http transport with `Proxy: http.ProxyFromEnvironment` for HTTP2, but I only got a slew of errors about cached connections

Fixes #1184


private/pkg/transport/http2client/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+	"encoding/base64"
+	"fmt"
+	"io"
+	"net"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+)
+
+// This file was adapted from https://github.com/grpc/grpc-go/blob/master/internal/transport/proxy.go
```
rubensf, line 28: This looks good to me in theory, but we should sanity check with responsible people on what conditions we need to meet - we're already using APACHE license on buf, but do we need to eg use the same license header from grpc-go?

rubensf (commented):
I have no issues with the code itself, but two concerns:
1 - The license comment below
2 - I wonder if that's something we should be adding to connect-go instead?

private/pkg/transport/http2client/proxy.go
bufdev: Here's an example of how we reference other licenses when appropriate https://github.com/bufbuild/buf/blob/main/private/pkg/diff/diff.go#L22=

bufdev:
@akshayjshah for visibility

akshayjshah:
Discussed at length with Luke and Julian on Friday, will update here too.

private/pkg/transport/http2client/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("reading server HTTP response: %v", err)
+	}
+	defer resp.Body.Close()
+	if resp.StatusCode != http.StatusOK {
+		dump, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			return nil, fmt.Errorf("failed to do connect handshake, status code: %s", resp.Status)
+		}
+		return nil, fmt.Errorf("failed to do connect handshake, response: %q", dump)
```
akshayjshah, line 98: In these two error messages, can we clarify that this is an HTTP CONNECT (vs. something to do with `connect-go` or the Connect RPC protocol)? We're likely to encounter this error when making requests with a `connect-go` client, which will be very confusing.

akshayjshah:
@rhbuf and @rubensf - this _shouldn't_ fix #1184. The changes here only affect clients configured to use H2C, which we only use between internal services. The CLI is, as far as I know, using TLS (as it should!) and not going down this code path at all.

I'll update #1184 with more a more comprehensive plan.

private/pkg/transport/http2client/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	return &bufConn{Conn: conn, r: r}, nil
+}
+
+// proxyDial dials, connecting to a proxy first if necessary. Checks if a proxy
+// is necessary, dials, does the HTTP CONNECT handshake, and returns the
+// connection.
+func proxyDial(netw, addr string) (net.Conn, error) {
```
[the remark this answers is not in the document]
lrewega: I think there are a few different concerns here. I think conflating the "scheme" and "protocol" makes this harder to think about. I'd expect to require supplying a custom `RoundTripper` which thus has to perform its own dialing and proxying. Our dialer can do whatever we want. I don't know what makes sense with a unix domain socket... what would we use to check if the request should be proxied? What's the "hostname"? The default `net/http` proxy ignores `localhost` which seems unsurprising, perhaps the same behaviour is desirable with unix domain sockets. I'd be curious to see how your example works with the standard library, if it even supports that in the default transport.

For HTTP/3, it's more or less the same, with CONNECT happening over HTTP/3 as well. Thus  I'm not sure that checking for TCP makes sense...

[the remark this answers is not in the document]
akshayjshah: > I don't know what makes sense with a unix domain socket... what would we use to check if the request should be proxied? What's the "hostname"? The default net/http proxy ignores localhost which seems unsurprising, perhaps the same behaviour is desirable with unix domain sockets.

Makes sense to me.

> I'd be curious to see how your example works with the standard library, if it even supports that in the default transport.

The standard lib definitely supports `unix://` URIs for HTTP, but I haven't checked whether `ProxyFromEnvironment` does anything unexpected. I'd expect domain sockets to be excluded from proxying, just like loopback.

> For HTTP/3, it's more or less the same, with CONNECT happening over HTTP/3 as well. Thus I'm not sure that checking for TCP makes sense...

The code in this PR _only_ supports an HTTP/1.1 CONNECT - we're not even supporting HTTP/2 right now. Unlikely to matter much in practice, but for now I'd consider delegating to `net.Dial` right away, with no proxying at all, if `netw` is not `tcp`.

rhbuf (commented):
@akshayjshah thanks for the review, good comments all round. I've provided your suggested fix for #1184 in 50022c2f4950bab18eb32d5fb61df555391c1182, let me know if it would be better to split that into a separate PR but I think it's ok to implement it here. 

I'm unsure if the following wouldn't be better to set some sensible defaults for timeouts on the client, but opted to go without since it didn't look like we were doing this before. Let me know what you think!
```go
var baseTransport http.RoundTripper
defaultTransport := http.DefaultTransport.(*http.Transport).Clone()
defaultTransport.TLSClientConfig = option.tlsConfig
defaultTransport.Proxy = option.proxy
```

akshayjshah:
Definitely agree that we _should_ have timeouts (and probably some better config for connection pooling too), but it's fine to address that another day.

Would you like me to take another look now?

rhbuf:
> Definitely agree that we _should_ have timeouts (and probably some better config for connection pooling too), but it's fine to address that another day.
> 
> Would you like me to take another look now?

yes please! although there's much less to look at now that we're moving the H2C changes to core

private/pkg/transport/http2client/client_option.go
```diff
@@ -51,3 +61,13 @@ func WithObservability() ClientOption {
...
+
+// WithProxy returns a new ClientOption to use
+// a proxy.
+//
+// The default is to use http.ProxyFromEnvironment
+func WithProxy(proxyFunc Proxy) ClientOption {
+	return func(option *clientOption) {
+		option.proxy = proxyFunc
+	}
+}
```
akshayjshah, line 73: If we never use this option, why have it? I think we want the CLI and everything in core to use `ProxyFromEnvironment` - maybe we can add this later on, when we know it's necessary.

akshayjshah (approved):
Thanks for going through a few rounds of review, Ross - lgtm!

I left an optional comment about cutting out some unused code. Also, it'd be nice if you kept some of the tests you had before - they're simple enough, and it's nice to have some protection from accidentally breaking this later on.

Either way, no need for more review from me.

PR #1264, opened 2022-07-05
Update webhook callback_url docstring

buildbreaker:
The user will just provide the host for the endpoint which implements the webhook event service. For example, the user will request with callback_url being `https://example.com/route/here` and we will call the endpoint `https://example.com/route/here/buf.alpha.webhook.v1alpha1.EventService/Event`


proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -104,6 +104,6 @@ message Webhook {
   string repository_name = 5;
   // The webhook repository owner name.
   string owner_name = 6;
-  // The subscriber's callback URL where notifications are delivered.
-  string callback_url = 7;
+  // The subscriber's callback host where notifications are delivered.
+  string callback_host = 7;
```
robbertvanginkel, line 108: `host` typically refers to only the domain part of a url, eg for `https://example.com/route/here` only `example.com` is the host.

If we want to support users supplying `https://example.com/route/here` to call `https://example.com/route/here/buf.alpha.webhook.v1alpha1.EventService/Event`, something like `callback_url_prefix` seems more appropriate (and in the comment explain that we will add the `pkg.service/method` to it)

buildbreaker: That is fair. You know -- what if we kept the previous name and I'll add some better doc strings around what specifically this can be?

unmultimedio: I think it's much clearer to ask for the complete `callback_url`, and for the time that we're only supporting `connect` backends, we validate the `/buf.alpha.webhook.v1alpha1.EventService/Event` suffix.

Later, when/if we support JSON or other protos, we can validate that suffix conditionally based on a new param. And still, that field will hold true to itself, the complete callback url.

proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -104,6 +104,8 @@ message Webhook {
   string repository_name = 5;
   // The webhook repository owner name.
   string owner_name = 6;
-  // The subscriber's callback URL where notifications are delivered.
+  // The subscriber's callback url prefix (e.g. `https://example.com/route/here`).
+  // The input will be prefixed by the webhook event RPC
+  // (e.g. the result will be `https://example/route/here/buf.alpha.webhook.v1alpha1.EventService/Event`).
```
unmultimedio, line 109: Not what I had in mind haha. I'd document something more like:

```suggestion
  // The subscriber's callback URL where notifications are delivered.
  // Currently we only support Connect-powered backends, so make sure that
  // your URL ends with `/buf.alpha.webhook.v1alpha1.EventService/Event`.
  //
  // Other delivery mechanisms (like JSON or other protos) may be added in the
  // future.
```

robbertvanginkel, line 109: Nit: I feel like its's worth calling out that we will use unary connect+json (maybe a link to the connect docs?) to call the `buf.alpha.webhook.v1alpha1.EventService/Event` rpc at this URL. The example is helpful, but the example shouldn't be the only thing that is telling me what the protocol/encoding/message type will be.

robbertvanginkel (approved):

proto/buf/alpha/registry/v1alpha1/webhook.proto
robbertvanginkel: The URL doesn't strictly specify encoding, `buf.alpha.webhook.v1alpha1.EventService/Event` could be called with either json or proto payloads.

proto/buf/alpha/registry/v1alpha1/webhook.proto
unmultimedio: We won't use JSON but connect+proto from [Peter's feedback](https://github.com/bufbuild/buf/pull/1249#discussion_r913959823).

proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -105,5 +105,11 @@ message Webhook {
   // The webhook repository owner name.
   string owner_name = 6;
   // The subscriber's callback URL where notifications are delivered.
+  // Currently we only support Connect-powered backends with application/protobuf as the content type.
+  // Make sure that your URL ends with `/buf.alpha.webhook.v1alpha1.EventService/Event`.
+  // For more information about connect see `https://connect.build`.
```
unmultimedio, line 110: Nit

```suggestion
  // For more information about Connect see `https://connect.build`.
```

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/webhook.proto
```diff
@@ -105,5 +105,11 @@ message Webhook {
...
+  //
+  // Other delivery mechanisms (like JSON or other protos) may be added in the
+  // future.
```
robbertvanginkel, line 113: >Other delivery mechanisms (like JSON or other protos)

This doesn't make much sense. Even with connect we would have to pick wether it is sent as proto or json.

The URL doesn't strictly specify encoding, `buf.alpha.webhook.v1alpha1.EventService/Event` could be called with either json or proto payloads.

buildbreaker: Removing

buildbreaker: It's probably not worth calling out. We can explain this in the future when we revisit that decision

PR #1265, opened 2022-07-06
BSR-317: add owner to CuratedPlugin

pkwarren:
In order to print status of plugins after push, we need to display the
owner.


mfridman (approved):
lgtm. Just need to `make generate` and commit those files

PR #1266, opened 2022-07-06
BSR-320: print curated plugin status after push

pkwarren:
After a successful push of a curated plugin to the BSR, we should print
out details of the newly published plugin (in particular, the revision
will be useful as it is incremented on every update).

### DIFF
--- a/private/buf/bufprint/curated_plugin_printer.go
+++ b/private/buf/bufprint/curated_plugin_printer.go
@@ -1,0 +1,110 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bufprint
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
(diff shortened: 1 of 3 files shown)


private/buf/bufprint/curated_plugin_printer.go
```diff
@@ -0,0 +1,110 @@
...
+			return nil
+		},
+	)
+}
+
+type outputCuratedPlugin struct {
+	Owner    string `json:"owner"`
+	Name     string `json:"name"`
+	Version  string `json:"version"`
+	Revision uint32 `json:"revision"`
```
pkwarren, line 100: Other useful fields might be the description, image digest, etc. For now this will print the basics.

@amckinney - Is it ok if the JSON representation prints more details than the tab formatted one? In particular the image digest is going to be pretty long - Docker supports this by truncating image digest by default and providing a `--no-trunc` flag to show the full digest.

mfridman: I'll let Alex be the authority on this one, but from my experience the tabbed output is usually a subset of the json output. 

So imo it's fine to expose the minimum in the tabbed output, and if a user needs additional data they reach for the json representation (typically piped into `jq` or whatever).

amckinney: I agree with @mfridman - it's fine for the JSON representation to be a superset (it often is).

For what it's worth, the default output could just be the fully-qualified name of the plugin (without any tab-delimited format), similar to the `buf push` command. That way it's immediately visible how the user could reference the plugin they just created in a `buf.gen.yaml`, `buf.plugin.yaml`, etc. For example, something like this (`buf push` is shown for clarity):

```
$ buf push
7e040f51304842f6b55fdea26225ac16
```

```
$ buf alpha plugin push
buf.build/library/go:v1.0.0-1
```

This is fine for now though - given that we still don't exactly know how the revision will be included at the end of the fully-qualified plugin reference, it's probably safer to start with this.

amckinney (approved):

PR #1267, opened 2022-07-06
BSR-321: add PLUGIN_VERSION docker build arg

pkwarren:
To reduce copying of versions between `buf.plugin.yaml` and
`Dockerfile.plugin`, set a build arg `PLUGIN_VERSION` which can be used
to fetch dependencies on 3rd party plugins.


private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -135,6 +135,9 @@ func (d *dockerAPIClient) Build(ctx context.Context, dockerfile io.Reader, plugi
 			},
 			Version:   types.BuilderBuildKit, // DOCKER_BUILDKIT=1
 			SessionID: buildkitSession.ID(),
+			BuildArgs: map[string]*string{
+				"PLUGIN_VERSION": &pluginConfig.PluginVersion,
```
pkwarren, line 139: This allows you to manage the version entirely in `buf.plugin.yaml` and write a `Dockerfile.plugin` like:

```dockerfile
FROM debian:bullseye-slim as build
ARG PLUGIN_VERSION
RUN test -n "${PLUGIN_VERSION}"
WORKDIR /app
RUN apt-get update
RUN apt-get install -y curl
RUN curl -sSL https://github.com/protocolbuffers/protobuf-go/releases/download/${PLUGIN_VERSION}/protoc-gen-go.${PLUGIN_VERSION}.linux.amd64.tar.gz -o protoc-gen-go.tar.gz
RUN tar -xvf protoc-gen-go.tar.gz
RUN chmod +x protoc-gen-go

FROM scratch
WORKDIR /app
COPY --from=build /app/protoc-gen-go .

ENTRYPOINT [ "/app/protoc-gen-go" ]
```

mfridman: Seems reasonable.

Slightly unrelated, but this gets us one step closer to templatizing the Dockerfile. Although a bit far-fetched .. I think it'd be super neat if we could take the release URL (binary location) and the version and build a Dockerfile for the given language. Assuming we were able to retrieve said binary.

Super simple for Go where you have a single executable binary. Not sure how this would work for other languages, but that's outside the scope of this PR. Getting ahead of myself.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -180,7 +197,7 @@ func buildDockerPlugin(t testing.TB, dockerClient Client, dockerfilePath string,
 	require.Nilf(t, err, "failed to open dockerfile")
 	pluginName, err := bufpluginref.PluginIdentityForString(pluginIdentity)
 	require.Nilf(t, err, "failed to create plugin identity")
-	pluginConfig := &bufpluginconfig.Config{Name: pluginName}
+	pluginConfig := &bufpluginconfig.Config{Name: pluginName, PluginVersion: "v1.0.0"}
```
mfridman, line 200: Should this be an arg we pass to the func?

mfridman (approved):
One question, otherwise lgtm.

amckinney (approved):

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
pkwarren: Good idea - updated.

PR #1268, opened 2022-07-07
BSR-322: gracefully handle plugin exists

pkwarren:
When a plugin already exists with the same metadata and image digest in
the BSR, we should handle the error more gracefully (it doesn't really
indicate an error).

### DIFF
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -34,6 +34,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/multierr"
+	"go.uber.org/zap"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -217,7 +218,16 @@ func run(
...
+		if connect.CodeOf(err) != connect.CodeAlreadyExists {
+			return err
+		}
+		// Plugin with the same image digest and metadata already exists
+		container.Logger().Info(
+			"plugin already exists",
+			zap.String("name", pluginConfig.Name.IdentityString()),
+			zap.String("digest", buildResponse.Digest),
+		)
+		curatedPlugin = currentRevision
```
pkwarren, line 230: When running `buf alpha plugin push` subsequent times without changing the yaml or Dockerfile, you'll get something like this:

```
INFO    plugin already exists   {"name": "bufbuild.internal/library/go", "digest": "sha256:f91a3e801aac7d715727e10819827f60fc5ad6a0975298982401defc41c2cc3f"}
Owner    Name  Version  Revision
library  go    v1.28.0  0
```

pkwarren: What is tricky is that we don't have a way to determine up front (before building the image, pushing it to the registry, and creating it in the BSR) that it doesn't already exist. I think for CI systems you might just want to have them run `buf alpha plugin push <yaml>` for each plugin stored in a repo, and have them update the plugin if changed or report that it is unchanged if nothing changed since the last build.

If we do return an error, I would imagine most consumers would need to add special handling (with a unique error code) to distinguish this error (and ignore it) from all the other types of errors that could happen.

mfridman (approved):

PR #1270, opened 2022-07-07
BSR-316: add support for plugin dependencies

pkwarren:
Update the buf.plugin.yaml configuration handling code to support
dependencies on other plugins. Update plugin push command to pass
dependencies to the BSR.

### DIFF
--- a/private/bufpkg/bufplugin/bufpluginconfig/config.go
+++ b/private/bufpkg/bufplugin/bufpluginconfig/config.go
@@ -17,6 +17,8 @@ package bufpluginconfig
 import (
 	"errors"
 	"fmt"
+	"math"
+	"strconv"
 	"strings"
 
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginref"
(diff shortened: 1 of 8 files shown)


private/bufpkg/bufplugin/bufpluginconfig/testdata/success/go/buf.plugin.yaml
```diff
@@ -1,6 +1,8 @@
 version: v1
-name: buf.build/grpc/go
-plugin_version: v1.5.0
+name: buf.build/library/go-grpc
+plugin_version: v1.2.0
+deps:
+  - buf.build/library/go:v1.28.0-0
```
pkwarren, line 5: Changed this to match what we might end up using as the naming convention (`protoc-gen-go-grpc` -> `go-grpc` plugin name) and some version numbers that track upstream.

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -214,9 +214,9 @@ func parsePluginDependency(dependency string, pluginIdentity bufpluginref.Plugin
 	if identity.Remote() != pluginIdentity.Remote() {
 		return nil, fmt.Errorf("plugin dependency %q must use same remote as plugin %q", dependency, pluginIdentity.Remote())
 	}
-	version, revisionStr, ok := strings.Cut(versionRevision, "-")
```
mfridman: This might have been a bug before, because Cut operates on the first occurrence. If the upstream semver had a pre-release component it would grab the wrong string.

Can we add a test for semver that contains a pre-release? Example:`v1.0.0-rc.1 ` 

In case we decide to change the delimiter that may be a valid semver identifier.

mfridman (commented):
Overall looks good. Left a [comment](https://github.com/bufbuild/buf/pull/1270#discussion_r916132710) for adding a specific test

Also, when sending the dependencies to the backend we want version and revision as separate entities, is this landing in this PR or a subsequent one?

pkwarren:
> Overall looks good. Left a [comment](https://github.com/bufbuild/buf/pull/1270#discussion_r916132710) for adding a specific test
> 
> Also, when sending the dependencies to the backend we want version and revision as separate entities, is this landing in this PR or a subsequent one?

I'm going to update this PR with the changes to the proto definitions so we don't need to repeat changes to parsing there.

private/bufpkg/bufplugin/bufpluginconfig/config.go
pkwarren: Fixed: https://github.com/bufbuild/buf/pull/1270/files#diff-f714c67709cd9aa7b1bcb0fc19e7415b04b23b30974a4261ab4e7f06be52cc93R175

mfridman (approved):
Great job.

PR #1271, opened 2022-07-08
run CI jobs in go 1.18/1.19

pkwarren:
Update all CI jobs to run in both go 1.18 and 1.19 and use the latest
debian bullseye version vs. buster.


rhbuf (commented):
What's the idea behind running against the RC? My best guess is to detect early warnings and help the go team out by alerting them. But I am worried that if the `ci` run fails for `1.19.0-rc.2` then it's going to block our builds, would that be the case? I see that Github actions has some [error handling](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures), but it only seems related to stopping other concurrent runs in the matrix in the even of a failure.

pkwarren:
Going to close this out for now - the 1.19 release is imminent and we can upgrade when ready.

PR #1273, opened 2022-07-12
use GetRepositoryCommitByReference for module resolver

cyinma:
Use GetRepositoryCommitByReference instead of GetModulePins for module resolver.

We expect the `resolveService` is only be used for dependency resolving, for something that looks up the commit by a reference we can use a more generic and straightforward endpoint `repositoryCommitService.GetRepositoryCommitByReference`.

### DIFF
--- a/private/bufpkg/bufapimodule/module_resolver.go
+++ b/private/bufpkg/bufapimodule/module_resolver.go
@@ -16,39 +16,39 @@ package bufapimodule
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
 	"github.com/bufbuild/buf/private/gen/proto/apiclient/buf/alpha/registry/v1alpha1/registryv1alpha1apiclient"
-	modulev1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/module/v1alpha1"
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufapimodule/module_resolver.go
```diff
@@ -57,26 +57,12 @@ func (m *moduleResolver) GetModulePin(ctx context.Context, moduleReference bufmo
...
-	return targetModulePin, nil
-}
-
-func getModuleIdentity(moduleIdentity bufmoduleref.ModuleIdentity) string {
-	return moduleIdentity.Remote() + "/" + moduleIdentity.Owner() + "/" + moduleIdentity.Repository()
+	return bufmoduleref.NewModulePin(
+		moduleReference.Remote(),
+		moduleReference.Owner(),
+		moduleReference.Repository(),
+		bufmoduleref.Main,
```
robbertvanginkel, line 64: Main does not seem right here, can we make this empty? Or take branch from `repositoryCommit`?

robbertvanginkel (commented):
This looks good implementation wise, but lets update the description with the motivation here  (which is that we expect resolve service to only be used for dependency resolving, and that something that looks up the commit for a reference can use a more generic endpoint).

robbertvanginkel (approved):

PR #1274, opened 2022-07-12
BSR-333: define plugin generation RPC

pkwarren:
Add a new RPC to perform remote code generation using curated plugins.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -114,6 +116,49 @@ service PluginCurationService {
   rpc GetLatestCuratedPlugin(GetLatestCuratedPluginRequest) returns (GetLatestCuratedPluginResponse);
 }
 
+// PluginCodeGenerationService generates code using remote plugins.
+service PluginCodeGenerationService {
+  // GenerateCode generates code using the specified remote plugins.
+  rpc GenerateCode(GenerateCodeRequest) returns (GenerateCodeResponse);
```
pkwarren, line 122: Open to naming here - I preferred this over something like `GeneratePlugins` because it more accurately refers to what the RPC does.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -114,6 +116,49 @@ service PluginCurationService {
...
+message GenerateCodeResponse {
+  // Responses from each plugin execution.
+  // The order of each response matches the order in the GenerateCodeRequest.
+  repeated PluginGenerationResponse responses = 1;
+}
+
+// Request for performing code generation using the specified plugin.
+message PluginGenerationRequest {
+  // The plugin to execute.
+  CuratedPluginReference plugin = 1;
```
pkwarren, line 150: Instead of the previous `PluginReference` type which combined remote/owner/name/version with options, split it out into two separate fields here.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -114,6 +116,49 @@ service PluginCurationService {
...
+  // The options to pass to the plugin. These will
+  // be merged into a single, comma-separated string.
+  repeated string options = 2;
+}
+
+// Response from code generation for a given plugin.
+message PluginGenerationResponse {
+  // Generated code from the plugin.
+  google.protobuf.compiler.CodeGeneratorResponse response = 1;
```
pkwarren, line 159: This felt better to break into a separate message type for future expansion - for example if we want to return stdout/stderr or any other fields. The previous `GeneratePluginsResponse` for example returned `repeated RuntimeLibrary` but it wasn't clear which plugin each library referred to. This allows us to group returned data by plugin.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -114,6 +116,49 @@ service PluginCurationService {
...
+  // Include Well-Known Types from the Image in generation.
+  //
+  // include_imports must be set if include_well_known_types is set.
+  bool include_well_known_types = 4;
+}
+
+message GenerateCodeResponse {
+  // Responses from each plugin execution.
+  // The order of each response matches the order in the GenerateCodeRequest.
+  repeated PluginGenerationResponse responses = 1;
```
pkwarren, line 144: Wavered a bit on whether to return `CuratedPluginReference` here too but went with simplicity of indexing to find the corresponding response.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: `GenerationService`
`CodeGenerationService`

Are a bit shorter and still reflect what the action does .. "generating code". I dunno, just a thought.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: I quite prefer this instead of overloading the `PluginReference` which contains details about identity _and_ code generation options.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
pkwarren: I like it - renamed to `CodeGenerationService`.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
pkwarren: Indexing once is an advantage but if you look here: https://github.com/bufbuild/buf/blob/3863d2fc5e0e96acd7079c7d82d5ce86a3fa7c29/proto/buf/alpha/registry/v1alpha1/generate.proto#L92-L98

Let's say you were generating code with multiple plugins as part of one request. It wasn't possible in the previous API to know which plugin each RuntimeLibrary was associated with (a single plugin can return 0-N RuntimeLibrary messages).

mfridman: Got it, makes sense to me.

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
amckinney: Sounds good - we'll need to keep this ordered with respect to the requests anyway (re: insertion points), so we can easily start without the `CuratedPluginReference`.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
amckinney: It's a bit verbose, but we might want to expand the field name to `plugin_reference`. We do this kind of naming elsewhere, too. We can keep the `curated` prefix out of it though.

If we ever support user-facing plugins, we'll definitely need to get rid of all this `Curated` prefix stuff (but we've already discussed that elsewhere).

pkwarren: Fixed.

PR #1277, opened 2022-07-13
BSR-333: initial support for code generation

pkwarren:
Add support for code generation using new curated plugins. Define a new
'plugin' key for use in buf.gen.yaml which can refer to either a local
or a remote curated plugin.


private/buf/bufgen/config.go
```diff
@@ -163,36 +163,67 @@ func validateExternalConfigV1(externalConfig ExternalConfigV1, id string) error
...
+				if err := checkPathAndStrategyUnset(id, plugin, pluginIdentifier); err != nil {
+					return err
+				}
+			} else {
+				// plugin.Plugin is a local plugin - verify it isn't using an alpha remote plugin path
+				if _, _, _, _, err := bufremoteplugin.ParsePluginVersionPath(pluginIdentifier); err == nil {
+					return fmt.Errorf("%s: invalid local plugin", id)
+				}
+			}
+		} else if plugin.Remote != "" {
```
amckinney, line 196: We often try to prevent using `else` if we don't have to - we can reduce nesting a bit by using `continue` in each statement since each of the `plugin.{Plugin,Remote,Name}` are mutually exclusive at this point.

For example,

```go
if plugin.Plugin != "" {
  ...
  continue
}
if plugin.Remote {
  ...
  continue
}
...
```

With that said, this is actually a better opportunity to use a `switch` (due to the mutual exclusivity):

```go
switch {
  case plugin.Plugin != "":
    ...
  case plugin.Remote != "":
    ...
  case plugin.Name != "":
   ...
  default:
    // Unreachable, but handle error anyway.
}
```

private/buf/bufgen/config.go
```diff
@@ -163,36 +163,67 @@ func validateExternalConfigV1(externalConfig ExternalConfigV1, id string) error
...
+			return fmt.Errorf("%s: plugin %s out is required", id, pluginIdentifier)
 		}
-		if plugin.Remote != "" {
-			if _, _, _, _, err := bufremoteplugin.ParsePluginVersionPath(plugin.Remote); err != nil {
+		if plugin.Plugin != "" {
+			if _, err := bufpluginref.PluginReferenceForString(pluginIdentifier); err == nil {
+				// plugin.Plugin is a remote plugin
+				if err := checkPathAndStrategyUnset(id, plugin, pluginIdentifier); err != nil {
+					return err
+				}
```
amckinney, line 189: We should avoid as much client-side validation as possible - we've found ourselves in a variety of issues due to things like this. We might require that the fully-qualified plugin reference is used (i.e. it contains a version+revision), but we should only enforce that on the server-side.

That way, if we ever support a way to resolve the latest version by omitting a revision entirely, old CLIs will "just work".

I would only validate the absolute minimum set of values on the client side - the remote, owner, and name. We could even omit the version so that we could open the opportunity to automatically resolve the latest version AND revision.

This same thinking applies elsewhere. In summary let's try to keep the client-side validation to an absolute minimum to open up opportunities to permit new behavior with just server-side changes.

private/buf/bufgen/generator.go
```diff
@@ -322,6 +327,52 @@ func (g *generator) execRemotePlugin(
...
+	}
+	responses, err := codeGenerationService.GenerateCode(
+		ctx,
+		bufimage.ImageToProtoImage(image),
+		[]*registryv1alpha1.PluginGenerationRequest{
+			{
+				PluginReference: bufplugin.PluginReferenceToProtoCuratedPluginReference(reference),
+				Options:         options,
+			},
+		},
```
amckinney, line 359: It looks like we're still just planning to send one plugin at a time (but concurrently), so we might not need the repeated field to begin with after all. But this is at least symmetrical with the old model, so it seems fine to leave it as-is.

amckinney (commented):
Overall looks great - the main thing I'm worried about is how much client-side validation we have here. Let's see what we can do to remove as much as possible.

amckinney:
As discussed offline / for posterity, we'll want to wait to merge this one until we're totally prepared for this to go out in the next release.

private/buf/bufgen/generator.go
pkwarren: I didn't look at this in detail before but if this is the expected calling pattern do you think we should change the API to match how it is used? Do we think there will be worthwhile optimizations of executing multiple plugins in batch on the backend?

amckinney: Technically updating a single `string` field to be `repeated` is a backwards wire compatible change ([link](https://developers.google.com/protocol-buffers/docs/proto#updating)), so we should be able to make this change later on if ever. The only thing is if we change the field name - starting with a single `string` would suggest that it should be called a `request` (rather than `requests`), which will be another form of code-level incompatibility.

Batching them would prevent multiple round trips which is nice. Plus, it doesn't really do us any harm to support this feature out the gate, so I'm OK with keeping this as `repeated` for now. We'll be the only caller for the foreseeable future, so we can always change it later on before the BSR API is actually consumed by others.

private/buf/bufgen/config.go
pkwarren: I've updated the implementation to relax client side validation to allow either a PluginIdentity or a PluginReference to be specified.

private/bufpkg/bufplugin/bufpluginref/bufpluginref.go
```diff
@@ -107,7 +119,7 @@ func parsePluginIdentityComponents(path string) (remote string, owner string, pl
 		return "", "", "", newInvalidPluginIdentityStringError(path)
 	}
 	plugin = strings.TrimSpace(slashSplit[2])
-	if plugin == "" {
+	if plugin == "" || strings.ContainsRune(plugin, ':') {
```
mfridman, line 122: Do we still need this `:` check?

mfridman:
> As discussed offline / for posterity, we'll want to wait to merge this one until we're totally prepared for this to go out in the next release.

@amckinney Afaics there are no breaking changes, any reason we don't want to merge this as a "hidden" feature, instead of having to build from this branch (for our own purposes) and having to update it with `main`?

There might be something more I'm missing, but figured I'd ask.

private/bufpkg/bufplugin/bufpluginref/bufpluginref.go
pkwarren: This felt like a good safeguard to prevent someone from attempting to parse a PluginReference as a PluginIdentity. Previously that would work without errors, even though the plugin name should never contain a `:` separator.

mfridman:
I'm not overly fond of the `plugin` keyword, but it's the best we've come up with .. so my suggestion is to just roll with it.

After typing these out a few times, it don't bug me that much.

mfridman (approved):

amckinney (approved):

amckinney:
Don't forget to add an entry to the `CHANGELOG.md`!

PR #1281, opened 2022-07-14
Add Proxy to httpclient akin to #1262

rubensf:


private/pkg/transport/http/httpclient/httpclient_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestProxy(t *testing.T) {
+	proxyResponse := []byte("hello")
+	// setup a proxy server, it doesn't actually have to proxy anywhere
+	proxy := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write(proxyResponse)
```
akshayjshah, line 31: Maybe we can assert that there's no error from the write, rather than throwing the error away?

private/pkg/transport/http/httpclient/httpclient_test.go
```diff
@@ -0,0 +1,50 @@
...
+	)
+
+	req, err := http.NewRequest("GET", "http://www.example.com", nil)
+	require.NoError(t, err)
+	resp, err := client.Do(req)
+	require.NoError(t, err)
+	require.Equal(t, http.StatusOK, resp.StatusCode)
+	respBody, err := ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	require.Equal(t, proxyResponse, respBody)
```
akshayjshah, line 49: If this ever fails, the test output is much better if we cast both `proxyResponse` and `respBody` to `string`.

akshayjshah (approved):
LGTM, some nits but no need for more review. This is great!

PR #1282, opened 2022-07-15
add hidden target argument for local arm64 dev

pkwarren:
In order to support building images on M1 for local development and testing, add a hidden `--target` CLI argument which can be used to override the default architecture.


mfridman (approved):
lgtm.

PR #1285, opened 2022-07-18
BSR-355: use image push digest

pkwarren:
Update to use the digest returned from push (instead of build) as they
don't match and remote execution fails with the build image digest.


amckinney (approved):
lgtm

PR #1290, opened 2022-07-22
BSR-374: start initial revision at 1

pkwarren:
When creating plugins, start initial revisions at 1 instead of 0.


srikrsna-buf:
We should also change `genversion` to ignore `1` instead of `0`

mfridman (approved):

pkwarren:
> We should also change `genversion` to ignore `1` instead of `0`

I think I've got this updated now - can you verify those changes?

PR #1298, opened 2022-07-26
ModuleReferenceForString support reference with slash

cyinma:
this make the `ModuleReferenceForString` support reference with slash (e.g. `buf.build/bufbuild/buf:test/draft`), as both tag and draft should support it

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleref/util.go
+++ b/private/bufpkg/bufmodule/bufmoduleref/util.go
@@ -20,7 +20,7 @@ import (
 )
 
 // parseModuleReferenceComponents parses and returns the remote, owner, repository,
-// and ref (branch, commit, or draft) from the given path.
+// and ref (branch, commit, draft, or tag) from the given path.
 func parseModuleReferenceComponents(path string) (remote string, owner string, repository string, ref string, err error) {
 	split := strings.Split(path, ":")
 	switch len(split) {
--- a/private/bufpkg/bufmodule/bufmoduleref/util.go
+++ b/private/bufpkg/bufmodule/bufmoduleref/util.go
@@ -20,25 +20,25 @@ import (
 )
 
 // parseModuleReferenceComponents parses and returns the remote, owner, repository,
-// and ref (branch or commit) from the given path.
+// and ref (branch, commit, or draft) from the given path.
 func parseModuleReferenceComponents(path string) (remote string, owner string, repository string, ref string, err error) {
-	remote, owner, rest, err := parseModuleIdentityComponents(path)
-	if err != nil {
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufmodule/bufmoduleref/util.go
```diff
@@ -20,25 +20,25 @@ import (
...
+	case 2:
+		ref = strings.TrimSpace(split[1])
 		if ref == "" {
 			return "", "", "", "", newInvalidModuleReferenceStringError(path)
 		}
-		return remote, owner, repository, ref, nil
+	default:
+		return "", "", "", "", newInvalidModuleReferenceStringError(path)
+	}
+	remote, owner, repository, err = parseModuleIdentityComponents(split[0])
```
buildbreaker, line 37: Ah. You're doing some "input cleaning" with the draft/tag before deriving the identity components

buildbreaker (commented):
Seems straightforward of a change to me but I'll let others chime in @cyinma

private/bufpkg/bufmodule/bufmoduleref/util.go
```diff
@@ -20,25 +20,25 @@ import (
 )
 
 // parseModuleReferenceComponents parses and returns the remote, owner, repository,
-// and ref (branch or commit) from the given path.
+// and ref (branch, commit, draft, or tag) from the given path.
 func parseModuleReferenceComponents(path string) (remote string, owner string, repository string, ref string, err error) {
```
doriable, line 24: Style nit, feel free to ignore: this makes sense to me, logically, but would it be possible/reasonable to refactor this not to use named returns? We're not using the default zero values and explicitly setting zero values in error return cases anyway, I think we might as well instantiate these explicitly in the function...

doriable (approved):
Code is reasonable, I just left a tiny style/nit, but stamping, since I am in agreeance with this change.
Would like a stamp from @amckinney to confirm consensus on supporting `/` in tag/draft names :)

private/bufpkg/bufmodule/bufmoduleref/util.go
cyinma: as per internal style guide, I think we prefer named returns when the function returns multiple values of the same type to make it more readable, and the default zero values of `ref` is actually used in the case of `len(split) == 1`, so I think it's reasonable to keep this as is?

amckinney (approved):
lgtm

private/bufpkg/bufmodule/bufmoduleref/util.go
doriable: That seems very reasonable to me!

PR #1305, opened 2022-07-28
[bsr-360] No timeout for studio agent command

unmultimedio:
Make sure timeout has no timeout by default. Builder's timeout is inaccessible within each `NewCommand` function, so we need to declare it from the caller side.

### DIFF
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -95,6 +95,10 @@ func NewRootCommand(name string) *appcmd.Command {
 		appflag.BuilderWithTimeout(120*time.Second),
 		appflag.BuilderWithTracing(),
 	)
+	noTimeoutBuilder := appflag.NewBuilder(
+		name,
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/buf.go
```diff
@@ -138,7 +142,7 @@ func NewRootCommand(name string) *appcmd.Command {
 				SubCommands: []*appcmd.Command{
 					convert.NewCommand("convert", builder),
 					migratev1beta1.NewCommand("migrate-v1beta1", builder),
-					studioagent.NewCommand("studio-agent", builder),
+					studioagent.NewCommand("studio-agent", noTimeoutBuilder),
```
unmultimedio, line 145: I found no way to modify that within the studio-agent scope, builder options/params are not accessible, and the expected input is an interface.

amckinney (approved):
lgtm - not ideal that we have to use a separate builder, but it seems like the only way right now.

PR #1306, opened 2022-07-29
[bsr-383] Fix flaky studio agent test

unmultimedio:
Apparently, when doing [`conn.Close()`](https://github.com/bufbuild/buf/blob/bfa9bc47d59baf3d2487660044c7d92ffbc192f4/private/bufpkg/bufstudioagent/bufstudioagent_test.go#L194) in the test, to fake a server-closed connection, it's mapped sometimes as `connect.CodeUnavailable` instead of `connect.CodeUnknown`, that's what's causing the flakiness.

Dig into `connect-go`, and it's being returned [in the envelope](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/envelope.go#L178), by reading the error [from the request](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/duplex_http_call.go#L144).

The error that's being triggered is on the Connect's [`httpClient.Do`](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/duplex_http_call.go#L244) invocation, a `net/url.Error`:

```
error(*net/url.Error) *{Op: "Post", URL: "http://127.0.0.1:56188", Err: error(*errors.errorString) *{s: "io: read/write on closed pipe"}}
```

Which later is unwrapped into a regular error string `"io: read/write on closed pipe"`, and then wrapped again into a [`connect.CodeUnavailable`](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/duplex_http_call.go#L251).

That's the one that's being returned, but when doing `conn.Close()` the test is expecting a [`BadGateway`](https://github.com/bufbuild/buf/blob/bfa9bc47d59baf3d2487660044c7d92ffbc192f4/private/bufpkg/bufstudioagent/bufstudioagent_test.go#L210).

Not sure if this is expected, or what's the original intention of the test [`t.Run("invalid_upstream",`](https://github.com/bufbuild/buf/blob/bfa9bc47d59baf3d2487660044c7d92ffbc192f4/private/bufpkg/bufstudioagent/bufstudioagent_test.go#L188). Ticket [BSR-383](https://linear.app/bufbuild/issue/BSR-383/flaky-test-in-bufstudioagent-testgo)

tbh, not sure if the fix is here, or in Connect by returning a `CodeUnknown` instead of `CodeUnavailable` on the [catch-all non-Connect error](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/duplex_http_call.go#L251). Related [commit](https://github.com/bufbuild/connect-go/commit/1aba8655deeef4a3477e50a4526d4e30e7cf00b6).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,6 @@
 - Change default for `--origin` flag of `buf beta studio-agent` to `https://studio.buf.build`
 - Change default for `--timeout` flag of `buf beta studio-agent` to `0` (no timeout). Before it was
   `2m` (the default for all the other `buf` commands).
-- Add `connect.CodeUnavailable` as an unhandled Connect error code for `buf beta studio-agent`,
-  which will trigger an `http.StatusBadGateway` response in case of a bad or unreachable upstream
-  server. Before the only unhandled Connect error code was `connect.CodeUnknown`.
 
 ## [v1.7.0] - 2022-06-27
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 ## [Unreleased]
 
 - Change default for `--origin` flag of `buf beta studio-agent` to `https://studio.buf.build`
+- Change default for `--timeout` flag of `buf beta studio-agent` to `0` (no timeout). Before it was
+  `2m` (the default for all the other `buf` commands).
+- Add `connect.CodeUnavailable` as an unhandled Connect error code for `buf beta studio-agent`,
+  which will trigger an `http.StatusBadGateway` response in case of a bad or unreachable upstream
+  server. Before the only unhandled Connect error code was `connect.CodeUnknown`.
 
 ## [v1.7.0] - 2022-06-27
 
--- a/private/bufpkg/bufstudioagent/bufstudioagent_test.go
+++ b/private/bufpkg/bufstudioagent/bufstudioagent_test.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -191,7 +192,9 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 		listening := make(chan struct{}, 1)
 		go func(listening chan<- struct{}) {
 			listening <- struct{}{}
+			fmt.Println("signal sent")
 			conn, err := listener.Accept()
+			fmt.Println("connection arrived")
 			require.NoError(t, err)
 			require.NoError(t, conn.Close())
 		}(listening)
@@ -207,7 +210,9 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 		request, err := http.NewRequest(http.MethodPost, agentServer.URL, bytes.NewReader(requestBytes))
 		require.NoError(t, err)
 		request.Header.Set("Content-Type", "text/plain")
+		fmt.Println("waiting before doing the request")
 		<-listening
+		fmt.Println("unblocked")
 		response, err := agentServer.Client().Do(request)
 		require.NoError(t, err)
 		assert.Equal(t, http.StatusBadGateway, response.StatusCode)
--- a/private/bufpkg/bufstudioagent/bufstudioagent_test.go
+++ b/private/bufpkg/bufstudioagent/bufstudioagent_test.go
@@ -25,7 +25,6 @@ import (
 	"net/http/httptest"
 	"strconv"
 	"testing"
-	"time"
 
 	studiov1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/studio/v1alpha1"
 	"github.com/bufbuild/connect-go"
@@ -142,7 +141,6 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 			nil,
 		),
 	)
-	agentServer.Client().Timeout = time.Second
 	defer agentServer.Close()
 
 	t.Run("forbidden_header", func(t *testing.T) {
--- a/private/bufpkg/bufstudioagent/bufstudioagent_test.go
+++ b/private/bufpkg/bufstudioagent/bufstudioagent_test.go
@@ -20,9 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
-	"fmt"
 	"io"
-	"net"
 	"net/http"
 	"net/http/httptest"
 	"strconv"
@@ -40,8 +38,9 @@ import (
 )
 
 const (
-	echoPath  = "/echo.Service/EchoEcho"
-	errorPath = "/error.Service/Error"
+	echoPath    = "/echo.Service/EchoEcho"
+	errorPath   = "/error.Service/Error"
+	unknownPath = "/unknown.Service/Unknown"
 )
 
 func TestPlainPostHandlerTLS(t *testing.T) {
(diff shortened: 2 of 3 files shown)


doriable:
Took a look at this and had a quick chat about this with @unmultimedio offline, summarising our points here:

- I think fix should definitely be on this side of things, since the flakiness seems to be caused by the fact that `conn.Close()` is called in a go routine
- We're going to look to clean-up the flow of the test, but it is also worth noting/considering that `CodeUnavailable` should maybe also be a `BadGateway` since it should deter retry mechanisms

private/bufpkg/bufstudioagent/bufstudioagent_test.go
```diff
@@ -141,6 +143,7 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 			nil,
 		),
 	)
+	agentServer.Client().Timeout = time.Second
```
unmultimedio, line 146: If we didn't do this, tests would hang waiting for the http client to wait for a response. Default is >30s.

This test is still flaky because sometimes we were getting:
```
waiting before doing the request
signal sent
unblocked, starting request
connection arrived
connection closed
... timeout
```

Which meant that the http client was the one hanging out. After adding this shorter timeout, is still flaky, but failures are:

```
waiting before doing the request
signal sent
unblocked, starting request
connection arrived
connection closed
request completed
--- FAIL: TestPlainPostHandlerH2C (1.05s)
    --- FAIL: TestPlainPostHandlerH2C/invalid_upstream (1.00s)
        /Users/jfigueroa/dev/buf/buf/private/bufpkg/bufstudioagent/bufstudioagent_test.go:221: 
            	Error Trace:	/Users/jfigueroa/dev/buf/buf/private/bufpkg/bufstudioagent/bufstudioagent_test.go:221
            	Error:      	Received unexpected error:
            	            	Post "https://127.0.0.1:62351": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
            	Test:       	TestPlainPostHandlerH2C/invalid_upstream
FAIL
```

We suspect the failure comes from [this defer](https://github.com/bufbuild/connect-go/blob/713199600fa7bc3a47957c910ddde9a85253036f/duplex_http_call.go#L240) never being triggered. cc @akshayjshah @pkwarren

akshayjshah (commented):
Caught up with Doria about this today.

I think that this approach to producing errors will always be flaky to one extent or another. The core of the problem is that it's very difficult to synchronize between the accept goroutine and the `net/http` code in `HTTPClient.Do` - even with the proposed changes, we don't have any guarantee that the connection is closed before the request issues the syscall to read data. `connect-go` tries to inspect the Go errors from `HTTPClient.Do` and assign meaningful error codes, but it's never going to be perfect.

In short, it's difficult to reliably make an HTTP request using a TCP connection which the remote has already closed. I can imagine some slightly more elaborate code to improve what we have, but I don't think it's worthwhile.

To me, the tests would be simpler to read, have no flakes, and be more exhaustive if we factored the error-to-HTTP-status mapping code out of `plainPostHandler.ServeHTTP` to a helper function. We can then easily test handling of _every_ connect error code, which serves as documentation of our intentions. (Reading this code, for example, I can't tell why we don't also map `ResourceExhausted` to 503 - did we just forget, or do we specifically not want to do that?) This approach sacrifices some "realism," but it doesn't look to me like that realism is particularly valuable here.

private/bufpkg/bufstudioagent/bufstudioagent_test.go
```diff
@@ -185,18 +187,9 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 		assert.Equal(t, "something", upstreamResponseHeaders.Get("grpc-message"))
 	})
 
-	t.Run("invalid_upstream", func(t *testing.T) {
-		listener, err := net.Listen("tcp", "127.0.0.1:")
-		require.NoError(t, err)
-		go func() {
-			conn, err := listener.Accept()
-			require.NoError(t, err)
-			require.NoError(t, conn.Close())
```
unmultimedio, line 194: This line is inherently flaky because of [the behavior of http clients handling closed connections](https://github.com/golang/go/issues/16100). So we can return the Unknown code, and expect and test the same behavior.

private/bufpkg/bufstudioagent/plain_post_handler.go
```diff
@@ -166,33 +166,23 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 			}
-			i.writeProtoMessage(w, &studiov1alpha1.InvokeResponse{
-				// connectErr.Meta contains the trailers for the
-				// caller to find out the error details.
-				Headers: goHeadersToProtoHeaders(connectErr.Meta()),
-			})
-			return
 		}
+		// Any issue connecting that is not a Connect error is assumed to be because
+		// the server is in some kind of bad and/or unreachable state.
```
unmultimedio, line 185: No need for checking `net.OpErr` or `url.Error`, all of those cases were doing the same thing as this catch-all. Same for the `Unknown` and `CodeUnavailable` connect codes.

unmultimedio:
@doriable @akshayjshah thanks for the revisions and help. As right now the only two codes that cannot be handled with an `invokeResponse` and `err.Meta()` are Unavailable and Unknown, it's easier to have those two conditions in the code inlined. But if that list grows, and other cases appear, we should definitely take Akshay's approach to having a map of connect codes to http status codes for the studio agent.

private/bufpkg/bufstudioagent/bufstudioagent_test.go
```diff
@@ -236,6 +227,13 @@ func newTestConnectServer(t *testing.T, tls bool) *httptest.Server {
 		},
 		connect.WithCodec(&bufferCodec{name: "proto"}),
 	))
+	mux.Handle(unknownPath, connect.NewUnaryHandler(
+		errorPath,
```
pkwarren, line 231: Should this be `unknownPath`?

doriable (approved):
This seems reasonable to me.
I think given the slight behavioural change of catching `connect.CodeUnavailable`, we may want to add a changelog entry for the `buf beta studio-agent` command, but I'm going to approve first on the code :)

private/bufpkg/bufstudioagent/bufstudioagent_test.go
unmultimedio: Yes! will fix

private/bufpkg/bufstudioagent/plain_post_handler.go
```diff
@@ -166,33 +166,23 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		// errors with the Unavailable code and client connection
-		// errors.
-		if netErr := new(net.OpError); errors.As(err, &netErr) {
-			http.Error(w, err.Error(), http.StatusBadGateway)
-			return
-		}
-		if urlErr := new(url.Error); errors.As(err, &urlErr) {
-			http.Error(w, err.Error(), http.StatusBadGateway)
-			return
-		}
```
pkwarren, line 183: I wonder if we need https://github.com/bufbuild/connect-go/issues/222 to accurately determine what to return here? I don't like that to get a test passing we're treating any CodeUnavailable errors (whether from client connection or returned from server) as StatusBadGateway.

I'd feel better if this fix was restricted to the unit tests and didn't change this logic.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@
 ## [Unreleased]
 
 - Change default for `--origin` flag of `buf beta studio-agent` to `https://studio.buf.build`
+- Change default for `--timeout` flag of `buf beta studio-agent` to `0` (no timeout). Before it was
+  `2m` (the default for all the other `buf` commands).
```
unmultimedio, line 7: Missing from https://github.com/bufbuild/buf/pull/1305/files

private/bufpkg/bufstudioagent/plain_post_handler.go
unmultimedio: If you check carefully the code that was written, the only logic I changed was adding `connect.CodeUnavailable` as another connect error code to trigger an `http.StatusBadGateway`. All the other cases: net.OpError, url.Error, or the catch-all in L196, all of them were doing the same thing, triggering an `http.StatusBadGateway`. All I did was simplify that code into a single catch-all in new L186.

unmultimedio:
> we may want to add a changelog entry for the `buf beta studio-agent` command

@doriable Added in [the latest commit](https://github.com/bufbuild/buf/pull/1306/commits/902ddbacff875c9a6259161f113e7878cce20ba7) :)

doriable (approved):

private/bufpkg/bufstudioagent/plain_post_handler.go
pkwarren: The main difference from before is that if you had a handler that returned CodeUnavailable (perhaps with error details) and it didn't wrap net.OpError or url.Error, we'd call `i.writeProtoMessage` with the error metadata.

Now we're treating CodeUnavailable (whether from a network error or from the handler) as StatusBadGateway. This means that people who use CodeUnavailable for whatever reason in their handler won't be able to dig into more details in the error.

akshayjshah: As I understand it, this PR is purely attempting to improve test flakiness. Is that correct? If so, it's important but not super urgent.

@pkwarren is correct about the error handling gotchas we're introducing here. If this change can wait a few days, or if we're comfortable reverting it in a few days, let's address bufbuild/connect-go#222 and use that logic here.

unmultimedio: Just had a zoom with @pkwarren, explained me the error logic here, thank you! I'll update this PR to just be skipping the flaky test, and improving comments.

pkwarren (approved):

PR #1307, opened 2022-07-29
Merge HTTP Clients for ease of use

smaye81:
This merges the two http clients in `buf` into one client that handles all the concerns.

The previous `http2client` was created to be able to use cleartext because the `http2.Transport` struct was the only way to configure `AllowHTTP`.  Therefore, users needed to use the `http2client` if they wanted cleartext and the regular `httpclient` for everything else.  

This essentially takes what both clients were doing and combines them into one interface.  If a callsite requests H2C, then the client will use `http2.Transport` with `AllowHTTP`.  For all other client constructions, the regular `http` client is used.  

Finally, instead of returning a struct with a nested `http.Client`, this simply returns the client straightaway.  The previous implementation only seemed to have the wrapping struct for use with functional parameters.  These properties were not used anywhere other than during construction of the client.  This allows us to simplify all the callsites to use a straight-forward `*http.Client` rather than an unnecessary custom `httpclient.Client` type.

When this PR is merged, this obviates the need for the `http2client` altogether and that package can be deleted.  Its usages will be refactored to use this `httpclient` interface.

Note this is a replica of https://github.com/bufbuild/buf/pull/1288 (which was stamped and merged), but had to be reverted for unrelated issues.


private/pkg/transport/http/httpclient/httpclient.go
```diff
@@ -16,65 +16,62 @@ package httpclient
...
 //
 // The default is to use no observability.
-func ClientWithObservability() ClientOption {
-	return func(client *client) {
-		client.observability = true
+func WithObservability() ClientOption {
+	return func(opts *clientOptions) {
+		opts.observability = true
+	}
+}
```
amckinney, line 51: I know the stutter is a little frustrating, but we might actually want to leave these options as-is with the `Client*` prefix. It's consistent with what you see in other packages like `bufmodule.ModuleWithTargetPaths`.

amckinney (approved):
lgtm - one nit on option naming, but seeing as that this was already stamped and merged in #1288, I'm less concerned about those smaller details.

PR #1314, opened 2022-08-02
remove PluginLanguage

pkwarren:
Use the existing RuntimeConfig oneof instead of having a separate
PluginLanguage defined for curated plugins.


smallsamantha (approved):

mfridman (approved):
A few suggestions for us to think through.

The runtime_config is evolving to be a knob to tune the BSR registry, maybe we should rename this to registry_config. It'd be nice (outside the scope of this PR) to have this be a well-defined enum we propagate all the way down to the BSR.

Furthermore, the language appears to be an orthogonal concern? I think this could be an optional field where we let the plugin author define a repeated string of allowed languages, and then in other places we can filter by language(s)

amckinney:
> The runtime_config is evolving to be a knob to tune the BSR registry, maybe we should rename this to registry_config. It'd be nice (outside the scope of this PR) to have this be a well-defined enum we propagate all the way down to the BSR.

I like it - `registry_config` is the more accurate name for this.

> Furthermore, the language appears to be an orthogonal concern? I think this could be an optional field where we let the plugin author define a repeated string of allowed languages, and then in other places we can filter by language(s)

In this framing, it _is_ kind of orthogonal. The one thing I'd question is the idea that a plugin should be able to generate code for multiple languages. This is certainly possible, but I don't think it's the common pattern we want to evangelize. With default options tied to the plugin itself, a plugin really ought to produce a single type of code. For plugins that do more (e.g. `protoc-gen-validate`), they really ought to be published as multiple plugins so that people can use a single plugin name for their language of choice (e.g. `protoc-gen-validate-{go,java,python}`).

pkwarren:
> The runtime_config is evolving to be a knob to tune the BSR registry, maybe we should rename this to registry_config. It'd be nice (outside the scope of this PR) to have this be a well-defined enum we propagate all the way down to the BSR.

I like that - I'll close this out and open a new PR to rename `runtime_config` -> `registry_config` and `PluginLanguage` -> `PluginRegistryType`.

> Furthermore, the language appears to be an orthogonal concern? I think this could be an optional field where we let the plugin author define a repeated string of allowed languages, and then in other places we can filter by language(s)

Yep - I'll get this added to the .proto definition and buf.plugin.yaml format.

pkwarren:
> This is certainly possible, but I don't think it's the common pattern we want to evangelize.

It will be true that most plugins generate one type of output, however connect-web is an example of a plugin that will generate both javascript and typescript.

amckinney:
> It will be true that most plugins generate one type of output, however connect-web is an example of a plugin that will generate both javascript and typescript.

Fair point, but that has the same semantics as `protoc-gen-validate`. With the default options point I made above, we might need to consider publishing the plugin as `protoc-gen-connect-web-{js,ts,dts}` (the three target values mentioned [here](https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-web)).

Otherwise a user couldn't just depend on the plugin - they'd need to override the default options (which is easy via the `buf.gen.yaml`), but then the revision of the plugin would need to be different if we wanted different behavior on the registries.

amckinney:
I take it back, it looks like the default behavior is to generate both `ts` and `js` (I noticed that in the same doc I linked above).

amckinney:
Multiple languages probably makes sense then - I just hope that users don't abuse it where it shouldn't be allowed (according to Hyrum's law they will).

mfridman:
> Fair point, but that has the same semantics as `protoc-gen-validate`. With the default options point I made above, we might need to consider publishing the plugin as `protoc-gen-connect-web-{js,ts,dts}` (the three target values mentioned [here](https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-web)).
>

To understand this a bit better. Did you mean distinct protoc-gen-* plugins or distinct plugins available on the BSR?

I.e., we could still have a single `protoc-gen-connect-web` plugin that accepts options to modify the generated output. But, when we **package and publish** this plugin to the BSR, we'd publish it as:

```
library/connect-web (default: target=js+dts)
library/connect-web-js
library/connect-web-ts
library/connect-web-dts
```

amckinney:
That's exactly right - a single plugin binary, but distinct plugins available on the BSR (just like you've described).

timostamm:
Please excuse the late entry to the discussion, I missed this one originally.

`protoc-gen-connect-web` is and odd case. It can indeed generate two different languages, but should we really publish several BSR plugins from it?

In the context of remote generation with the BSR npm registry, the default output of `protoc-gen-connect-web` (`.js` + `.d.ts`) is the only sensible choice. 

To elaborate: An NPM package with `.ts` files would be mostly useless. The package would not be usable for users without TypeScript, and if they do use TypeScript, it would be prohibitively slow to transpile all dependencies of a project. Therefore, NPM packages are typically transpiled to JavaScript _before_ publishing them. To still allow type checking, the package also includes Type Declaration files (`.d.ts`) which contain only type information for a piece of JavaScript code. The declaration files are useful to the TypeScript compiler and IDEs, but they do not impede running the JavaScript code in any way.

The default target is `js+dts` specifically for remote generation and plugin dependencies, but also because it is the most compatible output possible. 

Users who want to use remote plugin execution or local code generation can set `target=ts` to get slightly more readable code, at the cost of more transpilation work on their end.

PR #1316, opened 2022-08-03
rename Language to RegistryType

pkwarren:
Update the proto model for plugins to rename PluginLanguage to
PluginRegistryType and runtime_config to registry_config.


pkwarren:
Rework of #1314

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -100,10 +100,13 @@ message CuratedPlugin {
...
   string description = 11;
-  // The configuration for the runtime of the plugin.
-  RuntimeConfig runtime_config = 12;
+  // The configuration for the remote registry of the plugin.
+  RegistryConfig registry_config = 12;
   // The revision for this plugin version.
   uint32 revision = 13;
+  // The languages emitted by the plugin.
+  // Typically this will only contain a single language.
+  repeated string languages = 14;
```
amckinney, line 109: This should probably be modeled as an `enum`. Not as big of a deal for our internal `Plugin` interface for now (since we can change that more easily), but we should make sure that it is for the `.proto` API.

With that said, I realize this might be more painful than it's worth - whenever someone has a new language they want to introduce, we would need to update the API to support another language. But this really stands out as a place that we should stop and think: should we use a free-form string and lean on server-side validation to prevent unknown languages, or incorporate a large `enum` with a ton of different options?

amckinney (commented):
Looks good overall - we just should stop and think about how we want to model the `languages` field.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: Great point. A free-form string with server-side validation would make things easier for us, but it also means as a user I'd have to refer to documentation to figure out which languages are supported.

So a self-describing enum kind of makes sense here. I'd imagine after the first dozen there wouldn't be many more added.

Also, what about plugins that output json, or html (docs), etc. Would be categorize these as `json` or `html` or `markdown`? Or would we use a catch-all such as `text`?

pkwarren: Left out languages for now as we need to work out some of these issues - will address in a follow up.

mfridman (approved):

PR #1317, opened 2022-08-03
[BSR-409] Add Studio Request RPCs

unmultimedio:
Add Studio favorite requests RPCs to the registry backend.


proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,119 @@
...
+// StudioRequestService exposes the functionality to manage favorited Requests
+// from Studio.
+service StudioRequestService {
+  // Create registers a favorite Studio requests to the caller's BSR profile.
+  rpc Create(CreateRequest) returns (CreateResponse) {}
+  // Delete removes a favorite Studio request from the caller's BSR profile.
+  rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+  // List shows the caller's favorited requests.
+  rpc List(ListRequest) returns (ListResponse) {}
+}
```
unmultimedio, line 30: About the Search rpc, should we implement it now? If so, should we merge it with the existing [Search service](https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/search.proto#L96)?

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,119 @@
...
+// authentication.
+message CreateRequest {
+  // BSR repository ID this request is linked to.
+  string repository_id = 1;
+  // The friendly name of this request.
+  string name = 2;
+  // The target base URL of this request.
+  string target_base_url = 3;
+  // The target service name as defined in the BSR repository.
+  string service = 6;
```
stefanvanburen, line 44: nit: we can lower this to `4` (and lower the rest as well).

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,119 @@
...
+  // The request headers as defined in Studio.
+  map<string, string> request_headers = 9;
+  // The response body as seen in Studio.
+  string response_body = 10;
+  // The response headers as seen in Studio.
+  map<string, string> response_headers = 11;
+}
+
+// CreateResponse is the proto response representation of the Create method.
+message CreateResponse {
```
bufdev, line 58: This is a global name `CreateResponse` in the main registry package, this should be scoped as `CreateStudioRequestResponse` or the like

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,119 @@
...
+
+// StudioRequestService exposes the functionality to manage favorited Requests
+// from Studio.
+service StudioRequestService {
+  // Create registers a favorite Studio requests to the caller's BSR profile.
+  rpc Create(CreateRequest) returns (CreateResponse) {}
+  // Delete removes a favorite Studio request from the caller's BSR profile.
+  rpc Delete(DeleteRequest) returns (DeleteResponse) {}
+  // List shows the caller's favorited requests.
+  rpc List(ListRequest) returns (ListResponse) {}
```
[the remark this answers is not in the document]
bufdev: yes, scoping the naming as such is consistent with the rest of our APIs - consistency is the most important, and having `CreateRequest/Response` etc means that we're saying "this is the only create type in the entire registry API"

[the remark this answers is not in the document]
unmultimedio: You're right. I was trying to avoid stuttering

```
CreateStudioRequestRequest
CreateStudioRequestResponse
```

But since the resource name is `StudioRequest` there's no way out of it.

proto/buf/alpha/registry/v1alpha1/studio_request.proto
unmultimedio: Thanks, done

proto/buf/alpha/registry/v1alpha1/studio_request.proto
unmultimedio: You're right, fixed.

proto/buf/alpha/registry/v1alpha1/studio_request.proto
unmultimedio: Sure we can do that. cc @paul-sachs @doriable I also think list is good to start, then later bring search if needed.

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+  // The request headers as defined in Studio.
+  map<string, string> request_headers = 7;
+  // The response body as seen in Studio.
+  string response_body = 8;
+  // The response headers as seen in Studio.
+  map<string, string> response_headers = 9;
+}
+
+// CreateStudioRequestResponse is the proto response representation of the
+// CreateStudioRequest method.
```
stefanvanburen, line 60:
```suggestion
// CreateStudioRequestResponse returns the created StudioRequest.
```

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+
+// DeleteStudioRequestRequest is the proto request representation of the
+// DeleteStudioRequest method.
+message DeleteStudioRequestRequest {
+  // The Studio Request ID that will be deleted.
+  string id = 1;
+}
+
+// DeleteStudioRequestResponse is the proto response representation of the
+// DeleteStudioRequest method.
```
stefanvanburen, line 74: maybe just nix this one?

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+message DeleteStudioRequestResponse {}
+
+// ListStudioRequestsRequest is the proto request representation of the
+// ListStudioRequests method.
+message ListStudioRequestsRequest {
+  // The page token for paginating.
```
stefanvanburen, line 80: maybe
```suggestion
  // The page token for paginating.
  // The first page is returned if this is empty.
```
?

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+  string page_token = 1;
+}
+
+// ListStudioRequestsResponse is the proto response representation of the
+// ListStudioRequests method.
+message ListStudioRequestsResponse {
+  // The list of requests in the current page.
+  repeated StudioRequest requests = 1;
+  // The requests' next page token.
```
stefanvanburen, line 89:
```suggestion
  // The requests' next page token.
  // There are no more pages if this is empty.
```
maybe?

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+  google.protobuf.Timestamp update_time = 3;
+  // User ID who created and saved this request.
+  string author_user_id = 4;
+  // BSR repository ID this request is linked to.
+  string repository_id = 5;
+  // The friendly name of this request.
+  string name = 6;
+  // The target base URL of this request.
+  string target_base_url = 7;
+  // The BSR repository owner name (user or oganization).
```
stefanvanburen, line 110:
```suggestion
  // The BSR repository owner name (user or organization).
```

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,126 @@
...
+// from Studio.
+service StudioRequestService {
+  // CreateStudioRequest registers a favorite Studio requests to the caller's
+  // BSR profile.
+  rpc CreateStudioRequest(CreateStudioRequestRequest) returns (CreateStudioRequestResponse) {}
+  // DeleteStudioRequest removes a favorite Studio request from the caller's BSR
+  // profile.
+  rpc DeleteStudioRequest(DeleteStudioRequestRequest) returns (DeleteStudioRequestResponse) {}
+  // ListStudioRequests shows the caller's favorited requests.
+  rpc ListStudioRequests(ListStudioRequestsRequest) returns (ListStudioRequestsResponse) {}
```
stefanvanburen, line 31: based on your doc, did we want to also add `RenameStudioRequest`? (could also be post-mvp)

stefanvanburen (approved):
tiny nits (mostly comment related) - might be able to nix some of them?

but otherwise, looks good to me!

proto/buf/alpha/registry/v1alpha1/studio_request.proto
unmultimedio: Yeah! Almost forgot, thanks!

proto/buf/alpha/registry/v1alpha1/studio_request.proto
unmultimedio: rephrased.

unmultimedio:
@stefanvanburen thanks for the suggestions, fixed!

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,141 @@
...
+import "google/protobuf/timestamp.proto";
+
+// StudioRequestService exposes the functionality to manage favorited Requests
+// from Studio.
+service StudioRequestService {
+  // CreateStudioRequest registers a favorite Studio Requests to the caller's
+  // BSR profile.
+  rpc CreateStudioRequest(CreateStudioRequestRequest) returns (CreateStudioRequestResponse) {}
+  // RenameStudioRequest renames an existing Studio Request.
+  rpc RenameStudioRequest(RenameStudioRequestRequest) returns (RenameStudioRequestResponse) {}
```
unmultimedio, line 28: @paul-sachs @doriable as discussed in the zoom, we're not adding a full capability to update the favorite, but just rename it. For edits, user can just un-favorite and favorite a new one.

paul-sachs: Sure, as long as we're not making any decisions that makes it impossible to create a full edit endpoint. I can def imagine that we'll want edit functionality at some point, most MVP.

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -0,0 +1,141 @@
...
+  // The target RPC method name as defined in the BSR repository.
+  string method = 5;
+  // The request body as defined in Studio.
+  string request_body = 6;
+  // The request headers as defined in Studio.
+  map<string, string> request_headers = 7;
+  // The response body as seen in Studio.
+  string response_body = 8;
+  // The response headers as seen in Studio.
+  map<string, string> response_headers = 9;
```
paul-sachs, line 58: Looking at https://studio.buf.build/, there are two other fields that I think we may need:
- Under options we mark if the request is to be made with cookies or not
image
- What protocol was it made with
image
- What agent was associated with the request
image

Perhaps this was mentioned somewhere in notion about how we are handling these fields, but I would assume we should handle them?

unmultimedio: Nice point! No those weren't mentioned. I'm thinking, to account for these fields and to support potential new ones, just add `map<string, string> meta`? And even move the `response_body` and `response_headers` there (I had no strong opinions on storing responses values)? even timestamps and duration of requests if we're feeling that'd bring some value later?

paul-sachs: I am up for a generic meta if necessary. I've used it elsewhere before, though it does lose any type safety for those fields. Kinda depends on how dogmatic we want to be about making sure all our APIs are type safe. I could argue that if we're trying to be the "replace JSON" company, then we should stick to it and make all these fields very explicit, even if they all get squished and stored as generic key value pairs on the receiving end.

unmultimedio: You're right, I'll add them separately.

paul-sachs (approved):

PR #1319, opened 2022-08-04
Add versions to curated plugin response

mfridman:
This PR adds a `versions` field to the curated plugin response so we can display the requested plugin as well as a versions list. 

An argument could be made that we should have a separate endpoint to retrieve all the versions, probably utilizing a List endpoint and then doing filtering. But I think we should avoid proliferation of distinct RPCs and solve consumer objectives.


renahlee (approved):

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -218,6 +218,13 @@ message GetLatestCuratedPluginRequest {
 }
 
 message GetLatestCuratedPluginResponse {
```
unmultimedio, line 220: I guess this and L116 are not true anymore then?

```suggestion
message GetCuratedPluginResponse {
```

Although this'd be a breaking change. So may be just a clarification point and not an actual suggestion.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -218,6 +218,13 @@ message GetLatestCuratedPluginRequest {
...
-  // The latest plugin.
   CuratedPlugin plugin = 1;
+  // versions is a semver-sorted list in ascending order.
```
unmultimedio, line 222: I suggest showing latest first, when checking versions we usually see them from the most recent to the oldest. Same for the revisions.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: If you omit the version in the request, then it does return the latest curated plugin. And if you include the version, then it'll return the latest curated plugin by revision.

We might want to add revision as an optional, but that hasn't come up yet.

unmultimedio: Gotcha, thanks for the clarification :)

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: Fair, updated.

unmultimedio (approved):

PR #1325, opened 2022-08-08
Add "Limited Write" role to proto schema

stefanvanburen:
~~This adds the "Write Drafts" role to the repository roles.~~

NOTE: Changed the name to "Limited Write" per feedback below.


unmultimedio (approved):
Although I don't have full context on how drafts will work, I recommend looking for a second review haha.

proto/buf/alpha/registry/v1alpha1/role.proto
```diff
@@ -38,6 +38,7 @@ enum RepositoryRole {
   REPOSITORY_ROLE_ADMIN = 2;
   REPOSITORY_ROLE_WRITE = 3;
   REPOSITORY_ROLE_READ = 4;
+  REPOSITORY_ROLE_WRITE_DRAFTS = 5;
```
cyinma, line 41: I was thinking about if this need a comment as it's not as self described as others (and if we do that we should probably do it to the entire file), but that might be more like an implementation detail so I think it's good for now

cyinma (approved):
lgtm!

proto/buf/alpha/registry/v1alpha1/role.proto
robbertvanginkel, line 41: I'm not the biggest fan of the "Write drafts" name for a role, I feel like if I would see "write drafts" as a role in the UI that is very specificly related to one permission but very vague on what else they can do. Have we considered something like "limited write", "partial member", "draft contributor" or similar? In github, the roles other than read/write/admin are generic terms like "maintain", "triage" and you go to the docs table to find out what that gets you.

For context, we had some permissions system based on granular "scopes" in the past, were things like grant a user permission "x" but not "y" was easy and granular. This worked great from the API, but turned out to be very hard to represent in the web UI in a way that would scale. We then moved to the current system of "roles", where a role is a named relation you have to an object (and you only have one, so its easy to display/modify/manage from a UI), and we have some internal mapping of what permissions you should have based on your role.

cyinma: yeh I like "limited write" more

stefanvanburen: thanks for the feedback! definitely good to workshop this name a bit more as I don't think any of us were quite happy with "write drafts" but didn't have anything better.

Out of the ones you mentioned, I'd agree that I like "limited write" the best - "partial member" would be confusing with the organization role "member", and "draft contributor" is a bit long, and we don't use "contributor" much (I don't think).

Another "name" we mentioned was "Read and Write Drafts", which is obviously very literal, but long and not flexible.

It's tricky - I totally understand the move from scopes to roles, but as soon as you get back into wanting a pretty niche one I think it's necessarily tough :). Also, the fact that our roles have an implicit hierarchy (write = read + write, admin = read + write + admin) makes this particular one even more difficult.

I'll change this to "Limited Write" for now, but also leave this open so it can stew longer and see if we can come up with something better.

robbertvanginkel: Its definitely tricky, and I certainly also dont want to bikeshed this indefinitely. 

Beyond a reaosonable set of default roles I think you need to get into user configurable role territory, and at that point you can also punt the naming to the org/instance owner :). Having some sort of limited write role in the defaults seems sane to me.

robbertvanginkel (approved):

stefanvanburen:
going to go ahead and merge this as is, but obviously if someone comes up with a better name in the near-term I'd be happy to change it!

PR #1327, opened 2022-08-08
Avoid usage of the word request

unmultimedio:
It conflicts with the incoming `request` param
https://github.com/bufbuild/core/pull/5512#discussion_r940518909

Unblocking `make bufupgrade`.

bsr-409


pkwarren (approved):

stefanvanburen (approved):

proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -77,7 +77,7 @@ message CreateStudioRequestRequest {
 // CreateStudioRequestResponse returns the created StudioRequest.
 message CreateStudioRequestResponse {
   // Created Studio Request.
-  StudioRequest request = 1;
+  StudioRequest created_request = 1;
```
bufdev, line 80: why not name them both `studio_request`

unmultimedio: Given it's already typed (we already know it's a `StudioRequest`), thought it would give more useful context to named them after the operation.

PR #1328, opened 2022-08-09
add --override-remote flag to plugin push

pkwarren:
Update the 'buf alpha plugin push' command to add an '--override-remote'
flag to allow building/publishing a plugin to another BSR instance
(without requiring changes to the 'buf.plugin.yaml' file).

### DIFF
--- a/make/go/dep_golangci_lint.mk
+++ b/make/go/dep_golangci_lint.mk
@@ -7,9 +7,9 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/golangci/golangci-lint/releases 20220517 checked 20220520
+# https://github.com/golangci/golangci-lint/releases 20220804 checked 20220808
 # Check for new linters and add to .golangci.yml (even if commented out) when upgrading
-GOLANGCI_LINT_VERSION ?= v1.46.2
+GOLANGCI_LINT_VERSION ?= v1.48.0
 
 GOLANGCI_LINT := $(CACHE_VERSIONS)/golangci-lint/$(GOLANGCI_LINT_VERSION)
 $(GOLANGCI_LINT):
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -26,6 +26,7 @@ import (
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufplugindocker"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
+	"github.com/bufbuild/buf/private/pkg/netextended"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/storage"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
@@ -41,6 +42,7 @@ const (
 	errorFormatFlagName     = "error-format"
 	disableSymlinksFlagName = "disable-symlinks"
 	targetFlagName          = "target"
+	overrideRemoteFlagName  = "override-remote"
 )
 
 // NewCommand returns a new Command.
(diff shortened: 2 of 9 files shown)


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -110,6 +118,11 @@ func run(
 	if err := bufcli.ValidateErrorFormatFlag(flags.ErrorFormat, errorFormatFlagName); err != nil {
 		return err
 	}
+	if len(flags.OverrideRemote) > 0 {
+		if _, err := netextended.ValidateHostname(flags.OverrideRemote); err != nil {
```
pkwarren, line 122: This feels like good validation to perform on this arg - I'm sure Docker performs similar validation on the remote.

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -34,8 +34,13 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-// pluginsImagePrefix is used to prefix all image names with the correct path for pushing to the OCI registry.
-const pluginsImagePrefix = "plugins."
+const (
+	// DefaultTarget is the default target architecture for Docker images.
+	DefaultTarget = "linux/amd64"
```
pkwarren, line 39: Exported this so we can include the default in the usage statement.

private/bufpkg/bufstudioagent/bufstudioagent_test.go
```diff
@@ -84,6 +84,7 @@ func testPlainPostHandler(t *testing.T, upstreamServer *httptest.Server) {
 		request.Header.Set("Foo", "foo-value")
 		response, err := agentServer.Client().Do(request)
 		require.NoError(t, err)
+		defer response.Body.Close()
```
pkwarren, line 87: New lint failures.

private/pkg/transport/http/httpclient/httpclient_test.go
```diff
@@ -44,8 +44,9 @@ func TestProxy(t *testing.T) {
 	require.NoError(t, err)
 	resp, err := client.Do(req)
 	require.NoError(t, err)
+	defer resp.Body.Close()
 	require.Equal(t, http.StatusOK, resp.StatusCode)
-	respBody, err := ioutil.ReadAll(resp.Body)
+	respBody, err := io.ReadAll(resp.Body)
```
pkwarren, line 49: New lint failures.

make/go/dep_golangci_lint.mk
```diff
@@ -7,9 +7,9 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/golangci/golangci-lint/releases 20220517 checked 20220520
+# https://github.com/golangci/golangci-lint/releases 20220804 checked 20220808
 # Check for new linters and add to .golangci.yml (even if commented out) when upgrading
-GOLANGCI_LINT_VERSION ?= v1.46.2
+GOLANGCI_LINT_VERSION ?= v1.48.0
```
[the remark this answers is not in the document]
mfridman:
```
==> Upgrading 1 outdated package:
go 1.18.4 -> 1.19
```

I guess that makes 2

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -77,6 +87,16 @@ func newConfig(externalConfig ExternalConfig) (*Config, error) {
 			if err != nil {
 				return nil, err
 			}
+			if len(opts.overrideRemote) > 0 {
```
mfridman, line 90: Should the `bufpluginref` handle overriding the remote (maybe with a functional option passed via the `PluginReferenceForString` and co.), or did you want to scope the override to `bufpluginconfig` only?

mfridman (approved):

private/bufpkg/bufplugin/bufpluginconfig/config.go
pkwarren: Added some utility functions to extract this logic out of the main loop, but I did want to keep this close to `bufpluginconfig` to start since we really only want to perform this transformation when rewriting the ExternalConfig. If we find in the future we need to have more places that transform a PluginIdentity like this, I agree that extracting it out to that package will be better.

PR #1329, opened 2022-08-09
Add organization-level "Machine" role

stefanvanburen:
This role will only be available in environments that are configured for it, hence the comment.


proto/buf/alpha/registry/v1alpha1/role.proto
```diff
@@ -29,6 +29,8 @@ enum OrganizationRole {
   ORGANIZATION_ROLE_OWNER = 1;
   ORGANIZATION_ROLE_ADMIN = 2;
   ORGANIZATION_ROLE_MEMBER = 3;
+  // ORGANIZATION_ROLE_WRITER is not supported in certain environments.
+  ORGANIZATION_ROLE_WRITER = 4;
```
cyinma, line 33: I am thinking if `Writer` going to be too close to a specific resource role (write) and not giving us any flexibility on that in future when we have any new decision (new role etc), will it be better if this is called `Contributor`?

stefanvanburen: yeah, I could go either way: with "Writer" we're really descriptive but less flexible; in the future if we needed to change things I'd expect us to add another role. With "Contributor" we're less descriptive but more flexible.

seankimdev: Just adding a comment while passing by; we currently refer to the users who have been given organization's repository access as "Outside Contributors." Although, since this new role will only be available to single tenant customers, where outside contributors wouldn't exist, it may be okay...? but just thought it's worth noting.

proto/buf/alpha/registry/v1alpha1/role.proto
```diff
@@ -29,6 +29,8 @@ enum OrganizationRole {
   ORGANIZATION_ROLE_OWNER = 1;
   ORGANIZATION_ROLE_ADMIN = 2;
   ORGANIZATION_ROLE_MEMBER = 3;
+  // ORGANIZATION_ROLE_WRITER is not supported in certain environments.
```
robbertvanginkel, line 32: This comment is isn't helpful and imo will only evoke questions. We don't describe any of the impact/availability of any of the other roles and I don't think we should start with one off comments like this.

I think it is fine if people attempt to use this and receive an error message if the role isnt available in the instance.

proto/buf/alpha/registry/v1alpha1/role.proto
robbertvanginkel: Given the comment in  the design doc 

> We’ve envisioned this role as being assignable to a bot user that would be the only user within an organization (minus owners and admins) that can push to a repository — for example, in CI, after a merge to master

Maybe we just name this `ORGANIZATION_ROLE_{MACHINE,BOT,CI}`?

proto/buf/alpha/registry/v1alpha1/role.proto
stefanvanburen: removed 👍

proto/buf/alpha/registry/v1alpha1/role.proto
stefanvanburen: Out of those, I think `MACHINE` sounds the best. ("The Machine role has write access to all organization resources, regardless of base resource roles."). But it does have a similar issue of being less general in terms of the type of users it implies, given that it sounds like it should only be assigned to non-human accounts.

cyinma: both have their limitation, naming it `MACHINE` will give us flexibility on defining what that role can do to resources, and potentially we might be able to allow user to configure this in their organization in future just like base role, but yeh it will not make sense when someone assign it to a human account.
I personally like `MACHINE` more just because it's completely separated with the resource role, and is more aligned with another role `MEMBER`

stefanvanburen: I'm convinced! Changed this from "Writer" to "Machine".

robbertvanginkel (approved):

cyinma (approved):

PR #1332, opened 2022-08-11
Preserve single quotes with buf format

### LINKED ISSUE #1093: Preserve single quotes with buf format
before:
```protobuf
message LoginRequest {
  string name = 1 [
    (gogoproto.moretags) = 'binding:"required"',
    (google.api.field_behavior) = REQUIRED
  ];
  string passwd = 2 [
    (gogoproto.moretags) = 'binding:"required"',
    (google.api.field_behavior) = REQUIRED
  ];
}
```

after run buf format.
```protobuf
message LoginRequest {
  string name = 1 [
    (gogoproto.moretags) = "binding:\"required\"",
    (google.api.field_behavior) = REQUIRED
  ];
  string passwd = 2 [
    (gogoproto.moretags) = "binding:\"required\"",
    (google.api.field_behavior) = REQUIRED
  ];
}
```

it is alway use double quote for string, then Escape the double quote.

amckinney:
Fixes #1093

This adjusts the formatter logic to preserve single quotes for string literals that contain `(")` or `(\")` literals. In general, the formatter prefers double quotes, so if the string literal contains a mix of `(")` and `(')` literals, the string will be surrounded with `(")`.

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -1453,27 +1453,22 @@ func (f *formatter) writeStringValue(stringValueNode ast.StringValueNode) {
 
 // writeStringWithStyle writes a string value with surrounding single quotes or double quotes
 // based on the string value's content. The formatter prefers double quotes, but uses single
-// quotes if the content contains any (") or (\") literals and not any (') or (\') literals.
+// quotes if the content contains any (") literals and not any (') literals.
 //
 // For example,
 //
-//  1. f"o"'o' -> "f\"o\"'o'"      (") and (') are used - surround with (")
-//  2. f"oo" -> 'f"oo"'            (") is used - surround with (')
-//  3. f'oo' -> "f'oo'"            (') is used - surround with (")
-//  4. f\"o\"\'o\' -> "f\"o\"'o'"  (\") and (\') are used - surround with (")
-//  5. f\"o\"o -> 'f"o"o'          (\") is used - surround with (')
-//  6. f\'o\'o -> "f'o'o"          (\') is used - surround with (")
-//  7. foo -> "foo"                By default, use double quotes
+//  1. f"o"'o' -> "f\"o\"'o'"  (") and (') are used - surround with (")
+//  2. f"oo" -> 'f"oo"'        (") is used - surround with (')
+//  3. f'oo' -> "f'oo'"        (') is used - surround with (")
+//  4. foo -> "foo"            By default, use double quotes
 //
 func (f *formatter) writeStringWithStyle(value string) {
 	var (
-		singleQuote        = strings.ContainsRune(value, '\'')
-		doubleQuote        = strings.ContainsRune(value, '"')
-		escapedSingleQuote = strings.Contains(value, "\\'")
-		escapedDoubleQuote = strings.Contains(value, `\\"`)
+		singleQuote = strings.ContainsRune(value, '\'')
+		doubleQuote = strings.ContainsRune(value, '"')
 	)
 	var formattedString string
-	if (doubleQuote || escapedDoubleQuote) && (!singleQuote && !escapedSingleQuote) {
+	if doubleQuote && !singleQuote {
 		// Use a single quote if the content contains any (") or (\")
 		// literals, and not any (') or (\') literals.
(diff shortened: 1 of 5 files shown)


private/buf/bufformat/formatter.go
```diff
@@ -1443,12 +1443,44 @@ func (f *formatter) writeSpecialFloatLiteral(specialFloatLiteralNode *ast.Specia
...
+// based on the string value's content. The formatter prefers double quotes, but uses single
+// quotes if the content contains any (") or (\") literals and not any (') or (\') literals.
+//
+// For example,
+//
+//  1. f"o"'o' -> "f\"o\"'o'"      (") and (') are used - surround with (")
+//  2. f"oo" -> 'f"oo"'            (") is used - surround with (')
+//  3. f'oo' -> "f'oo'"            (') is used - surround with (")
+//  4. f\"o\"\'o\' -> "f\"o\"'o'"  (\") and (\') are used - surround with (")
+//  5. f\"o\"o -> 'f"o"o'          (\") is used - surround with (')
```
pkwarren, line 1464: Won't this be `f\"o\"o` and not `f"o"o` as written?

private/buf/bufformat/testdata/proto2/quotes/v1/quotes.proto
```diff
@@ -0,0 +1,25 @@
...
+  string name = 10001;
+  Foo foo = 10002;
+}
+
+message Foo {
+  string one = 1 [(name) = "f\"o\"'o'"];
+  string two = 2 [(name) = 'f"oo"'];
+  string three = 3 [(name) = "f'oo'"];
+  string four = 4 [(name) = "f\"o\"\'o\'"];
+  string five = 5 [(name) = "f\"o\"o"];
```
pkwarren, line 19: Shouldn't we preserve the user's input here since we don't need to rewrite?

private/buf/bufformat/testdata/proto2/quotes/v1/quotes.proto
```diff
@@ -0,0 +1,25 @@
...
+  string six = 6 [(name) = 'f\'o\'o'];
```
pkwarren, line 20: Shouldn't we preserve the user's input here since there is no extra quoting needed - `f'o'o`?

private/buf/bufformat/formatter.go
amckinney: Not with what I'm going for here - if the user escaped `"` characters, but the string didn't include any `'`, then we can safely surround it with `'` (as shown).

private/buf/bufformat/testdata/proto2/quotes/v1/quotes.proto
amckinney: Similar to the above, this is intentional but it is debatable.

private/buf/bufformat/testdata/proto2/quotes/v1/quotes.proto
amckinney: Again, intentional but debatable.

private/buf/bufformat/formatter.go
pkwarren: Where do we unquote the input though?

https://go.dev/play/p/sLN2nHYBHs5

amckinney: My mistake, I was interpreting this example as if it were surrounding in `"` (like [this](https://go.dev/play/p/Euv-btoqzzk)). But now I realize what you're calling out - I'll take a closer look.

private/buf/bufformat/testdata/proto2/quotes/v1/quotes.proto
amckinney: Also just to make sure you notice - this is the user's input whereas the `quotes.golden.proto` is what it's formatted into - so it does become `"f'o'o"`

private/buf/bufformat/formatter.go
amckinney: Ok, I saw the problem - we don't actually need to care about the explicit escape literals at all in this case. I simplified it down so that it's more clear.

pkwarren (approved):

PR #1341, opened 2022-08-15
Add option for cache to preserve external paths

amckinney:
This adds `bufcli.NewModuleReaderAndCreateCacheDirsWithExternalPaths` so that callers can preserve the external paths within the module cache. 

In general, this is _not_ something we want to do (hence the loud `WARNING` in the function comment), but it's required for some features (e.g. a language server that refers to files in the cache).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,6 +11,8 @@ linters-settings:
       - '^exec\.Cmd$'
       - '^exec\.Command$'
       - '^exec\.CommandContext$'
+      # This should only be used by github.com/bufbuild/buf-language-server
+      - '^bufcli\.NewModuleReaderAndCreateCacheDirsWithExternalPaths'
       # os.Rename does not work across filesystem boundaries
       # See https://github.com/bufbuild/buf/issues/639
       - '^os\.Rename$'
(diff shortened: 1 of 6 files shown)


bufdev (approved):
Approved based on offline conversation about updating comment, and adding this to the banned functions in golangci-lint for here and core.

.golangci.yml
```diff
@@ -11,6 +11,8 @@ linters-settings:
       - '^exec\.Cmd$'
       - '^exec\.Command$'
       - '^exec\.CommandContext$'
+      # This should only be used by github.com/bufbuild/buf-language-server
+      - '^bufcli\.NewModuleReaderAndCreateCacheDirsWithExternalPaths'
```
amckinney, line 15: Verified with the following:
```
private/buf/cmd/buf/command/build/build.go:142:11: use of `bufcli.NewModuleReaderAndCreateCacheDirsWithExternalPaths` forbidden by pattern `^bufcli\.NewModuleReaderAndCreateCacheDirsWithExternalPaths` (forbidigo)
	_, err = bufcli.NewModuleReaderAndCreateCacheDirsWithExternalPaths(nil, nil)
```

PR #1342, opened 2022-08-16
Add redirect import suffix to npm registry config

mfridman:
This PR adds a new field to the npm registry config for import rewrites.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -53,6 +53,10 @@ message NPMConfig {
   }
   // Optionally define the runtime libraries for the plugin.
   repeated RuntimeLibrary runtime_libraries = 1;
+  // Optionally define a redirect import suffix for the plugin, i.e. "connectweb.js" or "pb.js"
```
bufdev, line 56: Is "redirect" the right word? Partially it may be that I don't understand what this does from the comment here, so maybe the comment needs to be expanded

mfridman: I think either redirect or rewrite, the latter better reflects what this actually does "rewrite the imports".

@timostamm had a much more descriptive comment here.

https://github.com/bufbuild/protobuf-es/compare/main...tstamm/import-path-redirection#diff-4a8530d531211a57c4c1af19bf8b3685cb1bad75abece43cf708749eef74baa5R82-R111

Whatever we come up with, it should be same everywhere.

timostamm: I'm open to renaming this, it was named pretty much ad-hoc. It manipulates imports path _before_ they are added to a file - other than what we are currently doing in the BSR npm registry. But this distinction is pretty inconsequential. Mike, seems like you like "rewrite" better?

mfridman: If we had to choose one, I'd vote for:

```
rewrite_imports=<pattern>:<target>
```

The term "redirect" took me a bit to grok as I was building out the backend, whereas rewrite clicked immediately. I.e., "rewrite the import paths"

The other one I had in mind was `rewrite_imports=<pattern>:<target>` .. no strong opinion.

timostamm: Updated the branch, renaming the option to `rewrite_imports`.

timostamm (approved):

PR #1343, opened 2022-08-16
[bsr-465] Set a page size and reverse values when listing studio requests

unmultimedio:
Following how other list endpoints are defined, like organizations, repositories, commits, plugins, etc.

bsr-465


proto/buf/alpha/registry/v1alpha1/studio_request.proto
```diff
@@ -108,8 +108,14 @@ message DeleteStudioRequestResponse {}
...
 message ListStudioRequestsRequest {
+  // The default page size for studio requests is 20. You can set a custom value
+  // between 1-250.
+  uint32 page_size = 1;
   // The page token for paginating. The first page is returned if this is empty.
-  string page_token = 1;
+  string page_token = 2;
+  // The listed studio requests are order by created time in ascending order.
+  // You can reverse that order setting this value to true.
+  bool reverse = 3;
```
renahlee, line 118: (Potentially premature future proofing question) are there other fields we may want to sort on?

renahlee: @robbertvanginkel has informed me I have opened a 🥫 of 🪱 so I will slowly back away 🌚

unmultimedio: Yeah I guess that's slowly diving into GraphQL territory? I tried to follow the existing fields for other list RPCs.

renahlee (approved):

robbertvanginkel (approved):

Issue #1334, opened 2022-08-12
Error on formatting within /tmp


Ganitzsh:
I am using `nvim` with `null-ls` setup with buf for formatting. It was working great up until a month ago or so, and it stopped.

I created this issue https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1002 but according to the author the issue is most likely on the `buf` side. Not saying it is, but I'm looking for an answer.

Here's the proto file to format, it's very basic and nothing fancy:
```
syntax = "proto3";

package straudio.v1;

message AudioChunk {
     string name = 1;

   bytes data = 2;
}
```

Here's the output of the command when run within `nvim`:
```
[DEBUG Fri Aug 12 08:51:04 2022] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:346: spawning command "buf" at /Users/<me>/Dev/audio-streamer/protos/v1 with args { "format", "-w", "/tmp/null-ls_UQJFGT.proto" }
[TRACE Fri Aug 12 08:51:04 2022] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:217: error output: Failure: failed to enumerate module files: open /private/tmp/.vnc-vncservice: permission denied

[TRACE Fri Aug 12 08:51:04 2022] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:218: output: nil
```

As you can see, the error says:
```
[TRACE Fri Aug 12 08:51:04 2022] ...t/null-ls.nvim/lua/null-ls/helpers/generator_factory.lua:217: error output: Failure: failed to enumerate module files: open /private/tmp/.vnc-vncservice: permission denied
```

It works when I run `buf` directly on the file within it's source folder outside of `/tmp`

I have no idea why this file is being opened in the first place. I am suspecting a conflicting file name `/tmp/null-ls_UQJFGT.proto` causing `buf` to try and open something else?

I'm running macOS version 12.4 and nvim version 0.7.2

amckinney:
[...]

PR #1345, opened 2022-08-18
Add --stdin flag to 'buf format'

### LINKED ISSUE #1035: Format accepts input from stdin
I am working a lot with buf and have it also setup to work from inside Neovim and it would make life a bit more convenient if both `buf lint` and `buf format` would accept input from stdin. 

It would allow to show diagnostics live and on unsaved buffers and have a better integration with the `vim.lsp.buf.formatting()` functions as used in [null-ls](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/doc/BUILTINS.md#buf-1). 

Currently, this means that all changes are "undone" when running the formatting command on unsaved buffers as they revert back to what's on disc. Which happens a lot for me since I usually setup formatting like this
```lua
vim.cmd([[autocmd BufWritePre <buffer> lua vim.lsp.buf.formatting_sync(nil, 500)]])
```
which runs the formatting on the unsaved buffer before writing the buffer to disc every time I save a file. 

Maybe by adding an option `--stdin` or detecting when something gets piped in would be an option?

amckinney:
Fixes #1035

This adds a `--stdin` flag to `buf format` so that users can format the `.proto` content read from `stdin`, rather than requiring a file on disk. This makes it far easier to write developer tools that would otherwise need to create a temporary file containing the buffer's content (re: https://github.com/bufbuild/buf/issues/1334#issuecomment-1213539045).

Unsurprisingly, the `--output` and `--write` flags are not supported alongside `--stdin`. For anyone that needs to write the formatted content to disk, they can simply redirect the formatted result written to `stdout`.

### DIFF
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -1987,6 +1987,95 @@ func TestFormatExitCode(t *testing.T) {
 	assert.NotEmpty(t, stdout.String())
 }
 
+func TestFormatStdin(t *testing.T) {
+	stdin := bytes.NewBuffer(
+		[]byte(`
+syntax = "proto3";
+
+
+package foo;
+`),
+	)
+	stdout := bytes.NewBuffer(nil)
+	testRun(
+		t,
+		0,
+		stdin,
+		stdout,
+		"format",
+		"--stdin",
+	)
+	assert.Equal(
+		t,
+		`syntax = "proto3";
+
+package foo;
+`,
+		stdout.String(),
+	)
+
+	stdin = bytes.NewBuffer(
+		[]byte(`
+syntax = "proto3";
+
+
+package foo;
+`),
+	)
+	stdout = bytes.NewBuffer(nil)
+	testRun(
+		t,
+		0,
+		stdin,
+		stdout,
+		"format",
+		"--stdin",
+		"--diff",
+	)
+	assert.NotEmpty(t, stdout.String())
+
+	stdin = bytes.NewBuffer(
+		[]byte(`
+syntax = "proto3";
+
+
+package foo;
+`),
+	)
+	stdout = bytes.NewBuffer(nil)
+	testRun(
+		t,
+		bufcli.ExitCodeFileAnnotation,
+		stdin,
+		stdout,
+		"format",
+		"--stdin",
+		"--exit-code",
+	)
+	assert.NotEmpty(t, stdout.String())
+
+	stdin = bytes.NewBuffer(
+		[]byte(`
+syntax = "proto3";
+
+
+package foo;
+`),
+	)
+	stdout = bytes.NewBuffer(nil)
+	testRun(
+		t,
+		bufcli.ExitCodeFileAnnotation,
+		stdin,
+		stdout,
+		"format",
+		"--stdin",
+		"--diff",
+		"--exit-code",
+	)
+	assert.NotEmpty(t, stdout.String())
+}
+
 // Tests if the image produced by the formatted result is
 // equivalent to the original result.
 func TestFormatEquivalence(t *testing.T) {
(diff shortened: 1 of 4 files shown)


private/buf/cmd/buf/buf_test.go
```diff
@@ -1987,6 +1987,95 @@ func TestFormatExitCode(t *testing.T) {
...
+		`syntax = "proto3";
+
+package foo;
+`,
+		stdout.String(),
+	)
+
+	stdin = bytes.NewBuffer(
+		[]byte(`
+syntax = "proto3";
```
bufdev, line 2019: Can we add a test for multiple files concat'ed together, and either expect success or expect error? If error, document that `--stdin` can only take one file

bufdev (approved):

bufdev (changes requested):
Wait on second thought - should this not be `buf format -` to  be consistent, since `-` for inputs represents stdin? Or am I getting confused? This might be a special case, but wanted to check before we merge.

amckinney:
> Wait on second thought - should this not be buf format - to be consistent, since - for inputs represents stdin? Or am I getting confused? This might be a special case, but wanted to check before we merge.

Nah it's good that you called this out - I'll explain my original reasoning upfront, then we can decide if we want to change this (we probably do).

### Use `--stdin`

I mistakenly  thought the problem with leaning on `-` is that the input read from `stdin` would still need to be interpreted as a filepath - the stdin value is being read for the positional parameter (i.e. `buf format <input>`). For example,

```
buf format proto/buf/alpha/image/v1/image.proto
```

was thought to be equivalent to

```
echo proto/buf/alpha/image/v1/image.proto | buf format -
```

The `--stdin` flag was meant to tell `buf` that the _content_ of the `.proto` file should be read from `stdin`, so it can adjust its operation. That's why we also return an error if the user specifies `buf format some/file.proto --stdin` so that it's clear the input filepath is not used at all.

### Use `-`

Before `--stdin`, I actually went down the path of adding another `buffetch.Format` (e.g. `protosource`) that is only used for `buf format`, but ended up backing out of it since it doesn't align with the `storage.Bucket` abstractions tied up in `buffetch` and `bufwire`. That's solvable though.

However, now that I'm revisiting the other commands, I recall that the `-` input is actually interpreted as an `Image` / `bin` by default, hence:
```
buf build -o - | buf breaking --against -
```

Because `buf format`'s default behavior for `buf format -` is to read the input as an `Image` / `bin` (just like the other `buf` commands), we'd probably want to change the meaning of the `-` input for `buf format` to be `protosource` format by default. The `bin` format isn't even supported for `buf format` (because it acts upon the AST), so we wouldn't actually be breaking anyone.


That way someone could just write the following instead:
```
cat proto/buf/alpha/image/v1/image.proto | buf format -
```

Which is shorthand for:

```
cat proto/buf/alpha/image/v1/image.proto | buf format -#format=protosource
```

@bufdev what are your thoughts?

private/buf/cmd/buf/command/format/format.go
```diff
@@ -242,10 +252,62 @@ func run(
...
 		return err
 	}
+	if flags.Stdin && container.NumArgs() > 0 {
+		return fmt.Errorf("--%s cannot be used with a non-empty input %s", stdinFlagName, input)
+	}
+	runner := command.NewRunner()
+	if flags.Stdin {
+		// We need to read the raw bytes from the io.Reader upfront so that
+		// the content can be referenced more than once: once by the formatter,
+		// and again during the diff.
```
twilly, line 272: I smell something here. `--stdin` is implemented with a divergent different code path because its input is a non-seekable file. Standard input is not the only non-seekable file out there: sockets, tape-archive on literal tape, named pipes, bash's named descriptors (/dev/fd/<n>), etc.

One alternative is to implement a buffered, seeking reader. Wrap stdin with your reader and no more special code paths, I hope. I haven't groked all of the format package yet and `run` is particularly hard to follow.

This concept can extend further if you open a file input and attempt to seek to offset 0. That should, on POSIX systems, fail on pipes, sockets, and other file handles that are non-seekable. Commands like `buf format <(cat foo.proto bar.proto)` become possible.

twilly (approved):
This works but it would be nice to support process substitution and other inputs that cannot seek.

nedtwigg:
Sadface :(  Looks like this PR isn't going to get merged, but is #1035 still planned?

bufdev:
We have no near-term plans on this as there's some thinking we need to do on it, but we may get to it in a few months. We'll keep the issue open for now, we just need the time to budget to make sure we get this right.

PR #1349, opened 2022-08-19
add --cache-from/--pull/--dry-run options

pkwarren:
Update the `buf alpha plugin push` to allow for using cached images
published to a Docker registry, pulling the latest base images at build
time, and bypassing pushing to the BSR so local testing can run builds.

Additionally, add some rudimentary parsing of the `moby.buildkit.trace`
messages so we get more details when run with `--debug`.

### DIFF
--- a/private/bufpkg/bufplugin/bufplugindocker/docker.go
+++ b/private/bufpkg/bufplugin/bufplugindocker/docker.go
@@ -17,6 +17,7 @@ package bufplugindocker
 import (
 	"bufio"
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -169,11 +187,32 @@ func (d *dockerAPIClient) Build(ctx context.Context, dockerfile io.Reader, plugi
...
-			if err := json.Unmarshal([]byte(scanner.Text()), &message); err == nil && message.Error != nil {
+			if err := json.Unmarshal([]byte(scanner.Text()), &message); err != nil {
+				d.logger.Debug(scanner.Text())
+				continue
+			}
+			if message.Error != nil {
 				return message.Error
 			}
+			handled := false
+			if message.ID == "moby.buildkit.trace" && message.Aux != nil {
```
pkwarren, line 199: This isn't pretty - messages are double quoted base64 strings which need to be unmarshalled as a protobuf type. This does allow us to get a lot more debug information though when building to determine what might be failing.

bufdev (approved):

private/bufpkg/bufplugin/bufplugindocker/docker.go
mfridman: Oh, thank you do much for adding this!

PR #1353, opened 2022-08-22
add runner option to silence logging certain endpoints

rhbuf:
pass a list of endpoints to ignore by path, the goal of this is to
stop logging requests to /health and /metrics

### DIFF
--- a/private/pkg/transport/http/httpserver/zap.go
+++ b/private/pkg/transport/http/httpserver/zap.go
@@ -22,15 +22,26 @@ import (
 	"go.uber.org/zap"
 )
 
-func newZapMiddleware(logger *zap.Logger) func(http.Handler) http.Handler {
+func newZapMiddleware(logger *zap.Logger, silentEndpoints []string) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
 			logEntry := newLogEntry(logger, request)
 			wrapResponseWriter := middleware.NewWrapResponseWriter(newCheckedResponseWriter(logger, writer, request), request.ProtoMajor)
-			defer logRequest(logEntry, wrapResponseWriter, time.Now())
+			if !contains(silentEndpoints, request.URL.Path) {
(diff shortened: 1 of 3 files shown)


private/pkg/transport/http/httpserver/zap.go
```diff
@@ -22,17 +22,28 @@ import (
...
-			defer logRequest(logEntry, wrapResponseWriter, time.Now())
+			if !contains(silentEndpoints, request.URL.Path) {
+				defer logRequest(logEntry, wrapResponseWriter, time.Now())
+			}
 			next.ServeHTTP(wrapResponseWriter, middleware.WithLogEntry(request, logEntry))
 		})
 	}
 }
 
+func contains(l []string, s string) bool {
```
rhbuf, line 38: I was unsure about creating a dependency on https://pkg.go.dev/golang.org/x/exp/slices#Contains - perhaps it would it be better to convert to a map?

mfridman: Ye, I'd just `map[string]struct{}` and inline the `request.URL.Path` check.

private/pkg/transport/http/httpserver/zap.go
```diff
@@ -22,28 +22,19 @@ import (
...
 )
 
-func newZapMiddleware(logger *zap.Logger, silentEndpoints []string) func(http.Handler) http.Handler {
+func newZapMiddleware(logger *zap.Logger, silentEndpoints map[string]struct{}) func(http.Handler) http.Handler {
 	return func(next http.Handler) http.Handler {
 		return http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
 			logEntry := newLogEntry(logger, request)
 			wrapResponseWriter := middleware.NewWrapResponseWriter(newCheckedResponseWriter(logger, writer, request), request.ProtoMajor)
-			if !contains(silentEndpoints, request.URL.Path) {
+			if _, ok := silentEndpoints[request.URL.Path]; ok {
```
mfridman: Should this be `!ok`, i.e., if the path is not in the lookup defer the log?

mfridman (approved):

rubensf (approved):

PR #1354, opened 2022-08-22
allow specifying build arguments to plugin push

pkwarren:
Update plugin push command to allow for optional --build-arg arguments
to be passed to the build.

### DIFF
--- a/private/bufpkg/bufplugin/bufplugindocker/docker.go
+++ b/private/bufpkg/bufplugin/bufplugindocker/docker.go
@@ -168,13 +168,12 @@ func (d *dockerAPIClient) Build(ctx context.Context, dockerfile io.Reader, plugi
 		if len(target) == 0 {
 			target = DefaultTarget
 		}
-		buildArgs := map[string]*string{
-			"PLUGIN_VERSION": &pluginConfig.PluginVersion,
-		}
+		buildArgs := make(map[string]*string)
 		for _, arg := range params.buildArgs {
 			name, val, _ := strings.Cut(arg, "=")
 			buildArgs[name] = &val
 		}
+		buildArgs["PLUGIN_VERSION"] = &pluginConfig.PluginVersion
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -161,6 +168,13 @@ func (d *dockerAPIClient) Build(ctx context.Context, dockerfile io.Reader, plugi
 		if len(target) == 0 {
 			target = DefaultTarget
 		}
+		buildArgs := map[string]*string{
+			"PLUGIN_VERSION": &pluginConfig.PluginVersion,
+		}
+		for _, arg := range params.buildArgs {
+			name, val, _ := strings.Cut(arg, "=")
+			buildArgs[name] = &val
+		}
```
amckinney, line 177: Do we want to support overriding the `PLUGIN_VERSION` via `--build-args`? If not, we might need some more validation here. Otherwise looks fine as is.

amckinney (approved):
Sweet!

private/bufpkg/bufplugin/bufplugindocker/docker.go
amckinney: Also curious though, why not express this in the `buf.plugin.yaml` instead (e.g. via a `build_args` key)? If the `PLUGIN_VERSION` is specified there as `version`, then it'd be impossible for the two to overwrite each other.

There might be good reasons not to do this, but just thought I'd ask just in case.

pkwarren: Good call - I shuffled this around. Users shouldn't ever override `PLUGIN_VERSION` - that should always come from `buf.plugin.yaml`.

pkwarren: I need this for https://github.com/bufbuild/plugins/pull/8 - we set `DOCKER_ORG` as a build argument when publishing to a Docker registry and we need to pass it as a build argument to use the correct base images in gRPC/protoc plugins.

PR #1355, opened 2022-08-23
Add otel.TracerProvider to observability layer

doriable:
Add the `otel.TracerProvider` and a closer to the observability layer for
our applications.


private/pkg/observability/observabilityotel/tracer_provider_closer.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+func newTracerProviderCloser(tracerProvider *sdktrace.TracerProvider) *tracerProviderCloser {
+	return &tracerProviderCloser{
+		tracerProvider: tracerProvider,
+	}
+}
+
+func (t *tracerProviderCloser) Close() error {
+	// TODO: this should probably be passed down from the application entrypoint
```
rubensf, line 35: Don't leave hanging TODOs -- associate it with an issue to actually fix it, eg `// TODO (INF-123): fix this coz of a, b, and c.`, or `// TODO (#1234): blabla`. The issue can carry more dynamic context.

FWIW, I'm not sure if we need this TODO. I'd instead leave a `ctx Context` in the arguments, and wrap the closing for `chainCloser` - if this package is "self contained", it should also forward that context, but the way in which we _use_ it is where we lose the context (eg we don't have a `stop observability` endpoint that'd have a request-associated contexT)

doriable: After an offline discussion looking at the various ways we can do this, we've decided to annotate this clearly with a comment and will support a `context.Context` in the layer above if this is every required.

private/pkg/observability/observability.go
```diff
@@ -79,6 +86,11 @@ func Start(options ...StartOption) io.Closer {
 		view.RegisterExporter(traceViewExportCloser)
 		closers = append(closers, traceViewExportCloser)
 	}
+	for _, tracerProviderCloser := range startOptions.tracerProviderClosers {
+		tracerProviderCloser := tracerProviderCloser
+		otel.SetTracerProvider(tracerProviderCloser)
```
pkwarren, line 91: This sets the trace provider as the globally registered one - it feels like assigning this in a loop could indicate a misconfiguration if `len(tracerProviderClosers) > 1`.

private/pkg/observability/observabilityotel/observabilityotel.go
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+package observabilityotel
+
+import (
+	"github.com/bufbuild/buf/private/pkg/observability"
+	"go.opentelemetry.io/otel/sdk/trace"
+)
+
+func NewTracerProviderCloser(tracerProvider *trace.TracerProvider) observability.TracerProviderCloser {
```
pkwarren, line 22: It might be nice to keep this limited to using the trace API and not sdk - i.e. take https://pkg.go.dev/go.opentelemetry.io/otel/trace#TracerProvider and then do type assertions to an interface with `Shutdown(context.Context)` to determine if we should call that on shutdown.

This keeps the buf CLI free of SDK implementation details.

pkwarren (commented):
Looks good - one idea on keeping the SDK out of this implementation to allow use of other providers that might not implement shutdown (like https://github.com/open-telemetry/opentelemetry-go/blob/trace/v1.9.0/trace/noop.go#L27).

pkwarren (approved):

rubensf (approved):

private/pkg/observability/observabilityotel/observabilityotel.go
doriable: I'm a fan of this, going to have an interface here that includes `Shutdown(context.Context) error` and then we'll use that :)

PR #1366, opened 2022-08-29
[bsr-508] Add base audit event

unmultimedio:
So we can use the `AuditEvent` type for the audit logger interface.


proto/buf/alpha/registry/v1alpha1/audit_event.proto
```diff
@@ -0,0 +1,96 @@
...
+  // a plugin.
+  AUDIT_RESOURCE_TYPE_PLUGIN = 7;
+}
+
+// AuditResource is the affected resource by the audited event.
+message AuditResource {
+  // Type of resource that was affected by the audited event.
+  AuditResourceType type = 1;
+  // Id of the affected resource by the audited event.
+  string id = 2;
```
stefanvanburen, line 71: assuming `id` is fine here; was there a reason for `name` in the design doc though? (just double-checking, and similar comment for `AuditActor`)

proto/buf/alpha/registry/v1alpha1/audit_event.proto
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// AuditResourceType is the type of the resource that was affected by the
+// audited event.
+enum AuditResourceType {
+  // AUDIT_RESOURCE_TYPE_UNSPECIFIED is the default type for audited events.
+  AUDIT_RESOURCE_TYPE_UNSPECIFIED = 0;
+  // AUDIT_RESOURCE_TYPE_USER is the type for when the audited event affected a
+  // user.
```
stefanvanburen, line 44: complete nit: I might just nix these comments as they aren't adding much on top of the field names :-)

proto/buf/alpha/registry/v1alpha1/audit_event.proto
```diff
@@ -0,0 +1,96 @@
...
+  // AUDIT_ACTOR_TYPE_USER is the actor type for when the audited event is
+  // triggered by a user. This is the most common case.
+  AUDIT_ACTOR_TYPE_USER = 1;
+}
+
+// AuditActor is the actor who triggered or executed the audited event.
+message AuditActor {
+  // Type of actor who caused the audited event.
+  AuditActorType type = 1;
+  // Id of the actor who eecuted the action.
```
stefanvanburen, line 34:
```suggestion
  // Id of the actor who executed the action.
```

stefanvanburen (commented):
lgtm, just some minor notes

proto/buf/alpha/registry/v1alpha1/audit_event.proto
unmultimedio: I guess at that time I was drafting w/out fine grain detail. Now writing the actual message I see not much use on adding a `name`, since we might have resources or actors that don't have a `"name"` (think of a system admin at some point?). Actors right now [are always users](https://www.notion.so/bufbuild/Audited-Events-List-db1e266e05ce4d0db48638ec6091b9b2), and resources most of the times also have names (users, organizations, repos, plugins), except for members. Those cases, I'd say it's fine to use the usernames, but it seemed more appropriate to include that in the actual event payload.

I see it could be useful for [searching and filtering later](https://www.notion.so/bufbuild/Audit-Logging-e96b318c64224cd8b2a34887fcacc54a#42318760b882432eb08ff5ebce5cb01f) by names, so they could live in a specific column, but that's also possible with a fuzzy search.

I'm in doubt, so I preferred not to add it. wdyt?

proto/buf/alpha/registry/v1alpha1/audit_event.proto
unmultimedio: Agree, although that would leave the BSR docs w/out any description. Not sure if that's acceptable or if there's an unwritten rule of always adding comments to proto files. Would you remove just the "UNSPECIFIED" comments? or all of them?

proto/buf/alpha/registry/v1alpha1/audit_event.proto
stefanvanburen: yeah, I doubt it's necessary too - just thought I'd double-check as I saw that in the doc. 👍

proto/buf/alpha/registry/v1alpha1/audit_event.proto
stefanvanburen: I'd probably remove all; just a quick look at some of our other enums leads me to believe that we generally don't comment each value and instead generally just comment the enum itself. (See https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/role.proto, for example).

Admittedly I haven't seen best practices documented around this, but I think this is pretty in-line with what we're doing currently. Definitely subjective, though.

unmultimedio: Removed

proto/buf/alpha/registry/v1alpha1/audit_event.proto
unmultimedio: Fixed

proto/buf/alpha/registry/v1alpha1/audit_event.proto
```diff
@@ -0,0 +1,77 @@
...
+  // Unique id of the audited event.
+  string id = 1;
+  // Type of the audited event. It specifies "what" happened.
+  AuditEventType type = 2;
+  // Actor of the audited event. It specifies "who" did it.
+  AuditActor actor = 3;
+  // Resource of the audited event. It specifies "which resouce" was affected.
+  AuditResource resource = 4;
+  // Timestamp of the audited event. It specifies "when" it happened.
+  google.protobuf.Timestamp timestamp = 5;
```
unmultimedio, line 74: Other messages use to have `create_time`, but because they also have `update_time`. This won't be ever updated, so not sure if we're ok drifting from `create_time` to `timestamp`?

stefanvanburen: good point - there's a few that only have `create_time` ([ModulePin](https://github.com/bufbuild/buf/blob/ce0727633e9e3e4d1ac8591c1c8163d6dab4c4cd/proto/buf/alpha/module/v1alpha1/module.proto#L73), for example). I'd probably change to `create_time` for consistency. The only one that isn't `{create,update,expire}_time` is `event_time` from event.proto (at least, within ./proto/buf/alpha).

unmultimedio: I'll rename to `create_time` to be consistent.

stefanvanburen (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -0,0 +1,77 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.audit.v1alpha1;
```
unmultimedio, line 17: Moved to its own separated package, just like it was on the previous implementation. It allowed me to remove all the `Audit*` prefixes. cc @robbertvanginkel @stefanvanburen

unmultimedio: probably we want to make this stable (`v1`) when we finish implementing the first list of the proposed audited events(?)

renahlee (approved):

private/gen/proto/go/buf/alpha/audit/v1alpha1/event.pb.go
```diff
@@ -0,0 +1,576 @@
...
+	_ = protoimpl.EnforceVersion(20 - protoimpl.MinVersion)
+	// Verify that runtime/protoimpl is sufficiently up-to-date.
+	_ = protoimpl.EnforceVersion(protoimpl.MaxVersion - 20)
+)
+
+// ActorType is the type of actor that caused the audited event.
+type ActorType int32
+
+const (
+	ActorType_ACTOR_TYPE_UNSPECIFIED ActorType = 0
```
njhale, line 42: Looks like the `ACTOR_TYPE_` prefix in `proto/buf/alpha/audit/v1alpha1/event.proto` is creating a stutter here. Is this a common convention, or should we drop it?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -0,0 +1,77 @@
...
+
+import "google/protobuf/timestamp.proto";
+
+// ActorType is the type of actor that caused the audited event.
+enum ActorType {
+  ACTOR_TYPE_UNSPECIFIED = 0;
+  ACTOR_TYPE_USER = 1;
```
njhale, line 24: it looks like the `ACTOR_TYPE_` prefix [causes stutter downstream](https://github.com/bufbuild/buf/pull/1366/files#r959129560)

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: > probably we want to make this stable (v1) when we finish implementing the first list of the proposed audited events(?)

on it's own, that sounds reasonable. 

though I think we should still double check to make sure there aren't any additional considerations before moving to `v1` (e.g. support level).

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -0,0 +1,77 @@
...
+  // Type of actor who caused the audited event.
+  ActorType type = 1;
+  // Id of the actor who caused the audited event.
+  string id = 2;
+}
+
+// ResourceType is the type of the resource that was affected by the audited
+// event.
+enum ResourceType {
+  RESOURCE_TYPE_UNSPECIFIED = 0;
```
njhale, line 38: Similar to https://github.com/bufbuild/buf/pull/1366/files#r959129704

private/gen/proto/go/buf/alpha/audit/v1alpha1/event.pb.go
stefanvanburen: ye, it's just an unfortunate fallout of:

> Enum value names should be prefixed with the UPPER_SNAKE_CASE of the enum name. [Why?](https://docs.buf.build/lint/rules#enum_value_prefix)

https://docs.buf.build/best-practices/style-guide#enums

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: yeah totally, that'd be at the end of the implementation of the current mvp

private/gen/proto/go/buf/alpha/audit/v1alpha1/event.pb.go
njhale: ah, cool, okay 🙃 

thanks

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: Resolved by https://github.com/bufbuild/buf/pull/1366#discussion_r959546155

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: Resolved by https://github.com/bufbuild/buf/pull/1366#discussion_r959546155

PR #1367, opened 2022-08-29
create a separate options field for registry

pkwarren:
There are some options which we'll always want to specify when
generating code via the remote registry (Go/NPM/etc.). These options
will override any `default_opts` set in buf.plugin.yaml.

This will allow us to create new options for behavior specific to
working with the remote registry without causing all generated code
(local and via remote plugins) to change.


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -300,7 +300,7 @@ func run(
 		bufplugin.PluginToProtoPluginRegistryType(plugin),
 		plugin.Version(),
 		plugin.ContainerImageDigest(),
-		bufplugin.PluginOptionsToOptionsSlice(plugin.DefaultOptions()),
+		bufpluginconfig.PluginOptionsToOptionsSlice(plugin.DefaultOptions()),
```
pkwarren, line 303: Moved these utility methods to `bufpluginconfig` to avoid circular dependencies. They make more sense to live there since that's where we convert `[]string` options to/from `map[string]string`.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -93,6 +94,24 @@ type Config struct {
...
+	// Note that some legacy plugins don't always express their options
+	// as key value pairs. For example, protoc-gen-java has an option
+	// that can be passed like so:
+	//
+	//  java_opt=annotate_code
+	//
+	// In those cases, the option value in this map will be set to
+	// the empty string, and the option will be propagated to the
+	// compiler without the '=' delimiter.
+	Options map[string]string
```
pkwarren, line 114: Felt it made sense to add this at the top-level.

private/bufpkg/bufplugin/bufpluginconfig/testdata/success/go/buf.plugin.yaml
```diff
@@ -14,3 +14,5 @@ registry:
     deps:
       - module: google.golang.org/grpc
         version: v1.32.0
+  opts:
+    - separate_package=true
```
pkwarren, line 18: This is an example of a real option we'd use for go-grpc remote generation.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -69,6 +69,14 @@ message RegistryConfig {
     GoConfig go_config = 1;
     NPMConfig npm_config = 2;
   }
+  // Reserved for future remote registry types.
+  reserved 3 to 9;
```
pkwarren, line 73: Felt nice to group together the different registry config fields (since the oneof above will grow over time) but I can remove this and just set the options field number to 3 if preferred.

srikrsna-buf (approved):

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -93,6 +94,24 @@ type Config struct {
 type RegistryConfig struct {
 	Go  *GoRegistryConfig
 	NPM *NPMRegistryConfig
+	// Options is the set of options passed into the plugin for the
+	// remote registry. It overrides any DefaultOptions setting in
```
mfridman, line 98: I think `overrides` isn't correct here, because we're not going to use the `DefaultOptions` from the top-level config. It'll either use the `registry.opts` in the registry config, or none at all. There is nothing being overridden, right?

mfridman (approved):
Nit on comment, otherwise lgtm.

PR #1368, opened 2022-08-30
add more metadata on repository, organization, and user

cyinma:
fixes BSR-571

for the metadata on curated plugin, the only thing that needed to add would be license, and I would do it in a separated follow up PR so that we can also focus on necessary changes around `buf.plugin.yaml`


proto/buf/alpha/registry/v1alpha1/veracity.proto
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.registry.v1alpha1;
+
+enum Veracity {
```
bufdev, line 19: Needs docs on what this is. Also is veracity the right word here?

proto/buf/alpha/registry/v1alpha1/veracity.proto
```diff
@@ -0,0 +1,23 @@
...
+  VERACITY_UNSPECIFIED = 0;
+  VERACITY_BUF_OFFICIAL = 1;
```
bufdev, line 21: Values need docs as well. Also I might consider something other than `BUF_OFFICIAL` - perhaps just `OFFICIAL`?

Also should this be in the registry package? Are we going to want to use this in the module package, for example?

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -30,6 +31,12 @@ message User {
   string username = 4;
   // mutable
   bool deactivated = 5;
+  // description is the user configurable description of the user.
+  string description = 6;
+  // url is the user configurable URL in the description of the user.
```
bufdev, line 36: Will this url always include the scheme, will it never include the scheme, any other constraints? Document all constraints.

private/pkg/spdx/cmd/spdx-proto-data/main.go
```diff
@@ -0,0 +1,114 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
```
cyinma, line 1: I was thinking about should I check in this tool, as it is pretty one-off and didn't expect to regenerate with newer version of SPDX list, as the enum will break because of the new licenses (we might make this tool to handle that case, but it doesn't seems like that is something we care for now? the last update of the version here is more than a year ago), but I think it's still good to have some review on this, especially on the replacer, and we can exclude this if we have thoughts about that

proto/buf/alpha/registry/v1alpha1/veracity.proto
cyinma: I was trying to find the most accurate word for this, as `verification` doesn't seems accurate enough? any thoughts?

proto/buf/alpha/registry/v1alpha1/veracity.proto
cyinma: updated docs and to `OFFICIAL`.
I don't see that we will use it in the module package or other packages, as I think the furthest place that we plan to use it is in repository level, and probably not going indicate these status in a module

proto/buf/alpha/registry/v1alpha1/user.proto
cyinma: updated

proto/buf/alpha/registry/v1alpha1/veracity.proto
stefanvanburen: maybe ["provenance"](https://en.wiktionary.org/wiki/provenance)?

> 1. Place or source of origin.
> 4. ([computing](https://en.wiktionary.org/wiki/computing#Noun)) The copy history of a piece of data, or the intermediate pieces of data utilized to compute a final data element, as in a database record or web site (data provenance)

similar to "veracity" though, I don't think it's a common word. Maybe just "verification status" for simplicity?

cyinma: updated to `VerificationStatus` as I think it's the easiest one to understand for now, but happy to change if we can come up with a better name

private/pkg/spdx/cmd/spdx-proto-data/main.go
robbertvanginkel: I think like you point out, managing this list over time becomes difficult as you have to keep existing identifiers on their proto tags and can't just re-run this script. Any update would have to undo the stringing to get back the identifier (or take them from the comments and any reserved fields), match that against the new list, change the missing ones into "reserved", add new ones to the end.

All of this gives us strict enums over the wire, but it seems like both the frontend and the backend will have to map them back and forth to the spdx identifier strings (string in yaml file, some enum on the frontend to get the full name, etc). 

Given that the strings are well identified according to some external specification, maybe it will be more straightforward to not type this as an enum and use a the well defined identifiers as string?

cyinma: I think it's sane to use the identifier string over the wire, consider that both end need to map the enums back to string to make it usable, and it would be much easier for us to manage the list.
we might need to do some validation in both end but we can defer this implementation detail later.
updated.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +47,23 @@ message Repository {
...
+  // either a username or organization name.
+  string owner_name = 10;
+  // description is the user configurable description of the repository.
+  string description = 11;
+  // url is the user configurable URL in the description of the repository,
+  // always included the scheme and will not have a #fragment suffix.
+  string url = 12;
+  // license is the user configurable license of the repository, which should be one of
+  // the identifier defined in https://spdx.org/licenses
+  string license = 13;
```
robbertvanginkel, line 60: How are custom non-spdx licenses represented here?

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +47,23 @@ message Repository {
   bool deprecated = 8;
   // deprecation_message is the message shown if the repository is deprecated.
   string deprecation_message = 9;
+  // owner_name is the name of the owner of the repository,
+  // either a username or organization name.
+  string owner_name = 10;
```
robbertvanginkel, line 52: Why do we need to start including the owner name here?

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +47,23 @@ message Repository {
...
+  string url = 12;
+  // license is the user configurable license of the repository, which should be one of
+  // the identifier defined in https://spdx.org/licenses
+  string license = 13;
+  // license url is the user configurable url to the license text.
+  string license_url = 14;
+  // verification status of the owner of the repository, configurable by server admin.
+  VerificationStatus verification_status = 15;
+  // last commit time in main branch of the repository.
+  google.protobuf.Timestamp last_commit_time = 16;
```
robbertvanginkel, line 66: These two properties don't feel like they belong on the registry repository object, why are we adding them here rather than expecting the user to get them through the owner/listing commits?

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +47,23 @@ message Repository {
...
+  // always included the scheme and will not have a #fragment suffix.
+  string url = 12;
+  // license is the user configurable license of the repository, which should be one of
+  // the identifier defined in https://spdx.org/licenses
+  string license = 13;
+  // license url is the user configurable url to the license text.
+  string license_url = 14;
+  // verification status of the owner of the repository, configurable by server admin.
+  VerificationStatus verification_status = 15;
+  // last commit time in main branch of the repository.
```
robbertvanginkel, line 65: nit: branches arent a bsr concept anymore, if we include this should it be something like "last main commit"?

proto/buf/alpha/registry/v1alpha1/verification_status.proto
```diff
@@ -0,0 +1,27 @@
...
+
+package buf.alpha.registry.v1alpha1;
+
+// VerificationStatus is the verification status of an owner on if we recognize them,
+// an owner can be either user or organization.
+enum VerificationStatus {
+  VERIFICATION_STATUS_UNSPECIFIED = 0;
+  // OFFICIAL indicates that the owner is maintained by Buf.
+  VERIFICATION_STATUS_OFFICIAL = 1;
+  // VERIFIED_PUBLISHER indicates that the owner is a third-party that verified by Buf.
```
robbertvanginkel, line 25:
```suggestion
  // VERIFIED_PUBLISHER indicates that the owner is a third-party that has been verified by Buf.
```

proto/buf/alpha/registry/v1alpha1/repository.proto
cyinma: we can just do "custom", added to the comment

proto/buf/alpha/registry/v1alpha1/repository.proto
cyinma: this is kind of a preparation for repositories recommendation, where this is needed in order to link to the repository page by just the repository message.
we can exclude this from this PR but I think it makes sense to include it here while we are adding fields to repository, wdyt?

proto/buf/alpha/registry/v1alpha1/repository.proto
cyinma: make sense to me, but I assume frontend folks would like to prevent 2 more calls just for rendering the repository page? (cc @timostamm @paul-sachs ), how do you think if we add them to `GetRepositoryResponse` and `GetRepositoryByFullNameResponse`, just like repository counts?

proto/buf/alpha/registry/v1alpha1/repository.proto
robbertvanginkel: I think we should avoid adding our own special value here, if spdx doesn't have a placeholder value but they may in the future I'd rather not guess.

How about we rename this `license_spdx_identifier`, and if it is empty but there is a license url we count that as custom?

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -46,6 +47,23 @@ message Repository {
...
+  // description is the user configurable description of the repository.
+  string description = 11;
+  // url is the user configurable URL in the description of the repository,
+  // always included the scheme and will not have a #fragment suffix.
+  string url = 12;
+  // license is the user configurable license of the repository, which should be one of
+  // the identifier defined in https://spdx.org/licenses
+  string license = 13;
+  // license url is the user configurable url to the license text.
+  string license_url = 14;
```
robbertvanginkel, line 63: Is license not going to be a license file someone pushes, but a url they have to specify? What if the license of a repo/proto changes over time?

proto/buf/alpha/registry/v1alpha1/repository.proto
robbertvanginkel: I think we should avoid customizing the generic `buf.alpha.registry.v1alpha1.Repository` message with display fields that are meant for only certain usages of our frontend. Owner name seems still generically useful, but related to the comments below verification status of the owner or commit name seem more separate.

proto/buf/alpha/registry/v1alpha1/repository.proto
cyinma: do you mean we should read the LICENSE file on `buf push` instead of making it selectable by user?

robbertvanginkel: If we want to have docs on pkg.go.dev, the generated packages zips need to have a license file in them. At least in the goproxy, license is a per commit thing and not a repository wide setting. If a user would change their license for new commits, I'm doubtful that we should generate zips with this newer license for commits that were pushed earlier.

proto/buf/alpha/registry/v1alpha1/repository.proto
cyinma: discussed offline, we are going to remove verification status and last commit time from `Repository`, and might discuss on some endpoints that is more frontend specific later

proto/buf/alpha/registry/v1alpha1/repository.proto
timostamm: For the frontend, a single endpoint that provides all necessary information to render the page would be desirable. It is much less overhead on the wire, and less code to reason about when working on the actual UI.  
This becomes a more serious limitation in list views: Without last commit time and verification status embedded in the message, it is simply not feasible to show the information in lists. Without an owner name, it is not possible to link to the repository (because we would want canonical URLs with names).

It makes perfect sense to me to understand Repository as the normalized representation of the entity. The verification status obviously belongs on the object representing the owner from that POV, and adding it to both entities adds ambiguity. 

This looks like two opposing goals to me, and I'm not sure there is a simple answer. Compared to popular public APIs, the toll does seem quite high for the FE.

cyinma: @robbertvanginkel and I have discussed on this offline [previously](https://github.com/bufbuild/buf/pull/1368#discussion_r968887617), and we think we might end up needing something frontend specific. But that will need a separated discussion and I think we can move this forward for now

robbertvanginkel (approved):

PR #1371, opened 2022-08-31
Add plugin languages to curated plugin definition

mfridman:
This PR adds plugin languages to the plugin definitions, with more to come. The goal isn't to capture every language, just those we've built plugins for.

Relevant prior discussions:

https://github.com/bufbuild/buf/pull/1314
https://github.com/bufbuild/buf/pull/1316#discussion_r936961385


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -27,6 +27,18 @@ enum PluginRegistryType {
   PLUGIN_REGISTRY_TYPE_NPM = 2;
 }
 
+// PluginLang is used to specify the languages a plugin supports.
+enum PluginLang {
```
mfridman, line 31: We could name this `PluginLanguage` instead of this abbreviated `Lang` format. And use `language` in other places instead of `lang`, thoughts?

mfridman: Let's just use `Language/language` throughout for consistency.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -27,6 +27,18 @@ enum PluginRegistryType {
...
+  PLUGIN_LANG_UNSPECIFIED = 0;
+  PLUGIN_LANG_GO = 1;
+  PLUGIN_LANG_JAVASCRIPT = 2;
+  PLUGIN_LANG_TYPESCRIPT = 3;
+  PLUGIN_LANG_SWIFT = 4;
+  PLUGIN_LANG_CPP = 5;
+  PLUGIN_LANG_JAVA = 6;
+  PLUGIN_LANG_DART = 7;
```
pkwarren: Should add Kotlin, Ruby, ObjC, C#, PHP, Python, Rust (to come).

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -120,6 +139,8 @@ message CuratedPlugin {
   RegistryConfig registry_config = 12;
   // The revision for this plugin version.
   uint32 revision = 13;
+  // The languages supported by the plugin.
+  repeated PluginLanguage languages = 14;
```
cyinma, line 143: how do you think if we explicitly call this `output_languages`?

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -90,6 +93,28 @@ func PluginToProtoPluginRegistryType(plugin Plugin) registryv1alpha1.PluginRegis
 	return registryType
 }
 
+// LanguagesToProtoLanguages determines the appropriate registryv1alpha1.PluginRegistryType for the plugin.
+func LanguagesToProtoLanguages(languages []string) []registryv1alpha1.PluginLanguage {
+	languageToEnum := make(map[string]registryv1alpha1.PluginLanguage)
+	for pluginLanguageKey, pluginLanguage := range registryv1alpha1.PluginLanguage_value {
+		pluginLanguageKey := strings.TrimPrefix(pluginLanguageKey, "PLUGIN_LANGUAGE_")
```
mfridman, line 100: This might be considered brittle, and we might want to have an explicit string to proto map. But I'd argue the enum name is unlikely to change and it avoids having to remember to update a map somewhere.

Also, this doesn't fail with an error if the language is unknown. We probably don't want to block plugin uploads because a language isn't supported? Although we could just omit the language until it's added and be more explicit here.

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -90,6 +93,28 @@ func PluginToProtoPluginRegistryType(plugin Plugin) registryv1alpha1.PluginRegis
 	return registryType
 }
 
+// LanguagesToProtoLanguages determines the appropriate registryv1alpha1.PluginRegistryType for the plugin.
+func LanguagesToProtoLanguages(languages []string) []registryv1alpha1.PluginLanguage {
+	languageToEnum := make(map[string]registryv1alpha1.PluginLanguage)
```
pkwarren, line 98:
```suggestion
	languageToEnum := make(map[string]registryv1alpha1.PluginLanguage, len(registryv1alpha1.PluginLanguage_value)
```

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -90,6 +93,28 @@ func PluginToProtoPluginRegistryType(plugin Plugin) registryv1alpha1.PluginRegis
...
+	for pluginLanguageKey, pluginLanguage := range registryv1alpha1.PluginLanguage_value {
+		pluginLanguageKey := strings.TrimPrefix(pluginLanguageKey, "PLUGIN_LANGUAGE_")
+		pluginLanguageKey = strings.ToLower(pluginLanguageKey)
+		// Example:
+		// { go: 1, javascript: 2 }
+		languageToEnum[pluginLanguageKey] = registryv1alpha1.PluginLanguage(pluginLanguage)
+	}
+	protoLanguages := []registryv1alpha1.PluginLanguage{}
+	for _, language := range languages {
+		if pluginLanguage, ok := languageToEnum[language]; ok {
```
pkwarren, line 108: What should we do here if we don't find a language? Should this show up as an error to indicate we were only able to map some of them to the enum type?

pkwarren (approved):

private/bufpkg/bufplugin/bufplugin.go
mfridman: Ye, you're probably right. We'll want to error here, so if a user (us) refer to an unknown language it'll fail right away.

And then we'll have to file an issue in this repository to add that language.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
mfridman: Sure.

private/bufpkg/bufplugin/bufplugin.go
mfridman: Add a useful error message here:

```
invalid plugin output language: "cobol"
supported languages: cpp, csharp, dart, go, java, javascript, kotlin, objective_c, php, python, ruby, rust, swift, typescript
```

pkwarren (approved):

PR #1372, opened 2022-08-31
format: Handle special escaped characters

### LINKED ISSUE #1370: latest docker version buf formats text into invalid protobuf
The `latest` docker version buf will format text with escaped quote and line break into invalid protobuf file.

Here is a minimal sample to reproduce the issue:

original protobuf:
```protobuf
syntax = "proto3";

package bug.report.v1;

import "protoc-gen-openapiv2/options/annotations.proto";

option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
  info: {
    description: "This is \"title\"\nThis is second line \"aaaaa\"";
  };
};
```
After running `docker run -v $(PWD):/home/ubuntu/ -w /home/ubuntu bufbuild/buf:latest format -w`, the resulting protobuf is:
```protobuf
syntax = "proto3";

package bug.report.v1;

import "protoc-gen-openapiv2/options/annotations.proto";

option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
  info: {
    description: 'This is "title"
This is second line "aaaaa"';
  };
};
```
This issue is not seen on `1.7.0` version.

I'm not sure what is the real version of this `latest` docker container. It is only tagged as `latest`, and having different hash with `1.7.0` container. This is the link to the docker hub of this release: [link](https://hub.docker.com/layers/buf/bufbuild/buf/latest/images/sha256-e904205fb7b99dd4ed65d0a3243b36aefffbe9b6223a7608a26671e4914f4996?context=explore)

amckinney:
Fixes #1370

This adjusts the change made to fix https://github.com/bufbuild/buf/issues/1093 (which has not yet been released), so that we always preserve the raw string specified in the original source.

Note that we don't need a new `CHANGELOG.md` entry here because it was already introduced in #1332.

### DIFF
--- a/private/buf/bufformat/testdata/proto2/option/v1/option_message_field.golden.proto
+++ b/private/buf/bufformat/testdata/proto2/option/v1/option_message_field.golden.proto
@@ -20,7 +20,7 @@ message Test {
   repeated Simple r = 4;
   map<string, int32> m = 5;
 
-  optional bytes b = 6 [default = "\x00\x01\x02\x03\x04\x05\x06\afubar!"];
+  optional bytes b = 6 [default = "\0\1\2\3\4\5\6\7fubar!"];
 
   extensions 100 to 200;
 
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -1442,73 +1442,11 @@ func (f *formatter) writeSpecialFloatLiteral(specialFloatLiteralNode *ast.Specia
 }
 
 // writeStringLiteral writes a string literal value (e.g. "foo").
+// Note that the raw string is written as-is so that it preserves
+// the quote style used in the original source.
 func (f *formatter) writeStringLiteral(stringLiteralNode *ast.StringLiteralNode) {
-	f.writeStringWithStyle(stringLiteralNode.Val)
-}
-
-// writeStringValue writes a string value (e.g. "foo").
-func (f *formatter) writeStringValue(stringValueNode ast.StringValueNode) {
-	f.writeStringWithStyle(stringValueNode.AsString())
-}
-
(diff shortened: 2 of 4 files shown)


jhump:
@amckinney, leaf nodes all have a `Token()` method. You can use that with the `TokenInfo()` method on the file to get back details about the token's representation in the original source. These include the location of the token in the file, leading whitespace and comments, and also the original [raw text](https://github.com/jhump/protocompile/blob/204cda3a10b26b7e7f7c224524cb7d7056ab203b/ast/file_info.go#L294).

amckinney:
> @amckinney, leaf nodes all have a Token() method. You can use that with the TokenInfo() method on the file to get back details about the token's representation in the original source. These include the location of the token in the file, leading whitespace and comments, and also the original [raw text](https://github.com/jhump/protocompile/blob/204cda3a10b26b7e7f7c224524cb7d7056ab203b/ast/file_info.go#L294).

Awesome! This should actually be fairly straightforward then - we should be able to simplify a lot of this and always just use the `RawText` method.

private/buf/bufformat/testdata/proto2/option/v1/option_message_field.golden.proto
```diff
@@ -20,7 +20,7 @@ message Test {
   repeated Simple r = 4;
   map<string, int32> m = 5;
 
-  optional bytes b = 6 [default = "\x00\x01\x02\x03\x04\x05\x06\afubar!"];
+  optional bytes b = 6 [default = "\0\1\2\3\4\5\6\7fubar!"];
```
amckinney, line 23: This actually uncovered a small edge case from before - the original input was unnecessarily transformed, but now it's preserved.

jhump (approved):
Sweet!

PR #1375, opened 2022-09-01
Add organization created audit log event

stefanvanburen:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -73,5 +74,10 @@ message Event {
...
   google.protobuf.Timestamp event_time = 5;
   // Payload of the audited event. It specifies additional context on the event.
-  // TODO add oneof payload events.
+  oneof payload {
+    OrganizationCreatedPayload organization_created = 6;
+  }
 }
+
+// OrganizationCreatedPayload is the payload for an ORGANIZATION_CREATED event.
+message OrganizationCreatedPayload {}
```
stefanvanburen, line 83: I think it's fine to add this even if it's empty.

We could include the organization name?

stefanvanburen: (also, feel free to suggest a better naming scheme for this or the `oneof`)

unmultimedio: Yeah, `name` and `create_time` seem useful (`update_time` I feel adds no value for this event).

stefanvanburen: is `create_time` the same as `timestamp`?

njhale: > is create_time the same as timestamp?

I'm assuming this represents when the event was actually written to whatever store it was accessed from.

if that's the case, then: 

> Yeah, name and create_time seem useful (update_time I feel adds no value for this event).

is it important to an audit log consumer know _when_ an event was written to `$STORE`? seems like an implementation detail (and I expect that would differ between `StdoutStore`, `GCSStore`, `PostgresStore`, etc...).

unmultimedio: I'd include `name` and `create_time`, even if [it's exactly the same value as `event_timestamp`](https://github.com/bufbuild/core/pull/5867#discussion_r961132030), just to be explicit.

stefanvanburen: We could add both (and sorry, I misread, it's `event_time`, not `timestamp`). 

Maybe I'll clarify the docs that `event_time` is when the event is ... enqueued? recorded? (Not sure it matters, which is perhaps why I thought `event_time` could, in this case for example, be `Organization.created_at`, with no other timestamps.) wdyt @unmultimedio @njhale?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -73,5 +74,13 @@ message Event {
   // Time of the audited event. It specifies "when" it happened.
   google.protobuf.Timestamp event_time = 5;
   // Payload of the audited event. It specifies additional context on the event.
-  // TODO add oneof payload events.
+  oneof payload {
```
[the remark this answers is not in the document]
stefanvanburen: I'd prefer to [stay away from nested messages](https://docs.buf.build/best-practices/style-guide#recommendations) if we can. Un-nesting, I suppose we could do:

```proto
message EventMetadata {
  string event_id
  EventType type
  Actor actor
  Resource resource
  google.protobuf.Timestamp event_time
}

message OrganizationCreatedEvent {
  EventMetadata metadata
  
  string name
}

// ... other events in a similar vein

message Event {
  oneof event {
    OrganizationCreatedEvent organization_created
  }
}
```

I think I like the original approach a little better because it reduces the nesting to the common fields, which I think cover most of what we're looking to capture.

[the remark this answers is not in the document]
unmultimedio: I prefer too the original approach. The advantage from both of these approaches is that we could force event emitters to send exactly the payload we expect. But to be honest, we could do that also with the original approach, receiving common fields as separated params, and the `Payload<Event>` as the last param, and the logger builds it back into an `AuditEvent`. I thought we agreed on keeping the original approach for now.

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: (sorry, added that comment before I saw @unmultimedio's). I think that's reasonable. I'll add those fields.

unmultimedio: And now that `name` is back, we could have that name being the FQN, and still have separate fields for those in the payload?

```
{
  "event_time": "same as created_at",
  "actor" {
    "type": "USER",
    "name": "johndoe",
  },
  "resource" {
    "type": "ORGANIZATION",
    "name": "buf.build/acme", // for repositories/plugins/commits this can still be its FQN
  },
  "payload_organization_created": {
    "remote": "buf.build",
    "name": "acme",
    "owner": "johndoe", // do we have this?
    "created_at": "timestamp",
  }
}
```

stefanvanburen: added in [b14b34e](https://github.com/bufbuild/buf/pull/1375/commits/b14b34e772acfca2cd58bcfbf3652279ce2d8ddb)

stefanvanburen: yeah, I think FQN is best when available. (so probably no need for `"remote"` in this case? I think `"owner"` is implied?)

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -73,5 +74,15 @@ message Event {
...
+    PayloadOrganizationCreated organization_created = 6;
+  }
+}
+
+// PayloadOrganizationCreated is the payload for an ORGANIZATION_CREATED event.
+message PayloadOrganizationCreated {
+  // name is the name of the created organization.
+  string name = 1;
```
unmultimedio, line 85: Oh sorry, [I meant adding it back to the `actor` and `resource` messages](https://github.com/bufbuild/core/pull/5867#discussion_r961678823), not to each specific payloads. 😅

stefanvanburen: ah, yeah, I can add it there. sorry, not enough coffee yet 😄

stefanvanburen: Done in https://github.com/bufbuild/buf/pull/1375/commits/faa557bd7bcd70360f3a0c6e16aa678896240126!

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -68,10 +73,18 @@ message Event {
   EventType type = 2;
   // Actor of the audited event. It specifies "who" did it.
   Actor actor = 3;
-  // Resource of the audited event. It specifies "which resouce" was affected.
+  // Resource of the audited event. It specifies "which resource" was affected.
```
unmultimedio, line 76: 👓

unmultimedio (approved):
🚀

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: for context, looked at FQN with @unmultimedio and decided against it for now in favor of just attaching names where we think it's reasonable (for example, on a "repository create", perhaps we'll add `organization_id` / `organization_name`.)

proto/buf/alpha/audit/v1alpha1/event.proto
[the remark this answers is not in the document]
njhale: > I'd prefer to [stay away from nested messages](https://docs.buf.build/best-practices/style-guide#recommendations) if we can.

the nested message definitions in my example weren't really related to the point I was trying to make, if that was the hangup -- sorry.

```
message OrganizationCreatedEvent {
  EventMetadata metadata
  
  string name
}

message Event {
  oneof payload {
    OrganizationCreatedEvent organization_created
  }
}
```

going with ^, I was mostly wondering about our preference in access pattern for fields

the original:
- accessing common metadata: `event.`
- accessing event-specific data: `event.payload.organization_created.`

vs. proposed:
- accessing common metadata: `event.payload.metadata.`
- accessing event-specific data: `event.payload.organization_created.`

the proposed puts metadata and "additional context" at the same "level" in the field hierarchy.

> I thought we agreed on keeping the original approach for now.

@unmultimedio for sure; I'm just thinking out in the open. Let me know if you feel like it's becoming too noisy or otherwise not adding value.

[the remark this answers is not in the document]
unmultimedio: @njhale sure, let's keep throwing ideas here!

So what I see from whatever approach we take we should make sure to:
- Have a single type that includes it all for the `rpc ListAuditEvents() []AuditEvent`.
- Force type safety for loggers, to avoid mismatch in the `event_type` and the actual `payload.<SomeEvent>`.

With the current approach, I think we could accomplish both. We already have:

```proto
message Event {
  EventType event_type = 1;
  EventActor actor = 2;
  EventResource resource = 3;
  Time event_time = 4;
  oneof payload {
    PayloadOrganizationCreated organization_created = 5;
    // ...
  }
}
```

And the callers currently do:

```go
ev := audit.Event{
  ...
  // the risk is we could have a mismatch here
  EventType: audit.ORG_CREATED,
  OrganizationCreated: audit.PayloadOrganizationCreated{...},
}
s.auditLogger(ctx, ev)
```

So we could force them with many interfaces just as you proposed a few days ago, similar to:

```go
type OrganizationLogger interface {
  func LogOrganizationCreated(
    context.Context,
    // all the commons (see there's no EventType, it's implied and it would be hydrated in the logger)
    audit.Actor,
    audit.Resource,
    // then the specific payload
    audit.PayloadOrganizationCreated,
  )
  // ... other func LogOrganization<OtherEvents>()
}

// ... other interfaces, one per service?
```

and the caller won't have anymore the possibility to mix things up:

```go
payload := audit.PayloadOrganizationCreated{...}
s.organizationLogger.LogOrganizationCreated(ctx, actor, resource, payload)
```

wdyt?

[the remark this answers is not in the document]
stefanvanburen: could we do:

```go

  func Log(
    context.Context,
    audit.Actor,
    audit.Resource,
    audit.Payload, // interface
  )
```

a little more difficult for callers but makes it harder to make mistakes? I'd prefer not to add a bunch of separate interfaces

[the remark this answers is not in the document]
robbertvanginkel: Looking at the binary/json representation I think we should stick with the metadata in the envelope rather than in the oneof payloads:
```
{"organization_created": {"metadata": {"id": ..., "actor": ...}}, "name": "foo"}}
{"organization_deleted": {"metadata": {"id": ..., "actor": ...}}, "name": "foo"}}
```
vs
```
{"id": ..., "actor": ..., "type": "ORG_CREATED", "organization_created": {"name": "foo", ...}}
{"id": ..., "actor": ..., "type": "ORG_DELETED", "organization_deleted": {"name": "foo", ...}}
```
I think the bottom is easier to grep/jq/groq through if they wanted a text dump.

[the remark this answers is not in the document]
unmultimedio: > a little more difficult for callers but makes it harder to make mistakes? I'd prefer not to add a bunch of separate interfaces

@stefanvanburen you mean like the current catch-all interface we have now? I don't see how we can force the correct payload from the caller side with that interface, do you have something in mind?

PR #1376, opened 2022-09-02
Add audit events

stefanvanburen:
Now that we're somewhat decided on shared code, we can add the rest of the events that we initially want to log. Minor notes below.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -61,6 +62,25 @@ message Resource {
...
+  EVENT_TYPE_REPOSITORY_MEMBER_ADDED = 9;
+  EVENT_TYPE_REPOSITORY_MEMBER_ROLE_CHANGED = 10;
+  EVENT_TYPE_REPOSITORY_MEMBER_REMOVED = 11;
+  EVENT_TYPE_REPOSITORY_VISIBILITY_CHANGED = 12;
+  EVENT_TYPE_PLUGIN_CREATED = 13;
+  EVENT_TYPE_PLUGIN_DELETED = 14;
+  EVENT_TYPE_PLUGIN_PUSHED = 15;
+  EVENT_TYPE_USER_REGISTERED = 16;
+  EVENT_TYPE_USER_DELETED = 17;
+  EVENT_TYPE_USER_DEACTIVATED = 18;
```
stefanvanburen, line 81: split out deleted / deactivated. I guess this could just be a boolean on one of the events, i.e. `is_deleted` on `message PayloadUserDeactivated`, but figured this was fine. happy to change it.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+    PayloadPluginPushed plugin_pushed = 19;
+    PayloadUserRegistered user_registered = 20;
+    PayloadUserDeactivated user_deactivated = 21;
+    PayloadUserDeleted user_deleted = 22;
+    PayloadUserLoggedIn user_logged_in = 23;
+    PayloadUserLoggedOut user_logged_out = 24;
   }
 }
 
-// PayloadOrganizationCreated is the payload for an ORGANIZATION_CREATED event.
```
stefanvanburen, line 86: chose to go comment-less on the names for these payloads as I think they're somewhat self-documenting, but happy to add in something

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
 message PayloadOrganizationCreated {
-  // created_time is when the organization was created.
-  google.protobuf.Timestamp created_time = 1;
+  // create_time is when the organization was created.
+  google.protobuf.Timestamp create_time = 1;
```
stefanvanburen, line 126: we use `create_time`, not `created_time`, elsewhere. fixed here and below.

I'm still not sure if I actually see the need for the separation between this and `event_time`, tbh, but if we want to keep both of these I should probably add them to more of the messages below. (`delete_time`, `added_time`, `role_change_time`, etc.)

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+}
+
+message PayloadPluginDeleted {}
+
+message PayloadPluginPushed {
+  // owner_id is the id of the owner of the plugin.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the plugin.
+  string owner_name = 2;
+}
```
stefanvanburen, line 221: unsure if this is needed?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -61,6 +62,25 @@ message Resource {
...
+  EVENT_TYPE_REPOSITORY_CREATED = 6;
+  EVENT_TYPE_REPOSITORY_DELETED = 7;
+  EVENT_TYPE_REPOSITORY_PUSHED = 8;
+  EVENT_TYPE_REPOSITORY_MEMBER_ADDED = 9;
+  EVENT_TYPE_REPOSITORY_MEMBER_ROLE_CHANGED = 10;
+  EVENT_TYPE_REPOSITORY_MEMBER_REMOVED = 11;
+  EVENT_TYPE_REPOSITORY_VISIBILITY_CHANGED = 12;
+  EVENT_TYPE_PLUGIN_CREATED = 13;
+  EVENT_TYPE_PLUGIN_DELETED = 14;
+  EVENT_TYPE_PLUGIN_PUSHED = 15;
```
unmultimedio, line 78: Let's wait [to confirm](https://www.notion.so/Audited-Events-List-db1e266e05ce4d0db48638ec6091b9b2?d=4c291e06bee24fa1b84ef82de04ce13e#f4b78a2eac69445f994807f7d4fa1dfc) if those events make sense for plugins.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+
+message PayloadOrganizationDeleted {}
+
+message PayloadOrganizationMemberAdded {
+  // member_id is the id of the member added to the organization.
+  string member_id = 1;
+  // member_name is the name of the member added to the organization.
+  string member_name = 2;
+  // new_role is the role granted to the member added to the organization.
+  buf.alpha.registry.v1alpha1.OrganizationRole new_role = 3;
```
unmultimedio, line 137:
```suggestion
  // member_role is the role granted to the member added to the organization.
  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+  buf.alpha.registry.v1alpha1.OrganizationRole new_role = 4;
+}
+
+message PayloadOrganizationMemberRemoved {
+  // member_id is the id of the member removed from the organization.
+  string member_id = 1;
+  // member_name is the name of the member removed from the organization.
+  string member_name = 2;
+  // old_role is the old role of the member removed from the organization.
+  buf.alpha.registry.v1alpha1.OrganizationRole old_role = 3;
```
unmultimedio, line 157:
```suggestion
  // member_role is the role that the member had when removed from the organization.
  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+}
+
+message PayloadRepositoryCreated {
+  // create_time is when the repository was created.
+  google.protobuf.Timestamp create_time = 1;
+}
```
unmultimedio, line 163: You can add here the `owner_name` and `owner_id`. It could be different from the actor, it could be an organization.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+
+message PayloadRepositoryDeleted {}
```
unmultimedio, line 165: Same here, `owner_name` and `owner_id`. Not sure if we have them in context though.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+  string draft_name = 5;
+}
+
+message PayloadRepositoryMemberAdded {
+  // member_id is the id of the member added to the repository.
+  string member_id = 1;
+  // member_name is the name of the member added to the repository.
+  string member_name = 2;
+  // new_role is the role granted to the member added to the repository.
+  buf.alpha.registry.v1alpha1.RepositoryRole new_role = 3;
```
unmultimedio, line 186:
```suggestion
  // member_role is the role granted to the member added to the repository.
  buf.alpha.registry.v1alpha1.RepositoryRole member_role = 3;
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+  buf.alpha.registry.v1alpha1.RepositoryRole new_role = 4;
+}
+
+message PayloadRepositoryMemberRemoved {
+  // member_id is the id of the member removed from the repository.
+  string member_id = 1;
+  // member_name is the name of the member removed from the repository.
+  string member_name = 2;
+  // old_role is the old role of the member removed from the repository.
+  buf.alpha.registry.v1alpha1.RepositoryRole old_role = 3;
```
unmultimedio, line 206:
```suggestion
  // member_role is the old role of the member removed from the repository.
  buf.alpha.registry.v1alpha1.RepositoryRole member_role = 3;
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+}
+
+message PayloadPluginCreated {
+  // create_time is when the plugin was created.
+  google.protobuf.Timestamp create_time = 1;
+}
+
+message PayloadPluginDeleted {}
```
unmultimedio, line 214: Pls add owner details in both.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+
+message PayloadRepositoryDeleted {}
+
+message PayloadRepositoryPushed {
+  // owner_id is the id of the owner of the repository.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the repository.
+  string owner_name = 2;
+  // commit is the name of the created commit.
+  string commit = 3;
```
unmultimedio, line 173: `commit_name` is already in the resource. You're missing repo name and repo id.

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: I think those fields are fine. What I'm unsure is the event itself, already asked @pkwarren [in here](https://www.notion.so/Audited-Events-List-db1e266e05ce4d0db48638ec6091b9b2?d=4c291e06bee24fa1b84ef82de04ce13e#f4b78a2eac69445f994807f7d4fa1dfc).

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+  google.protobuf.Timestamp create_time = 1;
+}
+
+message PayloadUserDeactivated {}
+
+message PayloadUserDeleted {}
+
+message PayloadUserLoggedIn {}
+
+message PayloadUserLoggedOut {}
```
unmultimedio, line 234: Not sure if we can have additional details here like an IP address? Not blocking, this could be figured out in that specific ticket.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,135 @@ message Event {
...
+  string member_name = 2;
+  // new_role is the role granted to the member added to the organization.
+  buf.alpha.registry.v1alpha1.OrganizationRole new_role = 3;
+}
+
+message PayloadOrganizationMemberRoleChanged {
+  // member_id is the id of the member whose role was changed.
+  string member_id = 1;
+  // member_name is the name of the member whose role was changed.
+  string member_name = 2;
```
unmultimedio, line 144: This is not needed, it's within the resource.

stefanvanburen: ah, I was just thinking about it backwards, with the resource being the organization instead of the member. I'll change those, thanks!

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: yeah, figured I'd leave it out of here and we could decide when implementing.

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: I think I was thinking that the repository is the thing that the action "happens to", so was going with the repository as the resource. But we can change that to be the commit?

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: want me to pull them out for now then?

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: yeah, I'm sure we can pull them out somehow!

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: Nah, let's leave them and fix when needed.

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: Yeah, that was proposed in [the list](https://www.notion.so/bufbuild/Audited-Events-List-db1e266e05ce4d0db48638ec6091b9b2), use the commit as the resource. Why do you think it makes more sense to have the repo as the resource?

stefanvanburen: ah, sorry, I didn't look closely enough at that. I wasn't viewing commits as a "resource" - I guess they are, though. I don't have a strong feeling about it; we're capturing essentially the same information either way. I'll change it.

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: Since both values are exactly the same for this event, it might be redundant to add it to the payload as well. For other events, it might be worth to add additional timestamps, like when updating something we'd might want to add the `create_time`? or when deleting something we'd like to add `create_time` and most recent `update_time`?

stefanvanburen: I'll remove it from the places where it's redundant.

For the other timestamps, I might leave them out, because we should have already recorded a timestamp in the audit log previously for the same ID for create / update events, so we should be able to figure them out from there. Obviously that wouldn't be included directly in the log, but we could at least go back and figure it out.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,145 @@ message Event {
...
+message PayloadOrganizationDeleted {}
+
+message PayloadOrganizationMemberAdded {
+  // member_id is the id of the member added to the organization.
+  string member_id = 1;
+  // member_name is the name of the member added to the organization.
+  string member_name = 2;
+  // member_role is the role granted to the member added to the organization.
+  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
+}
```
unmultimedio, line 135: For this event, the member is the resource.
```suggestion
message PayloadOrganizationMemberAdded {
  // organization_id is the id of the organization with the new member.
  string organization_id = 1;
  // organization_name is the name of the organization with the new member.
  string organization_name = 2;
  // member_role is the role granted to the member added to the organization.
  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
}
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,145 @@ message Event {
...
+
+message PayloadOrganizationMemberRemoved {
+  // member_id is the id of the member removed from the organization.
+  string member_id = 1;
+  // member_name is the name of the member removed from the organization.
+  string member_name = 2;
+  // member_role is the role that the member had when removed from the organization.
+  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
+}
```
unmultimedio, line 155: Same here.
```suggestion
message PayloadOrganizationMemberRemoved {
  // organization_id is the id of the organization that the member was removed from.
  string organization_id = 1;
  // organization_name is the name of the organization that the member was removed from.
  string organization_name = 2;
  // member_role is the role that the member had when removed from the organization.
  buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
}
```

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,145 @@ message Event {
...
 }
+
+message PayloadRepositoryDeleted {
+  // owner_id is the id of the owner of the repository.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the repository.
+  string owner_name = 2;
+}
+
+message PayloadRepositoryPushed {
```
unmultimedio, line 171: Not blocking, just an open thought: the resource for this event is the new commit. It has an ID, but not sure if we expose that in anyway. Commits are always dealt with the name (hash). Food for thought, should we include the ID? or have an empty ID? or have the same value as ID and name?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -80,11 +100,145 @@ message Event {
...
+  string member_id = 1;
+  // member_name is the name of the member whose role was changed.
+  string member_name = 2;
+  // old_role is the old role of the member whose role was changed.
+  buf.alpha.registry.v1alpha1.RepositoryRole old_role = 3;
+  // new_role is the new role of the member whose role was changed.
+  buf.alpha.registry.v1alpha1.RepositoryRole new_role = 4;
+}
+
+message PayloadRepositoryMemberRemoved {
```
unmultimedio, line 206: Same as the organization members. The resource is the member.

proto/buf/alpha/audit/v1alpha1/event.proto
stefanvanburen: I guess I'd expect ID to be the "public" ID and name to be the hash? The both are exposed in protos, I believe: https://github.com/bufbuild/buf/blob/781efb570e17049fa88920f73f05dd008e760460/proto/buf/alpha/registry/v1alpha1/repository_commit.proto#L23-L32

unmultimedio: Nice! yes let's do that

unmultimedio (approved):
Nice work 💪🏼

PR #1390, opened 2022-09-08
Add `LimitWriteBucket`

srikrsna-buf:
Adds a `LimitWriteBucket` that limits the total size of bytes that can be written to the bucket.


private/pkg/storage/limit.go
```diff
@@ -0,0 +1,79 @@
...
+	return newLimitedWriter(writeBucket, limit)
+}
+
+type limitedWriteBucket struct {
+	WriteBucket
+	currentSize *atomic.Int64
+	limit       int64
+}
+
+func newLimitedWriter(bucket WriteBucket, limit int64) *limitedWriteBucket {
```
bufdev, line 38: `newLimitWriteBucket

bufdev, line 38: I would probably just make this an `int`, sort of a standard we have with interface types, and then correct <0 to be 0, ie `if limit < 0 { limit = 0 }`

srikrsna-buf: Updated the public function to accept an `int`

private/pkg/storage/limit.go
```diff
@@ -0,0 +1,84 @@
...
+		WriteObjectCloser: writeObjectCloser,
+		bucketSize:        bucketSize,
+		limit:             limit,
+	}
+}
+
+func (o *limitedWriteObjectCloser) Write(p []byte) (n int, err error) {
+	newSize := o.bucketSize.Add(int64(len(p)))
+	if newSize > o.limit {
+		return 0, fmt.Errorf("limit writer: write limit reached: limit: %d, exceeded by: %d", o.limit, newSize-o.limit)
```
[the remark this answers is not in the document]
srikrsna-buf: Yeah, I was going to update it with a variant of the second approach you suggested.

private/pkg/storage/limit.go
```diff
@@ -0,0 +1,92 @@
...
+	newBucketSize := o.bucketSize.Add(writeSize)
+	if newBucketSize > o.limit {
+		o.bucketSize.Sub(writeSize)
+		return 0, &errWriteLimitReached{
+			Limit:       o.limit,
+			ExceedingBy: newBucketSize - o.limit,
+		}
+	}
+	writtenSize, err := o.WriteObjectCloser.Write(p)
+	if writtenSize < int(writeSize) {
```
twilly, line 88: Casting int -> int64 -> int is a bit confusing. If we're going to use int64 for counting over plain int, then it's better to always cast upward (int -> int64).

```go
if int64(writtenSize) < writeSize {
```

private/pkg/storage/limit.go
```diff
@@ -0,0 +1,92 @@
...
+	limit int64,
+) *limitedWriteObjectCloser {
+	return &limitedWriteObjectCloser{
+		WriteObjectCloser: writeObjectCloser,
+		bucketSize:        bucketSize,
+		limit:             limit,
+	}
+}
+
+func (o *limitedWriteObjectCloser) Write(p []byte) (n int, err error) {
```
twilly, line 77: No need for named return parameters if you don't use the names. :)

```suggestion
func (o *limitedWriteObjectCloser) Write(p []byte) (int, error) {
```

twilly (commented):
Looking pretty good. A few tests and this thing should be good to go.

doriable (approved):

PR #1395, opened 2022-09-12
BSR-527 - Adds `PayloadRepositoryVisibilityChanged`

saquibmian:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -212,6 +214,15 @@ message PayloadRepositoryMemberRemoved {
   buf.alpha.registry.v1alpha1.RepositoryRole member_role = 3;
 }
 
+message PayloadRepositoryVisibilityChanged {
+  // repository_id is the id of the repository that the member was removed from.
```
saquibmian, line 218: Ah this comment is wrong.

saquibmian: Fixed

njhale (approved):
lgtm!

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -212,6 +214,15 @@ message PayloadRepositoryMemberRemoved {
   buf.alpha.registry.v1alpha1.RepositoryRole member_role = 3;
 }
 
+message PayloadRepositoryVisibilityChanged {
+  // repository_id is the id of the repository that the visibility was changed on.
+  string repository_id = 1;
+  // repository_name is the name of the repository that the visibility was changed on.
+  string repository_name = 2;
```
unmultimedio, line 221: @saquibmian You don't need repository id+name here, because that's what the resource is. Contrary to the `RepositoryMember*`, in which it's necessary because the resource is the member, not the resource. Also, currently we only have `public | private` visibility, so may be a bit redundant, but I'd have `old_visibility` and `new_visibility` for completeness.

saquibmian: All of the other `PayloadRepository*` events have the `id+name` combo, so I kept it for consistency with them. But I can add `old_visibility` and `new_visibility`.

PR #1396, opened 2022-09-12
BSR-431: update http log format

oliversun09:


private/pkg/transport/http/httpserver/zap.go
```diff
@@ -90,3 +119,31 @@ func getRequestFields(request *http.Request) []zap.Field {
 		zap.String("path", request.RequestURI),
 	}
 }
+
+// this gets request fields that are relevant but not covered by http request log
```
njhale, line 123: It's convention in go for comments to start with a capitalized first letter.

Additionally, even though this isn't exported, I would still follow the godocs comment convention and start the comment with the function name:

```go
// getTopLevelRequestFields returns zap Fields that are relevant to a request, but not are not a part of httpRequestLog.
func getRequestFields(request *http.Request) []zap.Field {
    return []zap.Field{
		zap.String("host", request.Host),
		zap.String("path", request.RequestURI),
	}
}
```

private/pkg/transport/http/httpserver/zap.go
```diff
@@ -62,11 +89,13 @@ func newLogEntry(logger *zap.Logger, request *http.Request) *logEntry {
 }
 
 func (l *logEntry) Write(status int, size int, _ http.Header, duration time.Duration, _ interface{}) {
+	httpField := zap.Object(
```
njhale, line 92: Looks like we can simplify this a bit more:

```go
func (l *logEntry) Write(status, size int, _ http.Header, duration time.Duration, _ interface{}) {
	l.logger.Info(
	    "request", 
	     zap.Object(
	        "httpRequest",
		    newHTTPRequestLog(l.request, status, size, duration),
		 ),
		 zap.String("host", l.request.Host),
		 zap.String("path", l.request.RequestURI),
	),
}

```

What do you think @oliversun09?

private/pkg/transport/http/httpserver/zap.go
```diff
@@ -90,3 +119,31 @@ func getRequestFields(request *http.Request) []zap.Field {
...
+
+// this gets request fields that are relevant but not covered by http request log
+func getTopLevelRequestFields(request *http.Request) []zap.Field {
+	return []zap.Field{
+		zap.String("host", request.Host),
+		zap.String("path", request.RequestURI),
+	}
+}
+
+func newHTTPRequestLog(r *http.Request, status int, responseSize int, duration time.Duration) *httpRequestLog {
```
njhale, line 131: nit: I suspect the current form is the result of following examples in the existing code, but I think we should condense parameter declaration with like types; e.g.

```go
func newHTTPRequestLog(r *http.Request, status, size int, duration time.Duration) *httpRequestLog {
// ...
}
```

njhale (commented):
Thanks for picking this up!

Just a few small comments.

private/pkg/transport/http/httpserver/zap.go
oliversun09: I removed this function because it's no longer needed after I simplified `Write`, as per your comment

private/pkg/transport/http/httpserver/zap.go
oliversun09: Fixed : )

private/pkg/transport/http/httpserver/zap.go
```diff
@@ -90,3 +119,23 @@ func getRequestFields(request *http.Request) []zap.Field {
 		zap.String("path", request.RequestURI),
 	}
 }
+
+func newHTTPRequestLog(r *http.Request, status, responseSize int, duration time.Duration) *httpRequestLog {
+	return &httpRequestLog{
+		requestMethod: r.Method,
+		requestURL:    getFullURL(r),
+		status:        status,
+		responseSize:  fmt.Sprintf("%v", responseSize),
```
joshcarp, line 128: generally `strconv.Itoa` is a lot better (and faster) than `fmt.Sprintf`,

joshcarp, line 128:
```suggestion
		responseSize:  strconv.itoa(responseSize),
```

private/pkg/transport/http/httpserver/zap.go
```diff
@@ -15,13 +15,40 @@
...
+// includes fields described in https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest
+type httpRequestLog struct {
+	requestMethod string
+	requestURL    string
+	status        int
+	responseSize  string
+	userAgent     string
+	remoteIP      string
+	latency       string
+	protocol      string
```
joshcarp, line 36: is there any reason why everything here are strings, 
can't they be their native types (responseSize as int etc) and then zap just logs it as its native type?

oliversun09: Actually I was following the format of [google cloud's log entry](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest). Also, I don't think we're going to do this update in the buf repo, and I should close this PR, see issue BSR-431. Thanks for the review though.

PR #1397, opened 2022-09-12
add httpserver.Run function

rhbuf:
The Run function only concerns itself with running a http server and does not attempt to perform any routing setup. The http.Handler that is passed is expected to be a chi.Mux or other complex router that will take care of routing for the server.

I've copied the RunnerOptions that matter for the server setup into RunOptions and split out server logic from the Run function into a `serve` function that can be called privately by the new Run function.

### DIFF
--- a/private/pkg/transport/http/httpserver/httpserver.go
+++ b/private/pkg/transport/http/httpserver/httpserver.go
@@ -174,3 +174,51 @@ func RunnerWithSilentEndpoints(silentEndpoints ...string) RunnerOption {
 		}
 	}
 }
+
+// RunOption is an option for a new Run.
+type RunOption = RunnerOption
+
+// RunWithShutdownTimeout returns a new RunOption that uses the given shutdown timeout.
+//
+// The default is to use DefaultShutdownTimeout.
+// If shutdownTimeout is 0, no graceful shutdown will be performed.
+func RunWithShutdownTimeout(shutdownTimeout time.Duration) RunOption {
+	return func(runner *runner) {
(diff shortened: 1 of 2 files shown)


private/pkg/transport/http/httpserver/httpserver.go
```diff
@@ -174,3 +174,51 @@ func RunnerWithSilentEndpoints(silentEndpoints ...string) RunnerOption {
...
+		runner.tlsConfig = tlsConfig
+	}
+}
+
+// Run will start a HTTP server listening on the provided listener and
+// serving the provided handler. This call is blocking and the run
+// is cancelled when the input context is cancelled, the listener is
+// closed upon return.
+//
+// The Run can be configured further by passing a variety of options.
```
mfridman, line 215: "The Run".

Maybe "The Run function can be.." ?

private/pkg/transport/http/httpserver/runner.go
```diff
@@ -96,7 +96,17 @@ func (s *runner) Run(
...
+	return s.serve(ctx, listener, mux)
+}
 
+// serve is intended to only be called within the httpserver package by
+// the Run function. Once callers have migrated away from NewRunner().Run()
+// it would be ideal to move this logic into the Run function.
+func (s *runner) serve(
+	ctx context.Context,
+	listener net.Listener,
+	mux http.Handler,
```
mfridman, line 108: Maybe pick one for consistency to describe an http.Handler? It's called `handler http.Handler` in the Run function

mfridman (approved):
Nits, looks sane to me.

PR #1398, opened 2022-09-12
BSR-527 - Adds old visibility to repository visiblity changed event.

saquibmian:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -219,8 +219,10 @@ message PayloadRepositoryVisibilityChanged {
   string repository_id = 1;
   // repository_name is the name of the repository that the visibility was changed on.
   string repository_name = 2;
```
unmultimedio, line 221: Also you can remove these 2. These will be in the resource field.

unmultimedio (changes requested):

proto/buf/alpha/audit/v1alpha1/event.proto
saquibmian: All of the other `PayloadRepository*` events have the `id+name` combo, so I kept it for consistency with them. You sure you want me to remove them?

unmultimedio: Not all of them, `Created` and `Deleted` have the owner's id+name. `Push` has both because its resource it's the commit, not the repo. `Member*` have just repo id+name (it's missing the owner's data). This event is like `Created` or `Deleted`. We should replace them with owner's info, and include missing owner's info to the `Member*` events :)

unmultimedio (approved):
Nice! Thank you for filling up the missing ones.

PR #1399, opened 2022-09-12
fix local generation using plugin key

pkwarren:
Update the local generation to look up the plugin name from the PluginName() accessor (which looks at the Plugin, Name, and Remote in order). This fixes local generation when using the new config key.

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -251,7 +251,7 @@ func (g *generator) execLocalPlugin(
 	response, err := appprotoexecGenerator.Generate(
 		ctx,
 		container,
-		pluginConfig.Name,
+		pluginConfig.PluginName(),
 		bufimage.ImagesToCodeGeneratorRequests(
 			pluginImages,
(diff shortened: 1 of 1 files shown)


private/buf/bufgen/generator.go
```diff
@@ -251,7 +251,7 @@ func (g *generator) execLocalPlugin(
 	response, err := appprotoexecGenerator.Generate(
 		ctx,
 		container,
-		pluginConfig.Name,
+		pluginConfig.PluginName(),
```
pkwarren, line 254: Before:
```
$ buf generate
Failure: plugin go-grpc: could not find protoc plugin for name
```

After:
```
$ buf generate
$ tree gen
gen
└── proto
    └── go
        ├── payment
        │   └── v1alpha1
        │       └── payment.pb.go
        └── pet
            └── v1
                ├── pet.pb.go
                └── pet_grpc.pb.go

6 directories, 3 files
```

mfridman (approved):

PR #1400, opened 2022-09-13
BSR-525 - Adds `RepositoryRole.None`

saquibmian:


proto/buf/alpha/registry/v1alpha1/role.proto
```diff
@@ -40,6 +40,7 @@ enum RepositoryRole {
   REPOSITORY_ROLE_WRITE = 3;
   REPOSITORY_ROLE_READ = 4;
   REPOSITORY_ROLE_LIMITED_WRITE = 5;
+  REPOSITORY_ROLE_NONE = 6;
```
njhale, line 43: is `REPOSITORY_ROLE_UNSPECIFIED` semantically equivalent to `REPOSITORY_ROLE_NONE`, or is `_UNSPECIFIED` just a catchall for cases where we don't want/need to be explicit in the proto message?

unmultimedio: I'm not sure what case is this trying to represent? I don't think the BSR allows for repository members without a role.

saquibmian:
Closing this PR for now, let me try out something else.

PR #1403, opened 2022-09-14
Include `LICENSE` file in the module on `buf push`

cyinma:
fixes BSR-575

This is to also push the `LICENSE` file that relative to the root of the module on  `buf push`, so that we can determine the license of a BSR repository based on the license in its latest main commit.

For license file name that we accept, we might want to also accept other names (e.g. `LICENSE.md`, `LICENSE.txt`), but I think we can just start with just the `LICENSE` for now


private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -35,6 +35,8 @@ import (
 const (
 	// DocumentationFilePath defines the path to the documentation file, relative to the root of the module.
 	DocumentationFilePath = "buf.md"
+	// LicenseFilePath defines the path to the license file, relative to the root of the module.
+	LicenseFilePath = "LICENSE"
```
bufdev, line 39: There's two issues with this:

- We didn't do this before because licenses are varied across many repos - some use `LICENSE`, others use `LICENSE.md`, `LICENSE.txt`, etc - it kind of needs to be variable to apply to the various choices users legitimately make here. You could get around this with having a "priority list", and having the module store both the license file name and the content, but not great.
- This assumes the license will be in the root of the module, which will not often be the case - this is never the case for us, for example. The license is typically at the root of the source repository, not the directory that is the root of the module. We can't expect users to copy their LICENSE into the root of the module.

timostamm: It's somewhat similar issue with buf.md? At least one user is [linking their README.md](https://github.com/authzed/api) to their buf.md, which seems like a viable option.

robbertvanginkel: The original proposal contained an idea to let users set license URLs through the BSR interface/api. While we need the licenses to be generated into registry artifacts for for example go modules, having some url that we need to resolve at generate time is problematic from a dependency/latency perspective as well as that the contents can change over time.

To have some license information available stored on the BSR for each commit, I think our options are:
1. Pick and document a license filename (this) 
1. Add some `license: ../License.txt` key to `buf.yaml`, read it on push and put the contents verbatim in the pushed `Module` (like what happens with the buf.md today)
1. add `license_spdix_id` to `buf.yaml` and include the identifier it on `buf push`. We choose to not support custom licenses.

Seems like we don't want to do 1, 2 seems a bit odd since we don't have any other keys that refer to files or are allowed to reach into parent directories. 3 seems straightforward but also restrictive. Open to alternatives.

bufdev: Talked offline, agreed on existing strategy (1 on robberts list)

private/bufpkg/bufmodule/util.go
```diff
@@ -72,6 +72,20 @@ func getDocumentationForBucket(
 	return string(documentationData), nil
 }
 
+func getLicenseForBucket(
```
twilly, line 75: This isn't materially different from `getDocumentationForBucket`.  Maybe it's time to put this logic as `getFileForBucket` and pass the doc and license file paths to it in `get(Documentation|License)ForBucket`.

twilly (approved):
LGTM. :)

doriable (approved):

PR #1408, opened 2022-09-15
storagearchive: Add max file size option for `Untar`

srikrsna-buf:
Adds a max file size option for `storagearchive.Untar`


private/pkg/storage/storagearchive/storagearchive.go
```diff
@@ -91,25 +103,51 @@ func Untar(
...
 		}
-		if !ok {
-			continue
-		}
-		if tarHeader.FileInfo().Mode().IsRegular() {
-			if tarHeader.Size < 0 {
-				return fmt.Errorf("invalid size for tar file %s: %d", tarHeader.Name, tarHeader.Size)
+		if !ok || !tarHeader.FileInfo().Mode().IsRegular() {
+			// Next also discards the file but doesn't apply the limit.
+			// Explicitly discard the file with limit applied.
```
twilly, line 119: We can apply a file limit without using `io.LimitReader`.

```go
if tarHeader.Size > options.maxFileSize {
    return ErrFileSizeLimit
}
```

No special casing requires if the default `options.maxFileSize` value is `math.MaxInt64`.

doriable (approved):

twilly (commented):
Post-land LGTM. :)

PR #1418, opened 2022-09-19
add rpc for SearchTagService

oliversun09:


proto/buf/alpha/registry/v1alpha1/search_tag.proto
```diff
@@ -0,0 +1,43 @@
...
+message SearchTagServiceSearchResponse {
+  repeated RepositoryCommit repository_commits = 1;
+  // There are no more pages if this is 0.
+  uint32 next_page_token = 2;
+}
+
+// SearchTagService is the search service that searches commits by tags
+service SearchTagService {
+  // Search searches in a repository
+  rpc Search(SearchTagServiceSearchRequest) returns (SearchTagServiceSearchResponse);
```
mfridman, line 42: We already have an existing `SearchService`, any reason to create another service specific for only tags?

https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/search.proto#L103-L107

Could be easier and cleaner to add `SearchTags` method to the existing one, unless this was a FE requirement?

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -104,6 +105,8 @@ message SearchResult {
 service SearchService {
   // Search searches the BSR.
   rpc Search(SearchRequest) returns (SearchResponse);
+  // SearchTag searches in a repository
+  rpc SearchTag(SearchTagRequest) returns (SearchTagResponse);
```
mfridman, line 109: Is this intended to search by reference, or will it, as the name suggested only search by commit tag?

A [reference](https://docs.buf.build/bsr/overview#referencing-a-module) can be a commit, tag and draft name (coming soon).

mfridman (approved):
once the [comment](https://github.com/bufbuild/buf/pull/1418#discussion_r975436693) is addressed feel free to merge.

Either `SearchCommitReference` or `SearchReference` sound good to me.

PR #1419, opened 2022-09-20
bufmodulebuild: include the bucket's config

twilly:
`bufmodulebuild.BuildForBucket` calls `bufmodule.NewModuleForBucket`
with a different bucket than the one passed to it. This mapped bucket
didn't include `buf.yaml`, hence the built `bufmodule.Module` didn't
have values for breaking or linting config.

`buf push` will now pass good lint and breaking config to the registry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Push lint and breaking configuration to the registry.
 
 ## [v1.8.0] - 2022-09-14
 
--- a/private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
+++ b/private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
@@ -17,6 +17,7 @@ package bufmodulebuild
 import (
 	"context"
 
+	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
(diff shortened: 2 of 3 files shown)


srikrsna-buf (approved):

doriable (approved):
Do we need to backfill anything for pushes that may have empty configs?

srikrsna-buf:
> Do we need to backfill anything for pushes that may have empty configs?

Won't that change the commit hash?

twilly:
> Do we need to backfill anything for pushes that may have empty configs?

We never received lint or breaking configuration, so there's nothing we can backfill on the BSR side of things. It does, however...

> Won't that change the commit hash?

Change the commit hash. A new commit will be allowed when a user runs `buf push`, which may be surprising to them. I should mention this in the changelog.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Push lint and breaking configuration to the registry.
```
doriable, line 5: We may want to position this as a bug fix -- "Fixed bug to resume pushing lint and breaking configuration to the registry" or something along those lines.

doriable:
Ah yeah, that would be the case... which is a bit of a problem because I believe the commits without lint/breaking configs would not be usable since they do not pass the validation o_o

PR #1423, opened 2022-09-20
Upgrade golangci-lint

robbertvanginkel:
See the release notes at https://github.com/golangci/golangci-lint/releases/tag/v1.49.0 for the changes in linters.

I did not add `interfacebloat`, a strongly opinionated linter that fails when there's more than 10 methods on an interface. `contextcheck` was re-enabled in this version of golangci-lint and we should fix the issues it brings up, temporarily disabled it for now.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -35,8 +35,7 @@ linters:
     - bidichk
     - bodyclose
     - containedctx
-    - contextcheck
-    - deadcode
+    # - contextcheck
     - decorder
     - depguard
     - dogsled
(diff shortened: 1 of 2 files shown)


pkwarren (approved):

.golangci.yml
```diff
@@ -35,8 +35,7 @@ linters:
     - bidichk
     - bodyclose
     - containedctx
-    - contextcheck
-    - deadcode
+    # - contextcheck
```
robbertvanginkel, line 38: https://linear.app/bufbuild/issue/TCN-492/fix-golangci-lint-context-check-issues-in-bufbuildbuf

PR #1427, opened 2022-09-21
Use latest protocompile in `buf format`, fix things up in the formatter

jhump:
There were a few bugs in the formatter that were uncovered with the new version of `protocompile`, mostly around trailing vs. leading comments.

For example, there was previously a bug in protocompile (https://github.com/bufbuild/protocompile/pull/33) where certain elements (like extension names enclosed in parentheses in an option name, full options declaration enclosed in brackets) would not get leading comments. The comments that _should_ be interpreted as leading would usually instead be interpreted as a trailing comment on the previous token.

That means that there were some cases in the formatter where it was not correctly handling leading comments on certain elements (such as custom option names).

Also, this was treating leading/trailing comments for inline comments in between tokens as if the parser's classification of trailing vs. leading were authoritative. If it was a trailing comment in the AST, it would format it with _no whitespace_ after the token, before the comment. Similarly, for leading comments, it would format it with nothing after the comment, before the token. But this classification of trailing vs. leading is just heuristic (and mostly done to mirror this distinction in source code info in descriptors generated by `protoc`). So now, the formatter will preserve whitespace: if a comment had no whitespace between it and a token, that will be preserved. If it did have whitespace, it will be canonicalized to a single space.

Getting everything looking right involved a greater degree of changes to the formatter than I would have liked. 🤷 

The good news is that these changes should mostly be idempotent for users: they shouldn't see a lot of new diff churn from formatting with this new version. As you'll see in the new golden files, the only new differences are the whitespace around inline comments. And this change tries to _preserve_ existing whitespace there, so if a file was previously formatted with the old version and then re-formatted with the new, there shouldn't be a diff.

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -83,9 +83,7 @@ func (f *formatter) P(elements ...string) {
 
 // Space adds a space to the generated output.
 func (f *formatter) Space() {
-	if !strings.ContainsRune("\x00 \t\n", f.lastWritten) {
-		f.pendingSpace = true
-	}
+	f.pendingSpace = true
 }
 
 // In increases the current level of indentation.
@@ -117,10 +115,16 @@ func (f *formatter) WriteString(elem string) {
 	if f.pendingSpace {
 		f.pendingSpace = false
 		first, _ := utf8.DecodeRuneInString(elem)
-		// We don't want dangling spaces before a newline, a
-		// comma, or a semicolon. So only print the space if the
-		// next character to write is not any of those.
-		if first != '\n' && first != ';' && first != ',' {
+
+		// We don't want "dangling spaces" before certain characters:
+		// newlines, commas, semicolons, and close parens/braces.
+		// Similarly, we don't want extra/doubled spaces or dangling
+		// spaces after certain characters: open parens/braces. So
+		// only print the space if the previous and next character
+		// don't match above conditions.
+
+		if !strings.ContainsRune("\x00 \t\n<[{(", f.lastWritten) &&
+			!strings.ContainsRune("\n;,)]}>", first) {
 			if _, err := f.writer.Write([]byte{' '}); err != nil {
 				f.err = multierr.Append(f.err, err)
 				return
@@ -538,7 +542,6 @@ func (f *formatter) writeMessageLiteral(messageLiteralNode *ast.MessageLiteralNo
 // an element in an array literal.
 func (f *formatter) writeMessageLiteralForArray(
 	messageLiteralNode *ast.MessageLiteralNode,
-	lastElement bool,
 ) {
 	var elementWriterFunc func()
 	if len(messageLiteralNode.Elements) > 0 {
@@ -1183,7 +1186,7 @@ func (f *formatter) writeCompositeValueForArrayLiteral(
 	case *ast.SignedFloatLiteralNode:
 		f.writeSignedFloatLiteralForArray(node, lastElement)
 	case *ast.MessageLiteralNode:
-		f.writeMessageLiteralForArray(node, lastElement)
+		f.writeMessageLiteralForArray(node)
 	default:
 		f.err = multierr.Append(f.err, fmt.Errorf("unexpected array value node %T", node))
 	}
@@ -1284,9 +1287,6 @@ func (f *formatter) writeBody(
 		}
 	}
 	f.Out()
-	if info := f.fileNode.NodeInfo(closeBrace); info.LeadingComments().Len() > 0 {
-		f.P()
-	}
 	closeBraceWriterFunc(closeBrace)
 }
 
@@ -1785,8 +1785,7 @@ func (f *formatter) writeInline(node ast.Node) {
 //
 //	message Foo {
 //	  string bar = 1;
-//
-//	// Leading comment on '}'.
+//	  // Leading comment on '}'.
 //	} // Trailing comment on '}.
 func (f *formatter) writeBodyEnd(node ast.Node) {
 	if _, ok := node.(ast.CompositeNode); ok {
@@ -1799,7 +1798,10 @@ func (f *formatter) writeBodyEnd(node ast.Node) {
 	defer f.SetPreviousNode(node)
 	info := f.fileNode.NodeInfo(node)
 	if info.LeadingComments().Len() > 0 {
+		f.P()
+		f.In()
 		f.writeMultilineComments(info.LeadingComments())
+		f.Out()
 	}
 	f.Indent()
 	f.writeNode(node)
@@ -1845,7 +1847,13 @@ func (f *formatter) writeBodyEndInline(node ast.Node) {
 	defer f.SetPreviousNode(node)
 	info := f.fileNode.NodeInfo(node)
 	if info.LeadingComments().Len() > 0 {
+		if f.previousHasTrailingComments() {
+			// line between prior node's trailing comments and upcoming leading comments
+			f.P()
+		}
+		f.In()
 		f.writeMultilineComments(info.LeadingComments())
+		f.Out()
 	}
 	f.Indent()
 	f.writeNode(node)
@@ -2039,6 +2047,13 @@ func (f *formatter) previousTrailingCommentsWroteNewline() bool {
 		return newlineCount(previousTrailingComments.Index(0).LeadingWhitespace()) > 0
 	}
 	return false
+}
+
+// previousHasTrailingComments returns true if the previous node included
+// trailing comments
+func (f *formatter) previousHasTrailingComments() bool {
+	previousTrailingComments := f.fileNode.NodeInfo(f.previousNode).TrailingComments()
+	return previousTrailingComments.Len() > 0
 }
 
 // stringForOptionName returns the string representation of the given option name node.
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -323,7 +323,7 @@ func (f *formatter) writePackage(packageNode *ast.PackageNode) {
 func (f *formatter) writeImport(importNode *ast.ImportNode) {
 	// We don't use f.writeStart here because the imports are sorted
 	// and potentially changed order.
-	f.writeBodyEndInline(importNode.Keyword)
+	f.writeElementInlineTrailing(importNode.Keyword, false)
 	f.Space()
 	// We don't want to write the "public" and "weak" nodes
 	// if they aren't defined. One could be set, but never both.
@@ -345,7 +345,7 @@ func (f *formatter) writeImport(importNode *ast.ImportNode) {
 func (f *formatter) writeFileOption(optionNode *ast.OptionNode) {
 	// We don't use f.writeStart here because the options are sorted
 	// and potentially changed order.
-	f.writeBodyEndInline(optionNode.Keyword)
+	f.writeElementInlineTrailing(optionNode.Keyword, false)
 	f.Space()
 	f.writeNode(optionNode.Name)
 	f.Space()
@@ -569,7 +569,7 @@ func (f *formatter) maybeWriteCompactMessageLiteral(
 	inArrayLiteral bool,
 ) bool {
 	if len(messageLiteralNode.Elements) == 0 || len(messageLiteralNode.Elements) > 1 ||
-		f.hasInteriorComments(messageLiteralNode) ||
+		f.hasInteriorComments(messageLiteralNode.Children()...) ||
 		messageLiteralHasNestedMessageOrArray(messageLiteralNode) {
 		return false
 	}
@@ -1101,7 +1101,7 @@ func (f *formatter) writeCompactOptions(compactOptionsNode *ast.CompactOptionsNo
 		f.inCompactOptions = false
 	}()
 	if len(compactOptionsNode.Options) == 1 &&
-		!f.hasInteriorComments(compactOptionsNode) {
+		!f.hasInteriorComments(compactOptionsNode.OpenBracket, compactOptionsNode.Options[0].Name) {
 		// If there's only a single compact scalar option without comments, we can write it
 		// in-line. For example:
 		//
@@ -1157,20 +1157,15 @@ func (f *formatter) writeCompactOptions(compactOptionsNode *ast.CompactOptionsNo
 	)
 }
 
-func (f *formatter) hasInteriorComments(n ast.Node) bool {
-	cn, ok := n.(ast.CompositeNode)
-	if !ok {
-		return false
-	}
-	children := cn.Children()
-	for i, child := range children {
+func (f *formatter) hasInteriorComments(nodes ...ast.Node) bool {
+	for i, n := range nodes {
 		// interior comments mean we ignore leading comments on first
 		// token and trailing comments on the last one
-		info := f.fileNode.NodeInfo(child)
+		info := f.fileNode.NodeInfo(n)
 		if i > 0 && info.LeadingComments().Len() > 0 {
 			return true
 		}
-		if i < len(children)-1 && info.TrailingComments().Len() > 0 {
+		if i < len(nodes)-1 && info.TrailingComments().Len() > 0 {
 			return true
 		}
 	}
@@ -1187,7 +1182,7 @@ func (f *formatter) hasInteriorComments(n ast.Node) bool {
 //	]
 func (f *formatter) writeArrayLiteral(arrayLiteralNode *ast.ArrayLiteralNode) {
 	if len(arrayLiteralNode.Elements) == 1 &&
-		!f.hasInteriorComments(arrayLiteralNode) &&
+		!f.hasInteriorComments(arrayLiteralNode.Children()...) &&
 		!arrayLiteralHasNestedMessageOrArray(arrayLiteralNode) {
 		// arrays with a single scalar value and no comments can be
 		// printed all on one line
@@ -1215,7 +1210,7 @@ func (f *formatter) writeArrayLiteral(arrayLiteralNode *ast.ArrayLiteralNode) {
 				}
 				if lastElement {
 					// The last element won't have a trailing comma.
-					f.writeBodyEnd(arrayLiteralNode.Elements[i])
+					f.writeElementEndLineTrailing(arrayLiteralNode.Elements[i], false)
 					return
 				}
 				f.writeStart(arrayLiteralNode.Elements[i])
@@ -1461,7 +1456,7 @@ func (f *formatter) writeCompoundStringLiteral(compoundStringLiteralNode *ast.Co
 			// wrote a newline, we need to add one here.
 			f.P()
 		}
-		f.writeBodyEnd(child)
+		f.writeElementEndLineTrailing(child, false)
 	}
 	f.Out()
 }
@@ -1487,10 +1482,10 @@ func (f *formatter) writeCompoundStringLiteralForSingleOption(compoundStringLite
 			f.P()
 		}
 		if i == len(compoundStringLiteralNode.Children())-1 {
-			f.writeBodyEndInline(child)
+			f.writeElementInlineTrailing(child, false)
 			break
 		}
-		f.writeBodyEnd(child)
+		f.writeElementEndLineTrailing(child, false)
 	}
 	f.Out()
 }
@@ -1513,10 +1508,10 @@ func (f *formatter) writeCompoundStringLiteralForArray(
 			}
 		}
 		if !lastElement && i == len(compoundStringLiteralNode.Children())-1 {
-			f.writeBodyEndInline(child)
+			f.writeElementInlineTrailing(child, false)
 			return
 		}
-		f.writeBodyEnd(child)
+		f.writeElementEndLineTrailing(child, false)
 	}
 }
 
@@ -1866,6 +1861,10 @@ func (f *formatter) writeInline(node ast.Node) {
 //	  // Leading comment on '}'.
 //	} // Trailing comment on '}.
 func (f *formatter) writeBodyEnd(node ast.Node) {
+	f.writeElementEndLineTrailing(node, true)
+}
+
+func (f *formatter) writeElementEndLineTrailing(node ast.Node, bodyEnd bool) {
 	if _, ok := node.(ast.CompositeNode); ok {
 		// We only want to write comments for terminal nodes.
 		// Otherwise comments accessible from CompositeNodes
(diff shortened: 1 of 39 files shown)


private/buf/bufformat/formatter.go
```diff
@@ -104,7 +114,22 @@ func (f *formatter) Indent() {
 
 // WriteString writes the given element to the generated output.
 func (f *formatter) WriteString(elem string) {
-	if _, err := fmt.Fprint(f.writer, elem); err != nil {
+	if f.pendingSpace {
```
jhump, line 117: I added code to try to preserve some of the whitespace conventions in inline comments. That means it's possible for it to now call `f.Space()` but then the next call wants to write a newline. And trailing whitespace looks ugly.

So we now defer actually writing the space and skip it if the next character to write is a newline.

private/buf/bufformat/formatter.go
```diff
@@ -116,11 +141,11 @@ func (f *formatter) SetPreviousNode(node ast.Node) {
 }
 
 // writeFile writes the file node.
-func (f *formatter) writeFile(fileNode *ast.FileNode) {
-	f.writeFileHeader(fileNode)
-	f.writeFileTypes(fileNode)
-	if fileNode.EOF != nil {
-		info := f.fileNode.NodeInfo(fileNode.EOF)
+func (f *formatter) writeFile() {
```
jhump, line 144: I removed the argument because it wasn't necessary since the file node is a field on `f`. In fact, `writeFileHeader` and `writeFileTypes` were also passed this argument but never even used it. So this is just minor clean up.

private/buf/bufformat/formatter.go
```diff
@@ -405,10 +430,10 @@ func (f *formatter) writeOptionPrefix(optionNode *ast.OptionNode) {
 //	(custom.thing).bridge.(another.thing)
 func (f *formatter) writeOptionName(optionNameNode *ast.OptionNameNode) {
 	for i := 0; i < len(optionNameNode.Parts); i++ {
-		if i == 0 {
-			// The comments will have already been written as a multiline
-			// comment above the option name, so we need to handle this case
-			// specially.
+		if f.inCompactOptions && i == 0 {
```
jhump, line 433: Previously, the option name rarely had a leading comment, which is why this bug was never noticed before.

We've only written all of the leading comments for compact options. Now that the new protocompile AST is likely to contain leading comments here, this bug was noticed (by seeing the formatter completely _drop_ a comment for a non-compact option declaration... oops!).

private/buf/bufformat/formatter.go
```diff
@@ -418,14 +443,19 @@ func (f *formatter) writeOptionName(optionNameNode *ast.OptionNameNode) {
 				f.writeInline(fieldReferenceNode.Name)
 			} else {
 				f.writeNode(fieldReferenceNode.Name)
+				if info := f.fileNode.NodeInfo(fieldReferenceNode.Name); info.TrailingComments().Len() > 0 {
```
jhump, line 446: Oops. This was also a place that was possibly dropping comments.

private/buf/bufformat/formatter.go
```diff
@@ -1009,6 +1039,10 @@ func (f *formatter) writeRange(rangeNode *ast.RangeNode) {
 //	  json_name = "something"
 //	]
 func (f *formatter) writeCompactOptions(compactOptionsNode *ast.CompactOptionsNode) {
+	f.inCompactOptions = true
+	defer func() {
+		f.inCompactOptions = false
+	}()
```
jhump, line 1045: This is kind of gross, to stash this state on `f` instead of on the stack. But weaving this flag all the way through numerous function calls, to get it into `writeOptionName` was even messier. So I went with the less smelly approach.

private/buf/bufformat/formatter.go
```diff
@@ -1226,6 +1260,7 @@ func (f *formatter) writeBody(
 		f.Indent()
 		f.writeNode(closeBrace)
 		f.writeTrailingEndComments(info.TrailingComments())
+		f.SetPreviousNode(closeBrace)
```
jhump, line 1263: Another bug. This code path was unexercised with the old version protocompile -- again because the comment would usually have been attributed as a trailing comment to previous token, so we'd not fall into this condition.

But with the new protocompile, it popped up as a mysterious missing newline in between elements (all because some logic that decides to add a newline wants to know if the last thing was a closing brace). Other call paths through this function do set this correctly after writing it.

private/buf/bufformat/formatter.go
```diff
@@ -1261,7 +1296,9 @@ func (f *formatter) writeOpenBracePrefix(openBrace ast.Node) {
 	info := f.fileNode.NodeInfo(openBrace)
 	if info.LeadingComments().Len() > 0 {
 		f.writeInlineComments(info.LeadingComments())
-		f.Space()
+		if info.LeadingWhitespace() != "" {
```
jhump, line 1299: This is part of the logic that tries to preserve whitespace conventions. You'll see this condition sprinkled in a few places.

private/buf/bufformat/formatter.go
```diff
@@ -1329,11 +1366,6 @@ func (f *formatter) writeFieldLabel(fieldLabel ast.FieldLabel) {
 	f.WriteString(fieldLabel.Val)
 }
 
-// writeBoolLiteral writes a bool literal (e.g. 'true').
-func (f *formatter) writeBoolLiteral(boolLiteralNode *ast.BoolLiteralNode) {
-	f.WriteString(strconv.FormatBool(boolLiteralNode.Val))
-}
```
jhump, line 1335: No such thing as a bool literal in the new version of protocompile. We don't actually know if `true` or `false` is a boolean literal until we do type checking, when interpreting options. It could also be the name of an enum value. So now the AST will just use identifier nodes whose text might be `"true"` or `"false"`.

private/buf/bufformat/formatter.go
```diff
@@ -1910,7 +1946,7 @@ func (f *formatter) writeMultilineComments(comments ast.Comments) {
 //	}
 func (f *formatter) writeInlineComments(comments ast.Comments) {
 	for i := 0; i < comments.Len(); i++ {
-		if i > 0 {
+		if i > 0 || comments.Index(i).LeadingWhitespace() != "" || f.lastWritten == ';' || f.lastWritten == '}' {
```
jhump, line 1949: It seemed more pleasing to the eye to always separate the comment with a space if it follows the end of a declaration (a semicolon or close brace).

private/buf/bufformat/formatter_test.go
```diff
@@ -49,11 +48,7 @@ func testFormatProto2(t *testing.T) {
 	testFormatNoDiff(t, "testdata/proto2/message/v1")
 	testFormatNoDiff(t, "testdata/proto2/option/v1")
 	testFormatNoDiff(t, "testdata/proto2/quotes/v1")
-
-	// TODO: Temporarily skipping this test since it's
-	// due to a bug in protocompile.
-	//
-	// testFormatNoDiff(t, "testdata/proto2/utf8/v1")
```
jhump, line 56: Bug is fixed! https://github.com/bufbuild/protocompile/pull/34

private/buf/bufformat/formatter_test.go
```diff
@@ -67,48 +62,45 @@ func testFormatProto3(t *testing.T) {
...
-			ctx,
-			readBucket,
-			"",
-			func(formattedFile storage.ReadObject) error {
-				originalPath := formattedFile.Path()
-				if strings.HasPrefix(filepath.Base(originalPath), "option_name") {
-					// TODO: Temporarily skipping this test since it's
-					// due to a bug in protocompile.
-					//
-					// https://github.com/jhump/protocompile/pull/3
```
jhump, line 90: Bug is fixed! https://github.com/bufbuild/protocompile/pull/33

private/buf/bufformat/formatter_test.go
```diff
@@ -67,48 +62,45 @@ func testFormatProto3(t *testing.T) {
...
+	t.Run(path, func(t *testing.T) {
```
jhump, line 65: Putting each file in its own test case provides much nicer output IMO. It also reduces the abruptness of a `require` assertion fail -- the test case that is immediately failed is just a single file instead of the entire function and all files.

private/buf/bufformat/formatter.go
```diff
@@ -75,7 +83,9 @@ func (f *formatter) P(elements ...string) {
 
 // Space adds a space to the generated output.
 func (f *formatter) Space() {
-	f.WriteString(" ")
+	if !strings.ContainsRune("\x00 \t\n", f.lastWritten) {
```
jhump, line 86: This prevents doubled whitespace from accidentally popping up, now that there are more calls to `f.Space()`. In particular, we explicitly put a space between some tokens, but then we might also try to add a space when printing inline comments, and two spaces looks bad.

private/buf/bufformat/testdata/proto2/enum/v1/enum.golden.proto
```diff
@@ -5,12 +5,12 @@ enum Type {
   // Another trailing comment on Type.
   // A third trailing comment on Type.
 
-  TYPE_UNSPECIFIED/* Comment before '=' */ = /* Comment after '=' */0 [deprecated = true];
+  TYPE_UNSPECIFIED /* Comment before '=' */ = /* Comment after '=' */ 0 [deprecated = true];
```
jhump, line 8: As can be seen below, the only diffs in the golden files are in whitespace around inline comments. If you look at the original files, you'll see we only expect a space in the golden output for cases where the original also had whitespace.

lrewega:
> The good news is that these changes should mostly be idempotent for users: they shouldn't see a lot of new diff churn from formatting with this new version. As you'll see in the new golden files, the only new differences are the whitespace around inline comments. And this change tries to preserve existing whitespace there, so if a file was previously formatted with the old version and then re-formatted with the new, there shouldn't be a diff.

Do the golden files need to be altered at all then? I only did a cursory read of the patch, but I would expect to add new tests that demonstrate preservation of whitespace, making it clear that previous formatting is still valid.

jhump:
> Do the golden files need to be altered at all then? 

@lrewega, Yes. The old logic did _not_ try to preserve whitespace. So the existing golden files are missing whitespace where the original source had it.

> I would expect to add new tests that demonstrate preservation of whitespace, 

The new golden files, compared to the existing originals, do demonstrate that.

> making it clear that previous formatting is still valid.

But to more specifically demonstrate that the previous formatting doesn't get churned, I just did manual testing. I pulled all of the files into a separate directory and run `buf format -w` and the golden files were not touched.

```
jhumphries@Joshuas-MacBook-Pro:/Users/jhumphries/src/tmp/test (master)> /Users/jhumphries/.cache/buf/Darwin/arm64/gobin/buf format -w
jhumphries@Joshuas-MacBook-Pro:/Users/jhumphries/src/tmp/test (master *)> git status | grep golden                                     
jhumphries@Joshuas-MacBook-Pro:/Users/jhumphries/src/tmp/test (master *)> git status              
On branch master
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   customoptions/options.proto
	modified:   proto2/enum/v1/empty.proto
	modified:   proto2/enum/v1/enum.proto
	modified:   proto2/enum/v1/enum_trailing_comment.proto
	modified:   proto2/enum/v1/enum_trailing_newline.proto
	modified:   proto2/enum/v1/enum_value_trailing_comment.proto
	modified:   proto2/extend/v1/empty.proto
	modified:   proto2/field/v1/option.proto
	modified:   proto2/group/v1/empty.proto
	modified:   proto2/header/v1/import.proto
	modified:   proto2/header/v1/option.proto
	modified:   proto2/header/v1/package.proto
	modified:   proto2/license/v1/enum.proto
	modified:   proto2/license/v1/message.proto
	modified:   proto2/message/v1/empty.proto
	modified:   proto2/message/v1/message.proto
	modified:   proto2/message/v1/message_extensions.proto
	modified:   proto2/message/v1/message_group.proto
	modified:   proto2/message/v1/message_import.proto
	modified:   proto2/message/v1/message_options.proto
	modified:   proto2/message/v1/multiple_nested.proto
	modified:   proto2/message/v1/nested.proto
	modified:   proto2/message/v1/sparse.proto
	modified:   proto2/option/v1/option.proto
	modified:   proto2/option/v1/option_array.proto
	modified:   proto2/option/v1/option_complex_array_literal.proto
	modified:   proto2/option/v1/option_compound_name.proto
	modified:   proto2/option/v1/option_message_field.proto
	modified:   proto2/option/v1/option_name.proto
	modified:   proto2/option/v1/single_compact_option.proto
	modified:   proto3/all/v1/all.proto
	modified:   proto3/file/v1/option.proto
	modified:   proto3/header/v1/header.proto
	modified:   proto3/literal/v1/compound_string.proto
	modified:   proto3/literal/v1/literal.proto
	modified:   proto3/literal/v1/literal_comments.proto
	modified:   proto3/literal/v1/special_literal.proto
	modified:   proto3/oneof/v1/multiple.proto
	modified:   proto3/oneof/v1/simple.proto
	modified:   proto3/range/v1/range.proto
	modified:   proto3/service/v1/empty.proto
	modified:   proto3/service/v1/empty_rpc.proto
	modified:   proto3/service/v1/service.proto
	modified:   proto3/service/v1/service_options.proto
	modified:   proto3/service/v1/simple.proto

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	buf.yaml

no changes added to commit (use "git add" and/or "git commit -a")
```

joshcarp (approved):
Seems pretty straight forward to me, but then again this is the first part of `bufformat` that I've seen

bufdev (approved):
Really what I looked at was the golden tests, which make sense.

Add something to the changelog, i.e. "Fixed bug in formatter where spaces before inline comments were not always properly added."

jhump:
Just pushed a minor commit based on feedback from @pkwarren: a "dangling" space before a semicolon looks funny. So we won't try to preserve that whitespace.

private/buf/bufformat/testdata/proto2/option/v1/option_message_field.golden.proto
```diff
@@ -48,10 +48,10 @@ message Test {
         // Trailing comment on '{'.
 
         // Leading comment on 'foo'.
-        foo: "goo"/* Leading comment on ',' */, // Trailing comment on foo
+        foo: "goo" /* Leading comment on ',' */, // Trailing comment on foo
 
         // Leading comment on extension name.
-        [/* One */foo.bar.Test.Nested._NestedNested._garblez/* Two */]/* Three */: "boo" // Trailing comment on extension name.
+        [ /* One */ foo.bar.Test.Nested._NestedNested._garblez /* Two */ ] /* Three */ : "boo" // Trailing comment on extension name.
```
pkwarren, line 54: nit: This is another place where adding extra space after `[` and before `]` looks like it could be trimmed.

```suggestion
        [/* One */ foo.bar.Test.Nested._NestedNested._garblez /* Two */] /* Three */ : "boo" // Trailing comment on extension name.
```

private/buf/bufformat/testdata/proto2/option/v1/option_name.golden.proto
```diff
@@ -11,12 +11,12 @@ extend google.protobuf.FieldOptions {
...
-  option (/* One */something/* Two */) = 1;
+  option (/* One */ something /* Two */) = 1;
 
   string name = 1 [
     // Leading comment on deprecated.
-    deprecated/* After deprecated */ = true,
+    deprecated /* After deprecated */ = true,
     // Leading comment on another.
-    (/* One */another/* Two */) = 2
+    ( /* One */ another /* Two */ ) = 2
```
pkwarren, line 20: nit: Looks nicer here if there aren't added spaces added after '(' and before ')'.

```suggestion
    (/* One */ another /* Two */) = 2
```

pkwarren (approved):
Very nice improvement (and good to see the new protocompile get pulled in with fixes for some previous bugs).

bufdev:
Don't forget changelog!

private/buf/bufformat/testdata/proto2/message/v1/message_options.proto
```diff
@@ -13,8 +13,30 @@ option (custom.message_thing_option) = {
 
  /* Leading comment on the 'foo' element */ foo:  1  /* Comment on bar */ bar:   2  /* Comment on truth */  truth:false    }  ;
 
+
+
+      option (custom.message_thing_option)=  { /*leading*/ foo: 1 /*trailing*/ } /*leading*/;/*trailing*/
```
jhump, line 18: @pkwarren, I updated the formatter like you requested, to omit whitespace after open parens and braces and before close parens and braces.

I then decided to overhaul this file with a bunch of new examples with various kinds of braces/brackets/parens, to make sure the output looked good. And I discovered a few issues.

Unfortunately, the fixes to these issues _will_ produce some diff churn for users .But I think it is worth it because the new canonical format is _much_ easier to read IMO.

Here are the things I fixed:
* If an array literal or message literal had an inline comment before the closing brace, it was printed in a rather verbose and bad-looking way:
   ```
   [
      123

   /* inline comment */
   ]
   ```
   So I've fixed this to look better IMO:
   ```
   [
      123
      /* inline comment */
   ]
   ```
* There was an exception for simple compact options to be printed inline, on a single line. But there was no such exception for simple (single element) arrays and messages. So I've added one. If the array or message has a single element/field whose value is not a nested array or element, it can be printed all on one line. You can see an example in field `d` of the complex option below.

pkwarren:
Once the CI test failure is fixed and a changelog entry is added this looks good to me. Much more readable formatting.

private/buf/bufformat/testdata/proto2/message/v1/message_options.proto
jhump: Welp there are some nasty issues in this code. One is that it interprets "trailing comment" to mean something a little different than what it is. And, honestly, this probably warrants a change in protocompile.

Currently, the lexer attempts to classify comments as "trailing" for previous token or "leading" for current/next token. It does this using heuristics that try to emulate how `protoc` does this, for purposes of matching its source code info in descriptors.

_However_, those heuristics are too complicated and cause the interpretation of a comment as leading vs trailing to vary with subtle newline changes. That means that the act of formatting the code like above (removing an extra newline for compact options, message literals, and array literals) and removing a newline can cause a comment to be interpreted in a re-format as a trailing comment (when it was previously interpreted as a leading comment). That means that the format is inherently unstable: reformatting in that case would change it further.

I think what we really want is that the AST only calls something a trailing comment if it is on the same line as the element. And then, when computing source code info for the AST, the compiler may have to "borrow" some leading comments from the next token to populate trailing comments like `protoc` does. This "borrow" step being in the lexer and impacting the AST is the root issue right now.

Also, this code is very hard to maintain. There are are _lots_ of helper functions that are not named what they are ostensibly _for_ instead of being named after what they actually do. For example, there is `writeBodyEnd`, which is used to write the end of a body... but also used in myriad other places, such as elements in a compound string literal 🤷.

The way the helper methods are shared/re-used, but without clear names, means its easy to make a seemingly reasonable change in a helper but it to have surprising side effects because of how its re-used in weird places. Anyhow, I feel like I've only made the code messier in trying to get this to look a little better. And now I know just how brittle it is.

I think I will put this PR on hold and go and make the above-described changes to protocompile -- so the AST is more predictable and simple with what gets classified as a trailing comment and move the logic to classify like `protoc` does into its `sourceinfo` package. That will be a much more stable point on which to implement formatter improvements.

jhump:
I think this will help: https://github.com/bufbuild/protocompile/pull/46

jhump:
Okay. I think I've finally wrestled this thing into shape.

It's not pretty. I ended up making quite a lot of changes with the _hope_ of simplifying some of the nasty edge cases and whitespace handling. I'm not sure how well I succeeded (though the last commit is a net removal of about 80 LOC...).

But I think the golden files are looking good, including with some of the additions I made to test some add'l edge cases.

jhump:
@pkwarren, @bufdev, I think this is finally ready to go. Though I probably need to double-check the differences in golden outputs and amend/improve the change log wording before merging... PTAL.

private/buf/bufformat/formatter.go
```diff
@@ -1861,31 +1917,41 @@ func (f *formatter) writeLineEnd(node ast.Node) {
...
+		compact = false
+		txt := comment.RawText()
+		if strings.HasPrefix(txt, "/*") && newlineCount(txt) > 0 {
+			lines := strings.Split(txt, "\n")
+			for j, line := range lines {
+				line = strings.TrimSpace(line)
+				if j == 0 || line == "*/" {
+					f.P(line)
+				} else {
+					f.P("   ", line)
```
pkwarren, line 1949: Does this need to take into account indentation level?

private/buf/bufformat/formatter.go
```diff
@@ -19,23 +19,33 @@ import (
...
 	previousNode ast.Node
 
+	// if true, a space will be written to the output unless the next character
+	// written is a newline (don't wait errant trailing spaces)
+	pendingSpace bool
+	// if true, the formatter is in the middle of printing compact options
+	inCompactOptions bool
+
+	pendingIndent int
+	inline        bool
```
pkwarren, line 47: It might be nice to add comments to these - there are some inline comments around use of defer that might make sense to move here?

private/buf/bufformat/testdata/proto2/message/v1/message_import.golden.proto
```diff
@@ -2,11 +2,9 @@ syntax = "proto2";
 
 import "custom.proto";
 
-message Thing {
-  // Trailing comment on '}'.
-
+message Thing { // Trailing comment on '}'.
```
pkwarren, line 5: This is one place where I feel like the original formatting reads a little nicer, however this does better to keep the original format of the file in place.

pkwarren (approved):
Really nice job with refactoring and fixing some edge cases. The golden files look much better to me and it is good that the bugs where comments were dropped is now fixed. The code in the formatter reads much nicer too.

Just a couple of minor comments (feel free to ignore the nit picks).

private/buf/bufformat/formatter.go
jhump: It does. `f.P(...)` prints the indentation, then the given args, then a newline. (If not args, indentation is skipped and it just prints empty line.)

private/buf/bufformat/formatter.go
jhump: done

private/buf/bufformat/formatter.go
```diff
@@ -65,17 +75,29 @@ func (f *formatter) P(elements ...string) {
...
-		f.WriteString(elem)
+		f.Indent(nil)
+		for _, elem := range elements {
+			f.WriteString(elem)
+		}
 	}
+	//_, file, line, ok := runtime.Caller(1)
+	//if ok {
+	//	f.WriteString(fmt.Sprintf("\t\t\t%s:%d\n", file, line))
+	//} else {
```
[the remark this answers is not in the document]
jhump: Doh! I'm glad you're looking closely. I've left this kind of crap in on several recent PRs 🤦 

Removed.

PR #1430, opened 2022-09-22
change page token in SearchTag and SearchDraft from uint32 to string

oliversun09:


mfridman (approved):
lgtm. Please change the comments above the fields since they are no longer `0` and empty strings

Issue #1443, opened 2022-09-27
Windows: Failing to resolve files in `buf.work.yaml` symbolic links


andrewparmet:
I'm using symbolic links in a `buf.work.yaml` to point to `.proto` files. On Mac OS and Linux `buf` is able to traverse the links and find the proto files, but on Windows I get `Failure: directory "build-extracted--include--protos-main" listed in buf.work.yaml contains no .proto files`. Is that supported functionality in `buf`? I've recreated the file structure @amckinney used in #550 on a Windows VM and I got the same error (`Failure: directory "two" listed in buf.work.yaml contains no .proto files`), so I don't think it's particular to my project setup.

Thanks!

CI build for reference: https://github.com/andrewparmet/buf-gradle-plugin/actions/runs/3129607613/jobs/5078908186

timostamm:
Thank you for the report, @andrewparmet! Symlink support was disabled on Windows because of limited support outside of Windows 10. We are changing this in #1445. Your use case should be solved on Windows 10 once the change is released.

andrewparmet:
Verified that symlink-based tests are now passing. Thanks!

PR #1445, opened 2022-09-27
Update storageos package to support following symlinks on Windows

tehbilly:
This should address #1443


bufdev:
It was intentional - I don't remember exactly the reasoning, but from our research, symlinks in Windows were (1) not normal (2) required elevated permissions, if I remember correctly. We didn't want to support what we didn't understand, and unless we have more research on the topic, we should stick to that.

bufdev (changes requested):
Blocking unless we have more data on symlinks in Windows

bufdev:
@doriable might have some recollection of this as well

tehbilly:
The creation of symlinks prior to Windows 10 required admin privileges, but [as of 2016](https://blogs.windows.com/windowsdeveloper/2016/12/02/symlinks-windows-10/) it can be done with either admin privs or by using [developer mode](https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development). However, there shouldn't be any permissions for reading/following the links unless they're explicitly added _(just like normal files)_.

The included tests, the ones I removed the `t.Skip` call from, do actually catch this case. If you check out the repo with git's `core.symlinks` set to `false` the tests will fail, but with it set to `true` the tests pass the same as on MacOS/Linux.

bufdev:
I'd want a few more pieces of input about this as I remember this being a universal/uncontroversial decision - do we have others who remember the context here with Windows? This is a one-way-door that we can't take back, so I don't want to make this decision lightly.

tehbilly:
Absolutely agree! If there were specific concerns it'd be nice to have a chance to address them. If not, perhaps we could add a warning when they're encountered on Windows, letting users know that they're skipped intentionally and they're not encountering a bug.

doriable:
I think the original decision was very much based on the fact that outside of Windows 10, symlinks are bespoke on NTFS. That being said, given that it does work on Windows 10, I can see the argument for enabling this. I think there should be an understanding that this will absolutely break on older systems and we will not be supporting that, but otherwise this seems like a reasonable course of action?

doriable (approved):

bufdev (approved):
We're just going to go ahead with this, having done more research

PR #1446, opened 2022-09-27
Rename repository to module throughout the surface layer of buf

joshcarp:
## TLDR of changes
- APIs, protos, generated client code are not changed and are still using `repository`
- Anything that requires objects directly from the API layer uses the term `repository ` (PrintRepositoryCommit function taking a RepositoryCommit)
- All comments and strings are using `module` unless specifically referring to something `RepositoryService`
- Nothing in `bufprint`, `buflock` or `bufmoduleref ` are changed because they are tightly coupled with the api layer (except comments referring to the structure of a fqn module name)
- fqn module names that refer to the current structure `remote/owner/repository` have changed to `remote/owner/name` and not `remote/owner/name` because `module` would be an ambiguous term with the fqn module name.
- flags that referenced `--repository` are now `--name` (up for discussion)

## List of changes from `repository` to `module`

- Any reference to what is currently called "module name" (which will soon be "fully qualified module") has changed from
`remote/owner/repository` -> `remote/owner/name` in all comments that aren't API comments (proto or generated)
- here `repository` is swapped for `name` when referring to the structure of the fully qualified module name

- Any reference to the specific object of a `repository` is now referred to as a `module`:
    - "Create a new repository on the BSR." -> "Create a new module on the BSR." 

- Commands, package names, filenames for the following commands have been changed from `repository -> module`

```
repositorycreate -> modulecreate
repositoryget -> moduleget
repositorylist -> modulelist
repositorydelete -> moduledelete
repositorydeprecate -> moduledeprecate
repositoryundeprecate -> moduleundeprecate
repositoryupdate -> moduleupdate
```

- Any reference to "repository" or "repo" in strings that are not in api files have been changed to "module" including any logs, error messages UNLESS they are referencing a specific field;

- `fmt.Errorf("a repository named %q already exists", name) -> fmt.Errorf("a module named %q already exists", name)`

<details>
 <summary>Flags (both string representations and go structures) that currently have the name "repository"  as this is more tightly coupled with the frontend cli and less with the API layer. `webhookcreate.go` and `webhooklist.go`</summary>

```diff

 type flags struct {
       OwnerName      string
-       RepositoryName string
+       Name    string
       Remote    string
 }

- repositoryFlagName   = "repository"
+ moduleFlagName       = "name"
```

</details>

## Places that are NOT changing from `Repository` to `Module`

- `buf/private/bufpkg/buflock`
- `private/buf/bufprint`
- `bufmoduleref`
- Identifiers that are heavily coupled with the API layer

<details>
<summary>
bufprint is heavily coupled with the api layer
</summary>

```
PrintRepositoryCommit(ctx context.Context, format Format, message *registryv1alpha1.RepositoryCommit) error {
```
- Above it would be weird for a `PrintModuleCommit` to be printing a `*registryv1alpha1.RepositoryCommit`.

</details>

<details>
<summary>
`bufmoduleref` is heavily coupled with the api layer
</summary>

- This is essentially an abstraction on the core api which uses `repository` heavily. I already tried changing this package to use `module` functionality and this was the ugliest I saw
```
func newModulePin(... module string ) (*modulePin, error) { // Module here
	return newModulePinForProto(
		&modulev1alpha1.ModulePin{
			Repository: module, // gets converted into repository here
		},
	)
}
```
- Whilst this doesn't seem too bad there very quickly becomes a mental tax on the developer of changing terms from `repository` to `module` within a package 

`ModuleIdentity` also is an abstraction on the API, so changing `Repository` to `Module` wouldn't work nicely
type ModuleIdentity interface {
	ModuleOwner

	Repository() string

	// IdentityString is the string remote/owner/name.
	IdentityString() string

	isModuleIdentity()
}

</details>

<details>
<summary>
Errors that reference a specific field on an object are NOT changed
</summary>

```
if protoModuleName.GetRepository() == "" {
	return errors.New("empty ModuleName.Repository")
}
```

</details> 

- The following would not make sense to change as `ModuleName.Repository` exists but `ModuleName.Module` does not
```
if protoModuleName.GetRepository() == "" {
	return errors.New("empty ModuleName.Repository")
}
```

- Error messages that are referring to an API level construct "repository service"
```
return nil, fmt.Errorf("a repository service is not available for %s", dependencyRemote)
```

- `ModuleIdentity` which has `Repository` method. Because this is only a slight abstraction on top of an API field it doesn't make much sense to change this method to `Module`. 

- newModulePinForProto or any modulePin identifiers; modulePin needs to use `repository` for compatibility reasons so changing anything from `repository` to `module` would be an eye sore.

## Transition points between `Repository` and `Module` 

- In `webhooklist.go` the `type flags struct {` has a `Name    string` flag
This gets converted into a `Repository` as follows:

```

results, _, err := service.ListWebhooks(ctx, flags.Name, flags.Owner, "")
	if err != nil {
		return err
	}

```

- in `module<crud>.go` it's even less subtle; because the argument to the function is the "fully qualified module name" the Use is stated as `Use:   cmdName + " <buf.build/owner/name>",` but it the argument gets passed as 
```
moduleIdentity, err := bufmoduleref.ModuleIdentityForString(container.Arg(0))
```


private/buf/bufcli/errors.go
```diff
@@ -114,7 +114,7 @@ func NewOrganizationNameAlreadyExistsError(name string) error {
 // NewRepositoryNameAlreadyExistsError informs the user that a repository
 // with that name already exists.
 func NewRepositoryNameAlreadyExistsError(name string) error {
-	return fmt.Errorf("a repository named %q already exists", name)
+	return fmt.Errorf("a module named %q already exists", name)
```
renahlee, line 117: Wondering if `a module with fully qualified name <FQN> ...` might be more appropriate here (but could also be more than a surface level rename 😅 ).

private/buf/cmd/buf/command/beta/registry/draft/draftdelete/draftdelete.go
```diff
@@ -36,8 +36,8 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name + " <buf.build/owner/repo:draft>",
-		Short: "Delete a draft of a BSR repository by name.",
+		Use:   name + " <buf.build/owner/name:draft>",
```
renahlee, line 39: (Know this is a simple find and replace but for clarity) is there a distinction between a `BSR module` versus a `module`?

private/buf/cmd/buf/command/beta/registry/module/moduledelete/moduledelete.go
```diff
@@ -36,8 +36,8 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name + " <buf.build/owner/repository>",
-		Short: "Delete a BSR repository by name.",
+		Use:   name + " <buf.build/owner/name>",
```
renahlee, line 39: For clarity of usage here, what is the `name` as an argument of this function?

renahlee (approved):
So quick 🚀

private/buf/cmd/buf/command/beta/registry/draft/draftdelete/draftdelete.go
joshcarp: technically there are local modules which are just the module before it's pushed to the bsr

bufdev (commented):
The module stuff looks basically complete. There's some snapshot stuff interleaved in here, is that on purpose? I still see a lot of commit stuff when I check out the branch.

joshcarp:
> There's some snapshot stuff interleaved in here, is that on purpose? 

sort of on purpose, it was on another branch that got merged into this one. https://github.com/bufbuild/buf/pull/1448

joshcarp:
snapshot PR is here: https://github.com/bufbuild/buf/pull/1452

joshcarp:
Closing because the decision has been reached that the rename becomes ambiguous when referring to `bsr modules` and local modules

PR #1450, opened 2022-09-28
Add options to doc proto

seankimdev:
- Adds the selected built-in options to be displayed in the bufdoc.
- For rendering implicitly deprecated elements (which its parent file is deprecated) in the package documentation, adds `file_deprecated` to each of the types (service, enum, message and file extension)

Product Description Notion: https://www.notion.so/bufbuild/Product-Description-17fa5fc450da49659bd048fa7dfbfaa9


timostamm (commented):
I have a couple of questions:

1. This replicates 7 messages from google/protobuf/descriptor.proto. So far we have avoided this - we introduced our own `message Field` instead of relying on `FieldDescriptor`. Are we going to update this if descriptor.proto changes? Related to this: Do we need to reserve field numbers?

2. What are we going to put into `repeated string custom_options`? Can't we simply put _all_ options in there? 

3. How do we want to represent options in the UI? We currently render a representation of the type that is very similar to the original source. Is the plan to render options similar to the original source? Do we want to preserve the formatting?

4. Do we want to enable links for custom options? For example, if I use one of the custom options from protoc-gen-validate, can I click on the option in the bufdoc UI to jump to the option definition?

For context, here are some examples for custom options:
https://github.com/envoyproxy/protoc-gen-validate#protoc-gen-validate-pgv
https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L185

seankimdev:
@timostamm, thank you for the questions.

> 1. This replicates 7 messages from google/protobuf/descriptor.proto. So far we have avoided this - we introduced our own `message Field` instead of relying on `FieldDescriptor`. Are we going to update this if descriptor.proto changes? Related to this: Do we need to reserve field numbers?

Can you elaborate on why we want to avoid replicating message names from `google/protobuf/descriptor.proto` for me? Would it help if we change the names used here slightly? If the built-in options from the `descriptor.proto` ever change, I assume we'd have to update these. 
I reserved the field numbers and names for the known built-in options which we are not including with this phase. Is there no benefit from doing so?

> 2. What are we going to put into `repeated string custom_options`? Can't we simply put _all_ options in there?

`custom_options` would include all other options, which we don't know the types of, besides the known built-in options. I thought it would be beneficial to explicitly define the types of built-in options than including them all as just strings. What do you think?

> 3. How do we want to represent options in the UI? We currently render a representation of the type that is very similar to the original source. Is the plan to render options similar to the original source? Do we want to preserve the formatting?

Yes, for the first phase, we want to render the options similar to the original source, preserving the current bufdoc format.
For the next phase, the "tag" type built-in options, like deprecated which we know what it means, Karol is working on a rough design idea to render them with customized UI.

> 4. Do we want to enable links for custom options? For example, if I use one of the custom options from protoc-gen-validate, can I click on the option in the bufdoc UI to jump to the option definition?
> 
> For context, here are some examples for custom options: https://github.com/envoyproxy/protoc-gen-validate#protoc-gen-validate-pgv https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L185

I haven't put much thought into the custom options yet as it's currently scoped to only the built-in options, but it seems reasonable to enable links for them. I guess we'll need to create a message for `custom_option` to include the address besides just a string then?

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -233,6 +265,14 @@ message EnumValue {
   // Paragraph newlines (double new lines) are respected, however single newlines are not.
   // Note that any leading and trailing `//` or spaces within a `/* */` block will be stripped.
   string description = 3;
+  EnumValueOptions enum_value_options = 4;
+}
+
+// EnumValueOptions provide information for the documentation of options for an enum value.
+message EnumValueOptions {
+  reserved "UninterpretedOptions";
```
doriable, line 273: Is there a reason why this particular version of `uninterpreted_options` is in camel case?

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -190,6 +198,21 @@ message Method {
...
   MethodRequestResponse response = 4;
+  MethodOptions method_options = 5;
+}
+
+// MethodOptions provide information for the documentation of options for a method.
+message MethodOptions {
+  reserved "uninterpreted_options";
+  bool deprecated = 1;
+  IdempotencyLevel idempotency_level = 2;
+  repeated string custom_options = 1000;
```
doriable, line 209: For the custom options, is there a reason why we are matching this to descriptor proto and using the field value `1000`? My understanding is for the file descriptor, specifically, 1000+ is for extensions. This should not be a concern here, right?

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: Nope, certainly not

timostamm:
> Can you elaborate on why we want to avoid replicating message names from `google/protobuf/descriptor.proto` for me? 

My concern is not about replicating message names, but messages. If we want descriptor.proto, we should simply use descriptor.proto. Replicating it with only slight differences will make it difficult to compare the two. Descriptors are not suitable to reconstruct the original source, and we will fight that problem in both BE and FE.

> Yes, for the first phase, we want to render the options similar to the original source, preserving the current bufdoc format. 

In that case, I recommend to remove all *Options messages, and replace them with a simple `repeated string options` field.

For example, field options would be `['json_name = "foo"', 'ctype = CORD']`, rendered as `[json_name = "foo", ctype = CORD]`.

Enum options would be `['allow_alias = true', 'deprecated = true']`, rendered as:

```
  option allow_alias = true;
  option deprecated = true;
```


> I haven't put much thought into the custom options yet as it's currently scoped to only the built-in options

In that case, I recommend to not define them here yet.

seankimdev:
> My concern is not about replicating message names, but messages. If we want descriptor.proto, we should simply use descriptor.proto. Replicating it with only slight differences will make it difficult to compare the two. Descriptors are not suitable to reconstruct the original source, and we will fight that problem in both BE and FE.

> In that case, I recommend to remove all *Options messages, and replace them with a simple `repeated string options` field.

Displaying them in source code style is the step one, but the next will involve adding more UX to it, such as a tooltip on hover to show the definition of the option. The reason for the *Options message structure is to allow its expansion. These messages would evolve to be different from the `descriptor.proto.`

> > I haven't put much thought into the custom options yet as it's currently scoped to only the built-in options
> 
> In that case, I recommend to not define them here yet.

The custom options would be displayed in the source code style.

All of these valid concerns are due to me not planning and providing enough context upfront. I jumped to the data structure too early. I will get back to this with more refined proposal.

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: Correct. I was initially thinking it'd be good to keep the field numbers similar to the descriptor proto, but it would rather create unnecessary confusion instead. Changed them.

seankimdev:
@timostamm, in order to scope this project phase to only the desired built-in options described in the [doc](https://www.notion.so/bufbuild/Product-Description-17fa5fc450da49659bd048fa7dfbfaa9), I've removed the custom_options field for now. 
I kept the message definitions for the option of each element for the following reasons:
1. To allow for its expansion with the addition of more built-in options down the road.
2. I think preserving the primitive types separately for each built-in options, such as a deprecated boolean, rather than listing them in a single array of strings, would make it easier for the frontend to use than parsing the array of strings.
3. The protosource, which the methods for the bufdoc currently gets the data from, already returns most of the options we need now in their primitive types. This structure will make it easy to map the returned value for each built-in type. 
Please let me know your thoughts!

timostamm:
> [...] I kept the message definitions for the option of each element for the following reasons:
> 
> 1. To allow for its expansion with the addition of more built-in options down the road.
> 2. I think preserving the primitive types separately for each built-in options, such as a deprecated boolean, rather than listing them in a single array of strings, would make it easier for the frontend to use than parsing the array of string options.

I'd argue that for the purpose of rendering options like they appear in the protobuf source, an array of strings is actually easier to handle in the FE because it doesn't need parsing or formatting at all. It will also be easier to add more built-in options in the future, because it will be a BE-only change.

> 3. The protosource, which the methods for the bufdoc currently gets the data from, already returns most of the options we need now in their primitive types. This structure will make it easy to map the returned value for each built-in type.

I agree that you don't have to format option values on the BE with this structure, but instead you have to do so on the FE, so it doesn't really reduce the work.

But I see from the doc (thanks for the context!) that custom styling for deprecated elements is a goal. A dedicated field is much better suited for that. You could have both - a `bool deprecated` _and_  a `repeated string options` for the rest. 

With the set of options we plan to surface, I'm definitely not blocking on making all of them fields. But I want to raise another edge case first: What happens to a message if the _file_ is marked deprecated?

For context/example, see [here](https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf-test/extra/deprecation-implicit.proto#L19) and [here](https://github.com/bufbuild/protobuf-es/blob/main/packages/protobuf-test/src/gen/ts/extra/deprecation-implicit_pb.ts#L24).

seankimdev:
Spoke with @timostamm offline.
We eventually want to show extra information about the options, such as their definitions, therefore its own message field is the way to go. We want to understand the options, and repeated strings is only good if we only want to render as source.
We don't want to include the options' definitions here, since they are very unlikely to change. Suggests hardcoding them in the frontend if needed.
In the case of option values that are based on enums, they use int32, because we cannot use proto2 enum in a proto3 message. The frontend will have to change the int value to the respective num values by importing from `@bufbuild/protobuf`.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -190,6 +196,13 @@ message Method {
   string description = 2;
   MethodRequestResponse request = 3;
   MethodRequestResponse response = 4;
+  MethodOptions method_options = 5;
+}
+
+// MethodOptions provides information for the documentation of options for a method.
+message MethodOptions {
+  bool deprecated = 1;
+  int32 idempotency_level = 2;
```
timostamm, line 205: This should have a comment saying that it holds a value of the enumeration `google.protobuf.MethodOptions.IdempotencyLevel`.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -303,6 +335,15 @@ message Field {
...
   string extendee = 9;
+  FieldOptions field_options = 10;
+}
+
+// FieldOptions provides information for the documentation of options for a field.
+message FieldOptions {
+  bool deprecated = 1;
+  bool packed = 2;
+  int32 ctype = 3;
+  int32 jstype = 4;
```
timostamm, line 346: These two should also have comments, similar to `idempotency_level` above.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -274,6 +301,11 @@ message MessageField {
   }
 }
 
+// MessageOptions provides information for the documentation of options for a message.
+message MessageOptions {
+  bool deprecated = 1;
```
timostamm, line 306: What happens to a message if the _file_ is marked deprecated - will we also show it as deprecated in the UI?

timostamm (commented):
This looks reasonable to me, but I'm still wondering about file deprecation.

seankimdev:
> This looks reasonable to me, but I'm still wondering about file deprecation.

Sorry, I completely missed responding to that. We currently don't have a plan to show the file level deprecation, since the current bufdoc design is only rendered per package documentation, and don't expose their files. The plan is to visit the file deprecation with the bufdoc v2, where we would consider showing the files.

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: hm that's a good question. If a file is deprecated, I think it makes sense for every element defined inside it to be deprecated.

timostamm: I agree. 

This is a very similar situation to protobuf-es. In JavaScript, you can deprecate all kinds of elements, but not an entire file. That is why treat all elements in a deprecated file as _implicitly_ deprecated.

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: Good idea. Added. Thanks!

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: Added

proto/buf/alpha/registry/v1alpha1/doc.proto
seankimdev: Right. I guess that doesn't require a change in this proto structure, does it?
I'd need to look into how I can map `deprecated = true` for the `implicitly` deprecated elements in the business layer. OR, would it be better off just overwriting them as deprecated in the frontend if the file is deprecated?

timostamm: You have two options here:
1. In the BE, set the `deprecated` field for types whose parent file is deprecated
2. Change the proto structure to include information about file level deprecation, and add logic in the FE

Do you think there are any showstoppers or disadvantages with either option for what you want to do?

seankimdev: The option 1 seems like the way to go, minimizing data transfer, unless we want to differentiate those `deprecated`  vs. `implicitly deprecated` elements rendered in the frontend. I'm not sure if its distinction would provide any values for consumers though. 🤔 What do you think?

timostamm: For option 2, consider the following proto file:

```proto
option deprecated = true;

message MyMessage {
  bool x = 1;
}
```

We would render it in bufdoc as:

## ~~MyMessage~~

```proto
message MyMessage {
  option deprecated = true;
  bool x = 1;
}
```

Do we _want_ to insert line 2, or do we only do that because we modeled the data structure in a way we have no other option? It doesn't seem correct to me to show something in the source that is not there.

seankimdev: Thank you for the example. I see it now. It definitely doesn't look correct to have a deprecated line in the source, which would be inevitable if we don't distinguish them in the data structure. We'll have to provide information about file level deprecation.

seankimdev: Added `FileOptions` to include `deprecated`

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -58,6 +58,12 @@ message FileInfo {
   string path = 1;
   bool is_dir = 2;
   repeated FileInfo children = 3;
+  FileOptions file_options = 4;
```
timostamm, line 61: Adding file options to this `message FileInfo` is not going to help us, unfortunately. If you take a look at `rpc GetPackageDocumentation`, you'll see that it doesn't return `FileInfo` anywhere, it only returns the messages, enumerations and services.

Each of the types has a couple of fields providing information about the file they are declared in:

```proto
  // file_path is the normalized path of the file containing the service.
  // This is used for navigating to the source code for the service.
  string file_path = 5;
  Location location = 6;
```

This is where file level deprecation is best added. 

Since package documentation is not concerned with files at all (it only needs the file path and location so that clients can link to source files), I do not think adding a `message FileOptions` is helpful. IMO, a plain `bool file_deprecated` field is sufficient.

seankimdev: Thanks Timo. I missed that the FileInfo is not returned from the `rpc GetPackageDocumentation`. I agree that the `message FileOptions` is not necessary in that case, since the `file_deprecated` is added separately to each of the types.
I think `message FileOptions` will be necessary when we decide to render other options at the file level, but it's not part of the current scope, so I'll remove it.

timostamm (approved):

doriable (approved):

SyrieBianco (approved):
So there's nothing I can add more to this conversation than has already happened, I approve these changes and hope we can unblock the backend to open up phase 2 of this project.

PR #1453, opened 2022-09-29
Rename repository, commit to module, snapshot in audit log protos

njhale:
Rename audit log enums to replace "repository" verbiage w/ "module"

See also: #1446


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -156,84 +156,84 @@ message PayloadOrganizationMemberRemoved {
   buf.alpha.registry.v1alpha1.OrganizationRole member_role = 3;
 }
 
-message PayloadRepositoryCreated {
-  // owner_id is the id of the owner of the repository.
+message PayloadModuleCreated {
```
njhale, line 159: The motivating [slack thread]() only mentioned the enums, but I assumed that we wanted to update the message names as well -- is this assumption valid? should I drop this change?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -156,84 +156,84 @@ message PayloadOrganizationMemberRemoved {
...
-  string repository_id = 3;
-  // repository_name is the name of the repository with the new contributor.
-  string repository_name = 4;
-  // contributor_role is the role granted to the contributor added to the repository.
+  // module_id is the id of the module with the new contributor.
+  string module_id = 3;
+  // module_name is the name of the module with the new contributor.
+  string module_name = 4;
+  // contributor_role is the role granted to the contributor added to the module.
   buf.alpha.registry.v1alpha1.RepositoryRole contributor_role = 5;
```
njhale, line 198: I didn't change the name of `buf.alpha.registry.v1alpha1.RepositoryRole` since it's more widely used and part of our "v1 supported" APIs; figured renaming it would be part of a larger effort (if at all possible)

saquibmian: How do I determine if something is part of our supported APIs? I would think as part of the rename, everything gets renamed?

njhale: > How do I determine if something is part of our supported APIs?

I believe everything in the `v1alpha1` registry API is being considered "v1", meaning we can't make backwards incompatible changes. I would need to defer to @smallsamantha for more detail though.

> I would think as part of the rename, everything gets renamed?

Agreed, but I think it's safer and more tightly scoped to modify _just_ the messages in the audit APIs.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -43,9 +43,9 @@ enum ResourceType {
   RESOURCE_TYPE_USER = 1;
   RESOURCE_TYPE_ORGANIZATION = 2;
   RESOURCE_TYPE_ORGANIZATION_MEMBER = 3;
-  RESOURCE_TYPE_REPOSITORY = 4;
-  RESOURCE_TYPE_REPOSITORY_CONTRIBUTOR = 5;
-  RESOURCE_TYPE_REPOSITORY_COMMIT = 6;
+  RESOURCE_TYPE_MODULE = 4;
+  RESOURCE_TYPE_MODULE_CONTRIBUTOR = 5;
+  RESOURCE_TYPE_MODULE_COMMIT = 6;
```
saquibmian, line 48: Commit is also changing to Snapshot ([PR](https://github.com/bufbuild/buf/pull/1448)). Are you planning a second PR for that change?

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: `v1alpha` is still alpha and subject to changes, BSR (UI and API) are not yet at v1, although breaking changes not much desirable, this module+snapshot thing is being changes almost everywhere.

saquibmian: I found this:
> - APIs, protos, generated client code are not changed and are still using repository
> - Anything that requires objects directly from the API layer uses the term repository  (PrintRepositoryCommit function taking a RepositoryCommit)

From https://linear.app/bufbuild/issue/TCN-507/rename-commit-to-snapshot-repository-to-module

So it looks like we don't need to rename this existing stuff, just audit logging. I won't block this PR on it, but could you confirm with @joshcarp?

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: Ah, missed that earlier. I think it's small enough to pull into this PR, no?

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: > So it looks like we don't need to rename this existing stuff, just audit logging.

This was my takeaway from a [conversation in yesterday's team recap](https://www.notion.so/bufbuild/BSR-Team-Recap-a0d0c15b347b47c58aa38e86027da7fc#4e4d6b8a48864a6a8d43b7124edc5afd) cc @doriable 

> I won't block this PR on it, but could you confirm with @joshcarp?

sure thing!

proto/buf/alpha/audit/v1alpha1/event.proto
njhale: just did 😈

proto/buf/alpha/audit/v1alpha1/event.proto
joshcarp: API changes were already attempted but deemed too messy to be done throughout everything.
This set of changes are meant to just be the surface level with only the frontend/buf cli being changed.
It was decided in a meeting with @renahlee and @SyrieBianco and I that a complete refactor has already been attempted and was deemed too messy so stuff like the API and the buf.yaml files should remain as `repository` and `commit` at the moment.

joshcarp: Although obviously if the API is getting promoted to stable soon then we should definitely talk about renaming it before that happens, but that's not the case with the cli tool as everything referencing `repository` and `commit` are still in beta

njhale:
This was put on ice for the moment.

PR #1455, opened 2022-09-29
move license SPDX ID from repository to repository commit

cyinma:
fixes BSR-571

license is a per commit concept, it can be changed between different commits in the same repository, so the license SPDX ID should actually tie to repository commit instead of repository


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -54,11 +54,6 @@ message Repository {
   // url is the user configurable URL in the description of the repository,
   // always included the scheme and will not have a #fragment suffix.
   string url = 12;
-  // license_spdx_id is the current license of the repository, based on the
-  // license file pushed, which should be one of the identifier defined in
-  // https://spdx.org/licenses, and will be empty if the license file is not
-  // presented or cannot be classified into a known license.
-  string license_spdx_id = 13;
```
cyinma, line 61: we might want to reserve the field name and number for this, but as this field is added very recently and we never actually use it, it should be fine to just remove it

doriable: If it has not been used, I think the breaking change is reasonable here.

doriable (approved):
This change looks reasonable to me!

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -42,6 +42,11 @@ message RepositoryCommit {
   repeated buf.alpha.registry.v1alpha1.RepositoryTag tags = 8;
   // The commit's draft name, if it is a draft commit.
   string draft_name = 9;
+  // license_spdx_id is the license of the commit, based on the
+  // license file pushed, which should be one of the identifier defined in
+  // https://spdx.org/licenses, and will be empty if the license file is not
```
twilly, line 47: Nit: "not set" over "empty" because the empty string be sent over the wire and may confer meaning.

twilly (approved):
Small comment nit, but this is certainly the right thing to do.

PR #1456, opened 2022-09-29
Osun/bsr 601/add proto definition for get resource

oliversun09:
Add GetResource service to allow FE to get the correct resource (either repo or plugin) by owner + name. See BSR-601


paul-sachs (approved):
LGTM, at least as far as contracts go.

proto/buf/alpha/registry/v1alpha1/resource.proto
```diff
@@ -0,0 +1,45 @@
...
+import "buf/alpha/registry/v1alpha1/repository.proto";
+
+message Resource {
+  oneof resource {
+    Repository repository = 1;
+    Plugin plugin = 2;
+  }
+}
+
+// ResourceService is a service that provides RPCs that allow BSR to query resources
```
stefanvanburen, line 29: maybe

```suggestion
// ResourceService manages resources.
```

since in the future it might do more?

stefanvanburen (approved):
lgtm!

proto/buf/alpha/registry/v1alpha1/resource.proto
oliversun09: You're right.

PR #1460, opened 2022-10-03
Remove plugin pushed audit log event

njhale:
BSR doesn't expose services/rpcs for _pushing_ plugins directly; remove related types from the audit logging protos.

Resolves BSR-624


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -76,7 +76,6 @@ enum EventType {
   EVENT_TYPE_REPOSITORY_VISIBILITY_CHANGED = 12;
   EVENT_TYPE_PLUGIN_CREATED = 13;
   EVENT_TYPE_PLUGIN_DELETED = 14;
-  EVENT_TYPE_PLUGIN_PUSHED = 15;
```
unmultimedio, line 79: This is not public yet, and no one is using it yet. I'd re-sort indexes.

unmultimedio (approved):

PR #1463, opened 2022-10-04
Use protocompile as the compiler

jhump:
This replaces the prior compiler (`github.com/jhump/protoreflect/desc/protoparse`) with a new one in `github.com/bufbuild/protocompile`.

- Weave build options through more of the API since there are now more flags that external callers are interested in than just "exclude source info".
- Update `buf build` and `buf generate` so they use descriptors that match those of protoc byte-for-byte.
- Update `buf lint` to generate *extra* comments, since the default source info from protocompile matches protoc. The level of comments generated by protoc is too low for linting, which wants extra comments (as captured by prior compiler, protoparse) for checking for nolint annotation comments.

Fixes TCN-294


private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -81,132 +79,73 @@ func (b *builder) build(
...
-	buildResults := getBuildResults(
+	buildResult := getBuildResult(
 		ctx,
 		parserAccessorHandler,
 		paths,
-		excludeSourceCodeInfo,
+		options,
 	)
-	var buildResultErr error
-	for _, buildResult := range buildResults {
```
jhump, line 91: Much of the red lines in this file are because `protocompile` natively handles concurrency. So a lot of logic in here that merged chunks of built results is no longer needed.

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -81,132 +79,63 @@ func (b *builder) build(
...
 ) *buildResult {
-	var errorsWithPos []protoparse.ErrorWithPos
-	var warningErrorsWithPos []protoparse.ErrorWithPos
-	var lock sync.Mutex
-	parser := protoparse.Parser{
-		IncludeSourceCodeInfo: !excludeSourceCodeInfo,
-		Accessor:              parserAccessorHandler.Open,
-		ErrorReporter: func(errorWithPos protoparse.ErrorWithPos) error {
-			// protoparse isn't concurrent right now but just to be safe
-			// for the future
```
jhump, line 191: `protocompile` _is_ concurrent. However, this explicit locking is unnecessary as it handles thread-safe access to the given reporter internally.
https://pkg.go.dev/github.com/bufbuild/protocompile@v0.0.0-20221004142453-757ef5688631/reporter#Reporter

CHANGELOG.md
```diff
@@ -12,6 +12,22 @@
 - Formatter uses a compact, single-line representation for array and message literals
   in option values that are sufficiently simple (single scalar element or field).
 - Include `LICENSE` file in the module on `buf push`.
+- New compiler that is XX% faster, uses XX% less memory, and addresses a few bugs
```
bufdev, line 15: Make sure we update before merging

go.mod
```diff
@@ -4,14 +4,14 @@ go 1.18
...
-	github.com/bufbuild/protocompile v0.0.0-20220927163942-e0e013a8d791
+	github.com/bufbuild/protocompile v0.0.0-20221004130402-a20e4810edfa
 	github.com/docker/docker v20.10.18+incompatible
 	github.com/go-chi/chi/v5 v5.0.7
 	github.com/gofrs/flock v0.8.1
 	github.com/gofrs/uuid v4.3.0+incompatible
 	github.com/google/go-cmp v0.5.9
 	github.com/jdxcode/netrc v0.0.0-20210204082910-926c7f70242a
-	github.com/jhump/protoreflect v1.12.1-0.20220721211354-060cc04fc18b
+	github.com/jhump/protoreflect v1.13.1-0.20220928232736-101791cb1b4c
```
bufdev, line 14: We should file an issue to rework the test in `bufimageutil_test.go` to not use `protoprint`, which I think is possible given what I see. A good CLI onboarding issue probably.

make/buf/all.mk
```diff
@@ -1,6 +1,5 @@
-# https://github.com/jhump/protoreflect/commits/master 20220416 checked 20220417
```
bufdev, line 1: Update to `https://github.com/bufbuild/protocompile/commits/main DATE_OF_LAST_COMMIT checked DATE_WE_LAST_CHECKED_FOR_A_NEW_COMMIT`

private/buf/cmd/buf/command/build/build.go
```diff
@@ -139,6 +140,21 @@ func run(
...
 	}
+	imageRef, err := buffetch.NewImageRefParser(container.Logger()).GetImageRef(ctx, flags.Output)
+	if err != nil {
+		return fmt.Errorf("--%s: %v", outputFlagName, err)
+	}
+	var buildOptions []bufimagebuild.BuildOption
+	if flags.ExcludeSourceInfo {
+		buildOptions = append(buildOptions, bufimagebuild.WithExcludeSourceCodeInfo())
+	} else if !flags.AsFileDescriptorSet {
+		// if not a descriptor set, not necessary to try to match protoc, so include extra comments
```
bufdev, line 151: I'm not sure we want to lock ourselves to the SourceCodeInfos that protoc produces - there's no exact definition as to what SourceCodeInfos a FileDescriptorSet will produce, and us producing extra is arguably better than what protoc does. I would not change the SourceCodeInfos dependent on the state of `--as-file-descriptor-set`, instead I'd remove the `WithExtraSourceCodeInfo` option altogether and make having extra source code info the default.

private/buf/cmd/buf/command/build/build.go
```diff
@@ -139,6 +140,21 @@ func run(
...
+		buildOptions = append(buildOptions, bufimagebuild.WithExtraSourceCodeInfo())
+	}
+	if flags.AsFileDescriptorSet && imageRef.ImageEncoding() == buffetch.ImageEncodingBin {
+		// if generating binary format, emit the same bytes as protoc
```
bufdev, line 155: Effectively same comment - I don't think we need to lock ourselves long-term into emitting the exact same bytes as protoc. We fulfill the contract that protoc has by producing functionally-equivalent FileDescriptorSets, our definition of what these are is just as good.

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -310,10 +311,10 @@ func run(
...
 		ref,
 		flags.Config,
-		flags.Paths,        // we filter on files
-		flags.ExcludePaths, // we exclude these paths
-		false,              // input files must exist
-		false,              // we must include source info for generation
+		flags.Paths,                             // we filter on files
+		flags.ExcludePaths,                      // we exclude these paths
+		false,                                   // input files must exist
+		bufimagebuild.WithCanonicalByteOutput(), // include source info, match protoc byte-for-byte
```
bufdev, line 317: Regardless of the answers to the above, why do we need to do this for `buf generate`?

private/bufpkg/bufcheck/bufbreaking/bufbreaking_test.go
```diff
@@ -173,29 +173,29 @@ func TestRunBreakingFieldSameJSONName(t *testing.T) {
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 7, 18, 7, 35, "FIELD_SAME_JSON_NAME"),
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 8, 20, 8, 37, "FIELD_SAME_JSON_NAME"),
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 9, 3, 9, 27, "FIELD_SAME_JSON_NAME"),
-		bufanalysistesting.NewFileAnnotation(t, "1.proto", 10, 28, 10, 45, "FIELD_SAME_JSON_NAME"),
-		bufanalysistesting.NewFileAnnotation(t, "1.proto", 11, 27, 11, 44, "FIELD_SAME_JSON_NAME"),
+		bufanalysistesting.NewFileAnnotation(t, "1.proto", 10, 28, 10, 46, "FIELD_SAME_JSON_NAME"),
```
bufdev, line 176: I'm concerned that any of these changed. If anything, I'd expect that part of our test for if this protoparse-to-protocompile conversion worked would be that this file would have no diff. Can you explain here a bit?

private/bufpkg/bufcheck/bufbreaking/testdata/breaking_field_same_json_name/1.proto
```diff
@@ -7,11 +7,11 @@ message Two {
   int32 two = 2 [json_name = "bar"];
   int32 three = 3 [json_name = "baz"];
   repeated int32 four = 4;
-  repeated int32 five = 5 [json_name = "bar"];
-  repeated int32 six = 6 [json_name = "baz"];
+  repeated int32 five = 5 [json_name = "bar2"];
```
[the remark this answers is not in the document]
bufdev: I think the is the answer to the above question, let me know.

private/bufpkg/bufimage/bufimagebuild/bufimagebuild.go
```diff
@@ -52,6 +53,36 @@ type BuildOption func(*buildOptions)
...
 func WithExcludeSourceCodeInfo() BuildOption {
 	return func(buildOptions *buildOptions) {
-		buildOptions.excludeSourceCodeInfo = true
+		buildOptions.sourceInfoMode = protocompile.SourceInfoNone
+	}
+}
+
+// WithExtraSourceCodeInfo returns a BuildOption that causes source code info to
+// be generated with extra comments.
+func WithExtraSourceCodeInfo() BuildOption {
```
bufdev, line 62: See comment above, I don't think we should have this option.

private/bufpkg/bufimage/bufimagebuild/bufimagebuild.go
```diff
@@ -52,6 +53,36 @@ type BuildOption func(*buildOptions)
...
+	return func(buildOptions *buildOptions) {
+		buildOptions.sourceInfoMode = protocompile.SourceInfoExtraComments
+	}
+}
+
+// WithParallelism indicates the maximum level of parallelism to use when building
+// an image. If unspecified, the maximum parallelism will be automatically set to
+// min(runtime.NumCPU(), runtime.GOMAXPROCS(-1)).
+func WithParallelism(parallelism int) BuildOption {
```
bufdev, line 71: Is this different than before? All parallelism in bufbuild/buf should be controlled by `pkg/thread`, so whatever `pkg/thread` has set for parallelism should be what we use within `bufimagebuild`.

private/bufpkg/bufimage/bufimagebuild/bufimagebuild.go
```diff
@@ -52,6 +53,36 @@ type BuildOption func(*buildOptions)
...
+}
+
+// WithCanonicalByteOutput indicates that file descriptor protos in the built
+// image should be marshaled to the protobuf binary format using a "canonical"
+// format, that matches byte-for-byte the output of protoc. This should only
+// be used by callers that intend to write a binary descriptor set file. When
+// true, options in the returned descriptor protos are not available
+// programmatically because the data is stored as (pre-serialized) unrecognized
+// fields (so that marshaling to bytes is in the right format).
+func WithCanonicalByteOutput() BuildOption {
```
bufdev, line 84: See comment above, I don't think we should have this option, we should choose what the correct answer is (canonical or not-canonical) and always do that. Happy to discuss offline.

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -278,108 +207,63 @@ func getBuildResult(
...
+		fileDescriptor, ok := nameToFileDescriptor[rootRelFilePath]
 		if !ok {
 			return nil, fmt.Errorf("no FileDescriptor for rootRelFilePath: %q", rootRelFilePath)
 		}
-		sortedDescFileDescriptors = append(sortedDescFileDescriptors, descFileDescriptor)
-	}
-	return sortedDescFileDescriptors, nil
-}
-
-func getSyntaxUnspecifiedFilenamesFromBuildResults(
```
bufdev, line 343: Why where these removed? I'm likely just not looking carefully enough, but `bufimagebuild` needs to produce a full `Image`, which includes these extra things that `FileDescriptorSets` do not have.

private/buf/cmd/buf/command/build/build.go
jhump: > there's no exact definition as to what SourceCodeInfos a FileDescriptorSet will produce

It is documented in `descriptor.proto`. And `protoparse` wasn't really conforming to it (intentionally -- I originally added the extra comments before I implemented the AST, because I wanted a way to get access to as much of the original source as possible).

I'm fine removing it. But it seemed valuable for maximizing trust that a descriptor set produced by buf matches that of protoc.

private/buf/cmd/buf/command/build/build.go
jhump: This was basically the fix for https://github.com/jhump/protoreflect/issues/214

If you don't think we should do that, it's easy to remove. But I thought there was value in it.

CHANGELOG.md
jhump: @bufdev, you could "request changes" on the PR, and that will block merging. (Or some other reviewer could do that.)

private/bufpkg/bufcheck/bufbreaking/bufbreaking_test.go
jhump: I had to change JSON names for the files to actually compile. I added dumb number suffixes, which made some of the spans grow by one.

private/bufpkg/bufimage/bufimagebuild/bufimagebuild.go
jhump: I'll rework it. I originally added this before discovering the `thread` package and seeing how it was used before. That is in fact how we set the _default_ parallelism. The only caller of this is a test (that sets it to 1 for deterministic output). I can instead have the test set the parallelism via the `thread` package (and defer to restore to prior setting).

private/buf/cmd/buf/command/build/build.go
bufdev: Let's discuss tomorrow or in person this week

private/buf/cmd/buf/command/build/build.go
bufdev: Let's discuss tomorrow/in person this week

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -447,9 +331,10 @@ func getImageFilesRec(
 		unusedDependencyIndexes = make([]int32, 0, len(unusedDependencyFilenames))
 	}
 	var err error
-	for i, dependency := range descFileDescriptor.GetDependencies() {
+	for i := 0; i < fileDescriptor.Imports().Len(); i++ {
```
pkwarren, line 334: The old protoreflect would just return non-public, non-weak dependencies with `GetDependencies()`. Do we need to filter those out here in this loop?

jhump: This isn't right. The old protoreflect returns all dependencies. The other accessors for public and weak return a subset of this.
https://pkg.go.dev/github.com/jhump/protoreflect@v1.13.0/desc#FileDescriptor.GetDependencies

.github/dependabot.yml
```diff
@@ -14,7 +14,7 @@ updates:
       interval: "weekly"
     ignore:
       # We manually update this from master and lock the commit in make/buf/all.mk
```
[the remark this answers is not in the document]
jhump: fixed

private/bufpkg/bufimage/bufimagebuild/bufimagebuild.go
jhump: Removed this option.

CHANGELOG.md
jhump: Updated with more details and actual stats. Related: https://github.com/bufbuild/protocompile/pull/64

private/bufpkg/bufcheck/bufbreaking/testdata/breaking_field_same_json_name/1.proto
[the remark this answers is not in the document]
jhump: Yep.

jhump:
Per convo with @bufdev, I've updated this PR to (1) always use "standard" source code info (same level of comments as protoc) and (2) not bother trying to emit "canonical" bytes for descriptor sets. We can easily roll either these back in if we change our minds. For the first item (amount of comments), I had to update a test since the linter can no longer access a comment on the individual request or response type definition in a method.

Since we don't need ways to customize these aspects of output, I removed the relevant build options and reverted a lot of the method signature changes (since those changes are no longer necessary/useful).

cc @pkwarren

bufdev (approved):

CHANGELOG.md
```diff
@@ -12,8 +12,19 @@
 - Formatter uses a compact, single-line representation for array and message literals
   in option values that are sufficiently simple (single scalar element or field).
 - Include `LICENSE` file in the module on `buf push`.
-- New compiler that is XX% faster, uses XX% less memory, and addresses a few bugs
-  where Buf would accept proto sources that protoc would reject:
+- New compiler that is faster and uses less memory than the outgoing one.
```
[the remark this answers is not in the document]
jhump: Moved it up.

pkwarren (approved):

PR #1464, opened 2022-10-04
Use CuratedPlugin in Resource, instead of Plugin

oliversun09:
Since `ResourceService` is not yet implemented, this change won't break any API

https://linear.app/bufbuild/issue/BSR-601/implement-a-getresource-endpoint


paul-sachs (approved):

robbertvanginkel (approved):

robbertvanginkel:
Can you put the linear issue in the description too?

PR #1465, opened 2022-10-04
[BSR-510] List Audited Events service

unmultimedio:


proto/buf/alpha/audit/v1alpha1/service.proto
```diff
@@ -0,0 +1,43 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.audit.v1alpha1;
+
+import "buf/alpha/audit/v1alpha1/event.proto";
+
+// AuditService is the Audit service.
+service AuditService {
```
unmultimedio, line 22: Not 100% certain now that this and the events definitions should live outside of `registry` pkg. Sounds like they make sense within it.

proto/buf/alpha/audit/v1alpha1/service.proto
```diff
@@ -0,0 +1,48 @@
...
+  // You can reverse that order setting this value to true.
+  bool reverse = 3;
+  // The start timestamp to filter events from.
+  google.protobuf.Timestamp start = 4;
+  // The end timestamp to filter events to.
+  google.protobuf.Timestamp end = 5;
+}
+
+message ListAuditedEventsResponse {
+  // The list of audited in the current page.
```
stefanvanburen, line 43: drive-by minor nit :-)

```suggestion
  // The list of audited events in the current page.
```

njhale (approved):
Looks good!

proto/buf/alpha/audit/v1alpha1/service.proto
saquibmian: I don't think we could make this `v1` while it depends on `registry/v1alpha1`. Unless we break that dependency, their versioning is tied and may as well live in the same package.

saquibmian (approved):
I think all of the auditing stuff should move into `registry/v1alpha1`, as it's coupled with the BSR itself. We can do that move in a follow up PR though, I won't block on that.

proto/buf/alpha/audit/v1alpha1/service.proto
unmultimedio: https://linear.app/bufbuild/issue/BSR-666/reorganize-audit-code-and-pkgs

PR #1466, opened 2022-10-04
Add request metadata to audit logging events

njhale:
Add user_agent, ip, and trace_id fields to audit logging messages.

Part of [BSR-560](https://linear.app/bufbuild/issue/BSR-560/add-ip-and-user-agent-to-audited-events) and [BSR-633](https://linear.app/bufbuild/issue/BSR-632/add-trace-or-request-id-to-the-event-metadata)


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -34,6 +34,10 @@ message Actor {
   string id = 2;
   // Name of the actor who caused the audited event.
   string name = 3;
+  // UserAgent is the User-Agent header associated with the request that caused the audited event, if any.
+  string user_agent = 4;
+  // IP is the IP address associated with the request that caused the audited event, if any.
+  string ip = 5;
```
unmultimedio, line 40: 1. Not certain if it should live here or at the base `Event`. Currently we only have `ACTOR_TYPE_USER`, so it kinda makes sense here, but it'd be empty if a future actor is just `ACTOR_TYPE_SYSTEM` or `ACTOR_TYPE_PERIODIC_JOB` or something similar?
2. Before adding them, do we have access to both pieces of data in the BSR to add them to each log?

njhale: > 1. Not certain if it should live here or at the base Event. Currently we only have ACTOR_TYPE_USER, so it kinda makes sense here, but it'd be empty if a future actor is just ACTOR_TYPE_SYSTEM or ACTOR_TYPE_PERIODIC_JOB or something similar?

I was already on the fence about that. I leaned this way because it keeps the root event message namespace clean. I think putting it in Actor makes still sense -- even for all of the types you've listed. Do you have any preference? 

> 2. Before adding them, do we have access to both pieces of data in the BSR to add them to each log?

[yup](https://github.com/bufbuild/core/pull/6271), although I was thinking that this info would be completely optional.

unmultimedio: 1. No preference, these fields in `Actor` are fine.
2. Yeah, optional is fine. I'd wait for that PR to be almost ready to merge this one.

saquibmian: The interesting part is that some system operations can be triggered by user action, but the user is not the actor (e.g., web hook firing, if we ever decide to audit that). What if we keep `Actor` simple (id, name), but we move all this stuff to a `Metadata` message that contains additional information about who/what caused the event in the system and potentially why. Something like:
```protobuf
message Metadata {
  string actor_ip =1 ;
  string actor_user_agent = 2;
  string event_trace_id = 3;
}
```
The idea being that all of this is optional and a best-effort. What say all y'all?

unmultimedio: Don't have a strong opinion on `Metadata` vs inlining in `Actor`.

btw that trace id comes from https://linear.app/bufbuild/issue/BSR-632/add-trace-or-request-id-to-the-event-metadata, @njhale do you see that as something you could merge/tackle in the same ticket, being that all of that is injecting metadata?

njhale: > btw that event_trace_id comes from https://linear.app/bufbuild/issue/BSR-632/add-trace-or-request-id-to-the-event-metadata, @njhale do you see that as something you could merge/tackle in the same ticket, being that all of that is injecting metadata?

@unmultimedio SGTM

> The interesting part is that some system operations can be triggered by user action, but the user is not the actor (e.g., web hook firing, if we ever decide to audit that). What if we keep Actor simple (id, name), but we move all this stuff to a Metadata message that contains additional information about who/what caused the event in the system and potentially why. Something like:

@saquibmian -- In the case of webhooks, who would the actor be? Is it really possible for there to be dissonance between the IP, User-Agent, and Actor? I _think_ there are only two possibilities:
1. "Delegation": Actor = BSR System Account, where (IP, User-Agent) = (BSR System IP, BSR System User Agent)
2. "Impersonation": Actor = Webhook Owner, where (IP, User-Agent) = (BSR System IP, BSR System User Agent)

(2. In the case of 2, the BSR System Account _pretends_ to be the Webhook Owner by authenticating as them using special privileges; which is something that you can do in k8s actually)

which both seem like accurate and meaningful depictions of the event to me; i.e. associating IP and User-Agent with Actor still "works" semantically.

I _think_ the semantics fail if we can find an event type that won't have an IP and User-Agent associated with it, but it doesn't seem like that's possible for any event written as the direct results of an RPC; of course, that only holds if we don't have some async process write an event directly (that's more of a pattern question).

njhale: @saquibmian I ended up creating an `EventMetadata` message type and putting everything in there.

@unmultimedio I added the `trace_id` field as well.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -97,27 +107,29 @@ message Event {
...
+    PayloadRepositoryContributorRemoved repository_contributor_removed = 17;
+    PayloadRepositoryVisibilityChanged repository_visibility_changed = 18;
+    PayloadPluginCreated plugin_created = 19;
+    PayloadPluginDeleted plugin_deleted = 20;
+    PayloadUserCreated user_created = 21;
+    PayloadUserDeactivated user_deactivated = 22;
+    PayloadUserDeleted user_deleted = 23;
+    PayloadUserLoggedIn user_logged_in = 24;
+    PayloadUserLoggedOut user_logged_out = 25;
   }
```
njhale, line 133: Field renumbering should be okay since this is `v1alpha1` and not exposed via an RPC until BSR-510 is completed (see https://github.com/bufbuild/buf/pull/1465)

saquibmian (approved):

PR #1468, opened 2022-10-05
Surface `deprecated` option in protosource

seankimdev:
For the purpose of providing `deprecated` option to the [`DocService.GetPackageDocumentation`](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.DocService.GetPackageDocumentation), which will be used for rendering in the bufdoc, this PR surfaces them in the `protosource` for the following elements:
- File
- Message
- Field
- Enum
- EnumValue
- Service
- Method


private/pkg/protosource/protosource.go
```diff
@@ -346,6 +346,7 @@ type Message interface {
 
 	MessageSetWireFormat() bool
 	NoStandardDescriptorAccessor() bool
+	IsDeprecated() bool
```
bufdev, line 349: Rename to `Deprecated` to be consistent with the other `bool` functions

private/pkg/protosource/enum.go
```diff
@@ -30,13 +31,15 @@ func newEnum(
...
 	allowAlias bool,
+	deprecated bool,
 	allowAliasPath []int32,
 	parent Message,
 ) *enum {
 	return &enum{
 		namedDescriptor:           namedDescriptor,
 		optionExtensionDescriptor: optionExtensionDescriptor,
 		allowAlias:                allowAlias,
+		deprecated:                deprecated,
```
seankimdev, line 42: This is placed in between `allowalias` and `allowAliasPath` to keep the ordering pattern consistent with the existing ones, where all options are listed first, and then their paths or locations follow.

private/pkg/protosource/field.go
```diff
@@ -33,6 +33,7 @@ type field struct {
 	jsType         FieldOptionsJSType
 	cType          FieldOptionsCType
 	packed         *bool
+	deprecated     *bool
```
timostamm, line 36: Why is this one a pointer, but the others are not?

Note that it is possible to set `[deprecated=false]` in protobuf, and using a pointer would preserve this case. I don't think we should do that here, though. Descriptors are not an accurate representation of the protobuf source, and deprecated = false is not meaningful.

seankimdev: I agree that the `deprecated = false` is not meaningful for us here, hence no reason to use a pointer. I'll change it.

doriable (approved):
This looks reasonable to me!

timostamm (approved):

PR #1475, opened 2022-10-10
[BSR-542] Allow injecting write callback to `storagemem`

unmultimedio:
Allow callback for write in `storagemem`, so it's more easily testable.


unmultimedio:
Not needed anymore, this might be a dangerous options. Solved locally by writing a struct that contains the bucket that intercepts `Write` operations.

PR #1476, opened 2022-10-11
Allow for buf beta convert to be used without input

joshcarp:
v simple change

- Remove default input from "." in convert as it doesn't need to be run in a buf module


cyinma (approved):
lgtm, but we will need to update the docs [here](https://docs.buf.build/reference/inputs) later as this is not using current directory as its `input` by default now

PR #1483, opened 2022-10-12
BSR-578/Protobuf definition for updating metadata

oliversun09:
Add metadata fields to `UpdateOrganizationSettingsRequest` and `UpdateRepositorySettingsByNameRequest`, add rpc `UpdateUserSettings`, to allow updating metadata.

- The issue is actually BSR-578 (the number in branch name is wrong)


proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -205,6 +205,12 @@ message UpdateOrganizationSettingsRequest {
   PluginRole plugin_base_role = 3;
   // optional, no update to this base role will be made if this is unspecified.
   TemplateRole template_base_role = 4;
+  // optional, no update to description will be made if this is unspecified.
+  string description = 5;
```
cyinma, line 209: how will we distinguish update to empty and no update?

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -205,6 +205,12 @@ message UpdateOrganizationSettingsRequest {
...
+  // optional, no update to url will be made if this is unspecified.
+  string url = 6;
+  // optional, no update to verification_status will be made if this is unspecified.
+  VerificationStatus verification_status = 7;
```
cyinma, line 213: I think it will be better to make this a separated endpoint for server admin only, as we are not going to let user update their own verification status

oliversun09: You're right, should that be a public API then?

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -205,6 +206,10 @@ message UpdateOrganizationSettingsRequest {
   PluginRole plugin_base_role = 3;
   // optional, no update to this base role will be made if this is unspecified.
   TemplateRole template_base_role = 4;
+  // Changes are only made to values specified in field_mask.
+  google.protobuf.FieldMask field_mask = 5;
+  string description = 6;
+  string url = 7;
 }
```
oliversun09, line 213: Note the last two fields look inconsistent with the ones above, as `repository_base_role` and the other two fields are not controlled by the field mask. I think this is fine because `organization` message doesn't have these roles as fields.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -295,8 +296,11 @@ message GetRepositorySettingsResponse {
 message UpdateRepositorySettingsByNameRequest {
   string owner_name = 1;
   string repository_name = 2;
-  // Changes are only made to specified values. Unspecified ones are left unchanged.
   Visibility visibility = 3;
+  // Changes are only made to values specified in field_mask.
+  google.protobuf.FieldMask field_mask = 4;
```
oliversun09, line 301: `visibility` is not affected by field mask, so that the existing code won't break

robbertvanginkel (commented):
I know there was some talk about using fieldmask internally, but I don't think this is a good experience for these endpoints:

- Some of these endpoints already treat the 0/_UNSPECIFIED state as an indication not to update. There's some comments now about only some fields being in scope of the field mask, but the experience overall is a bit inconsistent.
- The experience for using field mask varies pretty wildly by language, and in go in particular there's a need for extra libraries to create/read field masks.
- We haven't used fieldmasks throughout the api, these would be the first cases.

Based on the above, we should go with field presence (proto3's scalar optionals) for these. I think that will be less confusing and more ergonomic. We can consider field masks when we do a next iteration of the API.

robbertvanginkel (approved):

PR #1486, opened 2022-10-14
bufmodule: test NewModuleForBucket

twilly:
`NewModuleForBucket` didn't have any coverage, so here's some I added as
I was working on other things.

### DIFF
--- a/private/bufpkg/bufmodule/module_test.go
+++ b/private/bufpkg/bufmodule/module_test.go
@@ -1,0 +1,124 @@
+// Copyright 2020-2022 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/module_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/bufbuild/buf/private/pkg/storage/storagemem"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func testNewModuleForBucket(
+	t *testing.T,
+	desc string,
+	files map[string][]byte,
+	isError bool,
```
doriable, line 32: Any chance we could have this take some contents to test for against the error (as opposed to just a `nil` check fro the error)?

twilly: You mean to check for a specific error value? The current use is for checking against any error returned because there's no error specific to `bufmodule`. I'm not intending to ascribe "a buflock error should be expected"[1] but to verify errors are handled / propagated.

Value checking would be more interesting if I were to try checking the error paths through `getFileContentForBucket`, but I didn't get to that. Coverage isn't 100% yet. 

[1] Incidentally the error is `failed to read lock file: unknown lock file versions "v0"`, but it's not captured in an exported variable nor is there an error checking function in the package. The test would have to be a substring check of the `.Error()`.

doriable: Yeah, that's fair, since the error we're getting back is somewhat arbitrary and unspecific to the `bufmodule` package, it may not make sense to do a substring check :) Looks good to me otherwise!

doriable (approved):

PR #1487, opened 2022-10-14
BSR-680 - Remove `Subject` from auditing login/logout payloads

saquibmian:


unmultimedio (approved):

njhale (commented):
@saquibmian before I approve, could you explain the motivation for this change? Is a `subject` just not useful info?

saquibmian:
@njhale Basically if we don’t have to log it, we’d rather not. It can contain information that’s subject to GDPR (like an email address).

njhale (approved):

Issue #1432, opened 2022-09-22
Promote `buf beta convert` to stable


joshcarp:
Currently this shouldn't be a problem because `buf convert` does not already exist so this just involves removing `beta`

bufdev:
List of questions to address about current UX before we promote this to stable:

- How do well-known types work: https://github.com/bufbuild/buf/issues/1434
- What does it mean to provide a source without using the `--type` flag? From looking at the code, it seems like the `--type` flag is required, but this isn't specified in the docs.
- `--input` is meant for the encoded input, but we use "input" generally to mean the source files or Image, so this is confusing. Should address.
- If we have `-o` for output, and we keep `--input`, then `--input` should be alised to `-i`.
- Confusion/overlapping use between the `<source>` argument and the `--type` flag. There are times that `--type` means that the `<source>` argument is not used. For example, I tried to see if `buf build -o - | buf beta convert --type google.protobuf.FileDescriptorSet` would just work to address #1434, but it doesn't, you get this error message: `Failure: if a source isn't provided, the type needs to be a fully qualified path that includes the module reference; failed to parse the type: "google.protobuf.FileDescriptorSet" is not a valid fully qualified path`. There should be a clear split between source and type. Perhaps there should only be one of these.
- Documentation for `--error-format` should probably be better and more specific - in what cases does `convert` result in a build? I know the answer (when you are building the source) but we should probably be more explicit. This may be a `buf`-wide problem.
- Why does `buf beta convert` not allow you to use Images, instead only allowing you to use sources? I would think I'd be able to do `buf beta convert <(buf build -o -)#format=bin --type foo.Bar` for example.

bufdev:
Also need better documentation when you do `buf help beta convert`.

bufdev:
Right now, we have an `inverseEncoding` function that says "if the input is binary, output JSON, and vice versa". If we promote this to stable, this behavior will prevent us from later easily expanding to other encodings - although not impossible, the default behavior may be weirdly defined. We should make sure that users have to somehow explicitly define the input's encoding and the output's encoding. We usually do this by either file extension or options on the command line. Examples:

- `buf convert --input tmp.bin` - this would be parsed as a binary file
- `buf convert --input tmp.json` - this would be parsed as a JSON file
- `buf convert --input -` - this would default to binary
- `buf convert --input -#format=json` - this would be parsed as JSON

joshcarp:
[...]

PR #1494, opened 2022-10-17
Add support for well known types for buf beta convert

joshcarp:
Adds ability to call `buf beta convert` without specifying an input when a well known type is specified;

```
$ buf beta convert --type=google.protobuf.Duration --from <(echo \"3600s\")#format=json
> some proto bytes

$ buf beta convert --type=google.protobuf.Duration --from duration.bin
> "3600s"
```

### Input is defaulted to "."
This means that by default calling `buf beta convert --type=google.protobuf.Duration` will look for `google.protobuf.Duration` in the local directory first before resolving to a baked in well known type

Without explicit setting of input to be "." no errors will be reported on. This is to stop from `buf beta convert` failing to work in any directory that might have a broken proto file in. This means that:
- `buf beta convert .  --type=google.protobuf.Duration` will fail if `broken.proto` exists, for example but
- `buf beta convert --type=google.protobuf.Duration` will pass because the error from parsing `"."` will be discarded because the user did not specifically request for it to be parsed.


private/buf/bufcli/bufcli.go
```diff
@@ -801,6 +807,29 @@ func NewImageForSource(
 	return bufimage.MergeImages(images...)
 }
 
+// WellKnownTypeImage returns an image for a well known type source file.
+func WellKnownTypeImage(wkpath string) (bufimage.Image, error) {
+	descriptors, err := protoparse.Parser{}.ParseFiles(wkpath)
```
jhump, line 812: `protoparse` is the old compiler. Ideally, we'd figure out a way to make this work through the `bufimagebuild` package, since that consolidates all knowledge of the compiler (and also implements existing support for auto-discovery of the well known imports).

private/buf/buffetch/internal/reader.go
```diff
@@ -361,6 +363,18 @@ func (r *reader) getProtoFileBucket(
...
+		bucket := storage.NopReadBucketCloser(datawkt.ReadBucket)
+		terminateFileProvider, err := getTerminateFileProviderForBucket(ctx, bucket, "", nil)
+		if err != nil {
+			return nil, err
+		}
+		bucketCloser, err := newReadBucketCloser(bucket, "", "")
+		if err != nil {
+			return nil, err
+		}
+		return newReadBucketCloserWithTerminateFiles(bucketCloser, terminateFileProvider), nil
```
ElliotMJackson, line 376: extract method for early return to improve readability.

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -214,3 +221,19 @@ func inverseEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEnc
...
+		// not a statically known type
+		return ""
+	}
+	filePath := msgType.Descriptor().ParentFile().Path()
+	if datawkt.Exists(filePath) {
+		// found it!
+		return filePath
+	}
+	// not a well-known type
+	return ""
```
ElliotMJackson, line 238:
```suggestion
	if !datawkt.Exists(filePath) {
		// not a well-known type
		return ""
	}
	// found it!
	return filePath
```

left align happy path

ElliotMJackson (changes requested):

ElliotMJackson (approved):

private/buf/buffetch/internal/reader.go
```diff
@@ -361,6 +363,9 @@ func (r *reader) getProtoFileBucket(
 	if !r.localEnabled {
 		return nil, NewReadLocalDisabledError()
 	}
+	if datawkt.Exists(protoFileRef.Path()) {
```
bufdev, line 366: This means that you will always prioritize the built-in well-known types over whatever is on disk. As a counterexample of how this could break down, assume I have an old, incompatible version of `google/protobuf/timestamp.proto` on disk at `./google/protobuf/timestamp.proto`, and I specify `buf beta convert google/protobuf/timestamp.proto`. With this code, I will always use the built-in version and not the one on disk, whereas the rest of our code prioritizes the one on disk over the builtins.

private/buf/buffetch/internal/reader.go
```diff
@@ -361,6 +363,9 @@ func (r *reader) getProtoFileBucket(
...
+		return wellKnownTypeBucket(ctx)
```
bufdev, line 367: Why would I return the entire bucket if I have a single file? Shouldn't `buf beta convert google/protobuf/timestamp.proto --type google.type.Duration` fail?

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -214,3 +221,19 @@ func inverseEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEnc
 		return 0, fmt.Errorf("unknown message encoding %v", encoding)
 	}
 }
+
+// wktToPath returns the import path of the proto file for a well known type.
+func wktToPath(fulltype string) string {
+	msgType, err := protoregistry.GlobalTypes.FindMessageByName(protoreflect.FullName(fulltype))
```
bufdev, line 227: I'm skeptical of this approach We should never be using the global registry for anything we do here. The global registry changes depending on what happens to be linked into the binary.

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -152,6 +156,9 @@ func run(
 	if err != nil {
 		return err
 	}
+	if wkpath := wktToPath(flags.Type); input == "" && wkpath != "" {
```
bufdev, line 159: I believe this breaks the precedence rules for WKTs in input vs the builtins.

Also see our style guide, specifically the section on descriptive names, i.e. `wellKnownPath` would be correct here if this code persisted.

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -214,3 +221,19 @@ func inverseEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEnc
 		return 0, fmt.Errorf("unknown message encoding %v", encoding)
 	}
 }
+
```
bufdev, line 224: I can't comment a few lines up, but I'm a little surprised the `inverseEncoding` function is persisting to stable, I would have thought we removed this for stable, worth a discussion.

bufdev (changes requested):

private/buf/buffetch/internal/reader.go
jhump: The compiler's source accessor has the logic for falling back when needed. So maybe the change needed is in the file provider for OS, below, to simply _allow_ "file not found" issues for paths that exist in `datawkt`. WDYT?

private/buf/cmd/buf/command/beta/convert/convert.go
jhump: I had suggested this as an alternative to a hand-written hard-coded map of type names to file names. The limitation you mention is easily worked around using underscore imports.

But another option would be to have a code generator (one that is run alongside `storage-to-go`) generate that hard-coded index by actually compiling the files in `datawkt` first and examining all of the types therein. That way it is not hand-coded and will never be inaccurate/stale.

private/buf/cmd/buf/command/beta/convert/convert.go
joshcarp: https://github.com/bufbuild/buf/issues/1432#issuecomment-1255555035

input encoding and output encoding can be explicitly set. inverse encoding only gets called if output doesn't have a `-#format` option (or the format can't be inferred by the filename)

private/buf/cmd/buf/command/beta/convert/convert.go
joshcarp: If the input is unspecified it will try to resolve the well known type, if not it will use the input argument. 

This means that calls like previously mentioned `buf beta convert google/protobuf/timestamp.proto --type google.type.Duration` will fail because `google/protobuf/timestamp.proto` doesn't contain `google.type.Duration`

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -164,8 +167,30 @@ func run(
...
+	if container.NumArgs() == 0 {
+		if inputErr != nil {
+			resolveWellKnownType = true
+		}
+		if image != nil {
+			_, filterErr := bufimageutil.ImageFilteredByTypes(image, flags.Type)
+			if errors.Is(filterErr, bufimageutil.ErrImageFilterTypeNotFound) {
+				resolveWellKnownType = true
+			}
+		}
```
jhump, line 181: This seems odd. So if they provide inputs to identify the source of the well-known type, but they give the wrong inputs, we just ignore them and use the well-known type data? Seems like we should only fall back to the well-known type data when (1) no input files are named or (2) they name a file that is _in_ the well-known type data and _not_ found locally. Wouldn't both of those conditions result in `inputErr != nil` above?

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	image, fileAnnotations, err := bufimagebuild.NewBuilder(container.Logger()).Build(
+		ctx, moduleFileSet,
+		bufimagebuild.WithExcludeSourceCodeInfo(),
+	)
+	if err != nil {
+		return nil, err
+	}
+	if len(fileAnnotations) > 0 {
+		// stderr since we do output to stdouo
```
jhump, line 178:
```suggestion
		// stderr since we do output to stdout
```

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
```diff
@@ -0,0 +1,321 @@
...
+		if err := bufanalysis.PrintFileAnnotations(
+			container.Stderr(),
+			fileAnnotations,
+			"text",
+		); err != nil {
+			return nil, err
+		}
+		return nil, app.NewError(100, "")
```
jhump, line 186: Is this right? Mysterious magic numbers and empty string args aren't good for readability. Let's at least add a comment if this is intentional.

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	return protosource.NewFilesUnstable(ctx, bufimageutil.NewInputFiles(image.Files())...)
+}
+
+func getGolangFileData(
```
jhump, line 191: This seems like an ideal use for a template. I think that would be much more readable.

private/gen/data/datawkt/datawkt.gen.go
```diff
@@ -64,6 +64,81 @@ var (
...
 		},
 	}
+
+	messageNameToFilePath = map[string]string{
```
jhump, line 68: QQ: would this have been easier to just emit these to a separate generated file from the blob above? I naively would have thought we'd leave the existing generator alone for the data blob and just add another generated file...

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
joshcarp: Personally I don't like text templates too much. As soon as there's any sort of control flow that needs to happen it becomes too complicated for me. 

I've changed all the `buffer.WriteString(` for an anonymous function `p(` which is similar to how it's done with other go code generators i've seen and it makes it a lot more readable than `buffer.WriteString`

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+	return pathToData, nil
+}
+
+func getProtosourceFiles(
+	ctx context.Context,
+	container appflag.Container,
+	bucket storage.ReadBucket,
+) ([]protosource.File, error) {
+	// TODO: why is this not working? this is the path that should be used, not NewModuleForBucket
```
joshcarp, line 150: I am aware that this todo is here but I do not know what the difference between `bufmodule.NewModuleForBucket` and `bufmodulebuild.NewModuleBucketBuilder` is

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
jhump: > As soon as there's any sort of control flow that needs to happen it becomes too complicated for me.

I agree 100%... which is why this is the ideal use for one since it's just emitting a hard-coded map 😝 

(Not a blocking comment... just my 2 cents on readability for this function.)

private/buf/cmd/buf/testdata/convert/bin_json/convert_test.go
```diff
@@ -0,0 +1,200 @@
...
+	"strings"
+	"testing"
+)
+
+// Unfortunately os.Chdir is very flaky in tests and cannot be done for only one go routine
+// Putting a test file here was the best option I could come up with.
+func TestConvertDir(t *testing.T) {
+	cmd := func(use string) *appcmd.Command { return buf.NewRootCommand("buf") }
+	t.Run("default-input", func(t *testing.T) {
+		appcmdtesting.RunCommandExitCodeStdout(
```
bufdev, line 16: You can do i.e. `buf convert testdata/convert/bin_json ...` correct? If so, if you move this test up a few levels, convert to this. If not, talk to me about why so (offline, moving off this PR after this round of comments)

private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
bufdev: For context, this is mostly copy/pasted from `storage-go-data`, didn't want to spend time reinventing the wheel.

bufdev (approved):
Approving assuming comments are handled - consider me not blocking on this PR anymore

private/buf/bufcli/bufcli.go
```diff
@@ -801,6 +805,30 @@ func NewImageForSource(
...
+		sourceConfig.Build,
+	)
+	if err != nil {
+		return nil, err
+	}
+	image, _, err := bufimagebuild.NewBuilder(logger).Build(ctx, bufmodule.NewModuleFileSet(module, nil))
+	if err != nil {
+		return nil, err
+	}
+	return bufimageutil.ImageFilteredByTypes(image, wellKnownType)
```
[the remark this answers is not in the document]
joshcarp:
```
// ImageFilteredByTypes returns a minimal image containing only the descriptors
// required to define those types. The resulting contains only files in which
// those descriptors and their transitive closure of required descriptors, with
// each file only contains the minimal required types and imports.
```
so yes

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -148,11 +151,11 @@ func run(
 	if err := bufcli.ValidateErrorFormatFlag(flags.ErrorFormat, errorFormatFlagName); err != nil {
 		return err
 	}
-	input, err := bufcli.GetInputValue(container, flags.InputHashtag, "")
+	input, err := bufcli.GetInputValue(container, flags.InputHashtag, ".")
```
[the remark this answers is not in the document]
joshcarp: "." will default to the current input directory. i.e. if you had a `buf.yaml` in your current directory you could run `buf convert --in=foo.json --out=foo.bin`

private/buf/cmd/buf/command/beta/convert/convert.go
```diff
@@ -164,8 +167,30 @@ func run(
 		false, // externalDirOrFilePathsAllowNotExist
 		false, // excludeSourceCodeInfo
 	)
-	if err != nil {
-		return err
+	var resolveWellKnownType bool
+	// only resolve wkts if input was not set.
```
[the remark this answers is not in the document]
joshcarp: correct, that's handled at a much lower level by the compilers source accessor which has access to all of the wkts

PR #1499, opened 2022-10-17
Manifest handling.

twilly:
A manifest is a file that lists files and their hash digests. The canonical form is `<hash type>:<digest>[sp][sp]<path>\n` lexicographically ordered by path.

This package implements encoding manifests and accessing their content by path or digest. Manifests can be created from a bucket where the bucket is walked and a manifest is created for every file contained within.


private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+	"context"
+	"encoding"
+	"encoding/hex"
+	"errors"
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+
+	"github.com/bufbuild/buf/private/pkg/storage"
```
pkwarren, line 37: If the manifest package is meant to be small / self contained, it might be good to not pull in other dependencies if we can avoid it. Perhaps instead we could add methods to the storage package to create a manifest from a bucket?

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+		if err != nil {
+			return nil, multierr.Append(
+				newManifestError(lineno, "invalid hash"),
+				err,
+			)
+		}
+		if digest.Type() != shake256Name {
+			return nil, newManifestError(lineno, "unknown hash")
+		}
+		if len(digest.Bytes()) != 64 {
```
pkwarren, line 166: nit: might be good to make this a constant too (if it is meant to live alongside the expected digest name.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+func (m *Manifest) addDigest(path string, digest *Digest) {
+	m.paths[path] = digest
+	m.digests[digest.String()] = path
```
pkwarren, line 205: What happens if two paths have the same digest? It might also be good to add a comment to the `String()` method above that it is used as a hash key so shouldn't be changed (sometimes `String()` might be used just for debugging).

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+// AddContent adds a manifest entry for path by its content.
+func (m *Manifest) AddContent(path string, content io.Reader) {
+	m.hash.Reset()
+	// sha3.ShakeHash never errors, reading or writing. These checks are to
+	// satisfy linting, possibly exploding if some breaking behavior change
+	// happens.
```
pkwarren, line 213: I think we might just want to have this return an error (even if unlikely). That is probably preferred to a panic coming from lower level code.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+
+const shake256Name = "shake256"
+
+type Digest struct {
+	dtype  string
+	digest []byte
+	hexstr string
+}
+
+func NewDigestFromBytes(dtype string, digest []byte) *Digest {
```
pkwarren, line 52: Since the other `New*` methods return a possible error, I wonder if we should do the same and do basic validation here (dtype/digest non-empty?)

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+// AddContent adds a manifest entry for path by its content.
+func (m *Manifest) AddContent(path string, content io.Reader) {
+	m.hash.Reset()
+	// sha3.ShakeHash never errors, reading or writing. These checks are to
+	// satisfy linting, possibly exploding if some breaking behavior change
+	// happens.
+	if _, err := io.Copy(m.hash, content); err != nil {
+		panic(err)
+	}
+	digest := make([]byte, 64)
```
pkwarren, line 217: We could re-use a constant here.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+var _ encoding.TextMarshaler = (*Manifest)(nil)
+var _ encoding.TextUnmarshaler = (*Manifest)(nil)
+
+// NewManifest creates an empty manifest.
+func NewManifest() *Manifest {
+	return &Manifest{
+		paths:   make(map[string]*Digest),
+		digests: make(map[string]string),
```
pkwarren, line 125: nit: Maybe name `pathToDigest` and `digestToPath`?

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	digest := make([]byte, 64)
+	if _, err := hash.Read(digest); err != nil {
+		panic(err)
+	}
+	return NewDigestFromBytes("shake256", digest)
+}
+
+func TestRoundTripManifest(t *testing.T) {
+	// read a manifest using the unmarshaling method
```
pkwarren, line 62: nit: Should use `t.Parallel()` on all tests.

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,194 @@
...
+	retContent, err := manifest.MarshalText()
+	require.NoError(t, err)
+	assert.Equal(t, expect, string(retContent))
+}
+
+func testInvalidManifest(
+	t *testing.T,
+	desc string,
+	line string,
+) {
```
pkwarren, line 98: Could add `t.Helper()` here?

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+
+// NewManifestFromReader builds a manifest an encoded manifest reader.
+func NewManifestFromReader(manifest io.Reader) (*Manifest, error) {
+	m := NewManifest()
+	scanner := bufio.NewScanner(manifest)
+	scanner.Split(splitManifest)
+	lineno := 0
+	for scanner.Scan() {
+		lineno++
+		hash, path, found := strings.Cut(scanner.Text(), "  ")
```
pkwarren, line 152: What if the path contains multiple spaces?

pkwarren (commented):
Really clean - nice work.

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	return NewDigestFromBytes("shake256", digest)
+}
+
+func TestRoundTripManifest(t *testing.T) {
+	// read a manifest using the unmarshaling method
+	null := mkdigest([]byte{})
+	var manifestBuilder bytes.Buffer
+	for i := 0; i < 2; i++ {
+		fmt.Fprintf(&manifestBuilder, "%s  null%d\n", null, i)
```
unmultimedio, line 66: iiuc this test, this will result in this content:

```
  null0
  null1
```

If so I have 2 Qs:
- It has 2 spaces, does the `strings.Cut` result in `hash:""`, `path:" null0"`?
- The hash doesn't contain `:`, why is not triggering an error?

private/pkg/manifest/manifest.go
unmultimedio: [Related](https://github.com/bufbuild/buf/pull/1499/files#diff-536233f971a641a3a3fc1c6fb13958674f984e202f8dc0cf7e07c0f28ff741eaR18), curious why is it defined with 2 spaces? `<hash type>:<digest>[SP][SP]<path>[LF]`, also curious, why not have `path` first? since it's easier to grok sorted by path instead of by digest

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+
+// NewManifestFromBucket creates a manifest from a bucket.
+func NewManifestFromBucket(
+	ctx context.Context,
+	bucket storage.ReadBucket,
+) (*Manifest, error) {
+	m := NewManifest()
+	err := bucket.Walk(ctx, "", func(info storage.ObjectInfo) error {
+		path := info.Path()
+		obj, err := bucket.Get(ctx, path)
```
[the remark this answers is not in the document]
unmultimedio: And related, curious how is useful to get a Manifest from a whole bucket?

private/pkg/manifest/manifest.go
unmultimedio: Or even the switching its names, `paths` map doesn't hold paths, holds digests, accessed by it's path haha.

private/pkg/manifest/manifest.go
twilly: To be perfectly honest, I wasn't sure where to put it! It makes sense to be on its own than as a component in the storage/some-other module. There's some overlap with `Digest` and it's soon-to-exist definition in `proto/buf/alpha/module/v1alpha1/module.proto`, which likely will break that out into it's own component.

All that said, manifest construction from a storage bucket does make sense to move into the storage bucket to weaken the dependency. I'll give that a go.

private/pkg/manifest/manifest.go
twilly: I can make it a constant. Really there's a distinction of a hasher and a hash constructor that defines such parameters, but I don't want to boil that ocean until we really want more than one hash type. :)

private/pkg/manifest/manifest.go
twilly: You're right, there should be a list of paths for a digest.

I'll make a note on the Digest type about `String()`'s significance. It's a good call-out and it's why `addDigest` takes a `Digest` and not a `Stringer`. :)

private/pkg/manifest/manifest.go
twilly: This one is a bit frustrating because the hashing should never error. You'll see this don't-check-error even in Go's crypto libraries (e.g. [sha2](https://cs.opensource.google/go/go/+/refs/tags/go1.19.2:src/crypto/sha256/sha256.go;l=263-285), [sha3](https://cs.opensource.google/go/x/crypto/+/refs/tags/v0.1.0:sha3/hashes.go;l=67-97)).

However, keeping content in memory sucks too. Propagating any read error from `content` is useful.

private/pkg/manifest/manifest.go
twilly: This would be the wrong place to check for supported types (`Digest` doesn't care what you stick in it), but I can add the check in `Manifest.addDigest`.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+func NewManifestFromReader(manifest io.Reader) (*Manifest, error) {
+	m := NewManifest()
+	scanner := bufio.NewScanner(manifest)
+	scanner.Split(splitManifest)
+	lineno := 0
+	for scanner.Scan() {
+		lineno++
+		hash, path, found := strings.Cut(scanner.Text(), "  ")
+		if !found {
+			return nil, newManifestError(lineno, "invalid entry")
```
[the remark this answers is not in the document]
twilly: Also I don't want to process all lines in a manifest as the goal is not trying to lint or be comprehensive but rather be as restrictive in what's being accepted today so we can loosen up behavior later. Any manifest considered valid now requires to be supported forever.

I'm not sure scanning the whole manifest to find all, possibly cascading, errors is worthwhile.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+			return nil, multierr.Append(
+				newManifestError(lineno, "invalid hash"),
+				err,
+			)
+		}
+		if digest.Type() != shake256Name {
+			return nil, newManifestError(lineno, "unknown hash")
+		}
+		if len(digest.Bytes()) != 64 {
+			return nil, newManifestError(lineno, "short digest")
```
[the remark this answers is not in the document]
twilly: It's not a format string, but the nit is welcome. I'll report the expectation in the error message.

private/pkg/manifest/manifest.go
twilly: It's all about perspective! I'll make the names more approachable. :)

private/pkg/manifest/manifest_test.go
twilly: Add all the boiler plate!

I decided to see how that affected this set of tests. 7.2% faster and saving 9 microseconds on my machine. 🐰

private/pkg/manifest/manifest.go
twilly: > What if the path contains multiple spaces?

`strings.Cut` splits the string at the first occurrence of the separator. `path` can contain as many spaces as needed. It cannot, however, contain a line feed. This is the same restriction Go has for modules and it's a fair restriction in general (so many tools break with files containing newlines). That said, is this something we should support? Do we expect or want to support files with line feeds in them?

> [Related](https://github.com/bufbuild/buf/pull/1499/files#diff-536233f971a641a3a3fc1c6fb13958674f984e202f8dc0cf7e07c0f28ff741eaR18), curious why is it defined with 2 spaces? `<hash type>:<digest>[SP][SP]<path>[LF]`, also curious, why not have `path` first? since it's easier to grok sorted by path when paths are first

*Tradition*. Mostly.

Checksum tooling has long printed the hash before the filename. `cksum` of the ancient POSIX standard uses `<hash>[TAB][<bytes>[TAB]<path>[LF]`. `md5sum` and its family of tools of the GNU world uses `<hash>[SP]<mode><path>[LF]`. And this double space format is what Go uses for modules (and seems inspired by `md5sum` except text vs binary mode is crazy talk, so just space 🌟 ). We can choose any format we want and using the Go format works and looks fine to read.

Technically there's good reasons to put hashes first. First, digests are always the same size. A constant character column is one more way you could parse this manifest. Or to cut by `[SP][SP]` is just going to work (these are illegal characters). `cut -d ' ' -f 3- < manifest` will work and so will `sed 's/^[^ ]* [ \*]//'`, which will work for `md5sum`, Go, and these manifests alike.

Second, there's no path escaping. Placing the path after the hash means you have a fixed set of characters to expect. No need to escape for spaces. The only illegal character is the newline.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,269 @@
...
+func splitManifest(data []byte, atEOF bool) (int, []byte, error) {
+	// Return a line without LF.
+	if i := bytes.IndexByte(data, '\n'); i >= 0 {
+		return i + 1, data[0:i], nil
+	}
+
+	// EOF occurred with a partial line.
+	if atEOF && len(data) != 0 {
+		return 0, nil, errPartial
+	}
```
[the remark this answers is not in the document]
twilly: Yep. IMHO, for launching it's best to be as restrictive as possible. It will suck to parse slightly variant manifests down the road.

private/pkg/manifest/manifest.go
unmultimedio: If no validations are done here, and we'll always succeed returning a Digest, then I suggest returning a value.
```suggestion
func NewDigestFromBytes(dtype string, digest []byte) Digest {
```

private/pkg/manifest/manifest.go
unmultimedio: Thank you for the explanation! It makes total sense, mostly because same-size digests... beauty of monospace fonts! Just one remaining q:

> And this double space format is what Go uses for modules

Is this in `go.mod`, `go.sum` or other file? I see those files use single space `<path>[SP]<version or hash>`

private/pkg/manifest/manifest_test.go
unmultimedio: Nevermind, I'm being silly, `mkdigest([]byte{})` will generate a valid hash.

twilly: It'll result in a manifest with `shake256:xxxx    null<n>` because that's what `Digest.String` returns, which is what `null` is. It is a tad hard to read in a test though. I'll make it a little more explicit and less short-handy.

> It has 2 spaces, does the strings.Cut result in hash:"", path:" null0"?

Two spaces and strings.Cut takes that as the separator. It'll result in path as "null0" (without spaces). Same goes with why ":" is contained within but you cannot easily see it here.

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
unmultimedio: Then a nit haha
```suggestion
	// Force final empty line at the end.
	if atEOF && len(data) != 0 {
		return 0, nil, errNoFinalEmptyLine
	}
```

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
unmultimedio: > as the goal is not trying to lint or be comprehensive

I think the CLI is used as editor's linter for `*.proto` files? may be worth to leave the door open to lint manifests as well?

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,194 @@
...
+
+func TestEmptyManifest(t *testing.T) {
+	content, err := NewManifest().MarshalText()
+	assert.NoError(t, err)
+	assert.Equal(t, 0, len(content))
+}
+
+func TestAddContent(t *testing.T) {
+	manifest := NewManifest()
+	manifest.AddContent("null", bytes.NewReader(nil))
```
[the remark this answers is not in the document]
twilly: "path" is confusing as well because that's what it literally is called in the code. Trying to read with "path is path" isn't great. I went with "null" because it contains nothing. Would it help if it were more abstract, like "foo?"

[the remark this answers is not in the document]
unmultimedio: sure, or
```suggestion
	manifest.AddContent("my/foo/path", bytes.NewReader(nil))
```

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,194 @@
...
+		"partial record",
+		fmt.Sprintf("%s  null", mkdigest(nil)),
+	)
+}
+
+func TestBrokenRead(t *testing.T) {
+	expected := errors.New("testing error")
+	_, err := NewManifestFromReader(iotest.ErrReader(expected))
+	assert.Error(t, err)
+	assert.Equal(t, expected, err)
```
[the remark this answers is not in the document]
twilly: It's not technically equivalent but functionally is better for the contact. A+, would ErrorsIs again.

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
twilly: > If you're passing a storagemem in here, don't think you can do .Get inside of a .Walk, will cause race issues. You'll need to first grab all the paths in .Walk, and then loop to get each content.

Wow. That seems less than ideal. Maybe I'll fix that because that's not a constraint that's documented and what I think I'm doing isn't crazy. It's pretty common to list a directory and then do things like stat and read them. Walk + Get probably shouldn't interfere with each other.

> And related, curious how is useful to get a Manifest from a whole bucket?

Buckets are created for modules, for the proto files, config, docs, and license files. This construction is to be used when pushing as the push can build a manifest for a module bucket. That said, this will probably move the storage package.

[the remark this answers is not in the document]
unmultimedio: Yeah, they both use the same `RLock` ([Walk](https://github.com/bufbuild/buf/blob/f1becc1a99c5f8c8b1ea8013f9aba83885a16195/private/pkg/storage/storagemem/bucket.go#L61), [Get](https://github.com/bufbuild/buf/blob/f1becc1a99c5f8c8b1ea8013f9aba83885a16195/private/pkg/storage/storagemem/bucket.go#L44))

private/pkg/manifest/manifest.go
twilly: A small update: the dependency on storage comes from `NewManifestFromBucket` but it cannot live in the `storage` package either. Using `NewManifestFromBucket` implies using a storage implementation, like `storagemem`, and these packages depend on `storage`. Hence changing this dependency from `module -> storage` creates a cycle of `storage -> storagemem -> storage`.

I could create another package, perhaps as `private/pkg/manifest/storage`, that would import `manifest` and `storage`. This would break the direct dependencies on each other. But is it worth it? I don't know.

private/pkg/manifest/manifest.go
twilly: Doh, I missed this in the commits I just pushed. Hold on and I'll make this change (ret by value works for me).

private/pkg/manifest/manifest.go
twilly: It's in the `dirhash` "h1" hashing for go modules. https://pkg.go.dev/golang.org/x/mod@v0.5.1/sumdb/dirhash#Hash1

private/pkg/manifest/manifest.go
twilly: On second pass, returning a pointer makes for better intuitive use with the pointer receivers. Lines like this will break:

```go
fmt.Printf("%s", digest)
```

Stringing digests would need to be passed as `&digest`.

There's been some discussion on this topic in the general style guide (pointer all the things), so I'm going to stick to returning and using pointers for now.

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
twilly: I think we can add that as a feature later because manifests don't exist at the CLI side except for in memory or in private areas, like the cache.

twilly:
Updated with noted commits. Looks like the CI is broken with the new security release of Git, but I'll leave this here before rebasing so you two won't lose your threads.

pkwarren (approved):

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,289 @@
...
+	return m, nil
+}
+
+func (m *Manifest) addDigest(path string, digest *Digest) error {
+	if digest.Type() != shake256Name {
+		return fmt.Errorf("unsupported hash: %s", digest.Type())
+	}
+	if n := len(digest.Bytes()); n != shake256Length {
+		return fmt.Errorf("short digest: %d of %d bytes", n, shake256Length)
```
unmultimedio, line 218: Nit related to [this previous comment](https://github.com/bufbuild/buf/pull/1499#discussion_r999582015), it's erroring with `"short"` but the condition is `!=`, I'd use `different|unexpected`.

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
unmultimedio: I think while the `RLock` is there, you should first get the list of paths, and then outside of the `Walk` fetch its contents.

private/pkg/manifest/manifest_test.go
```diff
@@ -0,0 +1,219 @@
...
+func TestPaths(t *testing.T) {
+	t.Parallel()
+	m := NewManifest()
+	err := m.AddContent("path/one", bytes.NewReader(nil))
+	require.NoError(t, err)
+	err = m.AddContent("path/two", bytes.NewReader(nil))
+	require.NoError(t, err)
+	paths, ok := m.Paths(mkdigest(nil))
+	assert.True(t, ok)
+	assert.ElementsMatch(t, []string{"path/one", "path/two"}, paths)
```
unmultimedio, line 202: Nit: Add an assertion making sure they're sorted? or does the `ElementsMath` checks order too? If so, may be add them _unsorted_?

private/pkg/manifest/manifest.go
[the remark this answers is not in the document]
twilly: Turns out the RLock is fine. `Walk`, `Get`, and `Stat` takes read locks and `Put`, `Delete`, and `DeleteAll` take write locks. So it's fine to call `Get` in `Walk`'s callback. There is an opportunity to make `Put` and others work too, though I'll defer that work for now.

private/pkg/manifest/manifest_test.go
twilly: `ElementsMatch` is an unordered, set equality assertion: 

> ElementsMatch asserts that the specified listA(array, slice...) is equal to specified listB(array, slice...) ignoring the order of the elements. If there are duplicate elements, the number of appearances of each of them in both lists should match.

Order is not guaranteed by `Paths` since it was never said to be, but I'll make that explicit in its docs.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,289 @@
...
+)
+
+var errNoFinalNewline error = errors.New("partial record: missing newline")
+
+const (
+	shake256Name   = "shake256"
+	shake256Length = 64
+)
+
+type Digest struct {
```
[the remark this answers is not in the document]
twilly: Surprising. [TCN-623](https://linear.app/bufbuild/issue/TCN-623)

twilly:
A required rebase means we have a new PR for this change: #1528

PR #1500, opened 2022-10-18
Change field option packed type to optional

seankimdev:
- `packed` field was recently added for the internal use of surfacing the option in bufdoc. It has not been implemented yet, therefore does not break anything.
-  Set vs. unset matters for `packed`, because it has different default behaviours of `false` in proto2, and `true` in proto3. This change allows `packed` to be treated as `*bool` in Go, and a `boolean | undefined` in TS, so a `nil` can be communicated to the frontend as `undefined`, indicating that the option hasn't been specified in the source.


timostamm (approved):
Aptly summarized. LGTM.

PR #1502, opened 2022-10-18
Abstract out storage-go-data logic

bufdev:
This actually ended up being irrelevant to what I was doing but it's a cleanup nonetheless.


bufdev:
Actually never mind, I think this might be too much abstraction for this, closing.

PR #1503, opened 2022-10-18
rename repository commit license SPDX ID

cyinma:
fixes BSR-667


pkwarren (approved):
Change looks ok to me assuming nothing is relying on the persisted JSON representation of this message.

PR #1504, opened 2022-10-18
Generate MessageFilePath and EnumFilePath functions for the WKTs

bufdev:
Needs work and has a TODO, and the `stringPair` abstraction was too much, but @joshcarp you can take it from here, this would work as-is.


private/bufpkg/bufwkt/cmd/wkt-go-data/main.go
```diff
@@ -0,0 +1,362 @@
...
+		}
+		_, _ = buffer.WriteString(`},
+`)
+	}
+
+	_, _ = buffer.WriteString(`}
+
+	messageNameToFilePath = map[string]string{
+`)
+	for _, messageFullNameToFilePair := range messageFullNameToFile {
```
joshcarp, line 288: would you be against all of the `stringPair` stuff and passing around the maps, then sorting the maps when needed;

```go
func sortedKeys[T constraints.Ordered](m map[T]any) []T {
	ret := make([]T, 0, len(m))
	for _, v := range m {
		ret = append(ret, v)
	}
	sort.Slice(ret, func(i, j int) bool {
		return ret[i] > ret[j]
	})
	return ret
}


... 
for _, key := range sortedKeys(messageFullNameToFileMap){
  messageFullNameToFilePair := messageFullNameToFileMap[key]

...
```

joshcarp: has the benefit of not needing to populate `stringPair`

bufdev: yea i regretted the `stringPair` once I did it, adjust as you want

joshcarp:
Gonna open up a PR against `feat/convert-wkt` as that's where this is needed

joshcarp:
merged in #1494 so closing

PR #1506, opened 2022-10-19
avoid panic in OptionExtensionDescriptor

jhump:
I was looking through this code, just to get acquainted with it so I am prepared to talk to @seankimdev about potential upcoming changes, when I noticed this. It was a pretty easy fix, so I figured I'd go ahead and make the change.

Sadly, there are no tests in this package. I didn't add the first test because the process to create a `protosource.Message` appears non-trivial due to several layers between the underlying descriptor and the `protosource` layer. So I simply verified that no existing tests were broken by it. But this new check is the same as [the one that causes the panic](https://github.com/protocolbuffers/protobuf-go/blob/b0a944684d78a6b7c1f2bb6864fa910385d5b97a/internal/impl/message_reflect.go#L449-L452).


robbertvanginkel (approved):
Looking through some internal code I don't see any usage of this to validate this against either, but on face value and the code you linked this does seem reasonable!

PR #1514, opened 2022-10-19
BSR-689/Add license url to curated plugin

oliversun09:
Add `license_url` to curated plugin.

In terms of relationship between `license_url` and `spdx_license_id` (i.e. if one is set whether the other should be set as well), I'm currently going with no binding between the two,  making this an optional field. 
My reasons: 
1. A user might not always have the url to license at hand, or it might not be a public repo (actually, can someone confirm if that's the case?). 
2. A user's license might not fall under any of the types in SPDX. 
However, suggestions are welcome!


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -316,6 +316,7 @@ func run(
 		nextRevision,
 		outputLanguages,
 		pluginConfig.SPDXLicenseID,
+		pluginConfig.LicenseURL,
```
oliversun09, line 319: I did not validate `pluginConfig.LicenseURL` because `pluginConfig.SourceURL` is also not validated

private/bufpkg/bufplugin/bufpluginconfig/testdata/success/go/buf.plugin.yaml
```diff
@@ -18,3 +18,4 @@ registry:
   opts:
     - separate_package=true
 spdx_license_id: Apache-2.0
+license_url: https://github.com/grpc/grpc-go/blob/master/LICENSE
```
oliversun09, line 21: I only added `license_url` to this file, but not [`npm/buf.plugin.yaml`](https://github.com/bufbuild/buf/blob/main/private/bufpkg/bufplugin/bufpluginconfig/testdata/success/npm/buf.plugin.yaml) because A. I couldn't find the url to the license (though I could just use any url) B. It's an optional field and missing this field is still valid.

pkwarren: The vast majority of the projects will use a standard license (where we won't need a custom URL and knowing the SPDX license id will allow us to build a URL). In that case, we'd only need to specify a license URL if using a non-standard license.

Do we want to consider updating the spdx-go-data command to surface additional attributes from the JSON file: https://github.com/spdx/license-list-data/blob/36873db191f0ce8b6e54839738aeb628a4bb9da9/json/licenses.json#L3442-L3445

https://raw.githubusercontent.com/spdx/license-list-data/v3.18/json/licenses.json

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -148,6 +148,8 @@ message CuratedPlugin {
   // spdx_license_id is the license of the plugin, which should be one of the identifiers
   // defined in https://spdx.org/licenses
   string spdx_license_id = 15;
+  // license_url specifies where the source for the plugin can be found.
```
pkwarren, line 151:
```suggestion
  // license_url specifies an optional URL for the plugin's license (if not using a standard spdx_license_id).
```

pkwarren (changes requested):

private/bufpkg/bufplugin/bufpluginconfig/testdata/success/go/buf.plugin.yaml
oliversun09: Making the url optional sounds good. 

3.18 has [more licenses](https://github.com/spdx/license-list-data/blob/36873db191f0ce8b6e54839738aeb628a4bb9da9/json/licenses.json#L6247), but it seems both [3.12](https://github.com/bufbuild/buf/blob/main/private/pkg/spdx/cmd/spdx-go-data/main.go#L40) and 3.18 have the same amount of attributes for each licenses, a license from [3.12](https://raw.githubusercontent.com/spdx/license-list-data/v3.12/json/licenses.json) also has `see_also`:
```
{
      "reference": "./ODbL-1.0.json",
      "isDeprecatedLicenseId": false,
      "detailsUrl": "./ODbL-1.0.html",
      "referenceNumber": 342,
      "name": "Open Data Commons Open Database License v1.0",
      "licenseId": "ODbL-1.0",
      "seeAlso": [
        "http://www.opendatacommons.org/licenses/odbl/1.0/",
        "https://opendatacommons.org/licenses/odbl/1-0/"
      ],
      "isOsiApproved": false
}
```
We can still upgrade to 3.18 just for more licenses.

oliversun09: Will generating the JSON file be part of this task?

I think we can do something like:

if the user has only provided id, display title and url by looking up in JSON
if the user has provided both id and url, display title from lookup, and the provided url
if the user has provided only url: display `LICENSE` or the provided (or something better) as title, and the provided url

.github/workflows/ci.yaml
```diff
@@ -4,7 +4,7 @@ on: push
 # Ref: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions
 permissions: read-all
 env:
-  MAKEFLAGS: '-j 2'
+  MAKEFLAGS: "-j 2"
```
pkwarren, line 7: I think we can revert the changes to this file (were you attempting to fix the failing tests related to Git submodules?).

oliversun09: you're right, I had the wrong impression that I reverted all the changes I made to this file

pkwarren (approved):

oliversun09:
I will update `spdx-go-data` in a separate PR

PR #1518, opened 2022-10-20
Fix git unit test by using fake git server rather than file://

robbertvanginkel:
More recent versions of git fix a CVE by disabling some usage of the `file://` transport, see https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253. We were using this transport in tests.

Instead, use https://git-scm.com/docs/git-http-backend to serve up this repository locally so we don't have to use the file protocol. This should be a more accurate tests, since we mostly expect submodules to come from servers.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -136,3 +136,8 @@ issues:
     - linters:
         - containedctx
       path: private/bufpkg/bufmodule/bufmoduleprotocompile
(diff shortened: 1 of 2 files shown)


joshcarp (approved):
lgtm

jhump (approved):
Nice!

PR #1519, opened 2022-10-21
Add visibility proto fields to create and delete audit events

njhale:
Add the visibility field to the create and delete audit event messages for both plugins and repositories.

Part of [BSR-620](https://linear.app/bufbuild/issue/BSR-620/add-visibility-to-repoplugin-createddeleted-audited-events).


unmultimedio (approved):
lgtm

PR #1520, opened 2022-10-21
bufcli: improve error when BUF_TOKEN is invalid

### LINKED ISSUE #1479: Improve error message when BUF_TOKEN is set and is invalid
```
$ buf mod update
Failure: you are not authenticated. Create a new entry in your netrc, using a Buf API Key as the password. For details, visit https://docs.buf.build/bsr/authentication
```

This happened when my `BUF_TOKEN` was set to a value that was no longer a valid buf token. I went and messed with my `.netrc` (ie i deleted it) but still got this error message. When I did `export BUF_TOKEN=` (as I was trying to access a public repository), this error message no longer happened.

We should update the error message to specify that this is the fault of `BUF_TOKEN`, not `.netrc`.

jchadwick-buf:
Closes #1479.


private/buf/bufcli/errors.go
```diff
@@ -181,6 +182,9 @@ func wrapError(err error) error {
 		connectCode := connectErr.Code()
 		switch {
 		case connectCode == connect.CodeUnauthenticated, isEmptyUnknownError(err):
+			if authErr, ok := bufconnect.AsAuthError(err); ok && authErr.TokenEnvKey() != "" {
+				return fmt.Errorf(`Failure: the %[1]s environment variable is set, but is not valid. Set %[1]s to a valid Buf API key, or unset it. For details, visit https://docs.buf.build/bsr/authentication`, authErr.TokenEnvKey())
+			}
```
pkwarren, line 187: Working great locally:
```
$ BUF_TOKEN=abcd ../cmd/buf/buf push
Failure: the BUF_TOKEN environment variable is set, but is not valid. Set BUF_TOKEN to a valid Buf API key, or unset it. For details, visit https://docs.buf.build/bsr/authentication
```

private/bufpkg/bufconnect/errors.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// Unwrap returns the underlying error.
+func (e *ErrAuth) Unwrap() error {
+	return e.cause
+}
+
+// Error implements the error interface and returns the error message.
+func (e *ErrAuth) Error() string {
+	return e.cause.Error()
```
pkwarren, line 33: What if there is no underlying cause? Do we want to fall back to a default error message in that case?

private/bufpkg/bufconnect/errors.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+// TokenEnvKey returns the environment variable used, if any, for authentication.
+func (e *ErrAuth) TokenEnvKey() string {
+	return e.tokenEnvKey
+}
+
+// AsAuthError uses errors.As to unwrap any error and look for an *ErrAuth.
+func AsAuthError(err error) (*ErrAuth, bool) {
```
pkwarren, line 42: nit: I know this file has a simple struct/methods, but it might be good to add some basic tests (maybe just exercising `AsAuthError`/`Error`). Even better might be a test for the interceptor that verifies that the `ErrAuth` is instantiated with the right `tokenEnvKey`.

pkwarren (approved):
Looks great.

private/bufpkg/bufconnect/errors.go
jchadwick-buf: Done; there's now tests for both.

pkwarren (approved):

PR #1529, opened 2022-10-24
[TCN-632] Modules Proto: Push and Download blobs

twilly:
`PushRequest` and `DownloadResponse` now define a way to move modules as a manifest of the module's files. The content is modeled as set of blobs with the manifest linking blobs to path names.

These new manifest and blobs fields will eventually be populated and should be prefered values over the module field. We'll ignore their values until we can populate them.

This commit adds tests to the module reader since it must receive but ignore the new fields. I hope it helps clarify that nil values are expected to be passed into bufmodule.NewModuleForProto.


private/bufpkg/bufapimodule/module_reader_test.go
```diff
@@ -0,0 +1,126 @@
...
+func (m *mockDownloadService) Download(
+	ctx context.Context,
+	owner string,
+	repository string,
+	reference string,
+) (*v1alpha1.Module, *v1alpha1.Blob, []*v1alpha1.Blob, error) {
+	if m.err != nil {
+		return nil, nil, nil, m.err
+	}
+	return m.module, m.manifest, m.blobs, nil
```
unmultimedio, line 55: You're missing testing this, right?

unmultimedio:
Another issue of having this as a feature branch is that the backend uses the latest main stable commit of buf, so the RPC proto will be visible only until this is merged to main.

proto/buf/alpha/registry/v1alpha1/push.proto
```diff
@@ -39,6 +39,10 @@ message PushRequest {
   repeated string tracks = 6 [deprecated = true];
   // If non-empty, the push creates a draft commit with this name.
   string draft_name = 7;
+  // If set, `module` is ignored and the module's manifest is used instead.
+  buf.alpha.module.v1alpha1.Blob manifest = 8;
+  // blobs to close on the manifest
+  repeated buf.alpha.module.v1alpha1.Blob blobs = 9;
```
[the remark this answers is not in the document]
unmultimedio: Correct, but I'm talking about the case of a hash collision, same scenario as [the linked comment](https://github.com/bufbuild/buf/pull/1499#discussion_r998819721), with this request the backend cannot know which blob matches to which file:

```
{
  ...
  blob manifest = """
  shake256:abc123  path/to/file1
  ...
  shake256:abc123  path/to/file2
  """
  []blob blobs = [
    {hash: "shake256:abc123", content: "//contents of file1"},
    ...
    {hash: "shake256:abc123", content: "//contents of file2"},
  ]
}
```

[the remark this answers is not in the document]
twilly: I was thinking of creating a manifest with the blob (`manifest.NewManifestFromReader(manifestBlob)`). Then iterate over the blobs by creating a digest for them and calling `manifest.Paths(digest)`.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -20,6 +20,27 @@ import "buf/alpha/breaking/v1/config.proto";
 import "buf/alpha/lint/v1/config.proto";
 import "google/protobuf/timestamp.proto";
 
+enum HashKind {
+  HASH_KIND_UNKNOWN_UNSPECIFIED = 0;
```
unmultimedio, line 24: Nit
```suggestion
  HASH_KIND_UNSPECIFIED = 0;
```

proto/buf/alpha/registry/v1alpha1/push.proto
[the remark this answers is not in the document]
unmultimedio: Discussed in zoom: if 2 files have the same digest, it's because either:

- They have the same content (consume contents of whatever match in the array it's fine, it's the same), or
- We've found a collision in the hash, which is highly unlikely, and probably a reason to change hash kind

Note: []blobs will have just one entry per digest, even if many files share that same content.

private/bufpkg/bufapimodule/module_reader_test.go
```diff
@@ -0,0 +1,126 @@
...
+			assert.NotNil(t, module)
+		}
+		if expectError {
+			assert.Error(t, err)
+			if errorContains != "" {
+				assert.ErrorContains(t, err, errorContains)
+			}
+		} else {
+			assert.NoError(t, err)
+		}
```
unmultimedio, line 124: Nit: Related to [this thread](https://bufprivate.slack.com/archives/C02KXVDB23B/p1665589006759369), if `err` is populated you can assume that module value is non-populated, and not idiomatic to be checked. So you could:

```suggestion
		module, err := moduleReader.GetModule(ctx, pin)
		if expectError {
			assert.Error(t, err)
			if errorContains != "" {
				assert.ErrorContains(t, err, errorContains)
			}
			assert.Nil(t, module)
			return
		}
		assert.NoError(t, err)
		assert.NotNil(t, module)
```

unmultimedio (approved):
Looks good

private/bufpkg/bufapimodule/module_reader_test.go
unmultimedio: Got it, but I mean testing with a module that has some content.

private/bufpkg/bufapimodule/module_reader_test.go
twilly: The Download method can return nil error and a nil module. That's the intent of `expectNilModule`.

Related to your question on testing with module content, I didn't actually test for this case yet. I should though.

twilly: On second thought, I'll leave that for later. A module must exist right now and I'm not sure if I'll make nil module possible or not yet.

unmultimedio: > The Download method can return nil error and a nil module.

Oh alright, there's no `ModuleNotFoundErr`. Then feel free to dismiss my comment :)

PR #1530, opened 2022-10-24
Restore old PATH behavior on Go 1.19+

jchadwick-buf:
Go 1.19 introduces a security improvement that prevents it from resolving PATH queries via relative PATH entries. However, this is a behavior change for Buf. This change restores the original behavior.

Because `exec.ErrDot` only exists in Go 1.19+, we need conditional compilation for now, to continue supporting Go 1.18.

Resolves TCN-424.


jhump:
This was supposedly disabled by default in Go 1.19 ostensibly over security concerns. Do we have any users that have built buf using 1.19 and complained about this? If not, perhaps it's a reasonable stance to keep the 1.19 behavior and, when we are ready to start building releases with Go 1.19, put it in the release notes as a security-minded change?

jchadwick-buf:
@jhump 
> This was supposedly disabled by default in Go 1.19 ostensibly over security concerns. Do we have any users that have built buf using 1.19 and complained about this? If not, perhaps it's a reasonable stance to keep the 1.19 behavior and, when we are ready to start building releases with Go 1.19, put it in the release notes as a security-minded change?

Hmm, honestly, now that you say it, I'm not really sure what the use case for this is. If it were to do something like `PATH=. buf ...` presumably you could just do `PATH=$PWD buf ...` instead and it would be nearly as convenient. The issue was raised internally by @akshayjshah. So, @akshayjshah: WDYT about adopting the new behavior and documenting it instead?

akshayjshah:
In isolation, the 1.19 behavior is reasonable. However, Buf makes a lot of public statements about how much we value backward compatibility. Since it doesn't seem like the security concerns are significant in our context, I'm in favor of applying this change and keeping user-visible behavior unchanged (no matter how trivial the fix is).

I ran into this with a Makefile that set `PATH=./bin` when running Buf. It was easy enough to change to `PATH=$PWD/bin`, but it was annoying nonetheless.

I wonder whether we ought to have a regression test before we merge this, though.

jchadwick-buf:
I believe we discussed this in a meeting and decided to go ahead with reverting the behavior. However, I don't think we wound up reviewing the code so far, so I'm going to re-request reviews. If we changed our mind on reverting the behavior, that's fine.

jhump (approved):
LGTM

PR #1532, opened 2022-10-24
add license in the response of GetModuleDocumentation

cyinma:
fixes BSR-712

this consider `LICENSE` as part of the documentations, and included it in the response of `GetModuleDocumentation`


doriable (approved):
This seems reasonable to me :)

PR #1534, opened 2022-10-25
update to latest Go dependencies

pkwarren:
Update to the latest makego and the latest Go dependencies. Manually pin previous version of protoreflect (tests failed on last tagged version).


go.mod
```diff
@@ -14,25 +14,25 @@ require (
 	github.com/jhump/protoreflect v1.13.1-0.20220928232736-101791cb1b4c
```
pkwarren, line 14: FYI - running `make updategodeps` changed this to `v1.13.0`, which then caused failing tests. If we need to pin a specific version, I think we should do this in the makefile (or perhaps tag a newer release to be pulled in).

rubensf (approved):

PR #1540, opened 2022-10-26
bufcli: add test, storagemem: NewReadWriteBucket2

twilly:
Coverage for bufcli.ReadModuleWithWorkspacesDisabled
----------------------------------------------------

ReadModuleWithWorkspacesDisabled didn't have coverage and I intend to do some minor refactoring soon. Now's a great time to document its behavior.

ReadModuleWithWorkspacesDisabled took a `appflag.Container`, but that's a very large interface to create and inject. I didn't see any obvious test fixture to provide it., so instead of building that I reduced signature to its minimal interface: a logger and an app.EnvStdinContainer.

IMHO, it's pretty odd to depend on a large interface type from a different package. I'll leave that yak shave for another day.

storagemem.NewReadWriteBucketWithOptions
------------------------------

It wasn't possible to initialize an in-memory storage bucket like you could with NewReadBucket. This new constructor produces a read-write bucket with the functionally of the read-only bucket.

NewReadWriteBucketWithOptions consumes the code in NewReadBucket, making the later nothing more than a type cast.

This change is in support of testing ReadModuleWithWorkspacesDisabled.

### DIFF
--- a/private/buf/bufcli/bufcli_test.go
+++ b/private/buf/bufcli/bufcli_test.go
@@ -122,13 +122,13 @@ func testReadModuleWithWorkspacesDisabled(
 			assert.NotNil(t, module)
 			assert.NotNil(t, identity)
 			assert.NoError(t, err)
-		} else {
-			if expectedErr != nil {
-				assert.ErrorIs(t, err, expectedErr)
-			}
-			if expectedErrContains != "" {
-				assert.ErrorContains(t, err, expectedErrContains)
-			}
+			return
+		}
+		if expectedErr != nil {
+			assert.ErrorIs(t, err, expectedErr)
+		}
+		if expectedErrContains != "" {
+			assert.ErrorContains(t, err, expectedErrContains)
 		}
 	})
 }
--- a/private/buf/bufcli/bufcli_test.go
+++ b/private/buf/bufcli/bufcli_test.go
@@ -36,7 +36,7 @@ func (m *mockBucketProvider) NewReadWriteBucket(
 	_ string,
 	_ ...storageos.ReadWriteBucketOption,
 ) (storage.ReadWriteBucket, error) {
-	return storagemem.NewReadWriteBucket2(storagemem.WithFiles(m.files))
+	return storagemem.NewReadWriteBucketWithOptions(storagemem.WithFiles(m.files))
 }
 
 func TestReadModuleWithWorkspacesDisabled(t *testing.T) {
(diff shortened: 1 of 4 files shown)


private/pkg/storage/storagemem/storagemem.go
```diff
@@ -25,15 +25,41 @@ import (
...
 	return newBucket(nil)
 }
 
-// NewReadBucket returns a new ReadBucket.
-func NewReadBucket(pathToData map[string][]byte) (storage.ReadBucket, error) {
-	pathToImmutableObject := make(map[string]*internal.ImmutableObject, len(pathToData))
-	for path, data := range pathToData {
+// NewReadWriteBucket2 returns a new in-memory ReadWriteBucket. Errors are
+// returned with invalid options.
+func NewReadWriteBucket2(opts ...Option) (storage.ReadWriteBucket, error) {
```
unmultimedio, line 55: Weird constructor name, may be `NewReadWriteBucketWithOptions(WithFiles(...), )` a bit of stutter but clearer? and mark as deprecated the other one?

unmultimedio (approved):
lgtm! I'd recommend getting more 👀  on this.

private/buf/bufcli/bufcli_test.go
```diff
@@ -0,0 +1,134 @@
...
+			assert.NotNil(t, identity)
+			assert.NoError(t, err)
+		} else {
+			if expectedErr != nil {
+				assert.ErrorIs(t, err, expectedErr)
+			}
+			if expectedErrContains != "" {
+				assert.ErrorContains(t, err, expectedErrContains)
+			}
+		}
```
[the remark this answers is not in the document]
doriable: Could also do an early return in the first `if` clause also. I think both are defensible, but definitely not blocking :)

doriable (approved):
This looks reasonable to me, but I do agree with @unmultimedio about the `NewReadWriteBucket2` constructor name. Would be good to have a more explicit name and then document that we are deprecating the previous function and migrating to the one taking options.

private/pkg/storage/storagemem/storagemem.go
twilly: I'm fine with a different suffix. The number pattern is pretty common with C and other languages which don't have overloading but you're trying to describe n ways of construction. Just in case y'all run into this in some other library. :)

private/pkg/storage/storagemem/storagemem.go
```diff
@@ -75,5 +76,5 @@ func NewReadWriteBucket2(opts ...Option) (storage.ReadWriteBucket, error) {
 
 // NewReadBucket returns a new ReadBucket.
```
unmultimedio, line 77: Same deprecation notice here?

unmultimedio (approved):
Not blocking nit

PR #1541, opened 2022-10-26
Remove `httpserver.{Mapper,Runner}`

stefanvanburen:
This removes `httpserver.{Mapper,Runner}` and associated code, in favor of using `httpserver.Run`.


stefanvanburen:
(Putting this in draft until we're ready to go over on the core side of things)

private/pkg/transport/http/httpserver/httpserver.go
```diff
@@ -34,176 +37,64 @@ const (
...
 )
 
-// Mapper initializes a Router.
-type Mapper interface {
-	Map(chi.Router) error
+type runner struct {
+	shutdownTimeout   time.Duration
+	readHeaderTimeout time.Duration
+	tlsConfig         *tls.Config
+	maxBodySize       int64
```
mfridman, line 44: We could probably remove this, because it's a property of the middleware and less of the runner. I think it's trivial to find this instead of logging it.

stefanvanburen: yeah, I missed that this was already in the new `router` package; makes sense that it's part of the middleware. I'll remove it.

mfridman (approved):

mfridman:
lgtm. just need to remove this so it compiles.

https://github.com/bufbuild/buf/blob/aba00688f1e942094a654b8ec1fc849e0602f78b/private/buf/cmd/buf/command/beta/studioagent/studioagent.go#L198

PR #1544, opened 2022-10-27
[BSR-721] New hash blobs constructor and validator

unmultimedio:
Changes in this PR:
- Separated `Digest` (and new `Digester`) code into its own `digest.go` file
- Separated `NewDigestFromBlob` into `module.go` in case we wanna break that dependency later.
- Separated test files accordingly (manifest/digest/module).
- Exported `DigestType` type, and supported types (just shake256 for now). It didn't make sense to have it private if it was required to be used from the consumer in raw string form anyway.
- Created a new `Digester` interface that allows digesting content w/out the need to do it via a `Manifest`.
- Removed any "file content" concern from the `Manifest`, now it only cares about "entries" or "pairs" of paths+ and its digests. Meaning it doesn't have a `hash` and it doesn't receive `AddContent` anymore, just `AddEntry(path, digest)`. The digests can be built using the aforementioned `Digester`, in an upper layer that cares about content.
- Added checks to make sure manifest paths have a unique digest.
- Since `Manifest` doesn't care about content anymore, it also doesn't check digest types. That was moved down to the `Digest` constructors.

This results in a nicer, slimmer exported funcs for `Manifest`, which only cares about paths and manifests:
image

And for `Digest`, we have 3 constructors, all of those using the same `NewDigestFromBytes` underneath, which errors (for now) if the dtype is not supported. This allows for future extension just adding a new dtype, without breaking exported interface.

Two concerns still in my mind:
- It's possible to declare a `Digest{}` w/out going through a constructor, so not sure if we want to make `Digest` into an interface.
  - The `manifest.AddEntry` func checks that the passed digest is populated/valid at least, but that check feels out of place. I thought of doing a `func (*Digest) Valid() bool` but it felt even weirder, why would we allow a Digest to be invalid?
- If a `Digest` is allowed to be constructed outside of the pkg, then somehow `digest []bytes` and `hexstr string` could diverge? feels both like two sources of the same state.


private/pkg/manifest/manifest.go
```diff
@@ -76,6 +82,17 @@ func NewDigestFromString(typedDigest string) (*Digest, error) {
 	return NewDigestFromHex(hashfunc, digestStr)
 }
 
+func NewDigestFromBlobHash(hash *modulev1alpha1.Hash) (*Digest, error) {
```
twilly, line 85: I'm thinking we should put this into its own file, perhaps `module.go`? manifest depends on storage and with this adds a dependency on our proto package. Separating out that dep may help in breaking a cycle later.

Also this function needs docs. (I should get [TCN-623](https://linear.app/bufbuild/issue/TCN-623/bufbuildbuf-is-not-linting-for-comments-on-exported-types) in flight.)

private/pkg/manifest/manifest.go
```diff
@@ -93,6 +110,19 @@ func (d *Digest) Hex() string {
 	return d.hexstr
 }
 
+// Valid checks if the current digest matches a passed content, and any error
+// attempting to validate.
+func (d *Digest) Valid(content io.Reader) (bool, error) {
```
twilly, line 115: I'm not sure I like this style. It forces Digest to know how to construct hashers; in this case it always constructs a new one.

Instead Digests can have equality defined:

```go
func (d *Digest) Equal(other *Digest) bool {
    return d.dtype == other.dtype && bytes.Equal(d.digest, other.digest)
}
```

Then you can have a Shake digest constructor, akin to your `shake256DigestFrom`, that can return Digests for content. Validation becomes this:

```go
digest.Equal(newShake256Digester().Sum(content))
```

private/pkg/manifest/manifest.go
```diff
@@ -214,46 +244,49 @@ func (m *Manifest) addDigest(path string, digest *Digest) error {
...
-func (m *Manifest) Digest(path string) (digest *Digest, ok bool) {
-	digestValue, ok := m.pathToDigest[path]
-	digest = &digestValue
-	return digest, ok
-}
-
-// MarshalText encodes the manifest into its canonical form.
-func (m *Manifest) MarshalText() ([]byte, error) {
-	// order by path
+// Paths returns a sorted list of all paths in the manifest.
```
twilly, line 254: I generally dislike APIs that return values with a specific order when they must do extra work to make that happen. Sorting adds significant overhead for callers who don't care about order, and especially bad (sometimes O(n^2) bad) overhead if the caller desires a different order.

I'd change this to return without a call to `sort.Strings` and have `MarshalText` sort the value.

private/pkg/manifest/manifest.go
unmultimedio: > into its own file

You mean its own pkg? I had the same thought remembering your decision with `storage`. What about `private/pkg/manifest/manifestintegrations` for both `storage` and `proto`?

twilly: Nah, just putting this kind of support into its own file in the manifest package. That way it's clear where dep lies and can be moved later if the dep needs to be broken.

private/pkg/manifest/manifest.go
unmultimedio: I like that idea, but then I'm thinking we should take it one step further: have a `type DigestType string`, have a `type Digester interface{ Digest(content io.Reader) (*Digest, error) }`, export available digest types (for now just shake256), and make all Manifest/Digest constructors receive the type.

private/pkg/manifest/manifest.go
unmultimedio: Done

private/pkg/manifest/manifest.go
unmultimedio: Done, created a new `Digester` interface for this equality logic, much cleaner, thanks!

private/pkg/manifest/manifest.go
unmultimedio: Done

twilly (approved):
I love it! I knew the digest abstraction wasn't complete an you took it all the way. Fantastic!

I'm excited to add my own module mapper (bytes to blob) that I have in a change I'm working on.

PR #1545, opened 2022-10-27
Use connect.IsWireError for client error scenario

stefanvanburen:
Also unskips the test that was skipped due to this issue, which doesn't *seem* flaky now?

### DIFF
--- a/private/bufpkg/bufstudioagent/plain_post_handler.go
+++ b/private/bufpkg/bufstudioagent/plain_post_handler.go
@@ -169,25 +169,16 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// TODO(rvanginkel) should this context be cloned to remove attached values (but keep timeout)?
 	response, err := client.CallUnary(r.Context(), request)
 	if err != nil {
-		// TODO deal with this error handling using `connect.IsFromServer(err)`
-		// instead of these heuristics, once it's available. See
-		// https://github.com/bufbuild/connect-go/issues/222
-		//
 		// We need to differentiate client errors from server errors. In the former,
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufstudioagent/plain_post_handler.go
```diff
@@ -169,25 +169,16 @@ func (i *plainPostHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		if netErr := new(net.OpError); errors.As(err, &netErr) {
-			http.Error(w, err.Error(), http.StatusBadGateway)
-			return
-		}
-		if urlErr := new(url.Error); errors.As(err, &urlErr) {
+		// need to check *first* if it's not a wire error, so we can assume the
+		// request never left the client, or a response never arrived from the
+		// server. In those scenarios we trigger a `StatusBadGateway` to signal
+		// that the upstream server is unreachable or in a bad status...
+		if !connect.IsWireError(err) {
```
unmultimedio, line 181: I'm confused about this func name. A _wire_ error is communicating the error _actually came from the server_? I think older `IsFromServer` naming was [much clearer](https://github.com/bufbuild/connect-go/pull/374/files#r997322851)? cc @joshcarp

unmultimedio (approved):
Anyway, not blocking, I feel the exported connect func name is backwards. Thanks for this!

private/bufpkg/bufstudioagent/plain_post_handler.go
stefanvanburen: here's a bit of [discussion about the name](https://github.com/bufbuild/connect-go/pull/374#discussion_r997369512)

PR #1552, opened 2022-10-31
BSR-702/managed mode enabled false shouldnt fail buf gen

oliversun09:
https://linear.app/bufbuild/issue/BSR-702/enabled-configuration-should-not-fail-code-generation

In generate template V1, setting `enabled` to `false` in managed mode would fail the generation altogether, this PR makes it so that having `enabled: false` would be equivalent of having `managed` not defined at all.


private/buf/bufgen/config.go
```diff
@@ -43,44 +45,47 @@ func readConfig(
...
 		case ".json":
-			return getConfigJSONFile(override)
+			return getConfigJSONFile(logger, override)
 		case ".yaml", ".yml":
-			return getConfigYAMLFile(override)
+			return getConfigYAMLFile(logger, override)
 		default:
-			return getConfigJSONOrYAMLData(override)
+			return getConfigJSONOrYAMLData(logger, override)
 		}
```
oliversun09, line 53: Letting `getConfigJSONFile`, `getConfigYAMLFile` and `getConfigJSONOrYAMLData` use `provider.logger` would  also have worked (and would've reduced the number of changes), but I don't think that's good, because that logger is meant for `provider` only.

private/buf/bufgen/testdata/v1/go_gen_error1.yaml
```diff
@@ -1,7 +1,12 @@
 version: v1
```
oliversun09, line 1: I renamed `go_gen_error1.yaml` to `gen_success_7.yaml`, but it left the error files missing `go_gen_error1.yaml`., which is why I renamed `go_gen_error7.yaml` to `go_gen_error1.yaml` to fill the gap. (This is ok because there was not a strong logical order among the test files)

private/buf/bufgen/config.go
oliversun09: Defining something like a `ConfigReader` and injecting the logger into it could potentially reduce the need to pass the logger down through 3 function calls, but I'm not sure if that's necessary.

doriable: Yes, I think you are correct in not adding the logger to any of the config structures, since it is not extensively used. I agree with this approach 👍🏼

doriable (commented):
The approach looks good to me! Before I stamp, we just need to make sure that we have a changelog entry for this, since it will be changing user-facing behaviour. Thanks for putting this together! :)

private/buf/bufgen/config.go
```diff
@@ -229,11 +235,11 @@ func checkPathAndStrategyUnset(id string, plugin ExternalPluginConfigV1, pluginI
...
 }
 
-func newManagedConfigV1(externalManagedConfig ExternalManagedConfigV1) (*ManagedConfig, error) {
-	if !externalManagedConfig.Enabled && !externalManagedConfig.IsEmpty() {
-		return nil, errors.New("managed mode options are set but 'managed.enabled: true' is not set")
-	}
+func newManagedConfigV1(logger *zap.Logger, externalManagedConfig ExternalManagedConfigV1) (*ManagedConfig, error) {
 	if !externalManagedConfig.Enabled {
+		if !externalManagedConfig.IsEmpty() && logger != nil {
+			logger.Sugar().Warn("managed mode has options present, but is not enabled")
```
doriable, line 241: nit: we should keep the wording consistent with the previous message. maybe:

```suggestion
			logger.Sugar().Warn("managed mode options are set but are not enabled")
```

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Allow setting `enabled: false` under `managed` in `buf.gen.yaml`, in which 
+  case `buf generate` will ignore managed mode options.
```
doriable, line 6:
```suggestion
- When using managed mode, setting `enabled: false` now no longer fails `buf generate` 
  and instead prints a warning log and ignores managed mode options.
```

doriable (approved):
Small comment on the changelog statement, but otherwise this looks good to me. Going to approve now, no need for re-review. Thanks!

PR #1553, opened 2022-11-01
add GetRepositoriesMetadata rpc

cyinma:
fixes BSR-759

For context: there are multiple fields that we would like to display together with repository, e.g. the verification status of owner, or the last commit time. Putting these fields into `Repository` doesn't seems right as they are not actually part of the `Repository`, requiring frontend to retrieve these information from respective owner or commit also doesn't sounds feasible as it will create al lot more overhead to get all the necessary information, especially for listing. Multiple ways mentioned below have explored and this seems to be acceptable, but I am open to any suggestion.

1. separated object (let say `RepositoryWithMetadata`) which contains all the necessary information, and return it in new endpoints - this basically means reimplementing most endpoints so doesn't sounds right
2. add new service (or add new endpoints in existing display service) to serve these frontend specific information - this sounds future proof, and we have mentioned that we will eventually need this, but I still cannot figure any potential use cases except for metadata, which sounds much more tied to `RepositoryService` and `PluginCurationService` instead
3. (in this PR) new endpoint in a service to serve information related to that - this seems to be a least worst way to do it?


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -114,6 +130,9 @@ service RepositoryService {
   rpc GetRepositorySettings(GetRepositorySettingsRequest) returns (GetRepositorySettingsResponse);
   // UpdateRepositorySettingsByName updates the settings of a repository.
   rpc UpdateRepositorySettingsByName(UpdateRepositorySettingsByNameRequest) returns (UpdateRepositorySettingsByNameResponse);
+  // GetRepositoriesMetadata gets the metadata of the repositories in the request, the length of repositories in the
+  // request should match the length of the metadata in the response, but the order in the response is unspecified.
```
robbertvanginkel, line 134: Under what conditions do we need to load metadata from multiple repositories at the same time? 

I think we do have some other API endpoints where we match the order in request/responses, we should probably do the same here as well.

cyinma: let say in a repositories list, its likely that we will want to load last commit time and owner verification status on the list, hence loading metadata from multiple repositories at the same time.

for match the order in request/responses, that was what I thought originally too, but it doesn't seems to be the case for the `GetRepositoriesByFullName`, which also mention `Response order is unspecified`. However it seems like `CodeGenerationService.GenerateCode` and `GenerateService.GeneratePlugins` have the response order matches the request order, so I am open to change it.

robbertvanginkel (approved):

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -75,6 +76,21 @@ message RepositoryContributor {
   RepositoryRole implicit_role = 4;
 }
 
+message RepositoryMetadata {
```
timostamm, line 79: We also have [RepositoryCounts](https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/repository.proto#L59-L65). Are they going to stay seperate? Seems like they would belong here.

timostamm (commented):
> Putting these fields into Repository doesn't seems right as they are not actually part of the Repository

TBH, this option seems the most practical to @paul-sachs and myself. Are there any technical constraints?

robbertvanginkel:
> TBH, this option seems the most practical to @paul-sachs and myself. Are there any technical constraints?

The hesitation is that the `Repository` message isn't frontend specific, used in more context's, and we don't want to expand it to serve frontend design of the day. A verification status I could perhaps see as a property of the repo, but things like time and license id for the last commit seem like we would include them to make rendering a frontend easier rather than because it makes sense from a data model perspective.

paul-sachs:
> The hesitation is that the Repository message isn't frontend specific, used in more context's, and we don't want to expand it to serve frontend design of the day. 

That's fair. We definitely need to have a larger conversation about how we're building RPCs because these choices can definitely make the UI slower, and harder to reason about. That said, that conversation needs to happen not at a case by case basis, but holistically and thoughtfully. Not sure if there's a notion doc existing to gather these kinds of concerns but we should add this PR.

If we want to hold that these RPCs should be as self-contained and not cater to a specific UI, then choice 3. is probably the most sensible.

timostamm (approved):
Thanks for the context, Robbert.

PR #1556, opened 2022-11-02
update go to 1.19.3 and update go deps

pkwarren:


make/buf/all.mk
```diff
@@ -24,6 +24,7 @@ DOCKER_BINS := $(DOCKER_BINS) buf
 FILE_IGNORES := $(FILE_IGNORES) \
 	.build/ \
 	.ctrlp \
+	.idea/ \
```
pkwarren, line 27: Added this for Goland users 😄

ElliotMJackson (commented):
is it worth dropping the patch altogether?
```yaml
go-version: 1.19.x
```
&
```sh
DOCKER_IMAGE=golang:1.19-bullseye bash make/buf/scripts/release.bash
```

pkwarren:
> ```yaml
> go-version: 1.19.x
> ```

This would be ok I think, although everything is already automated with `make updategoversion GOVERSION=1.19.3` already so it isn't too bad.

> ```shell
> DOCKER_IMAGE=golang:1.19-bullseye bash make/buf/scripts/release.bash
> ```

We should keep this patch version - that way we know we're always pulling in the expected version (and don't have old cached versions being used).

jchadwick-buf (approved):

PR #1558, opened 2022-11-02
[BSR-721] manifest bucket and blob set

unmultimedio:


private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,95 @@
...
+		return nil
+	}); walkErr != nil {
+		return nil, walkErr
+	}
+	return m, nil
+}
+
+// ToBucket takes a map of digest to files' contents, and builds a storagemem
+// bucket from the manifest.
+func (m *Manifest) ToBucket(digestToFiles map[string][]byte) (storage.ReadBucket, error) {
```
unmultimedio, line 69: Uncertain about what to receive here. For this PR I'm receiving:

```
{
  "shake256:the-digest": "the actual file content",
}
```

But we could receive:

```
{
  "shake256:the-digest": io.Reader,
}
```

Although in most scenarios I imagine the caller already has the content in memory (like from the backend, in blob.Content), so no need for the indirection?

Also, not receiving:

```
{
  "file-path": "content or reader",
}
```

To avoid content duplication.

Another thing I considered, but not sure if it's a good idea to put that much logic into the Manifest, would be making sure the content matches the received content, by rehashing it. That way we could even just receive an array of de-duplicated contents.

twilly: `digestToFiles map[string][]byte` ties pretty hard to a string representation of a digest and content being in memory. Even `[]byte` is changed to an io.Reader, it's still a forced string representation.

`NewDigestFromBlobHash` makes a nice Digest type, so what if we make a Blob abstraction from a proto Blob? Something that has a Digest and can be an io.ReadCloser.

Warning: hastily thrown together code below! Figure out where to put it and maybe even make it work!

Let's define what a blob is and make a in-memory blob implementation.

```go
type Blob interface {
  Digest() *Digest
  Open(ctx context.Context) (io.ReadCloser, error)
}

type MemoryBlob struct {
  digest *Digest
  content []byte
}

var _ Blob = (*InMemoryBlob)(nil)

func NewMemoryBlob(digest *Digest, content []byte) *MemoryBlob {
  return &MemoryBlob{digest: digest, content: content}
}

func (memblob *MemoryBlob) Digest() *Digest {
  return memblob.digest
}

func (memblob *MemoryBlob) Open(_ context.Context) io.ReadCloser {
  return bytes.NewReader(memblob.content), nil
}
```

Then if we tie the open'd path to an io.ReadCloser:

```go
type storageCanBeSilly struct {
   path string
   file io.ReadCloser
}

func (silly *storageCanBeSilly) Path() string { return silly.path}
func (silly *storageCanBeSilly) ExternalPath() string { return silly.path}
func (silly *storageCanBeSilly) Read(p []bytes) (int, error) { return silly.file.Read(p) }
func (silly *storageCanBeSilly) Close() error { return silly.file.Close() }
```

Now we can implement a storage.ReadBucket by tying a manifest and an array of blobs together.

```go
type blobBucket struct {
  bag map[string]Blob
}

func NewBlobBucket(manifest *Manifest, blobs []Blob) storage.ReadBucket {
  bb := &blobBucket{bag: make(map[string]Blob])}
  for _, blob := range blobs {
    bb.bag[blob.Digest().String()] = blob
  }
  return bb
}

func (bb *blobBucket) Get(ctx context.Context, path string) (ReadObjectCloser, error) {
  digest, ok := bb.manifest.DigestFor(path)
  if !ok {
    return nil, errors.New("not found in manifest")
  }
  blob, ok := bb.bag[digest.String()]
  if !ok {
    return nil, errors.New("not found in blob bag")
  }
  file, err := blob.Open(ctx)
  if err != nil {
    return nil, err
  }
  return &storageCanBeSilly{path: path, file: file}, nil
}
```

This works for in-memory *and* can support on-disk or in GCS blobs. Those just need a different implementation if you want to be memory efficient.

unmultimedio: Thanks, that's a great suggestion. Pushed it!

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+		return nil
+	}); walkErr != nil {
+		return nil, walkErr
+	}
+	return m, nil
+}
+
+// NewBucket takes a manifest and an array of blobs and builds a readable storage bucket that
+// contains the files in the manifest.
+func NewBucket(m Manifest, blobs []Blob) (storage.ReadBucket, error) {
```
unmultimedio, line 80: I feel very strongly we should receive a map here instead
```suggestion
func NewBucket(m Manifest, blobs map[string]Blob) (storage.ReadBucket, error) {
```

That way we don't need to do the checking of duplicate blobs (and possible different content) for the same digest.

unmultimedio: Although, on second thought, in here we'd need to still check that key and value matches:

```
{
  "shake256:abc123": {"digest": "shake256:abc123", "content": "..."}
  |----- x -------|             |----- x -------|
}
```

So let's keep it as array I guess 😢 

For more background, I was thinking of converting everything to just `map[digest]content` and *that* would be the bag of blobs. But that has two downsides:

- A manifest couldn't any longer be represented as a Blob (or a weird map of 1 entry?)
- A digest(kind, bytes) as map key is not fun for proto, and we'd need to always use the string representation of it (would have this been solved by using multihash, where everything is just []byte?)

private/pkg/manifest/module.go
```diff
@@ -24,6 +28,50 @@ var hashKindToDigestType = map[modulev1alpha1.HashKind]DigestType{
...
+}
+
+var _ Blob = (*memoryBlob)(nil)
+
+// NewMemoryBlob takes a digest and a content, and turns it into an in-memory
+// representation of a blob, which returns the digest and an io.ReadCloser for
+// its content. It optionally validates that the passed digest and content
+// match.
+func NewMemoryBlob(digest Digest, content []byte, validateHash bool) (Blob, error) {
+	if validateHash {
```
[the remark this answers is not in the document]
twilly: I absolutely agree that blobs should have the option to validate. This is fine but I suggest a functional option pattern for validateHash. Reason being is validating from a stream can have one of two ways of validating. If it's seekable, it can be validated on construction. If not, it can only be validated with an error code on `Close`. So we have two ways of specifying validation: on construction and on close. Having a functional option for validation would make for a touch of consistency.[1]

Minor though. We can do this when we get to saving/reading blobs from persistent storage (cache, in the BSR, etc).

[1] Even when you have seekable input, validating on close can have performance impact, even with memory! Scanning memory twice will blow the CPU cache.

private/pkg/manifest/module.go
```diff
@@ -24,6 +28,50 @@ var hashKindToDigestType = map[modulev1alpha1.HashKind]DigestType{
...
+		content: content,
+	}, nil
+}
+
+func (b *memoryBlob) Digest() *Digest {
+	return &b.digest
+}
+
+func (b *memoryBlob) Open(ctx context.Context) (io.ReadCloser, error) {
+	return ioutil.NopCloser(bytes.NewReader(b.content)), nil
```
twilly, line 72: I caught this too in my hacked together comment's code. 😆

private/pkg/manifest/module_test.go
```diff
@@ -43,3 +45,60 @@ func TestDigestFromBlobHash(t *testing.T) {
 	require.NoError(t, err)
 	assert.Equal(t, digestFromContent.String(), digestFromBlobHash.String())
 }
+
+func TestNewMemoryBlob(t *testing.T) {
+	t.Parallel()
+	const content = "some file content"
```
twilly, line 51: For less casting?

```suggestion
	const content = []byte("some file content")
```

private/pkg/manifest/module_test.go
```diff
@@ -43,3 +45,60 @@ func TestDigestFromBlobHash(t *testing.T) {
...
+}
+
+func TestInvalidNewDigestFromBlobHash(t *testing.T) {
+	t.Parallel()
+	_, err := manifest.NewDigestFromBlobHash(nil)
+	require.Error(t, err)
+	_, err = manifest.NewDigestFromBlobHash(&modulev1alpha1.Hash{
+		Kind: modulev1alpha1.HashKind_HASH_KIND_UNSPECIFIED,
+	})
+	require.Error(t, err)
```
twilly, line 98: `assert.Error` since the test can correctly validate the next statement? Or subtest?

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+	"io"
+	"sort"
+
+	"github.com/bufbuild/buf/private/pkg/normalpath"
+	"github.com/bufbuild/buf/private/pkg/storage"
+	"github.com/bufbuild/buf/private/pkg/storage/storageutil"
+	"go.uber.org/multierr"
+)
+
+// manifestBucket is a storage.ReadBucket implementation from a manifest an an
```
twilly, line 29:
```suggestion
// manifestBucket is a storage.ReadBucket implementation from a manifest and an
```

private/pkg/manifest/storage.go
twilly: I totally understand the want for a map; it's the natural way of declaring a set in Go. Precisely the type we wish to describe a bag of blobs.

multihash would solve one problem in that it's not a composite of two values, but []byte isn't a valid map key type. We would need to different abstraction for a set.

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+
+// NewBucket takes a manifest and an array of blobs and builds a readable storage bucket that
+// contains the files in the manifest.
+func NewBucket(m Manifest, blobs []Blob) (storage.ReadBucket, error) {
+	digestToBlobs := make(map[string]Blob, len(blobs))
+	for _, b := range blobs {
+		digest := b.Digest().String()
+		_, alreadyPresent := digestToBlobs[digest]
+		if alreadyPresent {
+			return nil, fmt.Errorf("blob with digest %q duplicated", digest)
```
twilly, line 86: Given blobs with the same digest must have the same content, if `NewBucket` can (optionally) validate duplicated blobs, then we may be able to reduce duplicated code.

In line with `[]Blob` being *meh* and a map being impossibly *meh*, maybe a blob bag representing a set of blobs should be its own type. Pass in a `[]Blob` to construct. Have it optionally validate blobs and optionally deduplicate. Define `Blob.For(Digest) (Blob, ok bool)` for users like `NewBucket`.

WDYT?

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+		pathDigest, ok := m.DigestFor(path)
+		if !ok {
+			return nil, fmt.Errorf("path %q not present in manifest", path)
+		}
+		blob, ok := digestToBlobs[pathDigest.String()]
+		if !ok {
+			return nil, fmt.Errorf("manifest path %q with digest %q has no associated blob", path, pathDigest.String())
+		}
+		pathToBlob[path] = blob
+	}
```
twilly, line 105: I'm not sure building a map of path -> blob helps. manifest.DigestFor is a path -> digest, so with digestsToBlobs we have a complete path -> digest -> blob chain.

Validating a manifest is complete also means it's not possible to use an incomplete manifest to read what we do have. Minor but with the duplicated lookup I think this is better modeled as a validator. `Validate() bool` for bucket completeness would serve both needs and minimize memory.

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+
+func (m *manifestBucket) Stat(ctx context.Context, path string) (storage.ObjectInfo, error) {
+	_, ok := m.pathToBlob[path]
+	if !ok {
+		return nil, storage.NewErrNotExist(path)
+	}
+	// storage.ObjectInfo only requires path
+	return &manifestBucketObject{path: path}, nil
+}
+func (m *manifestBucket) Walk(ctx context.Context, prefix string, f func(storage.ObjectInfo) error) error {
```
twilly, line 135:
```suggestion
}

func (m *manifestBucket) Walk(ctx context.Context, prefix string, f func(storage.ObjectInfo) error) error {
```

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+	prefix, err := storageutil.ValidatePrefix(prefix)
+	if err != nil {
+		return err
+	}
+	paths := m.manifest.Paths()
+	sort.Slice(paths, func(i, j int) bool {
+		return paths[i] < paths[j]
+	})
```
twilly, line 143: I don't think paths are walked in any sort order. At least it's not defined by ReadBucket.Walk and I can't imagine another implementation trying to force such (e.g. storageos).

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,158 @@
...
+	for _, path := range paths {
+		if !normalpath.EqualsOrContainsPath(prefix, path, normalpath.Relative) {
+			continue
+		}
+		_, ok := m.pathToBlob[path]
+		if !ok {
+			return storage.NewErrNotExist(path)
+		}
```
twilly, line 151: This is impossible. I'd comment about that. :)

private/pkg/manifest/module_test.go
unmultimedio: I wish, but slices cannot be consts :/

private/pkg/manifest/storage.go
unmultimedio: I found this behavior in `storagemem` [here](https://github.com/bufbuild/buf/blob/61c2f49aa335dce8bc2311c116b84b61999c4ee5/private/pkg/storage/storagemem/bucket.go#L63-L70).

private/pkg/manifest/storage.go
unmultimedio: Yeah, 2 places in which I thought the same haha, I'll add a comment.

unmultimedio: Although... if we allow [incomplete manifests](https://github.com/bufbuild/buf/pull/1558/files#r1014506057) it could happen.

private/pkg/manifest/storage.go
unmultimedio: Removed

private/pkg/manifest/storage.go
unmultimedio: Nice catch! changed the map. Will push the validate functional arg in a bit

private/pkg/manifest/storage.go
unmultimedio: Done!

private/pkg/manifest/storage.go
unmultimedio: Done, allowed for incomplete manifests.

private/pkg/manifest/module.go
unmultimedio: TIL: `ioutil` is deprecated. `io` has the same functionality.

private/pkg/manifest/module.go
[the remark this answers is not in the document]
unmultimedio: Done, ptal

private/pkg/manifest/module_test.go
twilly: Ah, and you're preferring const over the many cast. Works for me.

private/pkg/manifest/module.go
```diff
@@ -24,6 +27,133 @@ var hashKindToDigestType = map[modulev1alpha1.HashKind]DigestType{
 	modulev1alpha1.HashKind_HASH_KIND_SHAKE256: DigestTypeShake256,
 }
 
+// Blob is a blob with a digest and a content.
+type Blob interface {
+	Digest() *Digest
+	Open(context.Context) (io.ReadCloser, error)
```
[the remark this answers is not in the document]
twilly: Correct-o-mundo! I don't mind not passing context either and leave it up to callers/handlers to figure that stuff out. I think threading it is less friction than trying to describe blocking open behavior.,

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+	return &manifestBucket{
+		manifest: m,
+		blobs:    blobs,
+	}, nil
+}
+
+func (m *manifestBucket) Get(ctx context.Context, path string) (storage.ReadObjectCloser, error) {
+	digest, ok := m.manifest.pathToDigest[path]
```
twilly, line 135: IMHO, I'd use the equivalent DigestFor. Less private member access, even if we're in the same module.

```suggestion
	digest, ok := m.manifest.DigestFor(path)
```

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,190 @@
...
+	if !ok {
+		return nil, storage.NewErrNotExist(path)
+	}
+	blob, ok := m.blobs.digestToBlob[digest.String()]
```
twilly, line 139: Same kind of thing for this. A blob set method `BlobFor(digest)`. Slightly less coupling. :)

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil, err
+	}
+	return &manifestBucketObject{
+		path: path,
+		file: file,
+	}, nil
+}
+
+func (m *manifestBucket) Stat(ctx context.Context, path string) (storage.ObjectInfo, error) {
+	digest, ok := m.manifest.pathToDigest[path]
```
twilly, line 154: Ditto in here, though the code duplication suggests a small func could be useful. Maybe `manifestBucket.blobFor(path) (Blob, error)`?

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,190 @@
...
+	prefix, err := storageutil.ValidatePrefix(prefix)
+	if err != nil {
+		return err
+	}
+	// walk order is not guaranteed
+	for _, path := range m.manifest.Paths() {
+		if !normalpath.EqualsOrContainsPath(prefix, path, normalpath.Relative) {
+			continue
+		}
+		digest, ok := m.manifest.pathToDigest[path]
```
twilly, line 175: 🔁

twilly (commented):
Only small things! This is looking good!

private/pkg/manifest/storage.go
unmultimedio: Moved this already to a private func, this would be the only usage, do you think that's still needed?

private/pkg/manifest/storage.go
unmultimedio: Did `manifestBucket.blobFor(path)` suggestion instead.

private/pkg/manifest/storage.go
twilly: The reason I was suggesting this as blob set method instead of a manifest method is to decouple access on private fields between types, in this case `m.blobs.digestToBlob`. It'll help if manifest buckets are ever placed in a package different from blobs.

Not a big deal but that was my thought process. :)

twilly (approved):

twilly:
Lint will be fixed when https://github.com/bufbuild/buf/pull/1572 lands.

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,192 @@
...
+				return nil, fmt.Errorf("manifest path %q with digest %q has no associated blob", path, pathDigest.String())
+			}
+		}
+	}
+	if config.noExtraBlobs {
+		for digestStr := range blobs.digestToBlob {
+			if _, ok := m.digestToPaths[digestStr]; !ok {
+				return nil, fmt.Errorf("blob with digest %q is not present in the manifest", digestStr)
+			}
+		}
```
unmultimedio, line 126: Decoupling private fields also requires something like `blobSet.Digests()` (similar to `manifest.Paths()`), which causes this weird looping as we've seen in paths already:

```go
for _, digest := blobSet.Digests() {
  blob, ok := blobSet.BlobFor(digest)
  if !ok {
    // this should/can never happen, but still feels like unnecessary boilerplate
  }
  // use blob
}
```

May be re-think exposing an iterator? or just expose a copy of the map? Since it's not critical, I'll leave it for later thought.

private/pkg/manifest/storage.go
unmultimedio: Done! [related](https://github.com/bufbuild/buf/pull/1558#discussion_r1019531518)

twilly (commented):
🥳

PR #1559, opened 2022-11-03
buf plugin config: add import_style in npm registry config

mfridman:
This PR adds an npm registry config to tune the import style in the package.json. Use by the BSR NPM registry implementation.

The other part of this PR reverts #1488, we don't want to use configs to drive hidden plugins as these files (in the future) may not be solely controlled by us.


private/bufpkg/bufplugin/bufplugin_test.go
```diff
@@ -53,10 +53,13 @@ func TestPluginRegistryRoundTrip(t *testing.T) {
 		},
 	})
 	assertPluginRegistryRoundTrip(t, &bufpluginconfig.RegistryConfig{
-		NPM: &bufpluginconfig.NPMRegistryConfig{},
+		NPM: &bufpluginconfig.NPMRegistryConfig{
+			ImportStyle: "module",
```
mfridman, line 57: If we're going to add enums, we might as well make this an explicit field which must be set for any npm registry compatible plugin. 99% of the time this should be "module" but we'll be explicit about it.

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -144,8 +143,13 @@ func PluginRegistryToProtoRegistryConfig(pluginRegistry *bufpluginconfig.Registr
 		}
 		registryConfig.RegistryConfig = &registryv1alpha1.RegistryConfig_GoConfig{GoConfig: goConfig}
 	} else if pluginRegistry.NPM != nil {
+		importStyle, err := goImportStyleToProtoImportStyle(pluginRegistry.NPM.ImportStyle)
```
pkwarren, line 146:
```suggestion
		importStyle, err := npmImportStyleToProtoImportStyle(pluginRegistry.NPM.ImportStyle)
```

mfridman: Go here was in reference to "go code" -> proto. But I can change this if it makes it clearer as suggested

pkwarren (approved):

PR #1560, opened 2022-11-03
internaltesting: also inject CONFIG_DIR envvar

twilly:
CONFIG_DIR is useful for controlling TLS.

### DIFF
--- a/private/buf/cmd/buf/internal/internaltesting/internaltesting.go
+++ b/private/buf/cmd/buf/internal/internaltesting/internaltesting.go
@@ -48,8 +48,9 @@ func NewEnvFunc(tb testing.TB) func(string) map[string]string {
 	tempDirPath := tb.TempDir()
 	return func(use string) map[string]string {
 		return map[string]string{
-			useEnvVar(use, "CACHE_DIR"): tempDirPath,
-			"PATH":                      os.Getenv("PATH"),
+			useEnvVar(use, "CACHE_DIR"):  tempDirPath,
+			useEnvVar(use, "CONFIG_DIR"): tempDirPath,
+			"PATH":                       os.Getenv("PATH"),
 		}
 	}
 }


private/buf/cmd/buf/internal/internaltesting/internaltesting.go
```diff
@@ -48,8 +48,9 @@ func NewEnvFunc(tb testing.TB) func(string) map[string]string {
 	tempDirPath := tb.TempDir()
 	return func(use string) map[string]string {
 		return map[string]string{
-			useEnvVar(use, "CACHE_DIR"): tempDirPath,
-			"PATH":                      os.Getenv("PATH"),
+			useEnvVar(use, "CACHE_DIR"):  tempDirPath,
+			useEnvVar(use, "CONFIG_DIR"): tempDirPath,
+			"PATH":                       os.Getenv("PATH"),
```
akshayjshah, line 53: Seems reasonable in isolation! Are there constants available for these env var strings?

akshayjshah (approved):

private/buf/cmd/buf/internal/internaltesting/internaltesting.go
twilly: CONFIG_DIR feeds [into the appname interface](https://github.com/bufbuild/buf/blob/main/private/pkg/app/appname/appname.go#L43-L48), as does CACHE_DIR and DATA_DIR. CONFIG_DIR is searched for application configuration, which really means [TLS configuration](https://github.com/bufbuild/buf/blob/main/private/buf/bufapp/bufapp.go#L28-L33).

What other kind of constraints are you thinking of?

akshayjshah:
I was asking a much lower-rent question: is there a Go constant for the
string literal “CONFIG_DIR” that we should be using here? From grepping the
code, it looks like no.

PR #1561, opened 2022-11-03
push: error if response's pin is missing

twilly:
It's possible to succeed but have a missing LocalModulePin in the response. Check for this and report an error.

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -16,6 +16,7 @@ package push
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/bufbuild/buf/private/buf/bufcli"
@@ -182,6 +183,9 @@ func run(
 			return nil
 		}
 		return err
+	}
+	if localModulePin == nil {
+		return errors.New("Missing local module pin in the registry's response.")
 	}
 	if _, err := container.Stdout().Write([]byte(localModulePin.Commit + "\n")); err != nil {
 		return err


private/buf/cmd/buf/command/push/push.go
```diff
@@ -183,6 +184,9 @@ func run(
 		}
 		return err
 	}
+	if localModulePin == nil {
+		return errors.New("Missing local module pin in the registry's response.")
+	}
```
unmultimedio, line 189: Don't see a scenario in which `service.Push` does `return nil, nil`, also in theory that [shouldn't happen](https://bufprivate.slack.com/archives/C02KXVDB23B/p1665590362168519?thread_ts=1665589006.759369&cid=C02KXVDB23B)? not very idiomatic.

twilly: `service.Push` isn't Go. Well, it is but only on the surface.

It's code generated from `proto/buf/alpha/registry/v1alpha1/push.proto` and return values are whatever the remote implementation gives you, if anything, because `PushResponse.LocalModulePin` is optional.

jchadwick-buf (approved):

private/buf/cmd/buf/command/push/push.go
doriable: This one is a bit of an interesting one for me, I think defensively, our clients should handle that case. In practice, it should never be empty because a push should not be able to succeed without being able to respond with what the new commit is, and thus should be handled on the server-side (which is what we do now). That being said, this error would indicate to us if the server-side no longer handles that case, so I'm in favour of this.

doriable (approved):

private/buf/cmd/buf/command/push/push.go
twilly: Exactly. It's a pretty silly response since a pin is needed, but only because that's our understanding today. Perhaps a new field replaces it and this old client won't know what to do. :)

PR #1562, opened 2022-11-03
bufcli: support HTTP registries

twilly:
Plain HTTP is useful for testing but the registry provider always used HTTPS even when the TLS config was nil.


doriable (approved):

jchadwick-buf (approved):
LGTM, seems to work.

PR #1563, opened 2022-11-03
bufcli: BucketAndConfigForSource

twilly:
BucketAndConfigForSource assumes ReadModuleWithWorkspacesDisabled's fetching and validating a module's content into a bucket.

Renames ReadModuleWithWorkspacesDisabled to ReadModule as workspaces are no longer involved. ReadModule becomes shorthand for constructing a builder and building a module.


doriable (approved):
This looks good to me. I also checked to make sure this wasn't being used elsewhere :)

PR #1564, opened 2022-11-03
push: add testing

twilly:
Push never had testing before. Now it does!

### DIFF
--- a/private/buf/cmd/buf/command/push/push_test.go
+++ b/private/buf/cmd/buf/command/push/push_test.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"path"
 	"strings"
+	"sync"
 	"testing"
 
 	"github.com/bufbuild/buf/private/buf/cmd/buf/internal/internaltesting"
@@ -42,6 +43,7 @@ import (
 
 type mockPushService struct {
 	t         *testing.T
+	lock      sync.Mutex
 	callbacks map[string]func(*registryv1alpha1.PushRequest)
 	called    map[string]struct{}
 	resp      map[string]*registryv1alpha1.PushResponse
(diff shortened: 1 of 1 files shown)


twilly:
Pending landing of  #1560,  #1561, and #1562.

private/buf/cmd/buf/command/push/push_test.go
```diff
@@ -0,0 +1,259 @@
...
+// Push pushes.
+func (m *mockPushService) Push(
+	_ context.Context,
+	req *connect_go.Request[registryv1alpha1.PushRequest],
+) (*connect_go.Response[registryv1alpha1.PushResponse], error) {
+	owner := req.Msg.Owner
+	cb, ok := m.callbacks[owner]
+	if ok {
+		cb(req.Msg)
+		m.called[owner] = struct{}{}
```
joshcarp, line 70: This fails `make racetest` here. 

Idk why it's not enabled on ci

private/buf/cmd/buf/command/push/push_test.go
```diff
@@ -0,0 +1,259 @@
...
+					func(env map[string]string) map[string]string {
+						env["BUF_TOKEN"] = "invalid"
+						buftransport.SetDisableAPISubdomain(env)
+						injectConfig(t, appName, env)
+						return env
+					},
+				)(appName),
+				tarball(map[string][]byte{
+					"buf.yaml":  bufYAML(t, URL, owner, "repo"),
+					"foo.proto": nil,
```
joshcarp, line 176: what does it mean for a file to have `nil` contents in a tarball?

twilly: `nil` is the empty slice, hence the same as `[]byte{}`, and means an empty file. "foo.proto" and "bar.proto" are both empty files, which is fine for the tests written here. That may change if I get to testing when two proto files do not have identical content.

joshcarp (approved):
LGTM ✅

private/buf/cmd/buf/command/push/push_test.go
```diff
@@ -0,0 +1,269 @@
...
+) {
+	t.Helper()
+	owner := strings.ReplaceAll(desc, " ", "_")
+	mock.respond(owner, resp)
+	mock.callback(owner, func(req *registryv1alpha1.PushRequest) {
+		assert.NotNil(t, req.Module, "missing module")
+	})
+	t.Run(desc, func(t *testing.T) {
+		t.Parallel()
+		appName := "test"
```
unmultimedio, line 171: This can be const

PR #1573, opened 2022-11-08
add proto definitions for new GetSchema and Reformat endpoints

jhump:
Resolves TCN-720

Resolves TCN-728


proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  string owner = 1;
+  // A repository or module name. Required.
+  string repository = 2;
+
+  // A reference to a version. Optional. If present, it may refer to either a
+  // tag or a commit.
+  //
+  // If absent, the latest version for the module identified by the above
+  // owner and repository fields is assumed.
+  string reference = 3;
```
akshayjshah, line 45: @smallsamantha Are the IDs for drafts also called "commits" or did we use a different term?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+message GetSchemaRequest {
+  // The module from which the schema is retrieved. Required.
+  Module module = 1;
+
+  // A version in the form of a commit ID (may not be a tag or other reference).
+  // If present, the response should only include a schema if the requested
+  // version resolves to something other than this commit. This can be used
+  // to detect newer versions, when the reference in id is empty/absent, or
+  // to detect changes to a tag.
+  string only_if_not_version = 2;
```
akshayjshah, line 57: Can we defer this, pending a decision on whether we want to expose these APIs via HTTP GET and standard caching? It seems like we're basically recreating the semantics of etags and If-None-Match.

jhump: >  It seems like we're basically recreating the semantics of etags and If-None-Match.

Indeed we are. FWIW, it's very low friction to define it this way now (since proto/connect is the well-trodden path for exposing stuff from BSR). The presence of this doesn't really complicate/change the actual server implementation logic or mean we're doing something incompatible with an HTTP GET future. After all, a server handler that supports an HTTP GET endpoint would still need basically this same metadata to actually know when to send back a 304.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: If there is better vernacular to use for these fields, I'd love to hear it. In particular, I'd like to make sure we are using the same terms here as BSR public docs might be using, to eliminate confusion in the API over terminology.

But I used these field names because they are consistent with how we refer to a version of a module today in [these](https://github.com/bufbuild/buf/blob/4b1c53e9b38ee78dbacbf3b1dd3810eab7f0de95/proto/buf/alpha/registry/v1alpha1/image.proto#L47-L49) [three](https://github.com/bufbuild/buf/blob/4b1c53e9b38ee78dbacbf3b1dd3810eab7f0de95/proto/buf/alpha/registry/v1alpha1/module.proto#L25-L28) [places](https://github.com/bufbuild/buf/blob/4b1c53e9b38ee78dbacbf3b1dd3810eab7f0de95/proto/buf/alpha/module/v1alpha1/module.proto#L57-L60).

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: So instead of "recreating the semantics of...", I view this as really _implementing_ those semantics.

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+// Provides endpoints for downloading dynamic schemas and for using schemas
+// in validation and data transformation functions.
+service SchemaService {
+  // GetSchema allows the caller to download a schema for one or more requested
+  // types, RPC services, or RPC methods.
+  rpc GetSchema(GetSchemaRequest) returns (GetSchemaResponse);
+
+  // Reformat allows the caller to indicate a type schema which will be used to
+  // transform given data to a different format.
+  rpc Reformat(ReformatRequest) returns (ReformatResponse);
```
elliotmjackson, line 30: Is `reformat` the best we can do? also, reformat what? it's a `SchemaService` so I'd be forgiven to assume its reformatting my schema. Following a verb+noun pattern...

**Verb**
- Translate
- Reformat
- Render
- Convert
- Decode
- Decipher

**Noun**
- Data
- Payload
- Message

I like `translateMessage` if we are trying to differ from the existing `convert` endpoint but otherwise `convert` is an equally appropriate verb.

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // matched that version, this will be absent, indicating that the resolved version
+  // has not changed.
+  google.protobuf.FileDescriptorSet schema_files = 2;
+}
+
+message ReformatRequest {
+  // The module which contains the schema of interest. Required.
+  Module module = 1;
+  // The message type of the input data. Required.
+  string message_name = 2;
```
elliotmjackson, line 122: comment and attribute use message name and message type interchangeably

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+
+  // The input data that is to be reformatted. Required.
+  bytes input_data = 3;
```
elliotmjackson, line 125: the attributes in this message become troublesome when we add a noun to the message name. The input bytes should match the chosen noun. ie `input_message`

elliotmjackson (commented):
I'd like to gather a greater understanding of `message` and how its used in the proto world. We refer to it like a first-class noun in the `ReformatRequest` - is this a widely adopted term?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // Callers typically use compiled-in versions of these files.
+  bool include_well_known_imports = 6;
+
+  // If true, the returned schema will be expanded to include custom option
+  // definitions for options present on the requested schema elements.
+  bool include_custom_options = 7;
+
+  // If true, the returned schema will include all known extensions for any
+  // extendable messages among the requested schema elements.
+  bool include_extensions = 8;
```
elliotmjackson, line 97: Devils advocate discussion: are these options necessary for an MVP?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+}
```
robbertvanginkel, line 98: The GetImage version of this endpoint that serves trimmed filedescriptors has 
https://github.com/bufbuild/buf/blob/4b1c53e9b38ee78dbacbf3b1dd3810eab7f0de95/proto/buf/alpha/registry/v1alpha1/image.proto#L54-L55

which is required to be off when using `types` (in this pr `element_names`). Should this endpoint have such an option too, or are we going to fixup the partial image generation to have support for synthetic sourcecodeinfo?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // any requested types (e.g. requested message types or request and response types
+  // for requested service and method names).
+  repeated string element_names = 5;
+
+  // If true, well-known imports will be included the returned set of files. If
+  // false or not present, these files will omitted from the response (e.g. the
+  // caller is assumed to already know the types they define).
+  //
+  // These files are a subset of the files in the "well-known types" module:
+  // https://buf.build/beta/wellknowntypes
```
robbertvanginkel, line 74: What's the rationale behind the subset? I think its only missing plugin.proto, which is probably not that useful but feels similar to returning descriptor.proto?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // tag or a commit.
+  //
+  // If absent, the latest version for the module identified by the above
+  // owner and repository fields is assumed.
+  string reference = 3;
+}
+
+message GetSchemaRequest {
+  // The module from which the schema is retrieved. Required.
+  Module module = 1;
```
robbertvanginkel, line 50: No other part of the API refers to a remote reference like this, the current standard for apis that need to fetch info related to a ref is that all three are individual toplevel fields: https://github.com/bufbuild/buf/blob/4b1c53e9b38ee78dbacbf3b1dd3810eab7f0de95/proto/buf/alpha/registry/v1alpha1/image.proto#L47-L49

There is a type like this in module.proto, but 

https://github.com/bufbuild/buf/blob/014e59c1d9aa7e6f5a9f8c0d8451d7d1d5283d32/proto/buf/alpha/module/v1alpha1/module.proto#L58-L65

why the api does not use it for requests is a decision made before me that I don't have good context on. I'm not sure we should break that convention in the `registry.v1alpha1` package.

proto/buf/alpha/registry/v1alpha1/schema.proto
robbertvanginkel: We have Convert in https://github.com/bufbuild/buf/blob/936c01c610632585a71c2b3a4cf4cea6a0db42b9/proto/buf/alpha/registry/v1alpha1/convert.proto#L28-L33, which I think is what @jhump is trying to avoid a conflict with. I don't think that api is really used since it was put out, and we may be able deprecate/remove that one.

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // These include all well-known types that have custom JSON formats and
+  // the types in "descriptor.proto", required for Protobuf reflection.
+  // Callers typically use compiled-in versions of these files.
+  bool include_well_known_imports = 6;
+
+  // If true, the returned schema will be expanded to include custom option
+  // definitions for options present on the requested schema elements.
+  bool include_custom_options = 7;
+
+  // If true, the returned schema will include all known extensions for any
```
robbertvanginkel, line 95: nit: since custom options are techinically extensions, does specifying true here get you all all options too? See https://github.com/bufbuild/buf/pull/885/files/d2251412e040eefd18a80539556f83e89c0cef60#r796174256 for some context, where Alex and I previously thought probably not.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: Probably not. To some degree I was trying to stake out the full interface, but we could implement it piecemeal. But it's fair to simply omit all of these. I can add a TODO comment to make reference to these additional features we expect to implement.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: The returned schema will not have source code info, which is why this is not an option. No, definitely do not want to try to fixup the sourceodeinfo -- that is non-trivial work, and the benefit is unclear. There may be some user agents doing dynamic RPC that want to show comments for elements. But I think that is something we could work in as a potential future feature, and we'll leave it out for now.

proto/buf/alpha/registry/v1alpha1/schema.proto
elliotmjackson: I have made TCN-735 and am looking into this now.

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -0,0 +1,196 @@
...
+  // FORMAT_BINARY, unrecognized fields can be retained and possibly
+  // included in the reformatted output (depending on the requested
+  // output format).
+  bool discard_unknown = 5;
+
+  oneof output_format {
+    BinaryOutputOptions output_binary = 6;
+    JSONOutputOptions output_json = 7;
+    TextOutputOptions output_text = 8;
+  }
```
jhump, line 141:
```
Failure: plugin go-api: plugin does not allow oneofs for request fields: OutputBinary
```
😭 🤮 

What can I do to work around this? This seems like a silly restriction. I know this is for our "de-structured" API clients... but I really don't like those anyway. Is there some way to just skip this file or service for that generator?

Not sure who to ask. Maybe @robbertvanginkel ?

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: To me `input_message` and `message_name` seem ambiguously named. The `message_name` is what selects the schema for processing the input data.

The other convert endpoint uses the word "payload". Is that better? Maybe `input_payload` (and `output_payload` in the response)?

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: Changed this to `ConvertMessage`.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: I've removed these last two and replaced with a TODO. I also made a note in a Linear ticket that MVP does not need to include selective inclusion of options/extensions.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: I saw stuff in `resolve.proto` make use of `LocalModuleReference`, so I've changed to re-use that type. Does that seem better?

proto/buf/alpha/registry/v1alpha1/schema.proto
robbertvanginkel: BSR-336 has a bit more details on this, I looked into supporting it once but its a disproportionate amount of work. Probably the quickest way to exempt it is in by https://github.com/bufbuild/buf/blob/e1cecca043cc170cd8e2c04bbcfb013f6ec54c78/make/buf/all.mk#L118 (although I think the cli only supports `--path` for includes?)

proto/buf/alpha/registry/v1alpha1/schema.proto
elliotmjackson: hmm i see what you mean...
```diff
-  bytes input_data = 3;
+  bytes message_data = 3;
```
 which then falls into the next attribute 
```diff
-  Format input_format = 4;
+  Format message_format = 4;
```
so if it looked like 
```proto
message ConvertMessageRequest {
  Module module = 1;
  string name = 2;
  bytes data = 3;
  Format format = 4;
...
}
```
do we hate it?

jhump: I think `name`, without any qualification`, is a bit vague. I prefer the mild redundancy of having "message" in the method name but also having it in the field name. I prefer keeping `input_` prefixes for fields that define the input, mainly to distinguish them and provide symmetry with analogous elements related to the output. There is, for example, also an `output_format` (in the request, though it's a oneof instead of a field) and an `output_data` (in the response).

Do you think the names as currently shown in the PR are unclear? Or are they inconsistent with naming of these same concepts in other places (other APIs, or docs, etc)?

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: Just promoted the link above to a draft PR: #1578.

But until all of that extra code gen and stuff is ripped out, I think the (admittedly kinda gross) workaround I pushed to this branch is okay. WDYT?

jhump:
@akshayjshah, @robbertvanginkel, @elliotmjackson: circling back. I tried to respond to some comments, and I made a few changes. If objections about parts of this persist, maybe we can schedule a Zoom to chat about it? Please take a look and let me know.

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: Removed for now. But I'd like to revisit this soon so this RPC can offer an efficient way for clients to poll for the latest without having to repeatedly re-download the schema when it hasn't changed.

jhump:
@robbertvanginkel , @elliotmjackson , I've got this merged up to latest main. I reverted the stuff related to code-gen, fighting with the `protoc-gen-go-api` plugin (and friends), since they are now all gone (💥).

For now, I've removed the field that @akshayjshah objected to (though I think we'll want to add this back pretty soon to have an efficient polling technique).

I've left these in `buf/alpha` for now since this really is an alpha endpoint. We can rearrange it with a stable layout and version number hopefully in short order.

Please take another look. As soon as this is merged, we can start implementing the endpoints in the BSR.

elliotmjackson (approved):

akshayjshah:
Works for me!

PR #1574, opened 2022-11-09
Promote buf convert to stable

joshcarp:
Promotes `buf beta convert` to stable because it's been out in beta for some time and there haven't been any bugs reported.
Keeps `buf beta convert` so that anyone who's using it may continue until we decide to remove it from beta.


private/buf/cmd/buf/buf.go
```diff
@@ -140,7 +141,7 @@ func NewRootCommand(name string) *appcmd.Command {
 				Use:   "beta",
 				Short: "Beta commands. Unstable and likely to change.",
 				SubCommands: []*appcmd.Command{
-					convert.NewCommand("convert", builder),
+					//convert.NewCommand("convert", builder),
```
elliotmjackson, line 144:
```suggestion
```

elliotmjackson, line 144:
```suggestion
```

elliotmjackson (changes requested):

private/buf/cmd/buf/buf.go
elliotmjackson: or do you want to uncomment it?

twilly (approved):
:shipit:

elliotmjackson (approved):

PR #1575, opened 2022-11-09
Add implicitly_deprecated field to the doc proto

seankimdev:
- This proto is only used internally for rendering the bufdoc. ~The existing `file_deprecated` field has not been used anywhere yet, therefore should be not affected by this change.~ `file_deprecated` is currently exposed in the [`GetPackageDocumentation`](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.DocService.GetPackageDocumentation) API. This PR should be merged together with the PR in core for updating the API.
- This diff replaces them with `implicitly_deprecated` fields, which will be used for the API to provide if the element is implicitly deprecated in the following cases:
    1. if file is deprecated, all elements within are deprecated 
    2. if enclosing message is deprecated, then enclosed message or enum is deprecated
    3. if service is deprecated, all rpcs within the service are deprecated
- This field is needed separately from the `deprecated` field in the `{element_name}Options` messages, which is used for providing explicit deprecation status for rendering options in the source code style.


proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -387,6 +389,6 @@ message FileExtension {
   Location location = 4;
   // fields are all the fields that are associated with the extension.
   repeated Field fields = 5;
-  // file_deprecated is the deprecation status of the file containing the field.
-  bool file_deprecated = 6;
+  // implicitly_deprecated is true if its enclosing file or parent element is deprecated.
```
pkwarren, line 392: In this case, I don't think extensions can have parent elements so I'd tweak the comment a bit here.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -184,8 +184,8 @@ message Service {
   Location location = 6;
   repeated Method methods = 7;
   ServiceOptions service_options = 8;
-  // file_deprecated is the deprecation status of the file containing the service.
-  bool file_deprecated = 9;
+  // implicitly_deprecated is true if its enclosing file or parent element is deprecated.
```
pkwarren, line 187: Services too can't be enclosed in a parent element that I'm aware of so I'd tweak this comment to only mention enclosing file.

pkwarren (approved):

doriable (approved):

proto/buf/alpha/registry/v1alpha1/doc.proto
timostamm: They can:

```proto
syntax = "proto3";
option (Foo.custom_file_option) = true;
import "google/protobuf/descriptor.proto";
message Foo {
    extend google.protobuf.FileOptions {
        bool custom_file_option = 123456;
    }
}
```

seankimdev: 😬 ah good catch... Thanks Timo.

seankimdev:
Correction: I just noticed that the `PackageDocumentation` API already included the `file_deprecated`, although it's not being used by our client. I will put this PR on hold from merging until the API changes in core is ready, so that it can follow immediately.

PR #1577, opened 2022-11-09
Add more export description

joshcarp:
Updates `buf export` command description to have more detail.


private/buf/cmd/buf/command/export/export.go
```diff
@@ -56,8 +56,27 @@ func NewCommand(
 	return &appcmd.Command{
 		Use:   name + " <input>",
 		Short: "Export the files from the input location to an output location.",
-		Long:  bufcli.GetInputLong(`the source or module to export`),
-		Args:  cobra.MaximumNArgs(1),
+		Long: bufcli.GetInputLong(`the source or module to export`) + `
+
+Examples:
```
marekbuild, line 61: These examples are nice. I think in Slack you mentioned that the output .proto files will not be an exact replica of the original.

Could we add a description in regard to that? Somehow list out or describe what is going to be missing or different?

joshcarp: The output .proto in this case will be an exact replica because the command doesn't allow for image inputs, only sources. In this case the output will be parity with the input proto in all cases

marekbuild: Even in this example:

```
buf export buf.build/<owner>/<repo> --output=<output-dir>
```

Where the information is coming from a repo (aka: from latest version of the module in that repo)?

marekbuild: So I'm totally good with this update. Looks great.

I'm actually asking about how can we "perfectly" reproduce the .proto files from a BSR module, when the BSR does not actually store the raw original files.

I though the BSR does some "transforming" before storage, and in that transform can loose some fidelity, is that right?

marekbuild (approved):
The docs are better with these examples, so this PR is a win.

PR #1578, opened 2022-11-10
Introduce path to start using plain Connect client stubs

jhump:
This will help us address the client-side of BSR-336.

The `connect.ClientConfig` and the generic `connect.MakeClient` function serve the previous role of `Provider`, in terms of making sure the client stubs are configured consistently. But they allow us to do so without extra code generation and to just leverage the code generated by `protoc-gen-connect-go`.

This change allows us to port over call sites to plain stubs piecemeal, one package at a time. As an example of how the new stuff is used, see the second commit, which ports of the `bufgen` package.

Once all packages are ported over (including usages from our internal core repo), we can completely remove the `protoc-gen-go-apiclient` and `protoc-gen-go-apiclientconnect` plugins and all associated generated code. At that point, the code in `bufcli` that currently configures the provider will be changed to instead just configure a client config.

We will still need the `protoc-gen-go-api` plugin, however, until the server side is similarly overhauled.


private/bufpkg/bufprotoplugin/cmd/protoc-gen-go-apiclientconnect/main.go
```diff
@@ -282,111 +305,3 @@ func generateServiceFile(helper protogenutil.NamedHelper, plugin *protogen.Plugi
 	}
 	return nil
 }
-
-func handleGlobal(helper protogenutil.NamedHelper, plugin *protogen.Plugin, goPackageFileSets []*protogenutil.GoPackageFileSet) error {
```
jhump, line 286: This stuff was completely unused AFAICT. So I decided to go ahead and remove it.

private/gen/proto/apiclientconnect/apiclientconnect.pb.go
```diff
@@ -1,85 +0,0 @@
...
-	return p.bufAlphaAuditV1alpha1Provider
-}
-
-func (p *provider) BufAlphaRegistryV1alpha1() registryv1alpha1apiclient.Provider {
-	return p.bufAlphaRegistryV1alpha1Provider
-}
-
-func (p *provider) BufAlphaWebhookV1alpha1() webhookv1alpha1apiclient.Provider {
-	return p.bufAlphaWebhookV1alpha1Provider
-}
```
robbertvanginkel, line 79: 👍

robbertvanginkel (approved):
This seems like a reasonable way forward!

private/pkg/connect/config.go
```diff
@@ -0,0 +1,75 @@
...
+	authInterceptorProvider func(string) connect_go.UnaryInterceptorFunc
+}
+
+func NewClientConfig(httpClient connect_go.HTTPClient, options ...ClientConfigOption) *ClientConfig {
+	var cfg ClientConfig
+	for _, opt := range options {
+		opt(&cfg)
+	}
+	return &cfg
+}
```
elliotmjackson, line 34:
```suggestion
func NewClientConfig(httpClient connect_go.HTTPClient, options ...ClientConfigOption) *ClientConfig {
	cfg  := ClientConfig{
		httpClient: httpClient,
	}
	for _, opt := range options {
		opt(&cfg)
	}
	return &cfg
}
```

elliotmjackson: httpClient appears to be ignored and then used in `MakeClient` later.

jhump: 🤦, good catch.

Indeed, nothing in this repo's tests actually exercises remote generation in a way that would have caught this. But I just tried it out and it panics.

jhump:
@elliotmjackson, @robbertvanginkel, after going through a few other packages, updating things to use this new client config instead of the provider, I thought of another change I think I want to make and want to get your input.

The `pkg/connect` is unfortunately named since it collides with the official `connect-go` package. I had thought the official package's name was actually `connect_go` since I had seen that in a few source files. But it turns out that was just coincidence. So as I update other callers, I ended up having to introduce an alias for the official connect package.

Maybe I should name it `pkg/connectclient` and name the functions/types `Config` and `Make`? So the config type changes from `connect.ClientConfig` -> `connectclient.Config`, and the factory func changes from `connect.MakeClient` to `connectclient.Make`. WDYT?

private/pkg/connectclient/connectclient.go
```diff
@@ -0,0 +1,80 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package connectclient
+
+import (
+	connect_go "github.com/bufbuild/connect-go"
```
robbertvanginkel, line 18: nit: could remove the alias now

robbertvanginkel (approved):

private/pkg/connectclient/connectclient.go
jhump: Yep, good call. Done.

PR #1582, opened 2022-11-10
remove default_opts from buf.plugin.yaml

pkwarren:
After using buf.plugin.yaml for a bit, we discovered that 'default_opts' had more potential to confuse users than make their lives easier. It also was unclear how a user could reset options to no options if a plugin used 'default_opts'.

As a result, we've migrated to remove all use of 'default_opts' so a plugin behaves the same when executed locally or remotely in buf.gen.yaml. We also make use of the 'opts' field under 'registry:' to set appropriate options for generated code by the Go module proxy or NPM registries.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -135,8 +132,7 @@ message CuratedPlugin {
   // the algorithm.
   // Ref: https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests
   string container_image_digest = 6;
-  // List of default plugin options in the form "{option}={defaultOption}"
-  repeated string default_options = 7;
+  reserved 7;
```
mfridman, line 135: I'm okay removing this entirely to avoid confusion. This is an unreleased feature and the only consumer is us.

If you do keep this, then we typically also include the field name, e.g.,

https://github.com/bufbuild/buf/blob/2b2aa72a9f3c8b73c807a650c556d8ec49c4c347/proto/buf/alpha/module/v1alpha1/module.proto#L69-L70

pkwarren: I'll just remove it - I didn't want to reshuffle all the field ids and just reserved the id to make it clear it could be used in a follow up if we wanted, but didn't reserve the JSON field name b/c I didn't think we'd rely on JSON serialization anywhere. Probably overthinking it.

mfridman (approved):
I'd still suggest reordering the fields, but otherwise this lgtm!

PR #1583, opened 2022-11-10
[bsr-721] Reader only interface

unmultimedio:
Related to https://github.com/bufbuild/buf/pull/1563

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -744,15 +744,12 @@ func ReadModule(
 	ctx context.Context,
 	logger *zap.Logger,
 	sourceBucket storage.ReadBucket,
+	sourceConfig *bufconfig.Config,
 ) (bufmodule.Module, error) {
-	config, err := bufconfig.GetConfigForBucket(ctx, sourceBucket)
-	if err != nil {
-		return nil, err
-	}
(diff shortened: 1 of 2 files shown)


private/buf/bufcli/bufcli.go
```diff
@@ -743,13 +743,16 @@ func BucketAndConfigForSource(
 func ReadModule(
 	ctx context.Context,
 	logger *zap.Logger,
-	sourceBucket storage.ReadBucketCloser,
-	sourceConfig *bufconfig.Config,
+	sourceBucket storage.ReadBucket,
```
unmultimedio, line 746: No need for a `ReaderCloser`, since `BuildForBucket` receives a `Reader` only.

private/buf/bufcli/bufcli.go
```diff
@@ -743,13 +743,16 @@ func BucketAndConfigForSource(
...
 ) (bufmodule.Module, error) {
+	config, err := bufconfig.GetConfigForBucket(ctx, sourceBucket)
+	if err != nil {
+		return nil, err
+	}
```
unmultimedio, line 751: Figured no need to receive a config, if it always comes from the bucket anyway?

twilly (approved):

PR #1584, opened 2022-11-10
[bsr-721] Generate a new blobs array per test

unmultimedio:
Fixes [this failure](https://github.com/bufbuild/buf/actions/runs/3440666825/jobs/5739385649#step:5:212). It is an `append` issue :facepalm: ... TIL that [sometimes `append` modifies the underlying array in the first param](https://pkg.go.dev/builtin#append), so `t.Parallel()` was messing the shared blobs state.

> The append built-in function appends elements to the end of a slice. **If it has sufficient capacity, the destination is resliced to accommodate the new elements.** If it does not, a new underlying array will be allocated. Append returns the updated slice.

### DIFF
--- a/private/pkg/manifest/module_test.go
+++ b/private/pkg/manifest/module_test.go
@@ -137,6 +137,7 @@ func TestNewBlobValidDuplicates(t *testing.T) {
 }
 
 func TestNewBlobInvalidDuplicates(t *testing.T) {
+	t.Parallel()
 	blobs := newBlobsArray(t)
 	incorrectBlob, err := manifest.NewMemoryBlob(
 		*blobs[0].Digest(),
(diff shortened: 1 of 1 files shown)


private/pkg/manifest/module_test.go
```diff
@@ -115,58 +115,43 @@ func TestInvalidNewDigestFromBlobHash(t *testing.T) {
...
-				duplicatedBlobs,
-				manifest.BlobSetWithContentValidation(),
-			)
-			require.NoError(t, err)
-			assert.NotNil(t, blobSet)
-		})
+func TestNewBlobValidDuplicates(t *testing.T) {
+	t.Parallel()
+	blobs := newBlobsArray(t)
+	duplicatedBlobs := append(blobs, blobs[0])
```
robbertvanginkel, line 131: nit: could re-use `blobs` as name now, whenever you see `x = append(<not x>, ...)` it is more potential overhead than the two separate variable names are helpful imo.

robbertvanginkel (approved):

private/pkg/manifest/module_test.go
unmultimedio: you're right, and actually this var is not explicitly needed, let me fix

twilly:
👍

PR #1585, opened 2022-11-11
Read manifest and blobs from bucket

unmultimedio:
Logic used in push command to go from a bucket to manifest and files blob is also useful in the backend. Moving it to the `manifest` pkg.

### DIFF
--- a/private/pkg/manifest/manifest.go
+++ b/private/pkg/manifest/manifest.go
@@ -194,6 +194,15 @@ func (m *Manifest) UnmarshalText(text []byte) error {
 	return nil
 }
 
+// Blob returns the manifest's blob.
+func (m *Manifest) Blob() (Blob, error) {
+	manifestText, err := m.MarshalText()
+	if err != nil {
+		return nil, err
+	}
+	return NewMemoryBlobFromReader(bytes.NewReader(manifestText))
+}
+
 func splitManifest(data []byte, atEOF bool) (int, []byte, error) {
 	// Return a line without LF.
 	if i := bytes.IndexByte(data, '\n'); i >= 0 {
--- a/private/pkg/manifest/module.go
+++ b/private/pkg/manifest/module.go
@@ -38,7 +38,6 @@ type Blob interface {
 	Digest() *Digest
 	Open(context.Context) (io.ReadCloser, error)
 	EqualContent(ctx context.Context, other Blob) (bool, error)
-	AsProtoBlob() (*modulev1alpha1.Blob, error)
 }
 
 type memoryBlob struct {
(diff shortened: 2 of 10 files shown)


private/pkg/manifest/manifest.go
```diff
@@ -194,6 +194,15 @@ func (m *Manifest) UnmarshalText(text []byte) error {
...
 
+// Blob returns the manifest's blob.
+func (m *Manifest) Blob() (Blob, error) {
+	manifestText, err := m.MarshalText()
+	if err != nil {
+		return nil, err
+	}
+	return NewMemoryBlobFromReader(bytes.NewReader(manifestText))
+}
+
```
twilly, line 205: 👏

private/pkg/manifest/module.go
```diff
@@ -166,6 +187,15 @@ func (s *BlobSet) BlobFor(digest string) (Blob, bool) {
 	return blob, true
 }
 
+// Blobs returns the blobs in the set as an array.
+func (s *BlobSet) Blobs() []Blob {
+	var blobs []Blob
```
twilly, line 192:
```suggestion
	blobs := make([]Blob, len(s.digestToBlob))
```
... should reduce memory allocations.

private/pkg/manifest/module.go
```diff
@@ -107,6 +113,21 @@ func (b *memoryBlob) EqualContent(ctx context.Context, other Blob) (_ bool, retE
...
+	}
+	return &modulev1alpha1.Blob{
+		Hash: &modulev1alpha1.Hash{
+			Kind:   hashKind,
+			Digest: b.digest.Bytes(),
+		},
+		Content: b.content,
+	}, nil
+}
+
```
twilly, line 130: I'm a little weary about adding mappers for an internal type. It ties it to a representation it doesn't necessarily care about. Imagine what a "fileBlob" or a "GCPBlob" implementation would be; they all would look quite similar.

```go
func AsProtoBlob(ctx contentex.Context, blob Blob) (*modulev1alpha1.Blob, error) {
	file, err := blob.Open(ctx)
	if err != nil {
		return nil, err
	}
	content, err := io.ReadAll(file)
	if err != nil {
		return nil, err
	}
	blob := &modulev1alpha1.Blob{
		Hash: &modulev1alpha1.Hash{
			Kind:   hashKind,
			Digest: blob.Digest().Bytes(),
		},
		Content: content,
	}
	if err := file.Close(); err != nil {
		return nil, err
	}
	return blob, nil
}
```

unmultimedio: thx! done

twilly (approved):
:shipit:

PR #1587, opened 2022-11-11
Return digest value

unmultimedio:


private/pkg/manifest/manifest_test.go
```diff
@@ -83,14 +83,14 @@ func TestAddEntry(t *testing.T) {
...
+	require.NoError(t, m.AddEntry(filePath, fileDigest)) // adding the same entry twice is fine
 
-	require.Error(t, m.AddEntry("", *fileDigest))
+	require.Error(t, m.AddEntry("", fileDigest))
 	require.Error(t, m.AddEntry("other/path", manifest.Digest{}))
-	require.Error(t, m.AddEntry(filePath, *mustDigestShake256(t, []byte("other content"))))
+	require.Error(t, m.AddEntry(filePath, mustDigestShake256(t, []byte("other content"))))
 
-	expect := fmt.Sprintf("%s  %s\n", fileDigest, filePath)
+	expect := fmt.Sprintf("%s  %s\n", fileDigest.String(), filePath)
```
unmultimedio, line 93: This is downside 1, is not an straight forward stringer.

private/pkg/manifest/digest_test.go
```diff
@@ -29,17 +29,18 @@ import (
 
 func TestNewDigestBytes(t *testing.T) {
 	t.Parallel()
+	nilDigest := mustDigestShake256(t, nil)
```
unmultimedio, line 32: This is downside 2, cannot chain to get `.Bytes()`, or `.Type()`, or similar.

private/pkg/manifest/module.go
```diff
@@ -35,7 +35,7 @@ var (
 
 // Blob is a blob with a digest and a content.
 type Blob interface {
-	Digest() *Digest
+	Digest() Digest
```
unmultimedio, line 38: This PR is just this change.

private/pkg/manifest/module.go
```diff
@@ -169,7 +170,8 @@ func NewBlobSet(ctx context.Context, blobs []Blob, opts ...BlobSetOption) (*Blob
 	}
 	digestToBlobs := make(map[string]Blob, len(blobs))
 	for _, b := range blobs {
-		digestStr := b.Digest().String()
+		digest := b.Digest()
+		digestStr := digest.String()
```
unmultimedio, line 174: From downside 1: this sometimes could be a bit verbose

unmultimedio (commented):
Only 2 downsides I see, the rest are wins imo.

private/pkg/manifest/manifest_test.go
unmultimedio: Although it could be fixed passing the pointer 🤔

private/pkg/manifest/module.go
twilly: And there's two implicit downsides here. One is Digests on the heap would be copied to stack. The other is a blob that doesn't have a digest must produce one. That latter is why chaining becomes possible: it requires blobs to have digests. Downside or upside is a matter of perspective with that one.

private/pkg/manifest/module.go
twilly: Digests has pointer receivers. If we really wanted to return Digest as values, I think we should be considering Digest values as immutable and define only value receivers. Mixing the two is causing headaches like these.

unmultimedio:
Converting to draft, wait to see if it's really needed.

unmultimedio:
Closing now, will reopen it if shows that pointers receivers will be a greater issue.

PR #1588, opened 2022-11-11
[BSR-701] add default except override to csharp namespace managed mode

oliversun09:
- add `csharp_namespace_prefix` to managed mode, allowing {default, except, override}
- `csharp_namespace_prefix` is optional, for backward compatibility


doriable (commented):
For C# namespace, we don't really need to define a default "prefix", since the outcome is that we want to set a namespace based on the package and proto file. We should be able to drop `default` entirely, and have the default behaviour, when managed mode is enabled to set the C# option. We want users to be able to exempt modules from this behaviour.

doriable (approved):
Looks good, thanks for implementing this!

felschr:
I think C# is a bit of a special case, because when importing from C# projects, directories & project names are irrelevant since imports are entirely evaluated via their namespaces.
In many other languages imports are either directory-based or module & package-based, so one can just put all the code generated by buf into a directory or a separate project directory. This essentially creates a scope around all generated code.

Let's assume a buf setup with lots of protobuf packages with different names, e.g. `apple.v1` & `banana.v1`.
Now, to import them into a C# project I have to use `using Apple.V1;`.
This can cause various naming conflicts, e.g. when there's already an existing `Apple` namespace in the importing project or a class called `Apple`.

I think an easy option to add a prefix would be desirable to avoid these kind of problems.

PR #1605, opened 2022-11-15
manifest: make Blob.EqualContent a pkg func

twilly:
Checking for blob equality doesn't really depend on the blob's implementation. It's a check if the digest and content are identical, both of which are exposed in the Digest and Open methods.

BlobEqual implements exactly this. It's a little tricky to read and compare content from two streams. They can throw errors at any point. Each read can return a different sized block. Reads can return an error with data. I hope I covered these edge cases and made sure to test them (100% coverage)!

### DIFF
--- a/private/pkg/manifest/module.go
+++ b/private/pkg/manifest/module.go
@@ -31,7 +31,6 @@ var (
 	digestTypeToHashKind = map[DigestType]modulev1alpha1.HashKind{
 		DigestTypeShake256: modulev1alpha1.HashKind_HASH_KIND_SHAKE256,
 	}
-	blockSize = 4098
 )
 
 // Blob is a blob with a digest and a content.
@@ -236,6 +235,7 @@ func NewMemoryBlobFromReaderWithDigester(content io.Reader, digester Digester) (
 // An error is returned if an unexpected I/O error occurred when opening,
 // reading, or closing either blob.
 func BlobEqual(ctx context.Context, a, b Blob) (_ bool, retErr error) {
+	const blockSize = 4098
 	if !a.Digest().Equal(*b.Digest()) {
 		// digests don't match
 		return false, nil
--- a/private/pkg/manifest/module.go
+++ b/private/pkg/manifest/module.go
@@ -31,13 +31,13 @@ var (
 	digestTypeToHashKind = map[DigestType]modulev1alpha1.HashKind{
 		DigestTypeShake256: modulev1alpha1.HashKind_HASH_KIND_SHAKE256,
 	}
+	blockSize = 4098
 )
 
 // Blob is a blob with a digest and a content.
 type Blob interface {
 	Digest() *Digest
 	Open(context.Context) (io.ReadCloser, error)
-	EqualContent(ctx context.Context, other Blob) (bool, error)
 }
 
 type memoryBlob struct {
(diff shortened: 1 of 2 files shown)


private/pkg/manifest/module.go
```diff
@@ -31,13 +31,13 @@ var (
 	digestTypeToHashKind = map[DigestType]modulev1alpha1.HashKind{
 		DigestTypeShake256: modulev1alpha1.HashKind_HASH_KIND_SHAKE256,
 	}
+	blockSize = 4098
```
unmultimedio, line 34: You can move this const to be within the func. [If it's a local usage it's ok to declare it in a smaller scope](https://www.notion.so/bufbuild/Go-Code-Standards-ff40382e6b104c1097a017f9565d54db#17b252823cd5404b83a50b82ed295c3d).

private/pkg/manifest/module.go
```diff
@@ -245,3 +229,71 @@ func NewMemoryBlobFromReaderWithDigester(content io.Reader, digester Digester) (
...
+			break
+		}
+	}
+	_, aErr := io.ReadAtLeast(aFile, aBlock, 1)
+	_, bErr := io.ReadAtLeast(bFile, bBlock, 1)
+	if aErr != io.EOF || bErr != io.EOF {
+		// a or b is longer, or we late errored
+		return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
+	}
+	return true, nil
```
unmultimedio, line 290: Couldn't we save the last check after the `for` loop, by checking both `aErr` and `bErr` are `io.EOF`?

```suggestion
		if aErr == io.EOF || bErr == io.EOF {
			// one of them is EOF...
			if aErr != bErr {
				// ...the other isn't
				return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
			}
			// ...the other too
			return true, nil
		}
	}
```

private/pkg/manifest/module_test.go
```diff
@@ -168,6 +171,208 @@ func TestBlobFromReader(t *testing.T) {
...
+		false,
+		false,
+	)
+	testBlobEqual(
+		t,
+		"mismatched equal-length content data with error",
+		iotest.DataErrReader(strings.NewReader("foo")),
+		iotest.DataErrReader(strings.NewReader("bar")),
+		false,
+		false,
```
unmultimedio, line 229: Don't fully understand why this one doesn't return error.

private/pkg/manifest/module.go
twilly: Unfortunately not because each blob has a choice of returning EOF with data or data with no error then zero data and EOF. That's what that final reads are forcing: try really hard at reading the next byte.

This is somewhat confusing since io.ReadAtLeast complicates things even more. I really should just be calling Read directly and making the equal return a positive evaluation.

private/pkg/manifest/module_test.go
twilly: DataErrReader forces strings.Reader to return n > 0 with error (io.EOF). io.EOF isn't an real error in the sense that it's a stream successfully terminated signal. That's why the data is different (false / does-not-equal) but the io.EOF from the paired read is absorbed (successfully reached end of both streams).

To help make sense, this test activates the bytes.Equal path (they don't) yet return no error because of nilEOF filtering.

```go
if !bytes.Equal(aBlock[:aN], bBlock[:bN]) {
	// Read content doesn't match. Don't forget to forward any errors
	// returned.
	return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
}
```

private/pkg/manifest/module.go
```diff
@@ -245,3 +228,73 @@ func NewMemoryBlobFromReaderWithDigester(content io.Reader, digester Digester) (
...
+		}
+	}
+	aN, aErr := aFile.Read(aBlock[:1])
+	bN, bErr := bFile.Read(bBlock[:1])
+	if aN == 0 && bN == 0 && aErr == io.EOF && bErr == io.EOF {
+		// a and b are at EOF with no more data for us
+		return true, nil
+	}
+	// either a or b are longer
+	return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
```
unmultimedio, line 291: Small nit: keep happy path to left.
```suggestion
	if aN != 0 || bN != 0 || aErr != io.EOF || bErr != io.EOF {
		// either a or b are longer
		return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
	}
	// a and b are at EOF with no more data for us
	return true, nil
```

unmultimedio (approved):

private/pkg/manifest/module.go
twilly: Normally I would put the positive case at the end of a func, but this long expression is easier to read in the positive form ("if all these things are true, we're good to go").

PR #1607, opened 2022-11-16
remove apiclient generators and generated code

jhump:
This will feel very good to merge! For now, it's probably best to wait on merging this until after BSR-866 is done, to make sure we don't have any other usages of these plugins outside of this repo.

Fixes BSR-798


elliotmjackson (approved):
−6,232 feels so good

PR #1615, opened 2022-11-19
buf format: preserve interior indentation in multi-line comments

### LINKED ISSUE #1612: buf format: changes in comments from 1.8.0 to 1.9.0
buf format behaves differently between 1.8.0 to 1.9.0. 

1.8.0

```
/*
Poll
    inputs:
        Instance ID

    outputs:
        Maybe a result
*/
```

became in 1.9.0

```
/*
   Poll
   inputs:
   Instance ID
   
   outputs:
   Maybe a result
*/
```

Comments should not be formated or is this behavior intended? This is a minor issue.

Thanks.

jhump:
The logic added in 1.9 was attempting to "fix" indentation for multi-line comments that were attached to an element that was moved. For example, if there's a trailing comment attached to a semicolon and the semicolon was moved as part of the re-format, the comment may need to be re-indented accordingly.

However, this re-indenting was too aggressive and turns out to be destructive for comments that have _intentional_ interior indentation.

The fix is to leave interior indentation alone by only re-indenting based on the _least_ indented line and leaving other lines in the comment indented relative to that. (Not sure if that sentence makes sense. Maybe reading the code will help clear it up?)

I'm not going to lie: I know the code in here (including some of the new code in this change) is hairy and feels brittle. Part of me really wants to just rewrite it all to be less complicated. But, alas, that's much easier said than done because the task at hand is complicated by its nature. (It would be simpler if this were less opinionated about line breaks and worked more like `gofmt`. But re-formatting line breaks -- sometimes joining multiple lines into one and sometimes breaking one line out across many -- and the associated possible movement of comments makes it kind of nasty.)

Fixes #1612 

Fixes TCN-778

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -20,6 +20,7 @@ import (
 	"io"
 	"sort"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/bufbuild/protocompile/ast"
@@ -128,6 +129,10 @@ func (f *formatter) Out() {
 // Indent writes the number of spaces associated
 // with the current level of indentation.
 func (f *formatter) Indent(nextNode ast.Node) {
+	// only indent at beginning of line
+	if f.lastWritten != '\n' {
+		return
+	}
 	indent := f.indent
 	if rn, ok := nextNode.(*ast.RuneNode); ok && indent > 0 {
 		if strings.ContainsRune("}])>", rn.Rune) {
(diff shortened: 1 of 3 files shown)


private/buf/bufformat/formatter.go
```diff
@@ -2023,11 +2016,99 @@ func (f *formatter) writeTrailingEndComments(comments ast.Comments) {
...
+	if strings.TrimSpace(s) == "*/" {
+		return 0, false
+	}
+	indent := 0
+	for _, r := range s {
+		switch r {
+		case ' ':
+			indent++
+		case '\t':
+			indent += 4
```
twilly, line 2103: Is this normal for proto? Not 8? It's at least some magic given common editors will prefix fill with tabs and then add spaces, but how many tabs and spaces to use depends on the configured tab width.

jhump: Just like you said: this depends on the editor (nothing to do with proto). I can change it to 8. Most editors I've used (like sublime, IntelliJ, and VSCode I think, too) use 4. But editors like vim or emacs will use tab stops with width 8. I guess it's probably less surprising to just use the standard 8, even if someone's editor is configured otherwise, so I'll change it.

jhump: Changed to eight. I also realized that incrementing by 8 is not correct since tab stops are at fixed intervals of 8, and not simply 8 sequential spaces. So I fixed that, too.

twilly (approved):
:shipit:

PR #1616, opened 2022-11-21
[BSR-581] Proto Definition for Set/List Recommended Resources (Plugins/Repos)

oliversun09:
Added proto definition for two new RPCs in `RecommendationService`.
* `ListRecommendedResources`: accepts empty request body, returns a list [`resource`s](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.Resource), each of which is either a [Repository](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.Repository) or a [Curated plugin](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.CuratedPlugin), in the order specified in `SetRecommnededResourcesRequest`. If some resources set by `SetRecommnededResources` were deleted or made private after the list was set, they will not be included in the response.
* `SetRecommnededResources`: accepts a list of `SetRecommendedResource`, each including an owner and a name, and returns an empty response. For each (owner, name) pair, there has to exist a repository or curated plugin with that name, and owned by that owner. It also needs to be public. This will set the list of recommended resources to this list of resources, in the same order.


proto/buf/alpha/registry/v1alpha1/recommendation.proto
```diff
@@ -37,12 +38,22 @@ message RecommendedTemplate {
   string template_id = 4;
 }
 
+// SetRecommendedResource is the information needed to configure a resource recommendation
+message SetRecommendedResource {
```
oliversun09, line 42: Do we want to call it resource or RepositoryOrCuratedPlugin?

cyinma (approved):
lgtm!

dimitropoulos (approved):
Tommy and Oliver and I had a good conversation about whether this should be separated by plugins vs repos, and landed that where it is now is ok.  I was confused because I was looking at an old mockup that made it look like it'd be (in places) exclusively showing repos (but not plugins) when in reality what we're thinking on is that it'd never be showing as a filtered list to the user.

PR #1625, opened 2022-11-24
BSR-704/default except override for optimize_for

oliversun09:
* Add default, except override to managed mode `optimize_for` option
* preserves backward compatibility with the current `optimize_for` option being string

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -496,13 +496,13 @@ func newModifier(
 		managedConfig.Override[bufimagemodify.CsharpNamespaceID],
 	)
 	modifier = bufimagemodify.Merge(modifier, csharpNamespaceModifier)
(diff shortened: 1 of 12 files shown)


private/buf/bufgen/generator.go
```diff
@@ -496,11 +496,13 @@ func newModifier(
...
 	)
 	modifier = bufimagemodify.Merge(modifier, csharpNamespaceModifier)
-	if managedConfig.OptimizeFor != nil {
+	if managedConfig.OptimizeForConfig != nil {
 		optimizeFor, err := bufimagemodify.OptimizeFor(
 			logger,
 			sweeper,
-			*managedConfig.OptimizeFor,
+			managedConfig.OptimizeForConfig.Default,
+			managedConfig.JavaPackagePrefix.Except,
```
doriable, line 504: This should be `OptimizeForConfig.Except`.

PR #1628, opened 2022-11-28
TCN-787 complete type filtering on buf generate

elliotmjackson:
Adds `types` attribute to the `buf.gen.yaml`:

```diff
version: v1
plugins:
  - name: go
    out: gen/go
+ types:
+    include:
+      - "buf.alpha.lint.v1.IDPaths"
```

and supply a `types` flag override to the cli:

```diff
Usage:
  buf generate <input> [flags]

Flags:
      --config string          The file or data to use for configuration.
      --disable-symlinks       Do not follow symlinks when reading sources or configuration from the local filesystem.
                               By default, symlinks are followed in this CLI, but never followed on the Buf Schema Registry.
      --error-format string    The format for build errors, printed to stderr. Must be one of [text,json,msvs,junit]. (default "text")
      --exclude-path strings   Exclude specific files or directories, for example "proto/a/a.proto" or "proto/a".
                               If specified multiple times, the union is taken.
  -h, --help                   help for generate
      --include-imports        Also generate all imports except for Well-Known Types.
      --include-wkt            Also generate Well-Known Types. Cannot be set without --include-imports.
  -o, --output string          The base directory to generate to. This is prepended to the out directories in the generation template. (default ".")
      --path strings           Limit to specific files or directories, for example "proto/a/a.proto" or "proto/a".
                               If specified multiple times, the union is taken.
      --template string        The generation template file or data to use. Must be in either YAML or JSON format.
+     --include-types strings     The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types.
```

resolves TCN-787 
resolves TCN-788


jhump (approved):
Wow, short and sweet. Glad that ended up being so succinct.

jhump:
This one is probably worth adding to the change log

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -248,6 +250,12 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		`The file or data to use for configuration.`,
 	)
+	flagSet.StringSliceVar(
```
bufdev, line 253: Is this what we want to do? We have plans to move more into the `buf.gen.yaml` into the future, this is going in the opposite direction. This feels like a major UI change

akshayjshah:
I _love_ that this is so short. I'd imagined this being part of our managed mode configuration - is that crazy?

jhump:
@akshayjshah, @bufdev, so should this be completely _not_ supported as command-line flag? Is there any precedent for things in the yaml being overridden via flag? If so, we could start this way and then add config to the yaml that can then be overridden via this flag. WDYT?

elliotmjackson:
@bufdev Going to bring your comment into this thread as i think its related.... That's a totally fair statement, shoving lists of types into the cli as flags will get messy under anything greater than very simple use. 

@akshayjshah
> I love that this is so short. I'd imagined this being part of our managed mode configuration - is that crazy?

Yes, I believe that is crazy. From my point of view, "Managed Mode" has 2 distinct characteristics:
1. It relates to file options
2. a consumer is opting into having buf apply a set of predetermined rules in a set-and-forget fashion

Type filtering doesn't share these characteristics IMO. I believe they are more likely elevated to be equal seniority with `plugins` like so:

```yaml
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/bufbuild/buf/gen/proto/go
types:
  - buf.alpha.module.v1alpha1.Module
plugins:
  - plugin: go 
    out: gen/proto/go
    opt: paths=source_relative
```

akshayjshah:
> It relates to file options

This is reasonable, but hasn't been my impression. I think of "managed mode" as all the Buf features that are effectively editing the protobuf schemas on the fly, and this feels squarely in that category. Put differently, it's all the stuff that has no equivalent in `protoc`.

As long as the type filters appear in the config somewhere though, I'm happy.

private/buf/bufcli/bufcli.go
```diff
@@ -63,7 +63,7 @@ import (
 
 const (
 	// Version is the CLI version of buf.
-	Version = "1.10.0-dev"
+	Version = "1.11.2-dev"
```
jhump, line 66: ?? Overly aggressive search-and-replace?

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -193,7 +193,6 @@ type flags struct {
 	IncludeWKT      bool
 	ExcludePaths    []string
 	DisableSymlinks bool
-	Types           []string
```
jhump: Would we not want to also provide a command-line flag, that can override what's in the config file?

jhump (approved):
Cool.

private/buf/bufcli/bufcli.go
elliotmjackson: a bit of end to end manual testing gone rogue 😆 

have converted back to draft

private/buf/cmd/buf/command/generate/generate.go
elliotmjackson: Maybe some literal interpretation of @bufdev comments above on my part
> We have plans to move more into the buf.gen.yaml into the future

"move" I guess could also mean "include"

private/buf/bufcli/bufcli.go
elliotmjackson: reverted in [731d00f](https://github.com/bufbuild/buf/pull/1628/commits/731d00f7abd8ddeb8d28c774669ac4ea6fa92d2d)

private/buf/cmd/buf/command/generate/generate.go
elliotmjackson: resolved in [731d00f](https://github.com/bufbuild/buf/pull/1628/commits/731d00f7abd8ddeb8d28c774669ac4ea6fa92d2d)

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -248,6 +250,12 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		`The file or data to use for configuration.`,
 	)
+	flagSet.StringSliceVar(
+		&f.Types,
+		"types",
+		nil,
+		"The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types.",
```
jhump, line 257: Need to point to types stanza in buf.gen.yaml and mention how this overrides.

jhump (approved):
Just one nit in flag doc. Otherwise LGTM

private/buf/cmd/buf/command/generate/generate.go
elliotmjackson:
```suggestion
		"The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types. Flag usage overrides buf.gen.yaml",
```

jhump (approved):
still lgtm

PR #1629, opened 2022-11-28
use gci for canonical grouping of imports

jhump:
The `goimports` tool does not collapse groups if one's IDE has split imports up. What it _does_ do is to break up a group if it contains both standard and custom imports (custom import is one with a dot in it, which includes all imports outside the Go standard lib) and then sort the groups.

This moves over to use `gci`, which _will_ collapse groups when re-grouping/re-ordering, instead of `goimports`. There's no other diff since there are no files in this repo that currently have un-grouped imports. But this change will make sure it stays that way.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,16 @@ linters-settings:
       - '^log\.'
       - '^print$'
       - '^println$'
+  gci:
+    # Section configuration to compare against.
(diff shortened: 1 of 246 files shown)


elliotmjackson (approved):

jhump:
Superceded by #1630

PR #1630, opened 2022-11-29
Enforce import grouping

akshayjshah:
Switch from `goimports` to `gci`, which actually enforces the import
grouping style required by Buf's style guide.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -44,10 +44,11 @@ linters:
     - exportloopref
     - forbidigo
     - forcetypeassert
+    - gci
     - gochecknoinits
     - gofmt
     - goheader
-    - goimports
+    # - goimports
     - gomodguard
     - goprintffuncname
     - gosec


bufdev:
As discussed in core PR, enforcement without automation of import sorting via make generate or typical IDE setups means more time for our devs on a consistent basis, with little to no benefit to us over removing this item from the style guide

jhump:
LOL, I actually did this _before_ opening a PR on core: #1629

jhump:
> without automation of import sorting via make generate or typical IDE setups means more time for our devs on a consistent basis

I don't see how this issue occurs on a consistent basis. It's a one-time setup to the IDE, and devs can instead easily run `golangci run --fix` ad hoc. The fact that nearly all imports are already correct (zero diff in repos other than core) implies that most people already have the right setup.

Just from a code hygiene perspective, the style guide item and the enforcement seems reasonable to me. While one doesn't always look at the imports section of the file (especially since IDEs often auto-manage them), if looking at code in GitHub or for any reason examining imports including the relatively common task of reviewing changes to imports, then having them grouped is a readability improvement IMO: it immediately gives signal at a glance about whether someone is adding a new stdlib import vs. third-party import. (The reason I like the three-group split, personally, is because it is even better for distinguishing true third-party imports from local/first-party imports at a glance.)

I think a tax on productivity, in practice, is near de minimis, especially compared to the other things one should be considering during a self-review for the style guide. A consistent and clean codebase can improve developer morale, which can be a productivity boost. So removing stuff from the style guide that seems aesthetic because there's any effort expended to enforce/improve it seems like a way to get to an unclean codebase "death by a thousand cuts"-style.

akshayjshah:
> LOL, I actually did this _before_ opening a PR on core: #1629

:facepalm: Sorry, I PR'ed without looking. Rookie move on my part.

akshayjshah:
This is unblocked now (per discussion on the core PR). @jhump, can you review/stamp?

jhump (approved):

bufdev (approved):

PR #1632, opened 2022-11-29
Improve Bucket and Module docs

unmultimedio:
Had some doubts when using this from push, had to read implementation, so just updating the funcs docs.


twilly (approved):
:shipit:

PR #1634, opened 2022-11-30
add new fields to GetSchemaRequest

jhump:
Adds additional request fields for the GetSchema endpoint to allow additional control over the filtered image and to enable efficient client-side caching.

Fixes TCN-804, TCN-806


jhump:
I pushed a commit with incompatible changes. This is okay since there are no consumers of the API yet. It is currently labeled "alpha" for a reason 😛 

The last commit, unrelated to the linked Linear tickets, flattens the module details and removes the functionality of filtering out well-known files.

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -32,48 +31,54 @@ service SchemaService {
...
+  // not send back any descriptors since the client already has them.
   //
-  // These files are a subset of the files in the "well-known types" module:
-  // https://buf.build/beta/wellknowntypes
+  // This allows a client to efficiently poll for updates: after the initial RPC
+  // to get a schema, the client can cache the descriptors and the resolved
+  // version. It then includes that version in subsequent requests in this field,
+  // and the server will only reply with a schema (and new version number) if/when
+  // the resolved version changes.
+  string if_not_version = 5;
```
bufdev, line 63: feels like something that should be in a header and  generic

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -32,48 +31,54 @@ service SchemaService {
...
-  //  * google/protobuf/descriptor.proto
-  //  * google/protobuf/duration.proto
-  //  * google/protobuf/empty.proto
-  //  * google/protobuf/field_mask.proto
-  //  * google/protobuf/struct.proto
-  //  * google/protobuf/timestamp.proto
-  //  * google/protobuf/wrapper.proto
+  // This flag is ignored if element_names is empty as the entire schema is always
+  // returned in that case.
+  bool exclude_custom_options = 6;
```
bufdev, line 72: probably dont need at launch? can discuss

proto/buf/alpha/registry/v1alpha1/schema.proto
jhump: _For posterity, here are remarks from a Zoom discussion:_

Indeed! In fact, if this were a simple HTTP GET endpoint, it would be an `If-Not-Modified` header. But an HTTP server has clear semantics for that header and can return a 304 status code to indicate that it decided to not return a response. But for RPC, those semantics don't really apply. This value necessarily impacts the behavior of the operation, causing it to return a nil file descriptor set in the response.

I don't really like putting behavior-changing inputs in metadata. It's more awkward to document since it has to just be prose in the method comment, it's not discoverable in an IDE since auto-completion won't allow a caller to discover it, and it's not type-safe (not a big deal in this case since the type of this field is just a string).

If there were a _standard_ for something like this, I'd be much more amenable to demoting this out of the request. But without such a standard for this kind of behavior, I think this really should be a request field.

In the future, we can wrap this RPC endpoint In a simple HTTP GET API, wherein this will indeed be a header. (And the version in the response message will  be an "Etag" response hedaer.) And when the handler replies with no file descriptor set, the HTTP handler can return a "304 Not Modified".

elliotmjackson (approved):

PR #1635, opened 2022-11-30
apimodule: test resolver and fix nil exception

twilly:
bufapimodule.moduleResolver.GetModulePin would panic if the response's RepositoryCommit is nil. This handles such a response with an error about the empty response. This also adds test coverage for GetModulePin.

Fixes: TCN-808

### DIFF
--- a/private/bufpkg/bufapimodule/module_resolver_test.go
+++ b/private/bufpkg/bufapimodule/module_resolver_test.go
@@ -16,7 +16,6 @@ package bufapimodule
 
 import (
 	"context"
-	"errors"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufapimodule/module_resolver_test.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil, errors.New("unexpected call")
+}
+
+func (m *mockCommitServiceClient) DeleteRepositoryDraftCommit(
+	_ context.Context,
+	_ *connect.Request[registryv1alpha1.DeleteRepositoryDraftCommitRequest],
+) (*connect.Response[registryv1alpha1.DeleteRepositoryDraftCommitResponse], error) {
+	m.t.Error("unexpected call: DeleteRepositoryDraftCommit")
+	return nil, errors.New("unexpected call")
+}
```
elliotmjackson, line 81:
```suggestion
type mockCommitServiceClient struct {
	registryv1alpha1connect.UnimplementedRepositoryCommitServiceHandler
	t       *testing.T
	refResp *registryv1alpha1.GetRepositoryCommitByReferenceResponse
}
```

twilly: Don't I wish it were that easy. All methods are required to fulfill `registryv1alpha1connect.RepositoryCommitServiceClient`.

jchadwick-buf: But embedding `registryv1alpha1connect.UnimplementedRepositoryCommitServiceHandler` should give you at least stub methods for the rest, right? Am I misunderstanding?

twilly: Yeah, I always put a space between fields and any embedded types, so I didn't see the suggestion's embed.

PR #1636, opened 2022-11-30
add AdminService CreateMachineUser rpc

cyinma:
fixes BSR-956


doriable (approved):

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -63,3 +65,11 @@ message UpdateOrganizationVerificationStatusRequest {
 }
 
 message UpdateOrganizationVerificationStatusResponse {}
+
+message CreateMachineUserRequest {
+  string username = 1;
+}
+
+message CreateMachineUserResponse {
+  User user = 1;
```
mfridman, line 74: Does the User object need an enum or a type to differentiate between a "regular" vs "machine" user?

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/admin.proto
cyinma: yes, but I am planning to include this in a later PR

PR #1639, opened 2022-12-01
schema.proto alternate proposal and comments

bufdev:
DO NOT MERGE

### DIFF
--- a/proto/buf/alpha/registry/v1alpha1/schema.proto
+++ b/proto/buf/alpha/registry/v1alpha1/schema.proto
@@ -37,18 +37,16 @@ service ConvertService {
 message GetFileDescriptorSetRequest {
   // always unwrap when easily possible for requests (ie dont use nested message)
   // TODO should these be here, or should we make this more generic?
-  string owner = 1;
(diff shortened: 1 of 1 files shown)


proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -21,20 +21,33 @@ import "google/protobuf/descriptor.proto";
 
 // Provides endpoints for downloading dynamic schemas and for using schemas
 // in validation and data transformation functions.
-service SchemaService {
-  // GetSchema allows the caller to download a schema for one or more requested
+// TODO don't like name
+service FileDescriptorSetService {
```
elliotmjackson, line 25:
```suggestion
service DescriptorService {
```

bufdev: some nuance here...better name but the problem is that a `DescriptorProto` should be named a `MessageDescriptorProto` but isn't https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/descriptor.proto#L98

PR #1650, opened 2022-12-06
TokenService include userID in requests

cyinma:
fixes BSR-962

### DIFF
--- a/proto/buf/alpha/registry/v1alpha1/token.proto
+++ b/proto/buf/alpha/registry/v1alpha1/token.proto
@@ -50,6 +50,7 @@ message CreateTokenRequest {
   google.protobuf.Timestamp expire_time = 2;
   // The ID of the user to create a new token, can only be
   // the user who perform this action, or a machine user.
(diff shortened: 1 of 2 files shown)


proto/buf/alpha/registry/v1alpha1/token.proto
```diff
@@ -48,6 +48,9 @@ message CreateTokenRequest {
   // The time until which the token should be valid.
   // Must be in the future. May be null for no expiry.
   google.protobuf.Timestamp expire_time = 2;
+  // The ID of the user to create a new token, can only be
+  // the user who perform this action, or a machine user.
+  string user_id = 3;
```
stefanvanburen, line 53: will this default to the user making the request, if empty? do we want to mention that in the comment?

cyinma: I think in the end we will want all clients to provide this field, but updated as empty should be accepted for now

stefanvanburen (approved):

PR #1653, opened 2022-12-07
[BSR-998] Add digest back to (local) module pins

unmultimedio:
We need to populate the digest in modules in order to put that information in the `buf.lock` file.


twilly:
This shouldn't be needed anymore. :)

PR #1657, opened 2022-12-07
TCN-858 explicitly define upload assets

elliotmjackson:
To determine the files to be uploaded, pipe the result of a find command into en env var, id only tested it on a single file in the past - when it has multiple files, it finds a new line character and terminates.

as such, release fails to upload assets. 

To move forward, im explicitly defining the upload assets so the release process can continue. Naturally a more elegant solution would be preferable but i think this still falls into the "win" category

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -144,17 +144,39 @@ jobs:
         run: |
           unset RELEASE_MINISIGN_PRIVATE_KEY
           unset RELEASE_MINISIGN_PRIVATE_KEY_PASSWORD
-      - name: Set FILES variable from tag
-        run: |
-          echo "FILES=$(find ${{github.workspace}}/.build/release/buf/assets -type f)" >> $GITHUB_ENV
       - name: Release
         id: ghr
         uses: softprops/action-gh-release@v1
+        env:
+          PATH: ${{github.workspace}}/.build/release/buf/assets
         with:
           token: ${{ steps.generate_token.outputs.token }}
           generate_release_notes: true # change to the CHANGELOG somehow
           files: |
-            ${{env.FILES}}
+            ${{env.PATH}}/buf-Darwin-arm64
+            ${{env.PATH}}/buf-Darwin-arm64.tar.gz
+            ${{env.PATH}}/buf-Darwin-x86_64
+            ${{env.PATH}}/buf-Darwin-x86_64.tar.gz
+            ${{env.PATH}}/buf-Linux-aarch64
+            ${{env.PATH}}/buf-Linux-aarch64.tar.gz
+            ${{env.PATH}}/buf-Linux-x86_64
+            ${{env.PATH}}/buf-Linux-x86_64.tar.gz
+            ${{env.PATH}}/buf-Windows-arm64.exe
+            ${{env.PATH}}/buf-Windows-x86_64.exe
+            ${{env.PATH}}/protoc-gen-buf-breaking-Darwin-arm64
+            ${{env.PATH}}/protoc-gen-buf-breaking-Darwin-x86_64
+            ${{env.PATH}}/protoc-gen-buf-breaking-Linux-aarch64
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -144,17 +144,39 @@ jobs:
...
           unset RELEASE_MINISIGN_PRIVATE_KEY
           unset RELEASE_MINISIGN_PRIVATE_KEY_PASSWORD
-      - name: Set FILES variable from tag
-        run: |
-          echo "FILES=$(find ${{github.workspace}}/.build/release/buf/assets -type f)" >> $GITHUB_ENV
       - name: Release
         id: ghr
         uses: softprops/action-gh-release@v1
+        env:
+          PATH: ${{github.workspace}}/.build/release/buf/assets
```
pkwarren, line 151: I'd recommend using something like `ASSETS_PATH` here instead of `PATH` (not sure if that would break command execution).

elliotmjackson: resolved in [e9ced1e](https://github.com/bufbuild/buf/pull/1657/commits/e9ced1e469041d3661458e71b4999b6f58e3d265)

pkwarren (approved):

.github/workflows/release.yaml
```diff
@@ -144,17 +144,39 @@ jobs:
...
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-breaking-Windows-arm64.exe
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-breaking-Windows-x86_64.exe
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Darwin-arm64
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Darwin-x86_64
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Linux-aarch64
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Linux-x86_64
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Windows-arm64.exe
+            ${{env.ASSETS_PATH}}/protoc-gen-buf-lint-Windows-x86_64.exe
+            ${{env.ASSETS_PATH}}/sha256.txt
+            ${{env.ASSETS_PATH}}/sha256.txt.minisig
```
pkwarren, line 179: It looks like the action supports glob style path patterns, so we might be able to do:

```yaml
files: ${{env.ASSETS_PATH}}/*
```

See https://github.com/softprops/action-gh-release/blob/9114792eb2fcdea3dbb139ffd081fcb856870dcf/src/util.ts#L76-L82 for more details.

elliotmjackson: 1 second... will try

elliotmjackson, line 179:
```suggestion
            ${{github.workspace}}/.build/release/buf/assets/*
```

.github/workflows/release.yaml
```diff
@@ -144,17 +144,39 @@ jobs:
...
           unset RELEASE_MINISIGN_PRIVATE_KEY
           unset RELEASE_MINISIGN_PRIVATE_KEY_PASSWORD
-      - name: Set FILES variable from tag
-        run: |
-          echo "FILES=$(find ${{github.workspace}}/.build/release/buf/assets -type f)" >> $GITHUB_ENV
       - name: Release
         id: ghr
         uses: softprops/action-gh-release@v1
+        env:
+          ASSETS_PATH: ${{github.workspace}}/.build/release/buf/assets
```
elliotmjackson, line 151:
```suggestion
```

.github/workflows/release.yaml
elliotmjackson: potentially the definition of overthinking it. resolved in [d4e3f0d](https://github.com/bufbuild/buf/pull/1657/commits/d4e3f0deada832e4dc248961e4392fad3af51f6a)

pkwarren (approved):

PR #1658, opened 2022-12-07
add support for installing released plugins

pkwarren:
Update `buf alpha plugin push` command to add support for installing plugins from released versions (zip file containing buf.plugin.yaml and image.tar). Remove all support for building Docker images (Docker CLI is preferred).


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -216,44 +196,39 @@ func run(
 		retErr = multierr.Append(retErr, client.Close())
 	}()
 
-	var createdImage string
-	if len(flags.Image) > 0 {
-		tagResponse, err := client.Tag(ctx, flags.Image, pluginConfig)
+	var imageToTag string
+	if flags.Image != "" {
+		imageToTag = flags.Image
```
pkwarren, line 201: The `--image` always takes precedence. If not specified, we fall back to looking for a pre-built image.tar in the directory where the `buf.plugin.yaml` is found.

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
```diff
@@ -193,22 +123,38 @@ func createClient(t testing.TB, options ...ClientOption) Client {
 	return dockerClient
 }
 
-func buildDockerPlugin(t testing.TB, dockerClient Client, dockerfilePath string, pluginIdentity string, pluginVersion string) (*BuildResponse, error) {
+func buildDockerPlugin(t testing.TB, dockerfilePath string, pluginIdentity string) (string, error) {
```
pkwarren, line 126: Removed code which used to build docker images in favor of just shelling out to the docker command to run some basic tests. This still allows us to keep coverage up for issues seen when pushing images, etc.

go.mod
```diff
@@ -14,7 +14,6 @@ require (
 	github.com/jhump/protoreflect v1.14.0
 	github.com/klauspost/compress v1.15.12
 	github.com/klauspost/pgzip v1.2.5
-	github.com/moby/buildkit v0.10.5
```
mfridman, line 17: 😌

private/bufpkg/bufplugin/bufplugindocker/docker_test.go
mfridman: At this point we're effectively testing the docker cli, do we still find these tests useful?

https://github.com/bufbuild/buf/blob/34fdcfa897728de29e527ba4e30c06d856ab654c/private/bufpkg/bufplugin/bufplugindocker/docker.go#L205-L208

Did the previous tests require you to have docker cli or the docker engine installed?

pkwarren: Yeah they did. We still exercise tests where we simulate errors from the docker registry push, etc. I think it is still useful to keep around.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
mfridman: I'd error to avoid surprising behavior. But we don't need to get the UX ideal right now, let's see how this will be used.

mfridman (approved):

PR #1659, opened 2022-12-08
image filtering: don't include entire enclosing message/service

jhump:
This addresses four of the TODOs I recently added to this code.

Two of them were about eliding the contents of messages that are _only_ present in the graph because they _contain_ a connected type (but are otherwise unused). For these cases, the message really only serves as a namespace, so it only needs to be present as a container for other referenced types.

Another of them was to allow the caller to provide a method name and to be able to prune its enclosing service of any other unreferenced methods.

The final TODO was to improve the logic around when to include custom options. Previously, if an extension was a custom option, it would only be included if referenced by actual option declarations on elements in the graph. But now such extensions will also be included, even if excluding custom options, if the operation is including known extensions and the options types (which custom options extend) are otherwise in the graph of types. (This is implemented via the notion of "implicit" vs. "explicit" members of the transitive closure.)

Finally, this overhauls the code that trimmed a file descriptor proto to be much more concise and more DRY (via a generic function and removing the `error` return value, which was never needed because these operations cannot fail).

I apologize for the big diff stats, but FWIW a big majority of the new lines come from new golden files for new tests.

Fixes TCN-886


private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -544,49 +474,129 @@ func (t *transitiveClosure) addElement(
...
+		}
+		if _, ok := t.elements[descriptor]; ok {
+			break // already in closure
+		}
+		t.elements[descriptor] = inclusionModeEnclosing
+		if err := t.exploreCustomOptions(descriptor, enclosingFile, imageIndex, opts); err != nil {
+			return err
+		}
+		// now move into this element's parent
+		descriptor = imageIndex.ByDescriptor[descriptor].parent
```
robbertvanginkel, line 535: I'm having a little trouble following the control flow here. From what I can see we only reach this method through:

```
descriptorInfo := imageIndex.ByDescriptor[descriptor]
...
// if this type is enclosed inside another, add enclosing types
if err := t.addEnclosing(descriptorInfo.parent, descriptorInfo.file, imageIndex, opts); err != nil {
	return err
}
// <no use of descriptorInfo.parent>
```

In the calling function (`addElement`), `descriptorInfo` is a `elementInfo` struct (no pointer).

I commented this out and test started failing so I must be misreading something, but what is going on here is not clear to me.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -544,49 +474,129 @@ func (t *transitiveClosure) addElement(
...
+	case *descriptorpb.MethodDescriptorProto:
+		descriptorType = "method"
+	default:
+		descriptorType = fmt.Sprintf("%T", d)
+	}
+	return fmt.Errorf("%s is unsupported filter type: %s", fullName, descriptorType)
+}
+
+func (t *transitiveClosure) addEnclosing(descriptor namedDescriptor, enclosingFile string, imageIndex *imageIndex, opts *imageFilterOptions) error {
+	for descriptor != nil {
```
jhump, line 521:
```suggestion
	// loop through all enclosing parents since nesting level
	// could be arbitrarily deep
	for descriptor != nil {
```

robbertvanginkel (approved):

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
jhump: Per F2F, there was confusion around this being a loop vs. a conditional. It is not really intuitive that it's a loop, so it's easy to misread `for` as `if` at the top. Added a comment to improve clarity.

PR #1661, opened 2022-12-09
bufgen: Batch generation

srikrsna-buf:
Closes TCN-823.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-- `buf generate` now batches to remote plugins calls.
+- `buf generate` now batches remote plugins calls.
 
 ## [v1.10.0] - 2022-12-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [UNRELEASED]
 
-- `buf generate` now batches remote plugins calls.
+- `buf generate` now batches remote plugin generation calls for improved performance.
 
 ## [v1.10.0] - 2022-12-07
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [UNRELEASED]
+
+- `buf generate` now batches to remote plugins calls.
+
 ## [v1.10.0] - 2022-12-07
 
 - When using managed mode, setting `enabled: false` now no longer fails `buf generate` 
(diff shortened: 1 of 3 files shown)


private/buf/bufgen/generator.go
```diff
@@ -169,15 +169,25 @@ func (g *generator) execPlugins(
 	jobs := make([]func(context.Context) error, 0, len(config.PluginConfigs))
 	responses := make([]*pluginpb.CodeGeneratorResponse, len(config.PluginConfigs))
 	requiredFeatures := computeRequiredFeatures(image)
+	remotePluginConfigTable := make(map[string]map[int]*PluginConfig, len(config.PluginConfigs))
```
pkwarren, line 172: Instead of two layers of maps, can we have a map of a slice of structs with something like:

```go
type remotePluginExec struct {
  ResponseIndex int
  Config *PluginConfig
}
```

Then for debugging, it is a little easier to figure out what the `int` is used for and we don't shuffle around the remote plugins to execute unnecessarily. I know order doesn't matter in plugin execution until we assemble the results at the end, but it would be nice for debugging if the order specified in the buf.gen.yaml file was kept.

private/buf/bufgen/generator.go
```diff
@@ -169,15 +169,25 @@ func (g *generator) execPlugins(
 	jobs := make([]func(context.Context) error, 0, len(config.PluginConfigs))
```
pkwarren, line 169: While you are here, can you remove the `appprotoexecGenerator` argument? It is already available on the generator itself and is unused here.

private/buf/bufgen/generator.go
```diff
@@ -380,14 +416,62 @@ func (g *generator) execRemotePluginV2(
...
+		codeGeneratorResponse := responses[i].GetResponse()
+		if codeGeneratorResponse == nil {
+			return nil, errors.New("expected code generator response")
+		}
+		responseSet[indexes[i]] = codeGeneratorResponse
+	}
+	return responseSet, nil
+}
+
+func getPluginGenerationRequest(
```
pkwarren, line 433: Thanks for refactoring these out.

pkwarren (approved):
Changes look good to me. This feels like something we should highlight in the changelog for next release.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [UNRELEASED]
+
+- `buf generate` now batches remote plugins calls.
+
```
srikrsna-buf, line 6: @pkwarren

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [UNRELEASED]
+
+- `buf generate` now batches remote plugins calls.
```
pkwarren, line 5:
```suggestion
- `buf generate` now batches remote plugin generation calls for improved performance.
```

pkwarren (approved):

PR #1664, opened 2022-12-09
Push a minimal set of blobs.

twilly:
This is an exercise in frustration to push only the set of blobs needed to build a module. The bug's cause was building a module did not expose the bucket used to build the module with, hence the manifest consumed the source bucket containing module-unrelated files.

bufmodulebuild tries to abstract away building but had, and still has, many rough edges. Some of note are:

 - Build configuration is in the bucket, but the build section of configuration is not read from the bucket.

 - Build configuration cannot be written to `buf.yaml`.

 - Incorrect build configuration in a bucket can make a bucket appear buildable when it is not.

 - Some build configuration is not in the build section nor in the bucket, instead being hard coded as options.

 - BuildForBucket would take options directly, even though it was a method. It's a plain function masquerading as a method, causing callers to believe they're injecting when really the tail call was constructing independent behavior.

I tried to remove passing configuration by value or to serialize configuration into a bucket if it must be passed as a value (e.g. in many tests). It turns out writing configuration from configuration structs is not possible; the build section is omitted. I suspect it's impossible to round-trip configuration to entities to configuration and back to identical entities.

Included in this commit is a compromise in exposing a minimal module bucket. It should produce correct buckets if configuration is passed that was read from the source bucket and no options are used to change the configuration's specified behavior. This is the case with `buf push` but not the case with some other commands, notably `buf lint` given it's `--config` option.

Changes include:

 - Remove bufcli.ReadModule wrapper on bufmodulebuild.  This wrapper didn't provide any value over directly invoking bufmodulebuild.

 - Remove logging support from bucket builder. Logging is not used, and should have been passed as a constructor option anyway.

 - Move BuildForBucket to be a function that construct a builder and calls a new BuildForBucket method.

 - Type alias ModuleBucketBuilder to avoid rewriting the world if the bucket builder needs a new method.

 - Expose the minified bucket by returning a struct with both a usable module and the bucket used to create it. This struct could include a usable configuration structure to inspect or write as a buf.yaml, but I'll leave that as an exercise for another unfortunate soul.

Fixes: TCN-796

### DIFF
--- a/private/buf/cmd/buf/command/push/push_test.go
+++ b/private/buf/cmd/buf/command/push/push_test.go
@@ -147,16 +147,16 @@ func TestPush(t *testing.T) {
 	)
 }
 
-func TestPushIsMinimal(t *testing.T) {
-	// Assert push only manifests the minimal amount of files needed to build
-	// the module.
+func TestPushIsSmallerBucket(t *testing.T) {
+	// Assert push only manifests with only the files needed to build the
+	// module as described by configuration and file extension.
 	t.Parallel()
 	server, mock := pushService(t)
 	t.Cleanup(func() {
 		server.Close()
 		mock.assertAllCallbacksCalled()
 	})
-	const owner = "minimal"
+	const owner = "smallerbucket"
 	mock.respond(
 		owner,
 		&registryv1alpha1.PushResponse{
--- a/private/buf/cmd/buf/command/push/push_test.go
+++ b/private/buf/cmd/buf/command/push/push_test.go
@@ -38,6 +38,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
+	"github.com/bufbuild/buf/private/pkg/manifest"
 	"github.com/bufbuild/buf/private/pkg/storage/storagemem"
 	connect_go "github.com/bufbuild/connect-go"
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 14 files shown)


private/buf/cmd/buf/command/push/push_test.go
```diff
@@ -146,6 +147,46 @@ func TestPush(t *testing.T) {
 	)
 }
 
+func TestPushIsMinimal(t *testing.T) {
+	// Assert push only manifests the minimal amount of files needed to build
+	// the module.
```
unmultimedio, line 152: Nit: May be clear to say that's not pushing a _minimal_ amount of files (sounds like some of them are optional and we're deciding not to push them), but instead we're only pushing _allowed_ files, so we're actually pushing _all of the allowed_ files, and ignoring unrelated ones.

unmultimedio (commented):
lgtm, but I don't have a lot of context for these build functions and possible consequences.

private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
```diff
@@ -21,55 +21,64 @@ import (
...
+func (b *moduleBucketBuilder) BuildForBucket(
 	ctx context.Context,
 	readBucket storage.ReadBucket,
 	config *bufmoduleconfig.Config,
-	moduleIdentity bufmoduleref.ModuleIdentity,
-	bucketRelPaths *[]string,
-	excludeRelPaths []string,
-	bucketRelPathsAllowNotExist bool,
-) (bufmodule.Module, error) {
+) (builtModule BuiltModule, err error) {
```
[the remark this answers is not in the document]
twilly: Named returns because it made returning a tuple of values as a struct on the stack pretty simple. But hey, I don't mind returning a struct ptr.

private/buf/cmd/buf/command/push/push_test.go
twilly: It's not even allowed but what's needed to build the module, according to the build section in the configuration. For example, excluded directories are not sent. "Minimal" may be misleading though, so I'll give this another wording spin.

unmultimedio: I see you changed to "smaller". Not strictly always smaller. User could have just one proto and `buf.yaml`, with no excluded directories, and the result built module could even be bigger (includes an empty `buf.lock`?), or at least the same amount of files.

In my mind what happens is then not an "allow" either (as I suggested) but a "curation" or "validation" or "preparation", meaning "let's see which files in this dir actually make part of the module": excluding directories, ignoring unrecognized files, generating missing `buf.lock` file, etc. So may be more like "curated" or "validated" bucket?

Not blocking ofc. But I think it helps for clarity.

twilly: We're not checking if the module is valid (e.g. buildable), so validated is misleading. "Curate" could work but that also implies correctness of some aspect, whatever it may be in the perspective of the curator.

Perhaps "subset" would be a more appropriate term. No correctness statement is being made and one subset of a set is the set itself. Minimal / smaller concepts don't come into play.

unmultimedio (approved):

PR #1665, opened 2022-12-09
optimize checking for current digest

pkwarren:
If the current digest has the same image ID as the image to push, we can greatly speed up looking through every tag (which may contain multiple duplicates - potentially one per version/revision of a plugin). This should greatly speed up the lookup in the case where the plugin already exists on the OCI registry.


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -376,7 +384,19 @@ func findExistingDigestForImageID(ctx context.Context, plugin *bufpluginconfig.C
 		return "", err
 	}
 	auth := &authn.Basic{Username: authConfig.Username, Password: authConfig.Password}
-	tags, err := remote.List(repo, remote.WithContext(ctx), remote.WithAuth(auth))
+	remoteOpts := []remote.Option{remote.WithContext(ctx), remote.WithAuth(auth)}
+	// First attempt to see if the current image digest matches the image ID
```
pkwarren, line 388: This is a lot faster in the case of the image already existing in the BSR with the same digest as before. It only has to check one digest (the current one) and not potentially search through dozens of tags.

robbertvanginkel (approved):

PR #1675, opened 2022-12-12
fix regression in plugin generation

pkwarren:
The 'Remote' field contains the full plugin name (including path and version information). For the GetRemoteHostname() method, we should parse the hostname portion and return it.

This resolves 404 errors when generating with older plugins.

### DIFF
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -26,6 +26,7 @@ import (
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
 	"github.com/bufbuild/buf/private/bufpkg/bufplugin/bufpluginref"
+	"github.com/bufbuild/buf/private/bufpkg/bufremoteplugin"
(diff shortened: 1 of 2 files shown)


private/buf/bufgen/bufgen.go
```diff
@@ -219,7 +220,13 @@ func (p *PluginConfig) GetRemoteHostname() string {
 	if reference, err := bufpluginref.PluginReferenceForString(p.Plugin, 0); err == nil {
 		return reference.Remote()
 	}
-	return p.Remote
+	if p.Remote == "" {
+		return ""
+	}
+	if remote, _, _, _, err := bufremoteplugin.ParsePluginVersionPath(p.Remote); err == nil {
+		return remote
+	}
```
unmultimedio, line 228: Could we log the `err` here, or shown as a warning to the CLI in case the remote is "unparsable"?

pkwarren: That should already happen earlier at this location: https://github.com/bufbuild/buf/blob/5d299811bbe0f329b30f3984b92211bf8442d053/private/buf/bufgen/config.go#L206-L212

unmultimedio (approved):

PR #1676, opened 2022-12-12
Add transform package

elliotmjackson:
Transform allows the caller to convert a given message data blob from one format to another by referring to a type schema for the blob. Utilizing the BSRs [SchemaService](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.SchemaService), the idea is to simplify integration and improve performance for the user.

A default client provides you with connection to the public Buf Schema Registry and an in memory cache will maintain the protoencoding `Resolver` for future transformations. You must provide it with a reference to a Buf Module and the expected format of the data you wish to convert and the resulting format you would like.

```go
	client, err := transform.DefaultClient(
		"bufbuild",
		"registry",
		"01b54e71e6b84514a9141323afdb95a1",
		registryv1alpha1.Format_FORMAT_BINARY,
		transform.ToDefaultJSONOutput(),
	)
	if err != nil {
		return err
	}
	result, err := client.ConvertMessage(ctx, "foo.bar.Message", inputData)
	if err != nil {
		return err
	}
	println(result)
```


pkg/transform/README.md
```diff
@@ -0,0 +1,131 @@
...
+
+```go
+package foo
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/bufbuild/buf/pkg/transform"
+	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
```
jhump, line 26: External code won't be able to import this since it's under `private`.

I just realized something much worse: it's not just that external code can't _import_ private packages, they also can't even be _linked_ to private packages. In other words, this `pkg/transform` cannot use any private packages, _even indirectly_, or else external binaries that try to use it will [blow up](https://github.com/bufbuild/buf/blob/main/private/usage/usage.go#L28-L32) during `init`.

So I think we have two options with how to proceed:
1. Remove all references to the private packages.
    * Don't use these types in this package's API. Create a separate enum type for the format, for example.
    * Add a step to `make generate` that generates only the schema service (can use the new type filtering stuff in `buf generate`) into some place like `pkg/transform/internal`).
    * We'd likely need a CI check that can make sure we never accidentally add a transitive dependency from pkg -> private since that would break users.
2. Move this package under private and punt on making it a library for external users.
    * Basically, we try to incorporate it into things only as an internal library first (like a sub-command of the buf CLI or a stand-alone executable that we provide; anything that lives under "github.com/bufbuild" is allowed to import the private packages.).

I think we may find the second option the path of least resistance for now. And we will need to have some serious conversations about how we can create open-source libraries. (This inevitably will involve pulling some stuff out of private; we'd likely want to mostly move it to internal packages, so that they don't become part of exported API that we have to support.)

pkg/transform/README.md
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const (
+	useEnumNumbers  = false
+	includeDefaults = false
+)
```
jhump, line 32: These are unused in the example code below. Delete?

pkg/transform/client.go
```diff
@@ -0,0 +1,306 @@
...
+// This allows a client to efficiently poll for updates: after the initial RPC
+// to get a schema, the client can cache the descriptors and the resolved
+// commit. It then includes that commit in subsequent requests in this field,
+// and the server will only reply with a schema (and new commit) if/when the
+// resolved commit changes.
+func IfNotCommit(in string) ClientBuilder {
+	return func(c *Client) {
+		c.ifNotCommit = in
+	}
+}
```
jhump, line 208: I don't think we'd expose this. Instead, this would be part of the client's internal cache management.

On first call, it retrieves a schema and version from the cache; if present, it includes that last seen version as this field in the RPC request.

On subsequent calls, it always includes the last seen version (and stores that version and new schema when it changes).

pkg/transform/client.go
```diff
@@ -0,0 +1,306 @@
...
+		c.client = client
+	}
+}
+
+// WithCache provided a default expiration duration and cleanup
+// interval. If the expiration duration is less than one (or NoExpiration),
+// the items in the cache never expire (by default), and must be deleted
+// manually. If the cleanup interval is less than one, expired items are not
+// deleted from the cache before calling c.DeleteExpired().
+func WithCache() ClientBuilder {
```
jhump, line 114: I think we'd want the caller to provide the cache, no? `WithDefaultCache` below could use a simple in-memory cache. But caller could provide some impl that is backed by redis or memcache, etc. The cache type could be a simple interface like:
```go
type SchemaCache interface {
    Load() (schema *descriptorpb.FileDescriptorSet, commit string, error)
    Save(schema *descriptorpb.FileDescriptorSet, commit string) error
}
```

pkg/transform/client.go
```diff
@@ -0,0 +1,306 @@
...
+	"net/http"
+
+	"github.com/bufbuild/buf/pkg/transform/internal/cache"
+	"github.com/bufbuild/buf/private/gen/proto/connect/buf/alpha/registry/v1alpha1/registryv1alpha1connect"
+	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
+	"github.com/bufbuild/connect-go"
+)
+
+// Client ..
+type Client struct {
```
jhump, line 30: I wonder if it's worth decomposing this a little further. Maybe we could have two different types: the `SchemaWatcher` and the `Converter`. So the client properties related to invoking the schema service RPC and caching the result could be in `SchemaWatcher`, and you then pass a `SchemaWatcher` along with from/to format options when creating a `Converter`. WDYT?

pkg/transform/internal/cache/cache.go
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cache
+
+import (
+	"sync"
+
+	"github.com/bufbuild/buf/private/pkg/protoencoding"
```
jhump, line 20: Oy, this will be a bit of a bear to untangle out of "private".  I think that's another bit of evidence that we should probably incubate this stuff inside of "private" and then figure out a path to open-sourcing it for external use as a library later.

pkg/transform/transform.go
```diff
@@ -0,0 +1,166 @@
...
+		return nil, fmt.Errorf("input message cannot be unmarshaled to %s: %w", c.inputFormat, err)
+	}
+	// Done!
+	return data, nil
+}
+
+// getProtoEncodingResolver allows the caller to download a schema for one or more requested
+// types, RPC services, or RPC methods.
+func (c *Client) getProtoEncodingResolver(ctx context.Context) (protoencoding.Resolver, error) {
+	if c.cache != nil {
```
jhump, line 69: I'm thinking maybe we _always_ use an in-memory cache, that is loaded/updated synchronously. And then we only fire off a goroutine to asynchronously save the results. We really only need to load the schema from remote source at startup (when there's nothing in memory for it).

We can't efficiently store a resolver into redis, for example: we'd have to serialize it as a file descriptor set. If only using redis (and not in-memory), this could be extremely inefficient if there are many messages to convert, because it would be constantly re-downloading the schema from redis and reconstructing the resolver.

jhump (commented):
I didn't go through everything with a fine-tooth comb. Just pointing out some high-level things to reconsider.

bufdev (changes requested):
This doesn't seem like it should be merged into bufbuild/buf, it seems to be an external library that is being just pushed into `private/pkg` wholesale. Issues I see at first glance:

- It's using remote packages for bufbuild/buf itself, when the rest of the code in bufbuild/buf does not use them
- It has complete copies of protoencoding and protodescriptor, two packages that live in bufbuild/buf already

Regardless of where we decide to put this, code must conform to style guide (comments on all public elements, proper ordering, etc) and if this is to become an external library, we need to consider the interface carefully including all options.

elliotmjackson:
Moved to bufbuild/prototransform#1

PR #1677, opened 2022-12-13
[BSR-701] default except override for objective c

oliversun09:
Add default, except and override for `objc_class_prefix`


doriable (approved):
Looks good, tested and ran locally :) Thanks! Make sure you update the changelog.

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
 ## [UNRELEASED]
+- Add `objc_class_prefix` option in managed mode, allowing `default`, `except` and `override`
+  in the same style as `optimize_for`.
```
doriable, line 5: Instead of having this refer back to `optimize_for` changelog statement, we should have one that expands on this. We can use the same wording and structure as the previous changelog statement, but it's better if this was a standalone statement.

doriable (approved):

PR #1680, opened 2022-12-14
New RPCs for Push and Download

unmultimedio:
## ⚠️ This is just a proof of concept to have something that compiles to use it in the backend.

To enrich existing conversation on new RPCs for tamper proofing abilities.

Fixes: TCN-905


proto/buf/alpha/registry/v1alpha1/download.proto
```diff
@@ -21,20 +21,38 @@ import "buf/alpha/module/v1alpha1/module.proto";
 // DownloadService is the download service.
 service DownloadService {
   // Download downloads.
+  // Soon to be deprecated.
```
pkwarren, line 24: We should state it is soon to be deprecated in favor of DownloadManifest.

proto/buf/alpha/registry/v1alpha1/download.proto
```diff
@@ -21,20 +21,38 @@ import "buf/alpha/module/v1alpha1/module.proto";
...
   rpc Download(DownloadRequest) returns (DownloadResponse);
+  // DownloadManifest downloads a module in the manifest+blobs encoding format.
+  rpc DownloadManifest(DownloadManifestRequest) returns (DownloadManifestResponse);
```
pkwarren, line 27: Do we want to name these `DownloadManifestAndBlobs` to be more accurate?

pkwarren (commented):
Looks good. For the build to go green we'll need to regenerate the code.

proto/buf/alpha/registry/v1alpha1/download.proto
unmultimedio: Yes could be, I'm more for shorter, so another option is `*CAS` suffix, but that could be cryptic?

proto/buf/alpha/registry/v1alpha1/download.proto
unmultimedio: Sure, we could expand on _why_ we're not deprecating it _yet_.

proto/buf/alpha/registry/v1alpha1/download.proto
pkwarren: I don't mind *CAS.

bufdev:
Can you expand on why we're making new RPCs for this?

bufdev:
If there's a need to change the behavior of `Download`, the cleanest way to do this would be to have a new `buf.alpha.registry.v1alpha2` package with a new `DownloadService`, as opposed to a new RPC. This may be prohibitive in terms of time to market, as a lot of stuff is organized around `buf.alpha.registry.v1alpha1`, but just of note, this _is_ adding tech debt.

unmultimedio:
> Can you expand on why we're making new RPCs for this?

This is to support tamper proofing abilities in an upcoming CLI version that will push/download blobs/files with a consistent encoding instead of modules in a proto message. I will share the document internally!

proto/buf/alpha/registry/v1alpha1/download.proto
```diff
@@ -21,20 +21,38 @@ import "buf/alpha/module/v1alpha1/module.proto";
 // DownloadService is the download service.
 service DownloadService {
   // Download downloads.
+  // Soon to be deprecated.
   rpc Download(DownloadRequest) returns (DownloadResponse);
+  // DownloadCAS downloads a module in the manifest+blobs encoding format.
+  rpc DownloadCAS(DownloadCASRequest) returns (DownloadCASResponse);
```
twilly, line 27: Instead of CAS, I'd say blobs or manifest. CAS has no meaning to the caller.

bufdev:
Not blocking this, but added my two cents in the notion doc

unmultimedio:
Closing in favor of https://github.com/bufbuild/buf/pull/1693

PR #1689, opened 2022-12-19
update ListUsers type filter to array

cyinma:
its likely that we will add additional user type in future and it will be much more useful if the type filter is an array. this is newly added and unused yet, so I am going to just change it in place


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -119,8 +119,8 @@ message ListUsersRequest {
   bool reverse = 3;
   // If the user_state_filter is unspecified, users of all states are included.
   UserState user_state_filter = 4;
-  // If the user_type_filter is unspecified, users of all types are included.
-  UserType user_type_filter = 5;
+  // If the user_type_filters is empty, users of all types are included.
+  repeated UserType user_type_filters = 5;
```
saquibmian, line 123: Quick q: what happens if one of these is unspecified? Would we filter those out before we check for emptiness?

saquibmian (commented):
Nice. We chatted offline, and while this does look incongruent with `user_state_filter`, it's probably better that this is an array and we only have to deal with migrating potentially one of them to not an an array.

proto/buf/alpha/registry/v1alpha1/user.proto
cyinma: it should be an invalid argument error IMO?

saquibmian: Sounds good!

saquibmian (approved):

PR #1691, opened 2022-12-20
Do not start spans from the CLI

doriable:
We should not be starting traces/spans from the CLI. This should be instrumented starting from the edge of our system.

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -604,7 +604,6 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 		return nil, err
 	}
 	client := httpclient.NewClient(
-		httpclient.WithObservability(),
 		httpclient.WithTLSConfig(config.TLS),
 	)
 	options := []connectclient.ConfigOption{


buildbreaker (approved):
Maybe my other teammates have differing opinions but I've always found client side tracing to be complex to use without having a specific problem to target.

robbertvanginkel:
I think it should be fine to also start traces from the client, we just need to handle them different at the edge. The recommendations from https://github.com/census-instrumentation/opencensus-specs/blob/8600b7497457a277e681552005556407e9666997/utils/HandleUntrustedRequests.md#what-to-do-with-trace-headers-from-an-untrusted-request & https://pkg.go.dev/go.opentelemetry.io/otel/trace#Link seem to suggest we should link them instead of treating them as parents.

I'm fine with removing them for now, but check BSR-683 for some follow ups.

doriable:
In this case, let's remove them for now, simply because we would not be using the same code to instrument CLI clients and backend clients. We can always come back and add this functionality correctly further down the line.

robbertvanginkel (approved):

mfridman (approved):

PR #1693, opened 2022-12-20
add new RPCs for Download/Push

pkwarren:
To support tamper proofing, add new Download and Push RPCs which contain the module manifest and blobs.


private/buf/cmd/buf/command/push/push.go
```diff
@@ -168,25 +169,67 @@ func run(
...
+) (*registryv1alpha1.LocalModulePin, error) {
 	clientConfig, err := bufcli.NewConnectClientConfig(container)
 	if err != nil {
-		return err
+		return nil, err
 	}
 	service := connectclient.Make(clientConfig, moduleIdentity.Remote(), registryv1alpha1connect.NewPushServiceClient)
+	// Check if tamper proofing env var is enabled
+	tamperProofingEnabled, err := strconv.ParseBool(container.Env(bufcli.BetaEnableTamperProofingEnvKey))
+	if err == nil && tamperProofingEnabled {
```
unmultimedio, line 207: Nits:

### `if err != nil`
- should we log it?, so a user with that ENV flag set to true would know that a regular push was used instead?
- should we fail and don't even try the push? otherwise how a user with that ENV set knows for sure the new RPC was used? how brittle is `strconv.ParseBool(container.Env(bufcli.BetaEnableTamperProofingEnvKey))`?

### `if tamperProofingEnabled`
- Log that the new RPC is being used? so the user with that ENV set knows for sure that's being used. This log could be removed whenever we turn on this by default.

private/buf/cmd/buf/command/push/push_test.go
```diff
@@ -229,13 +143,83 @@ func TestBucketBlobs(t *testing.T) {
...
+func createServer(t *testing.T, mock *mockPushService) *httptest.Server {
+	t.Helper()
 	mux := http.NewServeMux()
 	mux.Handle(
 		registryv1alpha1connect.NewPushServiceHandler(mock),
 	)
-	return httptest.NewServer(mux), mock
+	server := httptest.NewServer(mux)
+	t.Cleanup(func() {
+		server.Close()
```
unmultimedio, line 220: 👍🏼

private/bufpkg/bufapimodule/module_reader.go
```diff
@@ -61,15 +61,15 @@ func (m *moduleReader) GetModule(ctx context.Context, modulePin bufmoduleref.Mod
...
 	}
-	resp, err := m.download(ctx, modulePin)
-	if err != nil {
-		return nil, err
-	}
 	identityAndCommitOpt := bufmodule.ModuleWithModuleIdentityAndCommit(
 		moduleIdentity,
 		modulePin.Commit(),
 	)
 	if m.tamperProofingEnabled {
```
unmultimedio, line 68: Curious, when is this set? I assume whenever `moduleReader` is instantiated. If user sets/unsets the ENV value, does this get updated?

unmultimedio (commented):
lgtm! a few comments/questions

private/buf/cmd/buf/command/push/push.go
pkwarren: Good point on the `if err != nil` - updated to better report errors to the user so we don't switch behavior silently if the env var contains an invalid value. On the `if ...` logging, I'd prefer we wait on this (I expect us to provide this to a smaller number of users and then probably remove the flag once we feel ok).

private/bufpkg/bufapimodule/module_reader.go
pkwarren: This is from a previous PR but it gets set here: https://github.com/bufbuild/buf/blob/7d03938f0f3da7d3d2c12402c87df8f89f376cd2/private/bufpkg/bufapimodule/module_reader.go#L34-L39

When re-running the CLI, it will get the current value of the ENV value. It has no support for re-checking an env var set via https://pkg.go.dev/os#Setenv if that's what you are asking.

unmultimedio: > When re-running the CLI, it will get the current value of the ENV value.

You mean every time I run any `buf` command?

unmultimedio (approved):

unmultimedio (approved):
I'd wait for @twilly review on this though.

private/bufpkg/bufapimodule/module_reader.go
pkwarren: > You mean every time I run any `buf` command?

Yep!

twilly (approved):
Looks good. FYI the changes in the push tests are fine. I was minimizing spinning up a server since it'll open ports and that means it could run out of ports... but that's not a realistic concern.

FYI SA1019 can be dropped from .golangci.yml given we're dropping the deprecation mark.

twilly (approved):
Goodbye exceptions.

pkwarren:
> Looks good. FYI the changes in the push tests are fine. I was minimizing spinning up a server since it'll open ports and that means it could run out of ports... but that's not a realistic concern.

Yep that's what I figured. I think macOS used to limit to 256 FDs but hopefully our CI/CD workflows set a higher limit.

> FYI SA1019 can be dropped from .golangci.yml given we're dropping the deprecation mark.

Removed.

PR #1695, opened 2022-12-20
BSR-1136 - Adds `OrganizationRoleSource`

saquibmian:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -160,6 +162,10 @@ message PayloadOrganizationMemberRoleChanged {
   buf.alpha.registry.v1alpha1.OrganizationRole old_role = 3;
   // new_role is the new role of the member whose role was changed.
   buf.alpha.registry.v1alpha1.OrganizationRole new_role = 4;
+  // old_member_role_source is the old source of the role granted to the member.
+  buf.alpha.registry.v1alpha1.OrganizationRoleSource old_member_role_source = 5;
+  // new_member_role_source is the new source of the role granted to the member.
+  buf.alpha.registry.v1alpha1.OrganizationRoleSource new_member_role_source = 6;
```
unmultimedio, line 168: So an organization membership could be created by one source and then modified by another source?

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -49,6 +49,7 @@ message OrganizationUser {
   string organization_id = 2;
   // The role that the user has in the organization above.
   OrganizationRole organization_role = 3;
+  OrganizationRoleSource organization_role_source = 4;
```
unmultimedio, line 53: Just curious, in which scenarios do we need to expose this in RPC responses?

proto/buf/alpha/audit/v1alpha1/event.proto
saquibmian: Yes. There are two scenarios I can think of:

1. The user is added automatically by a group (source is `idp_group`), but then later someone else upgrades them to an admin (source will become `direct`).
2. The user is added manually first (source is `direct`) at a role _below the org default role_, if the next time they log in they are part of a group that's mapped to the org, they will be upgraded (source will be `idp_group`).

Basically, last write wins.

proto/buf/alpha/registry/v1alpha1/user.proto
saquibmian: I am thinking the UI will want to see this to power scenarios where we can show why a user is part of an org (e.g., they are part of a group).

unmultimedio: I ask because we'd need to check in the backend everytime we're moving a user proto and populate that value.

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/user.proto
saquibmian: Yeah, I'll have to do this up.

PR #1696, opened 2022-12-21
BSR-1094 - Adds IDP Group RPCs

saquibmian:
Groups are associated with a BSR Organization and will be used to automatically manage membership.


proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -45,6 +45,8 @@ message OrganizationMembership {
   OrganizationRole organization_role = 2;
 }
 
+message OrganizationIDPGroup {}
```
saquibmian, line 48: D'oh, this shouldn't have been committed.

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -73,6 +73,10 @@ service OrganizationService {
   rpc GetOrganizationSettings(GetOrganizationSettingsRequest) returns (GetOrganizationSettingsResponse);
   // UpdateOrganizationSettings update the organization settings including base roles.
   rpc UpdateOrganizationSettings(UpdateOrganizationSettingsRequest) returns (UpdateOrganizationSettingsResponse);
+  // AddOrganizationGroup adds an IdP Group to the organization.
```
saquibmian, line 76: I specifically did not include these in the `UpdateOrganizationSettings` RPC as that RPC is all-or-nothing, which is a poor interaction pattern. We should be using more scoped mutation RPCs.

mfridman: We should be cautious how many individual purpose-built RPC's we add where a single bulk one would suffice.

Having said that, it really depends how the FE / design UI will be built. Okay to merge this, but do loop in FE / design.

mfridman (approved):

PR #1698, opened 2022-12-22
Adds "buf curl" command

jhump:
This is a big one. It adds a new sub-command that has the functionality of grpcurl.

Main differences from grpcurl
1. Interface is much closer to `curl` -- the command-line usage and flags and the verbose output all more closely match.
2. Supports grpcweb and connect protocols.
3. Ability to indicate filenames for reading headers; it's even possible to combine headers and request body JSON payloads in a single file.
4. Instead of indicating proto descriptor sets or proto sources and import directories, one can specify schema in the same form as the `<input>` positional arg for `build` and `generate` commands. It can indicate a proto descriptor set, a directory of sources, or a remote module (BSR repo name). If a `buf.yaml` file is present, it will respect configuration therein. It accepts numerous flags related to defining the schema when not using server reflection (all match eponymous flags in the `build` and `generate` sub-commands).
5. Prefers v1 of gRPC reflection, falling back to v1alpha. (FWIW, grpcurl just uses v1alpha, at least as of today.)

I did a lot of testing against demo.connect.build, especially since it supports all three protocols (connect, grpc, and grpcweb).

But I also did testing using test servers I had written in the past:
1. `github.com/fullstorydev/grpcurl/internal/testing/cmd/testserver`: this was developed when testing `grpcurl`, since it allows for testing many scenarios, mostly around TLS and unix sockets.
2. `github.com/fullstorydev/grpcui/testing/cmd/testsvr`: this was developed to test all permutations of message types and response types in `grpcui`. I used it because it is easy to customize error output of the server, including error details (to test error output of `buf curl`).

Both of the above support only the grpc protocol. Neither supporst v1 of the grpc reflection service, only v1alpha. So this also allowed me to test the reflection "auto" version (which tries v1 and then falls back to v1alpha).

I have not added automated tests. All testing was done via exploratory testing. Input is welcome on approaches to automated tests for a lot of this.

One thing I noticed that I have not fixed yet: most connection-related errors result in a terrible error message of `write envelope: EOF`. This is in the Connect framework, where it gets an error trying to write an envelope for the request message. This seems like a bug in the Connect framework -- that's not just a terrible error message for buf CLI users, but it is also a terrible error message for users of the connect-go library.

Fixes TCN-773
Fixes TCN-774
Fixes TCN-775
Fixes TCN-776


private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -0,0 +1,832 @@
...
+If headers and the request body are both to be read from the same file (or both
+read from stdin), the file must include headers first, then a blank line, and
+then the request body.
+
+The URL can use either http or https as the scheme. If http is used then HTTP
+1.1 will be used unless the --http2-prior-knowledge flag is set. If https is
+used then HTTP/2 will be preferred during protocol negotiation and HTTP 1.1 used
+only if the server does not support HTTP/2.
+
+The default RPC protocol used will be gRPC. However this protocol cannot work
```
elliotmjackson, line 143: why not default to connect?

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -0,0 +1,831 @@
...
+		reflectBaseURL := baseURL
+		if f.ReflectBaseURL != "" {
+			reflectBaseURL = f.ReflectBaseURL
+		}
+		reflectHeaders, _, err := loadHeaders(f.ReflectHeaders, "", requestHeaders)
+		if err != nil {
+			return err
+		}
+		res, closeRes = resolverFromReflection(ctx, transport, clientOptions, reflectBaseURL, f.ReflectVersion, reflectHeaders, container.VerbosePrinter())
+		defer closeRes()
```
elliotmjackson, line 661: this defer statement, paired with the hard call of `closeRes()` [below](https://github.com/bufbuild/buf/pull/1698/files#diff-05c240e5d4191f595d5b747c112c314cf14ee4d77505cee99aac9500eed39527R717) resets the stream twice, It LOOKS safe to me but just want to make sure both are intentional

private/buf/cmd/buf/command/curl/curl.go
jhump: TBH, 99.9% of requests using the Connect protocol do not really need a `buf curl` tool: it was designed to be cURL-friendly and supports JSON out of the box. The main value of this seemed to be the fact that it can get the schema and then translate human-friendlier JSON to binary protobuf, which is mainly needed for grpc and grpcweb servers.

Though it's true that it does also provide a much easier way to do streaming than with cURL (which would require some shenanigans to wrap each JSON value in an envelope).

I also worried that Connect doesn't have enough adoption to be the default. I think most people making use of this feature would be targeting gRPC servers. I wonder what @akshayjshah thinks.

private/buf/cmd/buf/command/curl/curl.go
jhump: Yeah, intentional. I will clarify with a comment.

1. It is safe to call twice, even if it's not needed. So we call it eagerly to release the reflection stream.
2. If we need to unmarshal an Any or an extension in the JSON input, it will create a new reflection stream to resolve the descriptors for those. So the defer is necessary to make sure it shuts that stream down gracefully.

I mainly did this to make the output look cleaner if you use the `-v` flag. Without the first call, it's a bit confusing that reflection stream doesn't complete until _after_ the main RPC completes. However, if it does need to re-create a stream (like to unmarshal an unrecognized message in an Any or if it encounters an unknown extension), then that still happens. So maybe it's better to just always rely on defer (and then there are fewer reflection requests). WDYT?

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -0,0 +1,832 @@
...
+		"",
+		`The user agent string to send.`,
+	)
+	flagSet.StringSliceVarP(
+		&f.Headers,
+		headerFlagName,
+		"H",
+		nil,
+		`Request headers to include with the RPC invocation. This flag may be specified more `+
+			`than once to indicate multiple headers. Each flag value should have the form key=value. `+
```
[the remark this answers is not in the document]
jhump: Oops! My bad.

private/buf/cmd/buf/command/curl/curl.go
bufdev: I've gone back and forth on this, but I'd probably either (1) have no default or (2) default to connect, although I understand your motivations (and don't think they're off-base). happy to touch base briefly offline

private/buf/cmd/buf/command/curl/curl.go
jhump: I removed the explicit close. Now it's only closed from the defer. Since I added the `[#2]` snippet to the request tracing, it's more clear that the final call completion is the reflection, so not as confusing as when I first added that call.

private/buf/cmd/buf/command/curl/curl.go
jhump: Changed default to connect

elliotmjackson (approved):

bufdev:
From conversation:

- Add example usages, i.e `buf curl --schema ./proto, buf curl --schema buf.build/whatever, --protocol grpc, --whatever-else` to near top of usage.
- Add documentation to docs.buf.build?
- We already removed build-specific flags
- Change `--reflect-version` to `--reflect-protocol`
- Change protocols to `grpc-v1, grpc-v1alpha`
- Change meaning of "auto" to be the latest reflection protocol for whatever protocol you choose. This means in the future this can use a theoretical connect reflection protocol.
- Remove/comment out `--reflect-base-url` for now

bufdev:
@jhump I made two slight edits:

- Changed the wording of the help text for `--reflect` from "grpc server reflection" to "server reflection", to match the main command help text and for futureproofing.
- Added "Setting this flag implies ---reflect=false" to the help text for `--schema`.

Otherwise, looks great!! Let's merge it!

bufdev (approved):

PR #1701, opened 2022-12-23
Managed mode PHP except override

oliversun09:


private/bufpkg/bufimage/bufimagemodify/objc_class_prefix.go
```diff
@@ -78,7 +78,7 @@ func objcClassPrefix(
 			}
 			for moduleIdentityString := range overrideModuleIdentityStrings {
 				if _, ok := seenModuleIdentityStrings[moduleIdentityString]; !ok {
-					logger.Sugar().Warnf("%s override for %q was unused", ObjcClassPrefixID, moduleIdentityString)
+					logger.Sugar().Warnf("objc_class_prefix override for %q was unused", moduleIdentityString)
```
oliversun09, line 81: 😅

oliversun09:
A couple of followups that may or may not be worth doing:
1. order the fields in a meaningful way since there are quite a few of them
2. group the `AssertConfigsWithSame*` test helpers

prochac:
Hi, we use GitLab and its subdirs, and it doesn't work with this branch properly. 

```
➜  grpc git:(php) ✗ buf generate    
Failure: invalid php_namespace override key: module identity "gitlab.com/our-org/some-subdir/the-repo" is invalid: must be in the form remote/owner/repository
```

doriable:
Thanks for the heads-up, @prochac! I'll definitely make sure that we fix this up and test extensively before merging/release :)

doriable:
@prochac In this case, the module identity refers to a Buf module -- so the `remote/owner/repository` refers to the Buf Schema Registry remote, owner, and repository. The override is unfortunately not meant for Git references in this case.

doriable:
Tested locally, will need to make a small change in this case, since the `php_metadata_namespace` is currently not respecting overrides, which I believe it should.

doriable:
After reviewing this in further detail, I think given the unclear behaviour this should have with `php_metadata_namespace`, I'm closing this for now. If we get future requests for more detailed PHP file option overrides, I am happy to help with that.

PR #1703, opened 2022-12-24
Use protoencoding for marshal/unmarshal in bufcurl and turn off pretty printing

bufdev:
I was working through examples for `buf curl`, and noticed that it pretty-prints the output JSON. Whether we should do this or not is worthy of a quick conversation, but the `buf` tool's overall stance (including for `--error-format=json`) has been to print compact JSON, and leave it to the user to format their JSON with other tools if they want, sticking with the unix philosophy of doing one thing well, i.e. we give the user the most compact output possible, and they can use `jq` to format if they want:

```
$ buf curl https://demo.connect.build/buf.connect.demo.eliza.v1.ElizaService/Say -d '{"sentence":"Hello."}'
{"sentence":"Hello...I'm glad you could drop by today."}

$ buf curl https://demo.connect.build/buf.connect.demo.eliza.v1.ElizaService/Say -d '{"sentence":"Hello."}' | jq
{
  "sentence": "Hello, how are you feeling today?"
}
```

In going through the code to put up the proposal to make this change, I noticed that `bufcurl` is using `proto.Umarshal/Marshal` and `protojson` directly, which we want to avoid - we've strived to use `protoencoding` everywhere so we have the same behavior for Protobuf encoding everywhere, and can easily change it later. Example here being we always want to use deterministic wire marshaling, and we want to call `json.Compact` on the result for JSON marshaling due to the `protojson` purposefully non-determinstic JSON marshalling issue (and if this were to ever be fixed, we could easily change the usage of `json.Compact` in one place). This updates the code to use `protoencoding` everywhere in `bufcurl`, and I'll have a follow-up to do this in `bufstudioagent` (the only other place in bufbuild/buf that doesn't use `protoencoding`) as well.

Let me know your thoughts especially re: pretty-printing, we should come to agreement on that 100%.


jhump:
What other places are we using JSON? Are those places mainly for machines to consume? Since this is a debugging tool, it's likely that a human is examining the output. So I don't understand why we'd require users to use `jq` and provide another flag for the likely common case.

For the examples on the help (which are simple messages with a single string field), compact is fine. But as soon as you have anything more interesting/complex, the output becomes rather hard to read. I think a major point of providing a JSON interface, and converting to binary on the wire, was to aid human consumers.

bufdev:
Everywhere we're already using JSON is primarily for human consumption, the main places being:

- `--error-format=json` - when we output errors as JSON. Arguably for "machine" consumption sometimes, but this is also how users get detailed output.
- `buf build -o -#format=json` - outputting a FileDescriptorSet as JSON. This is almost always for human consumption (I use this quite often for that purpose even), and if I want it pretty-printed, I use `jq`
- `buf convert` - a very similar type of command to `buf curl` - will print compact JSON if converting to JSON

The argument is really two-fold:

- Unix philosophy of doing one thing well - we don't need to worry about formatting JSON, that is for another tool if you want it. Many users will use this for debugging, but others could use this for scripting. It's not `buf`'s job to format JSON, it's `buf`'s job to invoke the RPC and provide output for other tools to use.
- Consistency within the tool - other parts of the tool were already built with compact JSON, so it would be nicer for all commands to be consistent instead of one command to make a different choice.

private/buf/bufcurl/invoker.go
```diff
@@ -276,18 +276,16 @@ func (inv *invoker) handleResponse(data []byte, msg *dynamicpb.Message) error {
 	if msg == nil {
 		msg = dynamicpb.NewMessage(inv.md.Output())
 	}
-	if err := (proto.UnmarshalOptions{Resolver: inv.res}).Unmarshal(data, msg); err != nil {
+	if err := protoencoding.NewWireUnmarshaler(inv.res).Unmarshal(data, msg); err != nil {
 		return err
 	}
-	outputBytes, err := (protojson.MarshalOptions{Resolver: inv.res}).Marshal(msg)
+	// If we want to add a pretty-print option, we should perhaps make this a flag,o
+	// and use protoencoding,NewJSONMarshalerIndent if so
```
jhump, line 283:
```suggestion
	// If we want to add a pretty-print option, we should perhaps make this a flag
	// and use protoencoding.NewJSONMarshalerIndent
```

jhump:
If most JSON output is meant for human consumption, I don't understand why we require that users install something else (`jq`) to make the output readable.

As far as Unix philosophy, I'm certainly not arguing that we put all of `jq` functionality into `buf` -- `jq` is *way more* than a formatter. Formatting of the output seems like a sane responsibility of the _producer_ of the output.  And it's like one line of code to have the default behavior be what is likely the best experience for users.

As far as consistency, is there a downside to consistently pretty-printing everything instead of consistently not? If a human were even to _notice_ that the formatting is inconsistent, is the assumption that they'd actually prefer it to be always compact (vs. always pretty-printed)?

jhump (approved):
I've voiced my concerns and am not really sold that consistency trumps usability. But if my responses aren't compelling, go ahead and merge it -- I don't feel strongly enough to object further.

PR #1704, opened 2022-12-24
Use protoencoding for marshalling and unmarshalling

bufdev:
See https://github.com/bufbuild/buf/pull/1703 for the background here - this, plus #1703, will make it so that all of our marshalling/unmarshalling of Protobuf payloads goes through one common/consistent/vetted location.


doriable (approved):
lgtm

PR #1705, opened 2022-12-24
Remove auto value from reflect protocol and make unset the default

bufdev:
This changes the definition of `buf curl --reflect-protocol` slightly to make it such that this flag no longer has an `auto` value. Instead, this flag is defined as "use a specific reflection protocol. if no protocol is set, we will choose". This means there is no special value needed to be understood by our users - the default value for `--reflect-protocol` is no value.

As part of this change, this changes the `ReflectProtocol` type to be an enum in line with the rest of the codebase and our code standards (although we should make it more clear in our code standards, its a little obscure to be fair). We prefer `int` enums over `strings` because they do not allow people to do things such as cast user-provided flag values to their type directly :-) enums should be just that, and parsing should be commonly validated in the package that provides them. `bufanalysis` is a good example of this style (that `bufcurl` is now largely updated to reflect).

### DIFF
--- a/private/buf/bufcurl/reflection_resolver.go
+++ b/private/buf/bufcurl/reflection_resolver.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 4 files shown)


jhump (approved):
Seems fine 🤷

But I have to admit, it's not really clear to me that it provides much value. Is dogma around consistency really a good thing if it requires >100 LOC of boiler-plate for an enum of two values?

> We prefer int enums over strings because they do not allow people to do things such as cast user-provided flag values to their type directly :-)

Debatable that this is actually a desirable property :P
But this is not a hill I'm willing to die on, so I'll approve.

PR #1710, opened 2023-01-03
don't use remote package

jhump:
To avoid a dependency on the BSR when _building_ the BSR (which would greatly hamper efforts to fix the BSR when it's down), we need to remove the use of remote packages from this Go module.

While changing this, I also found a bug in the `--include-types` functionality of `buf generate`: the flag would be ignored if you didn't have a `types` stanza in your `buf.gen.yaml`. This fixes that bug and makes use of the feature, to avoid generating unneeded stuff in the [buf.build/grpc/grpc](https://buf.build/grpc/grpc) repo.


make/buf/all.mk
```diff
@@ -107,6 +107,7 @@ bufgenerateclean:: bufgeneratecleango
 .PHONY: bufgenerateprotogo
 bufgenerateprotogo:
 	$(BUF_BIN) generate proto --template data/template/buf.go.gen.yaml
+	$(BUF_BIN) generate buf.build/grpc/grpc --include-types grpc.reflection.v1.ServerReflection --template data/template/buf.go.gen.yaml
```
robbertvanginkel, line 110: 🎉

robbertvanginkel (approved):

PR #1711, opened 2023-01-04
Add plugin delete command

mfridman:
This PR adds a `buf alpha plugin delete ...` command to delete a curated plugin from the BSR registry.

### DIFF
--- a/private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
@@ -40,7 +40,7 @@ func NewCommand(
 	return &appcmd.Command{
 		Use:   name + " <buf.build/owner/plugin[:version]>",
 		Short: "Delete a plugin from the registry.",
-		Args:  cobra.MaximumNArgs(1),
+		Args:  cobra.ExactArgs(1),
 		Run: builder.NewRunFunc(
 			func(ctx context.Context, container appflag.Container) error {
 				return run(ctx, container, flags)
@@ -70,6 +70,11 @@ func run(
 	if err != nil {
 		return appcmd.NewInvalidArgumentError(err.Error())
 	}
+	if version != "" {
+		if err := bufpluginref.ValidatePluginVersion(version); err != nil {
+			return err
+		}
+	}
 	clientConfig, err := bufcli.NewConnectClientConfig(container)
 	if err != nil {
 		return err
@@ -91,9 +96,6 @@ func run(
 			return fmt.Errorf("the plugin %s does not exist", container.Arg(0))
 		}
 		return err
-	}
-	if _, err := fmt.Fprintln(container.Stdout(), "Plugin deleted."); err != nil {
-		return bufcli.NewInternalError(err)
 	}
 	return nil
 }
--- a/private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
@@ -72,7 +72,7 @@ func run(
 	}
 	if version != "" {
 		if err := bufpluginref.ValidatePluginVersion(version); err != nil {
-			return err
+			return appcmd.NewInvalidArgumentError(err.Error())
 		}
 	}
 	clientConfig, err := bufcli.NewConnectClientConfig(container)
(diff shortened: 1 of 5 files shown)


private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
```diff
@@ -0,0 +1,99 @@
...
+			func(ctx context.Context, container appflag.Container) error {
+				return run(ctx, container, flags)
+			},
+			bufcli.NewErrorInterceptor(),
+		),
+		BindFlags: flags.Bind,
+	}
+}
+
+type flags struct{}
```
mfridman, line 54: I didn't add a force flag here, but this is something we may add later on so kept the flag bits around.

private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (f *flags) Bind(flagSet *pflag.FlagSet) {}
+
+func run(
+	ctx context.Context,
+	container appflag.Container,
+	flags *flags,
+) error {
+	bufcli.WarnAlphaCommand(ctx, container)
+	identity, version, _ := strings.Cut(container.Arg(0), ":")
```
pkwarren, line 68: In other places, we do things like this to either parse as an identity or reference:

```go
var pluginIdentity bufpluginref.PluginIdertity
pluginVersion := ""
if identity, err := bufpluginref.PluginIdentityForString(container.Arg(0)); err == nil {
  pluginIdentity = identity
} else if reference, err := bufpluginref.PluginReferenceForString(container.Arg(0)); err == nil {
  pluginIdentity = reference
  pluginVersion = reference.Version()
} else {
  // error - not a valid plugin identity or reference
}
```

This will avoid passing invalid semver versions to the BSR.

private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
```diff
@@ -0,0 +1,99 @@
...
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <buf.build/owner/plugin[:version]>",
+		Short: "Delete a plugin from the registry.",
+		Args:  cobra.MaximumNArgs(1),
```
pkwarren, line 43:
```suggestion
		Args:  cobra.ExactArgs(1),
```

private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
```diff
@@ -0,0 +1,99 @@
...
+			Name:    pluginIdentity.Plugin(),
+			Version: version,
+		}),
+	); err != nil {
+		if connect.CodeOf(err) == connect.CodeNotFound {
+			return fmt.Errorf("the plugin %s does not exist", container.Arg(0))
+		}
+		return err
+	}
+	if _, err := fmt.Fprintln(container.Stdout(), "Plugin deleted."); err != nil {
```
pkwarren, line 95: In the existing `buf beta plugin delete`, we don't output anything on success. What do you think about keeping this behavior (or just logging this as a debug log).

mfridman: I was making this consistent with repository delete, where we output repository deleted.

https://github.com/bufbuild/buf/blob/f09b93d69e8b13d018a9ddc977690fbb5f438f74/private/buf/cmd/buf/command/beta/registry/repository/repositorydelete/repositorydelete.go#L107

I can remove, no strong opinion on output.

private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
mfridman: We can't use `PluginReferenceForString` because it requires a revision, and we don't expect the user to know the revision. I suppose this could always be a number like `1`, but that's not great.

Any objections if we add a `if version != "" { semver.IsValid }` check?

Fwiw I think we should do as little validation on the CLI and let the server-side validation reject the input.

mfridman: Instead of duplicating the validation, I exported them from the `bufpluginref`.

private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
mfridman: Removed the output, exit code 0 ftw.

pkwarren (approved):

PR #1713, opened 2023-01-05
Make silent runners faster

akshayjshah:
If we're using `private/pkg/command` to run a command silently, sending
stdout and stderr to `io.Discard` is needlessly expensive: it creates a
pipe and spawns a goroutine to copy from the pipe to the writer, which
then throws the data away. If we leave stdout and stderr as nil,
`os/exec` instead connects the process directly to `/dev/null`, which is
functionally identical and cheaper.

See [the os/exec.Command docs](https://pkg.go.dev/os/exec#Cmd:~:text=Otherwise%2C%20during%20the%20execution%20of%20the%20command%20a%20separate%20goroutine) for details.

### DIFF
--- a/private/pkg/command/command.go
+++ b/private/pkg/command/command.go
@@ -70,7 +70,7 @@ func RunWithStdin(stdin io.Reader) RunOption {
 
 // RunWithStdout returns a new RunOption that sets the stdout.
 //
-// The default is io.Discard.
+// The default is the null device (os.DevNull).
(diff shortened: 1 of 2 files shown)


akshayjshah:
In the spirit of fresh starts to the year, I'm trying to clear out my backlog of small todos...and this one's been on my list for quite a while. Not sure who's the best reviewer, so anyone's welcome to jump in!

robbertvanginkel (approved):

PR #1714, opened 2023-01-05
update makego

pkwarren:


private/buf/cmd/buf/command/alpha/plugin/plugindelete/plugindelete.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2020-2022 Buf Technologies, Inc.
+// Copyright 2020-2023 Buf Technologies, Inc.
```
pkwarren, line 1: FYI @joshcarp - these two newly added files were failing the build - looks like they were missed in copyright update.

joshcarp (approved):

joshcarp (approved):
thx

PR #1724, opened 2023-01-11
Add support for module tamper proofing

pkwarren:
Update buf CLI to support a beta feature (`BUF_BETA_ENABLE_TAMPER_PROOFING`) which enables supports for module tamper proofing. This requires BSR support and is disabled by default.

If enabled, the `buf push` and download service calls will be updated to send modules as a manifest and set of blobs (instead of a buf.alpha.module.v1alpha1.Module).


private/pkg/manifest/module.go
```diff
@@ -0,0 +1,321 @@
...
+// the digest has no blob in the set.
+func (s *BlobSet) BlobFor(digest string) (Blob, bool) {
+	blob, ok := s.digestToBlob[digest]
+	if !ok {
+		return nil, false
+	}
+	return blob, true
+}
+
+// Blobs returns the blobs in the set as an array.
```
pkwarren, line 204:
```suggestion
// Blobs returns a slice of the blobs in the set.
```

private/pkg/manifest/module.go
```diff
@@ -0,0 +1,321 @@
...
+	}, nil
+}
+
+// BlobEqual returns true if blob a is the same as blob b. The digest is
+// checked for equality and the content bytes compared.
+//
+// An error is returned if an unexpected I/O error occurred when opening,
+// reading, or closing either blob.
+func BlobEqual(ctx context.Context, a, b Blob) (_ bool, retErr error) {
+	const blockSize = 4098
```
pkwarren, line 259: Any reason we're using this block size? Should it be documented?

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,243 @@
...
+		if err != nil {
+			return err
+		}
+		defer func() { retErr = multierr.Append(retErr, obj.Close()) }()
+		blob, err := NewMemoryBlobFromReaderWithDigester(obj, digester)
+		if err != nil {
+			return err
+		}
+		blobs = append(blobs, blob)
+		if err := m.AddEntry(path, *blob.Digest()); err != nil {
```
pkwarren, line 71: nit: `return m.AddEntry(...)`.

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,243 @@
...
+	allManifestBlobs bool
+	noExtraBlobs     bool
+}
+
+// BucketOption are options passed when creating a new manifest bucket.
+type BucketOption func(*bucketOptions)
+
+// BucketWithAllManifestBlobsValidation validates that all manifest digests in
+// have a corresponding blob in the blob set. If this option is not passed, then
+// buckets with partial/incomplete blobs is allowed.
```
pkwarren, line 95:
```suggestion
// buckets with partial/incomplete blobs are allowed.
```

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,216 @@
...
+	}
+	return paths
+}
+
+// PathsFor returns one or more matching path for a given digest. The digest is
+// expected to be a lower-case hex encoded value. Returned paths are unordered.
+// Paths is nil and ok is false if no paths are found.
+func (m *Manifest) PathsFor(digest string) ([]string, bool) {
+	paths, ok := m.digestToPaths[digest]
+	if !ok {
```
pkwarren, line 153:
```suggestion
	if !ok || len(paths) == 0 {
```

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,243 @@
...
+				return nil, fmt.Errorf("path %q not present in manifest", path)
+			}
+			if _, ok := blobs.BlobFor(pathDigest.String()); !ok {
+				return nil, fmt.Errorf("manifest path %q with digest %q has no associated blob", path, pathDigest.String())
+			}
+		}
+	}
+	if config.noExtraBlobs {
+		for digestStr := range blobs.digestToBlob {
+			if _, ok := m.digestToPaths[digestStr]; !ok {
```
pkwarren, line 131: This should either use `m.PathsFor(digestStr)` or additionally validate that `len(paths) > 0`.

private/bufpkg/buflock/buflock.go
```diff
@@ -100,6 +108,7 @@ func ExternalConfigDependencyV1ForDependency(dep Dependency) ExternalConfigDepen
 		Owner:      dep.Owner,
 		Repository: dep.Repository,
 		Commit:     dep.Commit,
+		Digest:     dep.Digest,
```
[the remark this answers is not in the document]
unmultimedio: No cluster should be sending back any digest now, since it was disabled, and now in the new branch it's using the new field that stores the manifest digest. Although doing the check doesn't do any harm, I'd add it.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,216 @@
...
+
+var _ error = (*Error)(nil)
+
+func newError(lineno int, msg string) *Error {
+	return &Error{
+		lineno: lineno,
+		msg:    msg,
+	}
+}
+func newErrorWrapped(lineno int, err error) *Error {
```
unmultimedio, line 53: Nit
```suggestion
}

func newErrorWrapped(lineno int, err error) *Error {
```

private/pkg/manifest/module.go
```diff
@@ -0,0 +1,321 @@
...
+
+// NewMemoryBlob takes a digest and a content, and turns it into an in-memory
+// representation of a blob, which returns the digest and an io.ReadCloser for
+// its content.
+func NewMemoryBlob(digest Digest, content []byte, opts ...MemoryBlobOption) (Blob, error) {
+	var config memoryBlobOptions
+	for _, option := range opts {
+		option(&config)
+	}
+	if config.validateHash {
```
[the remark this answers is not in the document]
unmultimedio: Related: https://github.com/bufbuild/buf/pull/1558#discussion_r1014503044, sometimes we don't need to validate: https://github.com/bufbuild/buf/pull/1724/files#diff-f677e8abaa91a1b8f8a253e1fc2a6a581a43c657f253a8dc57926a8e3856813aR78

private/pkg/manifest/module.go
```diff
@@ -0,0 +1,321 @@
...
+	}
+	rc, err := b.Open(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("cannot open blob: %w", err)
+	}
+	content, err := io.ReadAll(rc)
+	if err != nil {
+		return nil, fmt.Errorf("cannot read blob contents: %w", err)
+	}
+	if err := rc.Close(); err != nil {
```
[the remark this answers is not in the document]
unmultimedio: agree

private/pkg/manifest/module.go
```diff
@@ -0,0 +1,321 @@
...
+			return false, aErr
+		}
+		if bErr != nil && bErr != io.EOF {
+			// unexpected read error
+			return false, bErr
+		}
+		if !bytes.Equal(aBlock[:aN], bBlock[:bN]) {
+			// Read content doesn't match. Don't forget to forward any errors
+			// returned.
+			return false, multierr.Append(nilEOF(aErr), nilEOF(bErr))
```
unmultimedio, line 298: seems to me here that `multierr.Append(nilEOF(aErr), nilEOF(bErr))` won't ever append any err? If EOF is ignored, and if other, it should have been returned in the previous 2 conditionals?

private/pkg/manifest/storage.go
unmultimedio: Good catch! Agree, we can use `blobs.Blobs` in the loop, and check `len(m.PathsFor) > 0` in the conditional.

unmultimedio:
push/download logic, and the `manifest` pkg looks good to me. the module builder part and the lock file populating the digest, I'd get other teammate to check that 😄

private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
```diff
@@ -53,22 +51,13 @@ func WithWorkspace(workspace bufmodule.Workspace) BuildModuleFileSetOption {
...
-	// If paths is empty, all files are built.
-	// Paths should be relative to the bucket, not the roots.
-	BuildForBucket(
-		ctx context.Context,
-		readBucket storage.ReadBucket,
-		config *bufmoduleconfig.Config,
-		options ...BuildOption,
-	) (bufmodule.Module, error)
-}
+type ModuleBucketBuilder = *moduleBucketBuilder
```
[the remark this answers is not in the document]
twilly: It's so much simpler to not have an interface exported. Keeping the interface is just code churn for, IMHO, no real benefit, along with real drawbacks (breaking other implementations, requiring recasting if a struct providing a particular interface does more than said interface).

This type alias keeps compatibility and later we can refactor to export `struct moduleBucketBuilder`.

private/pkg/manifest/module.go
twilly: It's a number. Is it a reasonable one? I don't know. It's one disk block size. It's less than one Ethernet frame. 🤷 

IMHO I wouldn't document it in the function docs; it's not behavior that should be relied upon. Maybe as a note on this line that it's open for tweaking?

private/pkg/manifest/module.go
twilly: I think you're right! I had this block in a different location before. I must have forgot about this when I moved it to where it is now.

```suggestion
			// Read content doesn't match.
			return false, nil
```

private/bufpkg/buflock/buflock.go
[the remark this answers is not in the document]
pkwarren: I'll leave it out so we have less business logic in the CLI.

twilly:
I think this is good for landing. I'm a +1 and will stamp after @jhump or @johanbrandhorst take a look.

private/pkg/manifest/manifest.go
```diff
@@ -0,0 +1,217 @@
...
+	"fmt"
+	"io"
+	"sort"
+	"strings"
+)
+
+var errNoFinalNewline = errors.New("partial record: missing newline")
+
+// Error occurs when a manifest is malformed.
+type Error struct {
```
jhump, line 39: None of this is exposed, aside from in the error message. Are there plans to add accessors? If not, do we need an exported type for this? You could also construct this using `fmt.Errorf` with `%w` directive to wrap an underlying error and add the line number and message.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -20,6 +20,27 @@ import "buf/alpha/breaking/v1/config.proto";
...
 
+enum HashKind {
+  HASH_KIND_UNSPECIFIED = 0;
+  HASH_KIND_SHAKE256 = 1;
+}
+
+// Hash represents a hash function's value.
+message Hash {
+  // kind describes the hash algorithm. e.g. "SHAKE256"
+  HashKind kind = 1;
```
jhump, line 31: Should these be called `Digest` and `DigestType` to match the names in the manifest package? Also, do we really need two different representations -- plain Go types/structs _and_ protos? Why aren't the generated protos enough?

proto/buf/alpha/registry/v1alpha1/module.proto
```diff
@@ -32,13 +32,11 @@ message LocalModuleReference {
...
 // It does not include a remote.
 message LocalModulePin {
-  reserved 5;
-  reserved "digest";
-
   string owner = 1;
   string repository = 2;
   string branch = 3;
   string commit = 4;
+  string digest = 5;
```
jhump, line 39: This is usually an anti-pattern in proto, to try to re-use reserved fields. What was it previously used for? Are we certain there will be no serialized protos around that might be using the old field and get incorrectly interpreted as the new one? If that did happen, what is the consequence?

jhump (commented):
I have completed one pass over this beast. I took a couple of notes on things I went to go look at in more detail, so I will try to do one more round this afternoon.

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,240 @@
...
+type bucketOptions struct {
+	allManifestBlobs bool
+	noExtraBlobs     bool
+}
+
+// BucketOption are options passed when creating a new manifest bucket.
+type BucketOption func(*bucketOptions)
+
+// BucketWithAllManifestBlobsValidation validates that all manifest digests in
+// have a corresponding blob in the blob set. If this option is not passed, then
```
jhump, line 91:
```suggestion
// BucketWithAllManifestBlobsValidation validates that all manifest digests
// have a corresponding blob in the blob set. If this option is not passed, then
```

jhump (commented):
I've double-checked, and it appears in the module reader that we are not actually _validating_ the digests (e.g. recomputing them to make sure everything matches the digests downloaded from the server). Is that intentional? Or did I miss something where that _is_ happening?

private/pkg/manifest/storage.go
```diff
@@ -0,0 +1,240 @@
...
+// bucket that contains the files in the manifest.
+func NewBucket(m Manifest, blobs BlobSet, opts ...BucketOption) (storage.ReadBucket, error) {
+	var config bucketOptions
+	for _, option := range opts {
+		option(&config)
+	}
+	if config.allManifestBlobs {
+		for _, path := range m.Paths() {
+			pathDigest, ok := m.DigestFor(path)
+			if !ok {
```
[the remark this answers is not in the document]
unmultimedio: We have several places for the `manifest.Paths` and the `blobset.Blobs` in which we follow this pattern. I also thought that a callback func would make this easier, should be a small change, probably an improvement after merge? cc @twilly @pkwarren

proto/buf/alpha/registry/v1alpha1/module.proto
twilly: > What was it previously used for?

It was used for digests too: b1 and b3 digests. They had "b1" and "b3" prefixes in their string form.

> Are we certain there will be no serialized protos around that might be using the old field and get incorrectly interpreted as the new one?

It gets better: are we certain there are/will-be no serialized YAML forms given these structs are mapped through proto form into/from `buf.lock`.

Answer: they can exist in pre-v1.0 releases modules. Even so, these values cannot be interpreted in our new digest form.

> If that did happen, what is the consequence?

The module with a b1/3 digest in their `buf.lock` would fail to build.

> This is usually an anti-pattern in proto, to try to re-use reserved fields.

A different field number would add wire level protection.

twilly:
> it appears in the module reader that we are not actually validating the digests (e.g. recomputing them to make sure everything matches the digests downloaded from the server). Is that intentional? Or did I miss something where that is happening?

Intentional so far since this feature isn't complete. Validation will happen at the module cache layer, which is the same location validation happens today for b3 hashes.

> On a related note: is it expected that the server will also validate (e.g. re-compute) the digests before accepting the push request?

Yes, it'll be validating on receive.

private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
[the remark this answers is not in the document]
twilly: The purpose was to support compatibility. That's less of a concern given this is landing.

I'm fine with also cutting to the chase skipping the multiple levels of indirection. Though before jumping to BuildForBucket, we should look at ModuleFileSetBuilder and ModuleIncludeBuilder. For consistency's sake they should be functions or refactored into some kind of Module Builder that can build modules, file sets, and includes.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -184,21 +247,32 @@ func run(
...
-		}
-		return err
+		return nil, err
 	}
-	if resp.Msg.LocalModulePin == nil {
-		return errors.New("Missing local module pin in the registry's response.")
+	return resp.Msg.LocalModulePin, nil
+}
+
+func manifestAndFilesBlobs(ctx context.Context, sourceBucket storage.ReadBucket) (*modulev1alpha1.Blob, []*modulev1alpha1.Blob, error) {
```
unmultimedio, line 255: Nit: This func would be useful for the backend to remove some duplicated code, could you move it to the `manifest` pkg and export it? Not blocker though.

private/pkg/manifest/digest.go
```diff
@@ -0,0 +1,137 @@
...
+func (d *shake256Digester) Digest(content io.Reader) (*Digest, error) {
+	d.hash.Reset()
+	if _, err := io.Copy(d.hash, content); err != nil {
+		return nil, err
+	}
+	digest := make([]byte, shake256Length)
+	if _, err := d.hash.Read(digest); err != nil {
+		// sha3.ShakeHash never errors or short reads. Something horribly wrong
+		// happened if your computer ended up here.
+		return nil, err
```
jchadwick-buf, line 134: nit: Maybe this should panic if it's not supposed to ever happen.

jchadwick-buf (commented):
Mostly LGTM, I don't have many comments that weren't already raised by others.

private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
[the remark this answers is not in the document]
pkwarren: I'm going to defer on any additional changes here - we can refactor as needed in a follow up.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -155,24 +161,81 @@ func run(
 		return err
 	}
 	moduleIdentity := sourceConfig.ModuleIdentity
-	module, err := bufcli.ReadModule(
+	builtModule, err := bufmodulebuild.NewModuleBucketBuilder().BuildForBucket(
```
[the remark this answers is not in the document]
pkwarren: This requires a `*bufmodulebuild.BuiltModule` while the other method returns a `bufmodule.Module`.

private/bufpkg/bufapimodule/module_reader.go
```diff
@@ -16,27 +16,90 @@ package bufapimodule
...
 	"github.com/bufbuild/connect-go"
 )
 
 type moduleReader struct {
 	downloadClientFactory DownloadServiceClientFactory
+	tamperProofingEnabled bool
+}
+
+// WithTamperProofing enables tamper proofing support (use of manifest/blobs).
+func WithTamperProofing() ModuleReaderOption {
```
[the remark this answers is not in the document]
pkwarren: Fixed.

private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
```diff
@@ -21,55 +21,64 @@ import (
...
 	"github.com/bufbuild/buf/private/pkg/normalpath"
 	"github.com/bufbuild/buf/private/pkg/storage"
 	"github.com/bufbuild/buf/private/pkg/storage/storagemem"
-	"go.uber.org/zap"
 )
 
+// BuiltModule ties a bufmodule.Module with the configuration and a bucket
+// containing just the files required to build it.
+type BuiltModule struct {
+	Module bufmodule.Module
```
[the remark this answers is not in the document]
pkwarren: Fixed.

private/pkg/manifest/manifest.go
pkwarren: Fixed.

proto/buf/alpha/module/v1alpha1/module.proto
pkwarren: Updated for symmetry with the manifest package.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
pkwarren: I ended up fixing this with a suggested change you made in a later comment.

private/buf/cmd/buf/command/push/push.go
pkwarren: Moved to `manifest.BucketToManifestAndFileBlobs`.

private/pkg/manifest/digest.go
pkwarren: Left this as-is for now. Callers can check an error - I don't want to use panic in library code if we can avoid it.

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -20,6 +20,27 @@ import "buf/alpha/breaking/v1/config.proto";
 import "buf/alpha/lint/v1/config.proto";
 import "google/protobuf/timestamp.proto";
 
+enum DigestType {
+  DIGEST_TYPE_UNSPECIFIED = 0;
+  DIGEST_TYPE_SHAKE256 = 1;
+}
+
+// Hash represents a hash function's value.
+message Hash {
```
jhump, line 29: s/Hash/Digest (for consistency with corresponding types in manifest package)

Also, I asked in previous comment but didn't get an answer: why have two separate representations (this one, generated from proto, and a hand-written version in pkg/manifest)? Seems like you could just use the generated proto messages and enums and not have to maintain hand-written types.

jhump:
> Intentional so far since this feature isn't complete. Validation will happen at the module cache layer, which is the same location validation happens today for b3 hashes.

It would be nice to add a TODO (with a reference to a linear ticket for implementing it), so that it's clear to readers that it's a known gap and so it's clear where it will happen.

private/pkg/manifest/storage.go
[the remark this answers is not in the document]
unmultimedio: Related: https://github.com/bufbuild/buf/pull/1558#discussion_r1019531518

proto/buf/alpha/registry/v1alpha1/module.proto
pkwarren: Updated to keep existing field reserved (in case they are stored on disk) and add a new field for the module digest.

proto/buf/alpha/registry/v1alpha1/module.proto
```diff
@@ -41,6 +41,6 @@ message LocalModulePin {
   string commit = 4;
   google.protobuf.Timestamp create_time = 7;
   string draft_name = 8;
-  // Module's digest (SHAKE256). Replacement for previous b1/b3 digests.
-  string module_digest = 9;
+  // Module's manifest digest. Replacement for previous b1/b3 digests.
+  string manifest_digest = 9;
```
unmultimedio, line 45: Has this message even had a field number 6? not present, not reserved.

pkwarren: Not according to history (as long as it wasn't renamed at some point).

private/buf/bufcli/bufcli.go
```diff
@@ -73,6 +74,9 @@ const (
 	alphaSuppressWarningsEnvKey = "BUF_ALPHA_SUPPRESS_WARNINGS"
 	betaSuppressWarningsEnvKey  = "BUF_BETA_SUPPRESS_WARNINGS"
 
+	// BetaEnableTamperProofingEnvKey is an env var to enable tamper proofing
+	BetaEnableTamperProofingEnvKey = "BUF_BETA_ENABLE_TAMPER_PROOFING"
```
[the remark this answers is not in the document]
unmultimedio: This is currently only accessed in-pkg. It's a const, so no sweat, but probably unexport for consistency?

unmultimedio (approved):

twilly (approved):

private/buf/bufcli/bufcli.go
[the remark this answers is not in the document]
pkwarren: It is used by a test in another package.

PR #1728, opened 2023-01-12
Update release.yaml to fetch changelog

elliotmjackson:
Small bash step to fetch changelog of current release version from changelog before putting it into the body of the release.

[example](https://github.com/elliotmjackson/buf/releases/tag/v1.12.0)

### DIFF
--- a/.github/actions/changelog-action/entrypoint.sh
+++ b/.github/actions/changelog-action/entrypoint.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+
+filename="CHANGELOG.md"
+printing=false
+#While loop to read line by line
+while IFS= read -r line; do
+  #If the line starts with ## & currently printing, disable printing
+  if [[ $line == \#\#* ]] && [[ $printing == true ]]; then
+    break
+  fi
+  # If printing is enabled, print the line.
+  if [[ $printing == true ]]; then
+    echo "$line"
+  fi
+  #If the line starts with ## & not currently printing, enable printing
+  if [[ $line == "## [$VERSION]"* ]] && [[ $printing == false ]]; then
+    printing=true
+  fi
+done <"$filename"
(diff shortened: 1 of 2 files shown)


bufdev:
Just to make sure this is covered - this does not fix the core issue. With this change, the release notes that will go out to watchers will now just be empty, instead of having the release notes that we want.

elliotmjackson:
yes, correct. Thats coming up next

.github/workflows/release.yaml
```diff
@@ -144,12 +144,14 @@ jobs:
         run: |
           unset RELEASE_MINISIGN_PRIVATE_KEY
           unset RELEASE_MINISIGN_PRIVATE_KEY_PASSWORD
+      - name: Fetch Changelog
+        run: .github/actions/changelog-action/entrypoint.sh > ${{env.VERSION}}-CHANGELOG.md
```
bufdev, line 148: There is no chance this gets published anywhere somehow correct? Ie we don't need to put this in some special temp directory?

elliotmjackson: the temp directory is the action itself which is crushed when it terminates

.github/actions/changelog-action/entrypoint.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
elliotmjackson: 😆 i think i managed to not conform with every part of the style guide

.github/workflows/release.yaml
elliotmjackson: I could delete the file in a subsequent test but i think its unnecessary

bufdev: Definitely unnecessary, just making sure

.github/actions/changelog-action/entrypoint.sh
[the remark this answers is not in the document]
bufdev: Well you're a champion of confirming to any style guide once you find it, it looks perfect now :-)

bufdev (approved):
Thanks for putting this together

PR #1729, opened 2023-01-12
Add optional name parameter to buf mod init

### LINKED ISSUE #1648: Optionally allow module name in `buf mod init` command
It'd be nice to have `buf mod init` take an optional arg or flag for the module name. E.g.,

```
buf mod init buf.build/mfridman/api
```

Or 

```
buf mod init --name buf.build/mfridman/api
```

tonyli233:
Fixes #1648.

As response to [this issue](https://github.com/bufbuild/buf/issues/1648), I added a new arg for `buf mod init`. 

`buf mod init` now can take an input 
E.g. `buf mod init buf.build/bufbuild/buf`


private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -158,3 +182,13 @@ func run(
...
+
+func getModuleName(name1, name2 string) (string, error) {
+	if name1 != "" && name2 != "" && name1 != name2 {
+		return "", fmt.Errorf("two different module name was supplied: %s and %s", name1, name2)
+	}
+	if name1 == "" {
+		return name2, nil
+	}
+	return name1, nil
+}
```
elliotmjackson, line 194: have another go at this, try conform to the style-guide and describe var names

private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -59,6 +59,8 @@ func NewCommand(
 type flags struct {
 	DocumentationComments bool
 	OutDirPath            string
+	// Special
+	InputHashtag string
```
elliotmjackson, line 63: is this an "inputHashTag"? whats the definition?

private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -115,7 +117,18 @@ func run(
...
+		return err
+	}
 	var writeConfigOptions []bufconfig.WriteConfigOption
+	if name != "" {
+		writeConfigWithName, err := bufconfig.WriteConfigWithName(name)
+		if err != nil {
+			return err
+		}
+		writeConfigOptions = append(writeConfigOptions, writeConfigWithName)
+	}
```
elliotmjackson, line 131:
```suggestion
	var writeConfigOptions []bufconfig.WriteConfigOption
	name := container.Arg(1)
	if name != "" {
		writeConfigWithName, err := bufconfig.WriteConfigWithName(name)
		if err != nil {
			return err
		}
		writeConfigOptions = append(writeConfigOptions, writeConfigWithName)
	}
```

private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -116,6 +116,17 @@ func run(
 		return appcmd.NewInvalidArgumentErrorf("%s already exists, not overwriting", existingConfigFilePath)
 	}
 	var writeConfigOptions []bufconfig.WriteConfigOption
+	name := ""
+	if container.NumArgs() > 0 {
+		name = container.Arg(0)
+	}
+	if name != "" {
```
elliotmjackson, line 123: well done

private/bufpkg/bufconfig/bufconfig.go
```diff
@@ -102,6 +102,16 @@ func WriteConfig(
 // WriteConfigOption is an option for WriteConfig.
 type WriteConfigOption func(*writeConfigOptions)
 
+func WriteConfigWithName(name string) (WriteConfigOption, error) {
```
marekbuild, line 105: Does this follow the style guide?

marekbuild:
@tonyli233 @elliotmjackson Do we use `--name` anywhere else?

elliotmjackson:
I'm unsure but we aren't using it here either... description is out of date, moving to draft

private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -43,9 +43,9 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name,
+		Use:   name + " <input>",
```
elliotmjackson, line 46: I'm not sure this counts as an "input" you will need to have a look at the precedent for `input` elsewhere in the codebase. This may be better off as `argument` or `arg`... I'm not sure, see what you can find

private/bufpkg/bufconfig/bufconfig.go
tonyli233: I think so. This conforms with the code surrounded. The parameter name is the module name passed in by user which has a key called ```name``` in the config file.

private/buf/cmd/buf/command/mod/modinit/modinit.go
tonyli233: I checked the precedence of ```input``` in the codebase. Looks like ```input``` refers to files in other commands. I didn't find ```argument``` or ```arg``` elsewhere. I think ```module-name``` will be a better way to describe.

elliotmjackson: [module-name](https://docs.buf.build/tour/push-a-module#module-names) looks appropriate.

private/bufpkg/bufconfig/bufconfig.go
tonyli233: I was trying to hide `moduleIdentity` in ```modinit.go``` since ´modinit.go´ is only setting the config with given parameters. 
I also agree that pulling the `bufmoduleref.ModuleIdentityForString(name)` call to `modinit.go` will make the functions uniform to other wrappers.

bufdev: All of our public types need documentation - looking at the style guide, this doesn't appear to be in there, so we should add

bufdev:
This needs some thinking as to what we should do here - this is a one-way door, and adding a new argument to a command is the ultimate one-way door.

elliotmjackson (changes requested):
blocking merge until @bufdev green light

private/bufpkg/bufconfig/bufconfig.go
tonyli233: The way of setting this config is changed. This function is no longer needed. 

I'll take a look at the style guide and keep in mind in future.

private/buf/cmd/buf/command/mod/modinit/modinit.go
```diff
@@ -43,9 +44,9 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name,
+		Use:   name + " <module-name>",
```
bufdev, line 47: @joshcarp do you like this?

bufdev (approved):
Approved pending @joshcarp's feedback

private/buf/cmd/buf/command/mod/modinit/modinit.go
joshcarp: TBH we shouldn't introduce a new placeholder `<module-name>`. 
We should either:
- Use `[name]` because it's consistent with the `name` field in `buf.yaml` 
- Use `[module]` because we have a somewhat consistent idea of what a module is, and the context here implies that you'd want to add a name
- Use `[buf.build/owner/repository]` if we want to be the most user friendly
- The problem with this last one is that there's a lot of complexity on the users to understand the difference between `repository` and `module`. IMO the slightly better option would probably be to avoid using `repository` altogether in this case (because we're technically not "creating" a repository on the BSR, yet)

I think the option that will cause the least confusion would be
```
  buf mod init [buf.build/owner/foobar]
```
This:
- gives the optional form (with square brackets)
- avoids the ambiguity in what a `name` is
- shows the form as what is should be `buf.build/owner/foobar`
- avoids the mental strain of understanding what a `repository` is

```suggestion
		Use:   name + " [buf.build/owner/foobar]",
```

elliotmjackson (approved):

bufdev (approved):

joshcarp (approved):
A small nit, but overall LGTM

PR #1737, opened 2023-01-16
Refactor BUF_TOKEN structure and interceptor constructor

tonyli233:
committed the changes suggested by @twilly in https://linear.app/bufbuild/issue/TCN-1008/add-support-for-buf-auth:
1. The `BUF_TOKEN` contains pairs of remote username and token and a default token that is used for every host. The `BUF_TOKEN` is interpreted as `${USER1}:${TOKEN1}@${REMOTE1},${USER2}:${TOKEN2}@${REMOTE2},${DEFAULT_TOKEN}`
2. The constructor for interceptor is refactored. The new constructor takes in a TokenSet to set the authorization header of the request.
3. Refactored test. Added coverage on zero value env and construct interceptor with provided token.

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -633,10 +633,11 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 	if err != nil {
 		return nil, err
 	}
+	machineFinder := bufconnect.NewMachineFinder(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, machineFinder),
 		),
 	)
 }
@@ -648,10 +649,11 @@ func NewConnectClientConfigWithToken(container appflag.Container, token string) 
 	if err != nil {
 		return nil, err
 	}
+	machineFinder := bufconnect.NewMachineFinder(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, machineFinder),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -629,10 +629,14 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
+	tokenSet, err := bufconnect.NewTokenSetFromContainer(container)
+	if err != nil {
+		return nil, err
+	}
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(bufconnect.AuthorizeWithAddress(container)),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
 		),
 	)
 }
@@ -640,10 +644,14 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
+	tokenSet, err := bufconnect.NewTokenSetFromString(token)
+	if err != nil {
+		return nil, err
+	}
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(bufconnect.AuthorizeWithProvidedToken(token)),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -646,14 +646,14 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
-	tokenSet, err := bufconnect.NewTokenProviderFromString(token)
+	tokenProvider, err := bufconnect.NewTokenProviderFromString(token)
 	if err != nil {
 		return nil, err
 	}
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenProvider),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -629,11 +629,10 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
-	newAuthorizationTokenOption := bufconnect.AuthorizeWithAddress(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(newAuthorizationTokenOption),
+			bufconnect.NewAuthorizationInterceptorProvider(bufconnect.AuthorizeWithAddress(container)),
 		),
 	)
 }
@@ -641,11 +640,10 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
-	newAuthorizationTokenOption := bufconnect.AuthorizeWithProvidedToken(token)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(newAuthorizationTokenOption),
+			bufconnect.NewAuthorizationInterceptorProvider(bufconnect.AuthorizeWithProvidedToken(token)),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -630,7 +630,7 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
-	tokenSet, err := bufconnect.NewTokenSetProviderFromContainer(container)
+	envTokenProvider, err := bufconnect.NewTokenProviderFromContainer(container)
 	if err != nil {
 		return nil, err
 	}
@@ -638,7 +638,7 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, netrcToken),
+			bufconnect.NewAuthorizationInterceptorProvider(envTokenProvider, netrcToken),
 		),
 	)
 }
@@ -646,7 +646,7 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
-	tokenSet, err := bufconnect.NewTokenSetProviderFromString(token)
+	tokenSet, err := bufconnect.NewTokenProviderFromString(token)
 	if err != nil {
 		return nil, err
 	}
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -633,11 +633,11 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 	if err != nil {
 		return nil, err
 	}
-	machineFinder := bufconnect.NewMachineFinder(container)
+	netrcToken := bufconnect.NewNetrcTokens(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, machineFinder),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, netrcToken),
 		),
 	)
 }
@@ -649,11 +649,10 @@ func NewConnectClientConfigWithToken(container appflag.Container, token string) 
 	if err != nil {
 		return nil, err
 	}
-	machineFinder := bufconnect.NewMachineFinder(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(tokenSet, machineFinder),
+			bufconnect.NewAuthorizationInterceptorProvider(tokenSet),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -634,11 +634,11 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 	if err != nil {
 		return nil, err
 	}
-	netrcToken := bufconnect.NewNetrcTokensProvider(container, netrc.GetMachineForName)
+	netrcTokenProvider := bufconnect.NewNetrcTokenProvider(container, netrc.GetMachineForName)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(envTokenProvider, netrcToken),
+			bufconnect.NewAuthorizationInterceptorProvider(envTokenProvider, netrcTokenProvider),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -50,6 +50,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/filelock"
 	"github.com/bufbuild/buf/private/pkg/git"
 	"github.com/bufbuild/buf/private/pkg/httpauth"
+	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/normalpath"
 	"github.com/bufbuild/buf/private/pkg/storage"
 	"github.com/bufbuild/buf/private/pkg/storage/storageos"
@@ -629,11 +630,11 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
-	tokenSet, err := bufconnect.NewTokenSetFromContainer(container)
+	tokenSet, err := bufconnect.NewTokenSetProviderFromContainer(container)
 	if err != nil {
 		return nil, err
 	}
-	netrcToken := bufconnect.NewNetrcTokens(container)
+	netrcToken := bufconnect.NewNetrcTokensProvider(container, netrc.GetMachineForName)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
@@ -645,7 +646,7 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
-	tokenSet, err := bufconnect.NewTokenSetFromString(token)
+	tokenSet, err := bufconnect.NewTokenSetProviderFromString(token)
 	if err != nil {
 		return nil, err
 	}
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -629,10 +629,11 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
+	newAuthorizationTokenOption := bufconnect.SetAuthTokenWithAddress(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProvider(container),
+			bufconnect.NewAuthorizationInterceptorProvider(newAuthorizationTokenOption),
 		),
 	)
 }
@@ -640,10 +641,11 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
+	newAuthorizationTokenOption := bufconnect.SetAuthTokenWithProvidedToken(token)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
-			bufconnect.NewAuthorizationInterceptorProviderWithToken(token),
+			bufconnect.NewAuthorizationInterceptorProvider(newAuthorizationTokenOption),
 		),
 	)
 }
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -629,7 +629,7 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
-	newAuthorizationTokenOption := bufconnect.SetAuthTokenWithAddress(container)
+	newAuthorizationTokenOption := bufconnect.AuthorizeWithAddress(container)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
@@ -641,7 +641,7 @@ func NewConnectClientConfig(container appflag.Container) (*connectclient.Config,
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
-	newAuthorizationTokenOption := bufconnect.SetAuthTokenWithProvidedToken(token)
+	newAuthorizationTokenOption := bufconnect.AuthorizeWithProvidedToken(token)
 	return newConnectClientConfigWithOptions(
 		container,
 		connectclient.WithAuthInterceptorProvider(
--- a/private/bufpkg/bufconnect/interceptors.go
+++ b/private/bufpkg/bufconnect/interceptors.go
@@ -43,78 +43,54 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
 	return interceptor
 }
 
-// MachineFinder finds the machine by name
-type MachineFinder interface {
-	getMachineForName(name string) (netrc.Machine, error)
+// TokenFinder finds the token for NewAuthorizationInterceptorProvider
+type TokenFinder interface {
+	// RemoteToken returns the remote token from the remote address.
+	// setFromEnvVar is true if the returned token is from the tokenEnvKey environment variable.
+	RemoteToken(address string) (token string, setFromEnvVar bool)
 }
 
-// NewMachineFinder returns a machineFinderImpl
-func NewMachineFinder(container app.EnvContainer) *machineFinderImpl {
-	return &machineFinderImpl{container: container}
+// NetrcTokens finds remote tokens from netrc machine.
+type NetrcTokens interface {
+	TokenFinder
+
+	GetMachineForName(address string) (netrc.Machine, error)
 }
 
-type machineFinderImpl struct {
+type netrcTokensImpl struct {
 	container app.EnvContainer
 }
 
-func (m *machineFinderImpl) getMachineForName(address string) (netrc.Machine, error) {
-	return netrc.GetMachineForName(m.container, address)
+// NewNetrcTokens returns a netrcTokensImpl
+func NewNetrcTokens(container app.EnvContainer) *netrcTokensImpl {
+	return &netrcTokensImpl{container: container}
 }
 
-// NewAuthorizationInterceptorProvider returns a new provider function which, when invoked, returns an interceptor
-// which will set the auth token into the request header by the provided option.
-//
-// Note that the interceptor returned from this provider is always applied LAST in the series of interceptors added to
-// a client.
-func NewAuthorizationInterceptorProvider(tokenSet *TokenSet, machineFinder MachineFinder) func(string) connect.UnaryInterceptorFunc {
-	return func(address string) connect.UnaryInterceptorFunc {
-		interceptor := func(next connect.UnaryFunc) connect.UnaryFunc {
-			return connect.UnaryFunc(func(
-				ctx context.Context,
-				req connect.AnyRequest,
-			) (connect.AnyResponse, error) {
-				authToken := ""
-				if tokenSet != nil {
-					token := tokenSet.lookUpRemoteToken(address)
-					if token != "" {
-						authToken = token
-					}
-				}
-				if authToken == "" && machineFinder != nil {
-					machine, err := machineFinder.getMachineForName(address)
-					if err != nil {
-						return nil, fmt.Errorf("failed to read server password from netrc: %w", err)
-					}
-					if machine != nil {
-						authToken = machine.Password()
-					}
-				}
-				if authToken != "" {
-					req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+authToken)
-				}
-				response, err := next(ctx, req)
-				if err != nil && tokenSet != nil && tokenSet.setBufToken {
-					err = &ErrAuth{cause: err, tokenEnvKey: tokenEnvKey}
-				}
-				return response, err
-			})
-		}
-		return interceptor
+func (nt *netrcTokensImpl) RemoteToken(address string) (string, bool) {
+	machine, err := nt.GetMachineForName(address)
+	if err != nil {
+		return "", false
 	}
+	if machine != nil {
+		return machine.Password(), false
+	}
+	return "", false
+}
+
+func (nt *netrcTokensImpl) GetMachineForName(address string) (netrc.Machine, error) {
+	return netrc.GetMachineForName(nt.container, address)
 }
 
 // TokenSet is used to provide authentication token in NewAuthorizationInterceptorProvider
 type TokenSet struct {
-	// setBufToken is true when the tokenEnvKey is defined in the environment, false otherwise.
-	setBufToken  bool
-	defaultToken string
-	tokens       map[string]authKey
+	// true: the tokenSet is generated from environment variable tokenEnvKey
+	// false: otherwise
+	setBufTokenEnvVar bool
+	defaultToken      string
+	tokens            map[string]authKey
 }
 
-type authKey struct {
-	username string
-	token    string
-}
+var _ TokenFinder = (*TokenSet)(nil)
 
 // NewTokenSetFromContainer creates a TokenSet from the BUF_TOKEN environment variable
 func NewTokenSetFromContainer(container app.EnvContainer) (*TokenSet, error) {
@@ -124,7 +100,7 @@ func NewTokenSetFromContainer(container app.EnvContainer) (*TokenSet, error) {
 		return nil, err
 	}
 	if bufToken != "" {
-		tokenSet.setBufToken = true
+		tokenSet.setBufTokenEnvVar = true
 	}
 	return tokenSet, nil
 }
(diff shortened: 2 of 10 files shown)


private/buf/bufcli/bufcli.go
```diff
@@ -629,21 +629,23 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
...
-			bufconnect.NewAuthorizationInterceptorProvider(container),
+			bufconnect.NewAuthorizationInterceptorProvider(newAuthorizationTokenOption),
 		),
 	)
 }
 
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
+	newAuthorizationTokenOption := bufconnect.SetAuthTokenWithProvidedToken(token)
```
bufdev, line 644: No need for variable here, this can be inlined

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,120 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
-			return connect.UnaryFunc(func(
-				ctx context.Context,
-				req connect.AnyRequest,
-			) (connect.AnyResponse, error) {
-				if token != "" {
-					req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+token)
-				}
-				return next(ctx, req)
-			})
+// SetAuthTokenOption is an option for NewAuthorizationInterceptorProvider
```
bufdev, line 65: Godoc should always be complete sentences, sentences should end in periods.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,120 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
 // NewAuthorizationInterceptorProvider returns a new provider function which, when invoked, returns an interceptor
-// which will look up an auth token by address and set it into the request header.  This is used for registry providers
-// where the token is looked up by the client address at the time of client construction (i.e. for clients where a
-// user is already authenticated and the token is stored in .netrc)
+// which will set the auth token into the request header by the provided option.
 //
 // Note that the interceptor returned from this provider is always applied LAST in the series of interceptors added to
 // a client.
-func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string) connect.UnaryInterceptorFunc {
+func NewAuthorizationInterceptorProvider(option SetAuthTokenOption) func(string) connect.UnaryInterceptorFunc {
```
bufdev, line 51: Options are generally things that are optional in constructors, whereas this appears to be a required parameter.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,120 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
-				ctx context.Context,
-				req connect.AnyRequest,
-			) (connect.AnyResponse, error) {
-				if token != "" {
-					req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+token)
-				}
-				return next(ctx, req)
-			})
+// SetAuthTokenOption is an option for NewAuthorizationInterceptorProvider
+type SetAuthTokenOption func(connect.AnyRequest, context.Context, connect.UnaryFunc, string) (connect.AnyResponse, error)
```
bufdev, line 66: I don't quite follow what the purpose of this is - why would we not just have two separate Interceptor providers?

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,120 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	remoteUsernames map[string]string
+	remoteTokens    map[string]string
+}
+
+func newTokenSetFromString(token string) (*tokenSet, error) {
+	tokenSet := &tokenSet{
+		remoteUsernames: make(map[string]string),
+		remoteTokens:    make(map[string]string),
+	}
+	tokens := strings.Split(token, ",")
```
bufdev, line 126: What if usernames, passwords, or remotes include `,`, `@`, or `:`? Are we sufficiently sure that this is impossible in each case?

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,120 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+				return nil, fmt.Errorf("cannot parse token: %s, repeated token for same BSR remote: %s", remoteToken, remoteAddress)
+			}
+			if _, ok := tokenSet.remoteUsernames[remoteAddress]; ok {
+				return nil, fmt.Errorf("cannot parse token: %s, repeated token for same BSR remote: %s", remoteToken, remoteAddress)
+			}
+			tokenSet.remoteTokens[remoteAddress] = remoteToken
+			tokenSet.remoteUsernames[remoteAddress] = username
+		} else {
+			if tokenSet.bufToken != "" {
+				return nil, fmt.Errorf("cannot parse token: %s, two buf token provided: %s and %s", token, u, tokenSet.bufToken)
```
bufdev, line 150: `two buf tokens`

use `%q` to quote strings for debugging purposes

bufdev (commented):
Why are we required to have the username in the case of multiple tokens within a `BUF_TOKEN`, but not required to have it in the case of a singular token?

private/bufpkg/bufconnect/interceptors.go
tonyli233: This option is required as we must either set the auth token by provided token or container.

bufdev: See comment below - I am not quite sure why this is preferred over having two constructors

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,123 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+		if token != "" {
+			req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+token)
 		}
-		return interceptor
+		return next(ctx, req)
+	}
+}
+
+// stubbed for testing
+var getMachineForName = netrc.GetMachineForName
```
twilly, line 81: Use dependency injection to capture this.

private/bufpkg/bufconnect/interceptors.go
tonyli233: We can't answer this yet. We need to make sure this is impossible and add it to the specs.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -43,64 +44,123 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
-				}
-				return next(ctx, req)
-			})
+// AuthorizeOption is an option for NewAuthorizationInterceptorProvider.
+type AuthorizeOption func(connect.AnyRequest, context.Context, connect.UnaryFunc, string) (connect.AnyResponse, error)
+
+// AuthorizeWithProvidedToken returns a new SetAuthTokenOption that will set the
+// provided token into the request header This is used for registry providers where the token is known at provider
+// creation (i.e. when logging in and explicitly pasting a token into stdin
+func AuthorizeWithProvidedToken(token string) AuthorizeOption {
```
marekbuild, line 71: Doc string should have a consistent column width where newlines start, for example the newlines should start at the same column as the end of `... that will set the`

Also the doc string should be consistent about closing `)` and terminating with a period.

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -21,18 +21,70 @@ import (
...
+	container = app.NewEnvContainer(map[string]string{})
+	_, err = NewAuthorizationInterceptorProvider(AuthorizeWithAddress(container))("fake")(func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		if req.Header().Get(AuthenticationHeader) != AuthenticationTokenPrefix+"password" {
+			return nil, errors.New("error auth token found")
+		}
+		return nil, nil
+	})(context.Background(), connect.NewRequest(&bytes.Buffer{}))
+	assert.NoError(t, err)
+	getMachineForName = netrc.GetMachineForName
+}
```
marekbuild, line 71: @tonyli233 , could we split this up into multiple test functions. It looks like it is executing multiple tests. The previous version of the test was very compact and easy to grok, not it is quite large.

private/bufpkg/bufconnect/interceptors.go
bufdev: Blocking the merging of this PR until this is resolved - I think the previous style, with two constructors, is correct, and there should be no new type named *Option introduced to this PR with one constructor deleted

private/bufpkg/bufconnect/interceptors_test.go
tonyli233: the tests are changed according to the new interceptor constructor

private/bufpkg/bufconnect/interceptors.go
tonyli233: The constructor is taking tokenset to provide the auth token in the most recent commit

tonyli233:
> Why are we required to have the username in the case of multiple tokens within a `BUF_TOKEN`, but not required to have it in the case of a singular token?

The old format of `BUF_TOKEN` is still supported.

private/bufpkg/bufconnect/interceptors.go
tonyli233: implementation is updated

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -25,14 +25,32 @@ import (
...
+	_, err := NewTokenSetFromString("default1,default2")
+	assert.Error(t, err)
+	_, err = NewTokenSetFromString("invalid@invalid@invalid")
+	assert.Error(t, err)
+	_, err = NewTokenSetFromString("invalid:invalid:invalid@remote")
+	assert.Error(t, err)
+	_, err = NewTokenSetFromString("user1:token1@remote,user2:token2@remote")
+	assert.Error(t, err)
+	_, err = NewTokenSetFromString("")
+	assert.NoError(t, err)
```
twilly, line 55: Yay, tests! Here's some things to cover:

 - Add a test for a valid single token.
 - When you have two default tokens, test for the expected error.
 - Same goes for when you have duplicates for the same remote.
 - Check for two different tokens and users for two different remotes.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -82,25 +71,66 @@ func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string
...
+				return nil, fmt.Errorf("cannot parse token: %s, two buf token provided: %q and %q", token, u, tokenSet.bufToken)
+			}
+			tokenSet.bufToken = u
 		}
-		return interceptor
 	}
+	return tokenSet, nil
+}
+
+func (t *TokenSet) obtainTokenFromRemoteAddress(remoteAddress string) (token string) {
```
twilly, line 131: `tokenForRemoteAddress(remoteAddress string) string {`

"obtain" isn't different from saying "get"

"from" implies the value is derived from remoteAddress when it's really a lookup.

You didn't use the named return value `token`. No need to have it either.

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -25,14 +25,32 @@ import (
...
+			return nil, errors.New("error auth token")
+		}
+		return nil, nil
 	})(context.Background(), connect.NewRequest(&bytes.Buffer{}))
+	assert.NoError(t, err)
+}
+
+func TestNewTokenSetFromEnv(t *testing.T) {
+	_, err := NewTokenSetFromContainer(app.NewEnvContainer(map[string]string{}))
+	assert.NoError(t, err)
```
twilly, line 42: Check for values here too, not just error.

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -25,14 +25,32 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestAuthorizationInterceptorProviderTokenErr(t *testing.T) {
-	_, err := NewAuthorizationInterceptorProvider(app.NewEnvContainer(map[string]string{
-		tokenEnvKey: "test1234",
-	}))("fake")(func(context.Context, connect.AnyRequest) (connect.AnyResponse, error) {
-		return nil, errors.New("underlying cause")
+func TestNewAuthorizationInterceptorProvider(t *testing.T) {
```
twilly, line 28: Check for no header when there is no token for the remote.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -82,25 +71,66 @@ func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string
...
+			if len(keyPairsAndRemoteAddress) != 2 {
+				return nil, fmt.Errorf("cannot parse token: %s, invalid remote token: %s", token, u)
+			}
+			keyPairs := strings.Split(keyPairsAndRemoteAddress[0], ":")
+			if len(keyPairs) != 2 {
+				return nil, fmt.Errorf("cannot parse token: %s, invalid remote token: %s", token, u)
+			}
+			remoteAddress := keyPairsAndRemoteAddress[1]
+			username := keyPairs[0]
+			remoteToken := keyPairs[1]
```
twilly, line 112: Try using strings.Cut, like in my comment above.

Do read rsc's "The Unreasonable Effectiveness of Cut" `https://github.com/golang/go/issues/46336` (escaped to avoid pinging that issue). It's a goodie.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -82,25 +71,66 @@ func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string
...
-					req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+token)
-				}
-				return next(ctx, req)
-			})
+// TokenSet is used to provide authentication token in NewAuthorizationInterceptorProvider
+type TokenSet struct {
+	envKey          string
+	bufToken        string
+	remoteUsernames map[string]string
+	remoteTokens    map[string]string
```
twilly, line 79: Instead of two maps, make it one map with a struct.

```go
type authkey struct {
  Username string
  Token string
}

type TokenSet struct {
  // ...
  tokens map[string]authkey
}
```

The neat part is you can break apart your parsing code since it's a dedicated type. If you were to follow the encoding package's form, it'd look like:

```go
func (ak *authkey) UnmarshalString(s string) error {
  ak.Username, ak.Token, found := strings.Cut(s, ":")
  if !found {
    return errors.New("invalid user:token format")
  }
  return nil
}
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -82,25 +71,66 @@ func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string
...
-			) (connect.AnyResponse, error) {
-				if token != "" {
-					req.Header().Set(AuthenticationHeader, AuthenticationTokenPrefix+token)
-				}
-				return next(ctx, req)
-			})
+// TokenSet is used to provide authentication token in NewAuthorizationInterceptorProvider
+type TokenSet struct {
+	envKey          string
+	bufToken        string
```
twilly, line 77: nit: `defaultToken`

"buf" doesn't really describe that it's a value that's used if there is no remote. And a remote isn't necessarily "buf" (single tenants).

private/bufpkg/bufconnect/interceptors.go
tonyli233: The function name is changed to `lookupRemoteToken`

private/bufpkg/bufconnect/interceptors_test.go
tonyli233: Added a test for checking valid single token and checking different remote. For `two default token` and `duplicate remote`, do you mean we should also check if it is giving the correct error message?

private/bufpkg/bufconnect/interceptors_test.go
tonyli233: Now also checking the token value by looking up from remotes.

private/bufpkg/bufconnect/interceptors_test.go
tonyli233: Now it checks the header for `AuthenticationHeader` is empty.

private/bufpkg/bufconnect/interceptors.go
tonyli233: This code can look way more elegant if we are using `strings.Cut`. But detecting invalid tokens like `invalid:invalid:invalid@remote` or invalid:invalid@invalid@remote is harder. But I think we don't have to valid date the token here as we are only trying to parse the tokens. The tokens are validated in a later step. Should I trying using `strings.Cut` and pass in whatever is parsed?

private/bufpkg/bufconnect/interceptors.go
tonyli233: now using dependency injection

private/bufpkg/bufconnect/interceptors_test.go
twilly: Checking for the expected error is what I was thinking. You don't have to but it can help if you want confidence that you're exercising the right check for the right kind of validity.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,38 +43,58 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
 	return interceptor
 }
 
+// MachineFinder finds the machine by name
+type MachineFinder interface {
+	getMachineForName(name string) (netrc.Machine, error)
+}
```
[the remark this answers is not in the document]
tonyli233: `NewAuthorizationInterceptorProvider` is refactored to the suggested implementation, except the definition of `RemoteToken(address string) (token string, setFromEnvVar bool)` of `TokenFinder`. The `setFromEnvVar` is true when the remote token returned is obtained from the `tokenEnvKey` environment variable. This is needed since we need to return different error message [here](https://github.com/bufbuild/buf/blob/13e8ff16083aad6ef2f4e3543d0669857279ebc6/private/buf/bufcli/errors.go#L185) if the token is obtained from the environment variable.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,38 +43,58 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
 				ctx context.Context,
 				req connect.AnyRequest,
 			) (connect.AnyResponse, error) {
-				envKey := tokenEnvKey
-				token := container.Env(envKey)
-				if token == "" {
-					envKey = ""
-					machine, err := netrc.GetMachineForName(container, address)
+				authToken := ""
+				if tokenSet != nil {
```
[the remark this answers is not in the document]
tonyli233: The implementation of `NewAuthorizationInterceptorProvider` is refactored to take multiple `TokenFinder`.

private/bufpkg/bufconnect/interceptors.go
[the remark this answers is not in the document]
tonyli233: This makes the `NetrcTokens` a little bit hard to test. So I made it as an interface and added `netrcTokensImpl` implementation for testing.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,144 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	// setFromEnvVar is true if the returned token is from the tokenEnvKey environment variable.
+	RemoteToken(address string) (token string, setFromEnvVar bool)
+}
+
+// NetrcTokens finds remote tokens from netrc machine.
+type NetrcTokens interface {
+	TokenFinder
+
+	GetMachineForName(address string) (netrc.Machine, error)
+}
```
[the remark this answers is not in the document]
tonyli233: The `NetrcTokensProvider` is refactored by the suggested way.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
 	return interceptor
 }
 
+// TokenFinder finds the token for NewAuthorizationInterceptorProvider
+type TokenFinder interface {
+	// RemoteToken returns the remote token from the remote address.
+	// setFromEnvVar is true if the returned token is from the tokenEnvKey environment variable.
```
twilly, line 49: Small doc nit: that's a private name and a detail. (and ugly detail, but one anyway)

```suggestion
	// RemoteToken returns the remote token from the remote address.
	// setFromEnvVar is true if the returned token is from an environment variable.
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	RemoteToken(address string) (token string, setFromEnvVar bool)
+}
+
+type NetrcTokensProvider struct {
+	container         app.EnvContainer
+	getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)
+}
+
+// NewNetrcTokensProvider returns a NetrcTokensProvider
```
twilly, line 58:
```suggestion
// NewNetrcTokensProvider returns a TokensProvider for a .netrc in a container.
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+func NewNetrcTokensProvider(container app.EnvContainer, getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)) *NetrcTokensProvider {
+	return &NetrcTokensProvider{container: container, getMachineForName: getMachineForName}
+}
+
+// RemoteToken finds the token from the .netrc file
```
twilly, line 63: Since this method is implementing an interface, the method behavior is documented there. The constructor can document the implementation's specifics.

```suggestion
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,144 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+			if tokenSet.defaultToken != "" {
+				return nil, fmt.Errorf("cannot parse token: %s, two buf token provided: %q and %q", token, u, tokenSet.defaultToken)
+			}
+			tokenSet.defaultToken = u
+		}
+	}
+	return tokenSet, nil
+}
+
+func (t *TokenSet) RemoteToken(address string) (string, bool) {
```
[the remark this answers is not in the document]
tonyli233: Should we include this in the scope of any near future work?

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+// TokenSetProvider is used to provide authentication token in NewAuthorizationInterceptorProvider
+type TokenSetProvider struct {
+	// true: the tokenSet is generated from environment variable tokenEnvKey
+	// false: otherwise
+	setBufTokenEnvVar bool
+	defaultToken      string
+	tokens            map[string]authKey
+}
+
+var _ TokenFinder = (*TokenSetProvider)(nil)
```
twilly, line 84: nit: You can do the same type check for NetrcTokensProvider.

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	if err != nil {
+		return "", false
+	}
+	if machine != nil {
+		return machine.Password(), false
+	}
+	return "", false
+}
+
+// TokenSetProvider is used to provide authentication token in NewAuthorizationInterceptorProvider
```
twilly, line 75: Since this isn't the only way to inject tokens for the auth interceptor:

```suggestion
// TokenSetProvider is used to provide set of authentication tokens.
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
 }
 
+// TokenFinder finds the token for NewAuthorizationInterceptorProvider
+type TokenFinder interface {
+	// RemoteToken returns the remote token from the remote address.
+	// setFromEnvVar is true if the returned token is from the tokenEnvKey environment variable.
+	RemoteToken(address string) (token string, setFromEnvVar bool)
+}
+
+type NetrcTokensProvider struct {
```
bufdev, line 53: this type doesnt need to be public, you can just make the constructor public and move this to a `netrc_tokens_provider.go` file

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
 	return interceptor
 }
 
+// TokenFinder finds the token for NewAuthorizationInterceptorProvider
+type TokenFinder interface {
+	// RemoteToken returns the remote token from the remote address.
+	// setFromEnvVar is true if the returned token is from the tokenEnvKey environment variable.
+	RemoteToken(address string) (token string, setFromEnvVar bool)
```
bufdev, line 50: Since the `setFromEnvVar` variable is static, you could split this out into another function to make the main call cleaner. This should also likely be named `TokenProvider` to be consistent:

```go
type TokenProvider interface {
  RemoteToken(address string) string
  IsFromEnvVar() bool
}
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+		return "", false
+	}
+	if machine != nil {
+		return machine.Password(), false
+	}
+	return "", false
+}
+
+// TokenSetProvider is used to provide authentication token in NewAuthorizationInterceptorProvider
+type TokenSetProvider struct {
```
bufdev, line 76: Is there any reason this needs to be public instead of using the `TokenFinder` interface? If not, make private, move to `token_set_provider.go`

private/bufpkg/bufconnect/interceptors.go
bufdev: Add a comment to the code to this effect for future readers

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -21,18 +21,109 @@ import (
...
 	"github.com/bufbuild/connect-go"
 	"github.com/stretchr/testify/assert"
 )
 
-func TestAuthorizationInterceptorProviderTokenErr(t *testing.T) {
-	_, err := NewAuthorizationInterceptorProvider(app.NewEnvContainer(map[string]string{
-		tokenEnvKey: "test1234",
-	}))("fake")(func(context.Context, connect.AnyRequest) (connect.AnyResponse, error) {
-		return nil, errors.New("underlying cause")
+type testMachineImpl struct{}
```
bufdev, line 29: no need to suffix this with `Impl`

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+		return nil, err
+	}
+	if bufToken != "" {
+		tokenSet.setBufTokenEnvVar = true
+	}
+	return tokenSet, nil
+}
+
+// NewTokenSetProviderFromString creates a TokenSetProvider by the token provided
+func NewTokenSetProviderFromString(token string) (*TokenSetProvider, error) {
```
bufdev, line 100: Also rename to `NewTokenProviderFromString` once `TokenFinder` is renamed to `TokenProvider` and this returns a `TokenProvider`

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	// false: otherwise
+	setBufTokenEnvVar bool
+	defaultToken      string
+	tokens            map[string]authKey
+}
+
+var _ TokenFinder = (*TokenSetProvider)(nil)
+
+// NewTokenSetProviderFromContainer creates a TokenSetProvider from the BUF_TOKEN environment variable
+func NewTokenSetProviderFromContainer(container app.EnvContainer) (*TokenSetProvider, error) {
```
bufdev, line 87: Also rename to `NewTokenProviderFromString` once `TokenFinder` is renamed to `TokenProvider` and this returns a `TokenProvider`

twilly (approved):
This should work and coverage is much improved. I left some small things, but +1 for getting the functionality and looking correct.

private/buf/bufcli/bufcli.go
```diff
@@ -629,21 +630,30 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
+	tokenSet, err := bufconnect.NewTokenSetProviderFromContainer(container)
```
bufdev, line 633: this should be renamed `envTokenProvider` once  `TokenFinder` is renamed to `TokenProvider` and `NewTokenProviderFromContainer` returns a new `TokenProvider`

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	RemoteToken(address string) (token string, setFromEnvVar bool)
+}
+
+type NetrcTokensProvider struct {
+	container         app.EnvContainer
+	getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)
+}
+
+// NewNetrcTokensProvider returns a NetrcTokensProvider
+func NewNetrcTokensProvider(container app.EnvContainer, getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)) *NetrcTokensProvider {
```
bufdev, line 59: `NewNetrcTokenProvider` and should return a `TokenProvider` once `TokenFinder` is renamed

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+
+// NewTokenSetProviderFromContainer creates a TokenSetProvider from the BUF_TOKEN environment variable
+func NewTokenSetProviderFromContainer(container app.EnvContainer) (*TokenSetProvider, error) {
+	bufToken := container.Env(tokenEnvKey)
+	tokenSet, err := NewTokenSetProviderFromString(bufToken)
+	if err != nil {
+		return nil, err
+	}
+	if bufToken != "" {
+		tokenSet.setBufTokenEnvVar = true
```
bufdev, line 94: If `bufToken == ""`, won't `RemoteToken` always return `""`? Whether or not a token came from an envVar feels like a property of a private constructor:



```go
func NewTokenProviderFromContainer(container app.EnvContainer) (TokenProvider, error) {
   return newTokenProviderFromString(container.Env(tokenEnvKey), true)
}

func NewTokenProviderFromString(token string) (TokenProvider, error) {
  return newTokenProviderFromString(token, false)
}

func newTokenProviderFromString(token string, isFromEnvVar bool) (TokenProvider, error) {
  ...
}
```

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+// NewTokenSetProviderFromString creates a TokenSetProvider by the token provided
+func NewTokenSetProviderFromString(token string) (*TokenSetProvider, error) {
+	tokenSet := &TokenSetProvider{
+		tokens: make(map[string]authKey),
+	}
+	tokens := strings.Split(token, ",")
+	for _, u := range tokens {
+		if keyPairs, remoteAddress, ok := strings.Cut(u, "@"); ok {
+			ak := authKey{}
+			err := ak.unmarshalString(keyPairs)
```
bufdev, line 108: error should be scoped to the if statement

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	return tokenSet, nil
+}
+
+// NewTokenSetProviderFromString creates a TokenSetProvider by the token provided
+func NewTokenSetProviderFromString(token string) (*TokenSetProvider, error) {
+	tokenSet := &TokenSetProvider{
+		tokens: make(map[string]authKey),
+	}
+	tokens := strings.Split(token, ",")
+	for _, u := range tokens {
```
bufdev, line 105: `for _, token := range tokens`

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +43,136 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
+	defaultToken      string
+	tokens            map[string]authKey
+}
+
+var _ TokenFinder = (*TokenSetProvider)(nil)
+
+// NewTokenSetProviderFromContainer creates a TokenSetProvider from the BUF_TOKEN environment variable
+func NewTokenSetProviderFromContainer(container app.EnvContainer) (*TokenSetProvider, error) {
+	bufToken := container.Env(tokenEnvKey)
+	tokenSet, err := NewTokenSetProviderFromString(bufToken)
```
bufdev, line 89: `tokenProvider, err`

bufdev (commented):
mostly style comments to get this consistent with the code base - good work

private/bufpkg/bufconnect/interceptors.go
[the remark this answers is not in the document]
twilly: I think so, but it's quite minor. I wouldn't spend much time without something bigger to tackle along with it.

tonyli233:
All the nits and style issues are corrected.

private/buf/bufcli/bufcli.go
```diff
@@ -629,21 +630,30 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
 // up the token in the container or in netrc based on the address of each individual client.
 // It is then set in the header of all outgoing requests from clients created using this config.
 func NewConnectClientConfig(container appflag.Container) (*connectclient.Config, error) {
+	envTokenProvider, err := bufconnect.NewTokenProviderFromContainer(container)
+	if err != nil {
+		return nil, err
+	}
+	netrcToken := bufconnect.NewNetrcTokensProvider(container, netrc.GetMachineForName)
```
bufdev, line 637: `netrcTokenProvider`

private/bufpkg/bufconnect/netrc_tokens_provider.go
```diff
@@ -0,0 +1,48 @@
...
+// netrcTokensProvider is used to provide remote tokens from .netrc.
+type netrcTokensProvider struct {
+	container         app.EnvContainer
+	getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)
+}
+
+var _ TokenProvider = (*netrcTokensProvider)(nil)
+
+// NewNetrcTokensProvider returns a TokenProvider for a .netrc in a container.
+func NewNetrcTokensProvider(container app.EnvContainer, getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)) TokenProvider {
```
bufdev, line 31: `NewNetrcTokenProvider`

private/bufpkg/bufconnect/netrc_tokens_provider.go
```diff
@@ -0,0 +1,48 @@
...
+
+package bufconnect
+
+import (
+	"github.com/bufbuild/buf/private/pkg/app"
+	"github.com/bufbuild/buf/private/pkg/netrc"
+)
+
+// netrcTokensProvider is used to provide remote tokens from .netrc.
+type netrcTokensProvider struct {
```
bufdev, line 23: `netrcTokenProvider`

private/bufpkg/bufconnect/netrc_tokens_provider.go
```diff
@@ -0,0 +1,48 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: `netrc_token_provider.go`

private/bufpkg/bufconnect/netrc_tokens_provider.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/bufbuild/buf/private/pkg/netrc"
+)
+
+// netrcTokensProvider is used to provide remote tokens from .netrc.
+type netrcTokensProvider struct {
+	container         app.EnvContainer
+	getMachineForName func(app.EnvContainer, string) (netrc.Machine, error)
+}
+
+var _ TokenProvider = (*netrcTokensProvider)(nil)
```
bufdev, line 28: nit: no need for this anymore as this is enforced by `NewNetrcTokenProvider`

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/bufbuild/buf/private/pkg/app"
+)
+
+// tokenSetProvider is used to provide set of authentication tokens.
+type tokenSetProvider struct {
```
bufdev, line 25: `staticTokenProvider`

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,97 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: `static_token_provider.go`

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,97 @@
...
+// tokenSetProvider is used to provide set of authentication tokens.
+type tokenSetProvider struct {
+	// true: the tokenSet is generated from environment variable tokenEnvKey
+	// false: otherwise
+	setBufTokenEnvVar bool
+	defaultToken      string
+	tokens            map[string]authKey
+}
+
+var _ TokenProvider = (*tokenSetProvider)(nil)
```
bufdev, line 33: nit: no need for this anymore since this is enforced by the constructors

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/bufbuild/buf/private/pkg/app"
+)
+
+// tokenSetProvider is used to provide set of authentication tokens.
+type tokenSetProvider struct {
+	// true: the tokenSet is generated from environment variable tokenEnvKey
+	// false: otherwise
+	setBufTokenEnvVar bool
+	defaultToken      string
+	tokens            map[string]authKey
```
bufdev, line 30: `tokenToAuthKey`

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,97 @@
...
+	return newTokenProviderFromString(container.Env(tokenEnvKey), true)
+}
+
+// NewTokenProviderFromString creates a tokenSetProvider by the token provided
+func NewTokenProviderFromString(token string) (TokenProvider, error) {
+	return newTokenProviderFromString(token, false)
+}
+
+func newTokenProviderFromString(token string, isFromEnvVar bool) (TokenProvider, error) {
+	tokenSet := &tokenSetProvider{
```
bufdev, line 46: `tokenProvider := &staticTokenProvider{`

bufdev (commented):
Looking great! Just a few more fine-tooth comb things and we should be good to go!

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -42,65 +39,41 @@ func NewSetCLIVersionInterceptor(version string) connect.UnaryInterceptorFunc {
...
 // NewAuthorizationInterceptorProvider returns a new provider function which, when invoked, returns an interceptor
-// which will look up an auth token by address and set it into the request header.  This is used for registry providers
-// where the token is looked up by the client address at the time of client construction (i.e. for clients where a
-// user is already authenticated and the token is stored in .netrc)
+// which will set the auth token into the request header by the provided option.
 //
 // Note that the interceptor returned from this provider is always applied LAST in the series of interceptors added to
 // a client.
-func NewAuthorizationInterceptorProvider(container app.EnvContainer) func(string) connect.UnaryInterceptorFunc {
+func NewAuthorizationInterceptorProvider(tokenFinders ...TokenProvider) func(string) connect.UnaryInterceptorFunc {
```
bufdev, line 55: `tokenProviders`

private/buf/bufcli/bufcli.go
```diff
@@ -629,21 +630,30 @@ func newConnectClientConfigWithOptions(container appflag.Container, opts ...conn
...
-			bufconnect.NewAuthorizationInterceptorProvider(container),
+			bufconnect.NewAuthorizationInterceptorProvider(envTokenProvider, netrcTokenProvider),
 		),
 	)
 }
 
 // NewConnectClientConfigWithToken creates a new connect.ClientConfig with a given token. The provided token is
 // set in the header of all outgoing requests from this provider
 func NewConnectClientConfigWithToken(container appflag.Container, token string) (*connectclient.Config, error) {
+	tokenSet, err := bufconnect.NewTokenProviderFromString(token)
```
bufdev, line 649: `tokenProvider`

private/bufpkg/bufconnect/netrc_token_provider.go
```diff
@@ -0,0 +1,46 @@
...
+
+package bufconnect
+
+import (
+	"github.com/bufbuild/buf/private/pkg/app"
+	"github.com/bufbuild/buf/private/pkg/netrc"
+)
+
+// netrcTokensProvider is used to provide remote tokenToAuthKey from .netrc.
+type netrcTokensProvider struct {
```
bufdev, line 23: still needs to be changed to `netrcTokenProvider`

private/bufpkg/bufconnect/token_set_provider.go
```diff
@@ -0,0 +1,95 @@
...
+	return newTokenProviderFromString(container.Env(tokenEnvKey), true)
+}
+
+// NewTokenProviderFromString creates a staticTokenProvider by the token provided
+func NewTokenProviderFromString(token string) (TokenProvider, error) {
+	return newTokenProviderFromString(token, false)
+}
+
+func newTokenProviderFromString(token string, isFromEnvVar bool) (TokenProvider, error) {
+	tokenSet := &staticTokenProvider{
```
bufdev, line 44: still need to be changed to `tokenProvider`

bufdev (commented):
a couple I missed (sorry about that, I really dislike when its multiple reviews for things that could have been caught) and a couple that are from the previous review that seem to have been resolved

tonyli233:
Fixed the nits and the ones that I missed before

bufdev:
There's still unresolved comments, address all the issues and then I'll take another look

bufdev (approved):
Thanks for getting through this!

bufdev:
Make sure to follow up with any necessary docs changes on docs.buf.build

tonyli233:
> Make sure to follow up with any necessary docs changes on docs.buf.build

I'll put it on the next scope of work.

PR #1738, opened 2023-01-16
Add PR title linter

bufdev:
We have had inconsistent PR titles across our various repositories - some start with capital letters, some don't, some end in periods, some don't. No one way is the right way, but from day one I did PR titles with capital letters and no periods at the end, and one way is better than no way, so @akshayjshah had the idea that we should just enforce this as part of linting with this action. Trying this on bufbuild/buf, and it it works, we'll add to our other repositories as well.

### DIFF
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -21,7 +21,7 @@ jobs:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
           title-regex: "^[A-Z].*[^.]$"
           on-failed-regex-fail-action: true
-          on-failed-regex-create-review: true
-          on-failed-regex-request-changes: true
+          on-failed-regex-create-review: false
+          on-failed-regex-request-changes: false
           on-failed-regex-comment: "PRs must start with a capital letter and not end with a period."
           on-succeeded-regex-dismiss-review-comment: "Thanks for helping keep our PR titles consistent!"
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-title.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - edited
+      - reopened
+      - synchronize
+jobs:
+  pr-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: morrisoncole/pr-lint-action@v1.7.0
+        with:
+          title-regex: "^[A-Z].*[^.]$"
```
akshayjshah: Should we also forbid other terminal punctuation, like question marks or exclamation points?

akshayjshah (approved):
IIRC the OSS discussion suggested that all workflows be in a single ci.yaml file - should this be an exception, like the review bypass workflows?

One other optional suggestion, no blocking concerns.

.github/workflows/pr-title.yaml
bufdev: Yea sgtm, I'm not crazy good in regex land though, so once we get this in feel free to adjust (or just push here)

bufdev:
I'd say we make this an exception so we can easily copy/paste it around. It also runs on different triggers than the ci.yaml jobs.

.github/workflows/pr-title.yaml
akshayjshah: Sure, can do. Might be nice to wait on copy-pasting this to every repo until we do this - otherwise it's twice as many PRs to ship and review. 

This should be a pretty trivial fix - add \?\! to the exclude set.

bufdev: Apparently @rubensf has some magic script that can do these types of rollouts widely, so @rubensf - coordinate with @akshayjshah once we have the final regex

bufdev: Also the exclude set as it is now just bans `.` at the end of the PR title, we likely want to allow `.` in the middle though

PR #1739, opened 2023-01-17
Forbid terminal punctuation in PR titles

akshayjshah:


.github/workflows/pr-title.yaml
```diff
@@ -19,9 +19,9 @@ jobs:
       - uses: morrisoncole/pr-lint-action@v1.7.0
```
rubensf, line 19: Might be worth to also use the sha here, particularly if we'll roll it wide across our repos.

.github/workflows/pr-title.yaml
```diff
@@ -19,9 +19,9 @@ jobs:
...
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
-          title-regex: "^[A-Z].*[^.]$"
+          title-regex: "^[A-Z].*[^.?!]$"
```
rubensf, line 22: We could be a bit more aggressive and forbid some other punctuation eg `-,;:`

rubensf (approved):

.github/workflows/pr-title.yaml
```diff
@@ -16,12 +16,12 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: morrisoncole/pr-lint-action@v1.7.0
+      - uses: morrisoncole/pr-lint-action@327d08270eeda69f390c2073935506556987b9a1 # v1.7.0
         with:
           repo-token: "${{ secrets.GITHUB_TOKEN }}"
-          title-regex: "^[A-Z].*[^.]$"
+          title-regex: "^[A-Z].*[^.?!,-;:]$"
```
bufdev, line 22: I don't think this is what we want - unless I'm misreading this, this just bans this character set at the end of the PR. What we want is to ban the PR ending in a period. The rest of the character set we might want to universally ban, but I don't know. Some examples:

- "Update buf.build to do X" - this should be allowed, but I think (?) your intent would be to ban, but I also think this regex (correctly) allows this
- "Specify X as key:value pairs" - I think this should be allowed, I think your intent is to ban, but I don't think this PR bans
- "Make foo-bar" - same as last two

I didn't mean for this to lead to extremely strict PR titles, just to make sure people stopped lowercasing PR titles and to stop ending them in periods, I think if we get more restrictive that might be beyond what we want to do. As a cut set, I'd say perhaps ";" should be universally banned, but idk about the rest.

bufdev: Oh, I think I'm mistaken - I think you actually DID intend for this PR to only enforce the last character, and in that context, this seems right.

bufdev (approved):

bufdev:
PR title misleading, three Pinocchios

.github/workflows/pr-title.yaml
akshayjshah: Yeah - I meant for readers to understand "Forbid terminal punctuation" as "Forbid punctuation at the end of the PR title." The latter would have been clearer.

PR #1740, opened 2023-01-17
support package names for filtering images

jhump:
When iterating with @bufdev on the functionality of the new BSR reflection endpoint, we came up with this, which describes being able to filter a schema by package name: https://github.com/bufbuild/reflect/blob/34fca2b20ed4a4380b5377ab891cea449f423d17/buf/reflect/v1beta1/file_descriptor_set.proto#L69-L81.

This provides that ability in the underlying `bufimageutil` package that implements the filtering.


private/bufpkg/bufimage/bufimageutil/image_index.go
```diff
@@ -102,11 +114,27 @@ func newImageIndexForImage(image bufimage.Image, opts *imageFilterOptions) (*ima
...
+			case *descriptorpb.EnumValueDescriptorProto, *descriptorpb.OneofDescriptorProto:
+				// do not add to package elements; these elements are implicitly included by their enclosing type
+			case *descriptorpb.FieldDescriptorProto:
+				// only add to elements if an extension (regular fields implicitly included by containing message)
+				includeInIndex = d.Extendee != nil
+			default:
+				includeInIndex = true
+			}
+
+			if includeInIndex {
```
jhump, line 130: Because of this, type filtering is actually now _faster_ than before. While there is some small overhead of indexing the packages, it is dwarfed by the improvement of no longer including fields and enum values in the index (the most numerous elements in a typical schema). So we do a lot less map insertions and map growing/re-allocating. The fields, oneofs, and enum values aren't actually useful in the index anyway since they can't be referenced by other elements, are not allowed elements to name in the include list, and are never filtered (an enum always includes all of its values; a message always includes all of its oneofs and regular fields).

Before:
```
BenchmarkFilterImage-10    	   28686	     41532 ns/op	   50227 B/op	     186 allocs/op
```
After:
```
BenchmarkFilterImage-10    	   45050	     26201 ns/op	   20824 B/op	     196 allocs/op
```

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -366,6 +405,30 @@ func (t *transitiveClosure) addFile(file string, imageIndex *imageIndex, opts *i
...
+	for _, descriptor := range pkg.elements {
+		if err := t.addElement(descriptor, "", false, imageIndex, opts); err != nil {
+			return err
+		}
+	}
+	for _, subPkg := range pkg.subPackages {
+		if err := t.addPackage(subPkg, imageIndex, opts); err != nil {
+			return err
+		}
+	}
```
robbertvanginkel, line 428: I don't think I follow why subpackages would be included here. This now seems to behave a bit like a `**/*` glob, whereas I expected something more like `*`: specifying a package would only be a shorthand for specifying all types in that package.

Following the comment on that schema as the reasoning: https://github.com/bufbuild/reflect/blob/34fca2b20ed4a4380b5377ab891cea449f423d17/buf/reflect/v1beta1/file_descriptor_set.proto#L75-L76
> Otherwise, the resulting schema contains only the named elements and all of their dependencies

subpackages aren't necessarily dependencies?

jhump: I think that's fine. But I want @bufdev to chime in too, just in case (since he helped wordsmith the description of the reflection service, which describes exactly this behavior).

One argument _for_ doing this is that packages in Protobuf are really nested namespaces -- so by that mental model, everything in package `foo.bar` _is also in_ package `foo`. But I can definitely see how that behavior could be surprising to most users and even undesirable in many cases.

bufdev: I'd argue it should behave as more of a glob, as it leads to two things:

1. This is how I think we think about packages in general, which @jhump just described well - they're really nested namespaces.
2. To double down on this point, consider the case of the message `foo.bar.Baz` and the nested message `foo.bar.Baz.Bat`. I think we'd all agree that a filter for `foo.bar.Baz` should include the information for `foo.bar.Baz.Bat`, and saying that `foo` includes `foo.bar` is a natural extension of this in the name of consistency.

robbertvanginkel: As a concrete counter point, go packages are nested (like directories) but referring to one doesn't implicitly refer to all subpackages. You need to use `/...` syntax to refer to the nested subset for building or querying.

Packages and sub-packages dont evoke the same nested thoughts for everyone, I rather think of them as a grouping of a types by prefix. When specifying just a package path, I'd expect it only selects all the types in that package (for the nested message example, I'd probably expect that to end up in the pacakge for which "root" nested message is). This matches shells (`ls foo` for toplevel, `ls foo/**/*` for recursive under foo), and go (`go build ./foo`, `go build ./foo/...`).

If someone wants to select a pakcage's contents without subpackages, this closes the door to that. I'd prefer to go the direction where we have an explicit recursive/nested syntax for who needs it.

jhump: > I think we'd all agree that a filter for `foo.bar.Baz` should include the information for `foo.bar.Baz.Bat`

I can actually see this both ways, just like the package. The main reason I would indeed _expect_ `foo.bar.Baz.Bat` to be included with `foo.bar.Baz` is because, per typical practices (and even recommendations), `foo.bar.Baz.Bat` will be _referenced_ by `foo.bar.Baz`, like via the type of a field. But if there were no such reference and `foo.bar.Baz.Bat` were actually a separate and independent type that just happened to be nested, I could see a case for not including it.

Anyhow, I guess that's a long-winded way of saying I can see it both ways. I'm happy to make sure this package implements whichever way we agree on. I do see @robbertvanginkel's argument, since I think many Protobuf users bring a mental model of packages from their programming language of choice, which likely differs from Protobuf's model. (Most concepts like this in Protobuf mirror C++, which is sometimes a shame as C++ tends to do things differently than most other languages.)

Is this worth a zoom chat to discuss, to all get on the same page?

robbertvanginkel: The way the type flag in the cli works is that if you give it a message type, it thins anything that isn't required to decode a message of the provided type. We follow edges depending on if the type is referenced, not based on package hierarchy.

Nested messages are also only included if they're required to decode its parent. In the cli today if you would have:
```
package foo
message Bar {
	message Baz {}
	Baz x = 1;
}
```
Then `buf build -o - --exclude-source-info --type foo.Bar | buf convert --type=google.protobuf.FileDescriptorSet | jq` will include `foo.Bar.Baz`, but if the where the nested message isn't used in the type
```
package foo
message Bar {
	message Baz {}
	string x = 1;
}
```
the output will omit it. Only in the reverse case (`--type foo.Bar.Baz`) the `foo.Baz` message will be included even though it isn't necessary to decode a `foo.Bar.Baz`, but Josh specifically did some work to have `foo.Baz` be an empty message in this case (so it only provides a namespace).

The current behavior provides you the smallest filedescriptorset possible to decode all messages/rpcs/services in the `type` param, based on type reachability in the graph. Extending this to packages by inclusion based on type name prefix rather than decoding/encoding reachability does not seem consistent to me.

Happy to zoom to get on the same page too.

jhump:
Per F2F, deciding to not include all sub-packages. Instead, we'll introduce some kind of glob syntax later to accomplish this. (We may want other kinds of filtering options, too, like field-level filtering or excluding elements.)

Opened https://github.com/bufbuild/reflect/pull/6 to capture changes in the spec for the `FileDescriptorSetService`. And updated this PR.

@robbertvanginkel, this is ready for another look.

robbertvanginkel (approved):

PR #1741, opened 2023-01-17
Swallow stderr for protoc-gen-swift

bufdev:


bufdev:
We should likely actually go a step further and run stderr through an effective pipe, and filter the specific error message that we think is incorrect.

bufdev:
This now specifically swallows the error message we care about. I've manually tested this in a bunch of scenarios, and it appears to swallow properly, but I have not test other stderr from protoc-gen-swift. @rebello95 can you try to test a case where protoc-gen-swift is known to print a different error message to stderr, and make sure the result is unchanged? There may be something wrong with newlines, ie i could see something like:

```
protoc-gen-swift: other error message
WARNING: 3.2...
```

And this being not properly detected.

private/pkg/app/appproto/appprotoexec/binary_handler.go
```diff
@@ -60,17 +60,61 @@ func (h *binaryHandler) Handle(
...
+				// https://github.com/apple/swift-protobuf/blob/c3d060478fcf1f564be0a3876bde8c04247793ae/Sources/protoc-gen-swift/main.swift#L244
+				//
+				// Note that our heuristic as to whether this is protoc-gen-swift or not for isProtocGenSwift
+				// is that the binary is named protoc-gen-swift, and buf/protoc will print the binary name
+				// before any message to stderr, so given our protoc-gen-swift heuristic, this is the
+				// error message that will be printed.
+				//
+				// Tested manually on Mac.
+				// TODO: Test manually on Windows.
+				[]byte("protoc-gen-swift: WARNING: unknown version of protoc, use 3.2.x or later to ensure JSON support is correct.\n"),
```
bufdev, line 109: I'm not actually sure if it's us or the plugin that produces the prefix, come to think of it, it may be the plugin.

rebello95: It's the plugin https://github.com/apple/swift-protobuf/blob/c3d060478fcf1f564be0a3876bde8c04247793ae/Sources/protoc-gen-swift/FileIo.swift#L19

rebello95 (approved):
I verified that it does properly pipe through [this error](https://github.com/apple/swift-protobuf/blob/44779d80a847bdda1555c0d7f4459d6b420b26f5/Sources/protoc-gen-swift/main.swift#L71) with the following config:

```
version: v1
plugins:
  - name: swift
    opt: Visibility=Public,Garbage=abc
    out: gen/swift
```

Output:

```
rebello@michaels-mbp eliza % buf generate    
Failure: plugin swift: Unknown generation parameter 'Garbage'
```

Without the garbage parameter, it succeeds without the filtered warning.

bufdev:
@rebello95 I did some cleanup, can you verify this still works?

rebello95:
@bufdev verified still works

PR #1754, opened 2023-01-24
Add system user type

pkwarren:
In order to automate certain background tasks on the BSR, add a system user type. These differ from personal users as they cannot login or issue tokens, and from machine users as they aren't manageable by server admins.


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -63,6 +63,7 @@ enum UserType {
   USER_TYPE_UNSPECIFIED = 0;
   USER_TYPE_PERSONAL = 1;
   USER_TYPE_MACHINE = 2;
+  USER_TYPE_SYSTEM = 3;
```
mfridman, line 66: If these users can't be interacted with, should they be exposed at the API layer, or is this an implementation detail of the BSR?

I presume we still might want to surface them in the UI, right?

mfridman (approved):
one question, otherwise lgtm.

proto/buf/alpha/registry/v1alpha1/user.proto
pkwarren: I think they should still be visible as owners of organization, but otherwise not able to be managed from admin UI or other pages.

PR #1756, opened 2023-01-26
Update BUF_TOKEN format

tonyli233:


private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -121,8 +119,6 @@ func TestNewTokenSetFromString(t *testing.T) {
 	assert.Error(t, err)
 	_, err = NewTokenProviderFromString("default1,default2")
 	assert.Error(t, err)
-	_, err = NewTokenProviderFromString("invalid@invalid@invalid")
-	assert.Error(t, err)
 	_, err = NewTokenProviderFromString("")
 	assert.NoError(t, err)
```
bufdev, line 123: Other test cases:

```
"token1" # should pass
"token1@host1" # should pass
"token1@host1,token2@host2" # should pass
"token1@host1,token2@host2," # should fail
",token1@host1,token2@host2," # should fail
",token1@host1" # should fail
"token1@host1," # should fail
"token1@" # should fail
"token1@host1@" # should fail
"@token1" # should fail
"@token1@host1" # should fail
"token1@host1,token2" # should fail
"token1@host1,token1@host2 # should pass
"token1@host1,token2@host1 # should fail

File: private/bufpkg/bufconnect/static_token_provider.go [hunk 2]
```diff
@@ -45,55 +36,48 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
 // The special characters `:`, `@` and `,` are used as the splitters. The usernames, tokens, and
 // remote addresses does not contain these characters since they are enforced by the rules in BSR.
 func newTokenProviderFromString(token string, isFromEnvVar bool) (TokenProvider, error) {
+	// Tokens for different remotes are separated by `,`. Using strings.Split to separate the string into remote tokenToAuthKey.
+	tokens := strings.Split(token, ",")
+	if len(tokens) <= 1 {
+		return staticTokenProvider{
```
bufdev, line 42: instead of doing if/then logic within `staticTokenProvider`, you should create two different structs:

```
type singleTokenProvider struct {
  token string  
  isFromEnvVar bool
}

type multipleTokenProvider struct {
  addressToToken map[string]string
  isFromEnvVar bool
} 
```

And implement `TokenProvider` on each.

Then, here, do a switch on `len(tokens)`

```
switch tokens := strings.Split(token, ","); len(tokens) {
  case 0:
    return nil, fmt.Errorf("...")
  case 1:
    return newSingleTokenProvider(tokens[0], isFromEnvVar)
  default:
    return newMultipleTokenProvider(tokens, isFromEnvVar)
}
```

you should also check in the `newSingleTokenProvider` to make sure that the token contains none the characters `,@:`, and also check that the result tokens in `newMultipleTokenProvider` contain those characters

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +36,48 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
+			setBufTokenEnvVar: isFromEnvVar,
+			token:             token,
+		}, nil
+	}
 	tokenProvider := &staticTokenProvider{
 		setBufTokenEnvVar: isFromEnvVar,
-		tokenToAuthKey:    make(map[string]authKey),
+		keyPairs:          make(map[string]string),
```
bufdev, line 49: `addressToToken`

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +36,48 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
+	// Tokens for different remotes are separated by `,`. Using strings.Split to separate the string into remote tokenToAuthKey.
+	tokens := strings.Split(token, ",")
+	if len(tokens) <= 1 {
+		return staticTokenProvider{
+			setBufTokenEnvVar: isFromEnvVar,
+			token:             token,
+		}, nil
+	}
 	tokenProvider := &staticTokenProvider{
 		setBufTokenEnvVar: isFromEnvVar,
```
bufdev, line 48: `isFromEnvVar`

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +36,48 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
-			if _, ok = tokenProvider.tokenToAuthKey[remoteAddress]; ok {
-				return nil, fmt.Errorf("cannot parse token: %s, repeated token for same BSR remote: %s", token, remoteAddress)
-			}
-			tokenProvider.tokenToAuthKey[remoteAddress] = ak
-		} else {
-			if tokenProvider.defaultToken != "" {
-				return nil, fmt.Errorf("cannot parse token: two buf token provided: %q and %q", token, tokenProvider.defaultToken)
-			}
-			tokenProvider.defaultToken = token
+		key, hostname, found := strings.Cut(token, "@")
```
bufdev, line 52: `Cut` will only split into two, no matter how many `@`s there are. You want to verify there is exactly one.

```
split := strings.Split(token, "@")
if len(split) != 2 {
  return nil, fmt.Errorf("...")
}
```

private/bufpkg/bufconnect/static_token_provider.go
tonyli233: Should we allow the case when `len(tokens)==0`? Is it possible to have empty token?

private/bufpkg/bufconnect/static_token_provider.go
bufdev: you should make sure that both sides of the split are non-empty as well

private/bufpkg/bufconnect/static_token_provider.go
tonyli233: Yeah I also just realized strings.split never return empty slice.

tonyli233: These cases are handled since they do not contain `@`

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +37,90 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
+// singleTokenProvider is used to provide set of authentication tokenToAuthKey.
+type singleTokenProvider struct {
+	// true: the tokenSet is generated from environment variable tokenEnvKey
+	// false: otherwise
+	setBufTokenEnvVar bool
+	token             string
+}
+
+func newSingleTokenProvider(token string, isFromEnvVar bool) (*singleTokenProvider, error) {
+	if strings.Contains(token, "@") {
```
bufdev, line 63: needs to check for `,` and `:` as well

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +37,90 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
+
+func (m *multipleTokenProvider) RemoteToken(address string) string {
+	return m.addressToToken[address]
+}
+
+func (m *multipleTokenProvider) IsFromEnvVar() bool {
+	return m.isFromEnvVar
+}
+
+type nonTokenProvider struct{}
```
bufdev, line 118: `nopTokenProvider` not `nonTokenProvider`, "nop" is short for "no-op"

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +37,90 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
+
+func (*nonTokenProvider) RemoteToken(string) string {
```
bufdev, line 120: nit: for nop structs, you can do the methods on the struct not the struct pointer, ie `func (nonTokenProvider)`, and return `nopTokenProvider` instead of `&nopTokenProvider`

private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -104,24 +104,42 @@ func TestNewAuthorizationInterceptorProvider(t *testing.T) {
...
 	assert.Equal(t, "default", token)
-	token = tokenSet.RemoteToken("remote")
-	assert.Equal(t, "token", token)
 }
 
 func TestNewTokenSetFromString(t *testing.T) {
 	_, err := NewTokenProviderFromString("default")
 	assert.NoError(t, err)
-	_, err = NewTokenProviderFromString("user1:token1@remote,user2:token2@remote")
+	_, err = NewTokenProviderFromString("token1@host1")
```
bufdev, line 118: tests should include verifying the values from the returned token provider, ie `assert.Equal(t, "token1", tokenProvider.RemoteToken("host1")`

private/bufpkg/bufconnect/static_token_provider.go
```diff
@@ -45,55 +37,90 @@ func NewTokenProviderFromString(token string) (TokenProvider, error) {
...
-	if !found {
-		return fmt.Errorf("cannot parse remote token: %s", s)
+func newMultipleTokenProvider(tokens []string, isFromEnvVar bool) (*multipleTokenProvider, error) {
+	addressToToken := make(map[string]string)
+	for _, token := range tokens {
+		split := strings.Split(token, "@")
+		if len(split) != 2 {
+			return nil, fmt.Errorf("invalid token: %s", token)
+		}
+		if split[0] == "" || split[1] == "" {
```
bufdev, line 96: `split[0]` needs to be checked for `:` and `,` as well for completeness

tonyli233:
Fixed the issues.

bufdev (approved):

PR #1760, opened 2023-01-26
Rename pkg/manifest to bufpkg/bufmanifest

bufdev:
First half of https://linear.app/bufbuild/issue/TCN-1064/pkgmanifest-should-be-moved-to-bufpkgbufmanifest. A change to core will be required after, which I will hand off.


pkwarren (approved):

twilly:
Can we instead move the protobuf types out of the package? That's the only dependency and the mapping funcs were there for convenience. We should have put those in a different package.

twilly (changes requested):
I don't think this is the right thing to do yet. Let me think some more before landing.

twilly:
PR #1762 is an alternative to wholesale moving the package. IMHO it's clearer but I'd like to know what others think.

bufdev:
Closing in favor of #1762

PR #1761, opened 2023-01-26
Move pkg/manifest to bufpkg/bufmanifest

pkwarren:
Move pkg/manifest as it depends on other buf specific types.


unmultimedio:
Is this different from https://github.com/bufbuild/buf/pull/1760?

bufdev:
Yea I already did this in #1760 @pkwarren lol, race condition

pkwarren:
Doh ok - didn't see that.

PR #1762, opened 2023-01-26
Move manifest<>proto mappers to bufmanifest

twilly:
The manifest module depended on protobuf when it shouldn't have. These mappers are the boundary between manifest types and protobuf types, meaning they should be in a different package in a scope that has visibility into both.

This places the mappers into bufmanifest, a new module that encapsulates the type boundaries and other utilities.

Fixes: TCN-1064


bufdev:
This seems fine to me

bufdev (commented):
Is `bufapimodule` the right package for this? It doesn't seem so to me, perhaps a new `bufmanifest` or something else?

twilly:
> Is `bufapimodule` the right package for this? It doesn't seem so to me, perhaps a new `bufmanifest` or something else?

This change (mappers in general) can be in any package that isn't the proto or manifest types (left or right side types).

I stuck these in `bufapimodule` because it makes some sense to co-locate with reading remote modules (`bufapimodule/module_reader.go` uses a bucket abstraction, `NewBucketFromManifestBlobs`, and is where to/as proto mapping is needed). `ToProtoManifestAndBlobs` becomes an outlier as it was originally moved from `command/push/push.go` use to be accessible in the BSR.

No objections here for slicing thinner and making `bufmodule`. I'll pop them over.

PR #1767, opened 2023-01-27
Custom `ChunkSize` for `Storage.WriteBucket.Put` via `PutOptions`

unmultimedio:
Allow `WriteBucket` users to pass options to set chunk size for multi-part uploads.


private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
 	Walk(ctx context.Context, prefix string, f func(ObjectInfo) error) error
 }
 
+// PutOptions are the possible options that can be passed to a Put operation.
+type PutOptions struct {
+	IfNotExists      bool
+	CustomChunkSize  bool
+	ChunkSizeInBytes int
```
bufdev, line 59: `ChunkSize int64`, sizes are typically bytes

private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
...
+}
+
+// PutOption are options passed when putting an object in a bucket.
+type PutOption func(*PutOptions)
```
bufdev, line 63: All other options structs in `storage` are private and at the end of the file, see `diff.go` as an example.

private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
...
+
+// PutIfNotExists only puts the object in the bucket if there is no other object
+// already in the same bucket path. Some implementations of
+// `storage.WriteBucket.Put` that are `ReadWriteBucket` support this option.
+// Check the implementation of the bucket you're using if you're passing this
+// option.
+func PutIfNotExists() PutOption {
```
bufdev, line 70: All other options functions follow the pattern `FunctionnameWithOptionname`, ie in this case it would be `PutWithIfNotExists`. I would likely change the name of the option to  `IfNotExist`, to match `os.IsNotExist`.

private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
...
+	return func(opts *PutOptions) {
+		opts.IfNotExists = true
+	}
+}
+
+// PutWithChunkBytesSize sets a custom chunk size in bytes for writing the
+// object. Some implementations of `storage.WriteBucket.Put` allow multi-part
+// upload, and allow customizing the chunk size of each part upload. Check the
+// implementation of the bucket you're using if you're passing this option.
+func PutWithChunkBytesSize(sizeInBytes int) PutOption {
```
bufdev, line 80: `PutWithChunkSize`

private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
...
+// option.
+func PutIfNotExists() PutOption {
+	return func(opts *PutOptions) {
+		opts.IfNotExists = true
+	}
+}
+
+// PutWithChunkBytesSize sets a custom chunk size in bytes for writing the
+// object. Some implementations of `storage.WriteBucket.Put` allow multi-part
+// upload, and allow customizing the chunk size of each part upload. Check the
```
bufdev, line 78: What does "check the implementation" mean? Callers of `storage` types should not have to concern themselves with what the implementation is at all, and all options should be defined in such a way that all implementations can reliably implement them. This is a core tenet of `storage` - callers only need to know about the API, not the implementation (really a core of all APIs). This may mean that this option is defined as "this is a suggested chunk size, implementations may choose to ignore this option".

private/pkg/storage/bucket.go
```diff
@@ -52,6 +52,38 @@ type ReadBucket interface {
 	Walk(ctx context.Context, prefix string, f func(ObjectInfo) error) error
 }
 
+// PutOptions are the possible options that can be passed to a Put operation.
+type PutOptions struct {
+	IfNotExists      bool
+	CustomChunkSize  bool
```
bufdev, line 58: Why is this needed over just checking if `ChunkSize` is not 0?

private/pkg/storage/bucket.go
```diff
@@ -61,7 +93,7 @@ type WriteBucket interface {
 	// The returned WriteObjectCloser is not thread-safe.
 	//
 	// Returns error on system error.
-	Put(ctx context.Context, path string) (WriteObjectCloser, error)
+	Put(ctx context.Context, path string, opts ...PutOption) (WriteObjectCloser, error)
```
bufdev, line 96: This is a breaking change, so before this PR is merged, all implementations will need to have PRs ready to be updated to this

bufdev (commented):
Is there any context as to why this is needed?

private/pkg/storage/bucket.go
bufdev: You could do something similar to `SetExternalPathSupported` as well, but in this case, that likely is not preferred

private/pkg/storage/bucket.go
unmultimedio: Right, but all implementations need to read and understand the passed options outside of this pkg, so probably exported struct with private fields and exported getters?

private/pkg/storage/bucket.go
unmultimedio: Yeah, @pkwarren was suggesting to also have an `UnsupportedOption` error top level here, so implementations can return it instead of ignoring a given option.

private/pkg/storage/bucket.go
unmultimedio: Because `= 0` could have a meaning, like in GCS it is disable multi-part and upload it all in a single request.

private/pkg/storage/bucket.go
twilly: Take care when casting. GCS uses `int` for chunk size.

private/pkg/storage/bucket.go
bufdev: Just to follow up on this - talked offline, @unmultimedio is right

pkwarren (approved):

twilly (approved):

PR #1776, opened 2023-01-30
Publish to NPM

paul-sachs:
Publish buf CLI to npm. This currently just adds a make target that allows us to publish.

```sh
VERSION=1.13.1 make npmpublish
```

TODOs
- [x] Hook up a github action to do this on release
- [ ] Add tests


.github/workflows/npm-publish.yaml
```diff
@@ -0,0 +1,31 @@
...
+    types: [ created ]
+
+jobs:
+  publish:
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: checkout
+        uses: actions/checkout@v3
+      - name: setup-node
+        uses: actions/setup-node@v3
```
paul-sachs, line 18: Woops, this action wasn't supposed to be committed since it clearly doesn't do what i want. Will update this one.

npm/@bufbuild/buf/scripts/install.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    // First check for the binary package from our "optionalDependencies". This
+    // package should have been installed alongside this package at install time.
+    binPath = require.resolve(`${pkg}/${subpath}`);
+  } catch (e) {
+    throw new Error(`[buf] Failed to find package "${pkg}" on the file system
+
+This can happen if you use the "--no-optional" flag. The "optionalDependencies"
+package.json feature is used by buf to install the correct binary executable
+for your current platform.
```
timostamm, line 69: I'm not convinced that this can only happen with the `--no-optional` flag. Do you remember when protoc-gen-es was written in Go? It used [this build script](https://github.com/bufbuild/protobuf-es/blob/761ce5ed5fabeae80a2bdd4c24f0b0da2dd5c8d0/make/scripts/go-build-npm.js), and [this shim](https://github.com/bufbuild/protobuf-es/blob/761ce5ed5fabeae80a2bdd4c24f0b0da2dd5c8d0/make/scripts/go-build-npm-shim.cjs), which is pretty similar to what you got here.

But we were also having issues with missing executables sometimes. From what I remember, some npm commands removed the optional dependencies that are not applicable for the current platform from the lock file. When this lock file was used to install on a different platform, npm did not install those packages (rightfully so, they were no longer in the lock file), and the binary couldn't be found.

This is the situation I meant with:

> Some npm commands remove unused optional dependencies from the lock file, and installing from a package-lock.json on a different platform yields in an unusable installation.

[Here is an issue](https://github.com/npm/cli/issues/4828) that seems related.

I might be missing something. Do we have a solution for this situation?

paul-sachs: Hmm, no i don't. Might've gone a little to overboard on trimming. I'll add a fallback.

paul-sachs: Fallback added.

make/buf/all.mk
```diff
@@ -26,7 +26,10 @@ FILE_IGNORES := $(FILE_IGNORES) \
 	private/buf/cmd/buf/command/alpha/protoc/test.txt \
 	private/buf/cmd/buf/workspacetests/other/proto/workspacetest/cache/ \
 	private/bufpkg/buftesting/cache/ \
-	private/pkg/storage/storageos/tmp/
+	private/pkg/storage/storageos/tmp/ \
+	**/node_modules/** \
```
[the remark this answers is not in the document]
paul-sachs: Woops, actually i just need to do `**/node_modules` and both git and docker will work as expected. Updating.

npm/@bufbuild/buf-darwin-arm64/README.md
```diff
@@ -0,0 +1,3 @@
+# buf
+
+This is a darwin-arm64 binary for buf. See https://github.com/bufbuild/buf for details.
```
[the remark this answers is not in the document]
bufdev: Unless @akshayjshah @timostamm @smaye81 @dimitropoulos have a better idea, I don't have a better idea, so sure works for me.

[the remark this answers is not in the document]
timostamm: An alternative would be to generate the package directories / files from some template, and only keep them for publishing to npmjs.com. But I think it's a good call to have them committed to the repository for transparency.

The location of these files is pretty much up to us (there is no requirement for them to be in a specific path). IMO `npm` is descriptive, and seems like a reasonable choice.

npm/@bufbuild/buf/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@bufbuild/buf",
+  "description": "The buf CLI is a tool for working with Protocol Buffers.",
```
timostamm, line 3: This description is what shows up in a search on npmjs.com, with `npm info`, and it's used by most IDEs when looking at imports or project dependencies.

Maybe we should use that chance to consolidate the descriptions across the different ecosystems? 

[On scoop](https://scoop.sh/#/apps?q=buf&s=0&d=1&o=true), we use:

> A tool for working with Protocol Buffers APIs, offering a range of features not found in the standard protoc compiler.

On [aur](https://aur.archlinux.org/packages/buf), it's 

> A new way of working with Protocol Buffers

And our home page says [about the CLI](https://buf.build/product/cli/):

> The Buf CLI is a one stop shop for your local Protocol Buffers needs

I'm only raising this because we can't retroactively change this description without cutting a release.

IMO we should definitely have a README.md here, because that is what shows up on npmjs.com ([example](https://www.npmjs.com/package/@bufbuild/protoc-gen-es)), and we have to cut a release to add / change it as well. 

Our top-level README.md is actually pretty nice, so perhaps it is sufficient to link it here? If we do that, we have to update [this link](https://github.com/bufbuild/buf/blob/a1a4fbd9ddebcded1a3ef1db30dcba7180919dcc/README.md?plain=1#L155) (and perhaps others), which would otherwise not work as expected when rendered on npmjs.com.

make/buf/scripts/npmpublish.bash
```diff
@@ -0,0 +1,69 @@
...
+	NPM_VERSION="${VERSION}-${REVISION}"
+fi
+
+update_npm_platform_files() {
+	mkdir -p "npm/@bufbuild/buf-${2}/bin"
+
+	if [[ ${2} =~ "win32" ]]
+	then
+		# Download the binaries
+		curl -Lf "https://github.com/bufbuild/buf/releases/download/v${VERSION}/buf-${1}.exe" -o "npm/@bufbuild/buf-${2}/bin/buf.exe"
```
timostamm, line 32: If I understand correctly, we have to make a regular release first, then run this as a follow-up to publish to npm, right? Seems okay to me for a start, just wanted to understand.

.github/workflows/npm-publish.yaml
```diff
@@ -0,0 +1,23 @@
+name: npm-publish
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: The version you intend to publish (eg x.y.z)
```
timostamm, line 7: We have to be very cautious with fat fingers when publishing to npmjs.com. It is basically impossible to unpublish a package without the help of npm support, so an inadvertent `npm-publish` with `v11.3.2` would be pretty difficult to fix up.

But if I read the code correctly, we should fail if the version isn't already released, is that correct?

npm/@bufbuild/buf/package.json
paul-sachs: > Maybe we should use that chance to consolidate the descriptions across the different ecosystems?

All up for consolidation. I'll defer to @bufdev or @akshayjshah on this one.

> IMO we should definitely have a README.md here, because that is what shows up on npmjs.com ([example](https://www.npmjs.com/package/@bufbuild/protoc-gen-es)), and we have to cut a release to add / change it as well.
> 
> Our top-level README.md is actually pretty nice, so perhaps it is sufficient to link it here? If we do that, we have to update [this link](https://github.com/bufbuild/buf/blob/a1a4fbd9ddebcded1a3ef1db30dcba7180919dcc/README.md?plain=1#L155) (and perhaps others), which would otherwise not work as expected when rendered on npmjs.com.

Hmm not sure i know exactly what you mean, but I do agree we should have a better README.md. If we could reuse the top level readme, that would be ideal.

make/buf/scripts/npmpublish.bash
paul-sachs: Yes, this is most useful when triggered from a release in the `npm-publish.yaml` github action. I wanted to split this process out from the rest of the release to keep things loose and flexible.

.github/workflows/npm-publish.yaml
paul-sachs: It will fail if a proper version isn't provided. I believe there is a little room for this to fail though, so I'll add a few extra checks.

npm/@bufbuild/buf/package.json
timostamm: I mean:

```sh
ln -s ../../../README.md npm/@bufbuild/buf
```

This would give you a link at npm/@bufbuild/buf/README.md pointing to README.md.

Not important.

But if we do use this README, we have to make sure all links are absolute.

paul-sachs: Ah right, with a symlink. Makes sense. I guess we'll lose that if it lives in it's own repo.

timostamm: Am not sure the symlink would have worked. Considering that we fetch binaries from the release page, we might just as well fetch the README.md from the repo?

paul-sachs: Sure, why not.

PR #1777, opened 2023-01-30
Add to event proto for logging created and deleted tokens

seankimdev:
- Adds the required event types, resources and payloads for logging token create/delete.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -319,3 +324,15 @@ message PayloadCuratedPluginCreated {
   // owner_name is the name of the owner of the plugin.
   string owner_name = 2;
 }
+
+message PayloadTokenCreated {
+  // owner_id is the id of the owner of the token.
+  string owner_id = 1;
+  // token_expiry_time is the expiry time of the token.
+  google.protobuf.Timestamp token_expiry_time = 2;
```
pkwarren, line 332: Do we want to add the token id (so we can correspond created/deleted audit events)?

seankimdev: The token id and name will be logged as the affected resource. Do we also want it logged in the payload?

pkwarren: Nope - sorry missed that part.

pkwarren (approved):

PR #1779, opened 2023-01-31
Migrate trace instrumentation to otel and remove extraneous observability code

doriable:
This migrates and simplifies the observability code:
- Move from Opencensus to Opentelemetry for tracing instrumentation
- On the CLI, we only need a logging exporter for our tracing instrumentation and debug logs, so this removes the extra code for other forms of instrumentation
- Migrate all spans to use Opentelemetry and properly record error messages to the spans, which is useful for extending our debug logs down the road


private/buf/buffetch/internal/reader.go
```diff
@@ -245,7 +253,8 @@ func (r *reader) getArchiveBucket(
 			return nil, err
 		}
 	default:
-		return nil, fmt.Errorf("unknown ArchiveType: %v", archiveType)
+		err := fmt.Errorf("unknown ArchiveType: %v", archiveType)
+		return nil, err
```
joshcarp, line 257:
```suggestion
		return nil, fmt.Errorf("unknown ArchiveType: %v", archiveType)
```

private/buf/buffetch/ref_parser.go
```diff
@@ -173,23 +180,28 @@ func newSourceOrModuleRefParser(logger *zap.Logger) *refParser {
...
 func (a *refParser) GetRef(
 	ctx context.Context,
 	value string,
 ) (Ref, error) {
-	ctx, span := trace.StartSpan(ctx, "get_ref")
+	ctx, span := a.tracer.Start(ctx, "get_ref")
 	defer span.End()
 	parsedRef, err := a.getParsedRef(ctx, value, allFormats)
 	if err != nil {
+		span.RecordError(err)
```
joshcarp, line 195: I know that we tend to dislike named return types but in all of these instances it'd make everything a lot easier:
```go

func (a *refParser) GetRef(
	ctx context.Context,
	value string,
) (ref Ref, err error) {
	defer func(){ 
		if err != nil {
			span.RecordError(err)
			span.SetStatus(codes.Error, err.Error())
		}
	}
	ctx, span := a.tracer.Start(ctx, "get_ref")
	defer span.End()
	parsedRef, err := a.getParsedRef(ctx, value, allFormats)
```

Also means that when the function grows there isn't the risk of someone forgetting to record the error

joshcarp: Another idea would be to use generics and whatever `func (ctx, string) (T, error)` function gets passed in gets traced with the name of the function:
```go
func TraceFunc[T any](tracer trace.Tracer, f func(ctx context.Context, value string) (T, error), ctx context.Context, value string) (T, error) {
	funcName := runtime.FuncForPC(reflect.ValueOf(f).Pointer()).Name()
	ctx, span := tracer.Start(ctx, funcName)
	retVal, err := f(ctx, value)
	if err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())
	}
	return retVal, err
}
```

joshcarp: This is wayyy outside the scope of this PR lol, just bike shedding :)

doriable: Yeah, this one is kind of difficult for me, because I didn't want to heavily warp the patterns and texture of the code base around the tracing instrumentation. There needs to be a decision made here around the way we want to do debug logs before we commit to larger changes. The debug log instrumentation patterns are also a little bit inconsistent, so it's also important to figure out the rules.

I'll take a look at places where it might greatly benefit from the `defer` error handler, but otherwise, I am inclined to try and keep it as consistent as possible with existing patterns. I do agree that it can be a bit brittle, especially for long functions, so I think that's where I'd like to start with the error handler.

joshcarp (commented):
This is a very large code change, and adding two lines to all these functions seems to be really messy and error prone. 

On one hand it's going to be a pain to keep this updated, and as more functionality is added to buf cli this will 100% get forgotten about. 
Personally I don't think that tracing in this manner adds anything; it's essentially recreating a stack trace. 

The tracing currently is recreating something that's achieved by https://github.com/pkg/errors which provides a stack trace of where the error was returned from. Unfortunately this package is archived now, IMO we need to think more about how we should be printing out error logs/traces in the CLI, and maybe there's some way of replicating what pkg/errors does and get stack traces for free.

private/buf/bufwire/image_writer.go
```diff
@@ -51,8 +52,14 @@ func (i *imageWriter) PutImage(
 	asFileDescriptorSet bool,
 	excludeImports bool,
 ) (retErr error) {
-	ctx, span := trace.StartSpan(ctx, "put_image")
+	ctx, span := otel.GetTracerProvider().Tracer("bufbuild/buf").Start(ctx, "put_image")
```
joshcarp, line 55: _maybe_ it might be worth putting consts like `"bufbuild/buf"` into a new package `github.com/bufbuild/buf/private/pkg/buftrace` package along with a helper function:

```go
def checkError(err error, span trace.Span){
	if err != nil {
		span.RecordError(err)
		span.SetStatus(codes.Error, err.Error())
	}
}
```
whatcha think?

joshcarp (approved):

private/buf/bufwire/image_writer.go
doriable: Hmmm -- not a bad thought, actually. Let me mull over this and see if we can put up a follow-up refactor...

Issue #1784, opened 2023-02-01
Fix appprotoexec go:build tags


bufdev:
https://github.com/bufbuild/buf/blob/d59feb33a081913be2817f7397e1f85287ce270c/private/pkg/app/appproto/appprotoexec/appprotoexec_go18.go#L15
https://github.com/bufbuild/buf/blob/d59feb33a081913be2817f7397e1f85287ce270c/private/pkg/app/appproto/appprotoexec/appprotoexec_go19.go#L15

The intention seems to be to do one thing for Golang <1.19, and one thing for Golang >=1.19, however the build tags just exclude Golang 1.19, meaning the same behavior will happen for ...,1.16,1.17,1.18,1.20,1.21,...

jchadwick-buf:
I'm going to take a look. I am a bit confused though: As far as I understood, the version build tags should work such that Go 1.19 *and later versions* have the `go1.19` flag, so Go 1.20 should also match `go1.19`.

From the [docs]:

> a term for each Go major release, through the current version: "go1.1" from Go version 1.1 onward, "go1.12" from Go 1.12, and so on.

Maybe the filename suffix constraints are accidentally kicking in and doing something weird?

[docs]: https://pkg.go.dev/cmd/go#hdr-Build_constraints (go command docs, Build constraints section)

bufdev:
Nope, you're right, closing.

PR #1785, opened 2023-02-01
Update to Go 1.20

bufdev:
Need to wait for the docker images to be published.
Also depends on https://github.com/bufbuild/buf/issues/1784


jchadwick-buf:
I'm suspicious of this:
```diff
-          go-version: 1.19.5
+          go-version: 1.20
```
I think it's loading Go 1.2 instead of 1.20, because YAML is fun. I think we need to add quotes to the 1.20.

jchadwick-buf:
Locally this also fails due to https://github.com/golangci/golangci-lint/issues/3420, so I assume that will hit us as well.

mfridman:
We can try and bump golanci to the latest version [`v1.51.0`](https://github.com/golangci/golangci-lint/releases/tag/v1.51.0) to resolve that issue.

https://github.com/bufbuild/buf/blob/d59feb33a081913be2817f7397e1f85287ce270c/make/go/dep_golangci_lint.mk#L12

jchadwick-buf:
Locally, that works for me and after doing so, `make all` passes. I think #1784 is invalid and not blocking this PR.

bufdev:
Docker images are now there

make/buf/all.mk
```diff
@@ -116,7 +116,7 @@ bufgeneratesteps:: \
 
 .PHONY: bufrelease
 bufrelease: $(MINISIGN)
-	DOCKER_IMAGE=golang:1.19.5-bullseye bash make/buf/scripts/release.bash
+	DOCKER_IMAGE=golang:1.20-bullseye bash make/buf/scripts/release.bash
```
bufdev, line 119: There is a `1.20.0-bullseye` as well, and we traditionally have locked to micros, but with our workflows for Docker, I don't think this actually makes as much sense. We may want to just lock to minors for Docker going forward, so we always push the latest.

.github/workflows/ci.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: setup-go
         uses: actions/setup-go@v3
         with:
-          go-version: 1.19.5
+          go-version: '1.20'
```
bufdev, line 17: We may want to update this to `1.20.x`

jchadwick-buf (approved):
LGTM as far as I can tell. Seems to work fine locally.

bufdev:
We're going to need to run this on a Windows box to figure out what is going on, I've re-run the job a few times and it appears as though there is an actual issue. On glancing at the 1.20 release docs (including changes to the stdlib), it's not immediately apparent of what change in 1.20 would have caused this.

PR #1787, opened 2023-02-02
When filtering image, include type references in Any messages in option values

jhump:
This adds code to explore messages that appear in option values, recursing through any nested structure to find instances of `google.protobuf.Any`. If it finds an instance and it refers to a known message type, add that as an edge to the graph.

If you check out the new testdata file `d.proto` with message `NormalMessageSyntaxValidType`: it does _not_ import the referred message type and uses the normal syntax instead of the extended syntax. But when we filter the image, we end up synthesizing the import because we recognize the type.

The testdata file `e.proto` includes a reference to a bogus, non-existing type. In this case, everything still works, but there is no edge it can add to the graph. (Consumers of the Any will, of course, be unhappy...)

The Any messages in the test outputs are ugly because `protoprint` doesn't try to produce the extended any syntax.

Fixes TCN-1101

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
+++ b/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
@@ -27,6 +27,10 @@ import (
 	"google.golang.org/protobuf/types/descriptorpb"
 )
 
+const (
+	anyFullName = "google.protobuf.Any"
+)
+
 var (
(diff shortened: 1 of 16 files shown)


robbertvanginkel (approved):
Nice! The protoprint output does get a little wieldy, didn't know it would resort to the `<>` message syntax.

In any case this is great!

PR #1793, opened 2023-02-03
Pull out logic from git-ls-files-unstaged, refactor to use better logic, and add logic to license-header

bufdev:
This PR does a few things:

- This pulls out the logic from `git-ls-files-unstaged` into the `git` package, so it can be reused.
- This uses @akshayjshah's logic in our OSS Makefile instead of the pre-existing logic, as there was a rare bug in the existing logic.
- This makes it so that if you call `license-header` with no arguments, `license-header` will by default do the equivalent of `git-ls-files-unstaged | xargs license-header`.
- This adds a `--ignore` flag to license-header that can replicate what we used `grep -e` for.

The result is we can make it so that our tools can just call `license-header` as opposed to doing other complicated logic with `git-ls-files, comm, git-ls-files-unstaged`.

I've QA'ed this manually and extensively, no worse than what was there before.

A few notes:

- This should also work on Windows now, but not for sure. But again, no worse than before, and we don't do development with these commands on Windows at the moment.
- The code style in `git` does not follow our current gold standard, especially with options, but instead it attempts to maximize local consistency, i.e. it matches the style already in `git`.
- Once we migrate our existing usage of `license-header`, we can get rid of the >1 args logic, so that there is no confusion. The current behavior with `--ignore` is a little weird, as this flag does not work with >1 args (this is documented).
- We could later extend `--ignore` to take regexes, but there's no need for us to do so at the moment.

### DIFF
--- a/private/pkg/git/cmd/git-ls-files-unstaged/main.go
+++ b/private/pkg/git/cmd/git-ls-files-unstaged/main.go
@@ -14,21 +14,16 @@
 
 // Package main implements a file lister for git that lists unstaged files.
 //
-// This does not list unstaged deleted files, and does list unignored files that are not added.
-// This ignores non-regular files.
-//
-// This is used for situations like license headers where we want all the potential git files
-// during development
+// See the documentation for git.Lister for more details.
 package main
 
 import (
 	"context"
-	"os"
 	"strings"
 
 	"github.com/bufbuild/buf/private/pkg/app"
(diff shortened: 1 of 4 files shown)


akshayjshah (approved):
LGTM.

To my eye, the logic would be clearer without the helper functions for string manipulation (`stringSliceExcept`, `filterIgnorePaths`, etc.). If helpers like this are the prevailing style in this section of the code, though, it's fine.

PR #1801, opened 2023-02-07
Add initial buf.plugin.yaml support for Maven registry

stefanvanburen:
There will be more added here, but this adds the initial message for the configuration and adds the registry to the enum.


private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -281,6 +295,15 @@ type ExternalNPMRegistryConfig struct {
 	ImportStyle string `json:"import_style,omitempty" yaml:"import_style,omitempty"`
 }
 
+// ExternalMavenRegistryConfig is the external registry configuration for a Maven plugin.
+type ExternalMavenRegistryConfig struct {
+	Deps []struct {
+		GroupID    string `json:"group_id,omitempty" yaml:"group_id,omitempty"`
+		ArtifactID string `json:"artifact_id,omitempty" yaml:"artifact_id,omitempty"`
+		Version    string `json:"version,omitempty" yaml:"version,omitempty"`
```
pkwarren, line 303: What do you think about using a single GAV string (which supports things like classifier, optional, type, etc.) instead? This would simplify configuration in `buf.plugin.yaml` and any validation client side and let us support more types of dependencies in the future without requiring a CLI release.

stefanvanburen: that makes sense to me - done in https://github.com/bufbuild/buf/pull/1801/commits/2e0cf94ec4a872892a85cdbe0bd8f62db442ad1a. Assuming we'll do validation server-side.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -281,6 +295,13 @@ type ExternalNPMRegistryConfig struct {
 	ImportStyle string `json:"import_style,omitempty" yaml:"import_style,omitempty"`
 }
 
+// ExternalMavenRegistryConfig is the external registry configuration for a Maven plugin.
+type ExternalMavenRegistryConfig struct {
+	Deps []struct {
```
pkwarren, line 300: What do you think about simplifying this in the config to just use `[]string`? It isn't extensible in the future but it results in a nicer config file IMO.

stefanvanburen: so instead of:

```yaml
registry:
  maven:
    deps:
      - gav: "io.grpc:grpc-core:1.52.1"
      - gav: "io.grpc:grpc-protobuf:1.52.1"
      - gav: "io.grpc:grpc-stub:1.52.1"
```

you'd have:

```yaml
registry:
  maven:
    - gav: "io.grpc:grpc-core:1.52.1"
    - gav: "io.grpc:grpc-protobuf:1.52.1"
    - gav: "io.grpc:grpc-stub:1.52.1"
```
?

I think I lean a little towards the former for congruity with the existing registry specs and the extensibility (which might come into play shortly with the `lite` runtime?).

pkwarren: No I was suggesting:

```yaml
registry:
  maven:
    deps:
      - "io.grpc:grpc-core:1.52.1"
      - "io.grpc:grpc-protobuf:1.52.1"
      - "io.grpc:grpc-stub:1.52.1"
```

stefanvanburen: oh! yeah definitely, that's better.

stefanvanburen: done in [`ea99592` (#1801)](https://github.com/bufbuild/buf/pull/1801/commits/ea995926af44faddb9391d66c6f5fa14c21ba274)

pkwarren (approved):

PR #1803, opened 2023-02-07
Add protoc_path in buf.gen.yaml

### LINKED ISSUE #1462: Allow specifying `protoc` path in `buf.gen.yaml` for built-in plugins
Currently, there is no way for the plugins built into `protoc` (java, cpp, python, etc) to override in `buf.gen.yaml` what protoc to point to, as confirmed in https://github.com/bufbuild/buf/issues/1155#issuecomment-1139712630.

My use case, which I believe isn't that uncommon, is to publish protobuf+gRPC stubs for Python using multiple version for legacy clients. While `path` works fine for gRPC plugins, there is no way to swap `protoc` for protobuf bindings.

```yaml
version: v1
plugins:
  - name: python
    path: /usr/bin/protoc-3.7    # fails
    out: out/python-legacy
  - name: grpc-python-1.32       # works fine
    out: out/python-legacy
```

```
Failure: plugin python: proto: cannot parse invalid wire-format data; proto: cannot parse invalid wire-format data
```

There are multiple workarounds suggested in https://github.com/bufbuild/buf/issues/1155 but neither scales well for huge and already-established workspaces.

Also, in https://github.com/bufbuild/buf/issues/1155:
> The solution long-term is likely that the builtin plugins are moved out of protoc, which it appears the Google team is working on, so we probably won't add anything more here. 

Can't be a solution for a production-grade tool, as it simply won't work for older plugins, which is a primary use case to support legacy clients.

Neither can remote plugin execution be an all-round solution as many enterprises can't allow exposing `*.proto` files for remote code generation nor/and have external dependencies which cannot be easily mirrored nor official plugins have older protoc/grpc version (see https://github.com/bufbuild/buf/issues/1289). I've also [heard](https://github.com/bufbuild/buf/issues/439) `api.buf.build` may not be available in China.

tonyli233:
Fixes #1462.

Added field `protoc_path` in `buf.gen.yaml` to allow specifying `protoc` path for built-in plugins as mentioned [here](https://github.com/bufbuild/buf/issues/1462).

### DIFF
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -184,6 +184,8 @@ type PluginConfig struct {
 	Path []string
 	// Required
 	Strategy Strategy
+	// Required
+	ProtocPath string
 }
 
 // PluginName returns this PluginConfig's plugin name.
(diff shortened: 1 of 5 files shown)


private/buf/bufgen/bufgen.go
```diff
@@ -184,6 +184,8 @@ type PluginConfig struct {
 	Path []string
 	// Required
 	Strategy Strategy
+	// Required
```
bufdev, line 187: This should not be required

private/buf/bufgen/generator.go
```diff
@@ -295,6 +295,12 @@ func (g *generator) execLocalPlugin(
 	if err != nil {
 		return nil, err
 	}
+	options := []appprotoexec.GenerateOption{
+		appprotoexec.GenerateWithPluginPath(pluginConfig.Path...),
+	}
+	if pluginConfig.ProtocPath != "" {
+		options = append(options, appprotoexec.GenerateWithProtocPath(pluginConfig.ProtocPath))
```
bufdev, line 302: You don't need to protect this with an if, as if it is empty, it is as if the option is not called

bufdev (commented):
Has this been manually QA'ed? I would take two protoc versions that have different output for one of the builtin plugins, and set `protoc_path` to one or the other to make sure it works. Test both absolute and relative paths.

tonyli233:
> Has this been manually QA'ed? I would take two protoc versions that have different output for one of the builtin plugins, and set `protoc_path` to one or the other to make sure it works. Test both absolute and relative paths.

Just manually QA'ed with different builtin plugins and with both absolute and relative paths. Different outputs were generated as expected.

bufdev (approved):

PR #1817, opened 2023-02-09
Enable version negotiation with Docker client

pkwarren:
Update Docker client to negotiate the client version and also fix typo in error handling.

### DIFF
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -188,7 +188,7 @@ func run(
 	if err != nil {
 		return err
 	}
-	client, err := bufplugindocker.NewClient(ctx, container.Logger(), bufcli.Version)
+	client, err := bufplugindocker.NewClient(container.Logger(), bufcli.Version)
 	if err != nil {
 		return err
 	}
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -188,7 +188,7 @@ func run(
 	if err != nil {
 		return err
 	}
-	client, err := bufplugindocker.NewClient(container.Logger(), bufcli.Version)
+	client, err := bufplugindocker.NewClient(ctx, container.Logger(), bufcli.Version)
 	if err != nil {
 		return err
 	}
--- a/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
@@ -199,7 +199,7 @@ func run(
 	if flags.Image != "" {
 		inspectResponse, err := client.Inspect(ctx, flags.Image)
 		if err != nil {
-			return nil
+			return err
 		}
 		imageID = inspectResponse.ImageID
 	} else {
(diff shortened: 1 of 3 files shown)


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -199,7 +199,7 @@ func run(
 	if flags.Image != "" {
 		inspectResponse, err := client.Inspect(ctx, flags.Image)
 		if err != nil {
-			return nil
+			return err
```
pkwarren, line 202: 🙄

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -225,6 +225,11 @@ func NewClient(logger *zap.Logger, cliVersion string, options ...ClientOption) (
 		client.WithHTTPHeaders(map[string]string{
 			"User-Agent": BufUpstreamClientUserAgentPrefix + cliVersion,
 		}),
+		// NOTE: This setting is not safe when a client is called from multiple goroutines:
```
bufdev, line 228: This feels like a massive potential bug in the future - are we sure we want to do this?

private/bufpkg/bufplugin/bufplugindocker/docker.go
```diff
@@ -225,6 +225,11 @@ func NewClient(logger *zap.Logger, cliVersion string, options ...ClientOption) (
...
+		// - https://github.com/moby/moby/issues/43729#issuecomment-1174446640
+		//
+		// Our usage is for the CLI with a single goroutine so we're ok to enable for compatibility.
+		client.WithAPIVersionNegotiation(),
```
mfridman, line 232: Are we sure we want to bake this into the CLI?

mfridman (approved):

private/bufpkg/bufplugin/bufplugindocker/docker.go
rhbuf: right now it just breaks when used against any server that is a lower version, not sure what our other mitigations are, the docker team [are aware of this](https://github.com/moby/moby/issues/43729#issuecomment-1174446640) so will hopefully fix in the near future
```
Error response from daemon: client version 1.42 is too new. Maximum supported API version is 1.41
```

pkwarren: We are only using this to push custom plugins - there are no other code paths that use these Docker abstractions and no concurrency in play there.

The alternative is that we hardcode an earlier version of the Docker API version that is compatible with most systems.

private/bufpkg/bufplugin/bufplugindocker/docker.go
rhbuf: unfortunately it cannot be set from the [env](https://pkg.go.dev/github.com/docker/docker/client#FromEnv) like most other options

mfridman: i guess you could set it manually `DOCKER_API_VERSION=1.x`, but that's obviously not great (nor recommended).

this seems reasonable, I just hope the negotiation isn't buggy.

private/bufpkg/bufplugin/bufplugindocker/docker.go
pkwarren: PTAL - I updated to force version negotiation when creating the client so reads/writes to those fields should no longer race.

private/bufpkg/bufplugin/bufplugindocker/docker.go
pkwarren: The updated code will skip setting the version if you've provided an override via the env var:

https://github.com/moby/moby/blob/d7573ab8672555762688f4c7ab8cc69ae8ec1a47/client/client.go#L227-L244

bufdev (approved):

PR #1818, opened 2023-02-09
Handle walking different volumes for Windows in normalpath

doriable:
This uses a sentinel value in checking the path equality in `normalpath` for Windows.
This lets us handles different volumes in the Windows paths.

This change has been tested on a Windows machine with Go 1.20.

### DIFF
--- a/private/pkg/normalpath/normalpath_windows.go
+++ b/private/pkg/normalpath/normalpath_windows.go
@@ -83,10 +83,17 @@ func EqualsOrContainsPath(value string, path string, pathType PathType) bool {
 
 	// Walk up the path and compare at each directory level until there is a
 	// match or the path reaches its root (either `/` or `.`).
-	for curPath := path; !strings.EqualFold(curPath, pathRoot); curPath = Dir(curPath) {
+	curPath := path
+	var lastSeen string
+	for {
 		if strings.EqualFold(value, curPath) {
 			return true
 		}
+		curPath = Dir(curPath)
+		if lastSeen == curPath {
+			break
+		}
+		lastSeen = curPath
 	}
 	return false
 }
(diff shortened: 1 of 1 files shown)


private/pkg/normalpath/normalpath_windows.go
```diff
@@ -83,10 +83,17 @@ func EqualsOrContainsPath(value string, path string, pathType PathType) bool {
 
 	// Walk up the path and compare at each directory level until there is a
 	// match or the path reaches its root (either `/` or `.`).
-	for curPath := path; !strings.EqualFold(curPath, pathRoot); curPath = Dir(curPath) {
```
jchadwick-buf, line 86: The approach itself seems logical enough to me. However, I think this means you can delete the `pathRoot` calculation above, right? Curious if/how the behavior changes with UNC paths, too. If it works for those, then this seems like a major simplification.

jchadwick-buf (changes requested):
Approach LGTM if it works for UNC paths.

private/pkg/normalpath/normalpath_windows.go
doriable: So I think that this should work with UNC paths, since `Dir` should return the volume name as `VolumeName` returns it: https://cs.opensource.google/go/go/+/master:src/path/filepath/path.go;l=654
So I think you're right, we'll be able to get rid of the `pathRoot` negotiation. Let me test this to make sure!

doriable: Just tested, seems like this works (it's good that we have test cases for this :D)

bufdev:
Can you merge this into the Go 1.20 PR and see if it works when you're done, and if so, merge that PR?

jchadwick-buf (approved):

PR #1821, opened 2023-02-10
Move bufwork directory check outside newConfigV1

rhbuf:
This check fails every time you run `getConfigForBucket` and hit `case 0` in the switch (no work config files found) because the `ExternalConfigV1{}` has no directories configured. Update the only other use case of `newConfigV1` to move the validation outside that function.


pkwarren (approved):

bufdev (changes requested):
This was behaving as intended - need more data as to why this is needed.

rhbuf:
This was a misconception on my part, resolved in https://github.com/bufbuild/buf/pull/1822

Issue #1799, opened 2023-02-07
Include rule name in breaking change output


codygibb:
When a breaking change is identified, buf should output the rule name (e.g. FIELD_NO_DELETE) in addition
to today's human-readable error output.

For example, if I have a message:
```
message Foo {
  string name = 1;
}
```
And I change the field number of `name` to `2`, buf prints:
```
testpackage/foo.proto:7:1:Previously present field "1" with name "name" on message "Foo" was deleted.
```
It would be helpful if buf _also_ printed the violated rule, maybe something like:
```
testpackage/foo.proto:7:1: FIELD_NO_DELETE: Previously present field "1" with name "name" on message "Foo" was deleted
```
This is useful for several reasons:
1. I can instantly identify the rule that failed instead of having to guess. This is especially important for overlapping/ambiguous rules.
2. I can easily find documentation about the rule, why it exists, what I should be doing instead, etc.
3. I can easily disable the rule if I need to via `except`.

codygibb:
Looks like `--error-format=json` is a good workaround for now (although having the rule name in the default output would still be nice):
```
$ buf breaking --against '.git#branch=main' --error-format=json | jq .
{
  "path": "testpackage/foo.proto",
  "start_line": 7,
  "start_column": 1,
  "end_line": 9,
  "end_column": 2,
  "type": "FIELD_NO_DELETE",
  "message": "Previously present field \"1\" with name \"name\" on message \"Foo\" was deleted."
}
```

bufdev:
We've considered this before, but can consider it again. We consiously did not do this in the past as it makes the messages long and somewhat less readable. If we do do this, we need to just add it to the message, as tools rely on the strict `FILE:LINE:COL:MESSAGE` format, so this would be in the form `FILE:LINE_COL:MESSAGE (ID)`, any thoughts there?

tonyli233:
I think the message now is pretty straight forward, and adding the rule name makes it less readable. I am not sure if it a common use case, so can we make it an optional thing? For example, we can add a new flag `with-rule-name`. We print the message in the `FILE:LINE:COL:MESSAGE (ID)` format if the flag is set, and print in the `FILE_LINE:COL:MESSAGE` format when the flag is not set.

bufdev:
I don't think another flag is what we want to do here, all of our relevant commands adjust the format with a single flag, `--error-format`, and we don't want to pollute the flag set more than we need to.

tonyli233:
[...]

PR #1823, opened 2023-02-10
Include rule name in breaking change output

tonyli233:
Proposed by [this issue](https://github.com/bufbuild/buf/issues/1799), I added `PrintOption` type for `PrintFileAnnotations` in `bufanalysis.bufanalysis.go`. Now printing file annotations can include rule name or exclude rule name depending on situation.

### DIFF
--- a/private/buf/cmd/buf/buf_test.go
+++ b/private/buf/cmd/buf/buf_test.go
@@ -405,11 +405,11 @@ func TestFailCheckBreaking1(t *testing.T) {
 		nil,
 		bufcli.ExitCodeFileAnnotation,
 		filepath.FromSlash(`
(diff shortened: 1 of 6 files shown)


joshcarp:
can you add some tests around this?

bufdev (changes requested):
We don't necessarily want to do this - the existing behavior is as intended, need to think about this more.

tonyli233:
> can you add some tests around this?

Yep I just refactored the old tests around this.

tonyli233:
> We don't necessarily want to do this - the existing behavior is as intended, need to think about this more.

Yeah so I'm keeping this PR as a draft until we have more thoughts.

bufdev:
Having considered this more, and re: https://github.com/bufbuild/buf/issues/1799, I think we're going to punt on any changes here for now, so will close this.

PR #1835, opened 2023-02-14
Fix support for unused imports linting in protoc-gen-buf-lint

jhump:
This will address https://github.com/bufbuild/rules_buf/issues/32. The problem is that the code generator request does not include the stored unused dependencies. So, since the check doesn't see that information, it never complains (even if the file does have unused imports, and the compiler that originally produced descriptor set warned about them).

So, just for `protoc-gen-buf-lint`, we introduce a new step to re-compute the unused dependencies and store them in the image proto. This is woven into `bufimage`, because that's the place where we have the information needed to do this. If we did it somewhere else, then we'd just be making it less efficient because we'd have to _re-create_ a resolver, instead of being able to use the one already created to re-parse unrecognized fields in the image.

This basically emulates the checks that the compiler does. The compiler populates the set of used imports while resolving all symbols in the file, during linking. So we basically do that, too, by visiting all references and marking associated imports as used. I've added test files that make use of every possible link I could think of -- field types, method input/output types, extendees, custom options, keys in `Any` messages, etc. They even include public imports. So I think this is pretty robust.

(In fact, I even unearthed an edge case in [these tests](https://github.com/bufbuild/buf/pull/1835/files#diff-677a988d530972605d439fb3b874951605fee6831d05a32a6e9c456630815a41R19-R29) that is currently _not_ correctly handled by the image filtering code (e.g. for `buf build -type ...`): if an `Any` is nested inside another in a custom option. That is non-trivial to solve in the image filtering code, so I just added a TODO for it in this branch.)

Note that the related lint check for whether a file has a syntax statement (which also does not work from the plugin or from the bazel rules_buf) cannot be implemented this way. This is a targeted change, just for reporting unused imports.


jhump:
@bufdev, @robbertvanginkel, whenever one of you gets a chance, you may be interested to see how this works. It's kind a similar to the type filtering if you squint hard enough.

FWIW, I have no idea what's going on with the Windows failure. Superficially, this test looks the same as `TestRunLint1`, which works fine on Windows. One key difference is that in the failing test, the files import each other. But that would be surprising if that didn't work on Windows... I do have a Windows personal laptop. Maybe I can try to repro on that machine and debug it... But if either of you have thoughts on what it could be, I'm all ears!

private/pkg/protoencoding/resolver.go
```diff
@@ -40,50 +40,58 @@ func newResolver(fileDescriptors ...protodescriptor.FileDescriptor) (Resolver, e
...
+		if err := registerDescriptors(types, messageDescriptor); err != nil {
 			return err
 		}
-		if err := addExtensionsToTypes(types, messageDescriptor.Extensions()); err != nil {
+	}
+
+	enumDescriptors := container.Enums()
+	for i, enumsLen := 0, enumDescriptors.Len(); i < enumsLen; i++ {
+		enumDescriptor := enumDescriptors.Get(i)
+		if err := types.RegisterEnum(dynamicpb.NewEnumType(enumDescriptor)); err != nil {
```
robbertvanginkel, line 82: Nice, the issues linked before seem closed as a wontfix but seems like this made it after all: https://go-review.googlesource.com/c/protobuf/+/237219

private/bufpkg/bufimage/import_tracker.go
```diff
@@ -0,0 +1,226 @@
...
+		t.markUsed(file, string(msgType.Descriptor().FullName()))
+		// process Any messages that might be nested inside this one
+		value := msg.Get(valueField).Bytes()
+		nestedMessage := msgType.New()
+		err = proto.UnmarshalOptions{Resolver: t.resolver}.Unmarshal(value, nestedMessage.Interface())
+		if err != nil {
+			// bytes are not valid; skip it
+			return
+		}
+		msg = nestedMessage // fall-through to recurse into this message
```
robbertvanginkel, line 220: I don't completely follow why we recurse based on the message value here. In what case is this required?

This is probably the analog of `findUsedImportsInOptionValue`, but for Any instead of extensions. However, if you find a concrete value of an any or an extension and you can find the file in which its defined, isn't that enough for marking files/types as imported?

robbertvanginkel (commented):
I haven't looked to the bufcurl part carefully, but the bufimageutil and tracker parts mostly make sense besides the one comment on concrete values.

private/bufpkg/bufimage/import_tracker.go
robbertvanginkel: Ohh thinking about this more does this capture the case of
```
import "a.proto";
import "b.proto";
import "c.proto";
string x = 1 [(some.ext) = {[type.googleapi.com/some.Foo]: {[some.other.ext]: true}}];
```
where
```
// a.proto
pkg some;
extends google.protobuf.FieldOptions {
    google.protobuf.Any ext = ....;
}
// b.proto
import "d.proto";
pkg some;
message Foo {
   Dmessage x = 1;
   extensions 10 to 100;
}
// c.proto
pkg some.other;
extends some.Foo {
    bool ext = 50;
}
// d.proto
pkg some;
message Dmessage{}
```
so that c is marked as used?

jhump: Yes, exactly. Also, it's not _just_ extensions that can be referenced this way: an `Any` could be nested inside another `Any`. It's even verified in the test cases:
https://github.com/bufbuild/buf/pull/1835/files#diff-677a988d530972605d439fb3b874951605fee6831d05a32a6e9c456630815a41R19-R29

This is also why I added a TODO to bufimageutil to eventually do the same thing.

rodaine (approved):

private/bufpkg/bufimage/bufimage.go
```diff
@@ -273,6 +273,19 @@ func WithNoReparse() NewImageForProtoOption {
 	}
 }
 
+// WithUnusedImportsComputation instructs NewImageForProto to compute unused imports
+// for the files. These are usually computed by the compiler and stored in the image.
```
bufdev, line 277: Is there room for us to do this in exactly one place throughout buf, without sacrificing performance? Ie could we in theory only do this here, and not do it in the compiler? Likely a dumb question but just asking

jhump: Not realistically. The compiler has to traverse all edges of the graph anyway when resolving elements during linking and interpretation of options. So reporting unused imports is effectively free.

But after compilation is done, there is no way to get at this info other than to re-examine every edge again in the descriptor.

So if we turned it off in the compiler, with the idea of this being the "only place we do it": (1) other users of `protocompile` would no longer get the benefit of this warning and (2) we'd just be adding runtime overhead to compilation (to visit everything in the graph again) and not actually reduce the surface area of code to maintain (since the compiler must still visit everything in the graph to link).

private/bufpkg/bufimage/bufimage.go
```diff
@@ -273,6 +273,19 @@ func WithNoReparse() NewImageForProtoOption {
...
+// But some sources of images may not include this information, so this option can be
+// used to ensure that information is present in the image and accurate.
+//
+// This option is NOT compatible with WithNoReparse: the image must be re-parsed for
```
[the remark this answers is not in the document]
jhump: Done

bufdev (approved):

PR #1840, opened 2023-02-15
Move appexecproto to bufpkg

robbertvanginkel:
The wasm handler will contain some buf specific code, so referring to that from pkg/app isn't allowed according to bandeps.

Move it to bufpkg (and change the name to match the packes there better).


robbertvanginkel:
See https://github.com/bufbuild/buf/tree/robbert/wasm-plugins-in-path for the wasm change that will be stacked on top of this. There's still a few things to do there such as tests for the yaml/generate syntax and modifying the bufwasm package to make more sense for the client (the current interface is copied from the internal repo for the server side).

Without this move of the appexecproto package, we'd end up with a bandeps violation:
```
github.com/bufbuild/buf/private/pkg/app/appproto/appprotoexec cannot depend on github.com/bufbuild/buf/private/bufpkg/bufwasm: Packages in private/bufpkg cannot be depended on by other packages except private/buf packages..
github.com/bufbuild/buf/private/pkg/app/appproto/appprotoexec cannot depend on github.com/bufbuild/buf/private/bufpkg/bufwasm: Packages in private/pkg cannot depend on packages outside of private/pkg..
github.com/bufbuild/buf/private/pkg/app/appproto/appprotoexec cannot depend on github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/wasmplugin/v1: Packages in private/pkg cannot depend on packages outside of private/pkg..
```

bufdev (approved):
This starts solving a cleanup problem of `private/pkg/app` that I wanted to tackle as well, so two for one.

PR #1841, opened 2023-02-15
Add bufmodule.ComputeStats function

jhump:
I need to throw a test case at this. But this at least shows how this can be accomplished (so we could also move this or adapt the code to an alternate input).


private/bufpkg/bufmodule/stats.go
```diff
@@ -0,0 +1,147 @@
...
+			return nil
+		},
+		nil,
+	))
+	var errs error
+	stats := &Stats{}
+	for _, info := range infos {
+		file, err := module.GetModuleFile(ctx, info.Path())
+		if err != nil {
+			errs = multierr.Append(errs, err)
```
mfridman, line 44: Should this continue in the unhappy path?

private/bufpkg/bufmodule/stats.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		func() {
+			defer func() {
+				errs = multierr.Append(errs, file.Close())
+			}()
+			astRoot, err := parser.Parse(info.Path(), file, handler)
+			if err != nil {
+				stats.NumFilesWithSyntaxErrors++
```
mfridman, line 52: Should this return instead of passing a `nil` file node?

private/bufpkg/bufmodule/stats.go
jhump: Oops, definitely not.

private/bufpkg/bufmodule/stats.go
jhump: Updated. In most cases, `parser.Parse` returns an error and a _non-nil_ file node. So, yes, we still want to examine the AST in that case. But I did add a check to see if the AST is nil, to prevent nil-pointer panics.

private/bufpkg/bufmodule/stats.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+	if stats.NumFiles == 0 && errs != nil {
+		return nil, errs
+	}
+	return stats, errs
+}
+
+// Stats represents some statistics/metrics about a module.
+type Stats struct {
```
mfridman, line 77: Should this also include the number of packages?

I found this an interesting (and potentially) useful metric, example:

```
owner     module      files  packages  services  methods  messages  enums
beta      googleapis  3149   305       823       4108     14328     2633
beta      envoy       560    365       52        91       1868      218
bufbuild  buf         41     8         34        166      463       27
bufbuild  eliza       1      1         1         3        6         0
```

jhump: Done

mfridman (approved):
As you said, needs a test to assert counts for a known valid and invalid module, but otherwise this is great!

I think the `Module` interface is a reasonable input. @bufdev may have some thoughts here on the interface...

Took this branch of a spin and got the numbers I'd expect for the [bufbuild/buf](https://buf.build/bufbuild/buf) module:

```
owner         files  packages  services  methods  messages  enums
bufbuild/buf  41     8         34        166      463       27
```

bufdev:
I'm merging this into the message-count branch and will take it from there, thanks for putting this together :-)

PR #1847, opened 2023-02-17
Update cache to use tamper proofing

pkwarren:
Update the buf CLI's cache to take advantage of tamper proofing features (manifest, blobs, and digests) to store files as content addressable storage. When enabled, the new cache directory is stored as:

```
  ~/.cache/buf/v2/module/{remote}/{owner}/{repo}
      blobs/
        {digest[:2]}/
          {digest[2:]} => The contents of the module's blob/manifest.
      commits/
        {commit} => The manifest digest
```


.golangci.yml
```diff
@@ -82,6 +82,10 @@ issues:
         - forbidigo
       # this is the one file we want to allow exec.Cmd functions in
       path: private/pkg/command/runner.go
+    - linters:
+        - forbidigo
+        # we use os.Rename here to rename files in the same directory
+      path: private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```
pkwarren, line 88: I looked at when this rule was added and it was when we attempted to rename files across filesystems. In this case, I'm renaming a file within the same directory where it will ultimately exist, so I think we should relax this check.

private/buf/bufcli/bufcli.go
```diff
@@ -589,19 +599,32 @@ func newModuleReaderAndCreateCacheDirs(
...
-		bufapimodule.NewModuleReader(
-			bufapimodule.NewDownloadServiceClientFactory(clientConfig),
-			moduleReaderOpts...,
-		),
-		bufmodulecache.NewRepositoryServiceClientFactory(clientConfig),
-		cacheModuleReaderOpts...,
+	repositoryClientFactory := bufmodulecache.NewRepositoryServiceClientFactory(clientConfig)
+	delegateReader := bufapimodule.NewModuleReader(
+		bufapimodule.NewDownloadServiceClientFactory(clientConfig),
+		moduleReaderOpts...,
```
pkwarren, line 605: This is the reader that actually talks to the BSR to pull modules.

private/buf/bufcli/bufcli.go
```diff
@@ -589,19 +599,32 @@ func newModuleReaderAndCreateCacheDirs(
...
 	)
+	var moduleReader bufmodule.ModuleReader
+	if tamperProofingEnabled {
+		moduleReader = bufmodulecache.NewCASModuleReader(
+			cacheModuleCASDirPath,
+			delegateReader,
+			repositoryClientFactory,
+			container.Logger(),
+			container.VerbosePrinter(),
```
pkwarren, line 614: If tamper proofing is enabled, we enable the CASModuleReader.

private/bufpkg/bufapimodule/module_reader.go
```diff
@@ -75,7 +75,12 @@ func (m *moduleReader) GetModule(ctx context.Context, modulePin bufmoduleref.Mod
 		if err != nil {
 			return nil, err
 		}
-		return bufmodule.NewModuleForBucket(ctx, bucket, identityAndCommitOpt)
+		return bufmodule.NewModuleForBucket(
+			ctx,
+			bucket,
+			identityAndCommitOpt,
+			bufmodule.ModuleWithManifestAndBlobs(bucket.Manifest(), bucket.Blobs()),
```
pkwarren, line 82: This is important - if we have manifest/blobs, we need to preserve them in the buf module. This allows us to store them to the cache with all the metadata.

unmultimedio:
Great! I'll take a loot at this work. Some questions:

> ```
>   ~/.cache/buf/v1/module (what's current module cache dir structure?)
>     cas/
>       blobs/
>         {digest[:2]}/
>           {digest[2:]}
>       commits/
>         {commit} => The manifest hex digest (a file with just that content, right?)
>       manifests/
>         {digest[:2]}/
>           {digest[2:]} (does these manifest files live also in the blobs dir? or just here?)
> ```

private/bufpkg/bufmodule/bufmodulecache/cache_stats.go
```diff
@@ -0,0 +1,48 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package bufmodulecache
+
+import "sync"
+
+type cacheStats struct {
```
pkwarren, line 19: Broke this out into a separate struct so it could be re-used across implementations.

private/bufpkg/bufmodule/bufmodulecache/repository_deprecation.go
```diff
@@ -0,0 +1,52 @@
...
+	"context"
+	"fmt"
+
+	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
+	registryv1alpha1 "github.com/bufbuild/buf/private/gen/proto/go/buf/alpha/registry/v1alpha1"
+	"github.com/bufbuild/connect-go"
+	"go.uber.org/zap"
+)
+
+func warnIfDeprecated(
```
pkwarren, line 27: Broke this out into a utility method that could be called by both implementations.

private/pkg/manifest/manifest.go
```diff
@@ -52,17 +52,9 @@ type Manifest struct {
 var _ encoding.TextMarshaler = (*Manifest)(nil)
 var _ encoding.TextUnmarshaler = (*Manifest)(nil)
 
-// New creates an empty manifest.
-func New() *Manifest {
```
pkwarren, line 56: Attempting to make this usable with the zero value.

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -110,6 +111,10 @@ type Module interface {
 	//
 	// This may be nil, since older versions of the module would not have this stored.
 	LintConfig() *buflintconfig.Config
+	// Manifest returns the manifest for the module (possibly empty).
+	Manifest() manifest.Manifest
+	// Blobs return the raw data for the module (possibly empty).
+	Blobs() manifest.BlobSet
```
unmultimedio, line 117: I guess we'll need to have some backend PR as well to cover this change, ready to be merged when this one is?

private/bufpkg/bufmodule/bufmodulecache/module_cacher.go
```diff
@@ -25,13 +25,25 @@ import (
 	"go.uber.org/zap"
 )
 
+type moduleCache interface {
```
bufdev, line 28: What is this for? If we need to keep it, it should not be in this file

private/pkg/manifest/storage.go
```diff
@@ -102,9 +102,19 @@ func BucketWithNoExtraBlobsValidation() BucketOption {
 	}
 }
 
+// ReadBucket extends the storage.ReadBucket interface to add an accessor for the Manifest.
```
bufdev, line 105: This also feels like it breaks the abstraction of a bucket by providing direct data access

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -146,7 +151,15 @@ func ModuleWithModuleIdentityAndCommit(moduleIdentity bufmoduleref.ModuleIdentit
 	}
 }
 
-// NewModuleForBucket returns a new Module. It attempts reads dependencies
+// ModuleWithManifestAndBlobs is used to create a Module with a manifest.Manifest and blobs to support tamper proofing.
+func ModuleWithManifestAndBlobs(manifest manifest.Manifest, blobs manifest.BlobSet) ModuleOption {
```
unmultimedio, line 155: Do we want to be an option or a separate constructor? What happens if someone does this?

```go
NewModuleForBucket( // or Proto
  ctx,
  someReadBucketWithFiles,
  ModuleWithManifestAndBlobs(differentManifest, differentBlobSet),
)
```

.golangci.yml
bufdev, line 88: We banned rename for a reason, can we provide justification as to why this case does not hit that reason (add in comments)

bufdev (commented):
Some preliminary comments - some may be off-base, this is just a quick pass and I have less context, but wanted to get it to you ASAP

.golangci.yml
unmultimedio: Nit: you can regex the exact string match, to avoid excluding the whole file.

private/buf/bufcli/bufcli.go
```diff
@@ -154,6 +155,12 @@ var (
 	// These digests are used to make sure that the data written is actually what we expect, and if it is not,
 	// we clear an entry from the cache, i.e. delete the relevant data directory.
 	v1CacheModuleSumRelDirPath = normalpath.Join("v1", "module", "sum")
+	// v1CacheModuleCASRelDirPath is the relative path to the cache directory for content addressable storage.
+	//
+	// Normalized.
+	// This directory replaces the use of v1CacheModuleDataRelDirPath, v1CacheModuleLockRelDirPath, and
+	// v1CacheModuleSumRelDirPath for modules which support tamper proofing.
+	v1CacheModuleCASRelDirPath = normalpath.Join("v1", "module", "cas")
```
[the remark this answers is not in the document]
pkwarren: Yep - I'm not great at naming things - just wanted this to be separate from the other 3 cache dirs.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,244 @@
...
+		return nil, storage.NewErrNotExist(manifestPath)
+	}
+	bucket, err := manifest.NewBucket(
+		*cacheManifest,
+		*blobSet,
+		manifest.BucketWithAllManifestBlobsValidation(),
+		manifest.BucketWithNoExtraBlobsValidation(),
+	)
+	if err != nil {
+		return nil, storage.NewErrNotExist(manifestPath)
```
unmultimedio, line 135: We should log err to know which validation failed

private/buf/bufcli/bufcli.go
```diff
@@ -589,19 +599,32 @@ func newModuleReaderAndCreateCacheDirs(
...
+		bufapimodule.NewDownloadServiceClientFactory(clientConfig),
+		moduleReaderOpts...,
 	)
+	var moduleReader bufmodule.ModuleReader
+	if tamperProofingEnabled {
+		moduleReader = bufmodulecache.NewCASModuleReader(
+			cacheModuleCASDirPath,
+			delegateReader,
+			repositoryClientFactory,
+			container.Logger(),
```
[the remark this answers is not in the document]
pkwarren: I was trying to move things around in order of importance, but I'll update to match the current constructor.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+	return bufmodule.NewModuleForBucket(ctx, bucket, bufmodule.ModuleWithManifestAndBlobs(*cacheManifest, *blobSet))
```
unmultimedio, line 137: This feels a bit redundant, if we have the bucket we shouldn't need the m+b, and viceversa, no need for the bucket, we're passing the m+b.

private/buf/bufcli/bufcli.go
```diff
@@ -589,19 +599,32 @@ func newModuleReaderAndCreateCacheDirs(
...
-		cacheModuleReaderOpts...,
+	repositoryClientFactory := bufmodulecache.NewRepositoryServiceClientFactory(clientConfig)
+	delegateReader := bufapimodule.NewModuleReader(
+		bufapimodule.NewDownloadServiceClientFactory(clientConfig),
+		moduleReaderOpts...,
 	)
+	var moduleReader bufmodule.ModuleReader
+	if tamperProofingEnabled {
+		moduleReader = bufmodulecache.NewCASModuleReader(
+			cacheModuleCASDirPath,
```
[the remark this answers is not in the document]
pkwarren: NewModuleReader takes read/write buckets (instantiated with storageosprovider). In this implementation, I'm relying on POSIX semantics around file renames to simplify the cache code. Overall, I'm not sure the previous code could be safe with any storage abstraction - there would be requirements on the same use of flock on the underlying filesystem.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,244 @@
...
+			// We've already written this blob
+			continue
+		}
+		blob, found := blobs.BlobFor(blobDigest.String())
+		if !found {
+			return fmt.Errorf("blob not found for path=%q, digest=%q", path, blobHexDigest)
+		}
+		if err := c.writeBlob(ctx, blob, blobsParentDir); err != nil {
+			return err
+		}
```
unmultimedio, line 193: We could skip this if the blob's already written

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -110,6 +111,10 @@ type Module interface {
 	//
 	// This may be nil, since older versions of the module would not have this stored.
 	LintConfig() *buflintconfig.Config
+	// Manifest returns the manifest for the module (possibly empty).
+	Manifest() manifest.Manifest
```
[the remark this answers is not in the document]
pkwarren: The manifest wraps the entire module - it isn't a 1-1 with a file within the module. This was the simplest way I could find to provide additional data at the time of caching to the cache layer. I can look again but I think some other interfaces (like bucket / path abstractions) make it harder to return a more specialized type here.

private/bufpkg/bufmodule/bufmodulecache/module_cacher.go
pkwarren: This was to ensure that the `cas_module_cacher` and `module_cacher` didn't diverge. When tamper proofing is enabled, we can remove the `module_cacher_*` entirely and remove this interface.

private/pkg/manifest/storage.go
pkwarren: This adds more metadata to a returned bucket that supports manifest/blobs. It isn't providing direct data access but just a separate way to represent a module.

.golangci.yml
pkwarren: Yep - see https://github.com/bufbuild/buf/pull/1847#discussion_r1110431719.

Given a temp dir, i.e. `~/.cache/buf/v1/module/{remote}/{owner}/{repo}/blobs`, we are creating a temporary file in `~/.cache/buf/v1/module/{remote}/{owner}/{repo}/blobs/{digest[:2]}/` and then renaming it to `~/.cache/buf/v1/module/{remote}/{owner}/{repo}/blobs/{digest[:2]}/digest[2:]`. This won't cross filesystem boundaries since we are creating the temp file in the same directory where it will ultimately live.

private/bufpkg/bufmodule/module.go
```diff
@@ -39,6 +40,8 @@ type module struct {
 	license              string
 	breakingConfig       *bufbreakingconfig.Config
 	lintConfig           *buflintconfig.Config
+	manifest             manifest.Manifest
+	blobs                manifest.BlobSet
```
unmultimedio, line 44: This is another data source. We should calculate these based on the sourceReadBucket, or viceversa, calculate the bucket based on the m+b.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: I thought about removing this, but we at least need the manifest at a higher level to compare the manifest of the cached bucket to the one defined in the `buf.lock` file. Maybe we create a shallow module which has a manifest and no blobs?

unmultimedio: I see no issue in exposing `Manifest` and `Blobs` getter in the module interface, that way you could compare if it's the same with the cached one.

private/bufpkg/bufmodule/module.go
unmultimedio: If modules are immutable once created, I think there's no issue with having the data in both formats (bucket and m+b) to avoid re-calculating every time. I guess my real concern is being able to pass 2 sources at the constructor.

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,97 @@
...
+	modulePin bufmoduleref.ModulePin,
+) (bufmodule.Module, error) {
+	module, err := c.cache.GetModule(ctx, modulePin)
+	if err == nil {
+		c.stats.MarkHit()
+		return module, nil
+	}
+	if !storage.IsNotExist(err) {
+		return nil, err
+	}
```
twilly, line 72: If `c.cache.GetModule` failed because of some strange error, say EIO on read, then falling back to the delegate is fine. This would also make the cache self healing if it's a persistent error that would be corrected if re-written.

It also means you don't need to wrap errors with `storage.NewErrNotExist` everywhere. It should simplify returns:

```go
return nil, storage.NewErrNotExist(fmt.Sprintf("digest mismatch - expected: %q, found: %q", manifestHexDigest, cacheManifestBlob.Digest().Hex()))
```

```go
	f, err := os.Open(manifestPath)
	if err != nil {
		if os.IsNotExist(err) {
			return nil, storage.NewErrNotExist(manifestPath)
		}
		return nil, err
	}
```

Becomes

```go
return nil, fmt.Sprintf("digest mismatch - expected: %q, found: %q", manifestHexDigest, cacheManifestBlob.Digest().Hex())
```

```go
	f, err := os.Open(manifestPath)
	if err != nil {
		return nil, err
	}
```

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,244 @@
...
+			}
+			return nil, err
+		}
+		manifestHexDigest = strings.TrimSpace(string(manifestDigestBytes))
+		if manifestHexDigest == "" {
+			return nil, storage.NewErrNotExist(commitPath)
+		}
+	}
+	manifestPath := normalpath.Join(moduleBasedir, manifestsDir, manifestHexDigest[:2], manifestHexDigest[2:])
+	f, err := os.Open(manifestPath)
```
twilly, line 76: Maybe we talked about this and I forgot, but why is manifest blobs treated special? It should work to store all manifests in `blobsDir`. If we want to be more transparent about the cache on disk, we could have the manifest dir be links or just empty.

(The order of writing needs to be changed to blobs then manifest blob, but that's simple enough.)

pkwarren: There will be no overlap between manifest contents and blob contents (we won't get any space savings if we move them to the same dir), and so storing in a separate directory makes it easier to navigate the cache from disk. Otherwise, this shouldn't matter much at all to the CLI.

private/buf/bufcli/bufcli.go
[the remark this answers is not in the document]
bufdev: We were using a file lock though, right? Let's catch up about this one offline

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
bufdev: Yea the constructor should not take the manifest

private/bufpkg/bufmodule/bufmodule.go
pkwarren: Will we rely on having these available in the backend for anything? We're only using them in the caching logic right now.

bufdev, line 117: nit: if we keep this, this should be named `BlobSet()`

private/buf/bufcli/bufcli.go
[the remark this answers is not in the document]
pkwarren: Updated to use storage.ReadWriteBucket along with a new `PutOption.PutAtomic`.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: Removed wrapping of the errors.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: We don't do this - otherwise the cache couldn't heal itself if one of the files didn't match the digest.

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
pkwarren: Updated to remove the wrapping of errors.

private/bufpkg/bufmodule/bufmodule.go
pkwarren: Done.

private/pkg/storage/storageos/bucket.go
```diff
@@ -184,12 +190,20 @@ func (b *bucket) Put(ctx context.Context, path string, _ ...storage.PutOption) (
 	} else if !fileInfo.IsDir() {
 		return nil, newErrNotDir(externalDir)
 	}
-	file, err := os.Create(externalPath)
+	var file *os.File
+	finalPath := ""
+	if putOptions.Atomic {
+		file, err = os.CreateTemp(externalDir, "."+filepath.Base(externalPath)+"*")
```
bufdev, line 196: this doesn't work with absolute paths

private/pkg/storage/bucket.go
```diff
@@ -73,6 +74,20 @@ func PutWithChunkSize(sizeInBytes int64) PutOption {
...
 
+// PutAtomic ensures that the Put fully writes the file before making it
+// available to readers. This happens by default for some implementations,
+// while others may need to perform a sequence of operations to ensure
+// atomic writes.
+//
+// The Put operation is complete and the path will be readable once the
+// returned WriteObjectCloser is written and closed (without an error).
+// Any errors will cause the Put to be skipped (no path will be created).
+func PutAtomic() PutOption {
```
bufdev, line 85: PutWithAtomic

private/pkg/manifest/manifest.go
twilly: This seems to do the trick. :)

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,94 @@
...
+		return module, nil
+	}
+	c.stats.MarkMiss()
+	module, err = c.delegate.GetModule(ctx, modulePin)
+	if err != nil {
+		return nil, err
+	}
+	if modulePinDigest := modulePin.Digest(); modulePinDigest != "" {
+		manifest := module.Manifest()
+		manifestBlob, err := manifest.Blob()
```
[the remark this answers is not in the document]
pkwarren: This was to work around inconsistencies with pointer and value receivers:

image

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,94 @@
...
+			// buf.lock module digest and BSR module don't match - fail without overwriting cache
+			return nil, fmt.Errorf("module digest mismatch - expected: %q, found: %q", modulePinDigest, manifestDigest)
+		}
+	}
+	if err := c.cache.PutModule(ctx, modulePin, module); err != nil {
+		return nil, err
+	}
+	if err := warnIfDeprecated(ctx, c.repositoryClientFactory, modulePin, c.logger); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
pkwarren: Agreed - the only thing I could think of is avoiding calls to the BSR to check deprecation on every build. This will only do it when we already have to reach out to the BSR to fetch the module.

private/pkg/manifest/storage.go
pkwarren: Removed this.

private/bufpkg/bufmodule/bufmodule.go
pkwarren: Created a separate constructor to create a module from a manifest and blobs.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: Created a new constructor so we no longer have an option for adding a manifest and blobs - we create the module from them.

private/bufpkg/bufmodule/module.go
pkwarren: Fixed.

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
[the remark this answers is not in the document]
pkwarren: Updated to make the module return pointers to manifest and blob set, which makes it easier to write these sort of things.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: Optimized this to first validate if the cache entry matches the digest, and if so we skip the write.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		retErr = multierr.Append(retErr, f.Close())
+	}()
+	cacheManifest, err := manifest.NewFromReader(f)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
pkwarren: Done.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,222 @@
...
+			return nil, err
+		}
+		contents, err := io.ReadAll(f)
+		if err != nil {
+			return nil, err
+		}
+		blob, err := manifest.NewMemoryBlob(*digest, contents, manifest.MemoryBlobWithDigestValidation())
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
pkwarren: Done.

twilly (approved):
+1 :)

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,272 @@
...
+		return nil, err
+	}
+	f, err := blob.Open(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		retErr = multierr.Append(retErr, f.Close())
+	}()
+	moduleManifest, err := manifest.NewFromReader(f)
```
unmultimedio, line 210: Feels weird to be reading the blob twice here. It's an interface issue I guess. Food for thought.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,272 @@
...
+func (c *casModuleCacher) PutModule(
+	ctx context.Context,
+	modulePin bufmoduleref.ModulePin,
+	module bufmodule.Module,
+) (retErr error) {
+	moduleManifest := module.Manifest()
+	manifestBlob, err := moduleManifest.Blob()
+	if err != nil {
+		return err
+	}
```
unmultimedio, line 102: Nit: You documented that `.Manifest` could be nil, you could check that here before asking its `.Blob` to avoid nil panic?

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,272 @@
...
+	digest := manifestBlob.Digest()
+	if digest == nil {
+		return errors.New("empty manifest digest")
+	}
+	if modulePinDigestEncoded := modulePin.Digest(); modulePinDigestEncoded != "" {
+		modulePinDigest, err := manifest.NewDigestFromString(modulePinDigestEncoded)
+		if err != nil {
+			return fmt.Errorf("invalid digest %q: %w", modulePinDigestEncoded, err)
+		}
+		if digest.String() != modulePinDigest.String() {
```
unmultimedio, line 112: Nit:
```suggestion
		if !digest.Equal(modulePinDigest) {
```

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -0,0 +1,272 @@
...
+	}()
+	digester, err := manifest.NewDigester(digest.Type())
+	if err != nil {
+		return false, err
+	}
+	cacheDigest, err := digester.Digest(f)
+	if err != nil {
+		return false, err
+	}
+	return digest.String() == cacheDigest.String(), nil
```
unmultimedio, line 191:
```suggestion
	return digest.Equal(cacheDigest), nil
```

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,99 @@
...
+		return nil, fmt.Errorf("required manifest/blobSet not set on module")
+	}
+	if modulePinDigest := modulePin.Digest(); modulePinDigest != "" {
+		manifestBlob, err := module.Manifest().Blob()
+		if err != nil {
+			return nil, err
+		}
+		manifestDigest := manifestBlob.Digest().String()
+		if manifestDigest != modulePinDigest {
+			// buf.lock module digest and BSR module don't match - fail without overwriting cache
```
unmultimedio, line 88: Nit
```suggestion
			// buf.lock module digest and cached BSR module don't match - fail without overwriting cache
```

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,99 @@
...
+	modulePin bufmoduleref.ModulePin,
+) (bufmodule.Module, error) {
+	module, err := c.cache.GetModule(ctx, modulePin)
+	if err == nil {
+		c.stats.MarkHit()
+		return module, nil
+	}
+	c.logger.Debug("module cache miss", zap.Error(err))
+	c.stats.MarkMiss()
+	module, err = c.delegate.GetModule(ctx, modulePin)
```
unmultimedio, line 72: Nit for readability
```suggestion
	cachedModule, err := c.cache.GetModule(ctx, modulePin)
	if err == nil {
		c.stats.MarkHit()
		return cachedModule, nil
	}
	c.logger.Debug("module cache miss", zap.Error(err))
	c.stats.MarkMiss()
	remoteModule, err := c.delegate.GetModule(ctx, modulePin)
```

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
```diff
@@ -0,0 +1,99 @@
...
+	if module.Manifest() == nil || module.BlobSet() == nil {
+		return nil, fmt.Errorf("required manifest/blobSet not set on module")
+	}
+	if modulePinDigest := modulePin.Digest(); modulePinDigest != "" {
+		manifestBlob, err := module.Manifest().Blob()
+		if err != nil {
+			return nil, err
+		}
+		manifestDigest := manifestBlob.Digest().String()
+		if manifestDigest != modulePinDigest {
```
unmultimedio, line 87: Nit: Use `.Equal`

private/buf/bufcli/bufcli.go
```diff
@@ -549,19 +556,22 @@ func newModuleReaderAndCreateCacheDirs(
 	cacheModuleDataDirPath := normalpath.Join(container.CacheDirPath(), v1CacheModuleDataRelDirPath)
 	cacheModuleLockDirPath := normalpath.Join(container.CacheDirPath(), v1CacheModuleLockRelDirPath)
 	cacheModuleSumDirPath := normalpath.Join(container.CacheDirPath(), v1CacheModuleSumRelDirPath)
+	cacheModuleDirPathV2 := normalpath.Join(container.CacheDirPath(), v2CacheModuleRelDirPath)
 	if err := checkExistingCacheDirs(
```
bufdev, line 560: Just a question - in theory, should a new version of `buf` need to create the v1 cache dirs at all? Will it once tamper proofing is everywhere?

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -110,6 +111,10 @@ type Module interface {
 	//
 	// This may be nil, since older versions of the module would not have this stored.
 	LintConfig() *buflintconfig.Config
+	// Manifest returns the manifest for the module (possibly nil).
```
bufdev, line 114: Can we extend this comment to say what the manifest is for (all files including license, buf.md, etc)? Same for BlobSet, what is in the BlobSet, since BlobSet is a generic concept

private/bufpkg/bufmodule/bufmodulecache/repository_deprecation.go
```diff
@@ -0,0 +1,52 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: Nit: would just make this `util.go`

bufdev (approved):

private/pkg/storage/storageos/bucket.go
```diff
@@ -184,12 +190,20 @@ func (b *bucket) Put(ctx context.Context, path string, _ ...storage.PutOption) (
 	} else if !fileInfo.IsDir() {
 		return nil, newErrNotDir(externalDir)
 	}
-	file, err := os.Create(externalPath)
+	var file *os.File
+	finalPath := ""
```
unmultimedio, line 194: Nit
```suggestion
	var finalPath string
```

unmultimedio (approved):
Looking great. No blocking comments, a few nits, if there are any upcoming PR those could be addressed there if you want.

private/buf/bufcli/bufcli.go
pkwarren: Nope - we'd only need the v2 directories. I can update to create only the necessary directories.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: We ended up doing it this way because we first validate the blob's contents against the digest before attempting to consume it as a manifest. I don't think we are really reading it twice here since the second open just returns an in-memory view of the blob's contents.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
pkwarren: We check it a layer higher in the cas_module_reader, but I added an extra check here.

private/bufpkg/bufmodule/bufmodulecache/cas_module_reader.go
pkwarren: This isn't the cached module - it is the one we've just retrieved from the BSR due to a cache miss.

twilly (approved):
+1

unmultimedio (approved):

PR #1852, opened 2023-02-22
Update descriptor.proto and Image mappings for editions

bufdev:
Fixes https://github.com/bufbuild/buf/issues/1845.

PR #1853, opened 2023-02-22
Handle new descriptor.proto syntax and options

bufdev:
This actually handles `syntax = "editions";` and the new options in `descriptor.proto` everywhere. This is a follow-up to #1852.

Note that `protosource.Locations` were not yet added for the new options, as (1) I don't believe they are handled by protocompile yet, so there will be no value (2) there is precedent in `protosource` for not having `Locations` for every option. If we eventually need those `Locations`, we can easily add them (assuming protocompile produces them).


private/pkg/protosource/file.go
```diff
@@ -244,16 +244,18 @@ func newFile(inputFile InputFile) (*file, error) {
...
 		// IsSyntaxUnspecified
+		f.syntax = SyntaxUnspecified
+	} else {
+		switch syntaxString := f.fileDescriptor.GetSyntax(); syntaxString {
 		case "", "proto2":
 			f.syntax = SyntaxProto2
 		case "proto3":
 			f.syntax = SyntaxProto3
+		case "editions":
+			f.syntax = SyntaxEditions
```
jhump, line 258: We should not allow this. This case is only valid when the  `edition` field is set to a valid edition, and there are no known/valid editions yet.

jhump (commented):
I gotta be honest: I'm skeptical about most of this stuff in this branch.

What is the purpose of the protosource package? It seemed to be wrappers around descriptor protos that provide fully-qualified names and source code info paths. Is it really necessary to repeat _everything_ in the descriptor model in there, all with hand-written code? Will anything actually _use_ the new fields/methods added there?

Also, I really don't think we should pretend to support editions anywhere: it could lead to unexpected problems since we really _don't_ (and can't) support editions.

private/pkg/protosource/alias_types.go
```diff
@@ -80,6 +80,34 @@ const (
 	// FieldOptionsJSTypeJSNumber is an alias.
 	FieldOptionsJSTypeJSNumber FieldOptionsJSType = 2
 
+	// FieldOptionsOptionRetentionUnknown is an alias.
```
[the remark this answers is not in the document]
bufdev: The goal here was to not depend on the types in `descriptorpb` at all in our lint or breaking code, this predated adoption of `google.golang.org/protobuf` in 2019. Adding these is just keeping this package up to date.

private/buf/bufgen/features.go
```diff
@@ -105,8 +105,8 @@ func featureName(feature pluginpb.CodeGeneratorResponse_Feature) string {
 }
 
 func fileHasProto3Optional(fileDescriptorProto *descriptorpb.FileDescriptorProto) bool {
-	if fileDescriptorProto.GetSyntax() != "proto3" {
-		// can't have proto3 optional unless syntax is proto3
+	if fileDescriptorProto.GetSyntax() != "proto3" && fileDescriptorProto.GetSyntax() != "editions" {
+		// can't have proto3 optional unless syntax is proto3 or editions
```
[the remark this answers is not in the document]
bufdev: OK, can revert this part.

private/pkg/protosource/enum.go
```diff
@@ -18,30 +18,33 @@ type enum struct {
...
-	values             []EnumValue
-	allowAlias         bool
-	deprecated         bool
-	allowAliasPath     []int32
-	reservedEnumRanges []EnumRange
-	reservedNames      []ReservedName
-	parent             Message
+	values                             []EnumValue
+	allowAlias                         bool
+	deprecatedLegacyJSONFieldConflicts bool
```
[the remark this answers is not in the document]
bufdev: `protosource` has complete coverage of the builtin options, this just keeps it up to date.

private/pkg/protosource/file.go
bufdev: `protosource` is not a compiler, it assumes it is being given a valid `FileDescriptorSet`. Given that assumption, and that a value of syntax can be `editions`, we should handle it.

bufdev:
Per offline discussion, reverted accepting of editions for proto3 optional, and got rid of all the alias types

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
```diff
@@ -17,69 +17,71 @@ package bufbreakingcheck
...
 		// an enum value, but we have no way of verifying this.
-		protosource.FieldDescriptorProtoTypeEnum: 11,
+		descriptorpb.FieldDescriptorProto_TYPE_ENUM: 11,
 	}
 
 	// https://developers.google.com/protocol-buffers/docs/proto3#json
 	// this is not just JSON-compatible, but also wire-compatible, i.e. the intersection
-	fieldDescriptorProtoTypeToWireJSONCompatiblityGroup = map[protosource.FieldDescriptorProtoType]int{
+	fieldDescriptorProtoTypeToWireJSONCompatiblityGroup = map[descriptorpb.FieldDescriptorProto_Type]int{
 		// fixed32 not compatible for wire so not included
```
jhump, line 65: nit: this comment (and below) suggest that fixed32 and fixed64 are not present, but they are present below.

jhump (approved):
LGTM!

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
bufdev: oh i think this comment meant not in the int types

PR #1855, opened 2023-02-22
Preserve unknown fields when converting between `FileDescriptorProto` and `ImageFile`

jhump:
This way, if a field is ever added to the descriptors and then serialized by a newer version of `buf`, an older version of `buf` won't mangle it and inadvertently drop those newer fields.

The place in `protodescriptor` is the code path used for converting an image file to a descriptor proto. And the place in `bufimage` is for the other direction.

### DIFF
--- a/private/bufpkg/bufimage/bufimagetesting/bufimagetesting_test.go
+++ b/private/bufpkg/bufimage/bufimagetesting/bufimagetesting_test.go
@@ -455,22 +455,19 @@ func TestBasic(t *testing.T) {
 	)
 	diff := cmp.Diff(protoImage, bufimage.ImageToProtoImage(newImage), protocmp.Transform())
 	require.Equal(t, "", diff)
-	// TODO: all of the below proto equality checks should use cmp.Diff
-	require.Equal(
-		t,
-		&descriptorpb.FileDescriptorSet{
-			File: []*descriptorpb.FileDescriptorProto{
-				testProtoImageFileToFileDescriptorProto(protoImageFileAA),
-				testProtoImageFileToFileDescriptorProto(protoImageFileImport),
-				testProtoImageFileToFileDescriptorProto(protoImageFileWellKnownTypeImport),
-				testProtoImageFileToFileDescriptorProto(protoImageFileAB),
-				testProtoImageFileToFileDescriptorProto(protoImageFileBA),
-				testProtoImageFileToFileDescriptorProto(protoImageFileBB),
-				testProtoImageFileToFileDescriptorProto(protoImageFileOutlandishDirectoryName),
-			},
+	fileDescriptorSet := &descriptorpb.FileDescriptorSet{
+		File: []*descriptorpb.FileDescriptorProto{
+			testProtoImageFileToFileDescriptorProto(protoImageFileAA),
+			testProtoImageFileToFileDescriptorProto(protoImageFileImport),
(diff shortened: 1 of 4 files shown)


private/bufpkg/bufimage/bufimagetesting/bufimagetesting_test.go
```diff
@@ -455,22 +455,19 @@ func TestBasic(t *testing.T) {
 	)
 	diff := cmp.Diff(protoImage, bufimage.ImageToProtoImage(newImage), protocmp.Transform())
 	require.Equal(t, "", diff)
-	// TODO: all of the below proto equality checks should use cmp.Diff
```
jhump, line 458: I had to address this TODO because now that the `ProtoReflect` method might be called, the internal, unexported `atomicMessageState` pointer field of the message can differ between the two messages being compared.

private/bufpkg/bufimage/util.go
```diff
@@ -368,6 +368,10 @@ func fileDescriptorProtoToProtoImageFile(
 			ModuleInfo:          protoModuleInfo,
 		},
 	}
+	// TODO: What if file descriptor's unknown fields contain the image file's buf_extension?
+	//       We should probably examine the unknown fields and discard field 8042 if found.
```
jhump, line 372: Curious what you think about this. Should I go ahead and address this now? Do you think this is a material risk?

bufdev: Hmmm. I don't think its a material risk, I purposefully chose a borderline-nonsensical large number to make sure there would not be any overlap. If it's easy to address (<10 min), might be worth doing just to get it done though.

jhump: Ok, I will go ahead and address this.

FWIW, I wasn't worried about a conflicting field being added to `FileDescriptorProto`. The risk I was thinking of was a scenario like so:
1. An `ImageFile` is serialized to a file.
2. That file gets deserialized into a `FileDescriptorProto`. (Like if we tell buf that a file is a file descriptor set when it is actually an image.)
3. The resulting `FileDescriptorProto` is then converted back to an `ImageFile`.

In a case like this, it would be possible for field 8042 to be present _both_ as a known field and an unknown field.

bufdev: Makes sense. Yea if we don't address this now, it's going to be a nasty bug someday.

jhump: Definitely more than 10 minutes to address. But also definitely more robust/correct now. PTAL.

bufdev: should this todo be deleted now?

private/bufpkg/bufimage/util.go
```diff
@@ -368,6 +370,57 @@ func fileDescriptorProtoToProtoImageFile(
...
+	resultFile.ProtoReflect().SetUnknown(stripBufExtensionField(fileDescriptorProto.ProtoReflect().GetUnknown()))
+	return resultFile
+}
+
+const bufExtensionFieldNumber = 8042
```
bufdev, line 379: consts at top of file

bufdev (approved):
approved pending two comments (will likely have to approve again to allow github to merge)

private/bufpkg/bufimage/util.go
jhump: Done

PR #1858, opened 2023-02-24
Fix formatting of C-style block comments that use prefix symbols on each line

### LINKED ISSUE #1839: buf format: comment whitespace is still a little wonky
Hi,

We've noticed that when upgrading buf from 1.8 to 1.9 (as documented in https://github.com/bufbuild/buf/issues/1612), whitespace behavior in multiline comment blocks changed.

It was addressed in 1.10 (per https://github.com/bufbuild/buf/pull/1615) but it looks like there are still persistent issues with multiline comment formatting.

For example, on 1.14, we're seeing this diff introduced in our protobuf:

```diff
 /*
- * Information that is associated with a vulnerability network that we need to be able to store,
- * filter, search, and then obtain for use.
- */
+   * Information that is associated with a vulnerability network that we need to be able to store,
+   * filter, search, and then obtain for use.
+*/
 message Asset {
   /*
-   * Meta information about the asset that can be used to identify and categorize the asset.
-   */
+     * Meta information about the asset that can be used to identify and categorize the asset.
+  */
   message Info {
     string id = 1;
     string name = 2;
     optional string description = 3;
     bool stable = 4;
     optional string context = 5;
     string type = 6;
   }
```

which really just looks....wonky. Here's a side-by-side comparison.

before (buf 1.7):
```protobuf
/*
 * Information that is associated with a vulnerability network that we need to be able to store,
 * filter, search, and then obtain for use.
 */
message Asset {
  /*
   * Meta information about the asset that can be used to identify and categorize the asset.
   */
  message Info {
    string id = 1;
    string name = 2;
    optional string description = 3;
    bool stable = 4;
    optional string context = 5;
    string type = 6;
  }
```

after (buf 1.8):
```protobuf
/*
   * Information that is associated with a vulnerability network that we need to be able to store,
   * filter, search, and then obtain for use.
*/
message Asset {
  /*
     * Meta information about the asset that can be used to identify and categorize the asset.
  */
  message Info {
    string id = 1;
    string name = 2;
    optional string description = 3;
    bool stable = 4;
    optional string context = 5;
    string type = 6;
  }
```

jhump:
Fixes #1839

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -2027,10 +2027,29 @@ func (f *formatter) writeComment(comment string) {
 		// find minimum indent, so we can make all other lines relative to that
 		minIndent := -1 // sentinel that means unset
 		// start at 1 because line at index zero starts with "/*", not whitespace
+		var prefix string
 		for i := 1; i < len(lines); i++ {
 			indent, ok := computeIndent(lines[i])
 			if ok && (minIndent == -1 || indent < minIndent) {
 				minIndent = indent
+			}
+			if i > 1 && len(prefix) == 0 {
+				// no shared prefix
+				continue
+			}
+			line := strings.TrimSpace(lines[i])
+			if line == "*/" {
+				continue
+			}
+			var linePrefix string
+			if len(line) > 0 && isCommentPrefix(line[0]) {
+				linePrefix = line[:1]
+			}
+			if i == 1 {
+				prefix = linePrefix
+			} else if linePrefix != prefix {
+				// they do not share prefix
+				prefix = ""
 			}
 		}
 		if minIndent < 0 {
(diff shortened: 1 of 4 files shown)


private/buf/bufformat/formatter.go
```diff
@@ -2048,9 +2067,45 @@ func (f *formatter) writeComment(comment string) {
...
+
+			// But if there IS a prefix, "|" for example, we'll left-align
+			// the prefix symbol under the asterisk of the comment start
+			// like this:
+
+			/*
+			 | This comment has a prefix before each line.
+			 | Usually the prefix is asterisk, but it's a
+			 | pipe in this example.
+			*/
```
jhump, line 2086: I had considered lining up the asterisk of the trailing `*/` so it is left-aligned underneath the other symbols. _However_, I couldn't even put an example like that in here because `gofmt` would unindent the `*/` by one space.

So this alignment of the `*/` terminator is inspired by `gofmt`. Its rules appear to be that the `*/` is exactly aligned with the `/*` _unless_ each comment line starts with an asterisk, in which case it is aligned exactly one space to the right of the `/*` so that the asterisks all line up.

pkwarren (approved):

pkwarren (approved):

Issue #558, opened 2021-09-15
Consider adding auto-commenting option for lint to add lint ignores


bufdev:
Basically, in addition to `--error-format=config-ignore-yaml`, add something like an `--auto-comment` flag that automatically adds `// buf:lint:ignore` lines. My intuition is that this would naturally require us to be able to consume the AST and rewrite the whole file, i.e. the formatter, but a customer came up with a solution that might just be generic enough (we need to go through edge cases though):

```go
		fileContent := ""
		for lineNumber, l := range fileLines {
			for i < len(lps) && lineNumber+1 == lps[i].StartLine {
				var prefix string
				runes := []rune(l)
				for i, c := range runes {
					if c != ' ' && c != '\t' {
						prefix = string(runes[0:i])
						break
					}
				}
				fileContent += fmt.Sprintf("%s// buf:lint:ignore %s %s\n", prefix, lps[i].RuleType, lps[i].Message)
				i++
			}
			fileContent += l
			fileContent += "\n"
		}

		if err := ioutil.WriteFile(path, []byte(fileContent), 0644); err != nil {
			panic(err)
		}
```

Enrico2:
I'm the said customer, here's the full version:
```go
package main

import (
	"bufio"
	"encoding/json"
	"fmt"
	"io"
	"io/ioutil"
	"os"
	"sort"
)

type lintProblem struct {
	Path      string `json:"path"`
	StartLine int    `json:"start_line"`
	RuleType  string `json:"type"`
	Message   string `json:"message"`
}

func file2lines(filePath string) ([]string, error) {
	f, err := os.Open(filePath)
	if err != nil {
		return nil, err
	}
	defer func(f *os.File) {
		err := f.Close()
		if err != nil {
			panic(err)
		}
	}(f)
	return linesFromReader(f)
}

func linesFromReader(r io.Reader) ([]string, error) {
	var lines []string
	scanner := bufio.NewScanner(r)
	for scanner.Scan() {
		lines = append(lines, scanner.Text())
	}
	if err := scanner.Err(); err != nil {
		return nil, err
	}

	return lines, nil
}

type lintProblems []lintProblem

func (es lintProblems) Len() int {
	return len(es)
}

func (es lintProblems) Less(i, j int) bool {
	return es[i].StartLine < es[j].StartLine
}

func (es lintProblems) Swap(i, j int) {
	es[i], es[j] = es[j], es[i]
}

func main() {
	var lines []string
[...]

PR #1859, opened 2023-02-25
Allow buf lint to start from a given git revision or patch

xiaoyawei:
## What's this PR about

Similar to what was released in [golangci-lint](https://github.com/golangci/golangci-lint/blob/master/CHANGELOG.md#september-2021), this PR allows buf linter to find new issues from a diff.

## Why is it needed

There are times when someone wants to add buf linter for an existing repo, with many existing issues. Some of such changes cannot even be fixed without a breaking change, and hereby has to stay as is. With this feature, people can add linter for an existing repo, which will work to detect new issues in the future.

## Reviewers

@jhump (Looks like you were the most active contributor recently. Not sure whether you will be the right reviewer and feel free to pass this PR to someone else)

### DIFF
--- a/private/buf/cmd/buf/command/lint/lint.go
+++ b/private/buf/cmd/buf/command/lint/lint.go
@@ -15,8 +15,15 @@
 package lint
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"os"
+
+	"github.com/golangci/revgrep"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"github.com/bufbuild/buf/private/buf/bufcli"
 	"github.com/bufbuild/buf/private/buf/buffetch"
@@ -28,8 +35,6 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/command"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
-	"github.com/spf13/cobra"
-	"github.com/spf13/pflag"
 )
 
 const (
@@ -38,6 +43,10 @@ const (
 	pathsFlagName           = "path"
 	excludePathsFlagName    = "exclude-path"
 	disableSymlinksFlagName = "disable-symlinks"
+	newFlagName             = "new"
+	newFromRevFlagName      = "new-from-rev"
+	newFromPatchFlagName    = "new-from-patch"
+	wholeFilesFlagName      = "whole-files"
 )
 
 // NewCommand returns a new Command.
@@ -62,11 +71,16 @@ func NewCommand(
 }
 
 type flags struct {
-	ErrorFormat     string
-	Config          string
-	Paths           []string
-	ExcludePaths    []string
-	DisableSymlinks bool
+	ErrorFormat       string
+	Config            string
+	Paths             []string
+	ExcludePaths      []string
+	DisableSymlinks   bool
+	DiffFromRevision  string
+	DiffPatchFilePath string
+	OnlyNew           bool
+	WholeFiles        bool
+
 	// special
 	InputHashtag string
 }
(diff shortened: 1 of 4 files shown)


private/buf/cmd/buf/command/lint/lint.go
```diff
@@ -175,3 +228,38 @@ func run(
...
+
+	var patchReader io.Reader
+	if diffPatchFilePath != "" {
+		patch, err := os.ReadFile(diffPatchFilePath)
+		if err != nil {
+			return nil, fmt.Errorf("can't read from patch file %s: %s", diffPatchFilePath, err)
+		}
+		patchReader = bytes.NewReader(patch)
+	}
+	checker := revgrep.Checker{
```
bufdev, line 249: `revgrep` seems to only work with a local directory, whereas `buf` works with local directories (git or not git), tarballs, zip files, remote git repositories, and images. I don't think `revgrep` works as a general solution for `buf` here unless I'm missing something

bufdev (changes requested):
We appreciate the efforts here! This does look like something that would be useful if it could work. Unfortunately I'm concerned that it won't work in `buf`'s case - this seems to rely on `buf` only processing local git repositories, whereas `buf` may be processing directories, tarballs, zip files, remote git repositories, or images. We'd also have to make sure the UX is carefully thought through.

The traditional way to accomplish what you'd like to with `buf` (ignore existing errors) is to do a run with `buf lint --error-format=config-ignore-yaml`, copying and pasting the result into your `buf.yaml`. Does this not work for you/what is insufficient here?

I've made some comments on the PR, but I think we're unlikely to accept it in its current form. Again, thanks for the efforts though.

xiaoyawei:
> The traditional way to accomplish what you'd like to with buf (ignore existing errors) is to do a run with buf lint --error-format=config-ignore-yaml, copying and pasting the result into your buf.yaml. Does this not work for you/what is insufficient here?

@bufdev thanks for the suggestion but I am afraid it is not sufficient. If I did what you suggested, then I can't prevent people from making bad changes in an existing file; instead, for new changes that break linting rules in an existing file, I would like `buf lint` can catch them as well.

What I was trying to achieve is very similar to that in `golangci-lint` (https://github.com/golangci/golangci-lint/blob/f4fec8c6d0700e6a3e20835d04995ea5c9ee9087/pkg/commands/run.go#L232-L244)

bufdev:
In a given file, they could have more bad changes for a the single lint rule that is ignore yes, but that seems like mostly an edge case right?

You're welcome to pursue this PR with (1) no dependencies on revgrep (2) making sure it fits idiomatically into the existing codebase (3) it accounts for all the possible [inputs](https://docs.buf.build/reference/inputs), it will be a significant change though, and we can't promise acceptance.

xiaoyawei:
@bufdev thanks for the feedback

> In a given file, they could have more bad changes for a the single lint rule that is ignore yes, but that seems like mostly an edge case right?

It can't be an edge case for sure. For example, in a file consisting of multiple RPC definitions, some RPC names could be badly formatted; new changes on that file could also possibly suffer from similar issues, especially given the fact that junior engineers tend to follow the existing naming format in the same file, no matter whether that's appropriate or not.

> You're welcome to pursue this PR with (1) no dependencies on revgrep (2) making sure it fits idiomatically into the existing codebase (3) it accounts for all the possible [inputs](https://docs.buf.build/reference/inputs), it will be a significant change though, and we can't promise acceptance.

I would suggest that we could probably this change for `git` only, and handle all edge case (e.g., this feature won't work otherwise `buf lint` is running on a local git repo). Down the road, this change is to solve issues in CI, where 1) 90+% of people use `git` for version management and 2) CI usually runs on a git repo.

It's fine that this change does not get accepted and I totally understand all the concerns from the perspective of the buf owners. I would probably use a forked version with my changes here in my team, which I think is allowed by the license (Apache-2.0) used by this project, if it takes too much efforts to get this PR through to the codebase.

However, I would like to ask you to give it a second thought, that this feature is a super useful one for CI, and common seen in many existing (and widely-used) linters.

bufdev:
For cases where you want to make sure it's just ignoring one specific area, you can use comment ignores https://docs.buf.build/configuration/v1/buf-yaml#allow_comment_ignores. We have an issue tracked in https://github.com/bufbuild/buf/issues/558 to auto-add comment ignores on a first pass, along with code that actually works for it - we'd be more likely to pursue that issue than implement this here.

We're not comfortable having divergent behavior just for `git` - that would be significantly problematic within this codebase. You're totally welcome to maintain your fork obviously for your team :-) I'm going to close this as it sounds like the (large amount of) work to get this working across all inputs isn't something you'd want to take on, which is totally reasonable.

PR #1864, opened 2023-02-27
Add user reactivated type to events proto

seankimdev:
- Required for logging users activated by SCIM


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -310,6 +312,8 @@ message PayloadPluginDeleted {
 
 message PayloadUserCreated {}
 
+message PayloadUserActivated {}
```
saquibmian, line 315: Perhaps this should be `UserReactivated`?

seankimdev: hm yeah I agree that would be more accurate. Let me change

saquibmian (approved):

PR #1869, opened 2023-02-28
Pin Token Exchange Workflow Version

rubensf:
This reduces our security risk


joshcarp (changes requested):
blocking til we've finished release

elliotmjackson (approved):

joshcarp (approved):

PR #1870, opened 2023-02-28
Trigger adjacent jobs in release process

elliotmjackson:
We had chained a number of jobs meaning that if one failed, they all failed. This PR changes that behaviour by having the following steps depend on post-release and then run independently.
- trigger npm publish
- trigger homebrew-buf release
- trigger vim-buf release
- trigger docs.buf.build release
- trigger buf-setup-action release

i achieved this by posting the `VERSION` as an output of the `post-release` job like [this](https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -200,7 +200,7 @@ jobs:
       - id: output_version
         name: Output version to adjacent flows
         run: |
-          VERSION="${{github.ref_name}}".replace("v", "")
+          VERSION="${{github.ref_name:1}}"
           echo "version=${VERSION}" >> "$GITHUB_OUTPUT"
   trigger-npm-publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


joshcarp (approved):

elliotmjackson:
[simple test of the scripting](https://github.com/elliotmjackson/buf-workspace-example/actions/runs/4296709832)

.github/workflows/release.yaml
```diff
@@ -195,6 +197,18 @@ jobs:
...
           body: Release complete for ${{github.ref_name}}
           token: ${{ steps.generate_token.outputs.token }}
+      - name: Get Version
+        uses: actions/github-script@v6
+        with:
+          script:
+            core.exportVariable('VERSION', "${{github.ref_name}}".replace("v", ""));
+      - id: output_version
+        name: Output version to adjacent flows
+        run: echo "version=${{env.VERSION}}" >> "$GITHUB_OUTPUT"
```
paul-sachs, line 207:
```suggestion
            core.setOutput('version', "${{github.ref_name}}".replace("v", ""));
```

paul-sachs (approved):

PR #1873, opened 2023-02-28
Add 'buf beta mod checksum' command

pkwarren:
Update the CLI to support a 'buf beta mod checksum' command, which will show the manifest digest for the Buf module. When run with '--debug', this will also log all entries in the manifest and their digests.


pkwarren:
Will reopen this if needed in the future.

PR #1874, opened 2023-02-28
Update 'buf mod update' to check digest changes

pkwarren:
If 'buf mod update' runs and resolves to the same module identity and commit, it should error out if the digest doesn't match what is recorded in the buf.lock file. While unlikely, this may indicate that a module has changed since last resolved against the BSR.

### DIFF
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -163,11 +163,15 @@ func run(
 		}
 		container.Logger().Warn(warnMsg)
 	}
-
-	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
+	// Before updating buf.lock file, verify that existing dependency digests didn't change for the same commit.
+	err = bufmoduleref.ValidateModulePinsConsistentDigests(ctx, readWriteBucket, dependencyModulePins)
+	if err != nil {
 		if bufmoduleref.IsDigestChanged(err) {
 			return err
 		}
+		return bufcli.NewInternalError(err)
+	}
+	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
 		return bufcli.NewInternalError(err)
 	}
 	return nil
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -16,7 +16,6 @@ package modupdate
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"strings"
 
@@ -166,8 +165,7 @@ func run(
 	}
 
 	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
-		var inconsistentDigestErr *bufmoduleref.DigestChangedError
-		if errors.As(err, &inconsistentDigestErr) {
+		if bufmoduleref.IsDigestChanged(err) {
 			return err
 		}
 		return bufcli.NewInternalError(err)
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -16,6 +16,7 @@ package modupdate
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -165,6 +166,10 @@ func run(
 	}
 
 	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
+		var inconsistentDigestErr bufmoduleref.InconsistentDigestError
+		if errors.As(err, &inconsistentDigestErr) {
+			return err
+		}
 		return bufcli.NewInternalError(err)
 	}
 	return nil
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -164,8 +164,7 @@ func run(
 		container.Logger().Warn(warnMsg)
 	}
 	// Before updating buf.lock file, verify that existing dependency digests didn't change for the same commit.
-	err = bufmoduleref.ValidateModulePinsConsistentDigests(ctx, readWriteBucket, dependencyModulePins)
-	if err != nil {
+	if err := bufmoduleref.ValidateModulePinsConsistentDigests(ctx, readWriteBucket, dependencyModulePins); err != nil {
 		if bufmoduleref.IsDigestChanged(err) {
 			return err
 		}
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -166,7 +166,7 @@ func run(
 	}
 
 	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
-		var inconsistentDigestErr bufmoduleref.InconsistentDigestError
+		var inconsistentDigestErr *bufmoduleref.DigestChangedError
 		if errors.As(err, &inconsistentDigestErr) {
 			return err
 		}
--- a/private/bufpkg/bufmodule/bufmodule.go
+++ b/private/bufpkg/bufmodule/bufmodule.go
@@ -446,28 +446,28 @@ func ModuleDigestB3(ctx context.Context, module Module) (string, error) {
 func ModuleToBucket(
 	ctx context.Context,
 	module Module,
-	readWriteBucket storage.ReadWriteBucket,
+	writeBucket storage.WriteBucket,
 ) error {
 	fileInfos, err := module.SourceFileInfos(ctx)
 	if err != nil {
 		return err
 	}
 	for _, fileInfo := range fileInfos {
-		if err := putModuleFileToBucket(ctx, module, fileInfo.Path(), readWriteBucket); err != nil {
+		if err := putModuleFileToBucket(ctx, module, fileInfo.Path(), writeBucket); err != nil {
 			return err
 		}
 	}
 	if docs := module.Documentation(); docs != "" {
-		if err := storage.PutPath(ctx, readWriteBucket, DocumentationFilePath, []byte(docs)); err != nil {
+		if err := storage.PutPath(ctx, writeBucket, DocumentationFilePath, []byte(docs)); err != nil {
 			return err
 		}
 	}
 	if license := module.License(); license != "" {
-		if err := storage.PutPath(ctx, readWriteBucket, LicenseFilePath, []byte(license)); err != nil {
+		if err := storage.PutPath(ctx, writeBucket, LicenseFilePath, []byte(license)); err != nil {
 			return err
 		}
 	}
-	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, module.DependencyModulePins()); err != nil {
+	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, writeBucket, module.DependencyModulePins()); err != nil {
 		return err
 	}
 	// This is the default version created by bufconfig getters. The versions should be the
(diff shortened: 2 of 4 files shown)


private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -165,6 +166,10 @@ func run(
 	}
 
 	if err := bufmoduleref.PutDependencyModulePinsToBucket(ctx, readWriteBucket, dependencyModulePins); err != nil {
+		var inconsistentDigestErr bufmoduleref.InconsistentDigestError
+		if errors.As(err, &inconsistentDigestErr) {
+			return err
```
pkwarren, line 171: The mechanics of this can change, but what I'm trying to avoid here is printing out an error message that indicates it is an internal error that needs to be reported to Buf's issues page for a well known error.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -364,6 +371,46 @@ func ValidateModulePinsUniqueByIdentity(modulePins []ModulePin) error {
...
 
+// ValidateModulePinsConsistentDigests verifies that module pins to the same commit don't change digests.
+// This is important to avoid MITM issues, where the module digest stored in a buf.lock file doesn't match
+// the module pin returned from the BSR.
+func ValidateModulePinsConsistentDigests(
+	ctx context.Context,
+	bucket storage.ReadWriteBucket,
+	modulePins []ModulePin,
+) error {
+	currentConfig, err := buflock.ReadConfig(ctx, bucket)
```
pkwarren, line 382: This is why we've changed the interface to return a `ReadWriteBucket` instead of a `WriteBucket`. Before updating the `buf.lock` file, we want to verify the current contents vs. the new contents.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -69,6 +69,13 @@ type FileInfo interface {
 	isFileInfo()
 }
 
+// InconsistentDigestError is an interface usable with errors.Is to determine
+// if module pin digests have changed unexpectedly.
+type InconsistentDigestError interface {
+	Error() string
+	inconsistentDigest()
```
twilly, line 76: Unexported field with an exported interface. Why restrict who can instantiate this type of error?

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -364,6 +371,46 @@ func ValidateModulePinsUniqueByIdentity(modulePins []ModulePin) error {
...
+		return err
+	}
+	if len(currentConfig.Dependencies) == 0 {
+		return nil
+	}
+	currentIdentityAndCommitToDigest := make(map[string]string, len(currentConfig.Dependencies))
+	for _, dep := range currentConfig.Dependencies {
+		key := fmt.Sprintf("%s/%s/%s:%s", dep.Remote, dep.Owner, dep.Repository, dep.Commit)
+		currentIdentityAndCommitToDigest[key] = dep.Digest
+	}
```
twilly, line 396: A `ModulePin` is also a `ModuleIdentity`. `dep.IdentityString()` would simplify the map key.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
twilly: Maybe take a `bucket storage.ReadBucket` instead? Write isn't needed in this particular function.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: Just doesn't make sense to raise this elsewhere. I was trying to do the equivalent of a `var ErrInconsistentDigest` but allow a custom error message.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: Yep - I thought it would make it easier to read w/ the symmetry between two different types, but this works.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -364,6 +383,54 @@ func ValidateModulePinsUniqueByIdentity(modulePins []ModulePin) error {
...
+	for _, dep := range currentConfig.Dependencies {
+		// Ignore dependencies with no digest
+		if dep.Digest == "" {
+			continue
+		}
+		// Ignore dependencies with an invalid digest.
+		// We want to replace these with a valid digest.
+		if _, err := manifest.NewDigestFromString(dep.Digest); err != nil {
+			continue
+		}
```
unmultimedio, line 415: Worth to log a warning about replacing an invalid digest?

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -364,6 +383,54 @@ func ValidateModulePinsUniqueByIdentity(modulePins []ModulePin) error {
...
+		key := fmt.Sprintf("%s/%s/%s:%s", dep.Remote, dep.Owner, dep.Repository, dep.Commit)
```
unmultimedio, line 416: I think either the `ModulePin` or the `ModuleReference` or both have a standard `String` method that outputs exactly this?

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -364,6 +383,54 @@ func ValidateModulePinsUniqueByIdentity(modulePins []ModulePin) error {
...
+		currentIdentityAndCommitToDigest[key] = dep.Digest
+	}
+	for _, pin := range modulePins {
+		if pin.Digest() == "" {
+			continue
+		}
+		key := fmt.Sprintf("%s:%s", pin.IdentityString(), pin.Commit())
+		if currentDigest, ok := currentIdentityAndCommitToDigest[key]; ok && currentDigest != pin.Digest() {
+			return &DigestChangedError{
```
unmultimedio, line 425: Nit: Would it be worth to group them in `multierr`? So all errors go out at once? Not sure how that would impact the caller ability to do `errors.As`

unmultimedio (approved):
Just nits, non blocking, nice work!

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -69,6 +70,24 @@ type FileInfo interface {
...
+type DigestChangedError struct {
+	// CurrentDigest is the digest found in the buf.lock file.
+	CurrentDigest string
+	// UpdatedPin is a module pin with a different digest than CurrentDigest for the same commit.
+	UpdatedPin ModulePin
+}
+
+func (e *DigestChangedError) Error() string {
+	return fmt.Sprintf(
+		"module %s commit %q digest changed unexpectedly: expected=%q, found=%q",
```
unmultimedio, line 83: It works nice!

```
➜  moneyapis git:(main) buflon mod update
Failure: module bufbuild.internal/jfigueroa/coinbase commit "574005a7f6404734860b372d4c7667e3" digest changed unexpectedly: expected="shake256:6c661888789858cb84eeac84bb86f958e9c14f6fbdb4ea38b94fc0a2e4f04d69b45da2981d204d1baf54f8e4257a7661e8a168e04e15e2d0ca03fb49425bbe5b", found="shake256:02468788b6c17d1824c6c9759b6f5f057fec736f1fb69e13d8802ca09e5c7935a70d3d8953cade9601d4b154318172d6327e6095d342faa7e4ec1089c04476f2"
```

Since no change is being done now to see in git diff what's the issue, what about this tweak?

```suggestion
		"module %s remote commit %q returned an unexpected digest: local buf.lock=%q, returned=%q",
```

Or something in those lines...

unmultimedio (approved):

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: Changed this implementation to be more sane.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: We could but I don't know what value it would have. I guess we could conceivably have really old lock files that still have a b3 digest in them, and replacing it shouldn't trigger warnings.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: It was ModulePin - updated.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: Updated - thanks!

unmultimedio (approved):

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -446,28 +446,28 @@ func ModuleDigestB3(ctx context.Context, module Module) (string, error) {
 func ModuleToBucket(
 	ctx context.Context,
 	module Module,
-	writeBucket storage.WriteBucket,
+	readWriteBucket storage.ReadWriteBucket,
```
bufdev, line 449: Why does this need a `ReadWriteBucket`? `ModuleToBucket` should only need to write

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -69,6 +71,24 @@ type FileInfo interface {
 	isFileInfo()
 }
 
+// DigestChangedError is returned if module pin digests have changed unexpectedly.
+type DigestChangedError struct {
```
bufdev, line 75: What's the benefit of doing this public struct style over the `New.../Is...` style that already exists in `storage` and `normalpath`?

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -69,6 +71,24 @@ type FileInfo interface {
...
+	// CurrentDigest is the digest found in the buf.lock file.
+	CurrentDigest string
```
bufdev, line 77: Do we need to use these outside of this package/outside of the error message?

private/bufpkg/bufmodule/bufmodule.go
pkwarren: This is because `PutDependencyModulePinsToBucket` calls the new `ValidateModulePinsConsistentDigests` to read the current `buf.lock` file, if present, before overwriting it.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: No real benefit - lemme switch to use that convention.

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
pkwarren: Nope - I just used them in the tests but arguably that's testing internal type behavior. I'll unexport.

twilly (approved):
LGTM :)

private/bufpkg/bufmodule/bufmodule.go
bufdev: This feels a bit off IMO - the purpose of this function is to write a module to a bucket, but it seems like it has hidden behavior that says "if that bucket already has a buf.lock, do something" - that would be surprising to me as a caller of this function. What is it verifying, is it saying "if the `module` variable does not match a pre-existing file from `buflock.ReadConfig` then error"? This feels like a separate function that should be called outside of `ModuleToBucket`

pkwarren: Do you have a suggestion on where this would be good to add? This is a bit of an edge case but we do want to intercept before writing a buf.lock file so we don't overwrite an existing digest for a given commit with a new one, which could be a security concern.

bufdev: It feels like you should do the validate call before calling ModuleToBucket, wherever you are calling ModuleToBucket. The only non-test place this is called is in module_cacher.go, so this should be easy

pkwarren: Updated to reduce the scope of the change to only the `buf mod update` command.

bufdev (approved):

PR #1876, opened 2023-03-02
Add `SCIMTokenService`

saquibmian:
This service enables the creation, deletion, and retrieval of a new type of token, SCIM tokens. These tokens can be used to authenticate and authorize with the BSR's SCIM 2.0 API.


proto/buf/alpha/registry/v1alpha1/scim_token.proto
```diff
@@ -0,0 +1,69 @@
...
+  // DeleteToken deletes an existing token.
+  //
+  // This method requires authentication.
+  rpc DeleteSCIMToken(DeleteSCIMTokenRequest) returns (DeleteSCIMTokenResponse);
+}
+
+message CreateSCIMTokenRequest {
+  // The time until which the token should be valid.
+  // Must be in the future. May be null for no expiry.
+  google.protobuf.Timestamp expire_time = 2;
```
cyinma, line 44: should this start from 1?

saquibmian: Ah yes, let me fix.

proto/buf/alpha/registry/v1alpha1/scim_token.proto
```diff
@@ -0,0 +1,69 @@
...
+
+message SCIMToken {
+  string id = 1;
+  google.protobuf.Timestamp create_time = 2;
+  google.protobuf.Timestamp expire_time = 3;
+}
+
+// TokenSCIMService is the SCIM Token service.
+service SCIMTokenService {
+  // CreateToken creates a new token suitable for authentication to the SCIM API.
```
seankimdev, line 29:
```suggestion
  // CreateToken creates a new token suitable for authentication to the SCIM API.
  //
  // This method requires authentication.
```
nit: assuming this requires authentication as well, but looks like it's copied over from the existing token.proto.

cyinma (approved):
lgtm!

seankimdev (approved):
Looks good to me!

seankimdev (approved):

PR #1879, opened 2023-03-02
Fix panic in image filtering

jhump:
🤦 Caught this while on-call. It happened in production when the BSR was handling an `ImageService.GetImage` request. I think this means use of `-type` param with `buf build` and `buf generate` will crash if the schema contains maps whose values are not message types 😢. So we'll probably want to create a patch release of the Buf CLI, too.

Fixes BSR-1543

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
+++ b/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
@@ -737,15 +737,17 @@ func (t *transitiveClosure) exploreOptionValueForAny(
 	opts *imageFilterOptions,
 ) error {
 	switch {
-	case fd.IsMap() && isMessageKind(fd.MapValue().Kind()):
-		var err error
-		val.Map().Range(func(_ protoreflect.MapKey, v protoreflect.Value) bool {
-			if err = t.exploreOptionScalarValueForAny(v.Message(), referrerFile, imageIndex, opts); err != nil {
-				return false
-			}
-			return true
-		})
-		return err
+	case fd.IsMap():
+		if isMessageKind(fd.MapValue().Kind()) {
+			var err error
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -737,15 +737,17 @@ func (t *transitiveClosure) exploreOptionValueForAny(
...
-	case fd.IsMap() && isMessageKind(fd.MapValue().Kind()):
-		var err error
-		val.Map().Range(func(_ protoreflect.MapKey, v protoreflect.Value) bool {
-			if err = t.exploreOptionScalarValueForAny(v.Message(), referrerFile, imageIndex, opts); err != nil {
-				return false
-			}
-			return true
-		})
-		return err
+	case fd.IsMap():
```
jhump, line 740: For map fields whose value was _not_ a map, it would previously fall-through and end up getting to line 760, where it would blow up with this:
```
type mismatch: cannot convert map to message
```
The value type is actually a map. But we end up calling the `Message()` method. Oops.

So now, if it's a map whose value is _not_ a message, it will no longer fall-through due to the condition change here.

unmultimedio (approved):
lgtm, although not sure if you want another review from someone more familiar with `FileDescriptor`s than me? 😅

jhump:
> not sure if you want another review from someone more familiar with FileDescriptors than me?

LOL, Robbert was my go to reviewer for this stuff before this week.

Maybe @bufdev or @srikrsna-buf?

bufdev (approved):

jhump:
@bufdev, I merged but just realized that this might be changelog-worthy since it fixes a crash. WDYT? If you agree, I'll open another PR w/ the changelog note. And then we should probably cut a point release. (The only other change to the CLI since v1.15.0 is #1874 which doesn't seem minor-version-worthy.)

PR #1884, opened 2023-03-05
Fix broken link for formatting

Brainisthekey:


bufdev:
There's a different link we want to use, we're going to look into this, thanks

PR #1892, opened 2023-03-06
Add Local WASM Generation and Configuration

twoism:
**Experimental Use Only**

This branch adds a WASM execution path to `buf generate` so that we can continue fleshing out this feature.

For now, the `plugin` key in a `buf.gen.yaml` is able to be pointed at a pre-compiled plugin WASM blob.

**Example:**

```
plugins:
  - plugin: wasmplugins/protocolbuffers_go@v1.28.1.wasm
    out: gen
    opt: paths=source_relative
```


go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	github.com/spf13/cobra v1.6.1
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.8.2
+	github.com/tetratelabs/wazero v1.0.0-rc.1
```
bufdev, line 25: Can we do a before and after comparison of the binary size?

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+// executor. The default is 8192 pages for 512MB.
+func WithMemoryLimitPages(memoryLimitPages uint32) PluginExecutorOption {
+	return func(pluginExecutor *PluginExecutor) {
+		pluginExecutor.runtimeConfig = pluginExecutor.runtimeConfig.WithMemoryLimitPages(memoryLimitPages)
+	}
+}
+
+// PluginExecutor wraps a wazero end exposes functions to compile and run wasm
+// plugins.
+type PluginExecutor struct {
```
bufdev, line 102: The rest of the codebase uses interfaces for types such as this

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+	runtimeConfig wazero.RuntimeConfig
+}
+
+// NewPluginExecutor creates a new PluginExecutor with a compilation cache dir
+// and other buf defaults.
+func NewPluginExecutor(_ context.Context, compilationCacheDir string, options ...PluginExecutorOption) (*PluginExecutor, error) {
```
bufdev, line 108: We're not using the context, why is it part of the constructor? I can get compatibility here, but our codebase is all private, and the rest of the codebase does not have this style.

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+	if compilationCacheDir == "" {
+		cache = wazero.NewCompilationCache()
+	} else {
+		var err error
+		cache, err = wazero.NewCompilationCacheWithDir(compilationCacheDir)
+		if err != nil {
+			return nil, err
+		}
+	}
+	const maxMemoryBytes = 1 << 29 // 512MB
```
bufdev, line 119: Consts should be at top of file, per convention here

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+				Stderr:   strings.ToValidUTF8(stderr.String(), ""),
+				Exitcode: exitErr.ExitCode(),
+			}
+		}
+		return err
+	}
+	return nil
+}
+
+type PluginExecutionError struct {
```
bufdev, line 248: Public types need comments

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+
+syntax = "proto3";
+
+package buf.alpha.wasmplugin.v1;
+
+// Metadata describes buf specific extensions for a wasm plugin.  A metadata may
+// be encoded in a custom WASM section named ".bufplugin", see
+// (https://webassembly.github.io/spec/core/binary/modules.html#binary-customsec)
+// for more info.
+message Metadata {
```
bufdev, line 23: Protobuf types should never be named `*data` - all Protobuf types are data

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+  // When no ABI is provided, buf will make a best effort guess based on
+  // the functions the wasm module exports.
+  WasmABI abi = 1;
```
bufdev, line 26: `wasm_abi`

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package buf.alpha.wasmplugin.v1;
```
bufdev, line 17: How are the types in this file used? Document

private/buf/cmd/buf/command/alpha/protoc/plugin.go
```diff
@@ -46,6 +47,7 @@ func executePlugin(
 	logger *zap.Logger,
 	storageosProvider storageos.Provider,
 	runner command.Runner,
+	wasmPluginExecutor *bufwasm.PluginExecutor,
```
bufdev, line 50: interface, see comment below

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -390,10 +392,15 @@ func run(
 			return err
 		}
 	}
+	wasmPluginExecutor, err := bufwasm.NewPluginExecutor(ctx, filepath.Join(container.CacheDirPath(), "wasmplugins"))
```
bufdev, line 395: We have a top-level construct that handles the creation of all cache dirs, I can't remember where, @pkwarren will know

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+// extensions.
+const CustomSectionName = ".bufplugin"
+
+// https://www.w3.org/TR/2019/REC-wasm-core-1-20191205/#sections%E2%91%A0
+const customSectionID = 0
+
+// EncodeBufSection encodes the pluginmetadata message as a custom wasm section.
+// The resulting bytes can be appended to any valid wasm file to add the new
+// section to that file.
+func EncodeBufSection(metadata *wasmpluginv1.Metadata) ([]byte, error) {
```
bufdev, line 49: global functions should go below structs/interfaces, see "Go Code Standards" in notion, section "Within a file"

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+	if _, ok := exportedFuncs["_start"]; ok {
+		return wasmpluginv1.WasmABI_WASM_ABI_WASI_SNAPSHOT_PREVIEW1
+	} else if _, ok := exportedFuncs["run"]; ok {
+		return wasmpluginv1.WasmABI_WASM_ABI_GOJS
+	}
+	return wasmpluginv1.WasmABI_WASM_ABI_UNSPECIFIED
+}
+
+// PluginExecutorOption are options
+type PluginExecutorOption func(*PluginExecutor)
```
bufdev, line 90: Options go below type and constructor by convention in this codebase

bufdev (commented):
Looking really good for first pass - mostly style comments. Overall, consistency within the codebase trumps all, most of the comments are around that

private/buf/cmd/buf/command/generate/generate.go
pkwarren: So far cache dirs are only for the module cache, see bufcli's `createCacheDirs` as the current function that is used to create the cache dirs.

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,255 @@
...
+
+// WithMemoryLimitPages provides a custom per memory limit for a plugin
+// executor. The default is 8192 pages for 512MB.
+func WithMemoryLimitPages(memoryLimitPages uint32) PluginExecutorOption {
+	return func(pluginExecutor *PluginExecutor) {
+		pluginExecutor.runtimeConfig = pluginExecutor.runtimeConfig.WithMemoryLimitPages(memoryLimitPages)
```
pkwarren, line 96: We might consider allowing this to be specified by `buf generate` as a CLI option.

private/bufpkg/bufpluginexec/bufpluginexec.go
```diff
@@ -140,6 +144,12 @@ func NewHandler(
 		}
 		return newBinaryHandler(runner, pluginPath, handlerOptions.pluginPath[1:]), nil
 	}
+	if isWASM(pluginName) {
```
pkwarren, line 147: Are we going to support executing remote wasm plugins or just local paths? If we want to start with local only, then we will want to do more path validation here.

private/buf/cmd/buf/command/generate/generate.go
pkwarren: I also think this should indicate it is the wasm plugins compilation cache. At first glance, I thought this might be a location where we cached remote wasm plugins for execution.

go.mod
pkwarren: Should we bump to rc.2 now?

private/buf/cmd/buf/command/alpha/protoc/protoc.go
```diff
@@ -192,6 +194,10 @@ func run(
 			}
 			span.End()
 		}
+		wasmPluginExecutor, err := bufwasm.NewPluginExecutor(filepath.Join(container.CacheDirPath(), "wasmplugins"))
```
pkwarren, line 197: Can this be updated to use the same compile cache path as below?

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,143 @@
...
+	request *pluginpb.CodeGeneratorRequest,
+) error {
+	ctx, span := h.tracer.Start(ctx, "plugin_proxy", trace.WithAttributes(
+		attribute.Key("plugin").String(filepath.Base(h.pluginPath)),
+	))
+	defer span.End()
+
+	path, err := validateWASMFilePath(h.pluginPath)
+	if err != nil {
+		span.RecordError(err)
```
pkwarren, line 67: nit: It makes it a lot easier to ensure spans are marked appropriately if we use a named error and a defer.

private/bufpkg/bufpluginexec/wasm_handler_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestValidateWASMFilePath(t *testing.T) {
+	wasmPath := os.TempDir() + "/test.wasm"
+	notWasmPath := os.TempDir() + "/test.txt"
```
pkwarren, line 27: nit: Might be nice to use t.TempDir() which will automatically get cleaned up.

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+
+// Meta describes buf specific extensions for a wasm plugin.  A metadata may
```
pkwarren, line 19:
```suggestion
// Meta describes buf specific extensions for a wasm plugin. Metadata may
```

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+  repeated File files = 3;
+}
+
+// WASM_ABI specifies the abi this module expects buf to provide.
+enum WasmABI {
+  WASM_ABI_UNSPECIFIED = 0;
+
+  // Plugins compiled against
+  // https://github.com/WebAssembly/WASI/releases/tag/snapshot-01.
+  WASM_ABI_WASI_SNAPSHOT_PREVIEW1 = 1;
```
pkwarren, line 45: Not sure if we want to use something more like `WASM_ABI_WASI_P1` to more closely follow https://github.com/golang/go/issues/58141.

pkwarren (commented):
Looks good - just a couple of minor comments.

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
```diff
@@ -0,0 +1,56 @@
...
+
+syntax = "proto3";
+
+package buf.alpha.wasmplugin.v1;
+
+// Meta describes buf specific extensions for a wasm plugin.  A metadata may
+// be encoded in a custom WASM section named ".bufplugin", see
+// (https://webassembly.github.io/spec/core/binary/modules.html#binary-customsec)
+// for more info.
+message Meta {
```
bufdev, line 23: Meta isn't an acceptable name - we can talk about this offline

go.mod
bufdev: I don't love that this adds 3MB but on balance, probably nothing we can do. We should evaluate if there's places we can save on our binary size as it's starting to get a bit big

private/buf/bufgen/generator.go
```diff
@@ -53,12 +54,13 @@ func newGenerator(
...
 	storageosProvider storageos.Provider,
 	runner command.Runner,
+	wasmPluginExecutor bufwasm.PluginExecutor,
 	clientConfig *connectclient.Config,
 ) *generator {
 	return &generator{
 		logger:                logger,
 		storageosProvider:     storageosProvider,
-		appprotoexecGenerator: bufpluginexec.NewGenerator(logger, storageosProvider, runner),
+		appprotoexecGenerator: bufpluginexec.NewGenerator(logger, storageosProvider, runner, wasmPluginExecutor),
```
bufdev, line 63: Not from this PR but would be nice if we could quickly cleanup: this field should be renamed pluginexecGenerator

private/buf/cmd/buf/command/alpha/protoc/protoc.go
bufdev, line 197: It feels like this cache path should be a const somewhere, and not in an individual command's code - @pkwarren thoughts on where this should go?

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -46,6 +48,7 @@ const (
 	disableSymlinksFlagName     = "disable-symlinks"
 	typeFlagName                = "type"
 	typeDeprecatedFlagName      = "include-types"
+	wasmCompilationCacheDir     = "wasmplugins-compiled"
```
bufdev, line 51: Same comment - this shouldn't live as a const in command/*, this should be somewhere else. Perhaps bufcli

bufdev, line 51: Also perhaps named "wasm-bin"

private/buf/cmd/buf/command/alpha/protoc/protoc.go
bufdev: Also perhaps just named "wasm" or "wasmplugin"

private/buf/cmd/buf/command/generate/generate.go
bufdev: Or "wasmplugin-bin" (matching other cache dir)

private/bufpkg/bufpluginexec/bufpluginexec.go
```diff
@@ -206,3 +218,9 @@ type handlerOptions struct {
 func newHandlerOptions() *handlerOptions {
 	return &handlerOptions{}
 }
+
+// looksLikeWASM is a minimal check for WASM plugins. A more stringent validation
+// of the file is done in the handlers Handle method.
+func looksLikeWASM(pluginName string) bool {
```
bufdev, line 224: This feels a little ish imo - we should have a definitive test before we do any logic here

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func (h *wasmHandler) Handle(
+	ctx context.Context,
+	container app.EnvStderrContainer,
+	responseWriter appproto.ResponseBuilder,
+	request *pluginpb.CodeGeneratorRequest,
+) error {
+	ctx, span := h.tracer.Start(ctx, "plugin_proxy", trace.WithAttributes(
```
bufdev, line 60: Should this be "plugin_proxy"?

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,143 @@
...
+		attribute.Key("plugin").String(filepath.Base(h.pluginPath)),
+	))
+	defer span.End()
+
+	path, err := validateWASMFilePath(h.pluginPath)
```
bufdev, line 65: This should probably be done in the constructor as opposed to every call

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,143 @@
...
+func validateWASMFilePath(path string) (string, error) {
+	path, err := filepath.Abs(path)
+	if err != nil {
+		return path, err
+	}
+	info, err := os.Stat(path)
+	if err != nil {
+		return path, err
+	}
+	if info.IsDir() || info.Size() == 0 || !strings.HasSuffix(path, ".wasm") {
```
bufdev, line 139: !IsRegular?

private/bufpkg/bufpluginexec/wasm_handler_test.go
bufdev: I'd say more than a nit heh :-)

bufdev (commented):
@twoism let's do a 15 mim video review tomorrow and get this 
across, will be much much easier heh, but in meantime adding a few more comments

private/bufpkg/bufpluginexec/wasm_handler.go
saquibmian: Do we not also want to `span.SetStatus(codes.Error, err.Error())` here?

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,253 @@
...
+		}
+		defer func() {
+			retErr = multierr.Append(retErr, closer.Close(ctx))
+		}()
+		var module api.Module
+		module, err = runtime.InstantiateModule(ctx, plugin.Module, config)
+		defer func() {
+			retErr = multierr.Append(retErr, module.Close(ctx))
+		}()
+	default:
```
saquibmian, line 206: Are you missing a `wasi.Run` or something? I'm not sure where the plugin actually gets executed in this block.

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -0,0 +1,253 @@
...
+			return fmt.Errorf("error instantiating wasi: %w", err)
+		}
+		defer func() {
+			retErr = multierr.Append(retErr, closer.Close(ctx))
+		}()
+		var module api.Module
+		module, err = runtime.InstantiateModule(ctx, plugin.Module, config)
+		defer func() {
+			retErr = multierr.Append(retErr, module.Close(ctx))
+		}()
```
saquibmian, line 205: Are you missing a `wasi.Run` or something? I'm not sure where the plugin actually gets executed in this block.

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
twoism: The wazero/experimental module is still `wasi_snapshot_preview1` so I am hesitant to change this. Think we should?

private/bufpkg/bufwasm/bufwasm.go
pkwarren: See the example here: https://pkg.go.dev/github.com/tetratelabs/wazero@v1.0.0-rc.2/imports/wasi_snapshot_preview1

> // InstantiateModule runs the "_start" function which is like a "main" function.

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
pkwarren: Nah we can leave it.

private/bufpkg/bufwasm/bufwasm.go
twoism: Execution is done in the handler, bufwasm.go just sets up the plugin for later execution.

https://github.com/bufbuild/buf/blob/64d8cda65727a0f6e12c441f2f1f239284a793fd/private/bufpkg/bufpluginexec/wasm_handler.go#L86

saquibmian: Ah okay, I was tripped up by the `gojs.Run`.

twoism: No worries, the two execution paths make things confusing

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,143 @@
...
+	requestData, err := protoencoding.NewWireMarshaler().Marshal(request)
+	if err != nil {
+		span.RecordError(err)
+		span.SetStatus(codes.Error, err.Error())
+		return err
+	}
+	pluginBytes, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
```
saquibmian, line 79: Did you intentionally not instrument this error and the one below this?

twoism: Those two err paths were the only two that did not have tracing but I cannot think of why we wouldn't have them traced. @pkwarren were you around for this code path?

pkwarren: I can't think of why we wouldn't - it was probably just an accident.

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,162 @@
...
+		return path, fmt.Errorf("invalid WASM file: %s", path)
+	}
+	return path, nil
+}
+
+// IsAlphaWasmEnabled returns if BUF_ALPHA_ENABLE_WASM is set to true.
+func IsAlphaWasmEnabled(container app.EnvContainer) (bool, error) {
+	enabled, _ := container.GetEnvBoolValue(AlphaEnableWasmEnvKey)
+	if !enabled {
+		return false, errors.New("alpha wasm support is disabled")
```
saquibmian, line 159: One of these values doesn't matter, I would either hoist the error up to the caller, or drop the bool and rename this to `RequireAlphaWasm(app.EnvContainer) error`.

proto/buf/alpha/wasmplugin/v1/wasmplugin.proto
twoism: Changing the name to `ExecConfig` after discussing with P

private/bufpkg/bufpluginexec/bufpluginexec.go
```diff
@@ -126,34 +130,42 @@ func GenerateWithProtocPath(protocPath string) GenerateOption {
...
+
+	// Initialize binary plugin handler when path is specified with optional args. Return
+	// on error as something is wrong with the supplied pluginPath option.
 	if len(handlerOptions.pluginPath) > 0 {
-		pluginPath, err := unsafeLookPath(handlerOptions.pluginPath[0])
-		if err != nil {
-			return nil, err
-		}
-		return newBinaryHandler(runner, pluginPath, handlerOptions.pluginPath[1:]), nil
+		return NewBinaryHandler(runner, handlerOptions.pluginPath[0], handlerOptions.pluginPath[1:])
```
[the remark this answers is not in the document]
twoism: I did that because I noticed the exported version version does the error handling that I removed and calls `newProtocProxyHandler` which is not needed in the other two cases.

private/bufpkg/bufpluginexec/wasm_handler.go
twoism: Clarified and updated offline.

private/bufpkg/bufpluginexec/bufpluginexec.go
twoism: I am adding comments around this check.

private/bufpkg/bufpluginexec/wasm_handler.go
twoism: That is the name in binary_handler.go as well, so I am hesitant to change it in this PR.

private/pkg/app/app.go
```diff
@@ -28,14 +28,18 @@ import (
...
 	// Returns empty string if the key is not set or the value is empty.
 	Env(key string) string
 	// ForEachEnv iterates over all non-empty environment variables and calls the function.
 	//
 	// The value will never be empty.
 	ForEachEnv(func(string, string))
+	// GetEnvBoolValue gets and parses the environment variable bool value for the key.
+	//
+	// Returns false and the parse error if the key is not set or the value is empty.
+	GetEnvBoolValue(key string) (bool, error)
```
pkwarren, line 42: I feel like this would be more useful if we passed a `defaultValue bool` argument here to indicate which value to use if the key isn't found in the environment.

pkwarren (commented):
Changes look great - one suggestion on the new function in env container.

private/pkg/app/app.go
twoism: +1

private/bufpkg/bufpluginexec/wasm_handler.go
bufdev: ok fair

private/bufpkg/bufpluginexec/wasm_handler.go
bufdev: I think I meant using this function instead https://pkg.go.dev/io/fs#FileMode.IsRegular

private/bufpkg/bufpluginexec/wasm_handler.go
```diff
@@ -0,0 +1,161 @@
...
+	"github.com/bufbuild/buf/private/pkg/protoencoding"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
+	"google.golang.org/protobuf/types/pluginpb"
+)
+
+// AlphaEnableWasmEnvKey is an env var to enable WASM local plugin execution
+const AlphaEnableWasmEnvKey = "BUF_ALPHA_ENABLE_WASM"
```
bufdev, line 38: This should be in `bufcli` with the rest of our env vars

private/pkg/app/app.go
```diff
@@ -39,7 +39,7 @@ type EnvContainer interface {
 	// GetEnvBoolValue gets and parses the environment variable bool value for the key.
 	//
 	// Returns false and the parse error if the key is not set or the value is empty.
-	GetEnvBoolValue(key string) (bool, error)
+	GetEnvBoolValue(key string, defaultVal bool) (bool, error)
```
bufdev, line 42: I didn't know this function existed on `EnvContainer` - separately, this function should be removed, but for this PR, don't add to it more. If we want something like this, then `EnvBool(envContainer EnvContainer, key string) (bool, bool)` or `EnvBool(envContainer EnvContainer, key string, default bool) bool` would be acceptable as a top-level function

bufdev: (which, if you want to do, you might as well delete this function from EnvContainer and then add EnvBool heh)

twoism: I added the method to begin with because `container.Env` returns an empty string when a key is not set which means that callers will always need to check for an empty string before passing it to `strconv.ParseX`, then handle the parse errors.

bufdev: Yea this is more philosophy of this package - the containers are meant to be the absolute minimal thing you need to represent what they are containing, if you want to do helpers on top of them, add them as global functions. See `Args` as an example - just being consistent here basically

twoism: 👍

bufdev (approved):

PR #1895, opened 2023-03-07
Clarify that format takes a source and not an input

joshcarp:
`buf format` does not take an `input` but a `source`

The following command fails
```
buf build | buf format - 
Failure: format was "bin" but must be one of [dir,git,mod,protofile,tar,targz,zip]
```

and therefore the correct terminology to use is `source`


private/buf/cmd/buf/command/format/format.go
```diff
@@ -183,15 +183,15 @@ type flags struct {
 	Output          string
 	Write           bool
 	// special
-	InputHashtag string
```
bufdev, line 186: `InputHashtag` is the same name across all commands, it's a special variable and matches with `bufcli.GetInputValue` naming, updating this would make `format` the one inconsistent command here - I'd like to keep this the same so you can `grep` for this in the future

bufdev (approved):

PR #1912, opened 2023-03-13
Add buf beta price command

bufdev:


private/buf/cmd/buf/command/beta/price/price.go
```diff
@@ -0,0 +1,185 @@
...
+		Stats:    stats,
+		NumTypes: stats.NumMessages + stats.NumEnums + stats.NumMethods,
+	}
+	buckets := tmplData.NumTypes / 100
+	if tmplData.NumTypes%100 != 0 {
+		tmplData.ChargeableIsRounded = true
+		buckets++
+	}
+	tmplData.ChargeableTypes = buckets * 100
+	tmplData.TeamsDollarsPerMonth = fmt.Sprintf("%.2f", float64(tmplData.ChargeableTypes)*teamsPerTypeDollarsPerMonth)
```
joshcarp, line 182: Not to be over engineering but if we do any more advanced stuff with money calculations we should probably use a decimal/money package as to not have rounding errors down the line

joshcarp (approved):

timostamm (approved):

private/buf/cmd/buf/command/beta/price/price.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/spf13/pflag"
+)
+
+const (
+	disableSymlinksFlagName     = "disable-symlinks"
+	teamsPerTypeDollarsPerMonth = float64(0.50)
+	proPerTypeDollarsPerMonth   = float64(1)
+	tmplCopy                    = `Current BSR pricing:
+
+  - Teams: $50/month per 100 types
```
saquibmian, line 39: The [pricing page](https://buf.build/pricing/) refers to this as `Team` (no plural).

private/buf/cmd/buf/command/beta/price/price.go
```diff
@@ -0,0 +1,189 @@
...
+
+Your sources have:
+
+  - {{.NumMessages}} messages
+  - {{.NumEnums}} enums
+  - {{.NumMethods}} methods
+
+This adds up to {{.NumTypes}} types.{{if .ChargeableIsRounded}}
+
+We bill in increments of 100 types. Types are summed up across
```
saquibmian, line 53: nit: 
```suggestion
We bill in increments of 100 types. Types are totaled across
```

private/buf/cmd/buf/command/beta/price/price.go
saquibmian, line 39: Can you indicate that this is USD?

bufdev: That's assumed by everyone reading from a tool like this, seems superfluous

jhump, line 39: This information seems like it should be downloaded from the server. If it's built-in to the CLI, then someone on an old version of the CLI cannot get accurate pricing projections in the face of any tweaks or changes to pricing in the future.

Maybe a better way to do this is to send all of the stats to a simple BSR endpoint to compute the estimated price?

bufdev: Yea that's likely a better way, but that's more time than we want to allocate to this now, this gets this across the line and out there in under an hour's worth of work. We mention that the pricing information may not be the latest in the copy.

PR #1916, opened 2023-03-14
Add Labels API definition

doriable:
This adds the Protobuf definition for the Labels API.
Would love to get some thoughts on the following:
- [x] The `LabelNamespace` enum -- thoughts on having an "all" kind of namespace for making `glob` requests or is "unspecified" sufficient?
- [x] Do we want to return the "to" commit for `MoveLabel`?


proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,68 @@
...
+
+package buf.alpha.registry.v1alpha1;
+
+import "google/protobuf/timestamp.proto";
+
+enum LabelNamespace {
+  LABEL_NAMESPACE_UNSPECIFIED = 0;
+  // This is for globs, '*', in requests. When a label is created
+  // it will be assigned to a specific namespace.
+  LABEL_NAMESPACE_ALL = 1;
```
rodaine, line 25: Can a label exist in an ALL namespace? If the other options are all concrete, I'd say stick with UNSPECIFIED for this use-case.

rodaine (approved):

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: A label cannot/should not be created in the `ALL` namespace. Yeah, I was a bit concerned that this could cause some bugs. At the data model layer, this option won't exist, but having it at the API layer _might_ be confusing.

twilly: I wouldn't expect a creation call to be made on ALL. Querying, though, could be a search for the name in all namespaces.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,68 @@
...
+message Label {
+  LabelNamespace namespace = 1;
+  string name = 2;
+  string commit_id = 3;
+}
+
+service LabelService {
+  rpc CreateLabel(CreateLabelRequest) returns (CreateLabelResponse);
+  rpc MoveLabel(MoveLabelRequest) returns (MoveLabelResponse);
+  rpc GetLabelByNameAndNamespace(GetLabelByNameAndNamespaceRequest) returns (GetLabelByNameAndNamespaceResponse);
```
twilly, line 37: This is a misnomer if namespace is committed, which we should support. E.g. this should be a valid call:

```json
{"name": "main"}
```

`GetLabel` should be sufficient since the request specifies what the parameters are.

```proto
rpc GetLabel(GetLabelRequest) returns (GetLabelResponse);
```

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,68 @@
...
+  LABEL_NAMESPACE_ALL = 1;
+  LABEL_NAMESPACE_TAG = 2;
+  LABEL_NAMESPACE_BRANCH = 3;
+  LABEL_NAMESPACE_GITCOMMIT = 4;
+}
+
+message Label {
+  LabelNamespace namespace = 1;
+  string name = 2;
+  string commit_id = 3;
```
twilly, line 31: Perhaps split the label from its value. commit_id is a label value and needs to be supported in Create, Move, and Get calls. E.g. in Move there's "from" and "to" but they're strings and not describing "from_commit_id" and "to_commit_id". If this value becomes a oneof definition, we'll have a ton of cruft.

```proto
message LabelValue {
  string commit_id = 1;
}
```

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,68 @@
...
+}
+
+message CreateLabelResponse {
+  string commit_id = 1;
+}
+
+message MoveLabelRequest {
+  Label label = 1;
+  string from = 2;
+  string to = 3;
```
twilly, line 53: See my comment about splitting a label by its name and its value.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,65 @@
...
+service LabelService {
+  rpc CreateLabel(CreateLabelRequest) returns (CreateLabelResponse);
+  rpc MoveLabel(MoveLabelRequest) returns (MoveLabelResponse);
+  rpc GetLabelByNameAndNamespace(GetLabelByNameAndNamespaceRequest) returns (GetLabelByNameAndNamespaceResponse);
+}
+
+message CreateLabelRequest {
+  Label label = 1;
+  string author = 2;
+  google.protobuf.Timestamp create_time = 3;
```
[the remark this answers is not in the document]
doriable: Yep, that is correct!

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,65 @@
...
+  LABEL_NAMESPACE_UNSPECIFIED = 0;
+  LABEL_NAMESPACE_TAG = 1;
+  LABEL_NAMESPACE_BRANCH = 2;
+  LABEL_NAMESPACE_GITCOMMIT = 3;
+}
+
+message Label {
+  LabelNamespace namespace = 1;
+  string name = 2;
+  string commit_id = 3;
```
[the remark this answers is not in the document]
doriable: This naming might be confusing, sorry -- it's the BSR commit that the label is attached to.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,65 @@
...
+  Label label = 1;
+  string from = 2;
+  string to = 3;
+}
+
+message MoveLabelResponse {}
+
+message GetLabelByNameAndNamespaceRequest {
+  string name = 1;
+  LabelNamespace namespace = 2;
```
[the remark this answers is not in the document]
doriable: My original thought here was that if we set the "all" or "unspecified" namespace for querying, it would query across namespaces. But to Tristan's point above, if we're only querying by name, then maybe this is moot.

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: Also, is there no circumstance where we'd want to query for a name within a specific namespace? If we are always only ever retrieving by name, we are priced into querying all labels in all namespaces always?

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,65 @@
...
+
+message MoveLabelRequest {
+  Label label = 1;
+  string from = 2;
+  string to = 3;
+}
+
+message MoveLabelResponse {}
+
+message GetLabelByNameAndNamespaceRequest {
```
pkwarren, line 58: I think this needs to have some other field to restrict it to a given repo. Right now, it doesn't have the context to search anything other than all labels w/ a given name+namespace across all repos.

proto/buf/alpha/registry/v1alpha1/labels.proto
twilly: > we would not be providing a namespace for querying, which also means we won't need the ALL namespace, right?

Yep, we can use omission to mean all.

> there no circumstance where we'd want to query for a name within a specific namespace? If we are always only ever retrieving by name, we are priced into querying all labels in all namespaces always?

I believe both kinds of calls will be made. A plain name would be often queried, but that shouldn't exclude including a namespace. Similarly to getting a single name, this can support getting a /namespace/. E.g. to list all tags:

```json
{"namespace": "LABEL_NAMESPACE_TAG"}
```

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,65 @@
...
+  string name = 1;
+  LabelNamespace namespace = 2;
+}
+
+message GetLabelByNameAndNamespaceResponse {
+  Label label = 1;
+}
```
twilly: Since Get could return more than one label:

```suggestion
message GetLabelByNameAndNamespaceResponse {
  repeated Label label = 1;
}
```

doriable: In that case, I'm going to adjust it to `GetLabels` :)

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: I'm currently using the `Reference` structure -- do you think we should still define a `LabelValue` structure? It's a bit confusing, because I expect that down the road, these should converge, but for now `Reference` might be too large. I think I've convinced myself to define a separate `LabelValue` message, lol.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,70 @@
...
+  LABEL_NAMESPACE_UNSPECIFIED = 0;
+  LABEL_NAMESPACE_TAG = 1;
+  LABEL_NAMESPACE_BRANCH = 2;
+  LABEL_NAMESPACE_GIT_COMMIT = 3;
+}
+
+message Label {
+  LabelNamespace namespace = 1;
+  string name = 2;
+  LabelValue label_value = 3;
```
twilly, line 31: Probably should omit this field. See comment in move label request.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,70 @@
...
+}
+
+message CreateLabelRequest {
+  Label label = 1;
+  optional string author = 2;
+  optional google.protobuf.Timestamp create_time = 3;
+}
+
+message CreateLabelResponse {
+  string commit_id = 1;
```
twilly, line 51: `LabelValue` type here.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,70 @@
...
+}
+
+message MoveLabelRequest {
+  Label label = 1;
+  LabelValue from = 2;
```
twilly, line 56: Looking at this again makes this a very interesting form. What happens when `label` contains a value that's different than `from`?

```proto
{
  "label": {"namespace": "LABEL_NAMESPACE_TAG", "name": "ponies", value: "dashie"},
  "from": "rainbow dash",
  "to": "rd"
}
```
Moving `LabelValue` out of `Label` makes it clear values are not supposed to be supplied where they don't make sense.

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,70 @@
...
+  LabelValue to = 3;
+}
+
+message MoveLabelResponse {}
+
+message GetLabelsRequest {
+  string name = 1;
```
twilly, line 63: This is also optional.

(They *all* are, but I get what Philip is saying about autodocumenting code.)

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,70 @@
...
+  string module_entity_id = 2;
```
twilly, line 64: `buf.alpha.registry.v1alpha1.Repository` seems more appropriate?

twilly: Also, a repo reference needs to be present in all the calls.

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: The intent here would be that you can only move a label if the `from` supplied matches the label value -- this helps prevent some of the race conditions if multiple move operations are being fired. That being said, I should probably demonstrate that `from` is optional.

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: Ah yeah, because we want to be able to query all labels in a namespace. Makes sense.

proto/buf/alpha/registry/v1alpha1/labels.proto
twilly: I guess that's true. I'll have to check what our indexing looks like there. Conceptually I was always scoping commit identifiers to a repo, but these ids are supposed to be globally unique.

Related, this module_entity_id is yet another way of scoping to a repository. The type I suggested is pretty big and other APIs represent it differently. Let's chat about which pattern we should go with and avoid introducing a third (or even fourth) way to represent a repo address on the wire.

doriable: Yeah -- we should definitely chat about this. I was a little bit concerned about bringing in the large types (I also thought about the `Reference` type in https://github.com/bufbuild/buf/blob/dce75c97b1ed32ffe321916ad611d32e0cad6737/proto/buf/alpha/registry/v1alpha1/reference.proto#L22-L36 but I had a couple of concerns regarding the scope of this type, so I ended up not doing that).

proto/buf/alpha/registry/v1alpha1/labels.proto
doriable: Yeah, this makes sense -- but also makes sense that we're doing `GetLabels` :)

proto/buf/alpha/registry/v1alpha1/labels.proto
twilly: What I'm raising is the label's from value is specified twice in this message. Once in the `label` field and once in the `from` field. My example exemplifies the question of what happens when the from value is duplicated but with different values?

doriable: Ah yeah, I see. Okay, this makes sense. Let me adjust this. For `GetLabels`, I do think we may want to return `LabelValue` alongside the name and namespace, so let me push up an update :)

proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -0,0 +1,73 @@
...
+message MoveLabelResponse {}
+
+message GetLabelsRequest {
+  string module_entity_id = 1;
+  optional string name = 2;
+  optional LabelNamespace namespace = 3;
+}
+
+message GetLabelsResponse {
+  message LabelWithValue {
```
twilly, line 68: nit: I generally dislike prepositions in a name. It's suggestive that we're not modeling the problem right.

We started with a definition of Label that contained both a name and value. Now the value is separate but not the name part.

LabelName: name in a namespace
LabelValue: value
Label: a name in a namespace and a value

That would make this message read a bit better:

```proto
message GetLabelsResponse {
  repeated Label labels = 1;
}
```

It'll change other messages but they too remain clear. Like Move could be:

```proto
message MoveLabelRequest {
  LabelName name = 1;
  optional LabelValue from = 2;
  LabelValue to = 3;
}
```

twilly (approved):
🚢

PR #1918, opened 2023-03-14
Dogfood Remote Packages (Initial Steps)

rodaine:


data/template/buf.go-client.gen.yaml
```diff
@@ -4,10 +4,11 @@ version: v1
 managed:
   enabled: true
   go_package_prefix:
-    default: github.com/bufbuild/buf/private/gen/proto/go
+    default: buf.build/gen/go/bufbuild/buf/bufbuild/connect-go
```
bufdev, line 7: This template shouldn't be needed at all anymore right?

data/template/buf.go.gen.yaml
```diff
@@ -3,7 +3,7 @@ version: v1
 managed:
   enabled: true
   go_package_prefix:
-    default: github.com/bufbuild/buf/private/gen/proto/go
+    default: buf.build/gen/go/bufbuild/buf/protocolbuffers/go
```
bufdev, line 6: Same - all templates should be deleted I believe

go.mod
```diff
@@ -70,3 +72,8 @@ require (
 	golang.org/x/sys v0.6.0 // indirect
 	golang.org/x/text v0.8.0 // indirect
 )
+
+replace (
+	buf.build/gen/go/bufbuild/buf/bufbuild/connect-go => ./private/gen/proto/connect
```
bufdev, line 77: Why is this happening?

make/buf/all.mk
```diff
@@ -97,7 +97,8 @@ bufgeneratedeps:: \
 
 .PHONY: bufgeneratecleango
 bufgeneratecleango:
```
bufdev, line 99: All the `bufgenerate*` commands should be done from `all.mk`

bufdev (commented):
This is likely just an intermediate step but commenting in case not

PR #1934, opened 2023-03-22
Add label value to CreateLabel RPC

doriable:
Adds `label_value` to `CreateLabel`, which is the commit reference we are
creating the label for.


twilly (approved):
:shipit:

Issue #34, opened 2020-02-28
Investigate replacing go-git with calls to git binary


bufdev:
See #21 and #14.

https://github.com/src-d/go-git seems to no longer be maintained, and going against the grain by using it has caused more problems than it has solved. While Buf aims to be self-contained and not shell out, `git` is a pretty widespread exception, including in the `go` binary itself.

One issue is temporary clones - for Buf usage in a long-running service - for example, a GitHub App that performs check runs (which we built internally), this could cause problems we need to make sure are handled (or just not do `git` operations in long-running services altogether for now).

bufdev:
Added as `--experimental-git-clone` on https://github.com/bufbuild/buf/commit/c75b6195beced011672dbffd4a0a11d004ccbca4, please test if you can.

willstott101:
go-git seems the be back on the menu btw: https://github.com/go-git/go-git

bufdev:
This is done on master, will do a release soon.

PR #1936, opened 2023-03-22
A read bucket for git trees

twilly:
Modules are built from buckets, meaning if we want to build a module from git, then we'll need a bucket for it. It's possible to checkout a git commit and build from a storageos bucket but there's a better way: read directly from git's storage.

go-git exposes an API for reading git objects, including git trees. Trees are how git represents and stores files. Gluing a storage bucket onto go-git's tree API means we can build a module out of a git repository without copying file content to a filesystem.

Here's a rough example of how to instantiate a bucket from a git repository's HEAD commit:

```go
import (
  "github.com/bufbuild/buf/private/pkg/git"
  gogit "github.com/go-git/go-git/v5"
)

var repo *gogit.Repository

head, _ := repo.Head()
commit, _ := repo.CommitObject(head.Hash())
tree, _ := commit.Tree()
bucket := NewTreeReader(tree)
```


bufdev (changes requested):
https://github.com/bufbuild/buf/issues/34

go-git was a disaster for us, I had to spend a lot of time and energy ripping it out early on - I am hard against depending on it in any way unless something major has changed.

bufdev:
Also note that go-git secretly doesn't use pure Golang in certain cases: https://github.com/go-git/go-git/commit/a0ff9a899d66243dfcfc30993ed9bf3d6c982400

bufdev:
For some context, @twilly and I caught up offline, we're going to find a fun solution cuz git-based buckets would be great

twilly:
Closing this in favor of a go-git-less edition, currently drafted as PR #2018.

PR #1937, opened 2023-03-23
Remove token creation from alpha cli

pgmitche:


mfridman:
Probably worth updating the [changelog](https://github.com/bufbuild/buf/blob/main/CHANGELOG.md)

Tokens must be created through the BSR UI.

mfridman (approved):

PR #1938, opened 2023-03-23
Add visibility flag to curated plugins

mfridman:
This PR adds a visibility flag to the `buf alpha plugin push` command. It will is now a mandatory flag.

```
--visibility string        The plugin's visibility setting. Must be one of [public,private]
```


pkwarren (approved):
Looks great!

PR #1939, opened 2023-03-23
Set default for curated plugin visibility

mfridman:
Remove visibility as a required flag, will re-introduce this once this command is promoted to `beta` (from `alpha`) and the backend implementation is complete.


pkwarren (commented):
I think visibilityFlagToVisibility might also need updates unless we set a default for the cobra arg.

pkwarren (approved):

PR #1942, opened 2023-03-27
Upgrade deps including wazero v1.0.0

pkwarren:


saquibmian (approved):
🎉

pkwarren:
Looked through the changelog - the one potential simplification we could make would be to remove https://github.com/bufbuild/buf/blob/8682e21f5d05b4a0b8469a30a23a38125be78858/private/bufpkg/bufwasm/bufwasm.go#L211-L213, however I didn't audit all the code paths to make sure that's safe.

PR #1950, opened 2023-03-29
Fix comment for `StaticTokenProvider`

saquibmian:
`BUF_TOKEN` does not contain usernames, and does not support mixed remote-less and remote tokens.


tonyli233 (approved):

private/bufpkg/bufconnect/static_token_provider_test.go
saquibmian, line 1: Looks like there are already tests for this in `interceptors_test.go`. I'm going to move those tests here.

bufdev (approved):

PR #1958, opened 2023-03-30
Make a few style changes to wasm package

pkwarren:
PR #1956 pulled in some great fixes for performance and fixing memory leaks. This updates the code to more closely follow the style used by Buf (in particular around multierr).

### DIFF
--- a/private/bufpkg/bufwasm/bufwasm.go
+++ b/private/bufpkg/bufwasm/bufwasm.go
@@ -149,27 +149,36 @@ func WithMemoryLimitPages(memoryLimitPages uint32) PluginExecutorOption {
 // CompilePlugin takes a byte slice with a valid wasm module, compiles it and
 // optionally reads out buf plugin metadata.
 func (e *WASMPluginExecutor) CompilePlugin(ctx context.Context, plugin []byte) (_ *CompiledPlugin, retErr error) {
-
 	// Configure the compilation cache, which must be closed after the runtime.
 	var cache wazero.CompilationCache
 	if e.compilationCacheDir == "" {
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -149,27 +149,36 @@ func WithMemoryLimitPages(memoryLimitPages uint32) PluginExecutorOption {
...
-		if retErr != nil {
-			return
+		var err error
+		cache, err = wazero.NewCompilationCacheWithDir(e.compilationCacheDir)
+		if err != nil {
+			return nil, err
 		}
 	}
+	defer func() {
+		if retErr != nil {
```
pkwarren, line 164: Instead of needing to wrap each error call site, this moves the close into a defer so if this method returns any error, we'll ensure to close the compilation cache and runtime.

private/bufpkg/bufwasm/bufwasm.go
```diff
@@ -149,27 +149,36 @@ func WithMemoryLimitPages(memoryLimitPages uint32) PluginExecutorOption {
...
+			retErr = multierr.Append(retErr, runtime.Close(ctx))
+		}
+	}()
 
 	// Note: before we start accepting user plugins, we should do more
 	// validation on the metadata here: file path cleaning etc.
-	var compiledModule wazero.CompiledModule
-	compiledModule, retErr = runtime.CompileModule(ctx, plugin)
```
pkwarren, line 170: Small style tweak here - we typically only use the named return variables in a defer.

saquibmian (approved):

Issue #1955, opened 2023-03-29
Add Camel Case Field Lint Rule 🐪


cobbinma:
Hi 👋 

I know this is at odds with https://protobuf.dev/programming-guides/style/#message-field-names but we have historically used camel case for field names in our protobuf repository and don't want to change for the sake of consistency.

At the moment we can ignore non snake case but not ensure camel case field names.

Could a lint rule be added to check fields are written in camel case?

`FIELD_CAMEL_CASE`

bufdev:
Thanks for the input here. We've taken the position with our lint rules that there should be one way to do things - there's plenty of disagreement on what the right way to do things is, but we think it is most important to present to the community a single way. With that said, a rule like `FIELD_CAMEL_CASE` being built into `buf` would present a different story to end users - it would say that there are multiple valid ways to style field names. As such, we don't want to add this linter.

What you can do:

- Ignore the offending fields and migrate them over time. From our position, this would be best - CamelCased fields are against standard Protobuf, and should be dealt with eventually, as opposed to continually enforcing that they remain CamelCased.
- Write a Protobuf plugin in the style of `protoc-gen-buf-lint` that lints for this specific rule, and use that in your codebase.

Going to close this issue, but we appreciate it (and the associated PR).

cobbinma:
Thanks @bufdev 👍 
https://github.com/cobbinma/protoc-gen-camel

PR #1960, opened 2023-03-30
Add Camel Case Field Lint Rule 🐪

cobbinma:
[#1955](https://github.com/bufbuild/buf/issues/1955)

Added `FIELD_CAMEL_CASE` lint rule to check that field names are in camel case format.


bufdev:
Closing per message on #1955

PR #1962, opened 2023-03-31
Audit delete repository commit

smallsamantha:
Add payload for repository commit deletion.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -53,6 +53,7 @@ enum ResourceType {
   RESOURCE_TYPE_CURATED_PLUGIN = 8;
   RESOURCE_TYPE_TOKEN = 10;
   RESOURCE_TYPE_SCIM_TOKEN = 11;
+  RESOURCE_TYPE_REPOSITORY_DRAFT_COMMIT = 12;
```
pkwarren, line 56: Should we re-use `RESOURCE_TYPE_REPOSITORY_COMMIT` here?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -95,6 +96,7 @@ enum EventType {
   EVENT_TYPE_USER_REACTIVATED = 25;
   EVENT_TYPE_SCIM_TOKEN_CREATED = 26;
   EVENT_TYPE_SCIM_TOKEN_DELETED = 27;
+  EVENT_TYPE_DELETE_REPOSITORY_DRAFT_COMMIT = 28;
```
pkwarren, line 99: Same - `EVENT_TYPE_REPOSITORY_COMMIT_DELETED`?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -354,3 +357,12 @@ message PayloadSCIMTokenCreated {
 }
 
 message PayloadSCIMTokenDeleted {}
+
+message PayloadDeleteRepositoryDraftCommit {
+  // owner_name is the name of the owner of the repository from which the draft will be deleted.
+  string owner_name = 1;
+  // repository_name is the name of the repository from which the draft will be deleted.
+  string repository_name = 2;
```
pkwarren, line 365: Should we add `owner_id` and `repository_id` here to be similar to the push commit message?

pkwarren:
Even though we only support deleting draft commits, it feels good to audit them using the same resource types I think.

smallsamantha:
> Even though we only support deleting draft commits, it feels good to audit them using the same resource types I think.

yup was on the fence on this but i like the idea of reusing the resource type, will update 👍

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -354,3 +357,12 @@ message PayloadSCIMTokenCreated {
 }
 
 message PayloadSCIMTokenDeleted {}
+
+message PayloadDeleteRepositoryDraftCommit {
```
unmultimedio, line 361: This should be exactly the same as `PayloadRepositoryCommitPushed`, but `PayloadRepositoryCommitDeleted`

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: jynx haha

pkwarren (approved):

PR #1963, opened 2023-03-31
Add repo owner and name to move label rpc

doriable:
We need to include the owner and repository name in the event
that we are not providing a label value, to ensure that we are
able to query for the correct label.


proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -60,6 +60,8 @@ message MoveLabelRequest {
   LabelName label_name = 1;
   optional LabelValue from = 2;
   LabelValue to = 3;
+  string repository_owner_name = 4;
```
smallsamantha, line 63: annoying nit 😬 : should this just be repository_owner to be consistent with other requests? ([repository commits](https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/repository_commit.proto#L80-L83), etc)

PR #1965, opened 2023-03-31
Better error message from `buf curl` when trying to use reflection over plain-text HTTP

jhump:
Before:
```
> buf curl http://localhost:30485/buf.knit.demo.swapi.film.v1.FilmService/GetFilms
Failure: failed to resolve symbol "buf.knit.demo.swapi.film.v1.FilmService": unknown: HTTP status 505 HTTP Version Not Supported
```

After:
```
> buf curl http://localhost:30485/buf.knit.demo.swapi.film.v1.FilmService/GetFilms
Failure: --reflect cannot be used with plain-text URLs (http) unless --http2-prior-knowledge flag is set
```

### DIFF
--- a/private/buf/cmd/buf/command/curl/curl.go
+++ b/private/buf/cmd/buf/command/curl/curl.go
@@ -420,18 +420,36 @@ func (f *flags) validate(isSecure bool) error {
 		return fmt.Errorf("grpc protocol cannot be used with plain-text URLs (http) unless --%s flag is set", http2PriorKnowledgeFlagName)
 	}
 
+	if f.Schema != "" && f.Reflect {
+		if f.flagSet.Changed(reflectFlagName) {
+			// explicitly enabled both
+			return fmt.Errorf("cannot specify both --%s and --%s", schemaFlagName, reflectFlagName)
+		}
+		// Reflect just has default value; unset it since we're going to use --schema instead
+		f.Reflect = false
+	}
+	if !f.Reflect && f.Schema == "" {
+		return fmt.Errorf("must specify --%s if --%s is false", schemaFlagName, reflectFlagName)
+	}
+	schemaIsStdin := strings.HasPrefix(f.Schema, "-")
 	if (len(f.ReflectHeaders) > 0 || f.flagSet.Changed(reflectProtocolFlagName)) && !f.Reflect {
 		return fmt.Errorf(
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -420,18 +420,36 @@ func (f *flags) validate(isSecure bool) error {
...
+			// explicitly enabled both
+			return fmt.Errorf("cannot specify both --%s and --%s", schemaFlagName, reflectFlagName)
+		}
+		// Reflect just has default value; unset it since we're going to use --schema instead
+		f.Reflect = false
+	}
+	if !f.Reflect && f.Schema == "" {
+		return fmt.Errorf("must specify --%s if --%s is false", schemaFlagName, reflectFlagName)
+	}
+	schemaIsStdin := strings.HasPrefix(f.Schema, "-")
```
jhump, line 434: I moved this up so that it happens before the other flag checks. Mainly, there's the possibility this logic _changes_ the value of `f.Reflect` (line 429). So we need that settled before we use that value to check other flags.

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -420,18 +420,36 @@ func (f *flags) validate(isSecure bool) error {
...
-	if _, err := bufcurl.ParseReflectProtocol(f.ReflectProtocol); err != nil {
-		return fmt.Errorf(
-			"--%s value must be one of %s",
-			reflectProtocolFlagName,
-			stringutil.SliceToHumanStringOrQuoted(bufcurl.AllKnownReflectProtocolStrings),
-		)
+	if f.Reflect {
+		if !isSecure && !f.HTTP2PriorKnowledge {
+			return fmt.Errorf("--%s cannot be used with plain-text URLs (http) unless --%s flag is set", reflectFlagName, http2PriorKnowledgeFlagName)
+		}
```
jhump, line 443: This is the actual fix: an additional check so we can issue a good error message, instead of trying and failing to use reflection over HTTP 1.1.

jhump: Is the original message unclear? The message I used is similar to others in this function -- since gRPC is also not allowed over HTTP 1.1.

bufdev (approved):

PR #1966, opened 2023-03-31
Manifest.Range: iterate entries (path, digest)

unmultimedio:
Range function for the manifest entries, so we can reduce the appearances of `.Paths()` usages like:

```go
for _, path := range m.Paths() {
  digest, ok := m.DigestFor(path)
  if !ok {
    // this will never happen
    return nil, fmt.Errorf("digest not found for path: %s", path)
  }
  // do something with path and digest...
}
```

into:

```go
m.Range(func(path string, digest manifest.Digest) error {
  // do something with path and digest...
})
```

### DIFF
--- a/private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
+++ b/private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
@@ -68,7 +68,7 @@ func (c *casModuleCacher) GetModule(
 	}
 	var blobs []manifest.Blob
 	blobDigests := make(map[string]struct{})
-	if err := manifestFromCache.IteratePaths(func(path string, digest manifest.Digest) error {
+	if err := manifestFromCache.Range(func(path string, digest manifest.Digest) error {
 		if _, ok := blobDigests[digest.String()]; ok {
 			// We've already loaded this blob
 			return nil
@@ -118,7 +118,7 @@ func (c *casModuleCacher) PutModule(
 	moduleBasedir := normalpath.Join(modulePin.Remote(), modulePin.Owner(), modulePin.Repository())
 	// Write blobs
 	writtenDigests := make(map[string]struct{})
-	if err := moduleManifest.IteratePaths(func(path string, digest manifest.Digest) error {
+	if err := moduleManifest.Range(func(path string, digest manifest.Digest) error {
 		blobDigestStr := digest.String()
 		if _, ok := writtenDigests[blobDigestStr]; ok {
 			return nil
--- a/private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
+++ b/private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
@@ -68,20 +68,19 @@ func (c *casModuleCacher) GetModule(
 	}
 	var blobs []manifest.Blob
 	blobDigests := make(map[string]struct{})
-	for _, path := range manifestFromCache.Paths() {
-		digest, found := manifestFromCache.DigestFor(path)
-		if !found {
-			return nil, fmt.Errorf("digest not found for path: %s", path)
-		}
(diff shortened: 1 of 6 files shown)


private/pkg/manifest/module_test.go
```diff
@@ -324,3 +333,17 @@ func newBlobsArray(t *testing.T) []manifest.Blob {
 	}
 	return blobs
 }
+
+// assertBlobsAreEqual makes sure blobs digest are the same (assuming they're
+// correctly built), regardless the order in the passed arrays.
```
pkwarren, line 338:
```suggestion
// correctly built), ignoring order in the blobs arrays.
```

private/pkg/manifest/module_test.go
```diff
@@ -324,3 +333,17 @@ func newBlobsArray(t *testing.T) []manifest.Blob {
 	}
 	return blobs
 }
+
+// assertBlobsAreEqual makes sure blobs digest are the same (assuming they're
```
pkwarren, line 337:
```suggestion
// assertBlobsAreEqual makes sure the blobs digest is the same (assuming they're
```

private/pkg/manifest/manifest.go
```diff
@@ -139,6 +140,15 @@ func (m *Manifest) Paths() []string {
 	return paths
 }
 
+// IteratePaths invokes a function for all the paths in the manifest, passing
+// the path and its digest. The order in which the paths are iterated is not
+// guaranteed.
+func (m *Manifest) IteratePaths(f func(string, Digest)) {
```
pkwarren, line 146: The paths function doesn't look too expensive - why are we introducing this variant? Is it to return the pre-calculated digest for each path?

unmultimedio: Is not because of how expensive it is, but because we have some code paths in which we want the digests, not only the paths, so we can drop unnecessary presence checkers. I'll change a few of the code paths in this same PR as well to make it more clear.

private/bufpkg/bufmodule/bufmodulecache/cas_module_cacher.go
```diff
@@ -68,20 +68,19 @@ func (c *casModuleCacher) GetModule(
 	}
 	var blobs []manifest.Blob
 	blobDigests := make(map[string]struct{})
-	for _, path := range manifestFromCache.Paths() {
-		digest, found := manifestFromCache.DigestFor(path)
-		if !found {
-			return nil, fmt.Errorf("digest not found for path: %s", path)
-		}
```
unmultimedio, line 75: This is the check we're skipping with the new func.

private/pkg/manifest/manifest.go
```diff
@@ -139,6 +140,18 @@ func (m *Manifest) Paths() []string {
 	return paths
 }
 
+// Range invokes a function for all the paths in the manifest, passing the path
+// and its digest. The order in which the paths are iterated is not guaranteed.
+// This func will stop iterating if an error is returned.
+func (m *Manifest) Range(f func(path string, digest Digest) error) error {
```
unmultimedio, line 146: Renaming func to follow suggestion in [this comment](https://github.com/bufbuild/buf/pull/1724#discussion_r1069914158). cc @jhump

private/pkg/manifest/manifest.go
```diff
@@ -130,7 +130,8 @@ func (m *Manifest) AddEntry(path string, digest Digest) error {
 	return nil
 }
 
-// Paths returns all paths in the manifest.
+// Paths returns all paths in the manifest. If you want to iterate the paths and
+// their digests, consider using `IteratePaths` instead.
```
saquibmian, line 134:
```suggestion
// their digests, consider using `Range` instead.
```

pkwarren (approved):

Issue #1930, opened 2023-03-21
Print values for empty JSON fields in FileAnnotations


devkvlt:
### Problem

When we run `buf lint --error-format=json` on an empty proto file, we get the following JSON output:

```json
{
  "path": "path/to/file.proto",
  "type": "PACKAGE_DEFINED",
  "message": "Files must have a package defined."
}
{
  "path": "path/to/file.proto",
  "type": "SYNTAX_SPECIFIED",
  "message": "Files must have a syntax explicitly specified. If no syntax is specified, the file defaults to \"proto2\"."
}
```

This output is missing 4 fields `start_line`, `start_column`, `end_line` and `end_column` which makes consuming this JSON by third party tools problematic as one may easily forget to check if these fields are not null before assigning them as numeric values. 

### Proposition

Add the missing fields with value 1 so that the output becomes:

```json
{
  "path": "path/to/file.proto",
  "start_line": 1,
  "start_column": 1,
  "end_line": 1,
  "end_column": 1,
  "type": "PACKAGE_DEFINED",
  "message": "Files must have a package defined."
}
{
  "path": "path/to/file.proto",
  "start_line": 1,
  "start_column": 1,
  "end_line": 1,
  "end_column": 1,
  "type": "SYNTAX_SPECIFIED",
  "message": "Files must have a syntax explicitly specified. If no syntax is specified, the file defaults to \"proto2\"."
}
```

bufdev:
I can see the logic here, but there's also the inverse logic of this: there is no line 1 for an empty file, so you are indicating something that doesn't exist. You could make the argument that we already do this for the text format though, where we print `1:1`, so changing this would make this consistent. Perhaps omitting empty values would be better, but we'd want to make sure we do this consistently across all of our JSON outputs from `buf`. What specific third party tool are you referencing that makes this difficult?

@akshayjshah @timostamm interested if you have any input as well.

For reference, if we do do this: We likely just want to change `printFileAnnotationAsJSON` in `private/bufpkg/bufanalysis/print.go`.

bufdev:
I'm wondering if we should though - the more I think about it, it's inconsistent with the text format (which prints 1:1).

I think I might be coming around to the argument presented in the description of this issue, but am worried this may be a breaking change, although it's in a gray area.

timostamm:
For JSON, I can totally see myself looking at the output and writing code that ingests it, only to find out later about unknown locations. 

[...]

PR #1970, opened 2023-04-03
Output at least 1 for line and column information for JSON and MSVS

bufdev:
Fixes https://github.com/bufbuild/buf/issues/1930

### DIFF
--- a/private/bufpkg/bufanalysis/print.go
+++ b/private/bufpkg/bufanalysis/print.go
@@ -213,25 +213,12 @@ func newExternalFileAnnotation(f FileAnnotation) externalFileAnnotation {
 	if f.FileInfo() != nil {
 		path = f.FileInfo().ExternalPath()
 	}
-	startLine := f.StartLine()
-	startColumn := f.StartColumn()
-	endLine := f.EndLine()
-	endColumn := f.EndColumn()
-	// If path is empty, we don't output anything
-	// This may just be exacerbating the problem in https://github.com/bufbuild/buf/issues/1930
-	// but it seems right, but not very reasoned
-	if path != "" {
-		startLine = atLeast1(startLine)
-		startColumn = atLeast1(startColumn)
-		endLine = atLeast1(endLine)
-		endColumn = atLeast1(endColumn)
-	}
 	return externalFileAnnotation{
 		Path:        path,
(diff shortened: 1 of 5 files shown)


private/bufpkg/bufanalysis/print.go
```diff
@@ -245,3 +255,10 @@ func printEachAnnotationOnNewLine(
 	}
 	return nil
 }
+
+func atLeast1(i int) int {
```
saquibmian, line 259: nit: you can use this in [fileAnnotation#String](https://github.com/bufbuild/buf/blob/main/private/bufpkg/bufanalysis/file_annotation.go#L91-L96) as well.

private/bufpkg/bufanalysis/print.go
```diff
@@ -216,12 +213,25 @@ func newExternalFileAnnotation(f FileAnnotation) externalFileAnnotation {
...
 		path = f.FileInfo().ExternalPath()
 	}
+	startLine := f.StartLine()
+	startColumn := f.StartColumn()
+	endLine := f.EndLine()
+	endColumn := f.EndColumn()
+	// If path is empty, we don't output anything
+	// This may just be exacerbating the problem in https://github.com/bufbuild/buf/issues/1930
+	// but it seems right, but not very reasoned
+	if path != "" {
```
saquibmian, line 223: This would be true if `FileInfo#ExternalPath()` returns `""`. Looking at the implementations, it doesn't seem like this should be the case as `ExternalPath` falls back to `Path`, which is always normalized, validated, and non-empty. Not to mention, the resulting annotation is also useless. I.e., if `path` is empty here, we have a bug. 

Either way, we don't need to guard this block.

private/bufpkg/bufanalysis/print.go
bufdev: done

private/bufpkg/bufanalysis/print.go
bufdev: removed

bufdev: generally, i want to do defensive programming in these scenarios, but normalized paths are a case in the codebase where we assume standard conventions

saquibmian (approved):

PR #1973, opened 2023-04-04
Add buf beta stats and delete buf alpha stats

bufdev:
Having a `stats` command that basically does `price` but locally is actually useful at least for me, we might as well expose it here and see if people use it.


private/buf/cmd/buf/command/beta/stats/stats.go
```diff
@@ -0,0 +1,141 @@
...
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <source>",
+		Short: "Get statistics a given source or module",
```
saquibmian, line 46:
```suggestion
		Short: "Get statistics for a given source or module",
```

saquibmian (commented):
Looks good, this is definitely more useful than the previous form. One small typo.

saquibmian (approved):

Issue #1906, opened 2023-03-10
Buf build reports adding "optional" as breaking change


mikhainin:
We've got the following
```shell
$ cat cat proto/antispamtextcheck.proto
syntax = "proto3";

package com.bumble.contract.service;

message CheckTextRequest {
  int64 collective_id = 2; // we will make it "optional int64 collective_id = 2;" later
}
```

```
buf build -o old.bin proto/
```

If we add "optional" we will get a really confusing error:
```shell
buf breaking --against old.bin proto
proto/antispamtextcheck.proto:6:3:Field "2" on message "CheckTextRequest" moved from outside to inside a oneof.
```

According to protobuf folks, adding "oneof"  is fine (link: https://github.com/protocolbuffers/protobuf/issues/10822)

However, if we consider this as indeed a breaking change, it looks like we can detect that this is not "oneof" (which isn't clear for users) but rather the usage of field presence. So, we can change the error message to make it clear. e.g.

>  Field "2" on message "CheckTextRequest" became "optional" which internally moves it inside a oneof

If possible, it would be great if we could make this particular check disableable, because in our case this is absolutely fine

bufdev:
Oneofs and breaking changes are a complicated subject: https://protobuf.dev/programming-guides/proto3/#updating

In `v1` of the buf breaking rules (which we are committed to not breaking), we took the simple approach and said that any changes around `oneofs` were both breaking for generated sources and wire compatibility. In reality, there is a bit more nuance around wire compatibility, which in a `v2`, we would look to clean up.

I agree that the error message could be better, and the breaking change rules that produce errors should be cleaned up here. I've attempted to address this in https://github.com/bufbuild/buf/pull/1978

bufdev:
Closing this as a "duplicate" of https://github.com/bufbuild/buf/issues/2393 - subscribe to that issue to follow along on our progress, they're both the same issue effectively.

PR #1978, opened 2023-04-06
Handle proto3 optionals better for breaking changes

bufdev:
Fixes https://github.com/bufbuild/buf/issues/1906.

This suppresses a few breaking change errors that were produced when they shouldn't be, and updates the message for when oneof changes involve proto3 optionals.

### DIFF
--- a/private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
+++ b/private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
@@ -291,11 +291,17 @@ func checkFieldSameOneof(add addFunc, corpus *corpus, previousField protosource.
 	if !previousInsideOneof && !insideOneof {
 		return nil
 	}
+	previousProto3Optional := previousOneof != nil && previousOneof.IsProto3OptionalSyntheticOneof()
+	proto3Optional := oneof != nil && oneof.IsProto3OptionalSyntheticOneof()
 	if previousInsideOneof && insideOneof {
 		if previousOneof.Name() != oneof.Name() {
-			// otherwise prints as hex
-			numberString := strconv.FormatInt(int64(field.Number()), 10)
-			add(field, nil, field.Location(), `Field %q on message %q moved from oneof %q to oneof %q.`, numberString, field.Message().Name(), previousOneof.Name(), oneof.Name())
+			// If we are inside a oneof, then this means that we renamed the field, which is
+			// a different check.
+			if !previousProto3Optional && !proto3Optional {
+				// otherwise prints as hex
+				numberString := strconv.FormatInt(int64(field.Number()), 10)
+				add(field, nil, field.Location(), `Field %q on message %q moved from oneof %q to oneof %q.`, numberString, field.Message().Name(), previousOneof.Name(), oneof.Name())
+			}
 		}
 		return nil
 	}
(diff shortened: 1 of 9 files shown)


private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -291,11 +291,17 @@ func checkFieldSameOneof(add addFunc, corpus *corpus, previousField protosource.
...
+	proto3Optional := oneof != nil && oneof.IsProto3OptionalSyntheticOneof()
 	if previousInsideOneof && insideOneof {
 		if previousOneof.Name() != oneof.Name() {
-			// otherwise prints as hex
-			numberString := strconv.FormatInt(int64(field.Number()), 10)
-			add(field, nil, field.Location(), `Field %q on message %q moved from oneof %q to oneof %q.`, numberString, field.Message().Name(), previousOneof.Name(), oneof.Name())
+			// If we are inside a oneof, then this means that we renamed the field, which is
+			// a different check.
+			if !previousProto3Optional && !proto3Optional {
+				// otherwise prints as hex
```
saquibmian, line 301: What is this comment? I know it came from the original source, but I don't know what it means.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -783,9 +805,14 @@ func checkOneofNoDelete(add addFunc, corpus *corpus, previousMessage protosource
 	if err != nil {
 		return err
 	}
-	for previousName := range previousNameToOneof {
+	for previousName, previousOneof := range previousNameToOneof {
 		if _, ok := nameToOneof[previousName]; !ok {
-			add(message, nil, message.Location(), `Previously present oneof %q on message %q was deleted.`, previousName, message.Name())
+			// If this is a synthetic oneof, deleting of the oneof does not signal that a oneof was deleted,
+			// instead it signals that the field was renamed or changed to optional, which are done in
```
saquibmian, line 811: Did you mean changed to non-optional?

bufdev: This `if` statement says that if previously we were non-optional and now are optional, so this comment is right, changed to optional

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -783,9 +805,14 @@ func checkOneofNoDelete(add addFunc, corpus *corpus, previousMessage protosource
...
+			// different checks.
+			if !previousOneof.IsProto3OptionalSyntheticOneof() {
```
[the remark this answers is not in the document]
bufdev: We could in theory refactor `protosource` to special-case proto3 optionals, and not produce oneofs for them, but that's a big change to what `protosource` is - `protosource` was meant to be a wrapper to `FileDescriptorSets` largely, but with easily-gettable location information and some pointers back up through nested types (fields, nested messages, etc). If you decided to special-case proto3 optionals in `protosource`, you'd also have to special-case maps, which are also syntactic sugar on top of repeated fields. I think this level of change is out of scope.

The above isn't actually a breaking change for `WIRE` or `WIRE_JSON` funny enough, but I guess it is for `FILE` and `PACKAGE`, that's annoying. Will need to ponder what to do with this.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
saquibmian:
```
If this is a synthetic oneof [...] the field was [...] changed to optional.
```
A synthetic oneof _is_ an optional. 

Additionally, if we were previously non-optional, then we'll fall into the `if` and will report an annotation, whereas I think your comment was meant to capture the cases that should _not_ fall into the `if`.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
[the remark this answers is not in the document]
saquibmian: Yeah, I figured there'd be a wider-reaching impact. I'm cool with filtering out synthetic oneofs like you're doing here.

I think we can catch the specific escape by doing a check that a field's `proto3Optional` value doesn't change, the same way we check labels?

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
saquibmian: ~~`field.Number()` is an `int`, why would `%d` turn it into a hex?~~
Figured it out, it's the `%q` that's interpreting it as a character-literal. Could just do `"%d"`, but I digress.

PR #1987, opened 2023-04-10
Update --type flag docs

bufdev:
Per @jhump, packages and methods are supported now. We could do more to clean up this flag doc, but this at least makes it accurate.

### DIFF
--- a/private/buf/cmd/buf/command/generate/generate.go
+++ b/private/buf/cmd/buf/command/generate/generate.go
@@ -275,7 +275,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		&f.TypesDeprecated,
 		typeDeprecatedFlagName,
 		nil,
-		"The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types. Flag usage overrides buf.gen.yaml",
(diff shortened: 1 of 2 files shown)


private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -275,7 +275,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		&f.TypesDeprecated,
 		typeDeprecatedFlagName,
 		nil,
-		"The types (message, enum, service) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types. Flag usage overrides buf.gen.yaml",
+		"The types (package, message, enum, service, method) that should be included in this image. When specified, the resulting image will only include descriptors to describe the requested types. Flag usage overrides buf.gen.yaml",
```
saquibmian, line 278: Can you also update the description for `typeFlagName`?

jhump: I pushed a change that includes this, and that also updates the flag of the same name (and same function) for `buf build`.

jhump (approved):

saquibmian (approved):

PR #1991, opened 2023-04-12
Fix error when installing plugin from zip file

pkwarren:
When checking for errors, we should have used errors.Is against storage.ErrClosed instead of os.ErrClosed. The image is closed automatically when it is uploaded via the http.Client, so we need to properly guard against double close.

Additionally, add some context to errors so we can easily identify where they occur. The previous logic would just print "already closed" which made it difficult to find the call site where the error was occurring.


private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
```diff
@@ -228,8 +230,8 @@ func run(
 			return err
 		}
 		defer func() {
-			if err := image.Close(); !errors.Is(err, os.ErrClosed) {
-				retErr = multierr.Append(retErr, err)
+			if err := image.Close(); err != nil && !errors.Is(err, storage.ErrClosed) {
```
pkwarren, line 233: This was the fix for the `already closed` error I encountered with local testing.

pkwarren:
It appears this regressed when https://github.com/bufbuild/buf/blame/c1deabb296b24c62c4aeb1a6341e9f649c181c8a/private/pkg/storage/storageos/bucket.go#L412 was updated to use `errors.Is` instead of `==` which I still believe is the proper fix.

private/buf/cmd/buf/command/alpha/plugin/pluginpush/pluginpush.go
saquibmian: Any way to test for this?

pkwarren: Not easily without wiring up a fair amount of supporting services. To run the CLI it would need to spin up the dependent BSR services and also a Docker client. This code path is less used now that we have implemented plugin sync in the BSR.

saquibmian (approved):

PR #1996, opened 2023-04-13
Digests() func in manifest to export all unique digests

unmultimedio:
Includes better usage of `Range` when all we want to do is iterate over unique digests.


pgmitche (approved):

twilly (approved):
:shipit:

PR #1998, opened 2023-04-14
Update Maven package version as part of release

pkwarren:
We have a Maven package which ships the Buf CLI (that is in turn used by the buf-gradle-plugin). When releasing a new version of Buf, we should call a workflow to open a PR to update the version of the Maven package as well.


buildbreaker (approved):
Nice!

PR #2001, opened 2023-04-14
Allow preserving java_multiple_files setting

pkwarren:
Currently, the JavaMultipleFiles image modifier will skip setting the value if it already matches the desired value (with `GetJavaMultipleFiles()`), however in some cases we want to preserve any settings specified in the source .proto files, and only overwrite it if unset.

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -562,6 +562,7 @@ func newModifier(
 		sweeper,
 		javaMultipleFilesValue,
 		managedConfig.Override[bufimagemodify.JavaMultipleFilesID],
+		false, // preserveExistingValue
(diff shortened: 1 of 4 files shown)


private/buf/bufgen/generator.go
```diff
@@ -562,6 +562,7 @@ func newModifier(
 		sweeper,
 		javaMultipleFilesValue,
 		managedConfig.Override[bufimagemodify.JavaMultipleFilesID],
+		false, // preserveExistingValue
```
pkwarren, line 565: Not exposing this to `buf generate` ATM - planning to use in the BSR for remote packages.

bufdev (approved):

PR #2002, opened 2023-04-14
Allow preserving java_outer_classname setting

pkwarren:
In some cases, we need to preserve the user specified setting for `java_outer_classname`. This is useful to allow compilation of code that relies on specific values and gives users more control over generated code/remote packages.

### DIFF
--- a/private/bufpkg/bufimage/bufimagemodify/java_outer_classname.go
+++ b/private/bufpkg/bufimage/bufimagemodify/java_outer_classname.go
@@ -36,6 +36,7 @@ func javaOuterClassname(
 	logger *zap.Logger,
 	sweeper Sweeper,
 	overrides map[string]string,
+	preserveExistingValue bool,
 ) Modifier {
 	return ModifierFunc(
 		func(ctx context.Context, image bufimage.Image) error {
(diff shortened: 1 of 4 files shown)


private/bufpkg/bufimage/bufimagemodify/java_outer_classname.go
```diff
@@ -65,11 +66,20 @@ func javaOuterClassnameForFile(
 	sweeper Sweeper,
 	imageFile bufimage.ImageFile,
 	javaOuterClassnameValue string,
+	preserveExistingValue bool,
 ) error {
+	if isWellKnownType(ctx, imageFile) {
+		// The file is a well-known type - don't override the value.
```
bufdev, line 72: What about if someone has the WKTs vendored? They might want to override in this case (I know this is pre-existing logic, but wanted to bring it up)

bufdev (approved):

private/bufpkg/bufimage/bufimagemodify/java_outer_classname.go
pkwarren: Haven't thought about it but I'd guess most people would expect to resolve the files from protobuf-java/protobuf-javalite artifact, which uses whatever settings which are present in the files. In most cases they do use pascal case so overriding might be a no-op, but in a few cases they are pluralized which wouldn't match our algorithm:

```
google/protobuf/descriptor.proto
45:option java_outer_classname = "DescriptorProtos";
...
google/protobuf/compiler/plugin.proto
48:option java_outer_classname = "PluginProtos";
```

I think keeping this safeguard in place is probably a good thing.

PR #2008, opened 2023-04-17
Remove repository owner and name from `MoveLabel` rpc

doriable:
We should resolve the repository through the `LabelValue` specified
by the `to` field, since this is not optional.
The repository owner and name were added as an unnecessary precaution.
The implementation will also be updated.


proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -60,8 +60,6 @@ message MoveLabelRequest {
   LabelName label_name = 1;
   optional LabelValue from = 2;
   LabelValue to = 3;
-  string repository_owner = 4;
-  string repository_name = 5;
```
unmultimedio, line 64: Should these be moved to the `CreateLabelRequest` instead?

doriable: `CreateLabelRequest` also takes a `LabelValue` in the form of a commit, so we are actually resolving the repo through the commit in that RPC already.

unmultimedio: Correct, and I assume the implementation checks the authz permissions for the parent repo of the matched commit?

doriable: Yes, correct.

twilly (approved):
Let's give it a go.

PR #2010, opened 2023-04-18
Add support to "buf curl" for --user and --netrc credentials flags

### LINKED ISSUE #1997: curl: support -u/--user
`buf curl` currently supports `-H`, which is great, but a nice feature of `curl` itself is being able to do `-u username:password` without needing to manually curate your own `-H 'Authorization: Basic xxxx'` header.

It's definitely a convenience and possible to do this with `-H`, but the syntax of `-u` is nice.

jhump:
This adds support for `-u`/`--user`, `-n`/`--netrc`, and `--netrc-file` flags to `buf curl`. These flags work the same way in this context as the flags of the same name in `curl`.

Resolves #1997.

### DIFF
--- a/private/buf/cmd/buf/command/curl/curl.go
+++ b/private/buf/cmd/buf/command/curl/curl.go
@@ -15,8 +15,10 @@
 package curl
 
 import (
+	"bytes"
 	"context"
 	"crypto/tls"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"io"
@@ -33,13 +35,16 @@ import (
 	"github.com/bufbuild/buf/private/buf/buffetch"
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"github.com/bufbuild/buf/private/pkg/app"
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/buf/private/pkg/command"
+	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/bufbuild/buf/private/pkg/protoencoding"
 	"github.com/bufbuild/buf/private/pkg/stringutil"
 	"github.com/bufbuild/buf/private/pkg/verbose"
 	"github.com/bufbuild/connect-go"
+	netrcparser "github.com/jdxcode/netrc"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"go.uber.org/multierr"
@@ -74,6 +79,9 @@ const (
 
 	// Header and request body flags
 	userAgentFlagName = "user-agent"
+	userCredsFlagName = "user"
+	netrcFlagName     = "netrc"
+	netrcFileFlagName = "netrc-file"
 	headerFlagName    = "header"
 	dataFlagName      = "data"
 	outputFlagName    = "output"
@@ -202,6 +210,9 @@ type flags struct {
 
 	// Handling request and response data and metadata
 	UserAgent string
+	UserCreds string
+	UseNetrc  bool
+	NetrcFile string
 	Headers   []string
 	Data      string
 	Output    string
@@ -362,7 +373,40 @@ one is provided`,
 		userAgentFlagName,
 		"A",
 		"",
-		`The user agent string to send`,
+		`The user agent string to send. This is ignored if a -H or --header flag is provided
+that sets a header named 'User-Agent'.`,
+	)
+	flagSet.StringVarP(
+		&f.UserCreds,
+		userCredsFlagName,
+		"u",
+		"",
+		`The user credentials to send, via a basic authorization header. The value should be
+in the format "username:password". If the value has no colon, it is assumed to just be
+the username, in which case you will be prompted to enter a password. This overrides
+the use of a .netrc file. This is ignored if a -H or --header flag is provided that sets
+a header named 'Authorization'.`,
+	)
+	flagSet.BoolVarP(
+		&f.UseNetrc,
+		netrcFlagName,
+		"n",
+		false,
+		`If true, a file named .netrc in the user's home directory will be examined to find
+credentials for the request. The credentials will be sent via a basic authorization header.
+The command will fail if the file does not have an entry for the hostname in the URL. This
+flag is ignored if a -u or --user flag is present. This is ignored if a -H or --header flag
+is provided that sets a header named 'Authorization'.`,
+	)
+	flagSet.StringVar(
+		&f.NetrcFile,
+		netrcFileFlagName,
+		"",
+		`This is just like use -n or --netrc, except that the named file is used instead of
+a file named .netric in the user's home directory. If a --netrc-optional flag is present,
+it will be respected and the command won't fail if no corresponding entry is found in the
+file. This flag cannot be used with the -n or --netrc flag. This is ignored if a -H or
+--header flag is provided that sets a header named 'Authorization'.`,
 	)
 	flagSet.StringSliceVarP(
 		&f.Headers,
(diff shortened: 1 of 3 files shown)


private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -497,6 +545,117 @@ func (f *flags) validate(isSecure bool) error {
...
+		}
+	}
+	if _, err := os.Stat(netrcFile); err != nil {
+		if !strings.Contains(err.Error(), netrcFile) {
+			// make sure error message contains path to file
+			return "", fmt.Errorf("could not read file: %s: %w", netrcFile, err)
+		}
+		return "", fmt.Errorf("could not read file: %w", err)
+	}
+	// NB: can't use buf/private/pkg/netrc for this because it only supports reading from $HOME/.netrc
```
bufdev, line 587: We should just extend pkg/netrc then, it's meant to be a wrapper so we don't expose the external lib throughout the codebase

bufdev:
Thanks for doing this

private/buf/cmd/buf/command/curl/curl.go
jhump: Fair. And it was trivial to just expose an existing unexported method. Done.

private/pkg/netrc/netrc.go
```diff
@@ -87,7 +87,11 @@ func GetFilePath(envContainer app.EnvContainer) (string, error) {
 	return filepath.Join(homeDirPath, netrcFilename), nil
 }
 
-func getMachineForNameAndFilePath(name string, filePath string) (_ Machine, retErr error) {
+// GetMachineForNameAndFilePath returns the Machine for the given name from the
+// file at the given path.
+//
+// Returns nil if no such Machine.
```
jhump, line 93:
```suggestion
// Returns nil if no such Machine or no such file.
```

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -202,6 +210,9 @@ type flags struct {
 
 	// Handling request and response data and metadata
 	UserAgent string
+	UserCreds string
```
bufdev, line 213: nit: almost all the variables in `flags` across our commands are named after the flag in practice, so just name this `User`.

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -202,6 +210,9 @@ type flags struct {
...
+	UseNetrc  bool
```
bufdev, line 214: same nit: `Netrc`

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -362,7 +373,40 @@ one is provided`,
...
+	)
+	flagSet.StringVarP(
+		&f.UserCreds,
+		userCredsFlagName,
+		"u",
+		"",
+		`The user credentials to send, via a basic authorization header. The value should be
+in the format "username:password". If the value has no colon, it is assumed to just be
+the username, in which case you will be prompted to enter a password. This overrides
+the use of a .netrc file. This is ignored if a -H or --header flag is provided that sets
```
bufdev, line 387: nit: if we're being pedantic, you should use `fmt.Sprintf` and pass in `headerFlagShortName` and `headerFlagName` (also there looks like there's no `headerFlagShortName` right now)

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -362,7 +373,40 @@ one is provided`,
...
 		"A",
 		"",
-		`The user agent string to send`,
+		`The user agent string to send. This is ignored if a -H or --header flag is provided
+that sets a header named 'User-Agent'.`,
+	)
+	flagSet.StringVarP(
+		&f.UserCreds,
+		userCredsFlagName,
+		"u",
```
bufdev, line 382: `userFlagShortName`

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -74,6 +79,9 @@ const (
 
 	// Header and request body flags
 	userAgentFlagName = "user-agent"
+	userCredsFlagName = "user"
```
bufdev, line 82: `userFlagName`

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -362,7 +373,40 @@ one is provided`,
...
+credentials for the request. The credentials will be sent via a basic authorization header.
+The command will fail if the file does not have an entry for the hostname in the URL. This
+flag is ignored if a -u or --user flag is present. This is ignored if a -H or --header flag
+is provided that sets a header named 'Authorization'.`,
+	)
+	flagSet.StringVar(
+		&f.NetrcFile,
+		netrcFileFlagName,
+		"",
+		`This is just like use -n or --netrc, except that the named file is used instead of
```
bufdev, line 405: Same comments about fmt.Sprintf, although tbh really consider this a nit, we don't need to do it if it takes too much time

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -497,6 +545,125 @@ func (f *flags) validate(isSecure bool) error {
 	return nil
 }
 
+func (f *flags) determineCredentials(
+	ctx context.Context,
+	container interface {
```
bufdev, line 550: cute heh

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -497,6 +545,125 @@ func (f *flags) validate(isSecure bool) error {
...
+		app.Container
+		appverbose.Container
+	},
+	host string,
+) (string, error) {
+	if f.UserCreds != "" {
+		// this flag overrides any netrc-related flags
+		var username, password string
+		pos := strings.IndexByte(f.UserCreds, ':')
```
bufdev, line 559: nit: can use `strings.SplitN`

bufdev (commented):
Mostly nits, sorry for the double review

private/buf/cmd/buf/command/curl/curl.go
jhump: done

private/buf/cmd/buf/command/curl/curl.go
jhump: done

private/buf/cmd/buf/command/curl/curl.go
jhump: done

private/buf/cmd/buf/command/curl/curl.go
jhump: done

private/buf/cmd/buf/command/curl/curl.go
jhump: Done. I applied this everywhere... except the long paragraphs of content for the main help content. That references numerous flags, including many references in the examples. That... didn't seem worth it, especially if it meant moving the actual flag name references to far away from the format indicator. If Go had python- or kotlin-style interpolation, that would be another story, but...

private/buf/cmd/buf/command/curl/curl.go
jhump: Yeah, I guess I'm in a habit of `strings.Index` to avoid the slice allocation. But that's definitely premature optimization for this bit of code. Done.

private/buf/cmd/buf/command/curl/curl.go
jhump: done

bufdev (approved):

jhump:
@bufdev, FYI, I just pushed a commit to add a blurb about this change to the changelog.

bufdev (approved):

mattrobenolt:
I just realized this doesn't work as expected. It tags the `Authorization` header, but it's not sending `Basic` since this is basic auth. Not sure how this interacts with `--netrc` or what expectations are there, but strictly for `--user`, the intention is to send along as basic auth.

Right now, if I do:

```
$ buf curl -v -u foo:bar ...
```

I end up with a request such as:

```
buf: > (#1) Authorization: Zm9vOmJhcg==
```

And the expectation is this would be:

```
Authorization: Basic Zm9vOmJhcg==
```

Otherwise, the auth is invalid.

mattrobenolt:
PR to fix: https://github.com/bufbuild/buf/pull/2096

Feel free to implement this however, but this gets it working. :)

jhump:
@mattrobenolt 🤦 , not sure how I missed this. I thought I had tried this out with real credentials to a BSR to verify, but maybe I imagined that. I know I did quite a lot of `buf curl -v` and inspecting the headers. I am embarrassed a bug this bad made it into a release. Thank you for the PR.

mattrobenolt:
haha @jhump, no worries. Happens to us all. 🫶

PR #2011, opened 2023-04-18
Add audit logging definitions for labels

doriable:
This adds definitions to our audit logging to support audit logs for
creating and moving labels.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -369,3 +374,35 @@ message PayloadRepositoryCommitDeleted {
...
+  // owner_id is the id of the owner of the repository on which the label will be created.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the repository on which the label will be created.
+  string owner_name = 2;
+  // repository_id is the id of the repository on which the label will be created.
+  string repository_id = 3;
+  // repository_name is the name of the repository from which the label will be created.
+  string repository_name = 4;
+  // label_namespace is the namespace of the label that was created.
+  string label_namespace = 5;
```
unmultimedio, line 388: Oh ok so you leave the namespace in a different field in the payload.

unmultimedio (changes requested):

proto/buf/alpha/audit/v1alpha1/event.proto
doriable: Yeah -- I prefer not to have them in the same field, and earlier, I had label name here as well, but I think the redundant data is not helpful.
Even though they are separate, I think from an audit log perspective, this is pretty reasonable.

unmultimedio (approved):

PR #2012, opened 2023-04-19
Support background processes with command.Runner

twilly:
This adds a new method to command.Runner: `Start`. This method is to support exec.Commands that run in the background, which is to say exec.Cmd.Start() would have been used over our previous uses of exec.Cmd.Run(). Much like exec.Cmd's Start method, command.Runner's Start does not block. Instead it returns a process handle that the caller can use to terminate the process at a later time.

Ref: BSR-1744


doriable (approved):
This seems very reasonable to me :)

private/pkg/command/runner.go
```diff
@@ -46,32 +48,26 @@ func newRunner(options ...RunnerOption) *runner {
...
 	r.semaphoreC <- struct{}{}
 	err := cmd.Run()
 	<-r.semaphoreC
 	return err
 }
 
+func (r *runner) Start(name string, options ...RunOption) (Process, error) {
+	runOptions := newRunOptions(options...)
+	cmd := exec.Command(name, runOptions.args...)
+	runOptions.Apply(cmd)
```
bufdev, line 63: This function ignores all parallelism that we set - one of the goals of `Runners` is to limit the number of concurrent tasks that can be sent off. If a caller sets `RunnerWithParallelism`, that should be respected. You'll have to rearchitect this function to accomplish this. You'll have to +1 the semaphore before you start the command, start the command, then kick of a goroutine to handle `cmd.Process.Wait`. The `terminator.Terminate` functionality will need to be rearchitected to handle the results of this `Wait` Once `Wait` returns, the semaphore needs to be -1'ed.

private/pkg/command/runner.go
```diff
@@ -96,3 +114,31 @@ func envSlice(env map[string]string) []string {
 	sort.Strings(environ)
 	return environ
 }
+
+type terminator struct {
```
bufdev, line 118: Why is this called `terminator` when the interface is called `process`?

Should be moved to a separate file `process.go` and renamed to `process` per style guide.

private/pkg/command/runner.go
```diff
@@ -84,8 +80,30 @@ type runOptions struct {
...
+	if len(runOptions.env) == 0 {
+		runOptions.env = emptyEnv
+	}
+	if runOptions.stdin == nil {
+		runOptions.stdin = ioextended.DiscardReader
+	}
+	return runOptions
+}
+
+func (r *runOptions) Apply(cmd *exec.Cmd) {
```
bufdev, line 97: No need for this change after splitting `RunOption` and `StartOption`.

bufdev (changes requested):

private/pkg/command/command.go
```diff
@@ -30,10 +43,15 @@ import (
...
-	// Run runs the external command.
+	// Run runs the external command. It blocks until the command exits.
 	//
 	// This should be used instead of exec.CommandContext(...).Run().
 	Run(ctx context.Context, name string, options ...RunOption) error
+
+	// Start runs the external command, returning a [Process] handle.
+	//
+	// This should be used instead of exec.Command(...).Start().
+	Start(name string, options ...RunOption) (Process, error)
```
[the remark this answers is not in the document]
twilly: The options are not specific to Start because all options map to exec.Cmd configuration. That's not just duplication of an option or two but a straight up duplication of the whole command package.

What about renaming RunOption to ExecOption? It's a little clearer and isn't tied to the name of the function.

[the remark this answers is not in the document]
bufdev: sure sgtm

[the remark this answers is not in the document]
twilly: I decided to take it a step further to better reflect what's going on with the difference between Run and Start. It's a but more like exec.Cmd in feel too, which is to say a command is prepared and then ran in two pieces.

twilly:
What I just pushed (ef89699c1d4e4f) is the refactoring that looks more similar to using exec.Command: a command is prepared and then it is ran. It also resolves the options style issue (two methods that want to share identical options controlling exec.Command).

I also dropped signaling to terminate a command. I tested what I really wanted to do more, which is call git cat-file, and it's pretty well behaved. Closing it's standard input will have it exit cleanly.

command.Run now looks exactly how exec.Cmd.Run is implemented: Start and Wait.

exec.CommandContext isn't called anymore. Instead Process handles the context cancellation itself.

There's not testing at all for this package. I may give that a go shortly.

private/pkg/command/process.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+func (p *process) Wait(ctx context.Context) error {
+	// There will not be a second call to wait. Always decrement process
+	// counters when we return.
+	if p.terminated {
+		return errors.New("process already terminated")
+	}
+	p.terminated = true
+	defer p.runner.decrement()
```
bufdev, line 60: If a caller doesn't call Wait, Runner will be in deadlock. The semaphore needs to be decremented regardless of the call to Wait

bufdev (commented):
The Run API as designed was meant to be foolproof, I don't think we need it refactored here, but see comment

private/pkg/command/process.go
twilly: Calling Run calls Wait. It's still the same foolproof execution and collection of a ephemeral process.

The caller manages to the call to Wait if they use Start. There's no way around that because the use case is to execute a process that lives beyond the calling function's return.

So you're right, if the caller doesn't collect the process with Wait, then the process remains accounted for. That's expected behavior.

bufdev: In the current implementation, a user is required to call `Process.Wait`. A user could call `Process.Start`, throw away the `Process` object, and never call `Process.Wait`, which will cause a deadlock as more and more `Process.Start` calls are made. This isn't acceptable - a user should not have to call `Process.Wait` for a deadlock not to occur.

You can deal with this by starting a goroutine in `Process.Start` that watches on `*exec.Cmd.Wait`, and decrements when it returns.

The goal of this package is to make it so that users are unable to mess up - this package, as currently implemented, lets users cause a major issue if they do not correctly use `Process`.

bufdev: Additionally, a user could call `Process.Run` and `Process.Start` on the same `Process` object, whereas since there was no stateful object returned in the previous API, it was not possible to misuse any returned stateful object. The API as it stands leaves more room for error for a critical function

twilly: I can setup a watcher routine as it will help in some cases of a forgotten close. It may produce some other odd behaviors (e.g. getting io.ErrClosedPipe) but the caller was broken anyway.

Managing a process is fundamentally stateful (hello halting problem, how I missed you). It's a problem even with the current command package: all you have to do is pass a background context (which are the majority being passed around) and leave open a stdin pipe if the program takes one. In fact, using a background context and calling Run in a goroutine was one of the ways I was going to create a long-lived process, but that felt a little hacky. I can still pursue that route if having command return a process handle of any kind feels worse.

For reference on why I'm looking at this kind of functionality, check out the latest draft PR for building modules directly from bare git storage: #2018

bufdev: We must limit the number of concurrent background processes, and any solution where the user has to call `Process.Wait` to prevent a potentially-system-side deadlock isn't acceptable

twilly:
Peter and I chatted a bit just to clear a few things. Here's the result: sticking with Run and Start sharing ExecOptions and the process handle spawning a goroutine to watch the process.

There still could be resources getting stuck if the user calls Start and not Wait if they also fail to close the process. At least if they try to close the process through other means and fail to call Wait, then our counter does get correctly decremented.

private/pkg/command/process.go
```diff
@@ -0,0 +1,64 @@
...
+	"context"
+	"os/exec"
+
+	"go.uber.org/multierr"
+)
+
+type process struct {
+	cmd          *exec.Cmd
+	doneCallback func()
+	wait         chan error
```
bufdev, line 27: This is an unbuffered channel, it will block unless there is a receiver already installed, meaning that the call in `monitor` will block unless someone has called wait. This has the same issue as  before - a potential deadlock unless the user calls `Wait`. This needs to be a buffered channel.

private/pkg/command/process.go
```diff
@@ -0,0 +1,64 @@
...
+	p.doneCallback()
+}
+
+func (p *process) Wait(ctx context.Context) error {
+	select {
+	case err, ok := <-p.wait:
+		// Process exited
+		if ok {
+			p.procErr = err
+			close(p.wait)
```
bufdev, line 53: If a user double-calls `Wait`, the program will panic. It should be documented that `Wait` can be called exactly once, and if it is called twice, an error should result and not a panic: https://cs.opensource.google/go/go/+/refs/tags/go1.20.3:src/os/exec/exec.go;l=882

private/pkg/command/process.go
twilly: Whoops, that came from a refactor where I passed error into the callback at first (doesn't seem useful, also doesn't encourage actually calling Wait).

I'll capture err and callback first. Sticks with style better. Also I'll add testing to actually cover the complete behavior of this.

private/pkg/command/process.go
twilly: The select won't panic as the receiver is in the <value>, <success> := form. It's here so we can tell when a call is double called rather than block forever or panic.

I'll make it so it'll return a specific error rather than capturing the exit code. The capture seemed useful but a permanent error is fine with me.

private/pkg/command/process.go
bufdev: The channel is still unbuffered - the  implementation as it currently is will still leave zombie goroutines

bufdev:
I/we can't afford more time on back and forth on this one, so I've just fixed the code so it doesn't produce zombie goroutines, and I've also made it so there are separate `RunOption` and `StartOption` types so there's no migration.

bufdev (approved):

PR #2014, opened 2023-04-20
Update to connect-go v1.7.0

pkwarren:
Update to the latest version of connect and add idempotency_level option to all RPCs that are safe for the BSR to be called with a GET.


proto/buf/alpha/registry/v1alpha1/convert.proto
```diff
@@ -29,7 +29,9 @@ enum ConvertFormat {
 service ConvertService {
   // Convert converts a serialized message according to
   // the provided type name using an image.
-  rpc Convert(ConvertRequest) returns (ConvertResponse);
+  rpc Convert(ConvertRequest) returns (ConvertResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
```
mfridman, line 33: Maybe not this one?

(this is a drive-by, mainly marked all the ones that take an Image, does it make sense for this to support anything other than POST, maybe .. not sure).

proto/buf/alpha/registry/v1alpha1/generate.proto
```diff
@@ -65,10 +65,14 @@ service GenerateService {
   // GeneratePlugins generates an array of files given the provided
   // module reference and plugin version and option tuples. No attempt
   // is made at merging insertion points.
-  rpc GeneratePlugins(GeneratePluginsRequest) returns (GeneratePluginsResponse);
+  rpc GeneratePlugins(GeneratePluginsRequest) returns (GeneratePluginsResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
```
mfridman, line 69: Maybe not this one?

proto/buf/alpha/registry/v1alpha1/schema.proto
```diff
@@ -23,11 +23,15 @@ import "google/protobuf/descriptor.proto";
...
-  rpc GetSchema(GetSchemaRequest) returns (GetSchemaResponse);
+  rpc GetSchema(GetSchemaRequest) returns (GetSchemaResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
 
   // ConvertMessage allows the caller to convert a given message data blob from
   // one format to another by referring to a type schema for the blob.
-  rpc ConvertMessage(ConvertMessageRequest) returns (ConvertMessageResponse);
+  rpc ConvertMessage(ConvertMessageRequest) returns (ConvertMessageResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
```
mfridman, line 33: Maybe not this one?

proto/buf/alpha/registry/v1alpha1/convert.proto
pkwarren: I think it just indicates they have no side effects, so they could be called as a GET. However, you might be right that we should leave off the bigger methods since there are limits to query parameters for most requests so even if we **could** send them as a GET, we probably shouldn't as we'd have to retry as a POST.

akshayjshah: FYI: we debated exposing an option to cap the size of GETs, and eventually decided not to export it in this release. It's still in the `connect-go` codebase and is fully tested, so we can easily export it if we need it.

akshayjshah:
If you're doing this, it's potentially also valuable to mark RPCs that are idempotent but _not_ side effect free (e.g., deletes).

I'd eventually like really good retry interceptors that work with these options.

proto/buf/alpha/registry/v1alpha1/convert.proto
pkwarren: I think that would be very useful, but we might also want to tweak the guidance for when to use this feature to recommend it be used with methods that have smaller request messages that stay way below the limits.

pkwarren:
> If you're doing this, it's potentially also valuable to mark RPCs that are idempotent but _not_ side effect free (e.g., deletes).

Mind if I do that as a follow up? I'd feel more comfortable doing that once I've examined the implementation of the RPCs to make sure we're safe.

mfridman (approved):

PR #2016, opened 2023-04-21
Add Swift registry plugin configuration

pgmitche:


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +158,43 @@ message MavenConfig {
...
+    // The minimum compatible platform versions of the runtime library.
+    message Platforms {
+      // The minimum compatible version of the MacOS platform.
+      string macos = 1;
+      // The minimum compatible version of the iOS platform.
+      string ios = 2;
+      // The minimum compatible version of the tvOS platform.
+      string tvos = 3;
+      // The minimum compatible version of the watchOS platform.
+      string watchos = 4;
```
stefanvanburen, line 179: where do we pull these values from? what happens when they're not set?

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +158,43 @@ message MavenConfig {
...
 
+message SwiftConfig {
+  // RuntimeLibrary describes the runtime library dependency of the generated code.
+  message RuntimeLibrary {
+    // The name of the runtime library package, github.com/apple/swift-protobuf.
+    string package = 1;
+    // The version of the runtime library, i.e. "1.21.0"
+    string version = 2;
+    // The products exported by the runtime library.
+    repeated string products = 3;
```
stefanvanburen, line 169: I vaguely understand why we only need the names of the products here, but should we future-proof ourselves a bit and make this a `message Product` with a `string name = 1` and then make this `repeated Product products`? wdyt?

It would make me feel better if we needed to encode more into these in the future to have it be extensible - looking at something like https://github.com/apple/swift-protobuf/blob/main/Package.swift#L16-L33

Lastly, do we need this field to be set (should we validate it as non-empty)?

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -239,10 +244,59 @@ func ProtoRegistryConfigToPluginRegistry(config *registryv1alpha1.RegistryConfig
...
 }
 
+func ProtoSwiftConfigToSwiftRegistryConfig(protoSwiftConfig *registryv1alpha1.SwiftConfig) (*bufpluginconfig.SwiftRegistryConfig, error) {
+	swiftConfig := &bufpluginconfig.SwiftRegistryConfig{}
+	runtimeLibs := protoSwiftConfig.GetRuntimeLibraries()
+	if runtimeLibs != nil {
+		swiftConfig.Deps = make([]bufpluginconfig.SwiftRegistryDependencyConfig, 0, len(runtimeLibs))
+		for _, runtimeLib := range runtimeLibs {
+			swiftConfig.Deps = append(swiftConfig.Deps, bufpluginconfig.SwiftRegistryDependencyConfig{
+				Module:   runtimeLib.GetPackage(),
```
stefanvanburen, line 264: It seems like these fields should line up, i.e. either it should be `package` or `module`, not both?

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -128,6 +130,15 @@ func newRegistryConfig(externalRegistryConfig ExternalRegistryConfig) (*Registry
...
+	} else if !isSwiftEmpty {
+		swiftRegistryConfig, err := newSwiftRegistryConfig(externalRegistryConfig.Swift)
+		if err != nil {
+			return nil, err
+		}
+		return &RegistryConfig{
+			Swift:   swiftRegistryConfig,
+			Options: options,
+		}, nil
 	}
```
stefanvanburen, line 142: hehe, we might want to switch this to a real `switch` at this point?

stefanvanburen (commented):
mostly questions, but I might also rope in someone slightly more familiar with the CLI side / defining the shape of these registry protos (@pkwarren?)

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
pgmitche: That's a great idea 🙏🏼 

This field doesn't need to be set. If the plugin source packages don't specify any platform version information, it can be omitted.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
pgmitche: We would need to specify them in the buf.plugin.yaml and source the information from any source plugin's own Package.swift manifest file. For example: https://github.com/bufbuild/connect-swift/blob/main/Package.swift#L21-L25

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -239,10 +244,78 @@ func ProtoRegistryConfigToPluginRegistry(config *registryv1alpha1.RegistryConfig
...
+func SwiftRegistryConfigToProtoSwiftConfig(swiftConfig *bufpluginconfig.SwiftRegistryConfig) *registryv1alpha1.SwiftConfig {
+	protoSwiftConfig := &registryv1alpha1.SwiftConfig{}
+	if swiftConfig.Dependencies != nil {
+		protoSwiftConfig.RuntimeLibraries = make([]*registryv1alpha1.SwiftConfig_RuntimeLibrary, 0, len(swiftConfig.Dependencies))
+		for _, dependency := range swiftConfig.Dependencies {
+			protoSwiftConfig.RuntimeLibraries = append(protoSwiftConfig.RuntimeLibraries, &registryv1alpha1.SwiftConfig_RuntimeLibrary{
+				Package:  dependency.Package,
+				Version:  dependency.Version,
+				Products: dependency.Products,
+				Platforms: []*registryv1alpha1.SwiftConfig_RuntimeLibrary_Platform{
```
pkwarren, line 298: This doesn't feel 1-1 with the conversion from a protobuf type. Don't we need to check each of the dependency.Platforms.* values and only add a platform entry if the version is non-empty?

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
```diff
@@ -192,6 +193,36 @@ type MavenRuntimeConfig struct {
...
+// SwiftRegistryDependencyPlatformConfig is the swift registry dependency platform configuration.
+type SwiftRegistryDependencyPlatformConfig struct {
+	// macOS specifies the version of the macOS platform.
+	MacOS string
+	// iOS specifies the version of the iOS platform.
+	IOS string
+	// TVOS specifies the version of the tvOS platform.
+	TVOS string
+	// WatchOS specifies the version of the watchOS platform.
+	WatchOS string
```
pkwarren, line 223: How stable is this list? Ideally we wouldn't need to update our CLI + schema + other systems to support new platform types.

pkwarren (commented):
Looks good - just a couple of comments.

private/bufpkg/bufplugin/bufpluginconfig/bufpluginconfig.go
pgmitche: @pkwarren as stable as apple's list of product offerings I suppose. I suppose this has only really been updated once this decade with the launch of watchOS.

private/bufpkg/bufplugin/bufpluginconfig/testdata/success/swift/buf.plugin.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - swift
+registry:
+  swift:
+    deps:
+      - source: github.com/apple/swift-protobuf
+        package: swift-protobuf
+        products: [SwiftProtobuf]
+        version: 1.21.0
+        platforms:
+          macos: v10_15
```
rebello95, line 18: Might be good to specify the other platforms here as well to validate

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +167,41 @@ message MavenConfig {
   repeated RuntimeConfig additional_runtimes = 3;
 }
 
+message SwiftConfig {
+  // RuntimeLibrary describes the runtime library dependency of the generated code.
```
rebello95, line 171:
```suggestion
  // RuntimeLibrary describes a runtime library dependency of the generated code.
```

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +167,41 @@ message MavenConfig {
...
+  message RuntimeLibrary {
+    // The source of the runtime library package, github.com/apple/swift-protobuf.
```
rebello95, line 173: Should we say this is an example, like the others below?

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +167,41 @@ message MavenConfig {
...
+    string source = 1;
+    // The name of the runtime library package, i.e. "swift-protobuf".
+    string package = 2;
+    // The version of the runtime library, i.e. "1.21.0"
```
rebello95, line 177:
```suggestion
    // The version of the runtime library, i.e. "1.21.0".
```

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +167,41 @@ message MavenConfig {
...
+    string version = 3;
+    // The products exported by the runtime library.
```
rebello95, line 179: Would add an example here to clarify what this is

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -63,6 +64,15 @@ enum NPMImportStyle {
   NPM_IMPORT_STYLE_COMMONJS = 2;
 }
 
+// SwiftPlatformType is used to specify the platform type for a swift plugin version
```
rebello95, line 67:
```suggestion
// SwiftPlatformType is used to specify the platform type for a Swift plugin version.
```

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -157,15 +167,41 @@ message MavenConfig {
   repeated RuntimeConfig additional_runtimes = 3;
 }
 
+message SwiftConfig {
+  // RuntimeLibrary describes the runtime library dependency of the generated code.
+  message RuntimeLibrary {
```
rebello95, line 172: Would "dependency" make more sense than "library" here? (Not sure what was done for other package managers, so I'd keep it consistent with whatever was done for Maven)

rebello95 (commented):
Proto changes LGTM with a few minor comments - feel free to proceed after addressing

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
pgmitche: Dependency would work just as well, but all other registries use the RuntimeLibrary terminology

rebello95 (approved):

private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -239,10 +244,94 @@ func ProtoRegistryConfigToPluginRegistry(config *registryv1alpha1.RegistryConfig
...
+	if swiftConfig.Dependencies != nil {
+		protoSwiftConfig.RuntimeLibraries = make([]*registryv1alpha1.SwiftConfig_RuntimeLibrary, 0, len(swiftConfig.Dependencies))
+		for _, dependency := range swiftConfig.Dependencies {
+			depConfig := &registryv1alpha1.SwiftConfig_RuntimeLibrary{
+				Source:   dependency.Source,
+				Package:  dependency.Package,
+				Version:  dependency.Version,
+				Products: dependency.Products,
+			}
+			platforms := make([]*registryv1alpha1.SwiftConfig_RuntimeLibrary_Platform, 0)
```
stefanvanburen, line 301: we aren't really getting anything from the `make` here. maybe we just change the appends below to `depConfig.Platforms = append(depConfig.Platforms, ...`, which would let us remove the `if len(platforms) > 0` check below?

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -263,6 +276,51 @@ func newMavenRegistryConfig(externalMavenRegistryConfig *ExternalMavenRegistryCo
...
+		Dependencies: dependencies,
+	}, nil
+}
+
+func swiftExternalDependencyToDependencyConfig(externalDep ExternalSwiftRegistryDependencyConfig) (SwiftRegistryDependencyConfig, error) {
+	if externalDep.Source == "" {
+		return SwiftRegistryDependencyConfig{}, errors.New("swift runtime dependency requires a non-empty source")
+	}
+	if externalDep.Package == "" {
+		return SwiftRegistryDependencyConfig{}, errors.New("swift runtime dependency requires a non-empty module name")
```
stefanvanburen, line 301:
```suggestion
		return SwiftRegistryDependencyConfig{}, errors.New("swift runtime dependency requires a non-empty package name")
```

stefanvanburen (approved):
truly minor nits, but otherwise :shipit:

stefanvanburen (approved):

PR #2018, opened 2023-04-21
A read bucket for git trees

twilly:
A read bucket for git trees

Modules are built from buckets, meaning if we want to build a module
from git, then we'll need a bucket for it. It's possible to checkout a
git commit and build from a storageos bucket but there's a better way:
read directly from git's storage.

Here's a rough example of how to instantiate a bucket from a git commit:

```go
var commitref object.ID
commitref.UnmarshalText(
	[]byte("<commit hash>"),
)
runner := command.NewRunner()
objects, _ := NewCatFileServer(runner)
defer objects.Close()
commit, _ := objects.Commit(commitref)
bucket := NewTreeReader(objects, commit.Tree)
```


saquibmian:
> It's possible to checkout a git commit and build from a storageos bucket but there's a better way: read directly from git's storage.

I agree this would be a better way, but I don't think we need a better way for buf sync. There's a lot in the contract of `Bucket` that is implicit and has to be done carefully, you will be better off building off of the existing `GitRef`-based Buckets and addressing specific optimizations on that once you've got a POC in?
Perhaps I'm missing something, though, happy to chat!

twilly:
> Perhaps I'm missing something, though, happy to chat!

This PR is tricky because it started as the first component from a demo of walking git history and pushing modules. That PR was #1936. Both it and the rest of the code being reworked used go-git to back it, which turned into a no-go. Now this PR is the same basis of a simple integration *but* with an implementation of Git that isn't go-git. e4e5ff079589c2f7bc645b73586faaf1b7752427, 4d4be5176291915df58ba2efb93a6256ce1cbf1e, and b635e1e5292771ed2ad5dbb206fb95f5ab64835c are those bits.

The bucket abstraction itself (46183f187c4ee9371f0edef2566f8f1bebeef2c2) is still important, and we'll chat more about it, but now it's also a demonstration of using the new Git implementation. That's definitely a required bit of code.

twilly:
### How to read this PR

#### By Feature

The first commits split the change by feature:
   - 07cfbdc285aec622c5a4b7f45325f76866d2487a:  Git types for decoding git objects 
   - 4d4be5176291915df58ba2efb93a6256ce1cbf1e:  Git tree walker
   - b635e1e5292771ed2ad5dbb206fb95f5ab64835c:  Git cat-file object service 
   - 46183f187c4ee9371f0edef2566f8f1bebeef2c2: A read bucket for git trees

Later commits fixed up some small issues, in particular a little cleaning up of cat-file in 00eff61ed64610970b7829d441f9055316fc2fad and fixing the integration tests which were hoping to have access to buf's own git repository. 

Git types for decoding git objects provides a way to open git commits, trees, and blobs. It also provides an opaque git hash type so we can follow links to commits and trees.

The git tree walker provides finding services to list all entries or locate a named entry in a tree and its subtrees.

The git cat-file object services provides access to a git repository's objects through [git-cat-file](https://git-scm.com/docs/git-cat-file) in batch mode. This command is modeled as starting up a service and creating connections to that service.

The read bucket for git trees is the first abstraction on top of all that. It provides a storage.ReadBucket implementation from any git tree and an object service. This abstraction can be passed into bufmodulebuild.BuildForBucket, meaning we can construct any number of bufmodule.Module from trees in a git commit without a working tree on disk. This is particularly important when determining if a module has changed from git commit to git commit as git trees hashes do not change if their contents have not changed, while almost every git commit contains some content change.

#### By Package

`pkg/git/object` contains the core git types for decoding git objects (commit, tree, blob) and hash identifiers for them.
`pkg/git/gittest` contains testing helpers and utilities. There's a tree object constructor from a map[string]string, an in-memory object store, and a wrapper for calling the host's git implementation to construct git repos.
`pkg/git` contains the tree walker, the tree read bucket, and git-cat-file service.
`bufpkg/bufmodule/bufmodulebuild` adds a module bucket builder integration test with git.

twilly (commented):
Let's get this into review before I spend way too much time polishing it.

PR #2027, opened 2023-04-25
Include DocumentationPath in module push

seankimdev:
Include `DocumentationPath` in the module on `buf push`.
Support fallback paths, `README.md` and `README.markdown`, for module documentation.

If `buf.md` file is not available, the fallback paths `README.md`, or `README.markdown` are checked in the order.

- [x] Add changelog
- [x] Get the server side update merged first before merging this.


private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -36,6 +36,8 @@ import (
 const (
 	// DocumentationFilePath defines the path to the documentation file, relative to the root of the module.
 	DocumentationFilePath = "buf.md"
+	// FallbackDocumentationPath defines the fallback path to the documentation file, relative to the root of the module.
+	FallbackDocumentationPath = "readme.md"
```
saquibmian, line 40: I think this is case sensitive and we probably want `README.md`.

saquibmian (commented):
The only part of this that's funky is when we do `ModuleToBucket` we don't know the original filename, so we would be writing it with `buf.md` when it could have been `README.md`. @bufdev can you see a scenario where that matters?

private/bufpkg/bufmodule/bufmodule.go
bufdev: definitely case-sensitive

bufdev:
Wait, might not matter for buf export, but yea I think I thought about this before and it was more complicated - @doriable might have insight as well (might involve docs)

saquibmian:
We could go about this in a few different ways:
1. `modulev1alpha1.Module` could gain a new field, `string fallback_documentation`
2. `modulev1alpha1.Module` could gain a new field, `string documentation_filename`
3. `modulev1alpha1.Module` could gain a new field, `DocumentationFile documentation_file` which has filename and content.

All three of these scenarios have the BSR storing the filename, we could default to `buf.md` if empty.

Are there alternative approaches?

saquibmian:
`buf export` to my knowledge doesn't output documentation? It just writes an `Image`, which I don't think has documentation?

private/bufpkg/bufmodule/bufmodule.go
seankimdev: Fixed

doriable:
> buf export to my knowledge doesn't output documentation? It just writes an Image, which I don't think has documentation?

Yes this is correct -- `buf export` will not affect documentation, so that is less of a concern.

bufdev:
You likely need to store the file path in the `Module` yea. It should be documented to be normalized, similar to other paths in Protobuf definitions

bufdev:
This may affect the digest as well

doriable:
I guess I'm questioning a little bit the user experience problem we are solving here. In our original discussions, we were concerned with just picking up `README.md` because some users may be storing internal-facing information in a `README.md` file alongside their proto files in a private repository. In the situation where the source code is closed, this may be crossing a boundary. It was a conscious decision to make a file that is both Buf branded and explicit so there is no leaking of information.

twilly:
> This may affect the digest as well

This will but that's OK. A new release will include this file when sending to/receiving from the BSR. Older clients will ignore it.

It does mean using an old and new client together will cause a ping-pong of commits, but that's expected behavior given each one is sending different content to the BSR.

seankimdev:
Putting this on hold until we find out how to proceed due to the concerns of its side effects and the user experience problem.

doriable:
Had an offline discussion with @bufdev. We came to the conclusion that private information in a `README` alongside a Buf module is likely an edge case, and that doing this fallback behaviour early is better than later. This can help ease of use when it comes to publishing documentation and emulates the same behaviour as Docker (getting your README from your source code repository).

With that being said, a couple of notes for this change:

- There should be a very clear CHANGELOG entry for this
- We should be storing the file name on the BSR side, and default to `buf.md` for all the existing data without a file name
- The fallback path should include `README.markdown` if we are going to go down this route, since both `md` and `markdown` are both valid extensions per RFC 7763. We would be following the same conventions as Go documentation (https://github.com/golang/pkgsite/blob/6862513e1b973c06468b98ed333cf9b80498b2a6/internal/postgres/searchdoc.go#L175-L179)

twilly:
>     * We should be storing the file name on the BSR side, and default to `buf.md` for all the existing data without a file name
> 
>     * The fallback path should include `README.markdown` if we are going to go down this route, since both `md` and `markdown` are both valid extensions per RFC 7763. We would be following the same conventions as Go documentation (https://github.com/golang/pkgsite/blob/6862513e1b973c06468b98ed333cf9b80498b2a6/internal/postgres/searchdoc.go#L175-L179)

Something to keep in mind: the more paths that we can consume for the "same" documentation, the more we have to support when trying to parse documentation. Case in point, the consideration of what to do when a user has `buf.md`, `README.md`, and `README.markdown` in the same module.

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -413,6 +421,7 @@ func ModuleDigestB3(ctx context.Context, module Module) (string, error) {
 			return "", err
 		}
 	}
+	// TODO: Should we include module.DocumentationPath here?
```
seankimdev, line 424: Should the documentation path be included in the module digest?

doriable: I think so -- I think it is desirable to have a push occur when the documentation path changes, so that should result in a digest change.

seankimdev: hm thinking if it would break anyone. I guess it would result in a digest change if you had an existing `buf.md` file but didn't change anything. Could that be a problem?

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -421,7 +421,11 @@ func ModuleDigestB3(ctx context.Context, module Module) (string, error) {
 			return "", err
 		}
 	}
-	// TODO: Should we include module.DocumentationPath here?
+	if docPath := module.DocumentationPath(); docPath != "" {
```
pkwarren, line 424: I do think for compatibility if `buf.md` was the file name, that's the behavior we had before, so it shouldn't update the b3 digest.

I think to address this, we'd change the conditional to be `docPath != "" && dockPath != "buf.md"` (w/ the constant for "buf.md").

seankimdev: but then wouldn't it skip updating the digest if the filename was changed from one of the `README`'s to `buf.md`?

pkwarren: We never looked at the other filenames for docs before this PR, so that should be safe.

seankimdev: I mean going forward, what if a module's filename is changed to `README.md` and then back to `buf.md`? wouldn't `buf.md` be excluded from the digest due to `docPath != "" && docPath != "buf.md"`?

private/bufpkg/bufmodule/bufmodule.go
seankimdev: Continuing in the other [thread](https://github.com/bufbuild/buf/pull/2027#discussion_r1180842068) below.

private/bufpkg/bufmodule/bufmodule.go
pkwarren: Yeah - in that case the documentation path won't be part of the digest, so the digest will still change.

What I was trying to avoid here is having the b3 digest change if the buf.md file was always used.

seankimdev: ah you are right, the digest will still change. I guess whether the documentation path is included or not isn't very important in this case as long as the digests are different?

private/bufpkg/bufmodule/bufmoduletesting/bufmoduletesting.go
```diff
@@ -124,10 +126,10 @@ var (
...
 	// TestDataWithDocumentation is the data that maps to TestDigestWithDocumentation.
 	//
-	// It includes a buf.md file.
+	// It includes a README.md file.
 	TestDataWithDocumentation = map[string][]byte{
-		TestFile1Path: []byte(`syntax="proto3";`),
-		"buf.md":      []byte(TestModuleDocumentation),
+		TestFile1Path:               []byte(`syntax="proto3";`),
+		TestModuleDocumentationPath: []byte(TestModuleDocumentation),
 	}
```
doriable, line 133: Should we have two tests for this, one for `buf.md` and the other for the fallback paths? That way, we can also ensure the digest doesn't change in the old code path.

bufdev: +1

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -36,6 +36,10 @@ import (
 const (
 	// DocumentationFilePath defines the path to the documentation file, relative to the root of the module.
 	DocumentationFilePath = "buf.md"
+	// FallbackDocumentationPathReadmeMD defines the fallback path to the documentation file, relative to the root of the module.
```
bufdev, line 39: Nit (optional if you want to clean this up), I likely would have done:

```go
const (
  DefaultDocumentationPath = "buf.md"
)

var (
  AllDocumentationPaths = []string{
    DefaultDocumentationPath,
    "README.md",
    "README.markdown",
  }
)
```

To make it easy to add in the future. I'd be consistent on using `Path` instead of `FilePath/Path`, as our goal is that in theory, you could have these in subdirs, although this is inconsistent with our choice in `buflock`.

private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
```diff
@@ -83,6 +83,8 @@ func (b *moduleBucketBuilder) BuildForBucket(
 	externalPaths := []string{
 		buflock.ExternalConfigFilePath,
 		bufmodule.DocumentationFilePath,
+		bufmodule.FallbackDocumentationPathReadmeMD,
```
bufdev, line 86: If you did the `AllDocumentationPaths` above, you'd do:

```go
externalPaths = append(
  AllDocumentationPaths,
  buflock.ExternalConfigFilePath,
  bufmodule.LicenseFilePath,
)
```

private/bufpkg/bufmodule/bufmodule.go
seankimdev: Definitely want to consider how it can be expanded in the future. I'll get them changed. Thank you for the feedback; really appreciate it.

private/bufpkg/bufmodule/bufmoduletesting/bufmoduletesting.go
seankimdev: Returning the existing test to use `buf.md`, and added another test for `README.md`.

bufdev (approved):

private/bufpkg/bufmodule/bufmodule.go
saquibmian: @pkwarren is there a reason for why we're avoiding the digest changing? Is it because we'll cache invalidate basically everything, or are checksums no longer going to match (i.e., we can never change the digest function for a module that has already been published)?
To me it seems odd that the docs are _sometimes_ included in the manifest and _sometimes_ are not (based on the filename).

pkwarren: We use the b3 digest server side to determine if we should short circuit and prevent an identical push. If we changed the hash here we'd allow a new push for a module that hadn't changed. This might also lead to a ping/pong effect pushing to the same module from two clients with different cli versions.

saquibmian: Don't we still have that problem with this PR? If you have a `README.md`, your module's hash is now computed differently once this lands (CLI vNext), and this can result in ping/pong with older CLI versions.

pkwarren: Yeah but it is probably more likely someone had a buf.md file before this change since that was the only supported doc path.

saquibmian: What about the scenario where someone didn't have a `buf.md`, but now wants to add one? The BSR won't accept the push because the digest didn't change. Granted we're in this problem today, it seems, but it's definitely undesirable behaviour.

I feel like it's reasonable to require users to use a consistent version of the buf CLI at push time, (i.e., in CI/Github Action/GitHub App).

pkwarren: The bsr would accept that change because GetDocumentation() would change from empty to the contents of the buf.md file.

pkwarren: So with this change, we have the behavior:

* buf push (1.17.0), buf.md docs: no changes
* buf push (1.18.0), buf.md docs: no changes

* buf push (1.17.0), README.* docs: no changes (doesn't support new path)
* buf push (1.18.0), README.* docs: changes (pushes new path and docs)

If we leave this change out, then identical modules (buf.md w/ same contents as before) will result in duplicate pushes to the BSR with no changes between them.

saquibmian: Oh I'm totally misreading this, it's the _path_ that we are debating hashing, not the docs themselves. 

Is the issue that we'll have a duplicate push to the BSR with no changes, or ping-pong? Ping-pong I see as less of a concern, as users will get strange behaviour if they are not consistent in the version of the CLI they use (e.g., in this case, docs appearing and disappearing and reappearing).

pkwarren: Agree as users will see that with new functionality. I just was trying to prevent pushing identical commits and maybe the likelihood of ping/pong effects would be higher for modules that have a buf.md file today.

seankimdev: Thank you both for reasoning through this again 🙏

saquibmian (approved):
Nice job!

PR #2028, opened 2023-04-26
Add documentation_path field to Module

seankimdev:
Add `documentation_path` field to the Module for storing a path, either `buf.md`, `README.md` or `README.markdown`, of a file that contains module documentation.


proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -65,6 +65,10 @@ message Module {
   //
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
   string license = 6;
+  // documentation_filename is the filename of the file which contains the module documentation.
+  //
+  // either `buf.md`, `README.md` or `README.markdown`.
+  string documentation_filename = 7;
```
bufdev, line 71: I would call this `documentation_path` in case in the future we want to support this being in another directory

seankimdev: Makes sense. I'll get that changed. Thanks!

doriable (approved):

proto/buf/alpha/module/v1alpha1/module.proto
```diff
@@ -65,6 +65,10 @@ message Module {
   //
   // string is used to enforce UTF-8 encoding or 7-bit ASCII text.
   string license = 6;
+  // documentation_path is the path of the file which contains the module documentation.
+  //
+  // either `buf.md`, `README.md` or `README.markdown`.
```
bufdev, line 70: Document that if empty, assumes buf.md

bufdev (approved):
Approved pending comment

proto/buf/alpha/module/v1alpha1/module.proto
seankimdev: Added

PR #2033, opened 2023-04-28
EXAMPLE ONLY DO NOT MERGE: Add a lint rule for protoc-gen-validate

bufdev:
This shows an example of how you could add a lint rule for protoc-gen-validate end-to-end. In this case, it adds a rule to `DEFAULT`, which we could do for PGV v2.0 without any breaking changes to `buf`. The actual rule is nonsense, this is just to show the code path.

Open issues:

- The tests in `buflint_test.go` are not really built for external files, and with how they are written, we'd have to copy/paste the validate.proto file into each test directory and then ignore it. We could make these tests workspace-aware, but that'd be even more of a lift. I'd suggest a new helper function `testLintWithValidate` that will just sideload `validate/validate.proto` into the `Bucket` being built for the module, and then add `"validate"` to the `IgnoreRootPaths` list on the `*buflint.Config`. I can wire this up if it's helpful once we get there.
- @jhump, we need to add location information for extensions to `protosource.OptionExtensionDescriptor`, likely in a new `OptionExtensionLocation` function:
 
```go
type OptionExtensionDescriptor interface {
  OptionExtension(protoreflect.ExtensionType) (interface{}, bool)
  OptionExtensionLocation(protoreflect.ExtensionType) Location
```

This is assuming that extensions have `SourceCodeInfos`, which I'm actually not sure about - you'd know though I think heh.


jhump:
> This is assuming that extensions have `SourceCodeInfos`, which I'm actually not sure about - you'd know though I think heh.

LOL. They do indeed have source code info... thanks to me :P https://github.com/protocolbuffers/protobuf/pull/4342

However, options that are messages (not just scalar values) get complicated. They can be de-structured across multiple option statements, so each such statement has its own source code info (and the path to it indicates the particular field inside the option message). If a message literal is used, the source code info span represents the entire message (no details about locations _inside_ the message literal).

So if you want info for _a particular field_ in an option message, there _may_ be a dedicated source code info row or you may have to look for an entry for the containing/ancestor message (in the latter case, you won't know where in the message literal that field was likely actually defined).

Similarly, if you want source code info for an option message (the whole thing), it _may_ have a dedicated source code info row or it may be spread across multiple entries.

FWIW, we could probably emit _way_ better source code info from `buf` for message literals because we have all of the location information for all of the fields _inside_ the message literal. But right now, we're just mirroring `protoc`.

bufdev:
We could go simple for now - we really just need a location that is at least around the option. Right now this code as written is using the location of the `message`, which isn't very granular (and likely has the "wrong" line number). If we just started with the "whole" option, that's completely good enough for now, we don't need this for a theoretical i.e. `buf lint --fix`.

PR #2034, opened 2023-04-28
Only cache Go artifacts once

pkwarren:
Right now, we're caching GOMODCACHE and GOCACHE in two places - our own cache rule and in actions/setup-go. Remove our custom caching and just use the built-in caching in setup-go.

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -6,10 +6,8 @@ permissions: read-all
 jobs:
   test:
     env:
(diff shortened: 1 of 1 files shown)


pkwarren:
See here: https://github.com/bufbuild/buf/actions/runs/4834234181/jobs/8615210532

It is caching the same artifacts in both "Post cache" and "Post setup-go".

bufdev (approved):

PR #2054, opened 2023-05-07
use docker cmd

itaischwartz:
CMD is preferred for projects such as buf, as it allows using a single container for multiple commands. 
Inspired by https://github.com/golangci/golangci-lint/blob/master/build/Dockerfile.


bufdev:
Thanks for the contribution, but this would be a breaking change to the image, and Buf is stable now, so we can't accept this.

PR #2059, opened 2023-05-08
Validate outer and absolute manifest paths

unmultimedio:


private/pkg/manifest/manifest.go
```diff
@@ -106,6 +107,13 @@ func (m *Manifest) AddEntry(path string, digest Digest) error {
 	if path == "" {
 		return errors.New("empty path")
 	}
+	path = filepath.Clean(path)
+	if strings.Contains(path, "..") {
+		return errors.New(`invalid sequence ".."  in path`)
```
pkwarren, line 112:
```suggestion
		return errors.New(`invalid sequence ".." in path`)
```

private/pkg/manifest/manifest.go
```diff
@@ -106,6 +107,13 @@ func (m *Manifest) AddEntry(path string, digest Digest) error {
 	if path == "" {
 		return errors.New("empty path")
 	}
+	path = filepath.Clean(path)
```
pkwarren, line 110: Should we use `filepathextended.RealClean` here?

pkwarren (approved):

twilly (approved):

pkwarren (approved):

PR #2066, opened 2023-05-10
Add buf push --create flag

bufdev:


doriable (approved):
Looks good to me! I also double checked the errors we're returning in the cases where the repository exists/permissions issues, and this seems very reasonable.

doriable (approved):
Looks good to me! I also double checked the errors we're returning in the cases where the repository exists/permissions issues, and this seems very reasonable.

doriable (changes requested):
Pushing up tests for this, blocking so that this doesn't get accidentally merged.

doriable (approved):
Re-approving to remove my change request, and requested additional reviewers for tests and changelog.

bufdev (commented):
Looks good from my perspective.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -199,6 +220,33 @@ func push(
 	if err != nil {
 		return nil, err
 	}
+	modulePin, err := push(ctx, container, clientConfig, moduleIdentity, builtModule, flags)
+	if err != nil {
+		// We rely on Push* returning a NotFound error to denote the repository is not created.
+		// This technically could be a NotFound error for some other entity than the repository
+		// in question, however if it is, then this Create call will just fail as the repository
+		// is already created, and there is no side effect. The 99% case is that a NotFound
+		// error is because the repository does not exist, and we want to avoide having to do
```
bufdev, line 229: Nit: I annoyingly had a typo in this comment. @doriable, I am OOO tomorrow morning, but if you want to get merged beforehand and we get a review, we should fix this first.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -199,6 +220,33 @@ func push(
...
+		// a GetRepository RPC call for every call to push --create.
+		if flags.Create && connect.CodeOf(err) == connect.CodeNotFound {
+			if err := create(ctx, container, clientConfig, moduleIdentity, flags); err != nil {
```
bufdev, line 232: I just realized one other annoying thing: in the case the comment refers to, or on a random race condition, where this create call is done in error on a repository that already exists, this will return a CodeAlreadyExists, which in turn will result in a message to stderr that the commit already exists and a zero exit code. This is wrong - in this case, we should error, with a message that stating we expected the repository to not exist but it did, and this should be changed to an internal error.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -139,6 +150,16 @@ func run(
 	if len(flags.Tags) > 0 && flags.Draft != "" {
 		return appcmd.NewInvalidArgumentErrorf("--%s (-%s) and --%s cannot be used together.", tagFlagName, tagFlagShortName, draftFlagName)
 	}
+	if flags.CreateVisibility != "" {
+		if !flags.Create {
+			return appcmd.NewInvalidArgumentErrorf("Cannot set --%s without --%s.", createVisibilityFlagName, createFlagName)
+		}
+		// We re-parse below as needed, but do not return an appcmd.NewInvalidArgumentError below as
+		// we expect valiation to be handled here.
+		if _, err := bufcli.VisibilityFlagToVisibility(flags.CreateVisibility); err != nil {
```
bufdev, line 159: Ugh, one extra thing - sorry: we should make sure the documentation says this flag is required if --create is set, and validate that it is non-empty immediately before this line, so that there's a proper error message explaining that --create-visibility is required if --create is set. We could also take the tact to default to private, but likely better to start as required.

private/buf/cmd/buf/command/push/push.go
doriable: Yeah, that's reasonable -- I can make that check when the error is checked.

private/buf/cmd/buf/command/push/push.go
bufdev: Oh, I think you meant making it required when you say "force setting", in which case we wouldn't be in agreement, lol

private/buf/cmd/buf/command/push/push.go
```diff
@@ -139,6 +150,16 @@ func run(
 	if len(flags.Tags) > 0 && flags.Draft != "" {
 		return appcmd.NewInvalidArgumentErrorf("--%s (-%s) and --%s cannot be used together.", tagFlagName, tagFlagShortName, draftFlagName)
 	}
+	if flags.CreateVisibility != "" {
+		if !flags.Create {
+			return appcmd.NewInvalidArgumentErrorf("Cannot set --%s without --%s.", createVisibilityFlagName, createFlagName)
+		}
```
unmultimedio, line 156: Nit: if both flags are required, and cannot be sent one w/out the other, may be it makes more sense having just a single flag called --create-with-visibility {REPO_VISIBILITY}? It'd save this double check.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -139,6 +150,16 @@ func run(
...
+		// We re-parse below as needed, but do not return an appcmd.NewInvalidArgumentError below as
+		// we expect valiation to be handled here.
```
unmultimedio, line 158:
```suggestion
		// we expect validation to be handled here.
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -199,6 +220,33 @@ func push(
...
 		return nil, err
 	}
+	modulePin, err := push(ctx, container, clientConfig, moduleIdentity, builtModule, flags)
+	if err != nil {
+		// We rely on Push* returning a NotFound error to denote the repository is not created.
+		// This technically could be a NotFound error for some other entity than the repository
+		// in question, however if it is, then this Create call will just fail as the repository
+		// is already created, and there is no side effect. The 99% case is that a NotFound
+		// error is because the repository does not exist, and we want to avoid having to do
+		// a GetRepository RPC call for every call to push --create.
```
unmultimedio, line 230: Why not instead do an invocation to `CreateRepository` RPC always if the `--create` flag is passed.

That should return either an OK (repo created), or an AlreadyExists error (I assume this `--create` flag is idempotent?), so with any of those two results, we continue with the push, otherwise, we fail the push command.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -167,7 +188,7 @@ func run(
 	if err != nil {
 		return err
 	}
-	modulePin, err := push(ctx, container, moduleIdentity, builtModule, flags)
+	modulePin, err := pushOrCreate(ctx, container, moduleIdentity, builtModule, flags)
```
unmultimedio, line 191: Nit also in favor of clarity: I'd call the `createRepo` func inside the `if flags.CreateVisibility != "" {` block, and in here I'd just keep doing a normal push.

bufdev: We generally do all the flag validation up front, so that the command fails fast on invalid input, and so that we take no action if the input is invalid.

private/buf/cmd/buf/command/push/push.go
bufdev: Actually just thought about it again (after implementing it and almost pushing it) - its better to first do the `push` call, because in the 99% case, this will result in only one RPC, i.e. the only time that `push` should fail is on the first instance where the repository is not created. For the remainder of the lifetime of the module, `push --create` will be unnecessarily be doing two RPCs instead of one, with the first RPC failing.

private/buf/cmd/buf/command/push/push.go
unmultimedio: Good point. Atm I cannot think of another option besides visibility, but you're right, in case of another flag it'd be very weird. And if we endup making private the default, then the `--create-with-visibility` can just become optional and only the `--create` would be required.

private/buf/cmd/buf/command/push/push.go
unmultimedio: Alright, I guess this is a tradeoff to protect network usage against users adding this flag always to CIs, when in reality they need to run it just once.

private/buf/cmd/buf/command/push/push.go
unmultimedio: And not only that, if we're not invoking createRepo RPC every time the flag is passed (as my other comment suggested), then my comment here makes not much sense, create would need to happen only if push fails.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -199,6 +222,33 @@ func push(
 	if err != nil {
 		return nil, err
 	}
+	modulePin, err := push(ctx, container, clientConfig, moduleIdentity, builtModule, flags)
+	if err != nil {
```
unmultimedio, line 226: Just one more nit to reduce nesting:

```go
pin, err := push(...)
if err == nil {
  return pin, nil // bail early, no need to create repo
}
// comment about the not found err
if err != "NotFound" {
  return nil, err
}
_, err := createRepo(...)
if err != nil  {
  return err
}
// repo created, try to push again
return push(...)
```

unmultimedio (approved):

private/buf/cmd/buf/command/push/push.go
bufdev: You want to return the NotFound if `--create` wasn't set, so then the logic gets a little confusing IMO, thoughts?

```go
pin, err := push(...)
if err == nil {
  return pin, nil // bail early, no need to create repo
}
// comment about the not found err
// logic takes some thinking when I read it
if !flags.Create || err != "NotFound" {
  return nil, err
}
// flags.Create && err == "NotFound"
// need people to understand the above comment
_, err := createRepo(...)
if err != nil  {
  return err
}
// repo created, try to push again
return push(...)
```

bufdev: Eck, I get it, I don't know though - may be a matter of taste. I like the current style better as its really explicit as to what's happening - attempt to push, if the repo isn't found and you have --create set, create and then try to push again. @doriable can be tiebreaker, any thoughts?

doriable: Haha, I purposefully didn't weigh in immediately. I do think it's a matter of style, and I would say the way it's currently structured is much closer to the rest of the CLI code, and since we do value consistency highly, I didn't immediately make the change this morning. I think in the future, if we are looking to put additional testing or test code paths a certain way that warrant a restructuring, we could revisit?

unmultimedio (approved):

PR #2073, opened 2023-05-11
Mark idempotent RPCs

oliversun9:
This adds `option idempotency_level = IDEMPOTENT;` to idempotent RPCs.
Ran `make generate`.
Most `Update*` and `Set*` aren't idempotent because they modify timestamps.


proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -40,8 +40,12 @@ message LabelValue {
 }
 
 service LabelService {
-  rpc CreateLabel(CreateLabelRequest) returns (CreateLabelResponse);
-  rpc MoveLabel(MoveLabelRequest) returns (MoveLabelResponse);
+  rpc CreateLabel(CreateLabelRequest) returns (CreateLabelResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
+  rpc MoveLabel(MoveLabelRequest) returns (MoveLabelResponse) {
+    option idempotency_level = IDEMPOTENT;
```
bufdev, line 47: Just making sure this is right @doriable

For quick reference: https://developer.mozilla.org/en-US/docs/Glossary/Idempotent

doriable: Yes, I believe this is correct: you're always required to set the commit the label is moving to, so calling the move to the same destination multiple times would be considered idempotent.

oliversun9: I didn't realize that 'from' was optional, in that case the timestamp will be updated on each call. Would it still be considered idempotent?

doriable: Ah yes, `from` is optional. You're right, it would not be.

oliversun9:
> Is this ready to go?

It just needs a final approval, it's ready for review

bufdev:
@mfridman @pkwarren can one of you check this over?

oliversun9:
Btw, [this](https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/webhook/v1alpha1/event.proto#LL27C51-L27C51) is not marked because it's not exactly an interface of BSR.

pkwarren (approved):
Changes look good to me. Saw a few opportunities for places where we could make additional RPCs idempotent and also improve the implementation of some calls, but this looks good to me.

pkwarren:
> Also @pkwarren after this PR, are all RPCs that have no side effects marked with `NO_SIDE_EFFECTS`, and all RPCs that have side effects but are idempotent marked `IDEMPOTENT` - this is my other question, just sanity checking

I think we might need to make a second pass as I'm seeing some more methods that should be able to be marked idempotent. I did check the ones in this PR though and they are safe to retry.

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -26,13 +26,17 @@ import "buf/alpha/registry/v1alpha1/verification_status.proto";
 service AdminService {
   // ForceDeleteUser forces to delete a user. Resources and organizations that are
   // solely owned by the user will also be deleted.
-  rpc ForceDeleteUser(ForceDeleteUserRequest) returns (ForceDeleteUserResponse);
+  rpc ForceDeleteUser(ForceDeleteUserRequest) returns (ForceDeleteUserResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
   // Update a user's verification status.
   rpc UpdateUserVerificationStatus(UpdateUserVerificationStatusRequest) returns (UpdateUserVerificationStatusResponse);
```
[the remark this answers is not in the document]
oliversun9: Chatted offline, will do this in a separate task

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -67,17 +67,27 @@ service OrganizationService {
...
   // AddOrganizationMember add a role to an user in the organization.
-  rpc AddOrganizationMember(AddOrganizationMemberRequest) returns (AddOrganizationMemberResponse);
+  rpc AddOrganizationMember(AddOrganizationMemberRequest) returns (AddOrganizationMemberResponse) {
+    option idempotency_level = IDEMPOTENT;
+  }
   // UpdateOrganizationMember update the user's membership information in the organization.
   rpc UpdateOrganizationMember(UpdateOrganizationMemberRequest) returns (UpdateOrganizationMemberResponse);
   // RemoveOrganizationMember remove the role of an user in the organization.
-  rpc RemoveOrganizationMember(RemoveOrganizationMemberRequest) returns (RemoveOrganizationMemberResponse);
+  rpc RemoveOrganizationMember(RemoveOrganizationMemberRequest) returns (RemoveOrganizationMemberResponse) {
```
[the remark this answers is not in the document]
oliversun9: You are right, I must have missed that

proto/buf/alpha/registry/v1alpha1/plugin.proto
```diff
@@ -172,15 +172,19 @@ service PluginService {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
   // CreatePlugin creates a new plugin.
-  rpc CreatePlugin(CreatePluginRequest) returns (CreatePluginResponse);
+  rpc CreatePlugin(CreatePluginRequest) returns (CreatePluginResponse) {
```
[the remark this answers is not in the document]
oliversun9: Sounds good

PR #2089, opened 2023-05-15
Throw ctx cancelled error in `thread.Parallelize`

unmultimedio:
Found a panic in a test, caused because of a DX assumption. From users perspective, doing:

```go
foo := make([]Foo, someLen)
// jobs preparation... each job fills one item in []foo
if err := thread.Parallelize(
  ctx,
  jobs,
  thread.ParallelizeWithCancel(cancelGetBlobs), // if one job fails, stop all jobs
); err != nil {
  return nil, err
}
// at this point the user could assume that if no errors happened, then []foo was successfully
// populated... well, if the ctx was cancelled, some/all items in []foo might be empty/nil.
// Even if the ctx was cancelled not by an error trying to fill []foo, but upstream from
// another caller.
```

~I will test in a dependent repo before making this ready for review.~ Tested.

### DIFF
--- a/private/pkg/thread/thread.go
+++ b/private/pkg/thread/thread.go
@@ -16,7 +16,6 @@ package thread
 
 import (
 	"context"
-	"fmt"
 	"runtime"
 	"sync"
 
@@ -74,7 +73,7 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
 	var wg sync.WaitGroup
 	var lock sync.Mutex
 	var stop bool
-	for jobIdx, job := range jobs {
+	for _, job := range jobs {
 		if stop {
 			break
 		}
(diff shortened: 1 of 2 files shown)


doriable (approved):
This is a good find! Stamping, but may want input from others also :)

private/pkg/thread/thread.go
```diff
@@ -87,16 +87,19 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
 		select {
 		case <-ctx.Done():
 			stop = true
+			retErr = multierr.Append(retErr, fmt.Errorf("jobs[%d]: context canceled", jobIdx))
```
bufdev, line 90: This should likely be `ctx.Err()` instead of a custom error, some people parse the specific context canceled function

private/pkg/thread/thread.go
```diff
@@ -87,16 +87,19 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
...
 		case semaphoreC <- struct{}{}:
 			select {
 			case <-ctx.Done():
 				stop = true
+				retErr = multierr.Append(retErr, fmt.Errorf("jobs[%d]: context canceled", jobIdx))
```
bufdev, line 95: same comment

bufdev (commented):
I'd probably delete the `jobIdx` stuff, see comments

private/pkg/thread/thread.go
```diff
@@ -77,7 +77,6 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
 		if stop {
 			break
 		}
-		job := job
```
bufdev, line 80: i would keep this here, only doing the set within `default` feels like a nasty bug in the future if we ever change this and refer to `job` somewhere else

bufdev: if anything id move it above the `if stop {`, always keeping it close to the `for` loop start

unmultimedio: I moved it closer to the usage to avoid declaring it on each loop w/out any usage. If using it elsewhere, pretty sure govet would lint it, and also this loop scope var issue it's supposed to [be fixed soon by the go team](https://github.com/golang/go/issues/60078). I'll move it back up, although I don't think it's quite necessary.

bufdev: oh fair enough...idk dont worry about it

PR #2093, opened 2023-05-16
BlobSet constructor defensive against nil

unmultimedio:


private/pkg/manifest/module.go
```diff
@@ -115,14 +116,30 @@ func BlobSetWithContentValidation() BlobSetOption {
...
 	digestToBlobs := make(map[string]Blob, len(blobs))
-	for _, b := range blobs {
+	for i, b := range blobs {
+		if b == nil {
+			if config.skipNilBlobs {
+				continue
+			}
+			return nil, fmt.Errorf("blobs[%d]: nil blob", i)
+		}
 		digestStr := b.Digest().String()
```
unmultimedio, line 143: nil-checking saves us from panicking in this line.

pkwarren (approved):

Issue #1920, opened 2023-03-15
Reading `protofile` from stdin doesn't work


joshcarp:
An error shows when running buf build with `-#format=protofile`. This should either be fixed, or not be supported and an error be returned.

```
cat proto/simple/v1/simple.proto | buf build -#format=protofile --output -    
Failure: path is empty
```

```proto
syntax = "proto3";
package simple.v1;

message Person {
  string simple = 1;
}

```

Issue #1915, opened 2023-03-14
Command buf format panics with `buf format -#format=protofile`


joshcarp:
cat proto/simple/v1/simple.proto | buf format -#format=protofile --output -

PR #2095, opened 2023-05-16
Add error message if protofile is attempted to be read or written from stdio

bufdev:
Fixes https://github.com/bufbuild/buf/issues/1920.
Fixes https://github.com/bufbuild/buf/issues/1915.

### DIFF
--- a/private/buf/buffetch/internal/proto_file_ref.go
+++ b/private/buf/buffetch/internal/proto_file_ref.go
@@ -14,6 +14,8 @@
 
 package internal
 
+import "github.com/bufbuild/buf/private/pkg/app"
+
 var (
 	_ ParsedProtoFileRef = &protoFileRef{}
 )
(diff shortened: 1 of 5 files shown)


private/buf/buffetch/internal/proto_file_ref.go
```diff
@@ -14,6 +14,8 @@
 
 package internal
 
+import "github.com/bufbuild/buf/private/pkg/app"
```
doriable, line 17: Nit: import formatting with parens.

doriable (approved):
Just left a small nit.

private/buf/buffetch/internal/proto_file_ref.go
bufdev: talked in person, not worrying about this, this is acceptable

PR #2096, opened 2023-05-17
Fix basic auth in buf curl

mattrobenolt:
basic auth needs the Basic prefix

Refs #2010


bufdev:
Could you add a quick line to the changelog in the form "Fix ...", similar to other fix lines?

jhump (approved):

PR #2098, opened 2023-05-17
Fix buf export with tamper proofing enabled

pkwarren:
Calls to buf export are failing with the tamper proofing feature enabled. We need to update to use the manifest digest from the response instead of the b3 digest.

Verified that buf export works both with and without the feature enabled.


unmultimedio (approved):
🚀 Nice catch!

PR #2104, opened 2023-05-18
Update JavaPackage modifier to support java_package

pkwarren:
The current version of the JavaPackage modifier doesn't support prepending a java_package_prefix to an existing value of the `java_package` file option. This makes sense for generation with remote plugins and managed mode, however for remote packages we may wish to preserve the user's `java_package` setting and use it instead of the protobuf package name in the final output. This can be useful to provide more control over the Java packages in the generated output, and also avoid known issues with Java keywords in package names that prevent successful compilation.


bufdev (changes requested):
See my comment in linear

PR #2111, opened 2023-05-22
Ignore b1/b3 buf.lock digests only

### LINKED ISSUE #2110: Silent success with invalid digest
Given any `buf.yaml` with dependencies and a `buf.lock`, we expect `buf export --config buf.yaml --output .` to fail if the digest is incorrect. However, if the digest is incorrect but malformed, the digest is simply ignored.

i.e. If you replace a character of the digest with a valid incorrect character, buf will warn that the digest does not match. But if you just delete a character of the digest, buf will ignore the digest and run the command successfully.

IMO this should still fail.

pkwarren:
Update the code that reads a buf.lock file from disk to only ignore previous b1/b3 module digests. If the lock file has an invalid manifest digest, we want to error early.

Fixes #2110.

### DIFF
--- a/private/bufpkg/buflock/buflock.go
+++ b/private/bufpkg/buflock/buflock.go
@@ -17,9 +17,9 @@ package buflock
 
 import (
 	"context"
+	"strings"
 	"time"
 
-	"github.com/bufbuild/buf/private/pkg/manifest"
 	"github.com/bufbuild/buf/private/pkg/storage"
 )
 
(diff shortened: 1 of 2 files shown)


private/bufpkg/buflock/buflock.go
```diff
@@ -152,10 +152,3 @@ func ExternalConfigDependencyV1Beta1ForDependency(dep Dependency) ExternalConfig
 type ExternalConfigVersion struct {
 	Version string `json:"version,omitempty" yaml:"version,omitempty"`
 }
-
-// isValidDigest returns true when the digest string is successfully parsed
-// by the `manifest` pkg. Older buf digests are not considered valid (b1/b3).
-func isValidDigest(digest string) bool {
```
pkwarren, line 158: This was too lenient before - it would also ignore a digest field in the newer format that wasn't valid (wrong encoding, length, algorithm, etc.). We only wanted to ignore buf.lock files which may have potentially contained a b1/b3 digest value.

saquibmian (approved):

PR #2114, opened 2023-05-23
Add a Git-backed `storage.ReadBucket` via `storagegit`

saquibmian:


private/pkg/git/git.go
```diff
@@ -268,3 +272,49 @@ func OpenObjectReader(
...
+// Tree is a git tree, which are a manifest of other git objects, including other trees.
+type Tree interface {
+	// Hash is the Hash for this Tree.
+	Hash() Hash
+	// Entries is the set of entries in this Tree.
+	Entries() []TreeEntry
+	// Traverse walks down a tree, following the name-path specified,
+	// and returns the terminal TreeEntry. If no entry is found, it returns
+	// ErrSubTreeNotFound.
+	Traverse(objectReader ObjectReader, names ...string) (TreeEntry, error)
```
bufdev, line 307: Why not `Walk`?

private/pkg/git/git.go
```diff
@@ -268,3 +272,49 @@ func OpenObjectReader(
...
+type FileMode uint32
+
+var ErrSubTreeNotFound = errors.New("subtree not found")
+
+// Tree is a git tree, which are a manifest of other git objects, including other trees.
+type Tree interface {
+	// Hash is the Hash for this Tree.
+	Hash() Hash
+	// Entries is the set of entries in this Tree.
+	Entries() []TreeEntry
```
bufdev, line 303: `TreeEntries`, or `Nodes` if renamed

private/pkg/git/tree.go
```diff
@@ -0,0 +1,93 @@
...
+
+func (t *tree) Entries() []TreeEntry {
+	return t.entries
+}
+
+func (t *tree) Traverse(objectReader ObjectReader, names ...string) (TreeEntry, error) {
+	return traverse(objectReader, t, names...)
+}
+
+func parseTree(hash Hash, data []byte) (*tree, error) {
```
bufdev, line 40: below the struct definition, before the methods

private/pkg/git/tree.go
```diff
@@ -0,0 +1,93 @@
...
+	t := &tree{
+		hash: hash,
+	}
+	for len(data) > 0 {
+		i := bytes.Index(data, []byte{0})
+		if i == -1 {
+			return nil, errors.New("malformed tree")
+		}
+		length := i + 1 + hashLength
+		entry, err := parseTreeEntry(data[:length])
```
bufdev, line 50: could use some comments as to the math here - overall, we could use with a bunch of inline documentation for how this all works, for ourselves to refer to in two years heh

private/pkg/git/tree.go
```diff
@@ -0,0 +1,93 @@
...
+	return t, nil
+}
+
+func traverse(
+	objectReader ObjectReader,
+	root Tree,
+	names ...string,
+) (TreeEntry, error) {
+	name := names[0]
+	names = names[1:]
```
bufdev, line 66: same thing about inline docs...assume this comment persists throughout the PR

private/pkg/git/tree_entry.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (e *treeEntry) Mode() FileMode {
+	return e.mode
+}
+
+func (e *treeEntry) Hash() Hash {
+	return e.hash
+}
+
+func parseTreeEntry(data []byte) (*treeEntry, error) {
```
bufdev, line 42: below struct

private/pkg/git/tree_test.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/base64"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestParseTree(t *testing.T) {
+	/*
```
bufdev, line 26: `t.Parallel()`

private/pkg/git/tree_entry_test.go
```diff
@@ -0,0 +1,74 @@
...
+package git
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestParseFileMode(t *testing.T) {
+	tests := []struct {
```
bufdev, line 24: `t.Parallel()`, all tests should be parallel unless there is a reason not to (and I thought we had a linter for this, but I guess its off?)

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,185 @@
...
+		data, err := b.objectReader.Blob(entry.Hash())
+		if err != nil {
+			return nil, err
+		}
+		return &namedReader{
+			info:   b.newObjectInfo(path),
+			reader: bytes.NewReader(data),
+		}, nil
+	case git.ModeSymlink:
+		// Symlinks are stored as blobs that reference the target path as a relative
```
bufdev, line 63: We don't always want to follow symlinks, its a security issue. For local storage, we have a system of options that feed down from the CLI  entry point, where we do want to follow symlinks, but are not enabled in core. We should evaluate what we want to do here.

private/pkg/storage/storagegit/storagegit.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// Provider provides storage buckets for a git repository.
+type Provider interface {
+	// NewReadBucketForTreeHash returns a new ReadBucket that represents
+	// the state of the working tree at the particular tree.
+	//
+	// Typically, callers will want to source the tree from a commit, but
+	// they can also use a subtree of another tree.
+	NewReadBucketForTreeHash(hash git.Hash) (storage.ReadBucket, error)
```
bufdev, line 29: `NewReadBucket`, all the other functions in `storage*` don't have qualifiers

private/pkg/storage/storagegit/provider.go
```diff
@@ -0,0 +1,41 @@
...
+// limitations under the License.
+
+package storagegit
+
+import (
+	"github.com/bufbuild/buf/private/pkg/git"
+	"github.com/bufbuild/buf/private/pkg/storage"
+)
+
+type provider struct {
```
bufdev, line 22: You'll likely have a `symlinks` option and `symlinksIfSupported` option added to `NewReadBucket`, similar to OS...but there might be a reason that symlinks are safe with git, but I'm not sure without research

private/pkg/git/tree.go
unmultimedio, line 66: for sanity check, make sure `len(names) >= 2`

private/pkg/git/git.go
```diff
@@ -268,3 +272,49 @@ func OpenObjectReader(
...
+	// Traverse walks down a tree, following the name-path specified,
+	// and returns the terminal TreeEntry. If no entry is found, it returns
+	// ErrSubTreeNotFound.
+	Traverse(objectReader ObjectReader, names ...string) (TreeEntry, error)
+}
+
+// TreeEntry is a reference to an object contained in a tree. These objects have
+// a file mode associated with them, which hints at the type of object located
+// at ID (tree or blob).
+type TreeEntry interface {
```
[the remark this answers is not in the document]
saquibmian: That's what `go-git` called it, but I can change this to `TreeNode`.

private/pkg/git/tree_entry.go
```diff
@@ -0,0 +1,82 @@
...
+	switch FileMode(mode) {
+	case ModeFile:
+	case ModeExe:
+	case ModeDir:
+	case ModeSymlink:
+	case ModeSubmodule:
+	default:
+		return 0, fmt.Errorf("unknown file mode: %o", mode)
+	}
+	return FileMode(mode), nil
```
unmultimedio: Nit:
```go
fileMode, ok := supportedFileModes[mode]
if !ok {
  return 0, fmt.Errorf("unknown file mode: %o", mode)
}
return fileMode, nil
```

private/pkg/git/tree_entry_test.go
```diff
@@ -0,0 +1,74 @@
...
+			desc: "submodule",
+			mode: ModeSubmodule,
+			txt:  "160000",
+		},
+	}
+
+	for i := range tests {
+		test := tests[i]
+		t.Run(test.desc, func(t *testing.T) {
+			t.Parallel()
```
unmultimedio: ❤️

private/pkg/storage/storagegit/bucket.go
saquibmian: Yes, that was an open question I meant to ask in this PR. I think I will add back the symlink options.

private/pkg/storage/storagegit/provider.go
saquibmian: Yeah I looked into it, and we have the same issues, Git will restore symlinks to the host fs and they may link to files outside of the repo, so same security issues there. I erroneously thought we always enable them, didn't notice the `--disable-symlinks` option. I'll add that back in.

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,185 @@
...
+	case git.ModeFile, git.ModeExe, git.ModeSymlink:
+		return b.newObjectInfo(path), nil
+	default:
+		// TODO: should this be an error? What kind of error can we throw here?
+		return nil, storage.NewErrNotExist(path)
+	}
+}
+
+func (b *bucket) Walk(ctx context.Context, prefix string, f func(storage.ObjectInfo) error) error {
+	walkChecker := storageutil.NewWalkChecker()
```
unmultimedio, line 97: TIL

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,185 @@
...
+		if err != nil {
+			return err
+		}
+		root = subTree
+	}
+	return walkTree(root, objectReader, prefix, walkFn)
+}
+
+func walkTree(
+	root git.Tree,
```
unmultimedio, line 131: Nit: this is not always the `root`, it's the current `parentNode`

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,185 @@
...
+	objectReader git.ObjectReader,
+	prefix string,
+	walkFn func(string, git.TreeEntry) error,
+) error {
+	for _, entry := range root.Entries() {
+		path := normalpath.Join(prefix, entry.Name())
+		switch entry.Mode() {
+		case git.ModeFile, git.ModeExe, git.ModeSymlink:
```
unmultimedio, line 139: Probably an option in the bucket constructor that defines if we should follow symlinks or not?

unmultimedio (commented):
Looking very nice, left mainly nits

private/pkg/git/tree.go
```diff
@@ -47,26 +35,38 @@ func parseTree(hash Hash, data []byte) (*tree, error) {
 			return nil, errors.New("malformed tree")
 		}
 		length := i + 1 + hashLength
-		entry, err := parseTreeEntry(data[:length])
+		entry, err := parseTreeNode(data[:length])
```
unmultimedio, line 38:
```suggestion
		node, err := parseTreeNode(data[:length])
```

private/pkg/storage/storagegit/bucket.go
```diff
@@ -131,9 +131,9 @@ func walkTree(
 	root git.Tree,
 	objectReader git.ObjectReader,
 	prefix string,
-	walkFn func(string, git.TreeEntry) error,
+	walkFn func(string, git.Node) error,
 ) error {
-	for _, entry := range root.Entries() {
+	for _, entry := range root.Nodes() {
```
unmultimedio, line 136:
```suggestion
	for _, node := range root.Nodes() {
```

private/pkg/storage/storagegit/provider.go
saquibmian: Done

private/pkg/git/git.go
```diff
@@ -28,9 +28,30 @@ import (
...
+	ModeFile FileMode = 010_0644
+	// ModeExec is a blob that should be written with the executable bit set.
+	ModeExe FileMode = 010_0755
+	// ModeDir is a tree to be unpacked as a subdirectory in the current
+	// directory.
+	ModeDir FileMode = 004_0000
+	// ModeSymlink is a blob with its content being the path linked to.
+	ModeSymlink FileMode = 012_0000
+	// ModeSubmodule is a commit that the submodule is checked out at.
+	ModeSubmodule FileMode = 016_0000
```
unmultimedio, line 46: Do we need the left padded zeroes in all of them? I think at least one zero can be removed in all of them?

private/pkg/git/git.go
```diff
@@ -28,9 +28,30 @@ import (
...
 )
 
+var ErrSubTreeNotFound = errors.New("subtree not found")
+
+// FileMode is how to interpret a tree node's object. See the Mode* constants
+// for how to interpret each mode value.
+type FileMode uint32
```
unmultimedio, line 53:
```suggestion
type ObjectMode uint32
```

So when declaring `ModeFile FileMode = 010_0644` does not read as cryptic.

private/pkg/git/git.go
```diff
@@ -300,3 +325,27 @@ func OpenObjectReader(
...
+// Tree is a git tree, which are a manifest of other git objects, including other trees.
+type Tree interface {
+	// Hash is the Hash for this Tree.
+	Hash() Hash
+	// Nodes is the set of nodes in this Tree.
+	Nodes() []Node
+	// Traverse walks down a tree, following the name-path specified,
+	// and returns the terminal Node. If no node is found, it returns
+	// ErrSubTreeNotFound.
+	Traverse(objectReader ObjectReader, names ...string) (Node, error)
```
unmultimedio, line 338: Could this be instead

```suggestion
	NodeAt(objectReader ObjectReader, path string /* relative to the tree */) (Node, error)
```

Which returns the terminal node, be it a file, dir, or whatever type it is, or a `NotFound` instead of the new `ErrSubTreeNotFound`? (I see you're only using it to hint a `storage.NewNotExistErr`)

Internally you could do all the `path.Components` dance.

private/pkg/git/object_reader.go
```diff
@@ -82,12 +82,24 @@ func (o *objectReader) Close() error {
 	)
 }
 
-func (o *objectReader) Commit(id Hash) (Commit, error) {
-	data, err := o.read("commit", id)
+func (o *objectReader) Blob(hash Hash) ([]byte, error) {
+	return o.read("blob", hash)
```
unmultimedio, line 86: Consider having this one, and commit, and tag, as global consts

private/pkg/git/tree.go
```diff
@@ -0,0 +1,110 @@
...
+		// and slice to the index of \0 + the length of a hash.
+		// That gives us a single node.
+		i := bytes.Index(data, []byte{0})
+		if i == -1 {
+			return nil, errors.New("malformed tree")
+		}
+		length := i + 1 + hashLength
+		node, err := parseTreeNode(data[:length])
+		if err != nil {
+			return nil, fmt.Errorf("malformed tree: %w", err)
```
unmultimedio, line 48:
```suggestion
			return nil, fmt.Errorf("parse node: %w", err)
```

private/pkg/git/tree.go
```diff
@@ -0,0 +1,110 @@
...
+	names ...string,
+) (Node, error) {
+	// split by the name of the next node we're looking for
+	// and the names of the descendant nodes
+	name := names[0]
+	if len(names) >= 2 {
+		names = names[1:]
+	} else {
+		names = nil
+	}
```
unmultimedio, line 80: I know I suggested receiving the whole path, but these renames still apply.
```suggestion
func traverse(
	objectReader ObjectReader,
	root Tree,
	pathComponents ...string,
) (Node, error) {
	if len(pathComponents) == 0 {
		return nil, errors.New("empty path components") // avoid panic
	}
	// split by the name of the next node we're looking for
	// and the names of the descendant nodes
	targetNode := pathComponents[0]
	var nextTargetNodes []string
	if len(pathComponents) >= 2 {
		nextTargetNodes := pathComponents[1:]
	}
```

private/pkg/git/tree_node.go
```diff
@@ -0,0 +1,86 @@
...
+		data is in the format
+			<mode><space><name>\0<hash>
+	*/
+	modeAndName, hash, found := bytes.Cut(data, []byte{0})
+	if !found {
+		return nil, errors.New("malformed node")
+	}
+	parsedHash, err := newHashFromBytes(hash)
+	if err != nil {
+		return nil, fmt.Errorf("malformed git tree node: %w", err)
```
unmultimedio, line 41:
```suggestion
		return nil, fmt.Errorf("malformed node hash: %w", err)
```

private/pkg/git/tree_test.go
```diff
@@ -0,0 +1,57 @@
...
+	assert.Len(t, tree.Nodes(), 3)
+	assert.Equal(t, tree.Nodes()[0].Hash().Hex(), "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391")
+	assert.Equal(t, tree.Nodes()[0].Name(), "a.proto")
+	assert.Equal(t, tree.Nodes()[0].Mode(), ModeFile)
+	assert.Equal(t, tree.Nodes()[1].Hash().Hex(), "e69de29bb2d1d6434b8b29ae775ad8c2e48c5391")
+	assert.Equal(t, tree.Nodes()[1].Name(), "b")
+	assert.Equal(t, tree.Nodes()[1].Mode(), ModeExe)
+	assert.Equal(t, tree.Nodes()[2].Hash().Hex(), "5c4c3b5f86fa2060081873bf2f06973dd139832b")
+	assert.Equal(t, tree.Nodes()[2].Name(), "c")
+	assert.Equal(t, tree.Nodes()[2].Mode(), ModeDir)
```
unmultimedio, line 57: Optional, could you go into `c` and parse that three as well to get `d.proto`?

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+}
+
+func (b *bucket) newObjectInfo(path string) storage.ObjectInfo {
+	return storageutil.NewObjectInfo(
+		normalpath.Normalize(path),
+		normalpath.Unnormalize(path),
+	)
+}
+
+type namedReader struct {
```
unmultimedio, line 182: Why `named` and not `objectReader`?

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+		// Symlinks are stored as blobs that reference the target path as a relative
+		// path. We can follow this symlink trivially.
+		data, err := b.objectReader.Blob(node.Hash())
+		if err != nil {
+			return nil, err
+		}
+		path = normalpath.Join(
+			normalpath.Base(path),
+			normalpath.Normalize(string(data)),
+		)
```
unmultimedio, line 78: Surprised this doesn't return optional error. What if the symlink content is not actually a path but a bunch of nonsense bytes? I guess we'd just return `NotFound` eventually, but I thought we could catch it here.

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+	case git.ModeFile, git.ModeExe:
+		return b.newObjectInfo(path), nil
+	case git.ModeSymlink:
+		if !b.symlinks {
+			return nil, storage.NewErrNotExist(path)
+		}
+		return b.newObjectInfo(path), nil
+	default:
+		// TODO: should this be an error? What kind of error can we throw here?
+		return nil, storage.NewErrNotExist(path)
```
unmultimedio, line 103: Submodule or unknown, in both scenarios returning `NotFound` sounds reasonable. And probably add a debug log saying which kind of mode was found?

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+	prefix = normalpath.Normalize(prefix)
+	if prefix != "." {
+		node, err := parent.Traverse(b.objectReader, normalpath.Components(prefix)...)
+		if err != nil {
+			if errors.Is(err, git.ErrSubTreeNotFound) {
+				return storage.NewErrNotExist(prefix)
+			}
+			return err
+		}
+		subTree, err := b.objectReader.Tree(node.Hash())
```
unmultimedio, line 132: Probably make sure that the returned node was a Dir before trying to read the tree?

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+			return err
+		}
+		return f(b.newObjectInfo(path))
+	})
+}
+
+func (b *bucket) walk(
+	parent git.Tree,
+	objectReader git.ObjectReader,
+	prefix string,
```
unmultimedio, line 120: I guess that partial prefixes won't work then, so for a valid dir `foo/bar/baz/*`, the prefix `foo/b` won't walk anything. Unsure how other buckets handle this.

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+		return b.newObjectInfo(path), nil
+	default:
+		// TODO: should this be an error? What kind of error can we throw here?
+		return nil, storage.NewErrNotExist(path)
+	}
+}
+
+func (b *bucket) Walk(ctx context.Context, prefix string, f func(storage.ObjectInfo) error) error {
+	walkChecker := storageutil.NewWalkChecker()
+	return b.walk(b.root, b.objectReader, prefix, func(path string, te git.Node) error {
```
unmultimedio, line 109:
```suggestion
	return b.walk(b.root, b.objectReader, prefix, func(path string, _ git.Node) error {
```

or even further, do you actually need to pass the node to the walk func?
```suggestion
	return b.walk(b.root, b.objectReader, prefix, func(path string) error {
```

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+		case git.ModeFile, git.ModeExe:
+			if err := walkFn(path, node); err != nil {
+				return err
+			}
+		case git.ModeSymlink:
+			if b.symlinks {
+				if err := walkFn(path, node); err != nil {
+					return err
+				}
+			}
```
unmultimedio, line 159: I know we don't "follow symlinks", but should we not walk them also?

private/pkg/storage/storagegit/bucket.go
```diff
@@ -0,0 +1,202 @@
...
+		case git.ModeDir:
+			subTree, err := objectReader.Tree(node.Hash())
+			if err != nil {
+				return err
+			}
+			if err := b.walkTree(subTree, objectReader, path, walkFn); err != nil {
+				return err
+			}
+		case git.ModeSubmodule, git.ModeUnknown:
+			// ignored
```
unmultimedio, line 169: Why not `default`?

unmultimedio (approved):
A bunch of rename suggestions and other nits. Nice work!

private/pkg/storage/storagegit/bucket.go
bufdev: If we don't follow them, we shouldn't walk them, same security concerns

private/pkg/storage/storagegit/storagegit.go
```diff
@@ -0,0 +1,55 @@
...
+// ReadBucket to have symlinks followed.
+func ProviderWithSymlinks() ProviderOption {
+	return func(provider *provider) {
+		provider.symlinks = true
+	}
+}
+
+type ReadBucketOption func(*readBucketOptions)
+
+func ReadBucketWithSymlinksEnabled() ReadBucketOption {
```
bufdev, line 46: Needs Godoc, I would also match naming of storageos

bufdev (approved):
Looks generally good to me! One comment, then defer to @unmultimedio for the rest.

private/pkg/git/tree_test.go
saquibmian: I could, but that would be another tree object, so that becomes redundant?

private/pkg/storage/storagegit/bucket.go
saquibmian: It's an objectinfo + reader, so the name felt appropriate. I'm cool with calling it anything, really :)

private/pkg/git/tree_test.go
unmultimedio: Yeah you're right

private/pkg/storage/storagegit/bucket.go
saquibmian: You're right, I don't. Let me clean up.

unmultimedio (approved):

PR #2115, opened 2023-05-24
Don't warn WKT imports

unmultimedio:


private/buf/cmd/buf/buf_test.go
```diff
@@ -2419,7 +2419,19 @@ func testRunStdout(t *testing.T, stdin io.Reader, expectedExitCode int, expected
...
+		func(use string) *appcmd.Command { return NewRootCommand(use) },
+		expectedExitCode,
+		expectedStderr,
+		internaltesting.NewEnvFunc(t),
+		stdin,
+		args...,
+	)
+}
+
+func testRunStdoutStderrNoWarn(t *testing.T, stdin io.Reader, expectedExitCode int, expectedStdout string, expectedStderr string, args ...string) {
```
unmultimedio, line 2434: This func doesn't output warnings. It sends `--no-warn` flag.

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -268,6 +272,32 @@ func (b *builder) warnInvalidImports(
...
 }
 
+var (
+	wktFiles map[string]struct{} // populated in isWKTFile func
+	wktOnce  sync.Once
+)
+
+func isWKTFile(filepath string) bool {
+	wktOnce.Do(func() {
+		wktFiles = map[string]struct{}{
```
saquibmian, line 282: We actually have this data already, it's just unexported in `datawkt`. I wonder if we could change `wkt-go-data` to create a new exported `FilePaths []string` variable that contains this.

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -268,6 +272,32 @@ func (b *builder) warnInvalidImports(
 	return nil
 }
 
+var (
```
bufdev, line 275: global vars always go at the top of a file (but in this case, they will be deleted)

private/bufpkg/bufimage/bufimagebuild/builder.go
bufdev: https://github.com/bufbuild/buf/blob/main/private/gen/data/datawkt/datawkt.gen.go#L158

bufdev (changes requested):

private/bufpkg/bufimage/bufimagebuild/builder.go
saquibmian: Actually yeah, that's basically exactly this function.

private/buf/cmd/buf/buf_test.go
bufdev: when did we add a `--no-warn` flag? is this part of the buf CLI?

private/bufpkg/bufimage/bufimagebuild/builder.go
bufdev: it in fact IS exactly this function :-)

saquibmian (approved):

private/buf/cmd/buf/buf_test.go
bufdev: oh got it, ok

bufdev (approved):

PR #2117, opened 2023-05-24
Add `git.TagIterator`

saquibmian:


doriable (approved):
Looks good to me, seems in-line with the other iterator implementations.

private/pkg/git/tag_iterator.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+		hashBytes = bytes.TrimSuffix(hashBytes, []byte{'\n'})
+		hash, err := parseHashFromHex(string(hashBytes))
+		if err != nil {
+			return err
+		}
+		// Tags are either annotated or lightweight. Depending on the type,
+		// they are stored differently. First, we try to load the tag
+		// as an annnotated tag. If this fails, we try a commit.
+		// Finally, we fail.
```
unmultimedio, line 79: Subjective, but I'd say lightweight tags are much more common than annotated ones? Consider switching order, first trying commits.

private/pkg/git/tag_iterator.go
```diff
@@ -0,0 +1,99 @@
...
+		if err == nil {
+			return f(tagName, hash)
+		}
+		if !errors.Is(err, errObjectTypeMismatch) {
+			return err
+		}
+		return fmt.Errorf(
+			"failed to determine target of tag %q; it is neither a tag nor a commit",
+			tagName,
+		)
```
unmultimedio, line 97: Weird if we ever see this.

unmultimedio (commented):
A late review, sorry!

private/pkg/git/tag_iterator.go
doriable: Yeah, I think the intention is that we are unlikely to hit this code path, but would still need to return an error here. I guess we could add a comment to the effect of `// should not hit this`. I suspect this error would mostly be to help us catch bugs and is useful to users on the off-chance there is some corruption in the `.git` directory.

PR #2118, opened 2023-05-24
Add parallelttest linter

bufdev:


.golangci.yml
```diff
@@ -189,3 +189,31 @@ issues:
...
       path: private/bufpkg/bufimage/bufimageutil/.*_test\.go
+    # This test shouldn't run in parallel as it allocates a lot of memory.
+    - linters:
+        - paralleltest
+      text: "missing the call to method parallel"
+      path: private/bufpkg/bufimage/bufimagebuild/builder_unix_test.go
+    # These tests calls helper functions that call t.Parallel() without creating
+    # subtests with t.Run(...). The linter fails to recognize that they already
+    # run in parallel.
+    - linters:
```
bufdev, line 200: Add comments to each justifying the exception

.golangci.yml
```diff
@@ -188,3 +189,50 @@ issues:
...
+        - paralleltest
+      text: "missing the call to method parallel"
+      path: private/buf/cmd/buf/workspace_test.go
+    # Parallelizing TestPlainPostHandlerTLS and TestPlainPostHandlerH2C
+    # makes this test flaky.
+    - linters:
+        - paralleltest
+      text: "missing the call to method parallel"
+      path: private/bufpkg/bufstudioagent/bufstudioagent_test.go
```
oliversun9, line 238: This test is already flaky in `main`. Run this loop locally
```
for i in {0..20}
do
    go test -timeout 30s -count=1 github.com/bufbuild/buf/private/bufpkg/bufstudioagent
done
```
and some of them will timeout. The error looks like: [this failure on pipeline for this branch](https://github.com/bufbuild/buf/actions/runs/5092280933/jobs/9153425478)

.golangci.yml
```diff
@@ -17,8 +17,8 @@ linters-settings:
       # Ban debug statements
       - '^fmt\.Print'
       - '^log\.'
-      - '^print$'
-      - '^println$'
+      - "^print$"
```
bufdev, line 20: Revert

.golangci.yml
```diff
@@ -188,3 +189,50 @@ issues:
...
+    # because they all call createZipFromDir on the same path.
+    - linters:
+        - paralleltest
+      text: "missing the call to method parallel"
+      path: private/buf/cmd/buf/workspace_test.go
+    # Parallelizing TestPlainPostHandlerTLS and TestPlainPostHandlerH2C
+    # makes this test flaky.
+    - linters:
+        - paralleltest
+      text: "missing the call to method parallel"
```
bufdev, line 237: Why do we need to match on text here? I think this can be removed

unmultimedio (approved):
All of this lgtm, I'd update the branch and file an issue for the issue that @oliversun9 points out, before merging.

oliversun9:
> All of this lgtm, I'd update the branch and file an issue for the issue that @oliversun9 points out, before merging.

Done, but I don't see the changes I just pushed on this PR for some reason, which seems like a GitHub problem

oliversun9:
@bufdev This PR is ready for merging (GitHub doesn't let me request your review because this is your branch).

.golangci.yml
bufdev: This question wasn't resolved, please don't mark as resolved unless it is.

.golangci.yml
```diff
@@ -188,3 +189,49 @@ issues:
...
+      path: private/bufpkg/bufcheck/buflint/buflint_test.go
+    # TestConfigInclusion and TestLicense already call T.parallel(), but the other
+    # tests in this file calls one of testBucketGetFileInfos, testBucketGetAllFileInfosError,
+    # testBucketGetFileInfosForExternalPathsError, and testDocumentationBucket,
+    # all of which call t.Parallel().
+    - linters:
+        - paralleltest
+      text: "missing the call to method parallel"
+      path: private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder_test.go
+    # Every test in this file calls testRunLint, which calls t.Parallel().
```
bufdev, line 211: For this and all other such comments,  move the `t.Parallel()` to the `Test.*` function, so that this ignore is not required.

bufdev (commented):
Blocked until question is resolved and new comment is resolved.

.golangci.yml
oliversun9: This linter also has other checks, such as [tc variable reinitialised](https://github.com/kunwardeep/paralleltest#tparallel-is-called-in-the-range-method-and-test-case-variable-tc-being-used-but-is-not-reinitialised-more-info), ensuring loop variable doesn't get captured by `t.Run`. The text is here to only exempt this specific check for this file.

Sorry about the resolution

oliversun9:
> Blocked until question is resolved and new comment is resolved.

Now it's ready, with the question resolved and new comment resolved.

PR #2122, opened 2023-05-25
Implement `buf alpha repo sync`

saquibmian:
Only base branch is supported today.

`sync` replicates pushed git state to the BSR, walking commits in order. New or moved tags are also pushed.


saquibmian:
Alright, this one is a rough draft, it's just to get it out there for initial assessment.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+	"github.com/bufbuild/connect-go"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+const (
+	errorFormatFlagName      = "error-format"
+	moduleFlagName           = "module"
```
bufdev, line 49: Why not first arg, similar to all other commands?

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+		fmt.Sprintf(
+			"The format for build errors printed to stderr. Must be one of %s",
+			stringutil.SliceToString(bufanalysis.AllFormatStrings),
+		),
+	)
+	flagSet.StringArrayVar(
+		&f.Modules,
+		moduleFlagName,
+		nil,
+		"The module(s) to sync to the BSR; you can provide a module override in the format "+
```
bufdev, line 101: I would expect sync to be congruent with other commands, namely taking a module or workspace as first arg

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+
+func sync(
+	ctx context.Context,
+	container appflag.Container,
+	modules []string,
+	createWithVisibility string,
+) error {
+	// Assume that this command is run from the repository root. If not, `OpenRepository` will return
+	// a dir not found error.
+	repo, err := git.OpenRepository(git.DotGitDir, command.NewRunner())
```
bufdev, line 150: THIS feels like a flag that defaults to .

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+			err,
+		)
+	}
+	_, err = container.Stderr().Write([]byte(
+		fmt.Sprintf("%s:%s\n", moduleIdentity.IdentityString(), pin.Commit)),
+	)
+	return err
+}
+
+func pushOrCreate(
```
bufdev, line 340: This looks copied from push - worth making into a common command in bufcli? Or will this diverge?

saquibmian: It's 90% a copy, yeah, namely the the message sent as part of `PushManifestAndBlobs` is different. I would definitely like to move this to a `bufpush` package.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+	sourceConfig, err := bufconfig.GetConfigForBucket(
+		ctx,
+		sourceBucket,
+	)
+	if err != nil {
+		// We found a module but the module config is invalid. We can warn on this
+		// and carry on. Note that because of resumption, we will typically only come
+		// across this commit once, we will not log this warning again.
+		container.Logger().Warn(
+			"invalid module at commit %s at path %s: %s\n",
```
saquibmian, line 276: lol what... I'll fix these

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+	ctx context.Context,
+	container appflag.Container,
+	clientConfig *connectclient.Config,
+	moduleDir string,
+	moduleIdentityOverride bufmoduleref.ModuleIdentity,
+	repo git.Repository,
+	branch string,
+	commit git.Commit,
+	storagegitProvider storagegit.Provider,
+	knownTags map[string][]string,
```
unmultimedio, line 248: iiuc this func is pushing a single commit, so you could receive only those tags here
```suggestion
	commit git.Commit,
	storagegitProvider storagegit.Provider,
	commitTags []string,
```

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+		moduleIdentity,
+		builtModule,
+		createWithVisibility,
+	)
+	if err != nil {
+		if connect.CodeOf(err) == connect.CodeAlreadyExists {
+			// Module has identical content. The BSR has already created the relevant labels
+			// for us, so we can simply carry on.
+			return nil
+		}
```
unmultimedio, line 324: For this scenario, in case the latest push has the same content, but there may be new tags, should we still try to tag it?

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+		// We failed to push. We fail hard on this because the error may be recoverable
+		// (i.e., the BSR may be down) and we should re-attempt this commit.
+		return fmt.Errorf(
+			"failed to push %s at %s %w",
```
unmultimedio, line 328: Nit
```suggestion
			"failed to push %s at %s: %w",
```

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+
+func push(
+	ctx context.Context,
+	clientConfig *connectclient.Config,
+	repo git.Repository,
+	commit git.Commit,
+	branch string,
+	tags []string,
+	moduleIdentity bufmoduleref.ModuleIdentity,
+	builtModule *bufmodulebuild.BuiltModule,
```
unmultimedio, line 396: I'm almost sure you don't need to pass the built module, only the bucket. I don't see you use anything else than just the bucket in L399.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,473 @@
...
+func NewCommand(
+	name string,
+	builder appflag.Builder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name,
+		Short: "Sync a Git repository to a registry",
+		Long: "Sync a Git repository's commits to a registry in topological order. " +
+			"Only modules specified via '--module' are synced.",
```
unmultimedio, line 64: I'd clarify that we only sync pushed commits to remotes named "origin", and also that it include all tags (even unpushed ones).

unmultimedio (approved):
LGTM, no blockers, only nits 😄

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
saquibmian: The BSR will handle that before it returns `AlreadyExists`. PR for that is open in core.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -0,0 +1,346 @@
...
+		fmt.Sprintf(
+			"The format for build errors printed to stderr. Must be one of %s",
+			stringutil.SliceToString(bufanalysis.AllFormatStrings),
+		),
+	)
+	flagSet.StringArrayVar(
+		&f.Modules,
+		moduleFlagName,
+		nil,
+		"The module(s) to sync to the BSR; you can provide a module override in the format "+
```
unmultimedio, line 101: Is this module dirs or module identities? I mean:

```
--module some/repo/dir:buf.build/user/overridemodule

or

--module buf.build/user/somelocalbufyaml:buf.build/user/overridemodule
```

unmultimedio: There's an interesting challenge here that we've thought at some point when syncing third-party modules. What to do, or how to handle commits in which the module dir changes... or even worse, when it's swapped with other module dir.

private/buf/bufsync/bufsync.go
```diff
@@ -0,0 +1,100 @@
...
+// identity override is optional.
+//
+// This option can be provided multiple times to sync multiple distinct modules.
+func SyncerWithModule(dir string, identityOverride bufmoduleref.ModuleIdentity) SyncerOption {
+	return func(s *syncer) error {
+		for _, existingModule := range s.modulesToSync {
+			if existingModule.dir != dir {
+				continue
+			}
+			if identityOverride == nil && existingModule.identityOverride == nil {
```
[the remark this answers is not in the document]
saquibmian: That's correct, these are distinct. See the dir guard right before this.

private/buf/bufsync/syncer.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/bufbuild/buf/private/pkg/storage"
+	"github.com/bufbuild/buf/private/pkg/storage/storagegit"
+	"go.uber.org/zap"
+)
+
+type syncer struct {
+	logger             *zap.Logger
+	repo               git.Repository
+	storageGitProvider storagegit.Provider
+	modulesToSync      []syncableModule
```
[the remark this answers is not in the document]
saquibmian: You're right, but I want to add more data to that struct eventually, like the sync restart point.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
saquibmian: Module dir. This in fact lets us handle the first case you've identified, but if the dir is swapped, we're out of luck.

### Example 1 (handled)

```
commit a: 
- module in dir `proto` points to `buf.build/bufbuild/bufapis`
commit b: 
- module in dir `proto` points to `buf.build/bufbuild/buf`
commit c: 
- module in dir `protos` points to `buf.build/bufbuild/buf`
```

To sync this "properly", the user would want to run
```
buf sync --module proto:buf.build/bufbuild/buf --module protos:buf.build/bufbuild/buf
```
 and we'll "follow" the module how they intended. "follow" because it's not strictly a follow, but the end result is the same.

### Example 2 (handled)

```
commit a: 
- module in dir `proto` points to `buf.build/bufbuild/buf`
commit b: 
- module in dir `proto/buf` points to `buf.build/bufbuild/buf`
- module in dir `proto/audit` points to `buf.build/bufbuild/audit`
```
To sync this "properly", the user would want to run
```
buf sync --module proto:buf.build/bufbuild/buf --module proto/buf:buf.build/bufbuild/buf --module proto/audit:buf.build/bufbuild/audit
```

### Example 3 (unhandled)

```
commit a: 
- module in dir `proto/buf` points to `buf.build/bufbuild/buf`
- module in dir `proto/audit` points to `buf.build/bufbuild/audit`
commit b: switch identity 
- module in dir `proto/audit` points to `buf.build/bufbuild/buf`
- module in dir `proto/buf` points to `buf.build/bufbuild/audit`
```

We cannot follow this on our flags alone, we would have to look at the commit and see if the `buf.yaml` moved, and try to follow that. Following the `buf.yaml` might be more resilient long term, but if the module identity is changed at the same time, this can get really tricky. Eventually if this becomes a thing, we may have to introduce a `buf.sync.yaml` for cases like this.

Related: we cannot sync on module identity alone, as people may have test files with the same identity, we don't want to sync those. We could exclude certain dirs from being walked, but that looks error prone. I suspect we'll see some of this come up during the alpha and we can re-evaluate then.

private/buf/bufsync/bufsync.go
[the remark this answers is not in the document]
unmultimedio: Oh you're right, I missed that dir guard!

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
unmultimedio: Thanks for the detailed explanation, yeah those edge cases will be tricky if/when they present themselves. Hopefully users don't switch their proto dirs too often with _another_ proto dir.

A `buf.sync.yaml` might sound like the perfect solution, but I think the whole reason of why `buf sync` is needed is to recover previous history into the BSR, which is most likely not buf-aware, and won't have such file in the history.

unmultimedio (approved):
💯

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
saquibmian: Yeah we'd read `buf.sync.yaml` from the `origin/head`, but we can come to that if needed.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
bufdev: You could make the same arguments for most of our commands (buf lint could take multiple inputs), but there's no reason to, and deviating from that standard without a clear reason to do so makes the UX inconsistent. Is there a clear reason to do so here?

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
bufdev: Comments should be resolved by the creator when there is a design disagreement

private/buf/bufsync/syncer.go
```diff
@@ -0,0 +1,174 @@
...
+	builtModule, err := bufmodulebuild.BuildForBucket(
+		ctx,
+		sourceBucket,
+		sourceConfig.Build,
+	)
+	if err != nil {
+		// We failed to build the module. We can warn on this
+		// and carry on. Note that because of resumption, we will typically only come
+		// across this commit once, we will not log this warning again.
+		s.logger.Warn(
```
bufdev, line 156: This is something that is conceivably part of control flow - it is conceivable that the caller of `bufsync` will want to know that a module was not built, and take some type of action. By logging here, we've effectively swallowed the warning. Instead, such data should be returned up the stack in a manner that allows the caller to decide what to do (including potentially log it)

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
bufdev: Please do so

saquibmian: This has diverged, calling an entirely different RPC now.

bufdev (approved):

private/buf/bufsync/bufsync.go
```diff
@@ -0,0 +1,144 @@
...
+	// the identity specified in the module config file will be used.
+	//
+	// Unnamed modules will not have their identity overridden, as they are
+	// not pushable.
+	IdentityOverride() bufmoduleref.ModuleIdentity
+}
+
+// NewModule constructs a new module that can be synced with a Syncer.
+func NewModule(dir string, identityOverride bufmoduleref.ModuleIdentity) (Module, error) {
+	path, err := normalpath.NormalizeAndValidate(dir)
```
saquibmian, line 64: I should move this into `newSyncableModule`. Let me do it up.

PR #2128, opened 2023-05-29
Add beta plugin command (remove alpha plugin)

mfridman:
This PR removes the old `buf alpha plugin {push,delete}` commands and promotes them to `buf beta plugin {push,delete}`.

Notable changes:

1. The host is now `buf.build` instead of `plugins.buf.build` (drop subdomain)
2. The plugin visibility is now mandatory, similar to `buf beta registry repository create` command

We might want to nest the `buf beta plugin` command under registry, like `buf beta registry plugin {push,delete}` .. but this is less an API command and more like `buf push`, so there's an argument to keep it under `buf beta plugin push` and eventually promote it to just `buf plugin push`


private/buf/cmd/buf/buf.go
```diff
@@ -146,6 +146,14 @@ func NewRootCommand(name string) *appcmd.Command {
 					stats.NewCommand("stats", builder),
 					migratev1beta1.NewCommand("migrate-v1beta1", builder),
 					studioagent.NewCommand("studio-agent", noTimeoutBuilder),
+					{
+						Use:   "plugin",
+						Short: "Manage plugins on the Buf Schema Registry",
+						SubCommands: []*appcmd.Command{
+							pluginpush.NewCommand("push", builder),
+							curatedplugindelete.NewCommand("delete", builder),
```
pkwarren, line 154:
```suggestion
							plugindelete.NewCommand("delete", builder),
```

Maybe as part of this we remove the old `buf beta registry plugin` and `buf beta registry template` commands since the old RPCs no longer support making changes?

pkwarren (approved):

Issue #2130, opened 2023-05-30
Add emit-defaults flag for `buf curl`


torkelrogstad:
I'd like an option (maybe even make it the default?) to emit fields with default values set, in the same vein as the `-emit-defaults` flag that `grpcurl` has. I'd be happy to code this up, if the PR would have a chance at being accepted.

torkelrogstad:
Wow, incredibly turnaround time! Thanks a lot.

bufdev:
This has been released in v1.20.0 https://github.com/bufbuild/buf/releases/tag/v1.20.0

PR #2132, opened 2023-05-30
Add emit-defaults flag to buf curl

bufdev:
Fixes https://github.com/bufbuild/buf/issues/2130.

The actual flag name we could revisit, and this just hacks it into the code, but it's pretty easy and makes sense to me.

### DIFF
--- a/private/buf/cmd/buf/command/curl/curl.go
+++ b/private/buf/cmd/buf/command/curl/curl.go
@@ -226,9 +226,9 @@ type flags struct {
 	NetrcFile string
 	Headers   []string
 	Data      string
-	Output    string
 
-	// Other options
+	// Output options
+	Output       string
 	EmitDefaults bool
 
 	// so we can inquire about which flags present on command-line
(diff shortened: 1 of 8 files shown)


private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -225,6 +228,9 @@ type flags struct {
 	Data      string
 	Output    string
 
+	// Other options
+	EmitDefaults bool
```
jhump, line 232: Ditto
```suggestion

	// Output options
	Output       string
	EmitDefaults bool
```

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -93,6 +93,9 @@ const (
 	dataFlagShortName      = "d"
 	outputFlagName         = "output"
 	outputFlagShortName    = "o"
+
+	// Other flags
+	emitDefaultsFlagName = "emit-defaults"
```
jhump, line 98: Maybe call it "Output flags" and move the above `outputFlag*` consts down to this section?
```suggestion

	// Output flags
	outputFlagName       = "output"
	outputFlagShortName  = "o"
	emitDefaultsFlagName = "emit-defaults"
```

jhump (approved):

jhump (approved):

PR #2139, opened 2023-05-31
Remove template version create command

pkwarren:
Update the Buf CLI to remove the 'buf beta registry template version create' command, which is no longer supported on the BSR (returns unimplemented).


mfridman (approved):
@bufdev we're going to clean up some of the unused (legacy remote generation) commands that are unimplemented or no longer apply, did you want to also review these changes?

mfridman (approved):

PR #2140, opened 2023-05-31
Rearrange packages and initial setup for buf.gen.yaml V2 implementation

oliversun9:
This PR rearranges some files and packages to prepare for the main implementation.
`bufgen` now looks like:
```
private/buf/bufgen
├── bufgenv1
│   └── code from the old bufgen package, with some of it moved to the new bufgen
├── bufgenv2
│   └── code from the prototype implementation, mostly untouched
└── code shared by both v1 and v2, and exported types / functions and etc.
```
It moved most stuff from `bufgen` to `bufgenv1`, keeping some common utilities in `bufgen`. There are also stuff that might be shared by both V2 and V1 but still moved to `bufgenv1`. They will be moved to `bufgen` for sharing when I get to that part of the implementation. For example, some of `generator.go` and `image_provider.go` might be reusable.

Removed old interface `Provider`, created `ConfigDataProvider`, so that the interface is not coupled with a specific config format and can be shared by both versions.

Add type alias `ConfigGetter[V any]`, which is a function that understands a certain config structure, interprets bytes as that structure and returns a certain field or the entire config.

`ReadFromConfig` takes a `ConfigGetter` as one of the parameters, reads the data and pass that to the `ConfigGetter`.


private/buf/bufgen/bufgenv1/config.go
```diff
@@ -0,0 +1,612 @@
...
+		logger,
+		provider,
+		readBucket,
+		getConfig,
+		options...,
+	)
+}
+
+func getConfig(
+	logger *zap.Logger,
```
oliversun9, line 51: the rest of this file is the same as `bufgen/config.go` before

private/buf/bufgen/config.go
```diff
@@ -16,51 +16,89 @@ package bufgen
...
+	data, id, err := provider.GetConfigData(ctx, readBucket)
+	if err != nil {
+		return nil, err
+	}
+	return configGetter(
+		logger,
+		encoding.UnmarshalYAMLNonStrict,
+		encoding.UnmarshalYAMLStrict,
+		data,
+		id,
```
oliversun9, line 88: The `id` here is the reason why interface `ConfigDataProvider`'s `GetConfigData` returns an id.

private/buf/cmd/buf/command/generate/v1.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"github.com/bufbuild/buf/private/bufpkg/bufimage/bufimageutil"
+	"github.com/bufbuild/buf/private/bufpkg/bufwasm"
+	"github.com/bufbuild/buf/private/pkg/app/appcmd"
+	"github.com/bufbuild/buf/private/pkg/app/appflag"
+	"github.com/bufbuild/buf/private/pkg/command"
+	"github.com/bufbuild/buf/private/pkg/storage/storageos"
+)
+
+func runV1(
```
oliversun9, line 35: This is the same as the old `run`

private/buf/bufgen/config_data_provider.go
```diff
@@ -27,20 +26,20 @@ import (
...
+	return &configDataProvider{
 		logger: logger,
 		tracer: otel.GetTracerProvider().Tracer("bufbuild/buf"),
 	}
 }
 
-func (p *provider) GetConfig(ctx context.Context, readBucket storage.ReadBucket) (_ *Config, retErr error) {
-	ctx, span := p.tracer.Start(ctx, "get_config")
+func (p *configDataProvider) GetConfigData(ctx context.Context, readBucket storage.ReadBucket) (_ []byte, _ string, retErr error) {
+	ctx, span := p.tracer.Start(ctx, "get_config_data")
```
oliversun9, line 42: @doriable does it look ok changing from `get_config` to `get_config_data`?

private/buf/bufgen/bufgen.go
```diff
@@ -44,8 +30,16 @@ const (
 	V1Version = "v1"
 	// V1Beta1Version is the string used to identify the v1beta1 version of the generate template.
 	V1Beta1Version = "v1beta1"
+	// V2Version is the string used to identify the v2 version of the generate template.
+	V2Version = "v2"
```
unmultimedio, line 34: Wondering if we should launch with a `v2alpha1` or `v2beta1` first, and then later support an stable v2.

unmultimedio (commented):
Only nits so far, I'm not very familiar with these pkgs but look good to me.

private/buf/bufgen/bufgen.go
oliversun9: That is good point. @bufdev has confirmed before that it is fine to go with `v2` directly

private/buf/bufgen/bufgen.go
```diff
@@ -85,243 +79,15 @@ func (s Strategy) String() string {
 }
 
 // Provider is a provider.
```
saquibmian, line 81:
```suggestion
// ConfigDataProvider is a provider for config data.
```

saquibmian (commented):
Overall LGTM, just a couple small things. Do a quick godoc pass to make sure all exported things are documented. I had to think to grok the new read config flow.

bufdev:
>  There are also stuff that might be shared by both V2 and V1 but still moved to bufgenv1

If in doubt, implement V1 in terms of V2, not V2 in terms of V1, ie bufgenv1 should depend on bufgenv2

private/buf/bufgen/bufgen.go
```diff
@@ -44,8 +30,16 @@ const (
 	V1Version = "v1"
 	// V1Beta1Version is the string used to identify the v1beta1 version of the generate template.
 	V1Beta1Version = "v1beta1"
+	// V2Version is the string used to identify the v2 version of the generate template.
+	V2Version = "v2"
 )
 
+// ExternalConfigVersion defines the subset of all config
+// file versions that is used to determine the configuration version.
+type ExternalConfigVersion struct {
```
bufdev, line 39: Helpers that should only be used by `bufgenv1` and `bufgenv2`, but not outside of these packages should be in `bufgen/internal` - is this used outside of `bufgen/*` or is this just an internal implementation detail?

private/buf/cmd/buf/command/generate/generatev1.go
```diff
@@ -0,0 +1,174 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: Why is this file called `generatev1.go`?

private/buf/cmd/buf/command/generate/generate_test.go
```diff
@@ -150,6 +150,18 @@ func TestOutputFlag(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestVersion(t *testing.T) {
+	testRunStdoutStderr(
+		t,
+		nil,
+		1,
+		``,
+		"Failure: v2 unimplemented",
```
bufdev, line 159: While developing, the error should be the same as any other unknown version, ie we should not be recognizing v2 from the perspective of the user at this point, if this is going to be merged into main

private/buf/cmd/buf/command/generate/generatev2.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: It doesn't make sense to have two separate entrypoints - you need to deduce within the single `run` command what version you have, and make inferences as to behavior based on that.

bufdev (changes requested):

private/buf/cmd/buf/command/generate/generatev1.go
oliversun9: This file is now removed, as per comment on having one entry point

private/buf/cmd/buf/command/generate/generate_test.go
oliversun9: Now treating `v2` as an unknown version

private/buf/bufgen/bufgen.go
oliversun9: `ExternalConfigVersion` is also used in [`bufmigrate`](https://github.com/bufbuild/buf/blob/main/private/buf/bufmigrate/v1beta1_migrator.go#L121). With some dependency injection (`Logger`, `ReadWriteBucket`, `ConfigDataProvider`) into `v1beta1Migrator`, we might be able to hide `ExternalConfigVersion` in `internal` and let the migrator call `ReadConfigVersion`. I left it as a TODO so that I can visit this when writing migrator to v2.

After inspecting `bufgen.go`, only `Strategy` is movable and it's moved to `internal`.
[`ReadFromConfig`](https://github.com/bufbuild/buf/pull/2140/files#diff-855d43fb4faf389939e0de63c9b9eb2a5a7ae640984bf23ded98ff5a2441026aR90) is not moved to `internal`, to avoid circular import, because `ReadFromConfig` has a parameter of type `ConfigDataProvider`, which is defined in `bufgen`.

bufdev:
Just to make sure: there's no external change in behavior here for existing users at this point, right? I want to make sure that bufgenv1 remains the same in terms of functionality.

How do you plan to do the conditional on v1 vs v2 behavior?

oliversun9:
> Just to make sure: there's no external change in behavior here for existing users at this point, right? I want to make sure that bufgenv1 remains the same in terms of functionality.
> 
> How do you plan to do the conditional on v1 vs v2 behavior?

Same functionality: Yes. Detail: 
* in `generate.go` the behavior is exactly the same, because we are not yet using the version after reading it. 
* In `bufgen`, the only thing new is `ReadFromConfig[V any]`, but the only change related to this in the code path is that `readConfig` is extracted and passed as value. Other changes only come from package renaming (`bufgen` to `bufgenv1`).

A high-level description of conditional on V1 and V2's behavior:
`run` in `generate.go` will
1. Get image from input: have some function in `bufgenv2` that maps each input config to a `Ref` (just like the `Ref` obtained from parsing the command-line input). Use the existing `imageConfigReader` to get images from these `Ref`s, just like before, except that we are reading many `Ref`s. Merge these images. `run` will only use the `Ref`s from the config file when input is not present on the command line.
2. Explicitly have a modifier calling `Modify` on this image: `bufgenv1` and `bufgenv2` will be able to build a `Modifier` from `v1` and `v2` managed mode config. `run` will call `NewModifier` from the respective package depending on the version.
3. Have some `Generator` calling `Generate` on this image and a list of plugins:
* Currently: `bufgenv1/generator.go` (previously `bufgen/generator.go`) applies managed mode (but now managed mode modifier is a separate thing), execute plugins and writes the responses. Its plugin exec logic accesses v1 plugin configs' fields, and sometimes determines the type of plugin by looking at whether a field is set (this resulted in having to do this kind of check in multiple places, and it is also hard to read).
* Plan: Have a generator with largely the same logic as before, doing plugin exec and response writing, but taking in plugin configs as interfaces. For example, 
```
type PluginConfig interface {
	Out() string
	Opt() string
	pluginConfig()
}
type LocalPluginConfig interface {
	PluginConfig
	Strategy() Strategy
	localPluginConfig()
}

type BinaryPluginConfig interface {
	LocalPluginConfig
	Path() []string
	binaryPluginConfig()
}
```
And both `bufgenv1` and `bufgenv2`'s `NewPluginConfig` will parse external plugin configs into different types of `PluginConfig`s, and checking a plugi config's type is more reliable.

bufdev:
How are you determining which package to call? For example, you might:

- Read the config in
- If v2 -> bufgenv2 package
- If v1  -> bufgenv1 package
- Prior to this, you can't read inputs etc

I might be missing something but I don't see the plan as it's currently laid out there, it might be somewhere else though.

oliversun9:
> How are you determining which package to call? For example, you might:
> 
> * Read the config in
> * If v2 -> bufgenv2 package
> * If v1  -> bufgenv1 package
> * Prior to this, you can't read inputs etc
> 
> I might be missing something but I don't see the plan as it's currently laid out there, it might be somewhere else though.

Getting the version of the config file isn't version specific or input specific. To get the version, I need the command line flag `--template`, if specified

I just need to pass this value (along with things like logger, read bucket and etc.) to `bufgen.ReadConfigVersion`, which lives in `bufgen`.
I also updated the plan and `run` in `generate.go` to reflect this.

bufdev:
Grab me tomorrow in-person, I'll come by your desk and we can hash this out live - I don't want to be a blocker here, I think it'll be easier than working through it on comments

bufdev (approved):

PR #2141, opened 2023-05-31
Add dag package

bufdev:
Competing idea to https://github.com/bufbuild/buf/pull/2138.

This started ad a vendoring of https://github.com/stevenle/topsort with proper license attribution, plus some cleanups to more closely match our code standards, but evolved into more functionality and more cleanups. We can edit this as we want for our needs. It's a super-simple package, and this is a universal concept.

@unmultimedio any thoughts about starting with this as a base?


bufdev:
I've spent some more time on this in support of the `direct-deps` branch, and I think this is the better way to go here - it's more full-fledged now, and is mostly up to our standards.

unmultimedio:
> I've spent some more time on this in support of the `direct-deps` branch, and I think this is the better way to go here - it's more full-fledged now, and is mostly up to our standards.

Sure! Let's go with this one if you see it works as we need it. What I'm missing from this package (probably I'm not thinking it or checking it through), are these two:

1. how to sort a complete graph, w/out giving it any starting point, something like: sort the whole tree, return the sorted modules on it. Useful for workspace push or sync managed modules. Probably it's enough to add everything under the same "root", and then request to have the "root" node sorted.
2. are non-dependent nodes in the tree lexicographically sorted or any other way to they're deterministic?

unmultimedio:
1. Yeah, that should be enough, thx
2. If `a->b, a->c` is `graph.Sort("a")` always `b, c, a`? or is the order non-deterministic? Because `c, b, a` is also a valid sort.

unmultimedio (approved):

PR #2143, opened 2023-05-31
Remove buf registry beta plugin/template commands

pkwarren:
Alpha plugin/template APIs have been replaced for some time with remote plugins and remote packages. Remove the old 'buf beta registry plugin/template' commands.


private/buf/cmd/buf/buf.go
```diff
@@ -199,7 +197,6 @@ func NewRootCommand(name string) *appcmd.Command {
 								Use:   "plugin",
 								Short: "Manage Protobuf plugins",
 								SubCommands: []*appcmd.Command{
-									plugincreate.NewCommand("create", builder),
```
mfridman, line 202: From the CLI side, can't we remove this entire block under `plugin`?

pkwarren: Removed - PTAL.

mfridman (approved):

bufdev (approved):

PR #2144, opened 2023-05-31
Add `SyncService` for sync-related RPCs

saquibmian:
- Remove git-metadata RPC from Push so that Push behaviour does not diverge for sync concerns.


proto/buf/alpha/registry/v1alpha1/sync.proto
```diff
@@ -0,0 +1,77 @@
...
+message SyncGitCommitRequest {
+  string owner = 1;
+  string repository = 2;
+  string branch = 3;
+  // Manifest with all the module files being pushed.
+  buf.alpha.module.v1alpha1.Blob manifest = 4;
+  // Referenced blobs in the manifest. Keep in mind there is not necessarily one
+  // blob per file, but one blob per digest, so for files with exactly the same
+  // content, you can send just one blob.
+  repeated buf.alpha.module.v1alpha1.Blob blobs = 5;
```
unmultimedio, line 63: Don't quite understand why these are present here.

- Don't seem to be strictly part of any git metadata, it's what's pushed.
- If manifest and blobs matches with latest, it'll attach the git metadata to that latest BSR commit,
- If not, then it'll create a new commit (basically push effect)?

doriable: Based on our conversation, I think we want a single push RPC that also handles the git metadata. So this will be called in lieu of `PushM+B` for sync.

saquibmian: The current patterns we have around this are based on where we're coming from, not where we're going. Pretty soon we'll allow tags to move, so `push --tag <foo>`, even with identical content, should move the tag to the latest commit. Internally that may call `Latest` + either `Push` or `SetTag`, but that RPC layout is suboptimal.

I think where we're headed is a world in which everyone uses `buf sync` exclusively and `push --tag` becomes obsolete. `push` would remain for bespoke/advanced use cases where you need to fall down to the primitive, and this would be used in cases where you're not syncing something, per se (like dynamically downloaded and pushed schemas, somewhat like managed modules).

PR #2145, opened 2023-05-31
Improve error message in plugin push command

mfridman:
Small improvement to the error string we surface to unauthenticated users trying to interact with the OCI registry.

Instead of:

```
Failure: GET https://{remote}/v2/{name}/tags/list?n=1000: unexpected status code 401 Unauthorized
```

It's now:

```
Failure: you are not authenticated. For details, visit https://buf.build/docs/bsr/authentication
```

### DIFF
--- a/private/buf/cmd/buf/command/beta/registry/plugin/pluginpush/pluginpush.go
+++ b/private/buf/cmd/buf/command/beta/registry/plugin/pluginpush/pluginpush.go
@@ -422,7 +422,7 @@ func findExistingDigestForImageID(
 		structuredErr := new(transport.Error)
 		if errors.As(err, &structuredErr) {
 			if structuredErr.StatusCode == http.StatusUnauthorized {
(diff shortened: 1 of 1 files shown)


private/buf/cmd/buf/command/beta/registry/plugin/pluginpush/pluginpush.go
```diff
@@ -421,6 +421,9 @@ func findExistingDigestForImageID(
 	if err != nil {
 		structuredErr := new(transport.Error)
 		if errors.As(err, &structuredErr) {
+			if structuredErr.StatusCode == http.StatusUnauthorized {
+				return "", errors.New("you are not authenticated. For details, visit https://buf.build/docs/bsr/authentication")
```
mfridman, line 425: We'll probably want to update all the https://docs.buf.build references in the future, @rakuista is this on your radar. Do we care?

pkwarren (approved):

PR #2147, opened 2023-06-01
Update to latest connect-go and makego

pkwarren:


make/go/base.mk
```diff
@@ -178,6 +178,9 @@ postgenerate::
 .PHONY: licensegenerate
 licensegenerate::
 
+.PHONY: format
```
mfridman, line 181: I haven't followed upstream changes, but I assume there were some changes..

pkwarren: Yep - these were upstream changes to makego to optimize CI build time and also support Knit.

PR #2148, opened 2023-06-02
Add .c, .cc, .h files for license header

buildbreaker:
Adding `.c`, `.cc`, and `.h` extensions for license headers


Alfus (approved):

Alfus (approved):

private/pkg/licenseheader/licenseheader.go
```diff
@@ -84,6 +84,9 @@ All rights reserved.`,
 		".kt":    "//",
 		".swift": "//",
 		".java":  "//",
+		".cc":    "//",
+		".c":     "//",
+		".h":     "///",
```
bufdev, line 89: Is this right? Just making sure

buildbreaker: @Alfus I was going off of protovalidate-cc's `.h` files. Is the triple slash the way to go?

Alfus: should be double slash, triple is for doc strings

PR #2150, opened 2023-06-02
Add audit event for plugin delete

pkwarren:
User uploaded plugins can be deleted - add an audit event to track this.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -343,6 +345,13 @@ message PayloadCuratedPluginCreated {
   string owner_name = 2;
 }
 
+message PayloadCuratedPluginDeleted {
+  // owner_id is the id of the owner of the plugin.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the plugin.
+  string owner_name = 2;
```
bufdev, line 352: I know this is copied from before, but why are we storing the id and name of the same entity? Also why not storing any info on the plugin itself? I assume there are reasons but am interested.

pkwarren: I'll dig in more - you are right I was mainly taking what was there for create events for delete.

pkwarren: When we create an audit event, it is wrapped in an auditv1alpha1.Event with the Resource set to the plugin id / name, the Actor set to the user id / name, and the payload contains additional information not found elsewhere (owner id / name).

bufdev (approved):

PR #2159, opened 2023-06-05
Test cached module identity

unmultimedio:
Test for https://github.com/bufbuild/buf/pull/2152

### DIFF
--- a/private/bufpkg/bufmodule/bufmodulecache/cas_module_reader_test.go
+++ b/private/bufpkg/bufmodule/bufmodulecache/cas_module_reader_test.go
@@ -228,7 +228,6 @@ func assertModuleIdentity(t *testing.T, module bufmodule.Module, expectedModuleI
 			t, expectedCommit, fileInfo.Commit(),
 			"unexpected commit for file %q", fileInfo.Path(),
 		)
-
 	}
 }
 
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmodulecache/cas_module_reader_test.go
```diff
@@ -212,6 +213,25 @@ func verifyBlobContents(t *testing.T, bucket storage.ReadWriteBucket, basedir st
 	assert.Equal(t, bb.Bytes(), cachedModule)
 }
 
+func assertModuleIdentity(t *testing.T, module bufmodule.Module, expectedModuleIdentity string, expectedCommit string) {
+	require.NotNil(t, module)
+	require.NotEmpty(t, expectedCommit)
+	fileInfos, err := module.SourceFileInfos(context.Background())
+	require.NoError(t, err)
+	for _, fileInfo := range fileInfos {
+		require.NotNil(t, fileInfo.ModuleIdentity())
```
unmultimedio, line 222: I was a little surprised that there was no way to do `module.Identity()`, but I needed to loop over the files. I assume there's no way to access a module identity from a module with no proto files?

bufdev: Modules always have `.proto` files - its invalid to have an empty module. `ModuleFileSets` inherit from `Modules`, and `ModuleFileSets` will have files from different `Modules`, meaning that you need to get identity at the file level given this inheritance.

pkwarren (approved):

PR #2164, opened 2023-06-06
Remote module identity required for sync

saquibmian:
When we sync a local module to a remote, we need to know the remote identity. This is found in the local `buf.yaml`, but that value can change over a module's history. `sync` needs a stable remote identity that it can sync to, so we are making the identity override required.

The UX of this command will be re-evaluated later before we go to beta.


doriable (approved):
This looks good to me :)

PR #2169, opened 2023-06-08
Add ResolveReference protos

joshcarp:
Add protos for ResolveReference. 
This api will return the package version for a plugin + module reference. 

Unsure of if `PluginRef` and `ModuleRef` should be the constituent parts of the module/plugin (owner, name, version), or if the plugin should take the raw string


proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -33,6 +33,10 @@ service ResolveService {
   rpc GetModulePins(GetModulePinsRequest) returns (GetModulePinsResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+
+  rpc ResolveReference(ResolveReferenceRequest) returns (ResolveReferenceResponse) {
```
joshcarp, line 37: Not sure if this should have a more descriptive name relating to registries? maybe something like `ResolveRegistryVersion`?

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +111,24 @@ message GetLocalModulePinsResponse {
...
+  string owner = 1;
+  string name = 2;
+  string version = 3;
+}
+
+message ModuleRef {
+  string owner = 1;
+  string repository = 2;
+  string reference = 3;
+}
```
stefanvanburen, line 130: could/should we use [`LocalModuleReference`](https://github.com/bufbuild/buf/blob/206dd163499d3c287b43718597fbf7775d3b3283/proto/buf/alpha/registry/v1alpha1/module.proto#L21-L29) and [`CuratedPluginReference`](https://github.com/bufbuild/buf/blob/206dd163499d3c287b43718597fbf7775d3b3283/proto/buf/alpha/registry/v1alpha1/plugin_curation.proto#L212-L221) for these? not sure how we feel about using messages across services.

Otherwise, `s/Ref/Reference`

doriable: Is there a reason why we would't need to send the `remote` for the module and plugin references?

joshcarp: i guess you only want to resolve it for one bsr instance, which would be the instance that you're sending the request to?
ie. you _couldn't_ do this:
if you resolve the following:
```
buf beta registry resolve buf.build/bufbuild/eliza petstore.buf.build/petstore/petstore-generator
```

bufdev: We don't have any CLI commands that take multiple inputs, and the rest of our API surface does not take a remote for routing

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +111,24 @@ message GetLocalModulePinsResponse {
   // This includes the transitive deps.
   repeated buf.alpha.module.v1alpha1.ModulePin dependencies = 2;
 }
+
+message ResolveReferenceRequest {
+  PluginRef plugin = 1;
```
bufdev, line 116: `s/Ref/Reference`, and name the fields after their types, ie `plugin_reference`

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +111,24 @@ message GetLocalModulePinsResponse {
...
+  ModuleRef module = 2;
```
bufdev, line 117: Same

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +111,24 @@ message GetLocalModulePinsResponse {
...
+}
+
+message PluginRef {
```
bufdev, line 120: We have a lot of representations for these, is there a reason these are special and aren't reused? Is it because of `remote` (which makes sense)

proto/buf/alpha/registry/v1alpha1/resolve.proto
bufdev: Regardless of the name, some docs on this as to what it is would be good

proto/buf/alpha/registry/v1alpha1/resolve.proto
joshcarp: Yeah

proto/buf/alpha/registry/v1alpha1/resolve.proto
doriable: Yeah, that's valid on the inputs. I guess we would check that the plugin and module input are routing to the same remote, and then we wouldn't need to pass the `remote` onto the API.

joshcarp: so `PluginReference` [already exists](https://github.com/bufbuild/buf/blob/fece8ff2cc6a184815a81551d94859384893e06c/proto/buf/alpha/registry/v1alpha1/generate.proto#L49) in this protobuf package, but it's got a field that I _don't_ want: `parameters`:

```
message PluginReference {
  // The owner of the plugin which identifies the
  // plugins to use with this generation.
  string owner = 1;
  // The name of the plugin which identifies the
  // plugins to use with this generation.
  string name = 2;
  // The plugin version to use with this generation.
  string version = 3;
  // The parameters to pass to the plugin. These will
  // be merged into a single, comma-separated string.
  repeated string parameters = 4;
}
```

I would use [LocalModuleReference](https://github.com/bufbuild/buf/blob/206dd163499d3c287b43718597fbf7775d3b3283/proto/buf/alpha/registry/v1alpha1/module.proto#L21-L29) but it's not referencing a local module? also [CuratedPluginReference](https://github.com/bufbuild/buf/blob/206dd163499d3c287b43718597fbf7775d3b3283/proto/buf/alpha/registry/v1alpha1/plugin_curation.proto#L212-L221) is every plugin that _can_ be uploaded to a BSR a "curated" plugin, because i assume this could refer to any plugin?

stefanvanburen: I think it's "Local" in the sense that it's within the same remote as the plugin (can't have different remotes), and I _think_ it has to be a "curated" plugin to be a part of remote packages - not sure if the recent custom plugin changes make that different though.

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +113,30 @@ message GetLocalModulePinsResponse {
...
+  // The owner of the plugin.
+  // example: bufbuild
+  string owner = 1;
+  // The name of the plugin.
+  // example: connect-go
+  string name = 2;
+  // The version of the plugin.
+  // If empty, this is a reference to the latest version.
+  // example: v1.0.0
+  string version = 3;
```
stefanvanburen, line 136: we also have plugin revisions - I could see either:

1. always using the latest revision (seems reasonable as we only do this generally to fix bugs; there's not much reason to use an older revision)
2. add a separate field for it (could be a `uint32`)
3. include it in the `version` string

what do you think? I think 1 is fine, but if so we might want to document it either on the message or on version.

joshcarp: There have been some issues with old revisions of plugins being broken and old versions being requested, i feel like always getting the latest and documenting is a good idea because it means that there'll be less cached versions everywhere and we have the freedom to update revisions without any client side changes. 

if this ends up being the wrong decision we can add another field for it and it wouldn't be a breaking change

stefanvanburen: agreed, sounds good to me.

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -107,3 +113,30 @@ message GetLocalModulePinsResponse {
...
+message ResolveReferenceRequest {
+  // The plugin reference to resolve.
+  ResolveReferencePlugin plugin_reference = 1;
+  // The module reference to resolve.
+  LocalModuleReference module_reference = 2;
+}
+
+// ResolveReferencePlugin is a plugin reference.
+// If the version is empty, this is a reference to the latest version.
+message ResolveReferencePlugin {
```
joshcarp, line 126: The name `PluginReference` conflicts with the type in https://github.com/bufbuild/buf/blob/4b05ef10993f94c52514c259cee664502bbbbc40/proto/buf/alpha/registry/v1alpha1/generate.proto#L22-L23 which is marked as deprecated.
As per alpha contract I assume this type can change names after release, but still support wire compatibility. 

This name will change name from `ResolveReferencePlugin` to `PluginReference` once the other file is deleted.

stefanvanburen (approved):

PR #2170, opened 2023-06-08
Resume sync from last sync point

saquibmian:
There are some important considerations:
- Sync is resumed _after_ the last known commit for a particular branch.
- Repositories that previously called `buf push` can still take advantage of resumption with `buf sync` if the BSR can find a sync point. The BSR currently looks for a tag on the BSR commit that _looks like_ a Git commit hash. In some cases this may not be correct, and users will have to manually create a new tag with the correct Git commit hash.
- Rebasing is not supported and currently fails the entire sync in a partial state.
- Sync points _must_ be encountered, otherwise sync fails after completion.

### DIFF
--- a/private/buf/bufsync/syncer.go
+++ b/private/buf/bufsync/syncer.go
@@ -157,7 +157,7 @@ func (s *syncer) Sync(ctx context.Context, syncFunc SyncFunc) error {
 		}); err != nil {
 			return fmt.Errorf("process commits: %w", err)
 		}
-		// If we have any sync points left, they were not encountered during sync, which is unexpected behaviour.
+		// If we have any sync points left, they were not encountered during sync, which is unexpected behavior.
 		for module, syncPoint := range syncPoints {
 			if err := s.errorHandler.SyncPointNotEncountered(module, branch, syncPoint); err != nil {
 				return err
--- a/private/buf/bufsync/syncer.go
+++ b/private/buf/bufsync/syncer.go
@@ -32,6 +32,7 @@ type syncer struct {
 	storageGitProvider storagegit.Provider
 	errorHandler       ErrorHandler
 	modulesToSync      []Module
+	syncPointResolver  SyncPointResolver
 
 	knownTagsByCommitHash map[string][]string
 }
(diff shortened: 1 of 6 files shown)


private/buf/bufsync/syncer.go
```diff
@@ -56,33 +57,112 @@ func newSyncer(
...
+// resolveSyncPoints resolves sync points for all known modules for the specified branch,
+// returning all modules for which sync points were found, along with their sync points.
+//
+// If a SyncPointResolver is not configured, this returns an empty map immediately.
+func (s *syncer) resolveSyncPoints(ctx context.Context, branch string) (map[Module]git.Hash, error) {
+	syncPoints := map[Module]git.Hash{}
+	// If resumption is not enabled, we can bail early.
+	if s.syncPointResolver == nil {
+		return syncPoints, nil
+	}
```
unmultimedio, line 69: Nit: Are the callers promised they they'll receive a not-nil map? Otherwise this looks a bit more clear:
```suggestion
	// If resumption is not enabled, we can bail early.
	if s.syncPointResolver == nil {
		return nil, nil
	}
	syncPoints := make(map[Module]git.Hash)
```

private/buf/bufsync/syncer.go
```diff
@@ -56,33 +57,112 @@ func newSyncer(
...
+		return nil, fmt.Errorf("resolve syncPoint for module %s: %w", module.RemoteIdentity(), err)
+	}
+	if syncPoint != nil {
+		// Validate that the commit pointed to by the sync point exists.
+		if _, err := s.repo.Objects().Commit(syncPoint); err != nil {
+			return nil, s.errorHandler.InvalidSyncPoint(module, branch, syncPoint, err)
+		}
+		return syncPoint, nil
+	}
+	return nil, nil
```
unmultimedio, line 108: Another nit to reduce nesting
```suggestion
	if syncPoint == nil {
		return nil, nil
	}
	// Validate that the commit pointed to by the sync point exists.
	if _, err := s.repo.Objects().Commit(syncPoint); err != nil {
		return nil, s.errorHandler.InvalidSyncPoint(module, branch, syncPoint, err)
	}
	return syncPoint, nil
```

doriable (approved):

doriable (approved):

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -226,32 +256,74 @@ func newErrorHandler(logger *zap.Logger) bufsync.ErrorHandler {
...
+	// commit has been garbage collected. In this case, let's present a better error message.
+	//
+	// We may want to provide a flag for sync to continue despite this, accumulating the error,
+	// and error at the end, so that other branches can continue to sync, but this branch is
+	// out of date. This is not trivial if the branch that's been rebased is a long-lived
+	// branch (like main) whose artifacts are consumed by other branches, as we may fail to
+	// sync those commits if we continue. So we now we simply error.
+	if errors.Is(err, git.ErrObjectNotFound) {
+		return fmt.Errorf(
+			"last known commit %s was not found for module %s; did you rebase?",
```
unmultimedio, line 301:
```suggestion
			"last synced commit %s was not found for module %s; did you rebase?",
```

unmultimedio (approved):
Looks very nice, missing some tests. Left a few nits.

private/pkg/git/git.go
```diff
@@ -46,7 +46,10 @@ const (
 	ModeSubmodule ObjectMode = 016_0000
 )
 
-var ErrTreeNodeNotFound = errors.New("node not found")
+var (
+	ErrTreeNodeNotFound = errors.New("node not found")
```
bufdev, line 50: needs godoc

bufdev (approved):
Generally fine

private/buf/bufsync/syncer.go
saquibmian: Yes, they are promised a non-nil map, otherwise they'll have to nil-check it. I commented with that in the godoc.

aside: I dislike that the zero value for maps is useless while the zero value for slices is useful. Seems like an implementation detail that surfaces in the language semantics.

doriable (approved):
Restamping :)

PR #2171, opened 2023-06-08
Add connect interceptor to log server warnings

mfridman:
This PR adds a connect interceptor to log arbitrary server warnings.

The CLI will recognize the `buf-warning-bin` header, which is expected to be a base64-encoded message. It uses the applog container so users can ignore these messages with `--no-warn`. Do we want this behaviour, or should we always print the warning regardless of user preferences? Probably not.

There's one edge case in that a single command may call multiple RPC's, and if multiple RPC's return warnings they will all be printed. We could add some ctx-based print once logic, but I suspect if multiple RPC's return warnings, those errors will be different anyway, so it's unlikely we'll want to print only the first one.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Add `buf-cli-warning-bin` header to detect and print server warning messages.
+- Add `buf-warning-bin` header to detect and print server warning messages.
 
 ## [v1.21.0] - 2023-06-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf-cli-warning-bin` header to detect and print server warning messages.
 
 ## [v1.21.0] - 2023-06-05
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Add `buf-warning-bin` header to detect and print server warning messages.
+- No changes yet.
 
 ## [v1.21.0] - 2023-06-05
 
--- a/private/bufpkg/bufconnect/bufconnect.go
+++ b/private/bufpkg/bufconnect/bufconnect.go
@@ -26,7 +26,7 @@ const (
 	CliVersionHeaderName = "buf-version"
 	// CLIWarningHeaderName is the name of the header carrying a base64-encoded warning message
 	// from the server to the CLI.
-	CLIWarningHeaderName = "buf-cli-warning-bin"
+	CLIWarningHeaderName = "buf-warning-bin"
 	// DefaultRemote is the default remote if none can be inferred from a module name.
 	DefaultRemote = "buf.build"
 )
--- a/private/bufpkg/bufconnect/bufconnect.go
+++ b/private/bufpkg/bufconnect/bufconnect.go
@@ -24,6 +24,9 @@ const (
 	AuthenticationTokenPrefix = "Bearer "
 	// CliVersionHeaderName is the name of the header carrying the buf CLI version.
 	CliVersionHeaderName = "buf-version"
+	// CLIWarningHeaderName is the name of the header carrying a base64-encoded warning message
+	// from the server to the CLI.
+	CLIWarningHeaderName = "buf-cli-warning-bin"
 	// DefaultRemote is the default remote if none can be inferred from a module name.
 	DefaultRemote = "buf.build"
 )
--- a/private/bufpkg/bufconnect/interceptors.go
+++ b/private/bufpkg/bufconnect/interceptors.go
@@ -16,7 +16,9 @@ package bufconnect
 
 import (
 	"context"
+	"encoding/base64"
 
+	"github.com/bufbuild/buf/private/pkg/app/appflag"
 	"github.com/bufbuild/connect-go"
 )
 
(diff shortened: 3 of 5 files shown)


private/bufpkg/bufconnect/bufconnect.go
```diff
@@ -24,6 +24,9 @@ const (
 	AuthenticationTokenPrefix = "Bearer "
 	// CliVersionHeaderName is the name of the header carrying the buf CLI version.
 	CliVersionHeaderName = "buf-version"
+	// CLIWarningHeaderName is the name of the header carrying a base64-encoded warning message
+	// from the server to the CLI.
+	CLIWarningHeaderName = "buf-cli-warning-bin"
```
joshcarp, line 29: Should we make this something like `bsr-warning-bin` or something, I imagine that we'd also want to surface this error if/when we make other integrations with non cli clients like studio?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf-cli-warning-bin` header to detect and print server warning messages.
```
pkwarren, line 5: Maybe this is overthinking it but do we want this to be an error detail to be able to provide more details in the future? Perhaps we start with warning message(s) but maybe at some point it adds other metadata (reference id for support, etc.)?

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -28,17 +30,33 @@ const (
...
-		) (connect.AnyResponse, error) {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
 			req.Header().Set(CliVersionHeaderName, version)
 			return next(ctx, req)
 		}
 	}
 	return interceptor
 }
 
+// NewCLIWarningInterceptor returns a new Connect Interceptor that logs CLI warnings returned by sever responses.
```
pkwarren, line 41:
```suggestion
// NewCLIWarningInterceptor returns a new Connect Interceptor that logs CLI warnings returned by server responses.
```

CHANGELOG.md
mfridman: Totally fair question, and here are some thoughts:

- The end result is a message string printed to the user, so whether it's a blob in a header or an error detail shouldn't matter
- Headers are easier to work with and are universally supported (maybe length? considerations). 
- If (extremely unlikely) there is a change in error details, then future servers would need to remain backwards-compatible with older clients
- It keeps it consistent with what we're doing with raw errors in `NewErrorInterceptor`, dumping the raw errors whichever way the server decides to format it (moot point because the same could be true for a base64-encoded string field)

private/bufpkg/bufconnect/bufconnect.go
bufdev: `buf-warning-bin`

mfridman: Do you mean renaming the header to something more generic, like `buf-warning-bin`? So it can be used universally across all our projects?

It's common to prefix headers with company name, so `buf-warning-bin` probably seems more appropriate.

bufdev: Yea make it more generic

bufdev: ie `buf-warning-bin`

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -28,17 +30,33 @@ const (
...
+func NewCLIWarningInterceptor(container applog.Container) connect.UnaryInterceptorFunc {
+	interceptor := func(next connect.UnaryFunc) connect.UnaryFunc {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+			resp, err := next(ctx, req)
+			if resp != nil && resp.Header() != nil {
+				encoded := resp.Header().Get(CLIWarningHeaderName)
+				if encoded != "" {
+					if warning, err := base64.StdEncoding.DecodeString(encoded); err == nil && len(warning) > 0 {
```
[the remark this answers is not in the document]
akshayjshah: To be fully safe, headers must be ASCII. That makes it annoying to include problematic user-supplied data in the error message (e.g., "username 💩 is not a valid Kafka topic").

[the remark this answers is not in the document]
mfridman: Covers any future case of allowing non-ASCII strings. If we don't do it, we'll probably regret it in the future for older CLI versions, and there is no downside outside of not seeing the raw text in the header. Since this is intended for the CLI, it seems like a reasonable tradeoff.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf-warning-bin` header to detect and print server warning messages.
```
bufdev, line 5: I don't know if this is worthy of a changelog update actually

private/bufpkg/bufconnect/interceptors.go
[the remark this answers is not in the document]
bufdev: OK got it

bufdev (approved):
Looks fine, I might actually remove the CHANGELOG entry

CHANGELOG.md
mfridman: Removed

bufdev (approved):

private/bufpkg/bufconnect/interceptors.go
```diff
@@ -28,17 +30,33 @@ const (
...
 	return interceptor
 }
 
+// NewCLIWarningInterceptor returns a new Connect Interceptor that logs CLI warnings returned by server responses.
+func NewCLIWarningInterceptor(container applog.Container) connect.UnaryInterceptorFunc {
+	interceptor := func(next connect.UnaryFunc) connect.UnaryFunc {
+		return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+			resp, err := next(ctx, req)
+			if resp != nil && resp.Header() != nil {
+				encoded := resp.Header().Get(CLIWarningHeaderName)
```
lrewega, line 47: Would `.Values()` be more appropriate? If our goal is to be future-compatible, perhaps we should expect multiple headers?

private/bufpkg/bufconnect/interceptors.go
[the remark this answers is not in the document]
lrewega: Should we use [connect.DecodeBinaryHeader](https://pkg.go.dev/github.com/bufbuild/connect-go#DecodeBinaryHeader) instead?

Relatedly:

> Following usual HTTP semantics, multiple base64-encoded values may be joined with a comma. When receiving such comma-separated values, split them with [strings.Split](https://pkg.go.dev/strings#Split) before calling DecodeBinaryHeader.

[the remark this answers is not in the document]
mfridman: Fair suggestion, updated in https://github.com/bufbuild/buf/pull/2171/commits/ea0d65ad6ad7cfee95719cb225cdf0407da65966

PR #2172, opened 2023-06-08
Handle Inputs for v2

oliversun9:
This PR adds logic for mapping an `ExternalInputConfig` to an `InputConfig`, and then to a `Ref`.

This PR does not modify `run` from `generate.go`, it will happen in the PR for plugins.


private/buf/buffetch/buffetch.go
```diff
@@ -360,3 +360,178 @@ func NewWriter(
 type getSourceBucketOptions struct {
 	workspacesDisabled bool
 }
+
+type RefBuilder interface {
+	// GetGitRef returns a Ref for a Git repository.
+	GetGitRef(
+		ctx context.Context,
+		format string,
```
bufdev, line 368: Why do these need formats?

private/buf/buffetch/internal/internal.go
```diff
@@ -157,8 +177,8 @@ type DirRef interface {
 }
 
 // NewDirRef returns a new DirRef.
-func NewDirRef(path string) (DirRef, error) {
-	return newDirRef("", path)
+func NewDirRef(format string, path string) (DirRef, error) {
```
bufdev, line 180: `format` is `dir`, why do we need to pass this? This seems incorrect unless I'm missing something

private/buf/buffetch/internal/internal.go
```diff
@@ -211,8 +262,8 @@ type ModuleRef interface {
 // NewModuleRef returns a new ModuleRef.
 //
 // The path must be in the form server/owner/repository/branch[:digest].
-func NewModuleRef(path string) (ModuleRef, error) {
-	return newModuleRef("", path)
+func NewModuleRef(format string, path string) (ModuleRef, error) {
```
bufdev, line 265: Same comment whole way down

private/buf/buffetch/buffetch.go
oliversun9: The `format` is passed to the internal constructor for the specific Refs, such as `internal.NewGitRef` and `internal.NewModuleRef`. 

They are used for error messages, for example: [`NewInvalidPathError(format, path)`](https://github.com/bufbuild/buf/blob/main/private/buf/buffetch/internal/module_ref.go#L41).
So that in V2, I pass in `module` and the error message will look like `invalid path xyz for module`.
And for parsing command line input argument, `mod` is passed and the error mesage will look like `invalid path xyz for mod`.

private/buf/buffetch/internal/internal.go
oliversun9: See [answer to the question above](https://github.com/bufbuild/buf/pull/2172#discussion_r1224720727)

private/buf/buffetch/internal/internal.go
```diff
@@ -84,6 +90,19 @@ type ArchiveType int
 // CompressionType is a compression type.
 type CompressionType int
 
+// NewCompressionType returns a CompressionType from the compression name,
+// and returns CompressionTypeNone if name is empty.
+func NewCompressionType(name string) (CompressionType, error) {
```
bufdev, line 95: `ParseCompressionType`

bufdev (approved):

PR #2176, opened 2023-06-09
Remove duplicate imports from formatter

catusax:
remove duplicate imports feat for `buf format`, will only preserve last import.


private/buf/bufformat/formatter.go
```diff
@@ -259,6 +259,11 @@ func (f *formatter) writeFileHeader() {
 		if i == 0 && f.previousNode != nil && !f.leadingCommentsContainBlankLine(importNode) {
 			f.P("")
 		}
+
+		if i > 0 && importNode.Name.AsString() == importNodes[i-1].Name.AsString() {
```
bufdev, line 263: Can you add a comment as to what this is doing, ie that this removes duplicates because the importNodes are already sorted?

bufdev (commented):
Thanks for this! Just one comment. It'd be great if you could add a test as well, but if not, we'll do it.

private/buf/bufformat/formatter.go
catusax: @bufdev just added testdata and comment

catusax:
Another question, i have read the gofmt commit : https://github.com/golang/go/commit/08925ce6ee16be5a3b937c0d55c2548bf30c5776

and their strategy is as follows: 
- Sort imports by import path, then import name, then comment.
- If two imports have the same import path and import name, and one of them has no comment, remove the import with no comment.

This commit will remove user's comment on import if its duplicate, i don't think it's the best solution. should i start working on a new strategy like above? @bufdev

bufdev:
That would be an improvement yep, go ahead!

catusax:
> That would be an improvement yep, go ahead!

@bufdev I have made a new commit with this strategy:
- sort imports by name
- then sort by import level ( import public > import > import weak )
- then sort by hasComment
- remove uncommented and lower level duplicate imports

please check it out

private/buf/bufformat/formatter.go
```diff
@@ -2197,6 +2225,36 @@ func (f *formatter) leadingCommentsContainBlankLine(n ast.Node) bool {
 	return newlineCount(info.LeadingWhitespace()) > 1
 }
 
+func (f *formatter) importHasComment(importNode *ast.ImportNode) bool {
+	return f.nodeHasComment(importNode.Keyword) ||
```
bufdev, line 2229: I think this should include the `importNode` directly as well, for completeness

private/buf/bufformat/formatter.go
```diff
@@ -2197,6 +2225,36 @@ func (f *formatter) leadingCommentsContainBlankLine(n ast.Node) bool {
...
+// importLevel maps import types to a level, so it can be compared and sorted.
+// `import public`=3, `import`=2; `import weak` =1
+func importLevel(node *ast.ImportNode) int {
+	switch {
+	case node.Public != nil:
+		return 3
+	case node.Weak != nil:
+		return 1
+	default:
+		return 2
```
bufdev, line 2254: The default import level should take precedence, so the precedence is `default, public, weak`.

private/buf/bufformat/formatter.go
```diff
@@ -2197,6 +2225,36 @@ func (f *formatter) leadingCommentsContainBlankLine(n ast.Node) bool {
...
+		return false
+	}
+	nodeinfo := f.fileNode.NodeInfo(node)
+	return nodeinfo.LeadingComments().Len() > 0 ||
+		nodeinfo.TrailingComments().Len() > 0
+}
+
+// importLevel maps import types to a level, so it can be compared and sorted.
+// `import public`=3, `import`=2; `import weak` =1
+func importLevel(node *ast.ImportNode) int {
```
bufdev, line 2247: `importSortOrder`

bufdev (commented):
A few comments, but this looks great overall! Once these comments are addressed, we'll get this merged! Thank you so much for your help.

private/buf/bufformat/formatter.go
bufdev: Add `f.nodeHasComment(importNode)` to the `||` list

catusax:
there is something wrong with the testdata and `nodeHasComment` function, the test reports
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x79bf29]

goroutine 233 [running]:
github.com/bufbuild/protocompile/ast.(*KeywordNode).Start(0x7?)
```
at this line https://github.com/catusax/buf/blob/70320a9cfbe8df718012a149fb4a6384e1983d3a/private/buf/bufformat/formatter.go#L2247

but i have already checked that `node` is not nil, i dont known how to solve it yet,
will try to solve this bug tomorrow

catusax:
@ bufdev i have fixed the testcase, please take a review again

bufdev (approved):
This looks great! Thank you so much for the contribution. We will get this out in the next release!

PR #2178, opened 2023-06-12
Dependencies warning fix: remote discovery for mod update and mod prune

unmultimedio:
This PR is composed of two fixes, one is consequence of the other:

1. Recently we introduced a warning if we detected an import that didn't come from local target files, or direct dependencies; this includes workspace modules, and more importantly, transitive dependencies.

This warning shouldn't be present for local unnamed modules, since they might be local-only, and not required to be pushed to any BSR. So that is the first fix, removing this warning from local unnamed modules.

2. In the case that one of those local modules depended on remote BSR modules, and those modules were enterprise (i.e. its remote was different than "buf.build"), then the commands `buf mod update` and `buf mod prune` would default to "buf.build" to fetch/clean the dependencies, which forced users to set a vanity name in their local modules that matched their remotes, so the command would recognize correctly the remote and fetch dependencies from the right place.

I'm introducing a `DiscoveryRemote` function that iterates over all declared direct dependencies in the `buf.yaml` file, and returns the first remote that's different than "buf.build" if any, so in case there is a local module that has remote dependencies, we don't force the users to name such modules, and it can remain being unnamed, which results in we successfully removing the annoying warning from point 1.

Note: Any named module won't go through that discovery process, and the remote in the name will be used.


private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -111,32 +111,13 @@ func run(
 		return err
 	}
 
-	remote := bufconnect.DefaultRemote
-	if moduleConfig.ModuleIdentity != nil && moduleConfig.ModuleIdentity.Remote() != "" {
-		remote = moduleConfig.ModuleIdentity.Remote()
-	} else {
-		for _, moduleReference := range moduleConfig.Build.DependencyModuleReferences {
-			if moduleReference.Remote() != bufconnect.DefaultRemote {
-				warnMsg := fmt.Sprintf(
```
unmultimedio, line 120: If it's an unnamed module, we moved this from being a warning to being a debug message. Most likely the user already knows they don't have a module name, and it's supposed to be local. The message is still there, just behind a `--debug` flag.

private/bufpkg/bufmodule/bufmoduleref/module_reference.go
```diff
@@ -94,6 +95,21 @@ func (m *moduleReference) IdentityString() string {
...
+func DiscoverRemote(refs []ModuleReference) string {
+	for _, ref := range refs {
+		if ref == nil {
+			continue
+		}
+		if ref.Remote() != bufconnect.DefaultRemote {
+			return ref.Remote()
+		}
+	}
+	return bufconnect.DefaultRemote
```
pkwarren, line 110: Do we want to infer here if there are multiple non-default remotes found?

private/bufpkg/bufmodule/bufmoduleref/module_reference.go
```diff
@@ -94,6 +95,21 @@ func (m *moduleReference) IdentityString() string {
 	return m.remote + "/" + m.owner + "/" + m.repository
 }
 
+// DiscoverRemote returns the first remote in a list of dependencies that is not "buf.build", if
+// any. If not, it defaults to "buf.build". It skips any nil reference. It's useful for targeting
+// right single-tenant BSR addresses.
```
pkwarren, line 100:
```suggestion
// single-tenant BSR addresses.
```

private/bufpkg/bufimage/bufimagebuild/builder.go
```diff
@@ -249,8 +249,13 @@ func (b *builder) warnInvalidImports(
 			)
 			if workspaceModule, ok := workspaceFilesToModule[importFilePath]; ok {
 				if workspaceModule == "" {
-					warnMsg += ", but is found in a local workspace unnamed module. Name the module and declare it in the deps key in buf.yaml."
+					// If dependency comes from an unnamed module, that is probably a local dependency, and
+					// that module won't be pushed to the BSR. We can skip this warning.
+					continue
 				} else {
```
pkwarren, line 255: For readability, it would be nice to remove the else/nesting here since we continue above.

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Remove an import warning in case such import statement points to a local unnamed module.
```
pkwarren, line 5:
```suggestion
- Remove an import warning in case the import points to a local unnamed module.
```

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
...
+- Discover the correct remote to use for fetching dependencies in the `mod update` and `mod prune`
+  commands, in case it's an unnamed module and there is a mix of remotes in the `buf.yaml`
```
pkwarren, line 7:
```suggestion
  commands, in case it's an unnamed module and there are a mix of remotes in the `buf.yaml`
```

pkwarren (commented):
Good catch - changes look good. Just a few comments.

private/bufpkg/bufmodule/bufmoduleref/module_reference.go
unmultimedio: Since the BSR doesn't support federation anymore, and we're finishing migrating a few customers that still have it enabled, I don't see a scenario in which we'd ever have more than one non-default remote. 99% of the cases they'll all have the same remote, either the public BSR, or their private remote address. In the remaining cases, there will be just two, public and theirs, we want to choose the latter.

pkwarren (approved):

private/buf/cmd/buf/command/mod/modprune/modprune.go
```diff
@@ -102,6 +93,24 @@ func run(
...
+		if config.ModuleIdentity != nil && config.ModuleIdentity.Remote() != "" {
+			remote = config.ModuleIdentity.Remote()
+		} else {
+			// At this point we know there's at least one dependency. If it's an unnamed module, discover
+			// the right remote from the list of dependencies.
+			remote = bufmoduleref.DiscoverRemote(config.Build.DependencyModuleReferences)
+			container.Logger().Debug(fmt.Sprintf(
+				`%q does not specify a "name". From your dependencies, Buf is defaulting to using remote %q for dependency resolution. This remote may be unable to resolve some of your dependencies. Did you mean to specify a "name: %s/..." on this module?`,
+				existingConfigFilePath,
+				remote, remote,
```
mfridman, line 106: Was this meant to be `bufconnect.DefaultRemote, remote`?

unmultimedio: Not really, we're telling the customer which remote we discovered (not always the default one). Probably the confusion comes from the expression "defaulting to"?

mfridman: Hmm ye. This reads a bit odd because it'll always be the same.. "buf is defaulting to X, and X may not be available. Did you mean to specify X?"

private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -189,6 +170,19 @@ func getDependencies(
...
 	}
+	var remote string
+	if moduleConfig.ModuleIdentity != nil && moduleConfig.ModuleIdentity.Remote() != "" {
+		remote = moduleConfig.ModuleIdentity.Remote()
+	} else {
+		// At this point we know there's at least one dependency. If it's an unnamed module, discover
+		// the right remote from the list of dependencies.
+		remote = bufmoduleref.DiscoverRemote(moduleConfig.Build.DependencyModuleReferences)
+		container.Logger().Debug(fmt.Sprintf(
+			`%q does not specify a "name". From your dependencies, Buf is defaulting to using remote %q for dependency resolution. This remote may be unable to resolve some of your dependencies. Did you mean to specify a "name: %s/..." on this module?`,
```
bufdev, line 181:
```
File path/to/file.yaml does not specify the "name" field. Based on the dependency buf.foo.com/bar/baz, it appears that you are using a BSR instance at buf.foo.com. Did you mean to specify "name: buf.foo.com/..." within path/to/file.yaml?
```

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Remove an import warning in case the import statement points to a local unnamed module.
```
bufdev, line 5: Remove

CHANGELOG.md
```diff
@@ -2,7 +2,10 @@
...
+- Discover the correct remote to use for fetching dependencies in the `mod update` and `mod prune`
```
bufdev, line 6: Remove

private/bufpkg/bufmodule/bufmoduleref/module_reference.go
```diff
@@ -15,6 +15,7 @@
 package bufmoduleref
 
 import (
+	"github.com/bufbuild/buf/private/bufpkg/bufconnect"
```
bufdev, line 18: bufmoduleref should not depend on bufconnect, bufmoduleref is a foundational package

private/bufpkg/bufmodule/bufmoduleref/module_reference.go
```diff
@@ -94,6 +95,21 @@ func (m *moduleReference) IdentityString() string {
 	return m.remote + "/" + m.owner + "/" + m.repository
 }
 
+// DiscoverRemote returns the first remote in a list of dependencies that is not "buf.build", if
+// any. If not, it defaults to "buf.build". It skips any nil reference. It's useful for targeting
+// single-tenant BSR addresses.
+func DiscoverRemote(refs []ModuleReference) string {
```
bufdev, line 101: This likely should be at a higher level than bufmoduleref

PR #2183, opened 2023-06-13
Add server breaking changes audit log event

mfridman:
This PR adds an audit log event for server breaking changes configuration changes.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -161,6 +164,7 @@ message Event {
     PayloadRepositoryLabelCreated repository_label_created = 35;
     PayloadRepositoryLabelMoved repository_label_moved = 36;
     PayloadCuratedPluginDeleted curated_plugin_deleted = 37;
+    PayloadServerBreakingChangesChanged server_breaking_changes_changed = 38;
```
mfridman, line 167: Don't like the `_changed` stutter, but it keeps it consistent with the rest of the definitions.

The alternative is to refer to this as `server_breaking_changed`, but don't think it's worth changing.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -415,3 +419,12 @@ message PayloadRepositoryLabelMoved {
   // from_commit_id is the id of the commit on which the label was moved from.
   string from_commit_id = 7;
 }
+
+message PayloadServerBreakingChangesChanged {
+  // enabled tracks whether breaking changes were enabled or disabled.
```
mfridman, line 424: @unmultimedio, is the audit event meant to capture the old state, or is it sufficient to record the new changes?

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -100,6 +102,7 @@ enum EventType {
   EVENT_TYPE_REPOSITORY_COMMIT_DELETED = 28;
   EVENT_TYPE_REPOSITORY_LABEL_CREATED = 29;
   EVENT_TYPE_REPOSITORY_LABEL_MOVED = 30;
+  EVENT_TYPE_SERVER_BREAKING_CHANGES_CHANGED = 32;
```
unmultimedio, line 105: So the breaking rules are server-wide? I thought they were org-wide.

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: What about "Server Breaking Protection Policy", so it'd result in `PayloadServerBreakingProtectionPolicyChanged`

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: We've been tracking both old and new properties, e.g.: https://github.com/bufbuild/buf/blob/8df996524679da8fca7c006c957fb4ce7cf0f3b2/proto/buf/alpha/audit/v1alpha1/event.proto#L182-L195

Although this being a boolean, I don't know, feels weird.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -54,6 +55,7 @@ enum ResourceType {
   RESOURCE_TYPE_TOKEN = 10;
   RESOURCE_TYPE_SCIM_TOKEN = 11;
   RESOURCE_TYPE_REPOSITORY_LABEL = 12;
+  RESOURCE_TYPE_SERVER = 13;
```
mfridman, line 58: @gilwong00 This PR captures the proto changes for adding an audit log event. 

The goal is to record a "server" change, specifically "breaking changes" updates. The outer payload contains generic things like who did it, timestamp, metadata (tracing, request, etc.) and the payload itself.

unmultimedio (commented):
Not 100% contextualized on this, I'm assuming these breaking categories can be switched on/off independently? what about separate events like:

```
EVENT_TYPE_SERVER_BREAKING_POLICY_ENABLED
EVENT_TYPE_SERVER_BREAKING_POLICY_DISABLED
```

And in the payload we have which category was turned on or off. Feels more in line with events like [these ones](https://github.com/bufbuild/buf/blob/8df996524679da8fca7c006c957fb4ce7cf0f3b2/proto/buf/alpha/audit/v1alpha1/event.proto#L74-L76), in which we have an on/off, and an additional "changed" if the change was not boolean.

mfridman:
@unmultimedio thanks for the suggestions 👍

It's not split into 2 events: enabling and disabling.

proto/buf/alpha/audit/v1alpha1/event.proto
mfridman:
```
PayloadServerBreakingChangePolicyEnabled
PayloadServerBreakingChangePolicyDisabled
```

unmultimedio (approved):

PR #2185, opened 2023-06-14
Update PluginGenerationRequest to include per-plugin include_imports …

oliversun9:
…and include_well_known_types.

This is needed for `buf.gen.yaml` `v2` which allows specifying `include_imports` and `include_well_known_types` per plugin.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -321,6 +323,12 @@ message PluginGenerationRequest {
   // The options to pass to the plugin. These will
   // be merged into a single, comma-separated string.
   repeated string options = 2;
+  // Include imports from the Image in generation.
+  bool include_imports = 3;
+  // Include Well-Known Types from the Image in generation.
+  //
+  // include_imports must be set if include_well_known_types is set.
+  bool include_well_known_types = 4;
```
pkwarren, line 331: Instead of ignoring these if set to true at the top-level, do we just want to make these `optional bool` and then if set (to either true or false), they'll override the setting at the top level?

oliversun9: Yes I think we should make these `optional bool`.
I made top level override each individual request because in buf gen yaml we want the behavior of command line flag `--include-imports` [override](https://github.com/bufbuild/extra/blob/main/buf-gen-yaml-v2/buf.gen.yaml#L123) each individual plugin's `include_imports` in `buf.gen.yaml`.
I also realized that this command line flag behavior is a separate thing to API design, so I agree with you on that as well.

pkwarren (approved):

PR #2188, opened 2023-06-14
Remove `noTimeoutBuilder`

### LINKED ISSUE #2187: Commands with `noTimeoutBuilder` don't bind global flags
The `noTimeoutBuilder` [here](https://github.com/bufbuild/buf/blob/23248b0e27a07ca34b28584e20624af2098ae04c/private/buf/cmd/buf/buf.go#L90-L93) used for `buf beta studio-agent` and `buf alpha repo sync` don't go through the `BindPersistentFlags` codepath [here](https://github.com/bufbuild/buf/blob/5efbb658d56a6f02f9fd0ffdc568d3ad79b74d75/private/buf/cmd/buf/buf.go#L100), which means that they don't bind flags like `--debug` or `--profile`, among others.

Quick fix would be removing that builder, just use the regular `builder`, and request the users to use `--timeout 0` to set no timeout to their operations, if required.

Related to https://github.com/bufbuild/buf/pull/1305

unmultimedio:
Basically reverts https://github.com/bufbuild/buf/pull/1305. Fixes #2187.

### DIFF
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -87,10 +87,6 @@ func NewRootCommand(name string) *appcmd.Command {
 		appflag.BuilderWithTimeout(120*time.Second),
 		appflag.BuilderWithTracing(),
 	)
-	noTimeoutBuilder := appflag.NewBuilder(
-		name,
-		appflag.BuilderWithTracing(),
-	)
 	globalFlags := bufcli.NewGlobalFlags()
 	return &appcmd.Command{
 		Use:                 name,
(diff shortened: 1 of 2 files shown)


bufdev (approved):

doriable (approved):

saquibmian:
Two things:

1. Is this not a breaking change? This breaks anyone who previously called `buf beta studio-agent`, they must now pass `--timeout 0`.
2. Requiring certain commands to be run with `--timeout 0` is poor UX.

unmultimedio:
> Is this not a breaking change? This breaks anyone who previously called buf beta studio-agent, they must now pass --timeout 0.

Being `beta` I don't think it qualifies as a breaking change. When doing the change or default no-timeout you could say the same. And all things considered, it was a worse active bug not being able to do `buf beta studio-agent --any-global-flag`.

> Requiring certain commands to be run with --timeout 0 is poor UX.

I'm not sure if I agree with you, it's certainly not nice if I have to run it everytime, but I think making it explicit is a nice property.

PR #2189, opened 2023-06-14
Remove reverse field from ListRepositoryBranchRequest

seankimdev:
Remove the reverse field as we don't plan on supporting it yet.
The rpc hasn't been implemented yet, so it is safe to be removed.


proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -16,11 +16,15 @@ syntax = "proto3";
...
 
+import "google/protobuf/timestamp.proto";
+
 message RepositoryBranch {
   // The name of the repository branch.
   string name = 1;
   // The name of the latest commit on the branch.
   string latest_commit_name = 2;
+  bool is_main_branch = 3;
+  google.protobuf.Timestamp last_update_time = 4;
```
doriable, line 27: We should add the documentation for this.

doriable (approved):
Looks reasonable, just left a comment regarding documentation, pre-emptively stamping :)

PR #2192, opened 2023-06-14
Add `BSR_HEAD` label namespace

srikrsna-buf:
Add a new namespace to track main commits on the BSR.


stefanvanburen (approved):

doriable (approved):
I'm okay with the name `BSR_HEAD`, going to tag @unmultimedio

unmultimedio (approved):
I like `BSR_STATUS` better 😅 because `HEAD|MAIN` gives me a sense of a pointer, just like a git branch, while a "status" it's exactly what we're after: approved|pending|declined.

wdyt?

doriable:
> I like `BSR_STATUS` better 😅 because `HEAD|MAIN` gives me a sense of a pointer, just like a git branch, while a "status" it's exactly what we're after: approved|pending|declined. Feels clearer.
> 
> wdyt? not blocking though.

I believe the status is tracked in the `Approval` namespace https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/labels.proto#L26

And this is only to track what is on BSR's `main` (so it is kind of like a pointer, lol).

PR #2196, opened 2023-06-15
Remove org-level breaking change option

mfridman:


pgmitche (approved):
This breaking change is fine, because no one should be using it at all

PR #2199, opened 2023-06-16
Add `Print` function to `diffmyers` package

srikrsna-buf:
Add `Print` function to `diffmyers` package to print the edits in the [unified](https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html) format.


private/pkg/diff/diffmyers/diffmyers.go
```diff
@@ -51,6 +53,107 @@ func Diff(from, to [][]byte) []Edit {
...
+		line     []byte
+		hunk     bool
+	}
+	// If the last line of from is not a newline append one.
+	if len(from) > 0 && from[len(from)-1] != nil {
+		last := from[len(from)-1]
+		if last[len(last)-1] != '\n' {
+			from[len(from)-1] = append(last, '\n')
+		}
+	}
```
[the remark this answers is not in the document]
srikrsna-buf: This is to add a newline to the end of the `from` file, in this case the `from` ends and there are a bunch of insertions. If we remove this line we'll end up having the insertion on the same line

joshcarp (approved):

private/pkg/diff/diffmyers/diffmyers.go
```diff
@@ -51,6 +53,107 @@ func Diff(from, to [][]byte) []Edit {
 	return shortestEdits(from, to, 0, 0)
 }
 
+// Print prints the edits in the unified diff format without the header.
+//
+// Ref: https://www.gnu.org/software/diffutils/manual/html_node/Detailed-Unified.html
+func Print(from, to [][]byte, edits []Edit) ([]byte, error) {
+	const contextThreshold = 2
```
stefanvanburen, line 60: not a huge deal, but it seems like the context for hunks in `diff` (at least, macOS/BSD `diff`) is 3 (and seems the same on GitHub, at least looking at the context given between hunks in this PR). Is there a reason we chose 2?

srikrsna-buf: It does 3 based on the logic that is when the threshold crossed 2 which is 3 we stop. But you are right this is a bit confusing we can update this to match 3 here.

PR #2200, opened 2023-06-16
Implement plugins v2

oliversun9:
Implemented plugins v2 and restructured `bufgen` package / directory.

The directory structure now:
* `bufgen` exposes `Generator` which figures out the version of the file and calls the corresponding generator of that version.
* `bufgen/internal`: common utilities shared by the entire package, such as reading config file's data, exposing `Strategy` type and etc.
* `bufgen/internal/plugingen`: 
  * exposes a `Generator` that executes the plugins and writes the responses, the same one as `bufgen` in `main`. It now takes in a slice of `PluginConfig` interface, as opposed to structs before.
  * exposes `PluginConfig` interface
* `bufgen/internal/bufgenv1`: same as before, but now `PluginConfig` interfaces are created
* `bufgen/internal/bufgenv2`: adds logic for creating `PluginConfig` from a config file


private/buf/bufgen/bufgen.go
```diff
@@ -40,6 +48,77 @@ type ExternalConfigVersion struct {
 	Version string `json:"version,omitempty" yaml:"version,omitempty"`
 }
 
+// Generator generates Protobuf stubs based on configurations.
```
oliversun9, line 51: The `Generator` stuff is moved from `bufgenv1.go` (previously known as `bufgen.go`)

private/buf/bufgen/bufgenv1/bufgenv1.go
```diff
@@ -130,76 +50,13 @@ func ReadConfigV1(
...
-	PluginConfigs []*PluginConfig
+	PluginConfigs []bufgen.PluginConfig
 	// Optional
 	ManagedConfig *ManagedConfig
 	// Optional
 	TypesConfig *TypesConfig
 }
 
-// PluginConfig is a plugin configuration.
-type PluginConfig struct {
```
oliversun9, line 141: `PluginConfig` is now an interface in `bufgen` package

private/buf/bufgen/bufgenv1/config.go
```diff
@@ -75,9 +75,6 @@ func getConfig(
 		if err := unmarshalStrict(data, &externalConfigV1); err != nil {
 			return nil, err
 		}
-		if err := validateExternalConfigV1(externalConfigV1, id); err != nil {
```
oliversun9, line 78: `validateExternalConfigV1` only validated plugin configs, now using `getPluginConfigs` to both validate and create plugin configs.

.golangci.yml
```diff
@@ -17,8 +17,8 @@ linters-settings:
       # Ban debug statements
       - '^fmt\.Print'
       - '^log\.'
-      - '^print$'
-      - '^println$'
+      - "^print$"
```
bufdev, line 20: ??

private/buf/bufgen/bufgenv1/config.go
```diff
@@ -143,49 +114,141 @@ func validateExternalConfigV1(externalConfig ExternalConfigV1, id string) error
...
+				)
 			} else {
-				// plugin.Plugin is a local plugin - verify it isn't using an alpha remote plugin path
+				// plugin.Plugin is a local plugin - verify it isn't using an legacy remote plugin path
 				if _, _, _, _, err := bufremoteplugin.ParsePluginVersionPath(pluginIdentifier); err == nil {
-					return fmt.Errorf("%s: invalid local plugin", id)
+					return nil, fmt.Errorf("%s: invalid local plugin", id)
+				}
+				if len(path) > 0 {
+					pluginConfig, err = bufgen.NewBinaryPluginConfig(
```
oliversun9, line 159: Going to check `err` after the `switch` statement

private/bufpkg/bufpluginexec/bufpluginexec.go
```diff
@@ -226,6 +253,27 @@ func NewBinaryHandler(runner command.Runner, pluginPath string, pluginArgs []str
 	return newBinaryHandler(runner, pluginPath, pluginArgs), nil
 }
 
+// NewProtocProxyHandler returns a new Handler that invokes the protoc built-in
+// plugin specified by pluginName, in the form of "java".
+func NewProtocProxyHandler(
```
bufdev, line 258: Why does this need to be public?

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -350,46 +334,152 @@ func run(
...
-	if len(fileAnnotations) > 0 {
-		if err := bufanalysis.PrintFileAnnotations(container.Stderr(), fileAnnotations, flags.ErrorFormat); err != nil {
+
+	var (
+		inputImages   []bufimage.Image
+		imageModifier bufimagemodify.Modifier
+		plugins       []bufgen.PluginConfig
+	)
+	switch configVersion {
+	case bufgen.V2Version:
```
bufdev, line 350: Does v2 actually work after this PR? If not, we shouldn't merge in a state to main that will accept it, we should still error until all the work is complete

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -350,46 +334,152 @@ func run(
...
-		return bufcli.ErrFileAnnotation
-	}
-	images := make([]bufimage.Image, 0, len(imageConfigs))
-	for _, imageConfig := range imageConfigs {
-		images = append(images, imageConfig.Image())
-	}
-	image, err := bufimage.MergeImages(images...)
-	if err != nil {
-		return err
+		// TODO: implement managed mode
```
bufdev, line 361: Yea since it's not done yet, we should still error on v2 for now

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -399,38 +489,111 @@ func run(
...
-	return bufgenv1.NewGenerator(
+	generator := bufgen.NewGenerator(
 		logger,
 		storageosProvider,
 		runner,
 		wasmPluginExecutor,
 		clientConfig,
-	).Generate(
+	)
+	for _, image := range inputImages {
```
bufdev, line 511: This represents a change in behavior for v1 - previously, we were merging the image and then calling `Generate` once, now we are looping through resultant images. This will result in extra calls to certain plugins that cannot tolerate them

bufdev (changes requested):
This PR is really difficult to review as there's so much going on - I don't want you to have to split it up, but we'll need to come up with a strategy to review it as I am concerned there are non-trivial changes in behavior to v1 being introduced. We'll need to get a walkthrough somehow of what is happening, and what has changed - it's difficult to diff this to what existed previously.

private/buf/cmd/buf/command/generate/generate.go
oliversun9: This actually merges into `bufgenv2`, and so I think it's ok to do this.
I can also merge into main, but I think it's easy to experiment with (and catch bugs) in such a state.

private/buf/cmd/buf/command/generate/generate.go
oliversun9: Actually, the behavior for `v1` config file is exactly the same:
in the switch case above (when version is `v1`), only one image is added to the list, and this image is already the result of a merge, see `getInputImage`: `image, err := bufimage.MergeImages(images...)`.

In a higher level, each input (one `input` from v2 config counts as one input, and the command line argument also counts as one input) corresponds to one `image` in this `for-loop`. It's worth noting that `getInputImage` transforms one `Ref` to one `Image`.

private/bufpkg/bufpluginexec/bufpluginexec.go
oliversun9: It does not need to be, but the other `NewXYZHandler`s aren't used outside of this package either. I made it public to stay consistent.
I am happy to make all of them unexported, but I would prefer to do this in a separate PR as this one is already difficult to review.

.golangci.yml
oliversun9: fixed

private/buf/bufgen/generator.go
```diff
@@ -431,10 +489,8 @@ func (g *generator) execRemotePluginsV2(
 		ctx,
 		connect.NewRequest(
 			&registryv1alpha1.GenerateCodeRequest{
-				Image:                 bufimage.ImageToProtoImage(image),
-				Requests:              requests,
-				IncludeImports:        includeImports,
```
oliversun9, line 436: `IncludeImports` and `IncludeWellKnownTypes` are set in each `request`

oliversun9:
@bufdev  I updated the PR description to summarize the refactor, and I'm also happy to chat offline on how the old `v1` code path is preserved in this PR, and address other concerns on this PR as well.

private/buf/bufgen/internal/bufgenv1/config.go
```diff
@@ -31,37 +31,103 @@ import (
...
+	PluginConfigs []plugingen.PluginConfig
+	// Optional
+	ManagedConfig *ManagedConfig
+	// Optional
+	TypesConfig *TypesConfig
+}
+
+// ManagedConfig is the managed mode configuration.
+type ManagedConfig struct {
+	CcEnableArenas          *bool
```
oliversun9, line 46: I'm going to un-export these in a separate PR.

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -40,31 +42,63 @@ func SilenceLinter() {
...
+// TODO: likely want something like *string or otherwise, see https://github.com/bufbuild/buf/issues/1949
+// overrideFunc is specific to a file option, and returns what thie file option
+// should be overridden to for this file.
+type overrideFunc func(bufimage.ImageFile) (string, error)
+
+// TODO: unexport these names
+
+// Config is a configuration.
+type Config struct {
+	Managed *ManagedConfig
```
oliversun9, line 57: this will be un-exported in a separate PR

bufdev (approved):
Code overall looks great...notes for the future:

- This is a bit of a departure from our existing setup, in that typically we don't leak things like containers and `bufwire` components to this level. However, I don't hate it, I think it may be a practical cleanup in this situation, and we can evolve working with this paradigm in the future as we try to deprecate `bufwire`.
- This approval is for the actual code - before we eventually merge `bufgenv2` into `main`, we're going to do extensive QA with to check for bugs, but more importantly to test our design assumptions. We may edit some of the UX at that point.

Great work!

PR #2216, opened 2023-06-21
Move the original bufimagemodify to bufimagemodify/bufimagemodifyv1

oliversun9:
This puts `bufimagemodifyv1` and `bufimagemodifyv2` under directory `bufimagemodify`, which is not a package itself. Also creates`bufimagemodify/internal` to host some helpers that will be shared, as well source code info location like `GoPackagePath`
Creates `bufimagemodify/bufimagemodifytesting` to host helpers used by tests that will be shared.

There is no change in behavior or new implementation in this PR.


bufdev (approved):
Nothing controversial here, let's just make sure all the public types are documented in subsequent PRs.

PR #2220, opened 2023-06-23
Fix PR trigger release github actions

joshcarp:
Fixes the PR triggering so that the PRs on other repos should occur automatically only needing merging after they get raised


rubensf (approved):
:lets-get-this-moving:

PR #2226, opened 2023-06-26
Implement Managed Mode v2

oliversun9:
About 62 files (out of 85) are test files added.
Changes in this PR:
* `bufgen/internal/bufgenv2`: filled the rest of the options, and some minor changes, possible to review on web UI
* `bufimagemodifyv1/<file option>.go`: moved helpers from these files to internal
*  `bufimagemodifyv2/bufimagemodifyv2.go`: implemented all `ModifyXYZ`


private/buf/bufgen/internal/bufgenv2/managed_mode.go
```diff
@@ -43,34 +41,77 @@ func applyManagementForFile(
...
 		}
 		var override bufimagemodifyv2.Override
 		if overrideFunc, ok := managedConfig.FileOptionToOverrideFunc[fileOption]; ok {
 			override = overrideFunc(imageFile)
 		}
-		// override can be nil at this point, in which case ModifyXYZ will either use
-		// our default value for the option, or leave the option alone, implicitly
-		// using Protobuf's default value.
+		// override can be nil at this point, which is a valid argument to ModifyXYZ
 		switch fileOption {
```
oliversun9, line 52: perhaps having a `map[fileOption]ModifyFunc` is better here, but that could potentially make the code a little bit (just a little) harder to read.

private/buf/bufgen/internal/bufgenv2/bufgenv2.go
```diff
@@ -166,7 +165,14 @@ func (g *Generator) Generate(
 		)
 	}
 	for _, inputImage := range inputImages {
-		// TODO: modify this image
+		if genConfig.Managed != nil {
```
bufdev, line 168: Shouldn't this be gated on if `managed.enabled = true`?

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -67,6 +43,18 @@ type ManagedConfig struct {
...
 }
 
+// disableFunc decides whether a file option should be disabled for a file.
+type disabledFunc func(FileOption, ImageFileIdentity) bool
+
+// overrideFunc is specific to a file option, and returns what thie file option
+// should be overridden to for this file.
+type overrideFunc func(ImageFileIdentity) bufimagemodifyv2.Override
+
+type ImageFileIdentity interface {
```
bufdev, line 53: What is this, and why is it needed? Is there a reason we can't just use `ImageFile`? If we keep this, why does it need to be private? If it needs to be public, it needs Godoc, including rationale for why it exists.

private/bufpkg/bufimage/bufimagemodify/internal/bufimagemodifytesting/bufimagemodifytesting.go
```diff
@@ -56,7 +72,51 @@ func AssertFileOptionSourceCodeInfoEmpty(t *testing.T, image bufimage.Image, fil
...
 
-func AssertFileOptionSourceCodeInfoNotEmpty(t *testing.T, image bufimage.Image, fileOptionPath []int32) {
+// AssertSourceCodeInfoWithIgnoreWKT skips well known types.
+func AssertSourceCodeInfoWithIgnoreWKT() AssertSourceCodeInfoOption {
+	return func(options *assertSourceCodeInfoOptions) {
+		options.ignoreWKT = true
+	}
+}
+
+type AssertSourceCodeInfoOption func(*assertSourceCodeInfoOptions)
```
bufdev, line 82: needs godoc

private/bufpkg/bufimage/bufimagemodify/internal/bufimagemodifytesting/bufimagemodifytesting.go
```diff
@@ -71,8 +131,31 @@ func AssertFileOptionSourceCodeInfoNotEmpty(t *testing.T, image bufimage.Image,
...
+	for _, location := range descriptor.SourceCodeInfo.Location {
+		if len(location.Path) > 0 && internal.Int32SliceIsEqual(location.Path, fileOptionPath) {
+			hasFileOption = true
+			break
+		}
+	}
+	assert.True(t, hasFileOption)
+}
+
+// GetTestImage returns an image from a directory.
```
bufdev, line 152: needs justification as to why we need this, and why its not part of a common testing package for images (which likely means we need a TODO to move such a function into a common `bufimagetesting`).

private/buf/bufgen/internal/bufgenv2/bufgenv2.go
oliversun9: Now using `Enabled` to gate image modifying (I kept the field public to stay consistent with the other fields, it will be made private and there is a TODO that will make these private)

private/buf/bufgen/internal/bufgenv2/config.go
oliversun9: This is made private now. The only thing `matchesPathAndModule` from the image file its `Path()` and `ModuleIdentity()`. 
Having this interface makes it easier in tests: I can define a struct that that implement these two methods, and I can create many instances of this struct, and call 
```
overrideFunc, ok := config.Managed.FileOptionToOverrideFunc[fileOption]
require.True(t, ok)
actual := overrideFunc(instanceOfThisStruct)
```
and compare `actual` with my expected override

private/bufpkg/bufimage/bufimagemodify/internal/bufimagemodifytesting/bufimagemodifytesting.go
oliversun9: Updated the doc explaining it allows skipping creating a module, and also added a TODO to move this to `bufimagetesting`.

private/buf/bufgen/internal/bufgenv2/testdata/managed/php_namespace_error_1.json
```diff
@@ -0,0 +1,18 @@
...
+    {
+      "binary": "protoc-gen-go",
+      "out": "gen/out"
+    }
+  ],
+  "managed": {
+    "enable": true,
+    "override": [
+      {
+        "file_option": "php_namespace",
```
bufdev, line 13: Commenting on just one file here - seeing this again, did we agree to have "value" and "prefix"? I thought we changed to only having "value"?

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,442 @@
...
+	var phpNamespaceValue string
+	switch t := override.(type) {
+	case valueOverride[string]:
+		phpNamespaceValue = t.get()
+	case nil:
+		phpNamespaceValue = internal.DefaultPhpNamespaceValue(imageFile)
+	default:
+		return errors.New("a valid override is required for php_namespace")
+	}
+	if internal.IsWellKnownType(imageFile) {
```
bufdev, line 410: Note to self: look at this function tomorrow

private/buf/bufgen/internal/bufgenv2/testdata/managed/php_namespace_error_1.json
oliversun9: `value` + `prefix` is what we have in the design doc. I saw you left comments on other parts of the doc but not this one, and so I assumed that you agreed with it. Of course we can always revisit this decision. My reasonings are:
* It would be slightly confusing if a user defines both `go_package` and `go_package_prefix` and only one takes effect, even though they go with different names. On the other hand, it would be more obvious that only one of `go_package.prefix` and `go_package.value` takes effect on a file.
* If all file options are defined in `descriptor.proto` (e.g. `go_package` is defined there but not `go_package_prefix`), the options will be more conisten. A user will not try to look for `go_package_prefix` in `descriptor.proto` and feel confused after not finding it.
* `objc_class_prefix` is defined in `descriptor.proto` but not `go_package_prefix`, this adds to the last point.

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,442 @@
...
 	marker Marker,
 	imageFile bufimage.ImageFile,
-	prefix string,
+	override Override,
 ) error {
+	var javaPackageValue string
+	switch t := override.(type) {
+	case prefixOverride:
+		javaPackageValue = internal.JavaPackageValue(imageFile, t.get())
+	case valueOverride[string]:
```
bufdev, line 77: A `value` for a straight `java_package` doesn't make sense - you're saying "for all code generated by this `buf.gen.yaml`, regardless of the package of a given `.proto` file, use exactly this java package value". Same for `go_package`, and others likely.

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,442 @@
...
+		javaPackageValue = t.get()
+	case nil:
```
bufdev, line 79: The nil case shouldn't be valid. All overrides should have a `value` or `prefix` - if you just want the default value for managed mode, you don't set an override. Same comment whole way down

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,442 @@
...
+		javaPackageValue = internal.JavaPackageValue(imageFile, internal.DefaultJavaPackagePrefix)
+	default:
+		return errors.New("a valid override is required for java_package")
```
bufdev, line 82: return `fmt.Errorf("unknown override type: %T", override)`

oliversun9:
Closing this because it's implemented in https://github.com/bufbuild/buf/pull/2344

PR #2227, opened 2023-06-26
Add ListGitCommitMetadataForReference rpc

doriable:
This adds an RPC to the `ReferenceService` that lists out
the git commits and metadata, if any, associated with the
given reference string.

This also removes the `GitMetadata` optional field in `RepositoryCommit` which
is not currently used.


proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -72,3 +78,17 @@ message GetReferenceByNameRequest {
...
+  // Owner of the repository the reference belongs to.
+  string owner = 2;
+  // Name of the repository the reference belongs to.
+  string repository_name = 3;
+}
+
+message ListGitCommitMetadataForReferenceResponse {
+  // List of git commits and metadata associated with the resolved reference.
+  repeated GitCommitMetadata git_commit_metadatas = 1;
+}
```
unmultimedio, line 94: I'd also return the BSR commit id or commit name (or both) that was resolved from the passed reference.

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/reference.proto
doriable: Oh that's a good idea -- I think ID should be reasonable, since if we need to resolve additional information, the ID would be most useful.

unmultimedio (approved):

PR #2231, opened 2023-06-28
Add module identity to external path

rhbuf:
The `ExternalPath` of an `ObjectInfo` "is not necessarily a file path, and should only be used to uniquely identify this file as compared to other assets, for display to users." ([doc](https://github.com/bufbuild/buf/blob/dac9471b15301ea7d7ddaafd95250244fa26d104/private/pkg/storage/bucket.go#L171-L173)) - Currently when a compile error happens due to duplicate file paths the context of which modules those files come from is not provided.
```
{
    "path": "b.proto",
    "start_line": 5,
    "start_column": 8,
    "end_line": 5,
    "end_column": 8,
    "type": "COMPILE",
    "message": "google/type/datetime.proto exists in multiple locations: google/type/datetime.proto google/type/datetime.proto"
}
```

This change adds the module identity to the external path to make it clearer for users that are experiencing the `errorExistsMultipleLocations` compile error:
```json
{
    "path": "b.proto",
    "start_line": 5,
    "start_column": 8,
    "end_line": 5,
    "end_column": 8,
    "type": "COMPILE",
    "message": "google/type/datetime.proto exists in multiple locations: bufbuild.internal/googleapis/googleapis:3a85c4762b07439aa574260257696aea/google/type/datetime.proto bufbuild.internal/rhenderson/googleapis:cc0a1eae8c714a5e9b590b84315ed810/google/type/datetime.proto"
}
```

I'm cautious to merge this change as `ExternalPath` appears to be used in some contexts where it is definitely expected to be a file path: https://github.com/bufbuild/buf/blob/4d59b5eacf91c675e4231c15fdf4a550a62452bf/private/buf/cmd/buf/command/format/format.go#L526

### DIFF
--- a/private/bufpkg/bufmodule/single_module_read_bucket.go
+++ b/private/bufpkg/bufmodule/single_module_read_bucket.go
@@ -40,6 +40,10 @@ func newSingleModuleReadBucket(
 	}
 }
 
+func (r *singleModuleReadBucket) Stat(ctx context.Context, path string) (storage.ObjectInfo, error) {
+	return r.StatModuleFile(ctx, path)
+}
+
 func (r *singleModuleReadBucket) StatModuleFile(ctx context.Context, path string) (*moduleObjectInfo, error) {
 	objectInfo, err := r.ReadBucket.Stat(ctx, path)
 	if err != nil {
--- a/private/bufpkg/bufmodule/module_object_info.go
+++ b/private/bufpkg/bufmodule/module_object_info.go
@@ -15,6 +15,8 @@
 package bufmodule
 
 import (
+	"fmt"
+
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule/bufmoduleref"
 	"github.com/bufbuild/buf/private/pkg/storage"
 )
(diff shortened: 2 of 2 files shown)


rhbuf:
@unmultimedio any suggestions on what I can do to increase confidence in this change?

unmultimedio:
> @unmultimedio any suggestions on what I can do to increase confidence in this change?

Probably if you have access to the `moduleObjectInfo` from the place where the error is triggered, you could manually put the module identity int he error msg instead of intercepting `ExternalPath`.

private/bufpkg/bufmodule/single_module_read_bucket.go
```diff
@@ -40,6 +40,10 @@ func newSingleModuleReadBucket(
 	}
 }
 
+func (r *singleModuleReadBucket) Stat(ctx context.Context, path string) (storage.ObjectInfo, error) {
```
bufdev, line 43: This shouldn't be needed at all - this is just calling into the original `Stat` via `StatModuleFile`. I see what you're doing here,  but I think this could have other unintended effects - if you're going to steathily edit the result of `ExternalPath`, you need to do this for every bucket call throughout `bufmodule` so this is consistent.

bufdev (changes requested):
I'm deeply concerned with this change - while `ExternalPath` was not meant to be used as anything other than an informational item, it appears that it has been used (ie in format) as an actual external path (incorrectly). Without a fuller audit, I am concerned this could break a lot of our code, and without a large benefit in return.

bufdev:
The right way to do this is likely us including some concept of "ID" on a bucket, optionally, and then when this error is created (also within the `storage` package), use this ID as part of the error message.

rhbuf:
Closing this out for now, we can re-approach with @bufdev's suggestion  in a separate PR

PR #2232, opened 2023-06-28
Fix bugs in ModuleFileSetBuilder and Workspace

bufdev:
See code comments. This is also starting to undo the mess in the workspace code, this is a long story.

We don't have super high confidence in this PR, but we think it puts us in a better state than we were in. It may cause bugs, although I have extensively QA'ed this.

### DIFF
--- a/private/buf/bufwork/bufwork.go
+++ b/private/buf/bufwork/bufwork.go
@@ -98,7 +98,9 @@ var (
 // WorkspaceBuilder builds workspaces. A single WorkspaceBuilder should NOT be persisted
 // acorss calls because the WorkspaceBuilder caches the modules used in each workspace.
 type WorkspaceBuilder interface {
-	// BuildWorkspace builds a bufmodule.Workspace for the given targetSubDirPath.
+	// BuildWorkspace builds a bufmodule.Workspace.
+	//
+	// The given targetSubDirPath is the only path that will have the configOverride applied to it.
 	BuildWorkspace(
 		ctx context.Context,
 		workspaceConfig *Config,
--- a/private/buf/bufwork/bufwork.go
+++ b/private/buf/bufwork/bufwork.go
@@ -117,10 +117,8 @@ type WorkspaceBuilder interface {
 }
 
 // NewWorkspaceBuilder returns a new WorkspaceBuilder.
-func NewWorkspaceBuilder(
-	moduleBucketBuilder bufmodulebuild.ModuleBucketBuilder,
-) WorkspaceBuilder {
-	return newWorkspaceBuilder(moduleBucketBuilder)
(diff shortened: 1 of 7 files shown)


doriable (approved):

private/buf/bufwork/workspace_builder.go
```diff
@@ -64,9 +60,6 @@ func (w *workspaceBuilder) BuildWorkspace(
 	allModules := make([]bufmodule.Module, 0, len(workspaceConfig.Directories))
 	for _, directory := range workspaceConfig.Directories {
 		if cachedModule, ok := w.moduleCache[directory]; ok {
-			if directory == targetSubDirPath {
```
bufdev, line 67: For posterity: workspaces were being created that specifically *excluded* the "target" module - this caused other bugs down the line. I have no idea why this worked, but a workspace should include all the modules within it.

private/buf/bufwork/workspace_builder.go
```diff
@@ -28,16 +28,12 @@ import (
 )
 
 type workspaceBuilder struct {
-	moduleBucketBuilder bufmodulebuild.ModuleBucketBuilder
-	moduleCache         map[string]*cachedModule
+	moduleCache map[string]*cachedModule
 }
 
-func newWorkspaceBuilder(
-	moduleBucketBuilder bufmodulebuild.ModuleBucketBuilder,
```
bufdev, line 36: For posterity: this `ModuleBucketBuilder` was never used.

private/buf/bufwork/bufwork.go
```diff
@@ -98,7 +98,9 @@ var (
 // WorkspaceBuilder builds workspaces. A single WorkspaceBuilder should NOT be persisted
 // acorss calls because the WorkspaceBuilder caches the modules used in each workspace.
 type WorkspaceBuilder interface {
-	// BuildWorkspace builds a bufmodule.Workspace for the given targetSubDirPath.
+	// BuildWorkspace builds a bufmodule.Workspace.
+	//
+	// The given targetSubDirPath is the only path that will have the configOverride applied to it.
```
bufdev, line 103: We likely want to eventually eliminate `targetSubDirPath` from workspace construction.

private/bufpkg/bufmodule/workspace.go
```diff
@@ -23,13 +28,30 @@ type workspace struct {
 }
 
 func newWorkspace(
+	ctx context.Context,
 	namedModules map[string]Module,
 	allModules []Module,
-) *workspace {
+) (*workspace, error) {
+	pathToExternalPaths := make(map[string][]string)
```
bufdev, line 35: This enforces the property that workspaces must have unique files

doriable (approved):

PR #2233, opened 2023-06-29
Change ModuleConfigReader to return an interface with a single Workspace

bufdev:
This changes `ModuleConfigReader` to return the new `ModuleConfigSet` interface, which has a `Workspace` function, and deletes `ModuleConfig.Workspace()`. This has the effect of returning 0 or 1 workspaces for the `ModuleConfigs` returned from the `ModuleConfigReader`.

There is no valid case where you can have an input to `ModuleConfigReader` that results in multiple logical workspaces, and this is by design - if you are building a set of `Modules`, this is because they came from a (single) workspace.

During this change, I believe a bug is fixed for where `--config` overrides were being applied to all `Modules` within a `Workspace`, even though the config override was only meant to apply to a single `Module`.

There's tons of cleanup work to do after this, and understanding we need to build for the `Workspace` code. I've added some comments to some key functions as I went through to help better understand what they do in the future.

@doriable, I manually QA'ed this extensively, with various inputs, and values for `--path, --exclude-path, --config`, and I wasn't able to break it. That, combined with the relatively-extensive testing we have for workspaces, gives me some level of confidence that this isn't making the situation worse. However, as with any edits we do to in the workspace code, this is still a relatively high-risk change.

### DIFF
--- a/private/buf/bufwire/bufwire.go
+++ b/private/buf/bufwire/bufwire.go
@@ -75,22 +75,28 @@ func NewImageConfigReader(
 	)
 }
 
-// ModuleConfig is an module and configuration.
+// ModuleConfig is a Module and configuration.
 type ModuleConfig interface {
 	Module() bufmodule.Module
 	Config() *bufconfig.Config
+}
+
+// ModuleConfigSet is a set of ModuleConfigs with a potentially associated Workspace.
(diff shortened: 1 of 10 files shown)


doriable (approved):
This seems very reasonable to me.

PR #2237, opened 2023-06-29
Delete bufmodulebuild.BuildForBucket

bufdev:
This function should never have existed - it doesn't fit in with the pattern of the rest of the codebase, it should have been in `bufmodulebuild.go`, `BuildOptions` should not be passed to its constructor, etc. This has been a longstanding thorn, and this just cleans it up.

### DIFF
--- a/private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
+++ b/private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
@@ -26,58 +26,36 @@ import (
 	"github.com/bufbuild/buf/private/pkg/storage/storagemem"
 )
 
-// BuiltModule ties a bufmodule.Module with the configuration and a bucket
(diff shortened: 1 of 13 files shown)


private/bufpkg/bufmodule/bufmodulebuild/module_bucket_builder.go
```diff
@@ -26,58 +26,36 @@ import (
...
-	bm, err := builder.BuildForBucket(
+	buildOptions := &buildOptions{}
+	for _, option := range options {
+		option(buildOptions)
+	}
+	return b.buildForBucket(
 		ctx,
 		readBucket,
 		config,
+		buildOptions,
```
unmultimedio, line 50: 👍🏼 Feels more natural to pass them here instead of within the builder.

unmultimedio (approved):

PR #2240, opened 2023-06-29
Improve private/buf/dag

bufdev:
- Add `NumNodes` to return the number of nodes.
- Add `NumEdges` to return the number of edges.
- Rename `Walk` to `WalkEdges`
- Fix a bug where `WalkEdges` incorrectly returned an error when there were only nodes in the graph with no edges.
- Add `WalkNodes` that walks the nodes. This allows access to nodes with no edges.
- Update `DOTString` to print nodes without edges.
- Add testing.

### DIFF
--- a/private/pkg/dag/dag.go
+++ b/private/pkg/dag/dag.go
@@ -266,6 +266,9 @@ func (g *Graph[Key]) DOTString(keyToString func(Key) string) (string, error) {
 	); err != nil {
 		return "", err
 	}
+	if len(nodeStrings) == 0 {
+		return "digraph {}", nil
+	}
 	buffer := bytes.NewBuffer(nil)
 	_, _ = buffer.WriteString("digraph {\n\n")
 	for _, nodeString := range nodeStrings {
--- a/private/pkg/dag/dag.go
+++ b/private/pkg/dag/dag.go
@@ -258,6 +258,7 @@ func (g *Graph[Key]) DOTString(keyToString func(Key) string) (string, error) {
 					fmt.Sprintf("%d", nextIndex),
 				)
 				nextIndex++
+				return nil
 			}
 			// This is a system error.
 			return fmt.Errorf("got node %v with %d inbound edges and %d outbound edges, but this was not processed during WalkEdges", key, len(inboundEdges), len(outboundEdges))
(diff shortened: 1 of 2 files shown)


private/pkg/dag/dag.go
```diff
@@ -195,6 +241,34 @@ func (g *Graph[Key]) DOTString(keyToString func(Key) string) (string, error) {
 	); err != nil {
 		return "", err
 	}
+	// We also want to pick up any nodes that do not have edges, and display them.
+	if err := g.WalkNodes(
+		func(key Key, inboundEdges []Key, outboundEdges []Key) error {
```
joshcarp, line 246:
```suggestion
		func(key Key, inboundEdges, outboundEdges []Key) error {
```

private/pkg/dag/dag.go
```diff
@@ -195,6 +241,34 @@ func (g *Graph[Key]) DOTString(keyToString func(Key) string) (string, error) {
...
+			//
```
joshcarp, line 247:
```suggestion
```

doriable (approved):

private/pkg/dag/dag_test.go
```diff
@@ -180,9 +180,9 @@ func TestWalk2(t *testing.T) {
 	)
 }
 
-func TestWalk3(t *testing.T) {
+func TestWalkEdges3(t *testing.T) {
```
joshcarp, line 183: nit: test names should be more descriptive instead of `Test<Num>`

private/pkg/dag/dag.go
```diff
@@ -93,11 +93,57 @@ func (g *Graph[Key]) ContainsNode(key Key) bool {
...
+}
+
+// WalkEdges visits each edge in the Graph starting at the source keys.
+//
+// f is called for each directed edge. The first argument is the source
+// node, the second is the destination node.
 //
 // Returns a *CycleError if there is a cycle in the graph.
-func (g *Graph[Key]) Walk(f func(Key, Key) error) error {
+func (g *Graph[Key]) WalkEdges(f func(Key, Key) error) error {
```
unmultimedio, line 141: This could be in the func signature instead of the doc

```suggestion
// f is called for each directed edge.
//
// Returns a *CycleError if there is a cycle in the graph.
func (g *Graph[Key]) WalkEdges(f func(sourceNode Key, destinationNode Key) error) error {
```

PR #2241, opened 2023-06-29
Uniqueness policy setting and state rpcs

pgmitche:


proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+  string organization_id = 2;
+  string repository_id = 3;
+  string filepath = 4;
+}
+
+enum CollisionType {
+  COLLISION_TYPE_UNSPECIFIED = 0;
+  COLLISION_TYPE_FQN = 1;
+  COLLISION_TYPE_FILEPATH = 2;
+}
```
pgmitche, line 65: @stefanvanburen I think this part of the change is probably the part that warrants some thought/feedback

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
   BreakingChangeCategory category = 3;
 }
 
+message UniquenessPolicy {
+  optional bool enabled = 1;
```
stefanvanburen, line 39: assuming `enabled` entails turning on uniqueness for both FQN && filepaths, right?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+}
+
+message UniquenessState {
+  bool has_collisions = 1;
+  repeated Collision collisions = 2;
```
stefanvanburen, line 44: we could probably reduce this to just the `repeated Collisions` which would make `has_collisions` == `len(collisions) != 0`?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+}
+
+message Collision {
+  string id = 1;
+  CollisionType type = 2;
+  string reference = 3;
```
stefanvanburen, line 50: What does reference mean here?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+  repeated CollisionSource sources = 4;
+}
+
+message CollisionSource {
+  string id = 1;
```
stefanvanburen, line 55: do we need the `id`s here and for `Collision`? my only hesitation is that if I resolve a collision and then later introduce "the same" collision, we aren't going to be assigning stable IDs, so I'm curious if you see a use-case for them. (happy to leave them in, just wondering!)

stefanvanburen (commented):
makes sense, just a few questions

proto/buf/alpha/registry/v1alpha1/admin.proto
pgmitche: either the fully qualified name (foo.Bar.Baz)
or proto filepath (a/b/c/v1alpha1/d.proto)

proto/buf/alpha/registry/v1alpha1/admin.proto
pgmitche: yeah, as far as I understand it, there's no ask to have these enabled separately.
I think the idea is that typical proto uniqueness enforcement, like FQN collisions and filename collisions are all avoided at once

bufdev (commented):
Everything here needs full documentation (all types, rpcs) - if I'm an external reader, I should understand what is going on.

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+  string organization_id = 2;
+  string repository_id = 3;
+  string filepath = 4;
+}
+
+enum CollisionType {
+  COLLISION_TYPE_UNSPECIFIED = 0;
+  COLLISION_TYPE_FQN = 1;
```
bufdev, line 63: Why using `FQN` abbreviation?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+  CollisionType type = 2;
+  string reference = 3;
+  repeated CollisionSource sources = 4;
+}
+
+message CollisionSource {
+  string id = 1;
+  string organization_id = 2;
+  string repository_id = 3;
+  string filepath = 4;
```
bufdev, line 58: We use `path` everywhere else, why are we using `filepath` here?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,35 @@ message BreakingChangePolicy {
...
+message Collision {
+  string id = 1;
+  CollisionType type = 2;
+  string reference = 3;
+  repeated CollisionSource sources = 4;
+}
+
+message CollisionSource {
+  string id = 1;
+  string organization_id = 2;
```
bufdev, line 56: Why is this `organization_id` and not `owner_id`?

proto/buf/alpha/registry/v1alpha1/admin.proto
bufdev: +1

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,45 @@ message BreakingChangePolicy {
...
+  // The type of collision.
+  CollisionType type = 2;
+  // The sources of the collision.
+  repeated CollisionSource sources = 3;
+}
+
+// CollisionSource is the source of a collision
+message CollisionSource {
+  string owner_id = 1;
+  string repository_id = 2;
```
pgmitche, line 63: Probably should be names instead, if kept

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,45 @@ message BreakingChangePolicy {
...
+  string path = 3;
+}
```
pgmitche, line 65: I'm not sure it's worth structuring the collision like this for the admin service.

I think we might be better off just building some informative error(s) in the application and returning them.

e.g. 

```
// UniquenessState is the state of collisions on the server.
message UniquenessState {
  // The existing collisions on the server.
  repeated string collisions = 1;
}
```

Where a collision is something like:
```
Fully qualified name "foo.Bar.Baz" is found { N } times: 
 - {owner}/{repository} in foo/bar/baz.proto
 - {owner}/{repository} in types/v1beta1/baz.proto
 ```
 ```
Proto file path "types/v1alpha1/datetime.proto" is found { N } times:
- {owner}/{repository}
- {owner}/{repository}
- {owner}/{repository}
```
 
@stefanvanburen, @rebello95  What do you think? This would be different to what users might see on push.

rebello95: I like the format of those errors, but I agree with @stefanvanburen that the API should be structured and allow the frontend decide how to show the info

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,45 @@ message BreakingChangePolicy {
...
+message Collision {
+  // The reference that has collisions.
+  string reference = 1;
+  // The type of collision.
+  CollisionType type = 2;
+  // The sources of the collision that the user is authorized to see.
+  repeated CollisionSource sources = 3;
+  // The number of collisions that were not returned.
+  // The user should contact a server admin for more information about these collisions.
+  uint64 num_other_collisions = 4;
```
stefanvanburen, line 54: 👍 on thinking about this, but maybe we should just start with `has_other_collisions` to keep things as locked down as possible to start with?

It also feels a little weird to me that a `message Collision` has a field named `num_other_collisions`, maybe it should be `has_other_collision_sources`?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -56,6 +95,16 @@ service AdminService {
...
   // Update breaking change policy for the server.
   rpc UpdateBreakingChangePolicy(UpdateBreakingChangePolicyRequest) returns (UpdateBreakingChangePolicyResponse);
+  // Get uniqueness policy for the server.
+  rpc GetUniquenessPolicy(GetUniquenessPolicyRequest) returns (GetUniquenessPolicyResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
+  // Update uniqueness policy enforcement for the server.
+  rpc UpdateUniquenessPolicy(UpdateUniquenessPolicyRequest) returns (UpdateUniquenessPolicyResponse);
+  // Get state of uniqueness collisions for the server
+  rpc GetServerUniquenessCollisions(GetServerUniquenessCollisionsRequest) returns (GetServerUniquenessCollisionsResponse) {
```
stefanvanburen, line 105: I don't love this, but I _think_ this should technically be `ListServerUniquenessCollisions` and have pagination a la https://google.aip.dev/132.

stefanvanburen (commented):
looking good, two other thoughts but I think this should be set otherwise?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -35,6 +35,45 @@ message BreakingChangePolicy {
...
+  // The user should contact a server admin for more information about these collisions.
+  bool has_other_collision_sources = 4;
+}
+
+// CollisionSource is the source of a collision
+message CollisionSource {
+  // The name of the owner of the source.
+  string owner_name = 1;
+  // The name of the repository of the source.
+  string repository_name = 2;
```
stefanvanburen, line 62: one final note was that I was curious on the relative usage of `owner_name` and `repository_name` versus something like `owner` and `repository`, and it _seems_ like we have _more_ of the latter, but I think we should keep this as-is because it's more explicit, and for something like `repository` we have both:

```proto
Repository repository = 1; // buf.alpha.registry.v1alpha1.Repository
```
and
```proto
string repository = 2; // aka, repository name, usually clarified in comments
```

tl;dr: keep these as-is

stefanvanburen (approved):

PR #2249, opened 2023-06-30
Update aur badge in README.md

joshcarp:


bufdev (changes requested):
I would just remove entirely since it's in the mainline AUR now

bufdev (approved):

PR #2252, opened 2023-07-01
Support .binproto/.textproto files in convert

Alfus:
These file types are used for CEL conformance tests; however, prototext support is a somewhat rare feature. This command lets .textproto files be converted to .json or .binproto files for implementations that cannot read prototext.

Example output:
```
% buf convert buf.build/alfus/cel --type dev.cel.expr.conformance.SimpleTestFile --from ../cel-spec/tests/simple/testdata/basic.textproto --to -#format=json --indent
{
  "name": "basic",
  "description": "Basic conformance tests that all implementations should pass.",
  "section": [
    {
      "name": "self_eval_zeroish",
      "description": "Simple self-evaluating forms to zero-ish values.",
      "test": [
        {
          "name": "self_eval_int_zero",
          "expr": "0",
          "value": {
            "int64Value": "0"
          }
        },
        ...
```

### DIFF
--- a/private/buf/bufconvert/bufconvert.go
+++ b/private/buf/bufconvert/bufconvert.go
@@ -30,10 +30,16 @@ import (
 const (
 	// MessageEncodingBin is the binary image encoding.
 	MessageEncodingBin MessageEncoding = iota + 1
+	// MessageEncodingText is the text image encoding.
+	MessageEncodingText
 	// MessageEncodingJSON is the JSON image encoding.
 	MessageEncodingJSON
 	// formatBin is the binary format.
 	formatBin = "bin"
+	// formatBinProto is also used for binary proto format.
+	formatBinProto = "binproto"
+	// formatText is the proto text format.
+	formatTextProto = "textproto"
 	// formatJSON is the JSON format.
 	formatJSON = "json"
 )
@@ -46,6 +52,8 @@ var (
 	// sorted
 	messageEncodingFormats = []string{
 		formatBin,
+		formatBinProto,
+		formatTextProto,
 		formatJSON,
 	}
 )
@@ -105,8 +113,10 @@ func getPathAndMessageEncoding(
 
 func parseMessageEncodingExt(ext string, defaultEncoding MessageEncoding) MessageEncoding {
 	switch strings.TrimPrefix(ext, ".") {
-	case formatBin:
+	case formatBin, formatBinProto:
 		return MessageEncodingBin
+	case formatTextProto:
+		return MessageEncodingText
 	case formatJSON:
 		return MessageEncodingJSON
 	default:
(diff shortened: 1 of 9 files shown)


private/buf/bufconvert/bufconvert.go
```diff
@@ -46,6 +52,8 @@ var (
 	// sorted
 	messageEncodingFormats = []string{
 		formatBin,
+		formatBinProto,
+		formatTextProto,
 		formatJSON,
```
joshcarp, line 57:
```suggestion
		formatJSON,
		formatTextProto,
```
because of the `//sorted` comment

private/buf/bufconvert/bufconvert.go
```diff
@@ -30,10 +30,16 @@ import (
...
 	// MessageEncodingBin is the binary image encoding.
 	MessageEncodingBin MessageEncoding = iota + 1
+	// MessageEncodingText is the text image encoding.
+	MessageEncodingText
 	// MessageEncodingJSON is the JSON image encoding.
 	MessageEncodingJSON
 	// formatBin is the binary format.
 	formatBin = "bin"
+	// formatBinProto is also used for binary proto format.
+	formatBinProto = "binproto"
```
joshcarp, line 40: I think this might raise some questions around being able to do things more than one way seeing that this is the same as just bin

bufdev: Yes - this is effectively just an alias for `bin`, which isn't necessary - we only have one way to do things here, and we chose the name `bin`. This should be deleted.

private/buf/bufconvert/bufconvert.go
```diff
@@ -30,10 +30,16 @@ import (
...
+	// formatTextProto is the proto text format.
+	formatTextProto = "textproto"
```
bufdev, line 42: `formatText = "text"`

private/buf/cmd/buf/command/convert/convert.go
```diff
@@ -221,15 +229,18 @@ func run(
 		image,
 		message,
 		outputMessageRef,
+		flags.Indent,
 	)
 }
 
 // inverseEncoding returns the opposite encoding of the provided encoding,
 // which will be the default output encoding for a given payload encoding.
-func inverseEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEncoding, error) {
```
bufdev, line 229: Why is this renamed? It seems like it is still attempting to do the inverse.

private/buf/cmd/buf/command/convert/convert.go
```diff
@@ -221,15 +229,18 @@ func run(
...
+func getDefaultEncoding(encoding bufconvert.MessageEncoding) (bufconvert.MessageEncoding, error) {
 	switch encoding {
 	case bufconvert.MessageEncodingBin:
 		return bufconvert.MessageEncodingJSON, nil
+	case bufconvert.MessageEncodingText:
```
bufdev, line 242: Adding a third option in here makes me wonder if we need to rethink the UX of this command. Why is `bin` the inverse of `text`, and vice-versa?

bufdev (changes requested):

private/buf/bufconvert/bufconvert.go
Alfus: I do not feel comfortable saying all .bin or .text files are proto encoded. Not do I want people to have to specify #format= when they are using more explicit and common files extensions.

PR #2256, opened 2023-07-04
Sync multibranch

unmultimedio:
Support multibranch sync in git repos. Loop over all of the remote branches, up to the latest pushed commits.

Commits will be assigned to a single branch, starting with the repo's base branch, which will travel from HEAD up via its first parent, until finding a commit w/out a parent (presumably the  first commit of the repo), then it will loop the remaining branches in lexicographical order, assigning commits in the same way, from HEAD via its first parent, until finding a commit that's been assigned to another branch already.


proto/buf/alpha/registry/v1alpha1/sync.proto
```diff
@@ -82,3 +82,11 @@ message SyncGitCommitResponse {
...
   GitSyncPoint sync_point = 1;
 }
+
+// Branch holds branch information when pushed as a sync git point.
+message Branch {
+  // Name of the branch.
+  string name = 1;
+  // Is this branch the default branch in the repo.
+  bool is_repo_default = 2;
+}
```
unmultimedio, line 92: @doriable @stefanvanburen I'm not 100% this is required. This is so the BSR knows which commits should be marked as "HEAD" label. Am I correct assuming that namespace label will only apply to the default repo branch?

proto/buf/alpha/registry/v1alpha1/sync.proto
```diff
@@ -82,3 +82,11 @@ message SyncGitCommitResponse {
   // SyncPoint is the latest syncpoint for the SyncGitCommit request.
   GitSyncPoint sync_point = 1;
 }
+
+// Branch holds branch information when pushed as a sync git point.
+message Branch {
+  // Name of the branch.
+  string name = 1;
+  // Is this branch the default branch in the repo.
+  bool is_repo_default = 2;
```
bufdev, line 91: This feels like it shouldn't be in the API

bufdev (commented):
What happens if the sync fails in the middle at various points? How does resume work?

proto/buf/alpha/registry/v1alpha1/sync.proto
unmultimedio: I'm not a fan of this either, but the BSR has no way of knowing which one is the default repo branch to mark it as so, that's only known by the CLI when navigating the git repo locally. Any suggestions? Another option is to make it configurable in the BSR, but that would open the possibility of having a different default branch in git vs BSR, which doesn't sound attractive imo.

bufdev: I think I'd need to understand what the behavior here is supposed to be a bit better - are we no longer having the concept of a single `main` branch on the BSR, instead letting the user name this as they please? Regardless, making this a per-RPC option seems bad and open to error, this should be a setting somewhere.

unmultimedio:
> What happens if the sync fails in the middle at various points? How does resume work?

It depends on the failure. If it's a transient failure like the BSR being down, then the BSR stores the git tags and resumes gracefully from the latest synced git commit. If it's an issue with the content in the commit itself, to keep respecting the correct order, the command won't block any more commits. Probably we'll need to add an "skip-failures" flag.

doriable (approved):
Approving this to unblock additional work on multi-branch sync and building upon the commit and branch ordering :)

PR #2257, opened 2023-07-05
Run copyfrommakego to update dependencies

pkwarren:
Update to connect-go 1.9.0 and the latest golangci-lint.


.golangci.yml
```diff
@@ -35,7 +35,7 @@ linters:
     - containedctx
     # - contextcheck
     - decorder
-    - depguard
+    # - depguard
```
pkwarren, line 38: depguard starting in 1.53.0 prevents importing anything other than the stdlib by default. If we want to configure this linter, we should do so in a follow up.

private/buf/buffetch/internal/reader.go
```diff
@@ -232,7 +232,7 @@ func (r *reader) getArchiveBucket(
 		if size < 0 {
 			data, err := io.ReadAll(readCloser)
 			if err != nil {
-				return
+				return nil, err
```
pkwarren, line 235:
```
private/buf/buffetch/internal/reader.go:235:5: naked return in func `getArchiveBucket` with 107 lines of code (nakedret)
                                return
```

This looks to be a valid issue caught by the linter - not sure why it just started reporting it now.

private/buf/cmd/buf/buf_test.go
```diff
@@ -2279,7 +2279,7 @@ func TestConvertRoundTrip(t *testing.T) {
 	)
 	t.Run("stdin and stdout", func(t *testing.T) {
 		t.Parallel()
-		stdin := bytes.NewBuffer([]byte(`{"one":"55"}`))
+		stdin := bytes.NewBufferString(`{"one":"55"}`)
```
pkwarren, line 2282: The is the new `mirror` linter that recommends appropriate functions if ones exist for both string/[]byte.

stefanvanburen (approved):

PR #2259, opened 2023-07-05
Update method that returns unexported struct

pkwarren:
Update the observabilityzap package to return an interface instead of a pointer to an unexported type.

### DIFF
--- a/private/pkg/observabilityzap/observabilityzap.go
+++ b/private/pkg/observabilityzap/observabilityzap.go
@@ -15,21 +15,30 @@
 package observabilityzap
 
 import (
+	"io"
+
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/propagation"
-	"go.opentelemetry.io/otel/sdk/trace"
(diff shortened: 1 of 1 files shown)


private/pkg/observabilityzap/observabilityzap.go
```diff
@@ -15,21 +15,30 @@
...
 
+// TracerProviderCloser is used to wrap a trace.TracerProvider with an io.Closer to use on shutdown.
+type TracerProviderCloser interface {
+	trace.TracerProvider
+	io.Closer
+}
+
 // Start creates a Zap logging exporter for Opentelemetry traces and returns
 // the exporter. The exporter implements io.Closer for clean-up.
-func Start(logger *zap.Logger) *tracerProviderCloser {
```
pkwarren, line 26: Noticed this was returning an pointer to an unexported type (which isn't expected on a public function). Updated to return an interface instead (that closer matches the code style).

doriable (approved):

PR #2265, opened 2023-07-07
Deprecate bin format and add binpb format as replacement

bufdev:
This deprecates the `bin` format and replaces it with the `binpb` format. `format=bin` and `.bin` files are still accepted, but `binpb` and `.binpb` are the new defaults.


Alfus (approved):

private/buf/bufconvert/bufconvert.go
```diff
@@ -28,14 +28,16 @@ import (
...
 	// MessageEncodingJSON is the JSON image encoding.
 	MessageEncodingJSON
-	// formatBin is the binary format.
-	formatBin = "bin"
+	// formatBinpb is the binary format.
+	formatBinpb = "bin"
 	// formatJSON is the JSON format.
 	formatJSON = "json"
+	// formatBin is the binary format's old form, now deprecated.
+	formatBin = "binpb"
```
jhump, line 40: nit: looks like the actual values for the two named constants are reversed

PR #2273, opened 2023-07-10
Add `buf mod` commands to pre-commit hooks

ayberkozgur:
The motivation is to be able to use `buf mod update` and `buf mod prune` as a pre-commit hook upon changes in the buf module and lock files. We currently have to do it with a repo local hook with system language.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -22,9 +22,15 @@
   entry: buf format -w --exit-code
   types: [proto]
   pass_filenames: false
-- id: buf-mod
-  name: buf mod
+- id: buf-mod-update
+  name: buf mod update
   language: golang
-  entry: buf mod
+  entry: buf mod update
+  files: '(buf\.lock|buf\.yaml)'
+  pass_filenames: false
+- id: buf-mod-prune
+  name: buf mod prune
+  language: golang
+  entry: buf mod prune
   files: '(buf\.lock|buf\.yaml)'
   pass_filenames: false
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -22,3 +22,9 @@
   entry: buf format -w --exit-code
   types: [proto]
   pass_filenames: false
+- id: buf-mod
+  name: buf mod
+  language: golang
+  entry: buf mod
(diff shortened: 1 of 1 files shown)


.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,9 @@
   entry: buf format -w --exit-code
   types: [proto]
   pass_filenames: false
+- id: buf-mod
+  name: buf mod
+  language: golang
+  entry: buf mod
+  files: '(buf\.lock|buf\.yaml)'
```
bufdev, line 29: Why would this be for changes to buf.lock? I think it should just be for buf.yaml but I could be wrong.

.pre-commit-hooks.yaml
```diff
@@ -22,3 +22,9 @@
   entry: buf format -w --exit-code
   types: [proto]
   pass_filenames: false
+- id: buf-mod
+  name: buf mod
+  language: golang
+  entry: buf mod
```
bufdev, line 28: As opposed to other commands in this list, this is just a partial command - can you explain how this works? Ie all above commands are actual commands that you run, while buf mod is a container for update, etc

ayberkozgur: Indeed; I plan that it should work via setting e.g `args: [update]` or `args: [prune]`. I think this is better compared to making one pre-commit entry for each subcommand of `buf mod`.

bufdev: For consistency's sake I might argue the opposite - given the number of commands that are relevant to precommit (small), adding two commands instead of one seems appropriate.

bufdev:
Will look again when comment about splitting commands is addressed, thanks

.pre-commit-hooks.yaml
ayberkozgur: I would argue for the opposite because there are many subcommands to `buf mod` that will potentially require one entry each there and this would pollute this file. But to me as a user of this repo it makes no difference so I can make the change you want.

ayberkozgur:
> Will look again when comment about splitting commands is addressed, thanks

@bufdev Done!

bufdev (approved):

PR #2274, opened 2023-07-10
Add `ListRepositoryTagsByReference` rpc

doriable:
This adds a new RPC for `ListRepositoryTagsForReference`.
There is an argument to simply add to the old RPC, but given the way
our APIs have been structured, I think adding a new RPC to address this
need is better than simply adding to the old RPC. Even though it would still
be backwards compatible, the new RPC provides greater clarity.


unmultimedio (approved):
Not 100% sure if a new RPC is better than just adding the missing reference param to the existing RPC. We'd need to know which places are using it, if any, and in the backend probably make it optional, so if it comes empty we can keep existing behavior?

Anyway, not blocking, just a thought.

doriable:
> Not 100% sure if a new RPC is better than just adding the missing reference param to the existing RPC. We'd need to know which places are using it, if any, and in the backend probably make it optional, so if it comes empty we can keep existing behavior?

Yeah, definitely. I think after the first run of sync, I'd like to document some of the API-related things, including this, and generally, should we be using the labels API, etc.
Appreciate this being not blocking :) Am definitely (attempting) to keep a log of all the things that are along these lines.

seankimdev (commented):
Sorry about the late review. Looks good to me. Thanks Doria

PR #2278, opened 2023-07-10
Remove remote generation alpha features

mfridman:
This PR removes the remote generation alpha code.

Please refer to the migration guide on how to migrate to the new **Remote Plugins** and **Remote Packages** features.

https://buf.build/docs/migration-guides/migrate-remote-generation-alpha


proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -88,26 +75,12 @@ message CuratedPluginSearchResult {
...
-  bool deprecated = 5;
-}
-
 message SearchResult {
   oneof item {
     RepositorySearchResult repository = 1;
     OrganizationSearchResult organization = 2;
     UserSearchResult user = 3;
     TeamSearchResult team = 4;
-    PluginSearchResult plugin = 5 [deprecated = true];
```
mfridman, line 109: I reserved the event fields since they are stored. But didn't add reserved for any of the others since the BSR should be the sole consumer of these. But opinions may vary though on how much of the legacy stuff we want to break.

private/buf/bufgen/generator.go
```diff
@@ -209,34 +208,12 @@ func (g *generator) execPlugins(
...
 		indexedPluginConfigs := indexedPluginConfigs
-		v1Args := make([]*remotePluginExecArgs, 0, len(indexedPluginConfigs))
 		v2Args := make([]*remotePluginExecArgs, 0, len(indexedPluginConfigs))
 		for _, param := range indexedPluginConfigs {
-			if param.PluginConfig.Plugin == "" {
-				v1Args = append(v1Args, param)
-			} else {
-				v2Args = append(v2Args, param)
+			if param.PluginConfig.Remote != "" {
+				return nil, fmt.Errorf("invalid plugin reference: %s", param.PluginConfig.Remote)
```
mfridman, line 214: We don't bother a roundtrip to the server, and instead fail if we detect any legacy plugin that has remote.

Suggestions are welcome for a nicer error message.

Given the work @oliversun09 is doing, I think we should avoid a large refactor within this code path.

private/buf/bufgen/generator.go
```diff
@@ -209,34 +208,12 @@ func (g *generator) execPlugins(
 	for remote, indexedPluginConfigs := range remotePluginConfigTable {
 		remote := remote
 		indexedPluginConfigs := indexedPluginConfigs
-		v1Args := make([]*remotePluginExecArgs, 0, len(indexedPluginConfigs))
 		v2Args := make([]*remotePluginExecArgs, 0, len(indexedPluginConfigs))
 		for _, param := range indexedPluginConfigs {
-			if param.PluginConfig.Plugin == "" {
```
bufdev, line 215: Don't we still need to support this in the config file? Or is that logic in the external -> internal config stuff?

mfridman: The `Plugin` key is still there, but I flipped the logic to check against the `remote` key to then use `param.PluginConfig.Remote` in the error message.

We could also push this much earlier to the external -> internal validation and remove `Remote` from internal packages entirely?

https://github.com/bufbuild/buf/blob/6c98e0e88168c6e17b8c82f489ab68cd705b943e/private/buf/bufgen/config.go#L211-L217

mfridman: Added a check in https://github.com/bufbuild/buf/pull/2278/commits/0ec5f4aa325c715f0644675617f0a0c914428a3b, I think there's an opportunity to cleanup `bufgen/{config,generator}.go`, but I suggest we do this in a separate PR or fold it into the v2 work.

We don't need the internal Remote config, and instead gate it at the external config layer.

### Example 1

```yaml
plugins:
  - remote: buf.build/protocolbuffers/plugins/go
  # or
  - remote: buf.build/protocolbuffers/go
```

We surface an error unconditionally for both new and old references since the `remote` field is being used in the external config.

> Failure: File "buf.gen.yaml": the remote field is no longer supported, see the migration guide: https://buf.build/docs/migration-guides/migrate-remote-generation-alpha/#migrate-to-remote-plugins

### Example 2

```yaml
plugins:
  - plugin: buf.build/protocolbuffers/plugins/go
```

> Failure: File "buf.gen.yaml": invalid plugin reference: buf.build/protocolbuffers/plugins/go

private/buf/bufgen/generator.go
```diff
@@ -342,73 +319,6 @@ type remotePluginExecutionResult struct {
...
-	pluginReferences := make([]*registryv1alpha1.PluginReference, len(pluginConfigs))
-	for i, pluginConfig := range pluginConfigs {
-		pluginReference, err := getPluginReference(pluginConfig.PluginConfig)
-		if err != nil {
-			return nil, err
-		}
-		pluginReferences[i] = pluginReference
-	}
-	generateService := connectclient.Make(g.clientConfig, remote, registryv1alpha1connect.NewGenerateServiceClient)
-	response, err := generateService.GeneratePlugins(
```
mfridman, line 363: The `GeneratePlugins` endpoint will be removed and has been replaced by `GenerateCode`

private/buf/bufgen/config.go
```diff
@@ -205,16 +205,16 @@ func validateExternalConfigV1(externalConfig ExternalConfigV1, id string) error
...
 		case plugin.Remote != "":
-			if _, _, _, _, err := bufremoteplugin.ParsePluginVersionPath(pluginIdentifier); err != nil {
-				return fmt.Errorf("%s: invalid remote plugin name: %w", id, err)
-			}
-			if err := checkPathAndStrategyUnset(id, plugin, pluginIdentifier); err != nil {
-				return err
-			}
+			// Remote generation alpha features have been deprecated, but we continue to detect
+			// the remote field to surface a better error message.
+			return fmt.Errorf("%s: the remote field is no longer supported, see the migration guide: %s",
```
bufdev, line 214: `the remote field no longer works as the remote generation alpha has been deprecated, see the migration guide to now-stable remote plugins`

bufdev (approved):

PR #2280, opened 2023-07-10
Implement Java package in the style of java_package_prefix, java_package_suffix and java_package

oliversun9:
This implements `java_package_prefix`, `java_package_suffix` and `java_package` for managed mode v2.
`mergeOverrideFuncs` in `private/buf/bufgen/internal/bufgenv2/config.go` is worthy of attention, which implements the precedence among the three.


private/buf/bufgen/internal/bufgenv2/bufgenv2.go
```diff
@@ -191,14 +195,14 @@ type ExternalConfigV2 struct {
 
 // ExternalManagedConfigV2 is an external managed mode configuration.
 type ExternalManagedConfigV2 struct {
-	Enable   bool                              `json:"enable,omitempty" yaml:"enable,omitempty"`
+	Enabled  bool                              `json:"enabled,omitempty" yaml:"enabled,omitempty"`
 	Disable  []ExternalManagedDisableConfigV2  `json:"disable,omitempty" yaml:"disable,omitempty"`
```
oliversun9, line 199: This is to stay in line with v1, which uses `enabled`

private/buf/bufgen/internal/bufgenv2/file_option.go
```diff
@@ -19,131 +19,201 @@ import (
...
+		managedOptionJavaPackagePrefix: "java_package_prefix",
+		managedOptionJavaPackageSuffix: "java_package_suffix",
+	}
+	managedOptionToFileOption = map[managedOption]fileOption{
+		managedOptionJavaPackage:       fileOptionJavaPackage,
+		managedOptionJavaPackagePrefix: fileOptionJavaPackage,
+		managedOptionJavaPackageSuffix: fileOptionJavaPackage,
+	}
+	managedOptionToOverrideParseFunc = map[managedOption]func(interface{}, managedOption) (bufimagemodifyv2.Override, error){
+		managedOptionJavaPackage:       parseValueOverride[string],
```
oliversun9, line 122: this map declares what value is expected from each option (not file option)

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -279,18 +287,34 @@ func mergeDisabledFuncs(disabledFuncs []disabledFunc) disabledFunc {
 
 func mergeOverrideFuncs(overrideFuncs []overrideFunc) overrideFunc {
 	// Last override listed wins
```
oliversun9, line 289: this function implements the rule that:
If a file matches both a rule on `java_package` and a rule on `java_package_prefix` / `java_package_suffix`, the last one wins.
If a file the last java package rule (one of the three) matched by a file is `java_package_suffix`, we will also apply the last seen `java_package_prefix`, and vice versa.

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -22,34 +22,27 @@ import (
...
-// TODO: likely want something like *string or otherwise, see https://github.com/bufbuild/buf/issues/1949
 // overrideFunc is specific to a file option, and returns what thie file option
 // should be overridden to for this file.
-type overrideFunc func(bufimage.ImageFile) (string, error)
+type overrideFunc func(imageFileIdentity) bufimagemodifyv2.Override
+
+// imageFileIdentity is an image file that can be identified by a path and module identity.
+// There two (path and module) are the only information needed to decide whether to disable
+// or override a file option for a specific file. Using an interface to for easier testing.
+type imageFileIdentity interface {
```
oliversun9, line 42: link to [discussion on a previous PR](https://github.com/bufbuild/buf/pull/2226#discussion_r1244148983)

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,154 @@
...
+	}
+	var javaPackageValue string
+	switch t := options.override.(type) {
+	case prefixOverride:
+		javaPackageValue = getJavaPackageValue(imageFile, t.get(), internal.DefaultJavaPackageSuffix)
+	case suffixOverride:
+		return errors.New("cannot modify java_package with a suffix but without a prefix")
+	case prefixSuffixOverride:
+		javaPackageValue = getJavaPackageValue(imageFile, t.prefix, t.suffix)
+	case valueOverride[string]:
```
oliversun9, line 134: > A value for a straight java_package doesn't make sense - you're saying "for all code generated by this buf.gen.yaml, regardless of the package of a given .proto file, use exactly this java package value". Same for go_package, and others likely.

I think it's possible that for smaller projects a user might want to put everything generated in one Java package (for example, while they are learning to use Protobuf), or they want files generated from a directory of proto files to be put in the same Java package. We could check for this in `bufgenv2` package, but I don't think it's necessary as long as we recommend using`java_package_prefix` over the other 2.

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,154 @@
...
+		javaPackageValue = t.get()
+	case nil:
+		javaPackageValue = getJavaPackageValue(imageFile, internal.DefaultJavaPackagePrefix, internal.DefaultJavaPackageSuffix)
```
oliversun9, line 137: > The nil case shouldn't be valid. All overrides should have a value or prefix - if you just want the default value for managed mode, you don't set an override. Same comment whole way down

Now `ModifyXYZ` has option parameters that specify an override. Now `nil` means user of this package did not pass an option of override.

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -140,11 +142,16 @@ func newManagedConfig(externalConfig ExternalManagedConfigV2) (*ManagedConfig, e
...
 	for _, externalOverrideConfig := range externalConfig.Override {
-		fileOption, err := ParseFileOption(externalOverrideConfig.FileOption)
+		managedOption, err := parseManagedOption(externalOverrideConfig.Option)
 		if err != nil {
 			// This should never happen because we already validated
 			return nil, err
 		}
+		fileOption, ok := managedOptionToFileOption[managedOption]
+		if !ok {
+			// this should not happen
```
bufdev, line 152: This should not happen yet, right?

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -171,103 +179,105 @@ func validateExternalManagedConfigV2(externalConfig ExternalManagedConfigV2) err
...
+		if externalOverrideConfig.Value == nil {
+			return errors.New("must specify an value to override")
+		}
 	}
 	return nil
 }
 
 func newDisabledFunc(externalConfig ExternalManagedDisableConfigV2) (disabledFunc, error) {
-	var selectorFileOption FileOption
+	if len(externalConfig.FileOption) == 0 && len(externalConfig.Module) == 0 && len(externalConfig.Path) == 0 {
```
bufdev, line 197: Everything should be `option` now, not `file_option`, unless I'm missing something...why would override and disable differ?

private/buf/bufgen/internal/bufgenv2/file_option.go
```diff
@@ -19,131 +19,203 @@ import (
...
-	// FileOptionCsharpNamespace is the file option csharp_namespace.
-	FileOptionCsharpNamespace
-	// FileOptionPhpNamespace is the file option php_namespace.
-	FileOptionPhpNamespace
-	// FileOptionPhpMetadataNamespace is the file option php_metadata_namespace.
-	FileOptionPhpMetadataNamespace
-	// FileOptionRubyPackage is the file option ruby_package.
-	FileOptionRubyPackage
+	// fileOptionJavaPackage is the file option java_package.
+	fileOptionJavaPackage fileOption = iota + 1
```
bufdev, line 29: why does this still exist?

private/buf/bufgen/internal/bufgenv2/file_option.go
```diff
@@ -19,131 +19,203 @@ import (
...
+	// fileOptionCsharpNamespace is the file option csharp_namespace.
+	fileOptionCsharpNamespace
+	// fileOptionPhpNamespace is the file option php_namespace.
+	fileOptionPhpNamespace
+	// fileOptionPhpMetadataNamespace is the file option php_metadata_namespace.
+	fileOptionPhpMetadataNamespace
+	// fileOptionRubyPackage is the file option ruby_package.
+	fileOptionRubyPackage
+	// managedOptionJavaPackage is the managed mode option java_package.
+	managedOptionJavaPackage managedOption = iota + 1
```
bufdev, line 53: I don't understand the distinction between file option and managed option, I'd expect we only have one type

private/buf/bufgen/internal/bufgenv2/config.go
oliversun9: I was thinking in disable a user can only disable options from `descriptor.proto`, like `java_package`,
and in override a user can override options like `java_package_prefix` and etc.

private/buf/bufgen/internal/bufgenv2/file_option.go
oliversun9: This is to keep allowing per-file overrides from v1

private/buf/bufgen/internal/bufgenv2/config.go
bufdev: This is the opposite of the direction we want to go in - this goes back to v1, where users have to put extra thought into what options apply where, whereas the goal of v2 is that everything just works, and you don't have to think about disable vs override as an example

private/buf/bufgen/internal/bufgenv2/file_option.go
bufdev: Why is v1 affecting v2? I'm not sure I understand

private/buf/bufgen/internal/bufgenv2/config.go
bufdev: I do not want to have to read documentation to know which options apply where

private/buf/bufgen/internal/bufgenv2/file_option.go
oliversun9: This option will be removed, as per comment above

private/buf/bufgen/internal/bufgenv2/file_option.go
oliversun9: There is also another reason:
Managed mode needs to know the relative order of rules with different file options:
```
override: # values omitted
  - file_option: java_package
  - file_option: java_package_prefix
  - file_option: java_package_suffix
 ```
has a different effect (per the spec) than
```
override:
  - file_option: java_package_prefix
  - file_option: java_package
  - file_option: java_package_suffix
 ```
The way I implemented this is to group file options together by the proto file option they modify (e.g. `j_p_prefix`, `j_p_suffix` and `j_p` belong to the same group). The grouping is achieved by the map `managedOptionToFileOption = map[managedOption]fileOption{...`, which will be renamed.
After [grouping them together by using this map](https://github.com/bufbuild/buf/pull/2280/files#diff-6ad607dc37b36d1f3e43504fb3b079819f8d2cca6cb92f7b16c2714ef2b64e86R150), I can merge the overrides of each group, like so in [mergeOverrideFuncs](https://github.com/bufbuild/buf/pull/2280/files#diff-6ad607dc37b36d1f3e43504fb3b079819f8d2cca6cb92f7b16c2714ef2b64e86R290).
Do you think this approach will reasonable after renaming `managedOption`? (another approach is to give each rule a number, which is also straighforward)

private/bufpkg/bufimage/bufimagemodify/bufimagemodifyv2/bufimagemodifyv2.go
```diff
@@ -15,49 +15,165 @@
...
 		descriptor.Options = &descriptorpb.FileOptions{}
 	}
-	descriptor.Options.JavaPackage = proto.String(getJavaPackageValue(imageFile, prefix))
+	descriptor.Options.JavaPackage = proto.String(javaPackageValue)
 	marker.Mark(imageFile, internal.JavaPackagePath)
 	return nil
 }
 
-func getJavaPackageValue(imageFile bufimage.ImageFile, prefix string) string {
+func getJavaPackageValue(imageFile bufimage.ImageFile, prefix string, suffix string) string {
```
oliversun9, line 168: Maybe this can be moved to `bufgenv2` package? wdyt?
If this move happens, package `bufimagemodifyv2` won't need to know anything about prefix and suffix. `ModifyXYZ` will be take a value and modify the file option on descriptor set directly to this value. However, `bufgenv2` would need to do stuff like converting a image file's proto package to Pascal cas.

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -120,154 +110,162 @@ func readConfigV2(
...
 	}, nil
 }
 
-func validateExternalConfigV2(externalConfig ExternalConfigV2, id string) error {
-	// TODO: implement this
-	return nil
-}
-
-func validateExternalManagedConfigV2(externalConfig ExternalManagedConfigV2) error {
-	if externalConfig.isEmpty() {
```
oliversun9, line 169: Now validation happens when the Config is being constructed / populated.

private/buf/bufgen/internal/bufgenv2/testdata/managed/match.yaml
```diff
@@ -0,0 +1,46 @@
...
+      module: buf.build/acme/weather
+    - file_option: java_package
+      path:  a/b/c.proto
+    - file_option: java_package
+      module: buf.build/acme/petapis
+      path: x/y/z.proto
+    - file_option: java_package
+      path: exact.proto
+  override:
+    - option: java_package
```
bufdev, line 21: Why is this `option` in `override` but `file_option` in disable? This should be the same thing everywhere. I thought we agreed on `file_option` in the latest iteration, but instructions remain the same: I should not have to think about what is in override vs disable, I shouldn't have to read the docs on this

oliversun9: sorry I forgot to change the field tag

bufdev (approved):
This mostly looks good to me. Actual functionality we will test before merging `bufgenv2` into `main`.

oliversun9:
> This mostly looks good to me. Actual functionality we will test before merging `bufgenv2` into `main`.

Sounds good!

PR #2282, opened 2023-07-11
Fix path handling in storagegit

bufdev:
- All paths should have `storageutil.Validate{Path,Prefix}` called on them at the entrypoint.
- `ExternalPath` being used potentially incorrectly here, it's a gray area, changing to `Path` for now.
- A few other normalization fixes.

### DIFF
--- a/private/pkg/storage/storagegit/bucket.go
+++ b/private/pkg/storage/storagegit/bucket.go
@@ -46,6 +46,10 @@ func newBucket(
 }
 
 func (b *bucket) Get(ctx context.Context, path string) (storage.ReadObjectCloser, error) {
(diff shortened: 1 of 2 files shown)


private/pkg/storage/storagegit/bucket.go
```diff
@@ -180,10 +187,11 @@ func (b *bucket) walkTree(
 	return nil
 }
 
+// path is expected to be normalized by calling cunrions
```
pkwarren, line 190: Typo in comment.

pkwarren (approved):

lrewega (approved):

PR #2283, opened 2023-07-11
Add new branch RPCs and change existing RPCs

doriable:
Adds a new RPC for branches, `GetDefaultBranch`, which returns the current default
branch mapped to BSR main.

Adds three new fields to `RepositoryCommit`:
- `tag_count` -- number of tags associated with the BSR commit
- `git_commit_metadata_count` -- number of git commits with metadata associated with the BSR commit
- `git_commit_no_metadata_count` -- number of git commit labels without metadata associated with the BSR commit 

Adds pagination of `ListGitMetadataForReference`.

Adds a new RPC for reference, `ListGitCommitsNoMetadataForReference`, which returns
a paginated list of git commit labels that do not have git metadata.


proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -37,6 +37,14 @@ service RepositoryBranchService {
   rpc ListRepositoryBranches(ListRepositoryBranchesRequest) returns (ListRepositoryBranchesResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // GetDefaultBranch returns the branch name that is mapped to the main/BSR_HEAD.
+  rpc GetDefaultBranch(GetDefaultBranchRequest) returns (GetDefaultBranchResponse) {
```
unmultimedio, line 41: This RPC name conflicts with L28-29 in which we say it's not "default" but "main" :/

I like "default" better though.

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -52,3 +60,26 @@ message ListRepositoryBranchesResponse {
   // There are no more pages if this is empty.
   string next_page_token = 2;
 }
+
+message GetDefaultBranchRequest {
+  // The ID of the repository whose branches should be listed.
+  string repository_id = 1;
```
unmultimedio, line 66:
```suggestion
  // The ID of the repository whose default branch should be listed.
  string repository_id = 1;
```

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -52,3 +60,26 @@ message ListRepositoryBranchesResponse {
...
+}
+
+message GetDefaultBranchResponse {
+  RepositoryBranch default_branch = 1;
+}
+
+message ListRepositoryNonDefaultBranchesRequest {
+  // The ID of the repository whose branches should be listed.
+  string repository_id = 1;
```
unmultimedio, line 75:
```suggestion
  // The ID of the repository whose non-default branches should be listed.
  string repository_id = 1;
```

unmultimedio (commented):
More to the why of these RPC, why not adding a filter to the existing `rpc ListRepositoryBranches`, like `bool default_branch` to "list" only the default one. And when listing the non-default, it could be filtered out by the client using the `is_main_branch` from L29.

seankimdev (approved):
Looks good to me

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
doriable: Ah that's true :< Hmm... yeah, I want to use the term default... tbh, I wish we could change the other RPC structure, lol... but it might be late for that. I think `Default` is a lot more explicit and clear for this use case though.

doriable: Had an offline discussion, we also decided to go with `Current`, since this makes it clear it's the current/latest default branch, and does not reflect the history of default branches if it's ever been changed.

seankimdev: Should the `is_main_branch` be removed? I think it may no longer be necessary with the addition of this new rpc. It also feels weird to have a field dedicated to indicate true only for one branch that is default(main).

seankimdev: > And when listing the non-default, it could be filtered out by the client using the is_main_branch from L29.
ah sorry I missed this part Julian wrote. I guess it would be necessary if we decide to go in that direction. Let's chat in the meeting.

proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -86,11 +92,37 @@ message ListGitCommitMetadataForReferenceRequest {
...
+  uint32 page_size = 4;
+  // The first page is returned if this is empty.
+  string page_token = 5;
+}
+
+message ListGitCommitsNoMetadataForReferenceResponse {
+  // ID of the BSR commit the reference resolved to.
+  string commit_id = 1;
+  // List of git commits labels without metadata associated with the resolved reference.
+  repeated Label git_commit_labels_no_metadatas = 2;
```
unmultimedio, line 125: Nit: we already know from the parent message that these labels come from a commit that doesn't have git metadata on it.
```suggestion
  // List of git commits labels without metadata associated with the resolved reference.
  repeated Label git_commit_labels = 2;
```

proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -64,6 +65,11 @@ service ReferenceService {
   rpc ListGitCommitMetadataForReference(ListGitCommitMetadataForReferenceRequest) returns (ListGitCommitMetadataForReferenceResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // ListGitCommitsNoMetadataForReference takes a string reference and returns all the git commit
+  // labels without metadata associated with the resolved reference commit.
+  rpc ListGitCommitsNoMetadataForReference(ListGitCommitsNoMetadataForReferenceRequest) returns (ListGitCommitsNoMetadataForReferenceResponse) {
```
unmultimedio, line 70: I assume this RPC would be invoked when `ListGitCommitMetadataForReference` returns empty for a given reference?

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -49,6 +49,12 @@ message RepositoryCommit {
   string spdx_license_id = 10;
   // The manifest digest of the commit.
   string manifest_digest = 11;
+  // Number of tags associated with the commit.
+  int64 tag_count = 12;
+  // Number of git commits with metadata associated with the BSR commit.
+  int64 git_commit_metadata_count = 13;
+  // Number of git commits without metadata associated with the BSR commit.
+  int64 git_commit_no_metadata_count = 14;
```
unmultimedio, line 57: I'm confused with this one, so if a commit has 2 git commits associated with it, it still has 2 git commit labels. Unless I'm missing something, I think `git_commit_no_metadata_count` is not necessary.

proto/buf/alpha/registry/v1alpha1/reference.proto
doriable: Yeah, so in the world where a user has `buf push`'d a BSR commit, that BSR commit reference will not have any git metadata, but may still have meaningful `GIT_COMMIT_SHA` labels that need to be displayed, so this RPC would be invoked for that circumstance.

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
doriable: I guess it's for the circumstance where the BSR commit has no git commit metadata associated with it, but still has git commit labels on it -- we would want to count those.

unmultimedio: Not sure how this is going to play out. For a BSR commit that was previously push, with, say 3 tags:

```
buf push --tag foo --tag bar --tag someGitCommitSHA1
```

and then sync runs, and the content is the same, attaching two git metadatas to the same BSR commit:

```
someGitCommitSHA2
someGitCommitSHA3
```

I guess the commit object would have:

```
{
  tag_count: 5, // foo, bar, and the 3 git shas
  git_commit_metadata_count: 2, // someGitCommitSHA2 and someGitCommitSHA3
  git_commit_no_metadata_count: 1 // someGitCommitSHA1
}
```

Am I right?

unmultimedio (approved):

doriable:
@unmultimedio Yes, this is exactly correct.

PR #2287, opened 2023-07-12
Reject buf mod updates that will result in multiple dependencies having equal file paths

oliversun9:
`buf mod update` fails when a file path exists in more than one modules (either from this module or from a dependency module)
```
$ ./dev mod update
./dev mod update
Failure: x.proto is found in multiple modules: buf.build/osun/a1, buf.build/osun/b1, and buf.build/osun/c1
Multiple files with the same path are not allowed because Protobuf files import each other by their paths, and each import path must uniquely identify a file.
$ ./dev mod update
Failure: x.proto is found in multiple modules: the current module, buf.build/osun/b1, and buf.build/osun/c1
Multiple files with the same path are not allowed because Protobuf files import each other by their paths, and each import path must uniquely identify a file.
```
In a follow-up PR, `buf mod update` will also include modules in the same workspace when checking for overlap.


private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -149,6 +154,43 @@ func run(
 		}
 		return bufcli.NewInternalError(err)
 	}
+	// Before updating buf.lock file, verify that no file path exists in more than one module.
+	moduleReader, err := bufcli.NewModuleReaderAndCreateCacheDirs(container, clientConfig)
+	if err != nil {
+		return bufcli.NewInternalError(err)
+	}
+	filePathToOwnerModule := make(map[string]string)
```
bufdev, line 162: I would instead make this  `pathToModuleIdentityStrings := map[string][]string`, and when iterating, store all the modules that a file path is in. At the end of this iteration over dependencies and the current module, `for path, moduleIdentityStrings := range pathToModuleIdentityStrings { if len(moduleIdentityStrings) > 1 { return an error with all modules included in the error string } }`. If the current module is nameless, use "the current module" as the identity string.

bufdev (approved):

PR #2292, opened 2023-07-13
Add --indent to `buf convert`

Alfus:


private/buf/bufwire/proto_encoding_writer.go
```diff
@@ -64,7 +65,13 @@ func (p *protoEncodingWriter) PutMessage(
 	case bufconvert.MessageEncodingBinpb:
 		marshaler = protoencoding.NewWireMarshaler()
 	case bufconvert.MessageEncodingJSON:
-		marshaler = protoencoding.NewJSONMarshaler(resolver)
+		var options []protoencoding.JSONMarshalerOption
+		if textOptions != nil {
+			if textOptions.Indent {
+				options = append(options, protoencoding.JSONMarshalerWithIndent())
```
bufdev, line 71: Honestly I think we should just make indent the new default

PR #2294, opened 2023-07-13
Specify 'binpb' instead of 'bin' in the convert cmd help

Alfus:
By fixing the constants.

Prev help output:
image
New output:
image


bufdev (approved):
Whoops.

PR #2309, opened 2023-07-20
Add system + local option to http client TLS

rubensf:
We have options to use http clients both for system certs or with user defined certs - but there are cases where we may want both.

### DIFF
--- a/private/pkg/cert/certclient/util.go
+++ b/private/pkg/cert/certclient/util.go
@@ -22,23 +22,51 @@ import (
 	"os"
 )
 
-// NewClientTLSConfigFromRootCertFiles creates a new tls.Config from a root certificate files.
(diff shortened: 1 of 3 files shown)


private/pkg/cert/certclient/util.go
```diff
@@ -18,11 +18,12 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"errors"
+	"fmt"
 	"os"
 )
 
 // NewClientTLSConfigFromRootCertFiles creates a new tls.Config from a root certificate files.
-func NewClientTLSConfigFromRootCertFiles(rootCertFilePaths ...string) (*tls.Config, error) {
+func NewClientTLSConfigFromRootCertFiles(alsoUseSystem bool, rootCertFilePaths ...string) (*tls.Config, error) {
```
bufdev, line 26: Can you make an option `WithSystemCertPool()`

rubensf: done, ptal

bufdev (approved):

PR #2313, opened 2023-07-21
GetLabelsInNamespace RPC

unmultimedio:
Cannot use existing `GetLabels` RPC because of the way the label messages data structure is built. Passing a name means in reality passing both a name+namespace, which I don't need in this scenario. The usecase is: Give me labels in namespace "foo", with an optional filter label names []string.


proto/buf/alpha/registry/v1alpha1/labels.proto
```diff
@@ -73,10 +78,26 @@ message MoveLabelResponse {}
 message GetLabelsRequest {
   string repository_owner = 1;
   string repository_name = 2;
+  // Optional label_name filter, in case you want to get labels in a namespace+name.
```
seankimdev, line 81: Where does the `namespace` come from for this request?

unmultimedio: Good question, it was also confusing for me at first. `message LabelName` [contains both namespace and name](https://github.com/bufbuild/buf/blob/cb7982633f38fa222e1141fcd9b04d004faaf1e9/proto/buf/alpha/registry/v1alpha1/labels.proto#L35-L38).

seankimdev: Ohh... I see

seankimdev (approved):
Lgtm, but I'd wait for @doriable to double check the rpc's use-case.

PR #2314, opened 2023-07-21
Remove parent path (for a FieldOptions) when there are no children field option paths left

oliversun9:
This is a follow-up to https://github.com/bufbuild/buf/pull/2311#discussion_r1270914076

For a field 
```
repeated int64 i8 = 127 [
  jstype = JS_STRING,
  retention = RETENTION_RUNTIME,
  debug_redact = true,
  lazy = true
];
```
If the source code info paths for all 4 options are removed, the path for the FieldOptions (`{..., 8}`) should also be removed, where [8 is the tag for field options](https://buf.build/protocolbuffers/wellknowntypes/file/main:google/protobuf/descriptor.proto#L261).


private/bufpkg/bufimage/bufimagemodify/internal/internal.go
```diff
@@ -83,7 +83,25 @@ func RemoveLocationsFromSourceCodeInfo(sourceCodeInfo *descriptorpb.SourceCodeIn
...
+	potentialFieldOptionToChildrenIndices := make(map[string][]int, len(paths))
+	// If multiple paths are the same, use the last one. This is fine because
+	// this is only useful for field options, where each FieldOptions' path does
+	// not appear twice.
+	keyToIndex := make(map[string]int, len(paths))
 	for i, location := range sourceCodeInfo.Location {
+		path := location.Path
+		keyToIndex[GetPathKey(path)] = i
+		if isPathMaybeForFieldOption(path) {
+			parentKey := GetPathKey(path[:len(path)-1])
```
jhump, line 95: If the source file used compound option names or repeated options, the length of the path could be greater than the path to options +1.

For example:
```
(custom.option.name).subfield_name.another_subfield_name = "abc"
```
In the above, this path will have three extra components beyond the path to the options message: one for the the `custom.option.name` tag, another for `subfield_name`, and a final one for `another_subfield_name`.

Similarly:
```
(custom.repeated.field) = 123
```
In the above, the path will have two elements beyond the options message: the tag for `custom.repeated.field` plus an index.

I think you may want to invert how you are identifying these: identify the "roots" as all paths to field options messages and store into a prefix tree instead of a map. With a prefix tree, you can then easily test subsequent paths to see if they are a sub-path of a root. When you find such a sub-path for an element that is _not_ being deleted, increment a count for that root. At the end, all roots in the prefix tree with a count of zero should be removed as well since they have no retained descendants.

BTW, I guess `isPathMaybeForFieldOption` already exists as I don't see the source in this PR. But this can be made definitive (instead of "maybe") as it's a fairly simple DFA to match any option. Similarly, you can definitively match a field options _message_ (the root path for any field option). The DFA for that looks like so, which should be straight-forward to encode as a static structure:
```mermaid
flowchart TD;
  Start -- "7 (tag extension)" --> fields;
  fields --"* (index)" --> field;
  Start -- "4 (tag message_type)" --> messages;
  messages -- "* (index)" --> message;
  message -- "3 (tag nested_type)" --> messages;
  message -- "2 (tag field)" --> fields;
  message -- "6 (tag extension)" --> fields;
  field -- "8 (tag options)" --> Accept;
```

private/bufpkg/bufimage/bufimagemodify/internal/internal.go
```diff
@@ -75,6 +72,7 @@ var fileOptionPath = []int32{8}
 
 // RemoveLocationsFromSourceCodeInfo removes paths from the given sourceCodeInfo.
 // Each path must be for either a file option or a field option.
+// TODO: in v1 there is no need to check for field options, maybe v1 and v2 don't need to share this function.
```
jhump, line 75: Let's move this into the function so it doesn't show up in Go docs.

private/bufpkg/bufimage/bufimagemodify/internal/file_options_trie_test.go
```diff
@@ -0,0 +1,342 @@
...
+				{20},            // 1
+				{20, 10},        // 2
+				{20, 15},        // 3
+			},
+			expectedIndices: []int{
+				1,
+				2,
+				3,
+				0,
+			},
```
jhump, line 61: Is the actual order of returned indices significant? Seems like we should just verify that expected indices equals [0..len(pathsToInsert)), so we could sort the return before comparing.

private/bufpkg/bufimage/bufimagemodify/internal/file_options_trie_test.go
```diff
@@ -0,0 +1,342 @@
...
+				{4, 0, 2, 2, 8, 5},
+			},
+			expectedIndicesWithoutDescendant: []int{
+				0,
+				3,
+				6,
+			},
+		},
+		{
+			description: "register more none-field option",
```
jhump, line 295: Might be good to have a case where expected indices are empty (i.e. all option paths retain at least one descendant)?

private/bufpkg/bufimage/bufimagemodify/internal/internal.go
```diff
@@ -83,8 +81,18 @@ func RemoveLocationsFromSourceCodeInfo(sourceCodeInfo *descriptorpb.SourceCodeIn
 	// are structured as expected, and collect all of the indices that
 	// we need to delete.
 	indices := make(map[int]struct{}, len(paths)*2)
+	// each path in this trie is for a FieldOptions message (not for a singular option)
+	fieldOptionsPaths := fieldOptionsTrie{}
```
jhump, line 85: I think it's more idiomatic to do `var fieldOptionsPaths fieldOptionsTrie` since the zero-value (nil slice) is fine here.

That's how I interpret [this part of the style guide](https://www.notion.so/bufbuild/Go-Code-Standards-ff40382e6b104c1097a017f9565d54db?pvs=4#36fdf7fc29054520b1ccdd35cfd38a01) anyways, which says to use `:=` for non-zero values and `var` for zero values.

private/bufpkg/bufimage/bufimagemodify/internal/internal.go
```diff
@@ -83,8 +81,18 @@ func RemoveLocationsFromSourceCodeInfo(sourceCodeInfo *descriptorpb.SourceCodeIn
...
 	for i, location := range sourceCodeInfo.Location {
-		if _, ok := paths[GetPathKey(location.Path)]; !ok {
+		path := location.Path
+		if isPathForFieldOptions(path) {
+			fieldOptionsPaths.insert(path, i)
+		}
+		if _, ok := paths[GetPathKey(path)]; !ok {
+			if isPathForSomeFieldOption(path) {
```
jhump: `isPathForFieldOption` could return a second bool to indicate whether it's a path to the options message directly or to a descendant (or maybe return a new enum type instead of two bools), so that you don't have to run the matcher 2x.

private/bufpkg/bufimage/bufimagemodify/internal/internal.go
```diff
@@ -83,8 +81,18 @@ func RemoveLocationsFromSourceCodeInfo(sourceCodeInfo *descriptorpb.SourceCodeIn
...
+				// This field option path is not marked, register it to its parent FieldOptions.
```
jhump, line 94: I guess "not marked" is what is meant by `!ok` in the condition? Does that mean that `paths` contains the items being _removed_? (I believe we should only register them if they are _not_ being removed.)

If so... that's kinda confusing. Usually a mark+sweep algorithm marks the live set, and then you sweep away everything unmarked. But is this instead marking the things to remove? (Please forgive my ignorance of what this package is or how it works.)

If `paths` are the items being removed, maybe call it `removedPaths` for clarity? And, if not, maybe `markedPaths`?

private/bufpkg/bufimage/bufimagemodify/internal/location_path_dfa.go
```diff
@@ -0,0 +1,112 @@
...
+	// https://github.com/protocolbuffers/protobuf/blob/29152fbc064921ca982d64a3a9eae1daa8f979bb/src/google/protobuf/descriptor.proto#L98
+	extensionTagInMessage = 6
+	// https://github.com/protocolbuffers/protobuf/blob/29152fbc064921ca982d64a3a9eae1daa8f979bb/src/google/protobuf/descriptor.proto#L215
+	optionsTagInField          = 8
+	pathTypeInvalid   pathType = iota + 1
+	pathTypeEmpty
+	pathTypeMessages
+	pathTypeMessage
+	pathTypeFields
+	pathTypeField
```
jhump: I don't think you actually need all of these. I think you can replace them all with `pathTypeNotFieldOption` since you really only care about the type being one of the last two items or not.

private/bufpkg/bufimage/bufimagemodify/internal/location_path_dfa.go
```diff
@@ -0,0 +1,112 @@
...
+
+type pathType int
+
+func isPathForFieldOptions(path []int32) bool {
+	return getPathType(path) == pathTypeFieldOptions
+}
+
+func isPathForSomeFieldOption(path []int32) bool {
+	return getPathType(path) == pathTypeSomeFieldOption
+}
```
jhump: Related to previous comment: I don't think you actually need these helpers at all. The place that calls them can instead call `getPathType` and then do its own comparisons of the type to `pathTypeFieldOptions` or `pathTypeSomeFieldOption` so it doesn't need to run the automata more than once.

private/bufpkg/bufimage/bufimagemodify/internal/location_path_dfa.go
```diff
@@ -0,0 +1,112 @@
...
+			return pathTypeInvalid
+		}
+		currentState, pathType = currentState(index)
+	}
+	return pathType
+}
+
+// locationPathDFAState takes an input and returns the next state and the path type
+// that ends with the input. It returns nil and pathTypeInvalid if the input is rejected.
+type locationPathDFAState func(int32) (locationPathDFAState, pathType)
```
jhump: Slick!

private/bufpkg/bufimage/bufimagemodify/internal/internal.go
oliversun9: That was my mistake, the helper doesn't need to know the concept of marking (the caller is a `MarkSweeper`). I updated it to `pathsToRemove` for clarity.

For context, in `buf generate`, if managed mode is enabled, it will call some number of `ModifyXYZ` functions, each `XYZ` being a file option or field option (we have a defined set of what `XYZ` can be).

For example, the branch that this branch is merging into, `mm-jstype`, implements `ModifyJSType`, which [modifies the jstype option on the compiled descriptor](https://github.com/bufbuild/buf/pull/2311/files#diff-0d68cfe9b7b680476f1736209b375fe7876766a166419efbfd591632d2a700c4R97), and then lets a `MarkSweeper`  mark the path to this field option. Once all such options are modified on the compiled descriptor set, managed mode let `MarkSweeper` sweep based on these marked paths, removing these paths from source code info locations. That is what this PR is about, expanding the capability of the sweeper to remove the paths to `FieldOptions` when all children field options are removed.

Let me know if you'd like to know more!

private/bufpkg/bufimage/bufimagemodify/internal/location_path_dfa.go
```diff
@@ -15,98 +15,92 @@
...
 func getPathType(path []int32) pathType {
-	pathType := pathTypeEmpty
+	pathType := pathTypeNotFieldOption
 	currentState := start
-	for _, index := range path {
+	for _, element := range path {
 		if currentState == nil {
-			return pathTypeInvalid
+			return pathTypeNotFieldOption
 		}
```
jhump: nit: you could early return (and short-circuit no-op transitions in the DFA) if `pathType == pathTypeFieldOption`.

jhump (approved):
A few comments, but otherwise LGTM!

private/bufpkg/bufimage/bufimagemodify/internal/location_path_dfa.go
```diff
@@ -0,0 +1,101 @@
...
+type locationPathDFAState func(int32) (locationPathDFAState, pathType)
+
+// getPathType returns the type of the path. It only accepts one of:
+// empty, messages, message, fields, field, field options and field option.
+func getPathType(path []int32) pathType {
+	pathType := pathTypeNotFieldOption
+	currentState := start
+	for _, element := range path {
+		if currentState == nil {
+			break
```
oliversun9, line 50: Now, if the DFA finishes running for either reason (bad path or path being a field option), we return the path type returned by the DFA.

> nit: you could early return (and short-circuit no-op transitions in the DFA) if pathType == pathTypeFieldOption

Now it does return early, without checking `pathType`, however.

Is this reasonable?

jhump: Yep, looks good!

PR #2320, opened 2023-07-25
Improve sync commits order

unmultimedio:
Improve the order in which we visit and sync git commits:

It used to be:
- Scan the whole repo, assign a single branch to each commit
- Sort all commits by author date
- Loop over sorted commits, skip until a sync point was found
- Once the sync point was found, start syncing

Now it is:
- Loop over all branches, starting with the default branch, then the rest A-Z
- For each branch:
  - Start in HEAD, travel backwards, until we find a git commit that is already synced per module
  - Keep track of modules to sync per commit
- Loop over "syncable commits", which may hold +1 modules on each
- On each syncable commit, loop over modules in the original order, sync them

Other changes:
- Removed `SyncPointNotEncountered` error handler. Now in the syncer logic we always fail if this happens in the default branch, or warn and continue if this happens in any other branch, as this could be a common scenario in PRs. If customers want to customize this behavior, the custom handler could be added again.
- Added `SyncerWithGitCommitChecker` option to the syncer. This checker will ask the BSR if certain git SHAs are already synced for a given module, so the syncer can calculate where to start syncing from.
- `git.ForEachCommit` no longer travel all commits in the time direction parent>children, but the other way, in topological direction children>parent. This way is easier to scan each branch and stop navigating the branch when we find a synced commit.
- Added `git.HEADCommit` to get the commit at HEAD for a given branch.


unmultimedio:
I'm missing tests, will add them next. But already tested locally, it's working as intended.

unmultimedio:
> I'm missing tests, will add them next. But already tested locally, it's working as intended.

Actually, in order to test this we will only require a git tree in certain forms, and hardcoding module identities. Since that soon will change (module identities will be read from git content), probably the tests will be mostly rewritten soon. How do you feel merging as is, and testing further after we do that module identity change?

unmultimedio:
> I'm missing tests, will add them next. But already tested locally, it's working as intended.

I just added a basic test to check the syncable commits finder stops when they find an already synced commit.

doriable (approved):
This looks good to me! And yeah, I think we're good to merge to make progress, and then we'll do integration testing.

PR #2323, opened 2023-07-26
Remove commit_sequence_id from RepositoryCommit

pkwarren:


proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -32,10 +32,8 @@ message RepositoryCommit {
   string name = 4;
   // The branch on which this commit was created.
   string branch = 5;
-  // The commit sequence ID for this commit. This
-  // is essentially what number commit this is on
-  // the branch.
-  int64 commit_sequence_id = 6;
+  reserved 6;
+  reserved "commit_sequence_id";
```
mfridman, line 36: This was exclusively used in the remote alpha gen template versions, we sunset the old registries and have stopped using sequence id's in the versioning schemas, so this is safe to remove.

mfridman (approved):

PR #2324, opened 2023-07-26
Add search content RPC

cyinma:


proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -103,6 +139,9 @@ service SearchService {
   rpc SearchDraft(SearchDraftRequest) returns (SearchDraftResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  rpc SearchContent(SearchContentRequest) returns (SearchContentResponse) {
```
mfridman, line 142: When we say "search content", it means search content within a module correct?

Maybe a more descriptive name might help (like `SearchModuleContent`), esp. if we continue expanding search then content could mean a lot of things.

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -89,6 +90,41 @@ message SearchResult {
   }
 }
 
+enum SearchContentFilter {
+  SEARCH_CONTENT_FILTER_UNSPECIFIED = 0;
+  SEARCH_CONTENT_FILTER_NAME = 1;
```
mfridman, line 95: What is name in this case? The name of the field, the name of the package, etc. ?

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -166,3 +205,21 @@ message SearchDraftResponse {
   // There are no more pages if this is empty.
   string next_page_token = 2;
 }
+
+message SearchContentRequest {
```
mfridman, line 209: Should this explicitly take the repo owner + name (or id), so we can filter down to content within the scope of that module?

cyinma: I thought about this (@pgmitche suggested this when we were discussing), and decided to add this later when we flushed out in our preferred UX and confirmed that there is a place for it, but this is definitely a reasonable ask

proto/buf/alpha/registry/v1alpha1/search.proto
cyinma: oops I renamed these things and forgot to also rename the filters, updated and added some comments

proto/buf/alpha/registry/v1alpha1/search.proto
cyinma: good call, updated

proto/buf/alpha/registry/v1alpha1/search.proto
mfridman: I guess it depends where this search feature is going to live within the UI. I assume it'll be scoped to a given repository, but could be wrong.

For searching content I imagine there are at least 3 levels (if not 4):

- global (entire server)
- organization
- repository
- _maybe_ even per package

If this is indeed scoped to any/all of the above, we should strongly consider starting with accepting an input and filter down as much as possible to only the scope we need.

cyinma: discussed offline, adding these in the request

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -89,6 +90,43 @@ message SearchResult {
...
+  string repository_id = 1;
+  string repository_owner = 2;
+  string repository_name = 3;
+  string proto_file_path = 4;
+  string fully_qualified_name = 5;
+  oneof document {
+    Service service = 6;
+    Method method = 7;
+    Enum enum = 8;
+    Message message = 9;
```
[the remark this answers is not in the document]
cyinma: good call out, I was thinking if that is searchable and if there is a good way to handle it, but I think its also good to include it, updated

mfridman (approved):

PR #2325, opened 2023-07-26
Change git commit information for reference RPC

doriable:
We are reworking the git commit information for reference RPC to be a
single, paginated RPC for a given reference.
Author and commiter information will be populated depending on if the data
is available.


proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -100,32 +94,24 @@ message ListGitCommitMetadataForReferenceRequest {
...
+  // Hash is the SHA1 hash of the git commit.
+  string hash = 1;
+  // Author is metadata associatd with the author of the git commit.
+  // This may not always be available, so it is not always populated.
+  optional GitIdentity author = 2;
+  // Commiter is the metadata sscoiated with the commiter of the git commit.
+  // This may not always be available, so it is not always populated.
+  optional GitIdentity commiter = 3;
+  // Tag names of the associated tags with the git commit.
+  repeated string tags = 4;
```
unmultimedio, line 116: I assume we will return this from labels in the tag namespace? I'm asking because those tags could be populated not only from the synced git metadata, but also from a regular `buf push --tag`. How do you feel moving this to the `ListGitCommitsForReferenceResponse`?

Sadly, the way things are in the BSR, I don't think we will be able to know exactly from which git commit (or if any) came a certain tag from, and a more product-led comment: do we want to distinguish git tags vs tags from buf push?

seankimdev: > I don't think we will be able to know exactly from which git commit (or if any) came a certain tag from

The UI design includes showing the associated tags for each row of git commit. Would it not be possible in that case?
image

Also, if we want to provide a way for users to see paginated lists of associated tags for a git commit, would it be worth making a separate rpc instead to return the tags? The behaviour would align with our decision to replace the `tags` field with `tagCount` from `RepositoryCommit`, and using `ListRepositoryTagsForReference` instead. We'd also want to add `tagCount` to the `GitCommitInformation` if we go in that direction.

doriable: Hmm, so I don't think we should be distinguishing git tags from BSR tags necessarily, since I don't think that degree of granularity really helps the user. I think maybe the point here is not to show the tags, and just show the tags on the BSR commit?

doriable: Since the tags are associated with the BSR commit and not necessarily associated with the git commit, right?

seankimdev: Yeah I suppose Git Commit SHA's are sufficient to provide the purpose of its UI to provide the relationship between the BSR commit and Git commits. The "Associated tags" section would provide all tags that are associated with the BSR commit, whether they have been pushed from git tags or buf push.
image

doriable: Yeah, I think that is probably the right way to go, tbh.

unmultimedio: Ok so just to summarize, we're ok that with the recent change to deduplicate information in the BSR, tags will be linked only to the BSR commit and not to individual git commits, right? Anyway, all those tags would point to the same BSR commit reference.

This also means when listing git commits associated to a BSR commit, we cannot show any tags on each git commit, only general tags for the BSR commit.

cc @akshayjshah

If we're ok with this, I think the original comment still stands: field `repeated string tags` cannot be part of the git commit info message, and should be moved to the parent message `ListGitCommitsForReferenceResponse`.

akshayjshah: If that's the compromise required to make tags synced from git and manually-applied tags behave the same, that's fine with me. If it's not terribly difficult, it would be nice to have a record _somewhere_ of git commit <--> tag mappings, even though we're not currently showing them in the UI.

unmultimedio: It's not difficult, but my fear is the constraints we need to apply to make sure that relationship remains consistently in sync after a git tag is removed or moved to another git commit.

proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -100,32 +94,22 @@ message ListGitCommitMetadataForReferenceRequest {
   string page_token = 5;
 }
 
-message ListGitCommitMetadataForReferenceResponse {
+message ListGitCommitsForReferenceResponse {
   // ID of the BSR commit the reference resolved to.
   string commit_id = 1;
   // List of git commits and metadata associated with the resolved reference.
-  repeated GitCommitMetadata git_commit_metadatas = 2;
+  repeated GitCommitInformation git_commit_metadatas = 2;
```
unmultimedio, line 101: Nit
```suggestion
  repeated GitCommit git_commits = 2;
```

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/reference.proto
doriable: I also don't think we should move the tag list to the parent response message -- it would not be paginated, and this may cause some issues. We already have `ListRepositoryTagsForReference` for that purpose.

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/reference.proto
doriable: I think we can have the data model support an indication of whether a tag is from git or not, but I don't think it directly affects this API at the the moment.

Discussing with @unmultimedio, we would have a table manage the tags that would be related to the labels. The labels would track the history of the tag, and the tags table would have metadata, such as whether or not it is from git.
I will have a separate issue to track that body of work, but in either case, for the current version of this API, we would not need to populate this information here.

seankimdev (approved):

proto/buf/alpha/registry/v1alpha1/reference.proto
unmultimedio: yeah, I agree.

> we would not need to populate this information here.

Once we have that additional table, we could populate those tags per commit in this response.

PR #2326, opened 2023-07-27
Get/Set repositories' default branch field

unmultimedio:
This allows `buf alpha repo sync` command to check if the local git default branch matches the BSR's, and abort if they don't.


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -162,6 +162,16 @@ service RepositoryService {
   rpc GetRepositoriesMetadata(GetRepositoriesMetadataRequest) returns (GetRepositoriesMetadataResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // GetRepositoryDefaultBranch retrieves the default_branch value for a given repository. By
+  // default a BSR repository's default branch is "main", which is the value used when a module is
+  // pushed to the BSR.
+  rpc GetRepositoryDefaultBranch(GetRepositoryDefaultBranchRequest) returns (GetRepositoryDefaultBranchResponse) {
+    option idempotency_level = NO_SIDE_EFFECTS;
+  }
```
unmultimedio, line 170: Instead of this RPC we could include the `default_branch` as a field in the `message Repository`, but then in the backend we'd need to make sure we always include the field in all responses. Thoughts? @doriable @seankimdev

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -162,6 +162,16 @@ service RepositoryService {
...
+  // SetRepositoryDefaultBranch allows repositories owners or admins to change the repository's
+  // default branch. This is useful for syncing git repositories with default branch names that are
+  // not "main".
+  rpc SetRepositoryDefaultBranch(SetRepositoryDefaultBranchRequest) returns (SetRepositoryDefaultBranchResponse);
```
unmultimedio, line 174: Similarly, we could append this field to the `UpdateRepositorySettingsByName` RPC, which is already behind `admin+` permissions. cc @seankimdev @doriable

proto/buf/alpha/registry/v1alpha1/repository.proto
seankimdev: Isn't this a duplicate rpc to the existing [GetCurrentDefaultBranch](https://buf.build/bufbuild/buf/docs/main:buf.alpha.registry.v1alpha1#buf.alpha.registry.v1alpha1.RepositoryBranchService.GetCurrentDefaultBranch)?

doriable: I am a fan of having `default_branch` as a field on the `Repository` message. I think it provides a lot more clarity.

Also, just noting that we have moved the discussion on whether this encroaches on `GetCurrentDefaultBranch` onto Slack.

proto/buf/alpha/registry/v1alpha1/repository.proto
doriable: Also in agreement with this.

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -39,7 +39,13 @@ service RepositoryBranchService {
...
     option idempotency_level = NO_SIDE_EFFECTS;
   }
-  // GetCurrentDefaultBranch returns the branch name that is mapped to the main/BSR_HEAD.
+  // GetCurrentDefaultBranch returns the branch name that is mapped to the main/BSR_HEAD. This might
+  // not be the same value in the repository's `default_branch` field, since that value can be
+  // changed at will by repository's owners/admins for syncing git repositories. This RPC retrieves
+  // the branch from the latest commit labeled as BSR_HEAD, even if that value differs from the one
+  // stored in the `default_branch` field.
+  //
+  // TODO: Rename this RPC to something more appropriate like "GetLatestHEADCommit".
```
seankimdev, line 48: We can probably remove or deprecate it instead, but we can decide later outside of this pr.

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -55,6 +55,10 @@ message Repository {
   // url is the user configurable URL in the description of the repository,
   // always included the scheme and will not have a #fragment suffix.
   string url = 12;
+  // default_branch in a BSR repository. It is used when syncing a git repository, to make sure both
+  // default branches (BSR and Git) are in sync. By default, every BSR repository is created with a
```
seankimdev, line 59: What happens if the default branches (BSR and Git) are out of sync? Would it reject the sync until the user manually changes the default branch in BSR to match?

seankimdev (approved):
Lgtm, just non-blocking question/comment.

proto/buf/alpha/registry/v1alpha1/repository.proto
unmultimedio: That's correct, but that's in the `buf sync` scope, in an upcoming PR.

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
unmultimedio: Agree, I'll leave it to @doriable and you to coordinate on that one 😄

proto/buf/alpha/registry/v1alpha1/repository.proto
unmultimedio: Here's that PR: https://github.com/bufbuild/buf/pull/2328

doriable (approved):

PR #2328, opened 2023-07-28
Buf sync: Make sure git default branches is in sync with BSR's

unmultimedio:
When running a sync command, before syncing anything, make sure the git repository's default branch name matches with the default branch for the remote BSR repository.


doriable (approved):
lgtm 👍🏼

PR #2329, opened 2023-07-31
Sync single branch by default

unmultimedio:
The expected most common use case for the `sync` command is in CI, on pull requests or individual pushes to remotes. This means we should sync a single branch by default, whatever one I'm currently checked in. Also added a flag `--all-branches` to go in the usual sync order and sync all remote branches.


private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -61,8 +62,9 @@ func NewCommand(
 		Use:   name,
 		Short: "Sync a Git repository to a registry",
 		Long: "Sync a Git repository's commits to a registry in topological order. " +
-			"Only commits belonging to the 'origin' remote are processed, which means that " +
-			"commits must be pushed to a remote. " +
+			"Only commits in the current branch that are pushed to the 'origin' remote are processed. " +
+			"Syncing all branches is possible using '--all-modules' flag." +
+			// TODO rephrase in favor of a default module behavior.
```
doriable, line 67: This TODO is already addressed, right? The default behaviour indicated here only syncs the default branch. Although, why is `--all-modules` flag used to sync all branches? Should this be `--all-branches`?

doriable (commented):
Changes look reasonable, just a nit/question.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
unmultimedio: > This TODO is already addressed, right? The default behaviour indicated here only syncs the default branch.

No, the TODO I'm referring to in here is the one about a default module at the root of the repo, instead of forcing the user to pass all modules using the `--module` flag.

> Although, why is --all-modules flag used to sync all branches? Should this be --all-branches?

Nice catch! yeah I meant `--all-branches`, I'll fix.

unmultimedio: Fixed.

doriable (approved):
lgtm!

PR #2336, opened 2023-08-02
Return better errors for unavailable error codes

pkwarren:
If we can determine that the error from login is due to a connect unavailable error, we should return more details to aid in troubleshooting. Update wrapError to return a better error message in case of a TLS certificate error (could be a MITM attack or an untrusted certificate).

### DIFF
--- a/private/buf/bufcli/errors.go
+++ b/private/buf/bufcli/errors.go
@@ -174,6 +174,8 @@ func wrapError(err error) error {
 			if dnsError := (&net.DNSError{}); errors.As(err, &dnsError) && dnsError.IsNotFound {
 				return fmt.Errorf(`%s Are you sure "%s" is a valid remote address?`, msg, dnsError.Name)
 			}
+			// If the unavailable error wraps a tls.CertificateVerificationError, show a more specific error message
+			// to the user to aid in troubleshooting.
 			if tlsErr := wrappedTLSError(err); tlsErr != nil {
 				return fmt.Errorf("tls certificate verification: %w", tlsErr)
 			}
--- a/private/buf/bufcli/errors_tls_go119.go
+++ b/private/buf/bufcli/errors_tls_go119.go
@@ -22,6 +22,8 @@ import (
 )
 
 // wrappedTLSError returns an unwrapped TLS error or nil if the error is another type of error.
+// This method is a workaround until we can switch to use errors.As(err, *tls.CertificateVerificationError),
+// which is a new error type introduced in Go 1.20. This can be removed when we upgrade to support Go 1.20/1.21+.
 func wrappedTLSError(err error) error {
 	wrapped := errors.Unwrap(err)
 	if wrapped == nil {


pkwarren:
When testing locally, the behavior is:

Before:
```
$ printf '...' | buf registry login <remote> --token-stdin --username <username>
Failure: invalid token provided
```

After:
```
$ printf '...' | buf registry login <remote> --token-stdin --username <username>
tls: failed to verify certificate: x509: “api.<remote>” certificate is not trusted
```

private/buf/bufcli/errors.go
```diff
@@ -174,7 +175,9 @@ func wrapError(err error) error {
 			if dnsError := (&net.DNSError{}); errors.As(err, &dnsError) && dnsError.IsNotFound {
 				return fmt.Errorf(`%s Are you sure "%s" is a valid remote address?`, msg, dnsError.Name)
 			}
-
+			if tlsErr := (&tls.CertificateVerificationError{}); errors.As(err, &tlsErr) {
+				return tlsErr
```
unmultimedio, line 179: Just to be safe from connect messages error rephrases in the future, let's wrap it in a message?
```suggestion
				return fmt.Errorf("tls certificate verification: %w", tlsErr)
```

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -188,6 +188,9 @@ func inner(
 	authnService := connectclient.Make(clientConfig, remote, registryv1alpha1connect.NewAuthnServiceClient)
 	resp, err := authnService.GetCurrentUser(ctx, connect.NewRequest(&registryv1alpha1.GetCurrentUserRequest{}))
 	if err != nil {
+		if connectErr := new(connect.Error); errors.As(err, &connectErr) && connectErr.Code() == connect.CodeUnavailable {
+			return connectErr
+		}
 		// We don't want to use the default error from wrapError here if the error
 		// an unauthenticated error.
 		return errors.New("invalid token provided")
```
unmultimedio, line 196: I thought you were going to invoke `wrapError` here.

unmultimedio (commented):
lgtm

private/buf/bufcli/errors.go
pkwarren: Done.

private/buf/bufcli/errors_tls_go119.go
```diff
@@ -0,0 +1,34 @@
...
+//go:build !go1.20
+
+package bufcli
+
+import (
+	"errors"
+	"strings"
+)
+
+// wrappedTLSError returns an unwrapped TLS error or nil if the error is another type of error.
```
bufdev, line 24: I'd like to see a bit more documentation here as to the motivation for this, so that in the future we understand why this is here in case we need to debug

pkwarren: Ok - this is just a temporary thing until Go 1.21 is released and we can remove and just use errors.As which will be a lot more readable. I'll update to add more details though.

bufdev (approved):

PR #2341, opened 2023-08-02
Add module content to search results

cyinma:


mfridman (approved):
One endpoint to rule them all.

I don't have much context, but on the surface, it feels odd to slam all the possible search items under one RPC.

Searching within a module seems different than searching across the BSR for resource names.

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -77,6 +79,21 @@ message CuratedPluginSearchResult {
   bool deprecated = 4;
 }
 
+message ModuleElementSearchResult {
+  string repository_id = 1;
+  string repository_owner = 2;
+  string repository_name = 3;
+  string proto_file_path = 4;
+  string fully_qualified_name = 5;
```
paul-sachs, line 87: I guess with the fully qualified name I can build urls like https://buf.build/acme/petapis/docs/main:pet.v1#pet.v1.PetStoreService, we kinda need to package name as well. Unless all fully qualified names can be transformed into package names easily. I'm not sure that's possible.

paul-sachs (approved):

Issue #1263, opened 2022-07-05
Support configuration for `go_package_prefix` package name


doriable:
Currently, we are create a package declaration based on the final two elements of the path of the package: 

> If the Protobuf file's package declaration conforms to the PACKAGE_VERSION_SUFFIX lint rule, the final two path elements are concatenated and included after the ; element in the go_package result. The above example generates a Go package with a package declaration equal to weatherv1, which enables you to import Go definitions from a variety of generated packages that would otherwise collide (a lot of Protobuf packages contain the v1 suffix).

https://docs.buf.build/configuration/v1/buf-gen-yaml#default

This may not be desirable for all project structures, in particular ones that are more deeply nested.
We should consider a configuration under `go_package_prefix` that allows the user some way to specify depth of the package tag, e.g.

```
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/acme/weather/private/gen/proto/go
    package_name_depth: 3 <default 2>
```

We could also consider exposing per package overrides for this as well.

kucjac:
I think, this is a proper place to discuss the implementation details as discussed in the: https://github.com/bufbuild/buf/pull/2343#issuecomment-1664171302.

doriable:
Given that we now have `v2` buf.gen.yaml support for overrides and disables, I think we can close this issue for now. If there are additional questions or concerns on this, we can re-open.

PR #2343, opened 2023-08-03
buf generate: go_package_prefix implemented package_depth option

kucjac:
This PR implements an option for generating Golang files, for the versioned protobuf packages.
It allows changing the depth of the package name generation for more nested `proto` packages.

Currently buf treats file:
```a.proto

package weather.v1alpha;
``` 

and generates its identity go_package as =";weatherv1alpha" which is perfectly fine and directly describes what this package does and what is its version.

However in a situation where the proto file package is more deeply nested i.e.:
```b.proto

package weather.admin.v1alpha;
```

This generates go_package as: `";adminv1alpha"`. For modules with more nested proto packages this makes golang package names a little not descriptive enough.

This PR provides `buf` generate option to override protobuf package name generation depths. By default, it takes a depth of **2** (i.e. `weather.v1alpha` -> `weatherv1alpha`). This PR allows modifying this depth manually, i.e.:

```buf.gen.yaml
version: v1
managed:
  enabled: true
  go_package_prefix:
    default: github.com/acme/weather/private/gen/proto/go
    except:
    package_depth:
      weather.admin.v1alpha: 3
plugins:
  - plugin: go
    out: gen/proto/go
    opt: paths=source_relative
```

With these options the `proto` package `weather.admin.v1alpha` will result with a `Golang` package identity => `weatheradmingv1alpha`, and thus option go_package = ";weatheradmingv1alpha"

This PR solves the problem of complex repositories with different submodules.
It also relates to the #1263.

The changes in this package breaks the API of the Golang function: 
`private/bufpkg/bufimage/bufimagemodify.GoPackage` by adding additional argument `packageDepthOverride map[string]uint`. If this argument has no value, this doesn't change a logic of the function.
The value for that argument is taken from the configurations, and provided in the `buf/private/buf/bufgen/generator.go:525` directly out of the configuration.

The configuration field `bufgen.GoPackagePrefixConfig.PackageDepth` is also affected by the `bufgen.ExternalGoPackagePrefixConfig.PackageDepth` field.

It also breaks the API of the Golang function:
`private/buf/pkg/bufimage/bufimagemodify.GoPackageImportPathForFile` by adding the number of the version parts to split. By default it takes the previous value of `2`.

Generation also logs a Warning message if none of the package_depth entries were used during generation.

The new options are included in the test cases of the `buf/private/bufpkg/bufimage/bufimagemodify` as well as in the e2e test on manually compiled `buf` CLI on real data.

### DIFF
--- a/private/buf/bufgen/bufgen.go
+++ b/private/buf/bufgen/bufgen.go
@@ -275,7 +275,8 @@ type GoPackagePrefixConfig struct {
 	Default string
 	Except  []bufmoduleref.ModuleIdentity
 	// bufmoduleref.ModuleIdentity -> go_package prefix.
-	Override map[bufmoduleref.ModuleIdentity]string
+	Override     map[bufmoduleref.ModuleIdentity]string
+	PackageDepth map[string]uint
 }
 
 // ObjcClassPrefixConfig is the objc_class_prefix configuration.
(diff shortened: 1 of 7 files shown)


bufdev:
Thanks for the thoughtful PR, we appreciate the effort here!

A few thoughts:

- Being able to customize the Golang package value as part of managed mode is something we'd love to consider.
- We're hard at work on `v2` of `buf.gen.yaml` at the moment, and don't anticipate any more changes to v1. We'd consider adding something like this as part of `v2` when that is ready.
- The actual UI I think we'd want to debate a bit. Instead of package depth, I could imagine we'd instead want an explicit name override, for example.

Can you file an issue with all of the above context so we can talk about this there?

kucjac:
Thanks for your reply. I'm glad to hear that you're working on the `v2`, as `buf` is a fantastic solution that finally makes protobuf handling easy.

I agree that the UI debate is a good point to start, I based my assumptions on the issue: https://github.com/bufbuild/buf/issues/1263, and I think we should continue discussion already over there, to not duplicate issues.

Let me know if I can help you somehow (in coding or discussion of the features) in the implementation of a new version.

PR #2344, opened 2023-08-03
Implement file options in managed mode v2

oliversun9:
This PR 
* moves `Override` from `bufimagemodifyv2` to `bufgenv2`, see https://github.com/bufbuild/buf/pull/2311#discussion_r1270887980.
* implements the all file options

Notes:
* In `v1`, `java_outer_classname` and `java_multiple_files` [preserve exisiting values in the proto source files](https://github.com/bufbuild/buf/pull/2002), meaning that if a proto file has `option java_outer_classname = xyz;`, managed mode will not modify this option for this file. This is not added to `v2` yet, and I don't think we should have this behavior in `v2` because with `disable` in `v2` users can solve this problem. (For example, generating for [message `GenericConnectionPoolProto` in envoy](https://buf.build/envoyproxy/envoy/file/main:envoy/extensions/upstreams/http/generic/v3/generic_connection_pool.proto) would fail if managed mode sets its source file's `java_outer_classname` to `GenericConnectionPoolProto` because it collides with the class name)
* `cc_enable_arenas` is still added, although deprecated, because I think users will less likely to migrate to `v2` if they see diffs on byte arrays in the generated code. It is added to avoid producing this diff, because this can affect users who do not generated cpp files at all. We can omit this option in `v2` docs so that new users won't see it or use it.
* `java_string_check_utf8` is still added, because it turned out to be NOT deprecated (I was wrong). For more information, this file option has no effect in `proto3` but it does in `proto2`.

Next steps:
* add a converter: I manually converted the sample `buf.gen.yaml` in managed mode section of the docs, and it turned out to require a certain (non-trivial) level of understanding of both `v1` and `v2` of this configuration. I think it would make a real difference if there is a converter.
* add integration tests


private/buf/bufgen/internal/bufgenv2/override.go
```diff
@@ -12,35 +12,37 @@
...
+}
+
 type prefixOverride string
 
 func newPrefixOverride(prefix string) prefixOverride {
 	return prefixOverride(prefix)
 }
 
-func (p prefixOverride) Get() string {
+func (p prefixOverride) get() string {
```
bufdev, line 29: nit: id keep these all as public, since the type is private they are not exposed, and this makes it easy to switch between this type being public and private in the future. I generally use private functions just for truly-internal-to-the-type stuff

private/buf/bufgen/internal/bufgenv2/file_option.go
```diff
@@ -59,8 +58,21 @@ const (
 	fileOptionPhpMetadataNamespaceSuffix
 	// fileOptionRubyPackage is the file option ruby_package.
 	fileOptionRubyPackage
+	// fileOptionRubyPackageSuffix is the file option ruby_package_suffix.
+	fileOptionRubyPackageSuffix
 	// groupJavaPackage is the file option group that modifies java_package.
 	groupJavaPackage fileOptionGroup = iota + 1
+	groupJavaOuterClassname
```
bufdev, line 65: Can you quickly explain why we need groups for all of these?

bufdev (commented):
Overall looks good, just a few comments

private/buf/bufgen/internal/bufgenv2/file_option.go
oliversun9: There can be multiple `file_option`s in `buf.gen.yaml` that modify the same file option on an image file's descriptor set (e.g. `java_package`, `java_package_prefix` and `java_package_suffix` all modify java package). The order among rules with different `file_options` matters if they modify the same file option.
For example:
```
override: ## values omitted
  - file_option: java_package
  - file_option: java_package_prefix
 ```
and
```
override:
  - file_option: java_package_prefix
  - file_option: java_package
```
have different effects.
This is achieved by putting override in different groups and for each group, before we call `mergeOverrideFunc`. [`mergeOverrideFuncs` can then decide what happens if the last two are prefix and suffix, or prefix and value and etc.](https://github.com/bufbuild/buf/pull/2344/files#diff-6ad607dc37b36d1f3e43504fb3b079819f8d2cca6cb92f7b16c2714ef2b64e86R404)

Another approach is to attach each rule with a number (the position it appears in the list of override rules).

bufdev (approved):
Mostly interested in final product when we QA it, but most of this looks like it's following existing patterns, so LGTM!

PR #2356, opened 2023-08-09
Buf sync: support module identity overrides

unmultimedio:
Accept an **optional** module identity override for the module flag. This is useful when syncing a repository, to set a target module identity, regardless of what HEAD content has.

This will be useful for dog fooding the sync command on our own repos, or other community repositories with valid Buf modules, in case we don't want or can't write to the git history, but still want to sync all activity across all branches.

Before:

```sh
buf alpha repo sync \
  # This would read the module name in the HEAD commit,
  # and only sync commits back in time that matched that name.
  --module my/proto/dir
```

Now:

```sh
buf alpha repo sync \
  # This will read the module name in the HEAD commit,
  # and only sync commits back in time that matched that name.
  --module my/proto/dir
  # This will use "buf.build/acme/foo" as the module target
  # identity for all branches, regardless of the module name
  # they have in the git commits, or even if they are unnamed.
  --module other/proto/dir:buf.build/acme/foo
```

### DIFF
--- a/private/buf/bufsync/bufsync.go
+++ b/private/buf/bufsync/bufsync.go
@@ -61,13 +61,20 @@ type ReadModuleError struct {
 }
 
 // Code returns the error code for this read module error.
-func (err *ReadModuleError) Code() ReadModuleErrorCode {
-	return err.code
+func (e *ReadModuleError) Code() ReadModuleErrorCode {
+	return e.code
 }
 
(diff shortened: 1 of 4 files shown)


doriable (approved):
Thanks for adding this back so fast! 🙏🏼

doriable (approved):

PR #2368, opened 2023-08-10
Enable all branches sync

unmultimedio:
We can enable all branches now that `main` synced once: https://github.com/bufbuild/buf/actions/runs/5822579536

### DIFF
--- a/.github/workflows/buf-shadow-sync.yaml
+++ b/.github/workflows/buf-shadow-sync.yaml
@@ -7,7 +7,6 @@ on:
 jobs:
   sync:
     runs-on: ubuntu-latest
-    if: github.ref == 'refs/heads/main' # TODO enable all refs once `main` is synced at least once
     steps:
       - uses: actions/checkout@v3
         with:


doriable (approved):

bufdev:
> We can enable all branches now that main synced once:

Are we saying that for users, they can't enable all branches from the start, instead they need to do a two-step process?

unmultimedio:
> > We can enable all branches now that main synced once:
> 
> Are we saying that for users, they can't enable all branches from the start, instead they need to do a two-step process?

Currently sync supports only two branches selection: current branch only, or all branches. And since nothing is synced when first onboarding, I can see two options to make sure the longest tree branch is assigned to `main` and not another branch:

1. What I did here. First turn it on only for the main branch, wait for a successful complete sync, and then turn it on for every other branch.
2. Run it the first time with `--all-branches`, even from a PR, which will pick the default branch first, and then before merging the PR remove that flag so no race conditions will happen cross-branches.

Probably there's a better way that improves this UX, but for now we can count this as the first improvement finding from dogfooding? hehe

PR #2369, opened 2023-08-10
Add server review flow grace period

stefanvanburen:
Minor notes below.


proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -40,6 +40,13 @@ message UniquenessPolicy {
   bool enabled = 1;
 }
 
+// GracePeriodPolicy is the policy for the review flow grace period on the server.
+message GracePeriodPolicy {
+  // The number of seconds of the grace period.
+  // Set to zero to disable the grace period.
+  uint32 grace_period_seconds = 1;
```
stefanvanburen, line 47: considered a `google.protobuf.Duration` here, but a) don't need to represent negative durations, and b) don't need nanosecond precision.

A `uint32` is ~4 billion seconds, which is >100 years, which is more than we're ever going to need to set this value to.

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -40,6 +40,13 @@ message UniquenessPolicy {
   bool enabled = 1;
 }
 
+// GracePeriodPolicy is the policy for the review flow grace period on the server.
+message GracePeriodPolicy {
```
pgmitche, line 44: To associate grace period with it's relevant feature
```suggestion
message BreakingChangeGracePeriodPolicy {
```

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -111,6 +118,8 @@ service AdminService {
   rpc RunServerUniquenessBackfillTask(RunServerUniquenessBackfillTaskRequest) returns (RunServerUniquenessBackfillTaskResponse) {
     option idempotency_level = IDEMPOTENT;
   }
+  // Update review flow grace period policy for the server.
+  rpc UpdateGracePeriodPolicy(UpdateGracePeriodPolicyRequest) returns (UpdateGracePeriodPolicyResponse);
```
pgmitche, line 122: Will the front end need to render an existing/currently set grace period? Do we need a `Get` for that?

proto/buf/alpha/registry/v1alpha1/admin.proto
pgmitche: Or governance, review flow. Whatever the umbrella for this is

proto/buf/alpha/registry/v1alpha1/admin.proto
stefanvanburen: added in https://github.com/bufbuild/buf/pull/2369/commits/b2df0de1e196a4fbb6b95c0f51b446c2c0e1e2b2, thanks!

proto/buf/alpha/registry/v1alpha1/admin.proto
stefanvanburen: prefixed with "ReviewFlow" in https://github.com/bufbuild/buf/pull/2369/commits/e47cbf039824eeaa77b18f57d597ab65f6a5cd10 :^)

pgmitche (approved):

PR #2370, opened 2023-08-10
Implement migration from buf.gen.yaml v1 to v2

oliversun9:
Implements migration from `v1` (add `v1beta1`) to `v2`. Migrate by appending `--migrate-v1` to the command you would normally run. For example:
```
$ buf generate --include-imports --include-wkt --type d.one.D1A --type d.one.D1B --path d/one --type d.one.D1X --exclude-path d/one/d_one_y.proto buf.build/osun/d1 --template buf.gen.yaml --output build --migrate-v1
Migration finshed, to generate with buf, run:
buf generate --template buf.gen.yaml --output build
```

Main changes are in `generate.go` and `bufgen/migrate.go`, changes in `bufgenv1` and `bufgenv2` are mainly exporting types and constants.


private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -321,10 +381,6 @@ func run(
 	if err != nil {
 		return err
 	}
-	var includedTypesFromCLI []string
-	if len(flags.Types) > 0 || len(flags.TypesDeprecated) > 0 {
```
oliversun9, line 325: These few lines of code aren't removed, they are moved up.

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -244,13 +248,13 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		&f.BaseOutDirPath,
 		baseOutDirPathFlagName,
 		baseOutDirPathFlagShortName,
-		".",
+		baseOutDirFlagDefault,
```
oliversun9, line 251: The other flag defaults aren't put into variables because they are zero-values of their respective types

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -298,16 +308,66 @@ func run(
...
+			updatedGenerateCommand += getStringFlagText(errorFormatFlagName, flags.ErrorFormat)
+		}
+		if flags.Config != "" {
+			updatedGenerateCommand += getStringFlagText(configFlagName, flags.Config)
+		}
+		if flags.DisableSymlinks {
+			updatedGenerateCommand += getBoolFlagText(disableSymlinksFlagName)
+		}
+		message := fmt.Sprintf("Migration finshed, to generate with buf, run:\n%s\n", updatedGenerateCommand)
+		_, err := container.Stderr().Write([]byte(message))
```
oliversun9, line 367: I assumed `Stderr` is the place to print it.

private/buf/buffetch/format.go
```diff
@@ -14,6 +14,33 @@
 
 package buffetch
 
+// TODO: remove the unexported formats below and only use these ones.
+// A good time to do so is when we merge txtpb and binpb into this branch.
+const (
+	// formatBin is the binary format.
```
oliversun9, line 20: Instead of renaming `formatXyz` to `FormatXYZ`, which will produce a lot of diff in this PR, I added `FormatXyz` as a temporary measure.
This will be cleaned up when we merge main into this branch, which will introduce `txtpb` and `binpb` to this branch.

private/buf/buffetch/ref_parser.go
```diff
@@ -180,6 +180,25 @@ func newSourceOrModuleRefParser(logger *zap.Logger) *refParser {
 	}
 }
 
+func (a *refParser) GetRefFormat(
```
oliversun9, line 183: This method is useful because now we can look at an input string, say `dir/file.git` and call this method to know that this is a git repo, and then we can set `externalConfigV2.GitRepo = dir/file.git`

private/buf/bufgen/internal/bufgenplugin/plugin_config.go
```diff
@@ -117,7 +118,10 @@ func (c *binaryPluginConfig) PluginName() string {
 	if c.name != "" {
 		return c.name
 	}
-	return c.path[0]
+	if len(c.path) == 1 {
+		return c.path[0]
+	}
+	return stringutil.SliceToString(c.path)
```
oliversun9, line 124: This update isn't strictly related to this PR, but this improves error reporting and in some other areas. For example, if a plugin is invoked by `go run github.com/example/cmd/protoc-gen-someplugin`, some error message will look like  `plugin [go, run, github.com/example/cmd/protoc-gen-someplugin] failed`, instead of `plugin go failed`

private/buf/bufgen/internal/bufgenv1/config.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 // Config is a configuration.
-type config struct {
+type Config struct {
```
oliversun9, line 35: Occurrences of `config` -> `Config` are the only changes in this file

private/buf/bufgen/internal/bufgenv1/configv1_test.go
```diff
@@ -34,7 +34,7 @@ import (
 
 func TestReadConfigV1Beta1(t *testing.T) {
 	truth := true
-	successConfig := &config{
+	successConfig := &Config{
```
oliversun9, line 37: Occurrences of config -> Config are the only changes in this file

private/buf/bufgen/internal/bufgenv2/config.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 // disableFunc decides whether a file option should be disabled for a file.
-type disabledFunc func(fileOption, imageFileIdentity) bool
+type disabledFunc func(FileOption, imageFileIdentity) bool
```
oliversun9, line 32: Occurrences of `fileOption` -> `FileOption` are the only changes in this file

private/buf/bufgen/internal/bufgenv2/file_option.go
```diff
@@ -24,42 +24,42 @@ import (
 )
 
 const (
-	// fileOptionJavaPackage is the file option java_package.
```
oliversun9, line 27: Occurrences of `fileOptionXyz` -> `FileOptionXyz` are the only changes in this file

private/buf/bufgen/internal/bufgenv2/managed_mode.go
```diff
@@ -52,14 +52,14 @@ func applyManagementForFile(
 		}
 		switch fileOptionGroup {
 		case groupJavaPackage:
-			if managedConfig.DisabledFunc(fileOptionJavaPackage, imageFile) {
```
oliversun9, line 55: Occurrences of fileOptionXyz -> FileOptionXyz are the only changes in this file

private/buf/bufgen/migrate.go
```diff
@@ -0,0 +1,553 @@
...
+		genTemplate: ExternalConfigFilePath,
+	}
+	for _, option := range migrateOptions {
+		option(options)
+	}
+	switch filepath.Ext(options.genTemplate) {
+	case ".json", ".yaml", ".yml":
+		// OK
+	default:
+		return errors.New("migration can only apply to a file on disk with extension .yaml, .yml or .json")
```
oliversun9, line 71: Note: we are not migrating anything if the template is passed on command line, i.e. `buf generate --template {"version": "v1"...`
We don't know where to put the file if there is no file to begin with. Maybe print to stdout/stderr? But that goes against the direction of making the `buf generate` command shorter.

private/buf/bufgen/migrate.go
```diff
@@ -0,0 +1,553 @@
...
+			break
+		}
+		// At this point, we know for certain that this plugin is not protoc-builtin.
+		// It's possible that the plugin is a valid binary plugin but not installed on
+		// the user's environment.
+		// TODO: we can also treat it as a binary plugin, leaving a comment like so:
+		// # we couldn't find protoc-gen-xyz locally and you should verify that it is
+		// # a binary installed locally
+		// binary_plugin: protoc-gen-xyz
+		return nil, fmt.Errorf("plugin protoc-gen-%s is not found locally and %s is not built-in to protoc", pluginName, pluginName)
```
oliversun9, line 245: Checking from `PATH` and then checking from the list of protoc plugins is how we have been deciding the type of plugin.

private/buf/buffetch/format.go
```diff
@@ -14,6 +14,33 @@
...
+	FormatBin = "bin"
```
bufdev, line 21: Optimally, these should not be public - the idea is all of this is hidden within `buffetch`. Exposing these breaks the `buffetch` model. I don't have any great answers, but it feels like we're really exposing `buffetch` internals too much. Perhaps this is the best we can easily do.

private/buf/bufgen/internal/bufgenv2/config.go
bufdev: Optimally, this shouldn't be public either.

private/buf/bufgen/migrate.go
```diff
@@ -0,0 +1,553 @@
...
+		}
+	case ".yaml", ".yml":
+		configV2Data, err = encoding.MarshalYAML(&externalConifgV2)
+		if err != nil {
+			return err
+		}
+		configV2Data = append([]byte(migratedV2FileYAMLHeader), configV2Data...)
+	default:
+		// This should not happen because we already checked this at the beginning of this function.
+		return errors.New("migration can only apply to a file on disk with extension .yaml, .yml or .json")
```
bufdev, line 130: Should print file name as part of error

private/buf/bufgen/migrate.go
```diff
@@ -0,0 +1,553 @@
...
+		inputConfig.Tarball = &path
+	case buffetch.FormatTargz:
+		inputConfig.Tarball = &path
+		compression := "gzip"
+		inputConfig.Compression = &compression
+	case buffetch.FormatZip:
+		inputConfig.ZipArchive = &path
+	case buffetch.FormatProtoFile:
+		inputConfig.ProtoFile = &path
+	// TODO: txtpb and jsonpb
```
bufdev, line 479: Should be in this PR optimally

private/buf/bufgen/migrate.go
```diff
@@ -0,0 +1,553 @@
...
+	path, options, err := bufref.GetRawPathAndOptions(input)
+	if err != nil {
+		return nil, err
+	}
+	format, err := buffetch.NewRefParser(logger).GetRefFormat(ctx, input)
+	if err != nil {
+		return nil, err
+	}
+	switch format {
+	case buffetch.FormatBin:
```
bufdev, line 451: This (and the below) feels like an abuse of the `buffetch` API  boundary - it feels like all of this stuff should somehow live at the buffetch layer. I don't know, I don't have any great answers to be honest.

bufdev:
Another note - can we change `--migrate-v1` to `--migrate`, and then later just migrate people to whatever the latest is?

private/buf/buffetch/format.go
oliversun9: This is the easiest way I found. The two alternatives I considered are:
1. Return a `RawRef`, which break the API even more, IMO
2. Lifted types like `ParsedGitRef` to `buffetch` from `buffetch.internal` and expose them. This also breaks the existing API of `bufftetch`.

private/buf/bufgen/migrate.go
oliversun9: `Txtpb` and `binpb` (not `jsonpb`, the comment was wrong) are now implemented in this branch as well

oliversun9:
> I expected the actual command to run after the migration is done, i.e. I expected `buf generate` to do the generation after the migration. Or potentially before, if we aren't sure the migration will be successful.

Now we continue to generate after migrating the template file. If the migration did not succeed, the user should either fix their template or remove `--migrate` if they wish to generate files.

> The template binary paths are now absolute, i.e. we went from `- name: connect-go` to `- binary: /Users/foo/.cache/buf/Darwin/x86_64/bin/protoc-gen-connect-go`. This isn't the best, it feels like we should be able to solve this

You are right, now it will just be protoc-gen-connect-go. It did not make sense to use an absolute path.

> `opt` went from a list to a single string - is this just what happens when there is only one value?

Originally I had a logic checking whether there is only one opt, but now I feel that it might be better just use whatever was in the original file (a string or a string slice)

> `strategy: directory` is added everywhere. This is the default, if it wasn't present before, it shouldn't be added.

This is fixed now.

> The `inputs` are going to be a bit weird. In this case, our templates were not in the same directory that we ran the command in. So now, we have `data/template/buf.go.gen.yaml` with an input `proto` - how is this expected to work? We might need to either heavily comment this on the migration, or do something else.

I added a message when this happens. `directory: %s is set based on input, and it should be interpreted as relative to the call site of buf.
For example, when running buf generate --template configs/buf.gen.yaml --migrate with directory: protos,
buf will look in protos, not config/protos.` I am not too sure if my wording will do well stressing our point and it looks a little verbose, any suggestions?

> Remove the generated comment at the top of the migrated file.

This is also fixed now.

bufdev:
Can we make `--migrate` a no-op with a warning if the migration has already happened, as opposed to an error?

bufdev:
Overall, this works incredibly well - good job.

Address the error vs warn comment, and I think that's sufficient. I am not crazy about some of the copy, but not worth bikeshedding on for a migration tool.

One other note: the migration doesn't preserve comments in the existing `buf.gen.yaml`. I don't think it's worth doing, and it also begs the question what comments to preserve.

bufdev (approved):
Fantastic job! Looking forward to the final QA of v2 :-)

PR #2371, opened 2023-08-10
Travel back from a custom git starting point

unmultimedio:
Add the ability to set a custom starting point when traveling back in the git tree. So far we have only needed to travel from HEAD of branches, but to be able to sync tags from recently synced git commits, we need to travel back from specific detached git commits.


private/pkg/git/git.go
```diff
@@ -300,6 +302,53 @@ type Repository interface {
 	Close() error
 }
 
+// ForEachCommitOption are options that can be passed to ForEachCommit.
+type ForEachCommitOption func(*forEachCommitOpts) error
+
+// ForEachCommitWithBranchStartPoint sets a branch name as a starting point to start the loop.
+func ForEachCommitWithBranchStartPoint(branchName string) ForEachCommitOption {
```
bufdev, line 309: How do we imagine the UX looking if we expect people to tell us the type of reference they're providing? I would prefer if we could differentiate programmatically

unmultimedio: I don't see right now a scenario in which external users would invoke this directly or via a command flag. So far "the users" are only us, users of the git package from a higher level, the sync command where we want to traverse the git tree from a custom starting points.

Now, about how to tell the type of provided reference, I thought about it and came up with two options:
1. Have options named differently: `...WithBranchStartPoint(branchName)`, `...WithHashStartPoint(hash)`, and as we support more types, we can add them later as new options.
2. Have a single option `...WithStartPoint(refName, refType)`, which in practice ends up with a few more exported types and consts, but at the end it's the same.

I chose the first one for more clarity in the option names, but I have no problem implementing the other one... probably it would make it a bit clearer, I'll try that one so we can check out the difference.

unmultimedio: Updated at https://github.com/bufbuild/buf/pull/2371/commits/92cc25ebf656f2e626dcec2f07c79c2d3c25b9b1, I think I like it better, but no strong take.

bufdev: I think the update is strictly worse - what we'd want is `WithStartReference(refName string)`, but if we have to specify the type of reference, a separate function is better.

What is this for if not to be used as a flag for the `sync` command?

unmultimedio: > I think the update is strictly worse - what we'd want is WithStartReference(refName string), but if we have to specify the type of reference, a separate function is better.

Ack, I'll revert it.

> What is this for if not to be used as a flag for the sync command?

It's not for a flag, it's for syncing older tags. Imagine these commits in a branch with a single module to sync:

- g (HEAD)
- f (to sync)
- e (to sync)
- d (sync start point)
- c (already synced)
- b (already synced)
- a (already synced) with tag "v1.2.3" that was created after this commit was synced

When running sync, we wouldn't ever look back past commit "d", so the tag "v1.2.3" in commit "a" would never be synced. With the change in this PR of setting a custom starting point, we could have an additional step the process:

1. find syncable commits: this would return commits [d, e, f, g]
2. before syncing them, do a "lookback" find, starting at "d" (**here we'll be using a custom starting point**), and sync missing tags if any. Internally we discussed X amount of commits or amount of time. Ideally we'd get to "a" and sync the missing tag.
3. then, actually sync commits d...g

It's much clearer and easier to implement as another step and not within step 1, because we'll be handling multiple modules, with potentially different starting points for each one.

This is not explicitly controlled by a flag, commit "d" is found by the command. We could implement a lookback flag to customize "how far to look back", but that doesn't set the starting point but the stop point.

unmultimedio: It's reverted now.

private/pkg/git/git.go
```diff
@@ -326,3 +375,20 @@ func OpenRepositoryWithDefaultBranch(name string) OpenRepositoryOption {
 		return nil
 	}
 }
+
+const (
```
bufdev, line 379: consts should always be at the top of the file, per the style guide

private/pkg/git/repository_test.go
```diff
@@ -66,19 +66,100 @@ func TestCommits(t *testing.T) {
...
+
 	assert.Equal(t, commits[0].Message(), "third commit")
 	assert.Contains(t, commits[0].Parents(), commits[1].Hash())
 	assert.Equal(t, commits[1].Message(), "second commit")
 	assert.Contains(t, commits[1].Parents(), commits[2].Hash())
 	assert.Equal(t, commits[2].Message(), "initial commit")
 	assert.Empty(t, commits[2].Parents())
+
+	t.Run("default_behavior", func(t *testing.T) {
+		var commitsFromDefaultBranch []git.Commit
```
bufdev, line 84: why are these not parallel?

private/pkg/git/git.go
unmultimedio: Even if not exported? I'll move them.

private/pkg/git/repository_test.go
unmultimedio: See comment in `.golancgi.yaml` [here](https://github.com/bufbuild/buf/pull/2371/files#diff-6179837f7df53a6f05c522b6b7bb566d484d5465d9894fb04910dd08bb40dcc9R216), it's because the git repo has a single object reader, not safe to use concurrently. I'll file a ticket to do that improvement.

bufdev: got it

bufdev (approved):

PR #2374, opened 2023-08-11
Syncer lookback to attach older tags

unmultimedio:
When syncing each branch, once all the syncable commits are calculated, add a new step to do a lookback process for each module to attach tags from recent commits.

The lookback process will stop when two conditions are met: at least 5 commits and at least 24 hours back; or when there are no more commits to lookback.

Once gathered those commits to backfill, it will send the tags present in those commits, to the existing synced git hashes.


private/buf/bufsync/bufsync.go
```diff
@@ -270,6 +280,26 @@ type ModuleDefaultBranchGetter func(
...
+// point for such module. The Syncer assumes that the "old" commit is already synced, so it will
+// attempt to attach existing tags using that git hash, in case they were recenly created or moved
+// there.
+//
+// A common scenario is SemVer releases: a commit is pushed to the default Git branch, the sync
+// process triggers and completes, and some minutes later that commit is tagged "v1.2.3". The next
+// time the sync command runs, this attacher would pick such tag and attach it to the correct BSR
+// commit.
+//
+// It's expected to return the BSR commit name to which the tags were attached.
```
doriable, line 293: Nit: instead of `OldTagsAttacher`, can we call this something like `BackfillTagsAttacher`?

```suggestion
// OldTagsAttacher is invoked when a commit with valid modules is found within a lookback threshold past the start sync
// point for such module. The Syncer assumes that the "old" commit is already synced, so it will
// attempt to attach existing tags using that git hash, in case they were recently created or moved
// there.
//
// A common scenario is SemVer releases: a commit is pushed to the default Git branch, the sync
// process triggers and completes, and some minutes later that commit is tagged "v1.2.3". The next
// time the sync command runs, this attacher would pick such tag and attach it to the correct BSR
// commit.
//
// It's expected to return the BSR commit name to which the tags were attached.
```

private/buf/bufsync/syncer.go
```diff
@@ -629,6 +653,142 @@ func (s *syncer) readModuleAt(
 	return builtModule, nil
 }
 
+// attachOlderTags takes syncable commits for a branch already calculated, and looks back for each
+// module a given amount of commits or timestamps, syncing tags in case they were created or moved
+// after those commits were synced.
+func (s *syncer) attachOlderTags(ctx context.Context, branch string, syncableCommits []*syncableCommit) error {
```
doriable, line 659: Nit: similarly with naming, can we call this `backfillTags`?

private/buf/bufsync/syncer.go
```diff
@@ -629,6 +653,142 @@ func (s *syncer) readModuleAt(
...
+// - mod1: commit a (first appearance)
+// - mod2: commit b (first appearance)
+// - mod3: commit c (first appearance)
+// - mod4: commit d (HEAD, did not appear)
+func (s *syncer) lookbackStartingPoints(ctx context.Context, branch string, syncableCommits []*syncableCommit) (map[string]git.Hash, error) {
+	branchModulesForSync, ok := s.branchesToModulesForSync[branch]
+	if !ok || len(branchModulesForSync) == 0 {
+		// branch should not be synced, or no modules to sync in that branch
+		return nil, nil
+	}
```
doriable, line 760: We already do this check one layer above, so we should not need to do this again?

private/buf/bufsync/syncer.go
```diff
@@ -629,6 +653,142 @@ func (s *syncer) readModuleAt(
...
 
+// attachOlderTags takes syncable commits for a branch already calculated, and looks back for each
+// module a given amount of commits or timestamps, syncing tags in case they were created or moved
+// after those commits were synced.
+func (s *syncer) attachOlderTags(ctx context.Context, branch string, syncableCommits []*syncableCommit) error {
+	branchModulesForSync, ok := s.branchesToModulesForSync[branch]
+	if !ok || len(branchModulesForSync) == 0 {
+		// branch should not be synced, or no modules to sync in that branch
+		return nil
+	}
```
doriable, line 664: Should this check be done one layer up in `syncBranch`?

unmultimedio: We can add it there, yes, although I think it's no harm to add it as well in the inner functions, in case those are reused sometime later by another func.

private/buf/bufsync/syncer.go
unmultimedio: Midway renaming funcs and vars, I realized the RPC is already called `AttachGitTags`, so probably that's why I chose that `attachOlderTags` naming. I'm going to rename all and let me know if it reads better even with the RPC named like that.

private/buf/bufsync/syncer.go
doriable: Ah yeah, I see -- and checking the map is pretty cheap. I think this is reasonable :)

doriable (approved):
This seems reasonable to me, and I find the rename to be much easier to grok, thank you :) (before, it wasn't entirely clear if it was "attaching old tags" vs. "attaching tags to old commits", whereas I think this provides a lot more clarity that it's new tags on existing commits).

private/buf/bufsync/syncer.go
unmultimedio: https://github.com/bufbuild/buf/pull/2374/files#r1294065870

doriable (approved):

PR #2377, opened 2023-08-14
Implement Type-Matching Lint Rule for `protovalidate`

elliotmjackson:
This pull request introduces a lint rule to ensure that `protovalidate` rules match the type of the field they are applied to. As discussed in issue bufbuild/protovalidate#28 there's a need to support linting protovalidate options as a first-class citizen of the buf ecosystem. 

**Further rules to follow, PR stands to establish implementation pattern.**

**Changes:**

1. **Rule Definition:** Implemented the type-matching lint rule in `private/bufpkg/bufcheck/buflint/internal/buflintv1/vars.go`.
    - The rule checks that an `int32` field, for example, does not get paired with string rules.

2. **Test Cases:** Added relevant test cases in `private/bufpkg/bufcheck/buflint/buflint_test.go`.
    - Cases include various mismatches, such as applying string rules to int32, int64 rules to string, etc.
    - Ensured all test cases passed successfully.

**TODO:**

- [ ] **Documentation Update:** Update the `buf lint` documentation to include this new lint rule and its description.
- [x] **proto update to v0.4.2**: Since this ticket has been open theres been a few breaking changes to the protovalidate protos. 

**Motivation:**

With the growing complexity of proto configurations, ensuring that proto fields are matched with appropriate validation rules becomes vital. This lint rule will streamline the development process by automatically detecting and flagging any mismatches, eliminating potential runtime errors.

Feedback from the community highlighted the importance of this feature. We are excited to bring this enhancement to `buf`.

Resolves https://github.com/bufbuild/buf/pull/2386


private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,96 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+		return nil, fmt.Errorf("could not convert expected validate.disabled into bool, was %T", extension)
+	}
+	data, err := proto.Marshal(fieldConstraints)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal expected validate.disabled into bytes: %v", err)
+	}
+	return data, nil
+}
+
+func getType(in protoreflect.Name) descriptorpb.FieldDescriptorProto_Type {
```
elliotmjackson: I'm very interested in advice on this, this feels too hand crafted but i dont know another solution

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -934,6 +934,29 @@ func TestCommentIgnoresCascadeOn(t *testing.T) {
 	)
 }
 
+func TestValidate(t *testing.T) {
+	t.Parallel()
+	t.Run("validate_rules_types_match", func(t *testing.T) {
```
bufdev: This `t.Run` has no effect - this just makes a subtest when there's only one test, and the rest of the tests in this file use a one-method-per-test strategy, so this is redundant.

Note: fixed this in a push.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -934,6 +934,29 @@ func TestCommentIgnoresCascadeOn(t *testing.T) {
 	)
 }
 
+func TestValidate(t *testing.T) {
```
bufdev: Should be named after the rule similar to other tests.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
 	}
 	return nil
 }
+
+var (
+	// fieldConstraintsDesc provides a Descriptor for validate.FieldConstraints.
+	fieldConstraintsDesc = (*validate.FieldConstraints)(nil).ProtoReflect().Descriptor()
+
+	// fieldConstraintsOneOfDesc provides the OneOfDescriptor for the type union
```
bufdev: This is named `Oneof` everywhere else, including the function it is calling, `Oneof` should be preferred over `OneOf`. Same concern for variables below.

Note: fixed this in a push.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+func getDataByExtension(field protosource.OptionExtensionDescriptor, extensionType protoreflect.ExtensionType) ([]byte, error) {
+	extension, ok := field.OptionExtension(extensionType)
+	if !ok {
+		return nil, nil
+	}
+	fieldConstraints, ok := extension.(*dynamicpb.Message)
+	if !ok {
+		return nil, fmt.Errorf("unexpected extension type for field got %T", extension)
+	}
+	data, err := proto.Marshal(fieldConstraints)
```
bufdev, line 1018: Why do we need to marshal (and the unmarshal) here? I would think we could return `*dynamicpb.Message` as `proto.Message` directly, and then use it for i.e. calling `ProtoReflect()`, is that not possible?

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+	if err != nil {
+		return nil, fmt.Errorf("marshal error for field: %v", err)
+	}
+	return data, nil
+}
+
+func getTypeName(in protoreflect.Name) descriptorpb.FieldDescriptorProto_Type {
```
bufdev, line 1025: This isn't actually returning the type name (as the function implies), this is returning the `FieldDescriptorProto_Type` given the field name specific to protovalidate's library. When I first saw this I thought of it as something more generic, but it just happens that the field names match up with values in `FieldDescriptorProto_Type`. If we want a convenience function like this, it's something more like `getFieldDescriptorProtoTypeForConstFieldName`.

"Type name" typically refers to the message, enum, or group name for a given field: https://github.com/bufbuild/buf/blob/54ff2e4ebd2d0aa5b31010f22550d60649713cd4/private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go#L346

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+			return fmt.Errorf("unmarshal error for field %q: %v", field.FullName(), err)
+		}
+		constraintsRefl := constraints.ProtoReflect()
+		setOneOf := constraintsRefl.WhichOneof(fieldConstraintsOneOfDesc)
+		if setOneOf == nil {
+			continue
+		}
+
+		got := field.Type()
+		expected := getTypeName(setOneOf.Name())
```
bufdev: This is where the function name gets confusing - since this isn't the type name, and is instead a `FieldDescriptorProto_Type`, I think this needs a bit more explanation. This is relying on the fact that the field names in protovalidate happen to line up with `FieldDescriptorProto_Types` - which is fine if we are spelling out that this is our assumption, but I think this needs some code comments and better variable naming if possible.

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match/a.proto
```diff
@@ -0,0 +1,26 @@
+syntax = "proto3";
+
+package a;
+
+import "buf/validate/validate.proto";
+
+message Foo {
```
bufdev, line 7: What is the linting we want to do for non-primitives (`enum, repeated, map, any, duration, timestamp`)? This rule doesn't take those into account.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
 	return nil
 }
+
+var (
+	// fieldConstraintsDesc provides a Descriptor for validate.FieldConstraints.
+	fieldConstraintsDesc = (*validate.FieldConstraints)(nil).ProtoReflect().Descriptor()
+
+	// fieldConstraintsOneOfDesc provides the OneOfDescriptor for the type union
+	// in FieldConstraints.
+	fieldConstraintsOneOfDesc = fieldConstraintsDesc.Oneofs().ByName("type")
```
bufdev: Is this not better named `fieldConstraintsTypeOneofDesc`?

Note: fixed this in a push.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
bufdev: `type oneof`, not `type union`.

Note: fixed this in a push.

private/pkg/protosource/option_extension_descriptor.go
```diff
@@ -39,9 +39,23 @@ func (o *optionExtensionDescriptor) OptionExtension(extensionType protoreflect.E
 	if extensionType.TypeDescriptor().ContainingMessage().FullName() != o.message.ProtoReflect().Descriptor().FullName() {
 		return nil, false
 	}
+
 	if !proto.HasExtension(o.message, extensionType) {
 		return nil, false
 	}
+
+	if extensionType.TypeDescriptor().Kind() == protoreflect.MessageKind {
```
bufdev, line 47: Can you explain (and in comments too) what is going on here/the motivation here a bit more?

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match/buf/validate/expression.proto
```diff
@@ -0,0 +1,92 @@
+// Copyright 2023 Buf Technologies, Inc.
```
bufdev, line 1: As opposed to needing to vendor this in every rule, in https://github.com/bufbuild/buf/pull/2033 I proposed we add a new helper function `testLintWithValidate`, that will sideload the protovalidate files into the bucket for the module. You'll likely then end up storing the validate files just once, and outside the specific sub-directory for the given rule, but for extra credit you could curl and cache the module, similar to what we do for googleapis in this repository - search around this repository for googleapis to see prior art.

private/bufpkg/bufcheck/buflint/internal/buflintbuild/buflintbuild.go
```diff
@@ -348,6 +348,12 @@ var (
 		"all files have a syntax specified",
 		newAdapter(buflintcheck.CheckSyntaxSpecified),
 	)
+	// ValidateRulesTypesMatchRuleBuilder is a rule builder.
+	ValidateRulesTypesMatchRuleBuilder = internal.NewNopRuleBuilder(
+		"VALIDATE_RULES_TYPES_MATCH",
```
bufdev, line 353: I don't have a great answer just yet and need to sleep on it, but I don't think this is the best name for this rule - need to brainstorm a bit as to what this rule specifically is and what the name should be. Just a note.

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/bufbuild/buf
 go 1.19
 
 require (
+	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.31.0-20230802163732-1c33ebd9ecfa.1
```
bufdev, line 6: I think we've said that we didn't want the buf CLI directly depending on remote packages due to the circular dependency issue that can introduce w.r.t. bootstrapping the buf toolchain - @rodaine do you have any opinions here?

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -963,3 +968,95 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+		constraintsRefl := constraints.ProtoReflect()
+		setOneOf := constraintsRefl.WhichOneof(fieldConstraintsOneOfDesc)
+		if setOneOf == nil {
+			continue
+		}
+
+		got := field.Type()
+		expected := getTypeName(setOneOf.Name())
+		if got != expected {
+			add(field, field.Location(), nil, "constraint type mismatch on %q, expected %q but got %s", field.FullName(), expected, got)
```
bufdev, line 1003: This should be using the `OptionExtensionLocation` that I believe @jhump somewhat-recently introduced, see the comment on https://github.com/bufbuild/buf/pull/2033.

bufdev (commented):
I pushed fixes for some of the lowest-rent comments to expedite - I was reviewing locally so it seemed easiest.

elliotmjackson:
@bufdev, this PR is still in draft mode and this is still a work in progress so isn't ready for review. I'll make sure to let you know as soon as it’s in a more finalized state and ready for your input. Thanks

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: this rule requires both positive and negative states to be validated to ensure that matched types are not being blocked. More tests coming before pr is ready for review

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
elliotmjackson: I dont like the string compare here, its overly complex and unstable when it comes to data structures and wkts. I anticipate this function will not be here by the time this pr is ready for review

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match/a.proto
elliotmjackson: these have not been implemented yet, this will be completed before the PR is ready for review

private/pkg/protosource/option_extension_descriptor.go
elliotmjackson: sure, somewhat of a stop-gap solution to prevent dynamic type mismatching panics. Before the PR is ready i will revisit this.

bufdev:
Note for reference: in bufbuild, all PRs put up are open for feedback regardless of state - if feedback isn't desired, keep as a branch without a PR.

elliotmjackson:
Understood, this PR stands to improve transparency. I'm always looking for feedback, i'd hate for you to waste your time committing to or commenting on unfinished work.

private/bufpkg/bufcheck/buflint/internal/buflintbuild/buflintbuild.go
elliotmjackson: modified to `VALIDATE_CONSTRAINTS_CHECK` - all ears for anything better. We also have the option of breaking up the rules for more granular control. I've document all possible rules [here](https://www.notion.so/bufbuild/protovalidate-first-class-buf-lint-support-d96aaa1815694b85b5e6696beded8f55?pvs=4)

go.mod
rodaine: My primary concern is the BSR depending on the BSR; less so concerned about the CLI. Right now, adding this dep to the CLI will currently make it a dependency of the BSR codebase.

As for these options, either has different tradeoffs in terms of complexity in order to avoid the circular deps through the BSR. Having the concrete types present will make the code easier to write and prevent runtime issues while the underlying PV types are still in a bit of flux.

bufdev: I think overall, it sounds like we should remove this dependency. Generating these deps within `private/gen` doesn't seem to be a major issue, unless I'm missing something here - let me know.

private/buf/cmd/buf/buf_test.go
```diff
@@ -522,6 +522,7 @@ RPC_REQUEST_RESPONSE_UNIQUE       DEFAULT                  Checks that RPC reque
 RPC_REQUEST_STANDARD_NAME         DEFAULT                  Checks that RPC request type names are RPCNameRequest or ServiceNameRPCNameRequest (configurable).
 RPC_RESPONSE_STANDARD_NAME        DEFAULT                  Checks that RPC response type names are RPCNameResponse or ServiceNameRPCNameResponse (configurable).
 SERVICE_SUFFIX                    DEFAULT                  Checks that services are suffixed with Service (suffix is configurable).
+VALIDATE_CONSTRAINTS_CHECK        DEFAULT                  Checks that fields contain constraints which match the same type.
```
bufdev, line 525: Re: previous comment, I think we still need to brainstorm on this a bit. Our other lint checks do not include `_CHECK` in them - I'd like to see some options/a brainstorm with respect to the naming here, can you provide that?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -459,29 +460,22 @@ func TestRunPackageLowerSnakeCase(t *testing.T) {
 
 func TestRunPackageNoImportCycle(t *testing.T) {
 	t.Parallel()
-	testLintWithModifiers(
```
bufdev, line 462: There should be no diff here - this function was formatted the same as the rest of this file, and there's no change here with respect to this PR.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -889,14 +883,9 @@ func TestCommentIgnoresOff(t *testing.T) {
 
 func TestCommentIgnoresOn(t *testing.T) {
 	t.Parallel()
-	testLintWithModifiers(
```
bufdev, line 892: Same, there should be no diff here.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -942,17 +931,104 @@ func TestCommentIgnoresCascadeOff(t *testing.T) {
 
 func TestCommentIgnoresCascadeOn(t *testing.T) {
 	t.Parallel()
+	testLintWithModifiers(t, "comment_ignores_cascade", func(config *bufconfig.Config) {
+		config.Lint.AllowCommentIgnores = true
+	}, nil, "")
+}
+
+func TestValidateRulesTypeDontMatch(t *testing.T) {
```
bufdev, line 939: The remainder of our test functions match the lint rule name, can you stick to this style?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -942,17 +931,104 @@ func TestCommentIgnoresCascadeOff(t *testing.T) {
...
+	t.Parallel()
+	testLintWithValidate(t, "validate_rules_types_match_fail",
```
bufdev, line 941: Can you stick to the same style by putting everything on newlines?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -942,17 +931,104 @@ func TestCommentIgnoresCascadeOff(t *testing.T) {
...
+}
+
+func TestValidateRulesTypeMatch(t *testing.T) {
+	t.Parallel()
+	testLintWithValidate(t, "validate_rules_types_match_success")
+}
+
+func TestValidateRulesTypesMatchUsageErrors(t *testing.T) {
+	t.Parallel()
+	testLintWithValidate(t, "validate_rules_types_match_errors",
```
bufdev, line 981: Same comments about newlines and about directory naming

private/bufpkg/bufcheck/buflint/buflint_test.go
bufdev, line 941: The rest of our testdata is in a directory named after the rule - can you stick to this pattern and make sure that all testdata subdirectories are named `name_of_rule` + potentially `_additional_context_needed_for_specific_test`?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -942,17 +931,104 @@ func TestCommentIgnoresCascadeOff(t *testing.T) {
...
+		bufanalysistesting.NewFileAnnotation(t, "a.proto", 39, 49, 39, 90, "VALIDATE_CONSTRAINTS_CHECK"),
+		bufanalysistesting.NewFileAnnotation(t, "a.proto", 40, 45, 40, 82, "VALIDATE_CONSTRAINTS_CHECK"),
+		bufanalysistesting.NewFileAnnotation(t, "a.proto", 41, 49, 41, 86, "VALIDATE_CONSTRAINTS_CHECK"),
+		bufanalysistesting.NewFileAnnotation(t, "a.proto", 42, 47, 42, 84, "VALIDATE_CONSTRAINTS_CHECK"),
+	)
+}
+
+func TestValidateRulesTypeMatch(t *testing.T) {
+	t.Parallel()
+	testLintWithValidate(t, "validate_rules_types_match_success")
```
bufdev, line 976: Same comment re directory naming

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -986,6 +1064,48 @@ func testLintWithModifiers(
 	)
 	require.NoError(t, err)
 
+	var validateReadWriteBucket storage.ReadWriteBucket
```
bufdev, line 1067: Based on how this function is laid out, this bucket is not specific to validate anymore - this variable naming confused me, can you update?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -962,6 +1038,7 @@ func testLint(
 		relDirPath,
 		nil,
 		nil,
+		"buf",
```
bufdev, line 1041: This should be in a deeper subdirectory to make sure it is clear to future developers what this is for - I'd suggest `deps/protovalidate/buf/validate/validate.proto`, with this value being changed to `deps/protovalidate`.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -986,6 +1064,48 @@ func testLintWithModifiers(
...
 
+	var validateReadWriteBucket storage.ReadWriteBucket
+	if dependencyPathPrefix != "" {
+		validateReadWriteBucket, err = storageosProvider.NewReadWriteBucket(
+			"testdata",
+			storageos.ReadWriteBucketWithSymlinksIfSupported(),
+		)
+		require.NoError(t, err)
+
+		// Copy the testdata directory to a temporary directory
```
bufdev, line 1075: Why are we copying it? This should just be a read-only thing, we should be able to just make a union bucket via `storage.MultiReadBucket` between the `readWriteBucket` and the additional dependency, and then pass this union bucket to `BuildForBucket`.

If we did/do need to copy, there is `storage.Copy`.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -986,6 +1064,48 @@ func testLintWithModifiers(
...
+				return err
+			}
+			defer put.Close()
+			_, err = put.Write(expectedData)
+			if err != nil {
+				return err
+			}
+			return nil
+		}
+		if err := validateReadWriteBucket.Walk(ctx, dependencyPathPrefix, fn); err != nil {
```
bufdev, line 1097: This all would be replaced with `storage.Copy`, but as mentioned, I don't think we need to (or should be) copying any data.

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -986,6 +1064,48 @@ func testLintWithModifiers(
...
+			require.NoError(t, err)
+		}
+
+		// Delete the temporary directory
+		defer func() {
+			if err := readWriteBucket.DeleteAll(ctx, dependencyPathPrefix); err != nil {
```
bufdev, line 1103: This looks to be a bug - this is deleting from the original `readWriteBucket`, which is generally on the path `testdata/name_of_rule` via `dirPath`.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -972,3 +977,48 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
 	}
 	return nil
 }
+
+// CheckValidateConstraintsCheck is a check function.
+var CheckValidateConstraintsCheck = newFilesWithImportsCheckFunc(checkValidateConstraintsCheck)
+
+func checkValidateConstraintsCheck(add addFunc, files []protosource.File) error {
+	m := buflintvalidate.NewValidateModule(files, add)
```
bufdev, line 985: Needs a descriptive name for the variable, such as `validateModule`, however I don't think `validateModule` is needed in this case.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -972,3 +977,48 @@ func checkSyntaxSpecified(add addFunc, file protosource.File) error {
...
+		for _, message := range file.Messages() {
+			for _, field := range message.Fields() {
+				data, ok, err := getDataByExtension(field, validate.E_Field)
+				if err != nil {
+					return err
+				}
+				if !ok {
+					continue
+				}
+				var constraints validate.FieldConstraints
```
bufdev, line 996: `constraints := &validate.FieldConstraints{}`

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
bufdev: Can we add some comments as to what this function is doing, so future readers can quickly get context? I know the private functions of `bufcheck` aren't the best-documented, but this function is particularly confusing at least to me as an external reader. Some comments about why we are doing the marshal/unmarshal as well would be helpful.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/buflintvalidate.go
```diff
@@ -0,0 +1,27 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buflintvalidate
+
+import "github.com/bufbuild/buf/private/pkg/protosource"
+
+func NewValidateModule(
```
bufdev, line 19: This type isn't needed, and IMO actually detracts from clarity. We're returning a private type, meaning the Godoc attached to the single (public) method won't be displayed, and we're not actually saving anything here - you could just have a `ValidateField` public interface (with documentation), with a `NewValidateField`. `*validateModule` is just storing two parameters, and doesn't store any state - this doesn't change usability in the `check` function. Going even further, I think you could just have a function `CheckFieldRules` that takes everything, as `*validateField` isn't that necessary either, but this is more a nit.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,53 @@
...
+
+	m.assertf(in == 0 ||
+		lt.IsNil() && lte.IsNil() &&
+			gt.IsNil() && gte.IsNil(),
+		"cannot have both `in` and range constraint rules on the same field",
+	)
+
+	if !lt.IsNil() {
+		m.assertf(gt.IsNil() || !reflect.DeepEqual(lti, gti),
+			"cannot have equal `gt` and `lt` rules on the same field")
```
bufdev, line 44: We haven't used backticks in any of our other lint rules - generally, if we need to quote something, we use double-quotes. I won't comment about this in rest of PR, but obviously take care of it globally.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"reflect"
+)
+
+// validateNumberField asserts that the given numbers are valid.
+func validateNumberField[T any](m *validateField, in, notIn int, ci, lti, ltei, gti, gtei T) {
+	m.checkIns(in, notIn)
+
+	c := reflect.ValueOf(ci)
```
bufdev, line 25: Please use descriptive variable names - it's not apparent to me what `c` or `ci` are.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/validate_field.go
```diff
@@ -0,0 +1,379 @@
...
+)
+
+// validateField is a validate Field.
+type validateField struct {
+	module   *validateModule
+	field    protosource.Field
+	location protosource.Location
+}
+
+// newValidateField returns a new validate validateField.
```
bufdev, line 51: None of these comments are public - they seem unnecessary, and inconsistent with our codebase

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
```diff
@@ -0,0 +1,87 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buflintvalidate
+
+import "strings"
+
+// WellKnownTypePackage is the proto package name where all Well Known Types
+// currently reside.
+const WellKnownTypePackage string = "google.protobuf."
```
bufdev, line 21: Why is this public? None of these appear to be used outside of this package. The only thing used outside of this package is `NewValidateModule` (soon to be `NewValidateField`), nothing else should be public.

This should also be in the same block as the rest of the consts.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
```diff
@@ -0,0 +1,87 @@
...
+	Int64ValueWKT  WellKnownType = "Int64Value"
+	UInt64ValueWKT WellKnownType = "UInt64Value"
+	Int32ValueWKT  WellKnownType = "Int32Value"
+	UInt32ValueWKT WellKnownType = "UInt32Value"
+	BoolValueWKT   WellKnownType = "BoolValue"
+	StringValueWKT WellKnownType = "StringValue"
+	BytesValueWKT  WellKnownType = "BytesValue"
+)
+
+var lookupTable = map[string]WellKnownType{
```
bufdev, line 53: `stringToWellKnownType`, similar to rest of codebase

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
```diff
@@ -0,0 +1,87 @@
...
+	"Int32Value":  Int32ValueWKT,
+	"UInt32Value": UInt32ValueWKT,
+	"BoolValue":   BoolValueWKT,
+	"StringValue": StringValueWKT,
+	"BytesValue":  BytesValueWKT,
+}
+
+// LookupWellKnownType returns the WellKnownType related to the provided Name. If the
+// name is not recognized, UnknownWKT is returned.
+func LookupWellKnownType(in string) WellKnownType {
```
bufdev, line 74: Doesn't need to be public

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match_errors/buf.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+lint:
+  use:
+    - VALIDATE_CONSTRAINTS_CHECK
+  ignore:
+    - buf
+
```
bufdev, line 7: Nit: extra newline

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match_success/buf.yaml
```diff
@@ -0,0 +1,7 @@
+version: v1
+lint:
+  use:
+    - VALIDATE_CONSTRAINTS_CHECK
+  ignore:
```
bufdev, line 5: Nit: optimally, `testLintWithModifiers` would automatically add an `ignore` for the dependency directory it added, which can be done via `configModifier`.

private/pkg/protosource/field.go
```diff
@@ -107,8 +107,40 @@ func newField(
 	}
 }
 
-func (f *field) Message() Message {
-	return f.message
+func (f *field) ParentMessage() Message {
+	return f.parentMessage
+}
+
+func (f *field) Embed(files ...File) Message {
```
bufdev, line 114: Why is this public? This function worries me - why are we taking a list of `Files` into a `Field`? This feels suceptible to user error.

I'd recommend we use the solution we discussed offline and not add this functionality in this way to `protosource`.

bufdev (changes requested):

private/buf/cmd/buf/buf_test.go
elliotmjackson: 1. `VALIDATE_CONSTRAINTS_CONSISTENCY`
1. `VALIDATE_VALIDATION_INTEGRITY`
1. `VALIDATE_FIELD_RULES_CHECK`
1. `VALIDATE_CONFORMANCE_ALL`
1. `VALIDATE_RULES_ALIGNMENT`
1. `VALIDATE_SYNTAX_HARMONY`
1. `VALIDATE_FIELD_CONSISTENCY`
1. `VALIDATE_RULES_COHERENCE`
1. `VALIDATE_VALIDATION_CORRECTNESS`
1. `VALIDATE_STRUCTURE_ALIGNMENT`

bufdev: Can you provide your thinking behind each and how they align with other rules? As an example, (3) continues to use the `CHECK` pattern

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: attempted to resolve in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e). lmk if you want different names

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e). Named `dependencyReadWriteBucket`

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: resolved in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/buflint_test.go
elliotmjackson: removed in [1b2c71c](https://github.com/bufbuild/buf/pull/2377/commits/1b2c71cceeb54702ea58bc69970a1a434fb6410e)

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
elliotmjackson: will revisit with conclusion of https://github.com/bufbuild/buf/pull/2377#discussion_r1305931404 and remove validateModule because of https://github.com/bufbuild/buf/pull/2377#discussion_r1305954165

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/buflintvalidate.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62). This now raises a problem with 2 style rules:
1. `buflintvalidate.go` must exist for the package but has no natural contents
2. `validate_field.go` must contain struct definition, constructors and methods. 

This is likely a smell because it doesn't conform to the style guide correctly. Let's consider actioning your nit on `validateField` this would expand the footprint of each function signature significantly - do you hate that more or less than the validateField methods?

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/util.go
elliotmjackson: resolved in [a3c77f8](https://github.com/bufbuild/buf/pull/2377/commits/a3c77f8869438dc13611808c7e3d10bcdbf8dc1f) and [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/util.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/validate_field.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
elliotmjackson: 100% oversight - thanks for the catch. resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/well_known_type.go
elliotmjackson: resolved in [182d2e2](https://github.com/bufbuild/buf/pull/2377/commits/182d2e21541d6084e75fd345150fa990cde60d62)

private/pkg/protosource/field.go
elliotmjackson: I think I've misunderstood our offline discussion - let me close out everything else and come back to this point.

private/bufpkg/bufcheck/buflint/testdata/validate_rules_types_match_errors/buf.yaml
elliotmjackson: resolved in [a3c77f8](https://github.com/bufbuild/buf/pull/2377/commits/a3c77f8869438dc13611808c7e3d10bcdbf8dc1f)

private/buf/cmd/buf/buf_test.go
elliotmjackson: github has split the thread, linking [here](https://github.com/bufbuild/buf/pull/2377#discussion_r1304635029) for clarity

private/pkg/protosource/field.go
bufdev:
```func IsFieldAMap(field Field, fullNameToMessage map[string]Message) bool {
	// For this to be a map field, it must be a repeated field
	// with a synthetic message as the type, that synthetic message
	// must be in the enclosing message for the field, and it must
	// have map_entry set to true.
	if field.Label() != descriptorpb.FieldDescriptorProto_LABEL_REPEATED {
		return false
	}
	if field.Type() != descriptorpb.FieldDescriptorProto_TYPE_MESSAGE {
		return false
	}
	message, ok := fullNameToMessage[field.TypeName()]
	if !ok {
		return false
	}
	return message.IsMapEntry()
}```

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/buflintvalidate.go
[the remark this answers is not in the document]
bufdev: When all other comments are addressed and this PR is otherwise good to go, we can revisit the name of the rule last

private/pkg/protosource/field.go
elliotmjackson: resolved in [ae405eb](https://github.com/bufbuild/buf/pull/2377/commits/ae405eb097ab57139d9f22eea4ec346740c1b3ea)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/buflintvalidate.go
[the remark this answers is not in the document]
elliotmjackson: Sorry this comment is super unclear. I was referring to the file `buflintvalidate.go`. the style guide wants the "most important" part of the package in here. Which would be `validateField` and its methods but that then breaks another rule requiring it be in its own snake cased file. Which one takes preference?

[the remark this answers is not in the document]
bufdev: That's not actually what the style guide says - the style guide says that if the package is small enough (which this one isn't), put everything in this file, otherwise put the public types in this file and lay out packages as you do otherwise, in this case with files named after their types.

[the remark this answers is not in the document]
elliotmjackson: resolved in [a5aeaa5](https://github.com/bufbuild/buf/pull/2377/commits/a5aeaa5a16b925b6ce59996c7e45ac455b380945)

PR #2379, opened 2023-08-16
Add duplicated dependency sentinel error

unmultimedio:
Add a sentinel for duplicated dependency error, so in the BSR we can determine if this error happens when validating a pushed module, and we can return a 4xx instead of a 5xx.


private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
```diff
@@ -168,3 +169,9 @@ func WithExcludePathsAllowNotExist(excludePaths []string) BuildOption {
 		buildOptions.pathsAllowNotExist = true
 	}
 }
+
+// IsErrDuplicatedDependency checks if the given error is `errDuplicatedDependency`, from the build
+// process.
+func IsErrDuplicatedDependency(err error) bool {
```
unmultimedio, line 175: Open for suggestions on naming, this seemed appropriate.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
```diff
@@ -121,7 +123,6 @@ func (m *moduleFileSetBuilder) build(
 			}
 			if _, ok := hashes[potentialDependencyModuleHash]; !ok {
 				dependencyModules = append(dependencyModules, potentialDependencyModule)
-			} else {
 				hashes[potentialDependencyModuleHash] = struct{}{}
 			}
```
unmultimedio, line 127: This seems like something that we missed [here](https://github.com/bufbuild/buf/pull/2232/files#diff-b267002b1d5ca9c7c22a25c7f3016a781a0322ba50822036c0e6aebf3f2d9fd8R122-R126). `hashes` is of type `map[string]struct{}`, doesn't make sense to add a key to the map when it's already there. Probably we wanted to append to `dependencyModules` and also mark it as present in the `hashes` map, but I'm not 100% sure. Pls confirm @bufdev

private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
bufdev: Call the error `ErrDuplicateDependency`, and just make it public so people can use `errors.Is`

unmultimedio: If the error is public `var err = errors.New(...)`, isn't that leaving it open for users to modify it? That's why I didn't exported it, just like in [here](https://github.com/bufbuild/buf/blob/3727d810c3f1de8bc1b39c7bc53b4b9fc6b83c38/private/pkg/storage/errors.go#L40-L43)... although that pkg also has other exported errors.

I'll export it anyway.

bufdev: I think by convention, public errors are expected not to be edited - even in that same file, we export some errors. The `IsNotExist` function is really to match `os.IsNotExist`, which a lot of people have traditionally used.

I actually like the `Is` pattern, and we have used it in `buf` in the past a decent amount, but my impression is `errors.Is` is the golang-idiomatic way forward nowadays.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
bufdev: yea this seems like a mistake on my part, good catch

bufdev (approved):

PR #2384, opened 2023-08-17
Use connect error message for AlreadyExists when pushing

unmultimedio:
Originally reported by @seankimdev, the CLI does not differentiate `AlreadyExists` errors, which right now could happen either because we're trying to push the same content, or because we're pushing a tag that already exists. It always assumes it's a duplicated content. This PR proposes using the error from the BSR response, which disambiguates it.

```sh
# clean push
$ echo "//\n" >> foo/v1/foo.proto && \
    buf push
b9c6867fdebb4bb5a7d61dd0364eedc3

# pushing the same content
$ buf push         
already_exists: new commit identical to latest on branch

# pushing new content with an existing tag
$ echo "//\n" >> foo/v1/foo.proto && \
    buf push --tag eb7195bd670182329c964cdb05ecd5cd2baf543d
already_exists: cannot create tag with name eb7195bd670182329c964cdb05ecd5cd2baf543d
```


bufdev:
> already_exists: cannot create a new label with name eb7195bd670182329c964cdb05ecd5cd2baf543d

As you mentioned, we currently do not have any concept of label as far as users are concerned, so this error message would confuse me (I think that answers your question?)

Message changes aren't considered breaking

unmultimedio:
We have fixed the BSR error message to match the expected "tag" nomenclature. Once deployed I'll update this PR description.

seankimdev (approved):

unmultimedio:
Updated error message in the PR description, no more `labels` reference.

doriable (approved):

private/buf/cmd/buf/command/push/push.go
```diff
@@ -193,7 +193,7 @@ func run(
 	if err != nil {
 		if connect.CodeOf(err) == connect.CodeAlreadyExists {
 			if _, err := container.Stderr().Write(
-				[]byte("The latest commit has the same content; not creating a new commit.\n"),
+				[]byte(fmt.Sprintf("%s\n", err.Error())),
```
bufdev, line 196: Why are we special-casing for connect.CodeAlreadyExists at all now?

unmultimedio: Because this error can come from 2 different places: identical content or existing tag, but it's assuming it's always coming from "identical content". See the PR description.

bufdev: I'm saying how is it any different than `return err`

bufdev (approved):

private/buf/cmd/buf/command/push/push.go
unmultimedio: Ah ok, I think it's just making sure to end with a new line.

PR #2386, opened 2023-08-18
Add Utility Functions for Detecting Map Fields from Files and Messages

elliotmjackson:
Add utility functions for checking if a field is a map field using file and message information.

These functions provide the ability to determine whether a given field is a map field based on both file and message contexts. They utilize the provided field and file/message information to determine whether the field is a map entry. The functions also handle the necessary checks for field type and labeling to accurately identify map fields.

The added functions are as follows:
- `isFieldAMapFromFiles`: Checks if the field is a map field using file information.
- `isFieldAMapFromMessages`: Checks if the field is a map field using message information.
- `getMessageFromFiles`: Retrieves the Message associated with a field from a list of files.

These functions enhance the capability to work with map fields in the codebase.

Furthermore, I changed `protosource.Field.Message()` to `ParentMessage()` which better describes the result/responsibility of this function


bufdev:
I believe this PR is fully subsumed by https://github.com/bufbuild/buf/pull/2377, is this correct?

PR #2389, opened 2023-08-21
Add element_type to module element search result

cyinma:


proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -42,6 +42,18 @@ enum OrderBy {
   ORDER_BY_NAME = 3;
 }
 
+enum ElementType {
```
bufdev, line 45: Everything we add needs documentation. Among the things that need to be addressed: What is the purpose of this? Why does this not already exist, and why did we need to add a new type?

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -42,6 +42,18 @@ enum OrderBy {
...
+  ELEMENT_TYPE_UNSPECIFIED = 0;
+  ELEMENT_TYPE_SERVICE = 1;
+  ELEMENT_TYPE_METHOD = 2;
+  ELEMENT_TYPE_MESSAGE = 3;
+  ELEMENT_TYPE_EXTENSION = 4;
```
bufdev, line 50: Can there be some docs on how extensions work with searching? Some examples?

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -42,6 +42,18 @@ enum OrderBy {
...
+  ELEMENT_TYPE_FIELD = 5;
+  ELEMENT_TYPE_ONEOF = 6;
```
bufdev, line 52: Is there some kind of priority given to different element types? Surfacing oneof names alongside message names would concern me.

proto/buf/alpha/registry/v1alpha1/search.proto
cyinma: discussed offline for the comments here

proto/buf/alpha/registry/v1alpha1/search.proto
```diff
@@ -86,6 +96,7 @@ message ModuleElementSearchResult {
   string proto_file_path = 4;
   string fully_qualified_name = 5;
   string package_name = 6;
+  SearchModuleElementType element_type = 7;
```
bufdev, line 99: `search_module_element_type`

bufdev (commented):
Good to go after field rename

mfridman (approved):

bufdev (approved):

PR #2390, opened 2023-08-21
Add integration tests for buf genenerate

oliversun9:
This PR
* fixes a bug where `file.proto: true` in a per-file override in a `buf.gen.yaml` in v1 gets migrated to `value: "true"`
* fixes a bug where in extreme cases in Windows, a plugin path with slashes, such as `node_modules/.bin/protoc-gen-es` (a plugin path in this form does not always cause an error on Windows)
* adds a binary `protoc-gen-plugin-config` next to `protoc-gen-insertion-point-writer` and `protoc-gen-insertion-point-receiver`, in order to test `strategy` (`dir` vs `all`).
* adds integration tests to `private/buf/cmd/buf/command/generate/generate_test.go`.
The rest are `testdata`.


private/buf/cmd/buf/command/alpha/protoc/internal/protoc-gen-plugin-config/main.go
```diff
@@ -0,0 +1,59 @@
+// Copyright 2020-2023 Buf Technologies, Inc.
```
bufdev, line 1: This shouldn't be in `alpha/protoc` - this is for the `buf alpha protoc` command, this should be somewhere for generate

bufdev (approved):

PR #2392, opened 2023-08-22
Add GetRepositoryDependencyDOTString RPC

seankimdev:
New RPC to get dependency graph for a repository. 
Returns a string DOT representation of the dependency graph, which is used to render DOT graph in the BSR UI.


doriable (approved):

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -367,3 +371,19 @@ message GetRepositoriesMetadataRequest {
...
+  string repository = 2;
+  // The name of the reference.
+  string reference = 3;
+}
+
+message GetRepositoryDependencyGraphResponse {
+  // The string DOT representation of the dependency graph for the repository.
+  //
+  // DOT language reference: https://graphviz.org/doc/info/lang.html
+  string repository_dependency_graph_dot = 1;
```
bufdev, line 388: `dot_string`

proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -166,6 +166,10 @@ service RepositoryService {
   rpc GetRepositoriesMetadata(GetRepositoriesMetadataRequest) returns (GetRepositoriesMetadataResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // GetRepositoryDependencyGraph gets the dependency graph for the repository.
+  rpc GetRepositoryDependencyGraph(GetRepositoryDependencyGraphRequest) returns (GetRepositoryDependencyGraphResponse) {
```
bufdev, line 170: I would call this `GetRepositoryDependencyDOTString`, to represent what it is. Not the best name, but we might want to use `GetRepositoryDependencyGraph` when we create a `Graph` type.

proto/buf/alpha/registry/v1alpha1/repository.proto
seankimdev: changed

proto/buf/alpha/registry/v1alpha1/repository.proto
seankimdev: changed

PR #2396, opened 2023-08-25
Add WithAllowV0 to PackageVersion construction

bufdev:
Needed for an external case where we need to be able to parse `v0`'s.


oliversun9 (approved):
LGTM

Issue #2403, opened 2023-08-29
`buf generate <input> --exclude-path="..."` doesn't seem to work for buf Schema Registry package


maoueh:
Using `buf.gen.yaml`:

```
version: v1
plugins:
  - plugin: buf.build/protocolbuffers/go:v1.31.0
    out: gen
```

If I do:

```bash
$ rm -rf gen
$ buf generate buf.build/streamingfast/substreams --exclude-path="sf/substreams/rpc"
...

$ file gen/github.com/streamingfast/substreams/pb/sf/substreams/rpc
gen/github.com/streamingfast/substreams/pb/sf/substreams/rpc: directory
```

Protos under `sf/substreams/rpc` (https://buf.build/streamingfast/substreams/tree/main:sf/substreams/rpc/v2) path are still generated.

If I use a raw Protobuf binary (the one used below is not exactly the same as the buf uploaded version, but both contains `sf/substreams/rpc` as a "path"):

```bash
$ rm -rf gen
$ buf generate "https://github.com/streamingfast/substreams/raw/develop/pb/system/system.pb#format=bin" --exclude-path="sf/substreams/rpc"

$ file gen/github.com/streamingfast/substreams/pb/sf/substreams/rpc
gen/github.com/streamingfast/substreams/pb/sf/substreams/rpc: cannot open `gen/github.com/streamingfast/substreams/pb/sf/substreams/rpc' (No such file or directory)
```

I would have though that `--exclude-path` would work even with bug Schema registry packages.

bufdev:
This should be fixed by #2404 and will go out in the next release. Apologies for the troubles.

maoueh:
No problem and thank you guys!

PR #2404, opened 2023-08-29
Fix module generation with --exclude-path

oliversun9:
This fixes https://github.com/bufbuild/buf/issues/2403.

The problem was that excluded paths were only used as a filter when 'target paths` are specified (the `--path` flag).


bufdev:
Can you grab me at my desk so we can go through this in-person tomorrow? I'm sure it's right, I just want to double check the logic, and easier/faster to do live I think, given I can get some quick explanations.

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 - Update `buf beta price` with the latest pricing information.
+- Fix issue where `buf generate buf.build/foo/bar --exclude-path some/path` does
```
bufdev, line 6: Fix issue where `buf generate --exclude-path` was not properly excluding paths for remote modules.

bufdev (approved):

PR #2419, opened 2023-09-04
Sync from local checkout

unmultimedio:
Remove the constraint that syncs only the content that's been pushed to a remote named "origin".


private/pkg/git/packed_refs.go
```diff
@@ -25,10 +25,10 @@ import (
 const (
 	// packedRefsHeader is the head for the `packed-refs` file
 	// based on https://github.com/git/git/blob/master/refs/packed-backend.c#LL1084C41-L1084C41
-	packedRefsHeader      = "# pack-refs with: peeled fully-peeled sorted "
-	tagRefPrefix          = "refs/tags/"
-	originBranchRefPrefix = "refs/remotes/origin/"
-	unpeeledRefPrefix     = '^'
+	packedRefsHeader  = "# pack-refs with: peeled fully-peeled sorted "
+	tagRefPrefix      = "refs/tags/"
+	branchRefPrefix   = "refs/heads/"
```
unmultimedio, line 30: Packed refs were also reading branches only from `origin` remote.

private/pkg/git/repository.go
```diff
@@ -333,12 +329,17 @@ func (r *repository) readPackedRefs() error {
 	return r.packedReadError
 }
 
+// detectDefaultBranch returns the repository's default branch name. It attempts to read it from the
+// `.git/refs/remotes/origin/HEAD` file, and expects it to be pointing to a branch also in the
+// `origin` remote.
 func detectDefaultBranch(gitDirPath string) (string, error) {
+	const defaultRemoteName = "origin"
```
unmultimedio, line 336: This is the only place in which we keep assuming `origin` remote name. If no explicit default branch is passed, then we read the HEAD of the default remote name (`origin`).

private/pkg/git/git.go
```diff
@@ -274,16 +274,13 @@ type TreeNode interface {
 
 // Repository is a git repository that is backed by a `.git` directory.
 type Repository interface {
-	// DefaultBranch is the default branch of the repository. This is either configured via the
-	// `OpenRepositoryWithDefaultBranch` option, or discovered from the value in
-	// `.git/refs/remotes/origin/HEAD`. Therefore, discovery requires that the repository is pushed to
-	// a remote named `origin`.
+	// DefaultBranch is the default branch of the repository. By default this is reads the value in
```
seankimdev, line 277:
```suggestion
	// DefaultBranch is the default branch of the repository. By default this reads the value in
```

private/pkg/git/repository.go
doriable: This is reasonable to me -- this is more a note for myself, but I think at some point, down the road, we may need to expose this so that it can be used by the BSR in some way, but this is a reasonable way of handling this right now.

doriable (approved):
This looks reasonable to me, I think there is a small documentation nit, but I am approving this for now.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -250,10 +249,10 @@ func sync(
 		}
 		_, err = container.Stderr().Write([]byte(
 			// from local                                        -> to remote
-			// <git-branch>:<git-commit-hash>:<module-directory> -> <module-identity>:<bsr-commit-name>
+			// <module-directory>:<git-branch>:<git-commit-hash> -> <module-identity>:<bsr-commit-name>
 			fmt.Sprintf(
 				"%s:%s:%s -> %s:%s\n",
-				moduleCommit.Branch(), moduleCommit.Commit().Hash().Hex(), moduleCommit.Directory(),
+				moduleCommit.Directory(), moduleCommit.Branch(), moduleCommit.Commit().Hash().Hex(),
```
unmultimedio, line 255: I missed this change from the previous PR

PR #2424, opened 2023-09-06
Delete branch and create_time from ModulePin

bufdev:
Neither of these are being used on client or server side, and there should be no concern with not serving them anymore.

This also cleans up some of the `bufmoduleref` validation functions.


proto/buf/alpha/registry/v1alpha1/module.proto
```diff
@@ -32,15 +30,15 @@ message LocalModuleReference {
 //
 // It does not include a remote.
 message LocalModulePin {
-  reserved 5;
-  reserved "digest";
+  reserved 3, 5, 7, 8;
+  reserved "branch",
+    "create_time"
+    "digest"
+, "draft_name";
```
unmultimedio, line 37: Pls check indentation here.

bufdev: https://github.com/bufbuild/buf/issues/2425

unmultimedio (approved):

PR #2426, opened 2023-09-06
Improve buf/manifest package and proto docs

pkwarren:
Update buf/manifest docs (and remove a few options which should have always been enabled). Improve documentation in the module .proto files.

### DIFF
--- a/private/pkg/manifest/blob.go
+++ b/private/pkg/manifest/blob.go
@@ -17,6 +17,7 @@ package manifest
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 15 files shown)


private/pkg/manifest/blob.go
```diff
@@ -97,69 +101,31 @@ type BlobSet struct {
 	digestToBlob map[string]Blob
 }
 
-type blobSetOptions struct {
-	validateContent bool
-	skipNilBlobs    bool
```
pkwarren, line 102: Both of these options should have probably defaulted to true all along, and there are no callers that are negatively affected by setting these to true so I'm simplifying the API surface area and removing the options.

bufdev (approved):

unmultimedio (approved):

PR #2433, opened 2023-09-11
Add `ListBSRHeadHistory`and `HasGitMetadata` rpcs and adjust `LocalResolveService` to return BSR_HEAD info

doriable:
This PR adds a new RPC for `BSR_HEAD` history, which returns a list
of commits specifically for the `BSR_HEAD`, agnostic of the branch.

It also adds `BSR_HEAD` information to the `LocalResolveService` pin.


proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -114,6 +114,9 @@ message LocalModuleResolveResult {
   buf.alpha.registry.v1alpha1.LocalModulePin pin = 2;
   // The type the reference resolved as.
   ResolvedReferenceType resolved_reference_type = 3;
+  // Indicates whether this reference is resolvable through BSR_HEAD.
+  // False indicates a non-default branch (formerly draft) commit.
+  bool bsr_head = 4;
```
unmultimedio, line 119: Nit
```suggestion
  bool is_bsr_head = 4;
```

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -163,3 +167,17 @@ message DeleteRepositoryDraftCommitRequest {
...
 message DeleteRepositoryDraftCommitResponse {}
+
+message ListBSRHeadHistoryRequest {
+  // The owner of the repository which the repository branch belongs to.
+  string repository_owner = 1;
+  // The name of the repository which the repository branch belongs to.
+  string repository_name = 2;
+  uint32 page_size = 3;
+  string page_token = 4;
+}
```
unmultimedio, line 178: Add `bool reverse`? And declare the default order in which it this returns commits (most recent to older).

doriable: Yeah, it makes sense to add reverse now, despite the fact that there currently isn't a use for it.

doriable: Added -- currently it's ordered in reverse by create time, but it will be in reverse order by parentage once we make the data model changes. Just noting this for documentation purposes :)

unmultimedio (approved):

unmultimedio (approved):

proto/buf/alpha/registry/v1alpha1/repository_commit.proto
```diff
@@ -163,3 +167,20 @@ message DeleteRepositoryDraftCommitRequest {
 }
 
 message DeleteRepositoryDraftCommitResponse {}
+
+message ListBSRHeadHistoryRequest {
```
bufdev, line 171: Why do we need another endpoint for this? What does this mean? Can we have docs on this as well?

doriable: This ended up being removed, but it was documented in the RPC definition in the service. In the future, I can add docs to the request and response messages as well.

PR #2441, opened 2023-09-13
Remove `HasGitMetadata` rpc and have it be a setting of `Repository`

doriable:
This removes the `HasGitMetadata` rpc and instead tracks this
as a part of the `Repository`. This removes the granular rpc in favour,
and instead keeps this scoped to the repository's settings information.


seankimdev (commented):
I think the commit is missing the removal of `HasGitMetadata` rpc?

doriable:
> I think the commit is missing the removal of `HasGitMetadata` rpc?

Ah yes, I undid that by accident -- pushing up an update.

PR #2445, opened 2023-09-14
Validate no duplicate module targets across directories

unmultimedio:
Make sure that in a branch scope, each module directory has a unique module target.

### DIFF
--- a/private/buf/bufsync/syncer.go
+++ b/private/buf/bufsync/syncer.go
@@ -258,11 +258,11 @@ func (s *syncer) prepareSync(ctx context.Context) error {
 	for _, branch := range s.sortedBranchesForSync {
 		identitiesInBranch := make(map[string][]string) // moduleIdentity:[]moduleDir
 		for _, moduleDir := range s.sortedModulesDirsForSync {
-			branchesToIdentities, ok := s.modulesDirsToBranchesToIdentities[moduleDir]
+			branchToIdentity, ok := s.modulesDirsToBranchesToIdentities[moduleDir]
 			if !ok {
 				continue // this module directory won't be synced by any branch
 			}
-			identity, ok := branchesToIdentities[branch]
+			identity, ok := branchToIdentity[branch]
 			if !ok || identity == nil {
 				continue // this module directory won't be synced by this branch
 			}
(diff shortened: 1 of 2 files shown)


private/buf/bufsync/syncer.go
```diff
@@ -251,6 +253,33 @@ func (s *syncer) prepareSync(ctx context.Context) error {
...
+	for _, branch := range s.sortedBranchesForSync {
+		identitiesInBranch := make(map[string][]string) // moduleIdentity:[]moduleDir
+		for _, moduleDir := range s.sortedModulesDirsForSync {
+			branchesToIdentities, ok := s.modulesDirsToBranchesToIdentities[moduleDir]
+			if !ok {
+				continue // this module directory won't be synced by any branch
+			}
+			identity, ok := branchesToIdentities[branch]
+			if !ok || identity == nil {
+				continue // this module directory won't be synced by this branch
```
saquibmian, line 267: nit: Why does `branchesToIdentities` contain a key for the branch if it won't be synced?

private/buf/bufsync/syncer.go
```diff
@@ -251,6 +253,33 @@ func (s *syncer) prepareSync(ctx context.Context) error {
 			}
 		}
 	}
+	// make sure no duplicate identities for different directories in the same branch
+	var duplicatedIdentitiesErr error
+	for _, branch := range s.sortedBranchesForSync {
+		identitiesInBranch := make(map[string][]string) // moduleIdentity:[]moduleDir
+		for _, moduleDir := range s.sortedModulesDirsForSync {
+			branchesToIdentities, ok := s.modulesDirsToBranchesToIdentities[moduleDir]
```
saquibmian, line 261: nit: branchToIdentity (there's only one identity, whereas I read this as several and I was confused)

saquibmian (approved):
LGTM, a couple small nits.

private/buf/bufsync/syncer.go
unmultimedio: Sure, it's a map 1:1, so many branches to many identities 1:1 haha, that's cool, I'll rename.

private/buf/bufsync/syncer.go
unmultimedio: This was more of a safeguard, to avoid any panic. It's true, if a module+branch combination don't have a target, there wouldn't be a key, we don't populate all of them exhaustively, only the ones that have target values, but the `identity == nil` doesn't hurt, wdyt?

PR #2447, opened 2023-09-15
Sync from custom remote

unmultimedio:
Introduce a new `--remote` flag that allows customizing a remote name to fetch the commits from. By default we still sync from local.


private/buf/bufsync/syncer.go
```diff
@@ -384,28 +392,24 @@ func (s *syncer) syncModuleInBranch(
 		if builtModule == nil {
 			return fmt.Errorf("syncable commit %s has no built module to sync", commitHash)
 		}
-		if builtModule.ModuleIdentity() == nil {
-			return fmt.Errorf("built module for commit %s has no module identity", commitHash)
-		}
-		modIdentity := builtModule.ModuleIdentity().IdentityString()
```
unmultimedio, line 390: This was redundant, all modules in the syncable commits slice had the same identities because they were being renamed to the target. We already have the target, so we can just don't rename anything, and use the target we already know.

private/buf/bufsync/syncer.go
```diff
@@ -475,39 +479,41 @@ func (s *syncer) branchSyncableCommits(
...
-				if err != nil {
-					return fmt.Errorf("override module in commit: %s, rename module: %w", readErr.Error(), err)
-				}
-				commitsForSync = append(commitsForSync, &syncableCommit{commit: commit, module: renamedModule})
-			default:
-				return fmt.Errorf("unexpected decision code %d for read module error %w", decision, readErr)
+		if readErr == nil {
+			commitsForSync = append(commitsForSync, &syncableCommit{commit: commit, module: builtModule})
+			return nil
+		}
```
unmultimedio, line 485: Just reducing nesting

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
```diff
@@ -150,13 +152,17 @@ func run(
 	} else if flags.Create {
 		return appcmd.NewInvalidArgumentErrorf("--%s is required if --%s is set.", createVisibilityFlagName, createFlagName)
 	}
+	if flags.Remote == "" {
+		return appcmd.NewInvalidArgumentErrorf("--%s cannot be empty.", remoteFlagName)
+	}
```
unmultimedio, line 157: This layer has the only check against empty remote. If we ever decide to support local syncs again, it's just a matter of fixing this layer.

private/pkg/git/git.go
```diff
@@ -47,13 +46,6 @@ const (
 	ModeSubmodule ObjectMode = 016_0000
 )
 
-const (
-	// refTypeHash is the reference type for a git SHA hash.
-	refTypeHash referenceType = "hash"
-	// refTypeBranch is the reference type for a git branch.
-	refTypeBranch referenceType = "branch"
-)
```
unmultimedio, line 55: Converted to structs implementing the same interface, because the branch reference has an optional remote name.

private/pkg/git/repository.go
```diff
@@ -44,7 +44,7 @@ type repository struct {
 	// packedOnce controls the fields below related to reading the `packed-refs` file
 	packedOnce      sync.Once
 	packedReadError error
-	packedBranches  map[string]Hash
+	packedBranches  map[string]map[string]Hash // remote:branch:hash (empty remote means local)
```
unmultimedio, line 47: When unpacking we know unpack all branches, alongside their remotes.

private/pkg/git/repository.go
```diff
@@ -153,41 +165,12 @@ func (r *repository) CurrentBranch() string {
...
-			commitID, err := NewHashFromHex(config.start.refName)
-			if err != nil {
-				return fmt.Errorf("new hash from %s: %w", config.start.refName, err)
-			}
-			startCommit, err = r.objectReader.Commit(commitID)
-			if err != nil {
-				return fmt.Errorf("read commit %s: %w", commitID.Hex(), err)
-			}
-		case refTypeBranch:
-			branch := normalpath.Unnormalize(config.start.refName)
```
unmultimedio, line 180: This turned out to fail in Windows tests. I'm a bit confused there, afaik windows creates branches refs `"foo/bar"`  as `foo\bar`, but it failed finding the branch. Did you remember if you see something similar in the first implementations of the git pkg @saquibmian ?

unmultimedio: Removing the `Unnormalize` line fixes the test, but that's what's confusing to me, isn't the branch supposed to be created normalized with backwards slashes?

private/pkg/git/repository.go
```diff
@@ -212,19 +194,18 @@ func (r *repository) ForEachCommit(f func(Commit) error, options ...ForEachCommi
...
 		}
 		if !info.Mode().IsRegular() {
 			return nil
 		}
-		tagName, err := filepath.Rel(dir, path)
+		tagName, err := normalpath.Rel(dir, path)
 		if err != nil {
 			return err
 		}
-		tagName = normalpath.Normalize(tagName)
```
unmultimedio, line 227: No need to normalize the received `tagName` if we used `normalpath.Rel` instead of `filepath.Rel`.

private/pkg/git/repository.go
unmultimedio: It's indeed related to normalization. Switched the usages of `path.Join` and other `path.` funcs to `normalpath.` pkg, which internally handles gracefully the OS separators.

private/buf/cmd/buf/command/alpha/repo/reposync/reposync.go
unmultimedio: Removed, by default we sync from local.

private/pkg/git/git.go
```diff
@@ -376,16 +383,45 @@ func OpenRepositoryWithDefaultBranch(name string) OpenRepositoryOption {
...
+
+type forEachCommitWithBranchStartPointOpts struct {
+	remote string
+}
 
 // reference is a single git reference used in ForEachCommit to declare an starting commit.
-type reference struct {
-	refName string
-	refType referenceType
+type reference interface {
```
saquibmian, line 400: This should go in `reference.go`.

private/pkg/git/git.go
```diff
@@ -376,16 +383,45 @@ func OpenRepositoryWithDefaultBranch(name string) OpenRepositoryOption {
...
+
+type branchReference struct {
+	name   string
+	remote string
+}
+
+func (r *branchReference) refType() string { return "branch" }
+func (r *branchReference) refName() string {
+	if r.remote != "" {
+		return r.remote + "/" + r.name
```
saquibmian, line 420: Use `normalpath.Join` here.

private/pkg/git/git_test.go
```diff
@@ -206,59 +206,59 @@ func createGitDirs(
 ) (string, string) {
 	tmpDir := t.TempDir()
 
-	submodulePath := filepath.Join(tmpDir, "submodule")
+	submodulePath := normalpath.Join(tmpDir, "submodule")
```
saquibmian, line 209: You can't use a normal path with the `os` package.

saquibmian (commented):
Let's chat about `path` -> `normalpath`

private/pkg/git/git.go
unmultimedio: How do you differentiate from other unexported func opts? The reference is also related just to func opts.

private/pkg/git/git_test.go
unmultimedio: You're right. Changed to only use `normalpath.Unnormalize` before passing it as elements to Join, and `filepath` pkg for the rest of it.

private/pkg/git/repository.go
```diff
@@ -125,17 +133,20 @@ func (r *repository) ForEachBranch(f func(string, Hash) error) error {
 		}
 		unpackedBranches[branchName] = struct{}{}
 		return f(branchName, hash)
-	}); err != nil {
+	}); err != nil && !errors.Is(err, fs.ErrNotExist) {
```
unmultimedio, line 136: If the `branchesDir` does not exist (like someone asking for branches in a remote name that does not exist), instead of failing, we just don't loop over any branch.

Also, it might be possible that we do have some packed branches for that repo.

doriable (approved):

private/pkg/git/git.go
saquibmian: Ah that's fair, you're right.

saquibmian (approved):
LGTM, I think we should do a pass before launching with a whole bunch of comments in the code.

PR #2457, opened 2023-09-18
Add `--branch` to `buf push`

doriable:
This aliases the `--draft` flag to `--branch` in `buf push`.
We are planning to start branch support with parity to `--draft`.
The description is also updated for the flag to reflect branch.


private/buf/cmd/buf/command/push/push.go
```diff
@@ -133,6 +134,17 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"Do not use. This flag never had any effect",
 	)
 	_ = flagSet.MarkHidden(trackFlagName)
+	flagSet.SetNormalizeFunc(aliasDraftToBranchFunc)
+}
+
+// aliasDraftToBranchFunc aliases the draft flag to branch.
```
bufdev, line 140: - are we sure we want to directly alias, and have no change in behavior?
- is there not a native way in pflag/cobra to do this?

doriable: For the initial step of releasing the branches UI, yes. There are server side changes we need to make for upcoming features, but it doesn't warrant a flag structure change.

This is the recommended way from Cobra's `pflag` package: https://github.com/spf13/pflag#mutating-or-normalizing-flag-names

bufdev: Once we do this, it's forever, so we can't later bring in any changes to `--branch` - just confirming we're comfortable with this

doriable: I had an original draft of this change where I added a new `--branch` flag, and then had a check return an error if the user attempted to pass both a `--draft` and `--branch`.
The normalize func here seems to be the recommended way to do a direct alias, but I think I would be more comfortable with the new flag and the check. Does that seem reasonable to you?

bufdev: Yea I would error if both are specified.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Update `buf beta price` with the latest pricing information.
 - Fix issue where `buf generate --exclude-path` was not properly excluding paths
   for remote modules.
+- Alias the `--draft` flag for `buf push` to `--branch` and update flag description.
```
bufdev, line 8: I would not call out the alias. More specifically, I would actually only do this when you are ready to deprecate `--draft` entirely, and then deprecate and hide `--draft`, and add `--branch.

bufdev (approved):

PR #2458, opened 2023-09-19
Add message documentation for GitCommitInformation

doriable:
Adds documentation to the `GitCommitInformation` message.


proto/buf/alpha/registry/v1alpha1/git_metadata.proto
```diff
@@ -28,6 +28,9 @@ message GitIdentity {
   google.protobuf.Timestamp time = 3;
 }
 
+// GitCommitInformation is the information associated with a Git commit.
+// This always includes the hash.
+// The author and/or committer user identities are included when available.
```
bufdev, line 33: Unrelated, but why are `author` and `committer` marked `optional`? Messages have presence by default.

doriable: Yeah, there was feedback from a different proto a while back to make this explicit, despite the fact that it's already the default. That being said, I think between the message and field documentation, it should be clear, I'll remove the `optional`.

bufdev (approved):

PR #2464, opened 2023-09-26
Add `UserCanReviewCommits` to AuthzService

stefanvanburen:
For the UI to determine which users should be able to see and use the review flow.


paul-sachs (approved):

bufdev (approved):
This type of RPC will disappear in the next version of our API, but given that it sticks with the pattern, OK for now.

PR #2465, opened 2023-09-28
Improve error message for invalid remotes

emcfarlane:
Extends error handling for invalid remotes to include `permission_denied` (is it safe to drop the code check?). Connect errors were being unwrapped back to simple form removing the message, so avoid unwrapping if not handled in the connect check. 

Before: 
```diff
-Failure: 403 Forbidden
+Failure: 403 Forbidden. Are you sure "github.com" (derived from module name "github.com/tmc/buf-fhir-example") is a Buf Schema Registry?
```
Issue: https://bufbuild.slack.com/archives/CRZ680FUH/p1695882656754589

### DIFF
--- a/private/buf/bufcli/errors.go
+++ b/private/buf/bufcli/errors.go
@@ -181,10 +181,9 @@ func wrapError(err error) error {
 			}
 			return errors.New(msg)
 		}
-		return fmt.Errorf("Failure: %s", connectErr.Message())
 	}
 
-	// Error was not a Connect error
+	// Error was not a known Connect error, so return it as-is.
 	return fmt.Errorf("Failure: %w", err)
 }
 
--- a/private/buf/bufcli/errors.go
+++ b/private/buf/bufcli/errors.go
@@ -141,7 +141,10 @@ func NewTokenNotFoundError(tokenID string) error {
 	return fmt.Errorf("a token with ID %q does not exist", tokenID)
 }
 
-func NewUnimplementedRemoteError(err error, remote string, moduleIdentity string) error {
+func NewInvalidRemoteError(err error, remote string, moduleIdentity string) error {
+	if connectErr, ok := asConnectError(err); ok {
+		err = connectErr.Unwrap()
+	}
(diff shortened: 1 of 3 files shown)


private/buf/bufcli/errors.go
```diff
@@ -181,10 +181,9 @@ func wrapError(err error) error {
 			}
 			return errors.New(msg)
 		}
-		return fmt.Errorf("Failure: %s", connectErr.Message())
```
timostamm, line 184: Looks like we did not print the error code before - it was just:

```
Failure: 403 Forbidden
```

I think this is intentional, see "intuitive message" comment at the top of the block.

With this change, we're adding the helpful message to the output, but we're also adding stutter in the form of "Failure: permission_denied: 403 Forbidden."

I don't see an easy way to retain the error without printing the code here, but I think we should make this possible. Someone more familiar with the code base than me might have an idea.

emcfarlane: Thanks I missed that, fixed the description. On the error site could avoid the `%w` to not wrap and interpret the connect error message: https://github.com/bufbuild/buf/blob/9bab8d4d668c993085b19636b2f18a6723c30cae/private/buf/bufcli/errors.go#L145
Also not sure if using `NewUnimplementedRemoteError` is the right message for this status code, and other errors on this path aren't wrapped. But for this issue it does make sense.

emcfarlane:
```go
func NewUnimplementedRemoteError(err error, remote string, moduleIdentity string) error {
	if connectErr, ok := asConnectError(err); ok {
		err = connectErr.Unwrap()
	}
	return fmt.Errorf("%w. Are you sure %q (derived from module name %q) is a Buf Schema Registry?", err, remote, moduleIdentity)
}
```

emcfarlane: Pushed up a fix to always unwrap the connect error. Removes the `permission_denied` code.

bufdev (approved):

PR #2470, opened 2023-10-02
Add ProtoYAML support to buf

Alfus:


go.mod
```diff
@@ -41,9 +42,12 @@ require (
 )
 
 require (
+	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.31.0-20230914171853-63dfe56cc2c4.1 // indirect
```
bufdev, line 45: We can't do this directly without a `replace` - see what we did in https://github.com/bufbuild/buf/pull/2473. I'd recommend we get #2473 merged first.

bufdev (commented):
- The only tests I see are around `buf convert`, can we do some round trip tests around `buf build`? I believe we already have some.
- What happens if I use snake case keys instead of camel case keys in my YAML file? `protojson` handles both, wondering if we have an answer for this. Optimally, we'd have a test for this as well (assuming we handle both).

go.mod
Alfus: Merged in

Alfus:
> * The only tests I see are around `buf convert`, can we do some round trip tests around `buf build`? I believe we already have some.

Added a test that uses the yaml format as both output and input to/from buf build. Is this what you mean, or do you want more tests, somewhere else?


> * What happens if I use snake case keys instead of camel case keys in my YAML file? `protojson` handles both, wondering if we have an answer for this. Optimally, we'd have a test for this as well (assuming we handle both).

protoyaml handles both, and tests that in it's repo.

bufdev (approved):

PR #2471, opened 2023-10-02
Add get and update user plugin preferences rpcs

gilwong00:
This adds two new endpoints `GetUserPluginPreferences` and `UpdateUserPluginPreferences` to the `UserService`


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,19 @@ message OrganizationUser {
   OrganizationRoleSource organization_role_source = 4;
 }
 
+message UserSelectedPlugin {
```
bufdev, line 57: Is there an existing type for this? What is a user-selected plugin? Needs docs explaining what this is at the minimum

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,19 @@ message OrganizationUser {
...
+  // owner of the plugin
+  string owner = 1;
+  // name of the plugin
+  string name = 2;
+}
+
+message UserPluginPreference {
```
bufdev, line 64: What is this? needs docs at the minimum

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,19 @@ message OrganizationUser {
...
+  string user_id = 1;
+  PluginLanguage selected_language = 2;
```
bufdev, line 66: Asymmetry between `UserSelectedPlugin` having `Selected` in it vs. this not, likely could be explained by understanding what API is and is for via docs

bufdev (changes requested):

proto/buf/alpha/registry/v1alpha1/user.proto
gilwong00: we can actually dont need this type anymore. Removed.

gilwong00:
@bufdev cleaned up this PR and added more docs around the new types

proto/buf/alpha/registry/v1alpha1/user.proto
bufdev: That is this used for? The documentation just says what the fields are, I'm not sure what this structure is for, and how it fits into the Buf product.

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,19 @@ message OrganizationUser {
   OrganizationRoleSource organization_role_source = 4;
 }
 
+// UserPluginPreference is a nested structure that contains the
```
bufdev, line 57: What is nested in this context?

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,19 @@ message OrganizationUser {
...
+// users preferred or selected language as well as a list of plugins
+// selected for that language
+message UserPluginPreference {
+  // The id of the user
+  string user_id = 1;
+  // The selected language that the user has indicidated
```
bufdev, line 63: I'm not sure what a preferred language for plugins is. What is this used for? How does this fit into the Buf product?

gilwong00: This is for the new Generated SDKs tab. We changed the flow so a user first selects their preferred language and then we display a list of plugins pertaining to that language. We want to persist the language and the plugins the user selected so the next time they visit this page we dont need to have them go through the selection process again

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -105,6 +119,12 @@ service UserService {
   }
   // UpdateUserSettings update the user settings including description.
   rpc UpdateUserSettings(UpdateUserSettingsRequest) returns (UpdateUserSettingsResponse);
+  // GetUserPluginPreferences gets plugin preferences for the user.
```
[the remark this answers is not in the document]
gilwong00: plugin preferences is the language and the list of plugins a user has selected in the new Generated SDKs tab

proto/buf/alpha/registry/v1alpha1/user.proto
bufdev: The point here is that any context should be added to the docs - I should be able to read these docs and understand immediately what is going on here.

proto/buf/alpha/registry/v1alpha1/user.proto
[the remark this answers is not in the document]
bufdev: Should be in docs

proto/buf/alpha/registry/v1alpha1/user.proto
gilwong00: Understood. Added a explanation that this type is meant for Generated SDKs.

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -53,6 +54,17 @@ message OrganizationUser {
   OrganizationRoleSource organization_role_source = 4;
 }
 
+// UserPluginPreference contains the users selected language as well as a list of plugins selected for that language.
+// This is used for Generate SDKs where the user can choose a language and corresponding plugins to be persisted.
```
bufdev, line 58: I still don't know what this means - what does it mean for a plugin to be persisted? What does it mean to choose a language?

I'd check out https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/asset/v1/remote_asset.proto as a good example of Protobuf API documentation.

bufdev (approved):

PR #2472, opened 2023-10-02
Drop actions cache for setup-go

emcfarlane:
v4 enables caching by default: https://github.com/actions/setup-go#v4


emcfarlane:
This is working as intended, looks llke it still downloads.

PR #2473, opened 2023-10-02
Add protovalidate lint rules

oliversun9:
This PR adds `PROTOVALIDATE` rule to `buf lint`.
It checks 
- all CEL expressions compile
- rules permit some value. For example, `gt: 5` and `lt: 1` reject all values
- rules match the type of the field. Wrapper types can also match rules for its value type
- rules are not obviously redundant. For example, if `const` is specified, other rules are obviously redundant. However, if `string.min_len` is 5 and `string.min_bytes` is 4, the latter is redundant but this is not obvious. (I initially had a quite many redundancy checks. I removed them after discussing with @rodaine)

If the `.proto` author wrote options in non-aggregate form, `buf lint` can pinpoint. If the options are written in the aggregate form, then everything between the braces gets highlighted, since that's how source code locations are generated.

Non-aggregate
 ```
  uint32 a = 5 [
    (buf.validate.field).uint32.lt = 10,
    // only the next line is highlighted
    (buf.validate.field).uint32.const = 1,
    (buf.validate.field).uint32.gt = 8
  ];
  ``` 
  
Aggregate
  ```
  uint32 a = 5 [
    // the next 5 lines are highlighted
    (buf.validate.field).uint32 = {
      lt: 10,
      const: 1,
      gt: 8
    }
  ];
```

### CODE REVIEW

File: go.mod [hunk 1]
```diff
@@ -6,10 +6,12 @@ require (
 	connectrpc.com/connect v1.11.1
 	connectrpc.com/otelconnect v0.5.0
 	github.com/bufbuild/protocompile v0.6.0
+	github.com/bufbuild/protovalidate-go v0.3.3-0.20231002212436-9023bd754f7f
```
oliversun9, line 9: This is from `go get ...@main`, perhaps we should tag a release on `protovalidate-go` first?

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	fieldDescriptor, ok := descriptor.(protoreflect.FieldDescriptor)
+	if !ok {
+		// this should never happen
+		return nil, fmt.Errorf("%s is not a field", descriptor.FullName())
+	}
+	return fieldDescriptor, nil
+}
+
+func getMessageConstraints(message protosource.Message) (*validate.MessageConstraints, bool, error) {
```
jhump, line 201: Why not return `*validate.MessageConstraints, error`? It looks like the bool is only true when the message != nil, so it's superfluous, no?

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
```diff
@@ -0,0 +1,311 @@
...
+	messageConstraintsMessageUntyped, found := message.OptionExtension(messageConstraintsExtensionType)
+	if !found {
+		return nil, false, nil
+	}
+	messageConstraintsMessage, ok := messageConstraintsMessageUntyped.(protoreflect.Message)
+	if !ok {
+		// this should never happen
+		return nil, false, fmt.Errorf("field extension expected to be `protoreflect.Message`, but has type %T", messageConstraintsMessageUntyped)
+	}
+	data, err := proto.Marshal(messageConstraintsMessage.Interface())
```
jhump, line 211: This seems to do the same thing as in @rodaine's recent PR in protovalidate itself. Could you instead somehow use the protovalidate runtime to extract the constraints, instead of repeating all of this?

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
```diff
@@ -0,0 +1,311 @@
...
+const (
+	// https://buf.build/bufbuild/protovalidate/docs/main:buf.validate#buf.validate.FieldConstraints
+	celFieldTagInFieldConstraints = 23
+	// https://buf.build/bufbuild/protovalidate/docs/main:buf.validate#buf.validate.MessageConstraints
+	celFieldTagInMessageConstraints = 3
+)
+
+var (
+	messageConstraintsExtensionType = dynamicpb.NewExtensionType(validate.E_Message.TypeDescriptor())
+	fieldConstraintsExtensionType   = dynamicpb.NewExtensionType(validate.E_Field.TypeDescriptor())
```
jhump, line 42: Definitely worth comments as to what these are here for, since `validate.E_Message` and `validate.E_Field` already implement `protoreflect.ExtensionType`.

My understanding (from other conversations) is that they need to be _dynamic_ extensions to avoid the type-assertion-panic that might otherwise happen when querying their value from the results of compilation.

You have `getReflectMessageDescriptor` in here, so you already have a way to get the `protoreflect.MessageDescriptor` that corresponds to a particular `protosource.Message`. Why not instead use that instead _with_ the protovalidate runtime to avoid having to repeat so much of what's in protovalidate here?

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
oliversun9: There is a case where the message doesn't have this option, but nothing else went wrong, and in this case the helper would return `nil, false, nil`. The caller benefits from knowing whether it should move on (this options isn't present) or it should throw an error (something has gone wrong).

I agree that returning `*validate.MessageConstraints, error` makes its signature cleaner. In that case, it would return some kind of `NotFound` error. I searched for `NotFound` in the codebase and couldn't find a standardized `NotFound` error. Do you have any suggestions?

go.mod
rodaine: https://github.com/bufbuild/protovalidate-go/releases/tag/v0.3.3

oliversun9: Thanks for releasing, resolved in [662bbbc](https://github.com/bufbuild/buf/pull/2473/commits/662bbbc55e6ac76b3a9144d80bbda692b3d5f2ad).

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
oliversun9: You are right, I opened https://github.com/bufbuild/protovalidate-go/pull/59 and will use `DefaultResolver` to resolve these constraints.

oliversun9: One implication of using `DefaultResolver` from `protovalidate-go` is that in `buf lint` we will need to use remote package for `protovalidate`, since `protovalidate` uses it.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
oliversun9: As another benefit of using `protovalidate-go`, these two variables are no longer needed. Resolved in [59e3caf](https://github.com/bufbuild/buf/pull/2473/commits/59e3caf2b1e1b4f024c52a17ec8372ce6e6b051e)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
oliversun9: Resolved in [5e33b02](https://github.com/bufbuild/buf/pull/2473/commits/5e33b02a5044a09ee659acb1d7bde7395490ffe1)

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/cel.go
oliversun9: These functions are replaced by direct calls to `DefaultResolver{}.Resolver...`. Resolved in [5e33b02](https://github.com/bufbuild/buf/pull/2473/commits/5e33b02a5044a09ee659acb1d7bde7395490ffe1).

bufdev (commented):
This is approved given the following:

- Add a test that will fail given the custom parsing of cel-go's error messages.
- We get the PR for the rest of protovalidate rules up in parallel. I'd like to merge them together, and to evaluate the final name for the respective lint rules (and perhaps combine them into one).

go.mod
```diff
@@ -80,3 +80,7 @@ require (
 	google.golang.org/genproto/googleapis/api v0.0.0-20230913181813-007df8e322eb // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20230913181813-007df8e322eb // indirect
 )
+
+// The Buf CLI is the one library that should not depend on BSR generated SDKs
+// as this could cause bootstrapping errors.
+replace buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go/buf/validate => ./private/gen/proto/go/buf/validate
```
jhump, line 86: Is this true? What bootstrapping errors do we expect? If we're using the `buf` CLI to work on `buf`, we're working with a binary where the remote packages are already resolved and compiled, so I don't see how this is a problem here.

I was under the impression that it's the core repo that must not depend on generated SDKs. Note that this directive here will do nothing to prevent a cyclic dependency in core since replace directives in dependencies are ignored by the go tool.

oliversun9: >we're working with a binary where the remote packages are already resolved and compiled

The `buf` CLI's binary itself will be fine, since like you said, it's just a binary.
> Note that this directive here will do nothing to prevent a cyclic dependency in core since replace directives in dependencies are ignored by the go tool.

Yes I think this is the concern and we will also need to add a replace directive there, before or around the time when this PR is merged.

jhump: > Yes I think this is the concern

So we can remove the replace directive here?

oliversun9: Great question, and with the reasoning above I think it's probably fine? @bufdev what do you think?

bufdev (approved):

PR #2486, opened 2023-10-10
Fix release workflow yaml

doriable:
This fixes the formatting on the release workflow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -91,8 +91,8 @@ jobs:
         with:
           app_id: ${{env.APP_ID}}
           private_key: ${{ secrets.TOKEN_EXCHANGE_GH_APP_PRIVATE_KEY }}
-          repositories: >-
-            [ ${{ github.repository }} ]
+          repositories:
+            - ${{ github.repository }} ]
           permissions: >-
             {"contents": "write"}
       - name: Set VERSION variable from tag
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -91,7 +91,8 @@ jobs:
         with:
           app_id: ${{env.APP_ID}}
           private_key: ${{ secrets.TOKEN_EXCHANGE_GH_APP_PRIVATE_KEY }}
-          repositories: [ ${{ github.repository }} ]
+          repositories:
+            - ${{ github.repository }} ]
```
joshcarp, line 95:
```suggestion
            - ${{ github.repository }} 
```

oliversun9, line 95: Do we still need the `]`?

doriable: No definitely not, moved through this a bit too quickly! ^^"

.github/workflows/release.yaml
```diff
@@ -320,7 +327,8 @@ jobs:
         with:
           app_id: ${{secrets.BUF_BUILD_RELEASE_APP_ID}}
           private_key: ${{ secrets.BUF_BUILD_RELEASE_APP_PRIVATE_KEY }}
-          repositories: [ "bufbuild/buf.build" ]
+          repositories:
+            "bufbuild/buf.build"
```
joshcarp, line 331:
```suggestion
            - "bufbuild/buf.build"
```

joshcarp (approved):

Issue #2480, opened 2023-10-08
Calling buf linter or generate on a remote git repository returns a dependency does not exist error


talgendler:
Hi, 
After upgrading to `v1.27.0` version I'm no longer capable to lint or generate code from a remote repository if the proto in context is dependent on other protos not in the same buf module but in a different one (both defined in `buf.work.yaml`).
_I simplified all the examples below._

Given an API repository with multiple modules (`buf.work.yaml`) up until now I was able to generate code from a remote (github) API repository by running this command:
```
buf generate ssh://git@github.com/company/api.git#branch=main,subdir=mod1 --path mod1/mod1/public/v2/public.proto
```

Now I get this error:

```
mod1/mod1/public/v2/public.proto:5:8:http/errors/v1/http_errors.proto: does not exist
```

This `api.git` repository only contains proto and buf definitions.

The above `public.proto` is dependent on others and it imports them like this:
```proto
syntax = "proto3";

package mod1.public.v2;

import "http/errors/v1/http_errors.proto";
import "http/generic/v1/http_generic.proto";

service ...
```

There directory structure of the api repo (simplified):

```
.
├── README.md
├── buf.work.yaml
├── http
│   ├── buf.yaml
│   └── http
│       ├── errors
│       │   └── v1
│       │       └── http_errors.proto
│       └── generic
│           └── v1
│               └── http_generic.proto
└── mod1
    ├── buf.yaml
    └── mod1
        └── public
            ├── v1
            │   └── public.proto
            └── v2
                └── public.proto
```

`buf.work.yaml` content:

```yaml
version: v1
directories:
  - http
  - mod1
```

If I add an additional `--path` pointing to the first imported dependency the call completes without errors.

```
buf generate ssh://git@github.com/company/api.git#branch=main,subdir=mod1 --path mod1/mod1/public/v2/public.proto --path http/http/errors/v1/http_errors.proto
```

bufdev:
Thanks for letting us know about this. I'm unable to reproduce this given the instructions - can you please provide a specific, reproducible example of your issue that we can run? A good way to do this, for example, is to create a temporary repository i.e. github.com/talgendler/buf-dependency-bug, that contains code and instructions that we can run that demonstrates the issue. If you're able to provide this, we can look further, but I am unable to replicate an equivalent issue with the latest release.

talgendler:
Thank you @bufdev, I created a new simple repository for this issue that can be found here:

[...]

PR #2490, opened 2023-10-11
Add ID to module

oliversun9:
Fixes https://github.com/bufbuild/buf/issues/2480
This fixes the problem where in some occasions dependencies from workspace are not loaded correctly when `--path` is passed. This happens when the `path` filter doesn't match any path in either directory, which results in the `Module`s from these two directories having no `TargetFileInfo`. Their hashes, used for identification within the workspace, are now the same as the hash for an empty module. After one is added, the other gets ignored.

In `getSourceModuleConfig` and `BuildWorkspace`, build modules with `ID`.
Note: it's not necessary to do this in `getSourceModuleConfig`, but added for consistency across local modules.


private/buf/bufwire/module_config_reader.go
```diff
@@ -622,6 +622,7 @@ func (m *moduleConfigReader) getSourceModuleConfig(
 		}
 		buildOptions = append(buildOptions, bufmodulebuild.WithExcludePaths(bucketRelPaths))
 	}
+	buildOptions = append(buildOptions, bufmodulebuild.WithID("dir:"+subDirPath))
```
doriable, line 625: For the IDs, I think we want to expose it as a part of the `Ref`, which means we'll need to expose it from the depths of `buffetch/internal` upwards.
This way, we actually also expose the type of ref this is, since this is not necessarily a `dir` ref.

private/bufpkg/bufmodule/module.go
```diff
@@ -364,6 +365,13 @@ func (m *module) Commit() string {
 	return m.commit
 }
 
+func (m *module) ID() string {
+	if m.moduleIdentity != nil {
+		return "mod:" + m.moduleIdentity.IdentityString()
+	}
+	return m.id
+}
```
doriable, line 373: If we change the `ID` to come from the ref, then we won't need this default, since any module built with that option will have the ref type + `subDir` as the ID.

doriable (changes requested):
The build option seems correct to me, but as mentioned in the comments, we'll need to get the ID from the ref, which means exposing that from `buffetch/internal` upwards on the `SourceOrModuleRef` interface in `buffetch`.

private/buf/buffetch/buffetch.go
```diff
@@ -93,6 +93,8 @@ type PathResolver interface {
 // Ref is an image file or source bucket reference.
 type Ref interface {
 	PathResolver
+	// ID is the Ref's ID.
```
bufdev, line 96: What does this mean? What is an ID? What is its form? What can I depend on and not depend on in terms of its form (likely nothing, likely that it's just an opaque string). ie explain all of this in the Godoc.

Since this is an internal package, an explanation of the form of each ID for each different `Ref` type, just for users' reference, would also be helpful here, but make it clear in the Godoc that the form of the ID is not stable and should not be relied on.

private/buf/buffetch/internal/internal.go
```diff
@@ -86,6 +86,8 @@ type CompressionType int
 
 // Ref is a reference.
 type Ref interface {
+	// ID is the ref's ID.
```
bufdev, line 89: Same comment on Godoc

private/buf/buffetch/internal/module_ref.go
```diff
@@ -61,6 +61,10 @@ func newDirectModuleRef(format string, moduleReference bufmoduleref.ModuleRefere
 	}
 }
 
+func (r *moduleRef) ID() string {
+	return r.moduleReference.IdentityString()
```
bufdev, line 65: This isn't unique necessarily - a module ref may include a branch, tag, or commit associated with it. Optimally, an ID would be the full ID with the commit, but regardless we can't just have the identity.

private/bufpkg/bufmodule/bufmodulebuild/bufmodulebuild.go
```diff
@@ -173,3 +173,10 @@ func WithExcludePathsAllowNotExist(excludePaths []string) BuildOption {
 		buildOptions.pathsAllowNotExist = true
 	}
 }
+
+// WithID returns a new BuildOption that specifies the module's Id.
```
bufdev, line 177: ID is always capitalized.

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -189,6 +193,13 @@ func ModuleWithModuleIdentityAndCommit(moduleIdentity bufmoduleref.ModuleIdentit
 	}
 }
 
+// ModuleWithID is used to construct a Module with an id.
```
bufdev, line 196: What is an ID on a Module? What is its form? What is this for? What's the motivation

Same general comments - I'd like us to be able to come back to the code in this PR in 6-12 months and be able to immediately understand the purpose of all of this without digging through PRs.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
```diff
@@ -59,12 +57,8 @@ func (m *moduleFileSetBuilder) build(
...
-	hashes := make(map[string]struct{})
-	moduleHash, err := protoPathsHash(ctx, module)
-	if err != nil {
-		return nil, err
-	}
-	hashes[moduleHash] = struct{}{}
+	moduleIds := make(map[string]struct{})
+	moduleIds[module.ID()] = struct{}{}
 	if workspace != nil {
 		// From the perspective of the ModuleFileSet, we include all of the files
```
bufdev, line 63: This long comment I wrote is likely now out of date - the same level of explanation needs to be here for what we're currently choosing, and alternatives.

bufdev:
Overall comment needs to be added to the effect of "IDs should be derived from data, ie we pull them from storage buckets, digests, etc, this is just a hack to weave through some form of quasi-unique identifier for Modules to fix this specific issue, and here's a link to the issue, we'll address this in the bufmodule refactor"

private/buf/buffetch/buffetch.go
```diff
@@ -216,6 +225,12 @@ type ReadWriteBucketCloser internal.ReadWriteBucketCloser
 // ReadBucketCloserWithTerminateFileProvider is a ReadBucketCloser with a TerminateFileProvider.
 type ReadBucketCloserWithTerminateFileProvider internal.ReadBucketCloserWithTerminateFileProvider
 
+// TerminateFileProvider provides TerminateFiles.
+type TerminateFileProvider = internal.TerminateFileProvider
+
+// TerminateFile is a terminate file.
+type TerminateFile = internal.TerminateFile
```
oliversun9, line 232: These are exported so we can have a mock `buffetch.Reader`

private/buf/buffetch/internal/single_ref.go
```diff
@@ -123,6 +123,10 @@ func newDirectSingleRef(
 	}
 }
 
+func (r *singleRef) ID() string {
+	return r.path
```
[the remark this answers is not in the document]
oliversun9: > why is that not just an image with a format?

I think the path itself can identify the image file. Or is this question about the naming of `singleRef`? I'm not sure why it's called that.

private/buf/bufwire/module_config_reader_test.go
```diff
@@ -179,11 +180,187 @@ func TestExcludePathsForModule(t *testing.T) {
...
+				require.Len(t, workspace.GetModules(), len(testcase.expectedWorkspaceModuleIds))
+				for i, expectedWorkspaceModuleID := range testcase.expectedWorkspaceModuleIds {
+					require.Equal(t, expectedWorkspaceModuleID, workspace.GetModules()[i].ID())
+				}
+			}
+		})
+	}
+}
+
+type fetchReader struct {
```
bufdev, line 358: We should not be mocking this - this is going to make refactors extremely difficult, and adds little value. I rather delete the tests entirely than add this, if those were the two options.

private/buf/buffetch/buffetch.go
bufdev: Re-unexport these - having to export these to write tests in another package sets off alarm bells for me. It likely means we should pursue a different testing strategy here.

oliversun9: unexported in [1c288c4](https://github.com/bufbuild/buf/pull/2490/commits/1c288c4f2e089a52abbe8187cb5c1119b235e5a9)

private/buf/bufwire/module_config_reader_test.go
oliversun9: added integration tests in [1e9928f](https://github.com/bufbuild/buf/pull/2490/commits/1e9928fb4f58c1929fd8f2ca189154579fc67e7c)

private/buf/cmd/buf/workspace_test.go
```diff
@@ -302,48 +304,6 @@ func TestWorkspaceNestedArchive(t *testing.T) {
 	)
 }
 
-func TestWorkspaceGit(t *testing.T) {
-	t.Skip("skip until the move to private/buf is merged")
-	// Directory paths specified as a git reference within a workspace.
```
oliversun9, line 307: This is re-enabled, but moved to workspace_unix_test.go because there seems to be a problem with git clone on the Windows CI, see [workflow run](https://github.com/bufbuild/buf/actions/runs/6510804063/job/17685247791). [This](https://stackoverflow.com/a/40909460) might fix it.

oliversun9: This change doesn't reduce the number of tests because it's already skipped on main.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
```diff
@@ -59,12 +57,8 @@ func (m *moduleFileSetBuilder) build(
...
 ) (bufmodule.ModuleFileSet, error) {
 	var dependencyModules []bufmodule.Module
-	hashes := make(map[string]struct{})
-	moduleHash, err := protoPathsHash(ctx, module)
-	if err != nil {
-		return nil, err
-	}
-	hashes[moduleHash] = struct{}{}
+	moduleIds := make(map[string]struct{})
+	moduleIds[module.ID()] = struct{}{}
```
bufdev, line 61: What if `ID` is empty?

private/buf/buffetch/internal/single_ref.go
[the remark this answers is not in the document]
saquibmian: Yeah no you're good, the `path` is sufficient.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
oliversun9: Added `getModuleIdentifier` to use path hash as a fallback

bufdev:
Subsumed by https://github.com/bufbuild/buf/pull/2490

PR #2495, opened 2023-10-13
Workspace directory

oliversun9:
Fixes https://github.com/bufbuild/buf/issues/2480
This fixes the problem where in some occasions dependencies from workspace are not loaded correctly when `--path` is passed. For example, `buf build workspacedir/a --path workspacedir/a/foo.proto` when `workspacedir/b` and `workspacedir/c` also exist. In `module_file_set_builder.go` module b and module c would be hashed based on their `TargetFileInfos`, which are both empty slices due to the `--path` filter. Same hash means only one of them gets added to the list of potential dependencies.

In this PR, workspace modules are distinguished by their directories relative to the workspace directory. This is accomplished by adding `WorkspaceDirectory` to interface `bufmodule.Module`.

### DIFF
--- a/private/buf/cmd/buf/testdata/workspace/success/dir/buf.work.yaml
+++ b/private/buf/cmd/buf/testdata/workspace/success/dir/buf.work.yaml
@@ -1,4 +1,5 @@
 version: v1
 directories:
+  - a/proto
   - other/proto
   - proto
--- a/private/buf/cmd/buf/workspace_test.go
+++ b/private/buf/cmd/buf/workspace_test.go
@@ -116,7 +116,8 @@ func TestWorkspaceDir(t *testing.T) {
 			t,
 			nil,
 			0,
-			filepath.FromSlash(`testdata/workspace/success/`+baseDirPath+`/other/proto/request.proto
+			filepath.FromSlash(`testdata/workspace/success/`+baseDirPath+`/a/proto/a/v1/a.proto
+		    testdata/workspace/success/`+baseDirPath+`/other/proto/request.proto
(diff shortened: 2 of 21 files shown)


private/buf/cmd/buf/workspace_test.go
```diff
@@ -302,48 +304,6 @@ func TestWorkspaceNestedArchive(t *testing.T) {
 	)
 }
 
-func TestWorkspaceGit(t *testing.T) {
```
oliversun9, line 305: This is re-enabled, but moved to workspace_unix_test.go because there seems to be a problem with git clone on the Windows CI, see [workflow run](https://github.com/bufbuild/buf/actions/runs/6510804063/job/17685247791). [This](https://stackoverflow.com/a/40909460) might fix it. This change doesn't reduce the number of tests because it's already skipped on main.

private/buf/cmd/buf/testdata/workspace/success/dir/buf.work.yaml
```diff
@@ -1,4 +1,5 @@
 version: v1
 directories:
+  - a/proto
```
oliversun9, line 3: Added a third module to test for the behavior when there are two potential dependency modules, which is what the bug is about. I verified that this fails on main and passes on this branch.

oliversun9:
Verified that running `buf lint ssh://git@github.com/talgendler/buf-remote-dependency-bug.git#branch=main,subdir=service --path service/service/private/v1/private.proto` with this branch succeeds.

bufdev (approved):

PR #2504, opened 2023-10-17
Add workaround for macOS archive xattrs

### LINKED ISSUE #2387: buf build xyz.tar fails if a file in this archive has an extended attribute
Files created in vscode, and potentially other apps, have an [extended attribute](https://apple.stackexchange.com/questions/450118/what-is-the-com-apple-provenance-xattr-extended-attribute-and-how-can-i-dele). Making a tape archive (`.tar`) and invoke `buf build` on such an archive will fail.

Steps to reproduce:
Run the following in the terminal built in to mac (not the one from vscode)

```
% mkdir dir1 dir2                              
% mv somewhere/vscode.proto dir1/vscode.proto #### assuming such a file exists and is created by vscode
% echo "message Foo {}" > dir1/vscode.proto    
% echo "message Bar {}" > dir2/terminal.proto  
% ls -al@ dir1/vscode.proto                    
-rw-r--r--@ 1 oliversun  staff  15 18 Aug 16:46 dir1/vscode.proto
	com.apple.provenance	11 
% ls -al@ dir2/terminal.proto                  
-rw-r--r--  1 oliversun  staff  15 18 Aug 16:46 dir2/terminal.proto
% tar -cvf t1.tar dir1                         
a dir1
a dir1/vscode.proto
% tar -cvf t2.tar dir2                         
a dir2
a dir2/terminal.proto
% buf --version                                
1.22.0
% buf build t2.tar 
% buf build t1.tar
dir1/._vscode.proto:1:1:invalid control character
dir1/._vscode.proto:1:1:syntax error: unexpected error
dir1/._vscode.proto:1:2:invalid control character
...
```

Verified on `1.22.0` and `1.26.2-dev` (latest main).

jchadwick-buf:
This commit introduces a workaround for an issue with .tar archives created with libarchive on Darwin/macOS. When encountering a file that has extended attributes, such as the com.apple.provenance attribute, it will add a corresponding AppleDouble file with the prefix "._" in the same directory. Because these files have the same filename otherwise, the buf tool will see these when they correspond to proto files and subsequently fail to parse them as protobuf IDL.

Due to the fact that libarchive is used by default with the version of tar that ships with macOS, and the provenance extended attribute is set by macOS under many conditions when SIP is enabled, archives with these files are likely to occur on macOS.

.zip archives created by libarchive do not seem to have this issue. However, .zip archives created by Archive Utility.app have a very similar behavior when encountering extended attributes: it will place AppleDouble files in a separate directory tree under the MACOSX directory, but also with the "._" prefix. Since this can be handled by the same logic, this behavior is extended to ZIP files as well.

Closes #2387.

### DIFF
--- a/private/pkg/storage/storagearchive/storagearchive.go
+++ b/private/pkg/storage/storagearchive/storagearchive.go
@@ -234,8 +234,10 @@ func isAppleExtendedAttributesFile(fileInfo fs.FileInfo) bool {
 	// Archive Utility.app has a similar behavior when creating .zip archives,
 	// except they are placed under a separate MACOSX directory tree.
 	// Here, both are handled by just ignoring all files with a "._" prefix.
-	// This is a reasonable compromise because protobuf IDL files with a "._"
-	// prefix are very unlikely to occur.
+	// This is a reasonable compromise because files that live in a Module
+	// (.proto files, configuration files such as buf.yaml, README files) are
+	// almost never prefixed with ._, and fixing this issue in this manner
+	// outweighs the slight incorrectness.
 	return strings.HasPrefix(fileInfo.Name(), "._")
 }
 
--- a/private/pkg/storage/storagearchive/storagearchive.go
+++ b/private/pkg/storage/storagearchive/storagearchive.go
@@ -21,7 +21,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"io/fs"
 	"math"
+	"strings"
 
 	"github.com/bufbuild/buf/private/pkg/normalpath"
 	"github.com/bufbuild/buf/private/pkg/storage"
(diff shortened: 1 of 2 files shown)


private/pkg/storage/storagearchive/storagearchive.go
```diff
@@ -219,6 +227,18 @@ func Unzip(
...
 }
 
+func isAppleExtendedAttributesFile(fileInfo fs.FileInfo) bool {
+	// On macOS, .tar archives created with libarchive will contain additional
+	// files with a prefix of "._" if there are files with extended attributes
+	// and copyfile is enabled.
+	// Archive Utility.app has a similar behavior when creating .zip archives,
+	// except they are placed under a separate MACOSX directory tree.
+	// Here, both are handled by just ignoring all files with a "._" prefix.
+	// This is a reasonable compromise because protobuf IDL files with a "._"
```
bufdev, line 237: "because files that live in a Module (`.proto` files, configuration files such as `buf.yaml`, README files) are almost never prefixed with `._`, and fixing this issue in this manner outweighs the slight incorrectness.

bufdev (commented):
Add a changelog entry as well

private/pkg/storage/storagearchive/storagearchive.go
jchadwick-buf: Done, thanks.

bufdev (approved):

PR #2510, opened 2023-10-23
Fix shutdown_finished duration field

pkwarren:
As written, the duration value is calculated when the defer function is registered. We should update it to be calculated when the defer is executed instead.


jchadwick-buf (approved):
LGTM.

PR #2511, opened 2023-10-23
Add integration guide url to buf plugin config

mfridman:
This PR extends the buf.plugin.yaml configuration to allow for a `integration_guide_url`.

This field is intended to link to an external integration guide for the plugin, e.g.,

https://buf.build/docs/bsr/remote-packages/swift#doc-content


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -381,6 +381,8 @@ message CreateCuratedPluginRequest {
   string license_url = 16;
   // Visibility indicates whether the plugin is public or private.
   CuratedPluginVisibility visibility = 17;
+  // The URL leading to the integration guide of the plugin, if available.
+  string integration_guide_url = 18;
```
pkwarren, line 385: Shouldn't this also be added to `CuratedPlugin` message type?

mfridman: Sure, I can add it in the same PR. I wanted to break this into 2 PR's, the config+request side, but no issue also updating the core type.

pkwarren (approved):

bufdev (approved):

PR #2520, opened 2023-10-27
Display warning when reading a buf.lock with a b3 digest

oliversun9:
Right now, if the CLI reads a buf.lock with a b3 digest, it treats digest as empty. This PR displays a warning when it's encountered.

Added a check that reads the lock file and checks whether any digest starts with `b1-` or `b3-` to
- `getSourceModuleConfig`, called by `buf beta graph`, `buf beta price`, `buf beta stats`, `buf build`, `buf convert` , `buf export`, `buf format`, `buf generate`and `buf lint`, when the source / input is a module directory or workspace directory.
-  `run` in `buf push`.


private/bufpkg/buflock/lock_file.go
```diff
@@ -52,8 +54,15 @@ func readConfig(ctx context.Context, readBucket storage.ReadBucket) (_ *Config,
 			return nil, fmt.Errorf("failed to unmarshal lock file at %s: %w", V1Version, err)
 		}
 		config := &Config{}
+		var isB3DigestFound bool
 		for _, dep := range externalConfig.Deps {
 			config.Dependencies = append(config.Dependencies, DependencyForExternalConfigDependencyV1(dep))
+			if !isB3DigestFound && strings.HasPrefix(dep.Digest, "b3-") {
```
oliversun9, line 60: This only checks b3, perhaps we should check for b1 as well?

oliversun9: The mapper functions checks b1 as well https://github.com/bufbuild/buf/blob/main/private/bufpkg/buflock/buflock.go#L90

private/bufpkg/bufmodule/bufmodule.go
```diff
@@ -211,10 +212,11 @@ func ModuleWithWorkspaceDirectory(workspaceDirectory string) ModuleOption {
 // from a lock file in the read bucket.
 func NewModuleForBucket(
 	ctx context.Context,
+	logger *zap.Logger,
```
bufdev, line 215: It's not OK to take a `logger` in the constructor for `Module` here - we don't take loggers in entity constructors. You'll want to figure out how to test for this somewhere else.

bufdev (commented):
The proliferation of `loggers` here would lead me to think there's a cleaner way to do this, we're adding `logger` in places we don't want it. Why not just read the `lock` file at some top level and print a warning if b3 is found, as opposed to weaving this through all the logic? A `buflock.CheckDeprecatedDigests` function you can call?

private/bufpkg/bufmodule/bufmodule.go
oliversun9: Added a `buflock.CheckDeprecatedDigests`, see PR description

bufdev (approved):
Yea this looks way better :-) Are you sure these are the only places we need to call it?

oliversun9:
> Yea this looks way better :-) Are you sure these are the only places we need to call it?

I think all commands that read a `buf.lock` are covered. I updated the PR description to give the full list of commands where this warning can be displayed.

saquibmian (commented):
There's also `buf alpha sync` [here](https://github.com/bufbuild/buf/blob/ee6ddfba2119d4293ae4d8c2220b91ee9f9060a2/private/buf/bufsync/syncer.go#L599). That one's a little tricky because we scan commits in the past and we don't want to show the warning for all of them. We probably want to warn based on some timestamp? If commit was created after Nov 1st? I don't know how @bufdev feels about that, though.

oliversun9:
> There's also `buf alpha sync` [here](https://github.com/bufbuild/buf/blob/ee6ddfba2119d4293ae4d8c2220b91ee9f9060a2/private/buf/bufsync/syncer.go#L599). That one's a little tricky because we scan commits in the past and we don't want to show the warning for all of them. We probably want to warn based on some timestamp? If commit was created after Nov 1st? I don't know how @bufdev feels about that, though.

What if we use a global variable to keep track of whether this warning has been printed, so that we only print it once? (or `sync.Once`)

bufdev:
> What if we use a global variable

Mutable global variables to signal state are banned

doriable:
> That one's a little tricky because we scan commits in the past and we don't want to show the warning for all of them. We probably want to warn based on some timestamp? If commit was created after Nov 1st?

Kind of inclined to say that if you're using `buf sync`, you shouldn't see warnings of this nature at all. And let's say we implement "after a certain date (e.g. Nov. 1)", that commit will still be in their history, so even if they fix the issue, if they ever resync to a different repo, for example, they'll see it. I might leave `buf sync` out of this one, tbh.

saquibmian:
> if you're using `buf sync`, you shouldn't see warnings of this nature at all

The point of this is to give people an indication that they should migrate off of b3 digests. How do you think we should do that for repositories using `buf sync`?

As an alternative to all of this, we could instead show the warning in the BSR UI? It would be visible only to org members/admins. They are more likely to see it anyway, as users aren't in the habit of seeing warnings in CI runs.

doriable:
> As an alternative to all of this, we could instead show the warning in the BSR UI? It would be visible only to org members/admins. They are more likely to see it anyway, as users aren't in the habit of seeing warnings in CI runs.

Yeah, I think I'm more in agreeance with that sort of approach, because yeah, exactly, dealing with warnings in a CI run that aren't outright failures are often not priorities. And also, it's something they would need to handle by checking updates into source anyway (same with any dependency upgrade).

saquibmian:
We chatted about this, and decided that we're not going to warn from sync, largely because it's highly likely that no one sees those anyway. The high value places to warn are `build` and `lint`, which are done.

Separately, we will consider a warning in the BSR UI, visible to repository members.

saquibmian (approved):

PR #2524, opened 2023-10-30
Add VCS commit as reference type

doriable:
This adds VCS commit as one of the reference types returned
by the `ReferenceService`. If we can differentiate the VCS commit
as a type, then we can use `GetReferenceByName` for sync and break
our dependency on `LabelService`, which will allow us to stop exposing labels.


unmultimedio (approved):

bufdev (approved):

saquibmian (commented):
I forgot to submit my pending review...

Issue #2513, opened 2023-10-25
Buf cannot resolve dependencies when we use workspaces


shuheiktgw:
Hi, team 👋  Thank you so much for your kind support. We just recently started using buf, but one problem we face is if we use a workspace that depends on another workspace, buf seems to fail to resolve the dependencies. I've created a repository to reproduce the problem.

- https://github.com/shuheiktgw/buf-test

If you run `buf generate` at the root, it seems buf fails to resolve the dependency to google/rpc/code.proto. Would you guide us on what is wrong with our configuration? I found a similar issue, https://github.com/bufbuild/buf/issues/687, but it just stopped using a workspace to avoid the issue. 

```
buf generate 
proto_deps/shuheiktgw/test1/v1/message.proto:6:8:google/rpc/code.proto: does not exist
```

I'm currently using buf 1.27.1.

```
buf --version
1.27.1
```

Thanks, in advance!

shuheiktgw:
OK, I think I've finally figured out what was wrong with the configuration (but let me leave it open to confirm if my understanding is correct). What I needed to do was

1. Create `buf.build/shuheiktgw/test1` repository, and push the latest changes to create a commit
2. Declear the following dependencies in proto/buf.yaml and run `buf mod update` to create buf.lock

```
+deps:
+  - buf.build/shuheiktgw/test1
```

Then `buf generate` at the root works just fine. I thought workspaces work fine without creating actual repositories, but IIUC, buf needs them to create a correct buf.lock and resolve dependencies.

timostamm:
Hey @shuheiktgw, I can confirm that your solution is correct. We're looking to improve the behavior in the future.

doriable:
Hi, so this behaviour is greatly improved for `v2` modules and workspaces: https://buf.build/docs/configuration/v2/buf-yaml
Dependencies for `v2` workspaces is now defined across the entire workspace, so it should make it much easier to resolve.

Going to close this since migrating to `v2` should address this issue, migration guide available here: https://buf.build/docs/migration-guides/migrate-v2-config-files. Please feel free to re-open if there are questions and/or concerns!

PR #2529, opened 2023-10-31
DO NOT MERGE: Fix issue where transitive dependencies in a workspace were not added to ModuleFileSet

bufdev:
Fixes https://github.com/bufbuild/buf/issues/2513.

Issue: does not deal with transitive deps with different commits for same module.


private/buf/bufcli/bufcli.go
```diff
@@ -404,7 +404,7 @@ func NewWireImageConfigReader(
 		storageosProvider,
 		NewFetchReader(logger, storageosProvider, runner, moduleResolver, moduleReader),
 		bufmodulebuild.NewModuleBucketBuilder(),
-		bufimagebuild.NewBuilder(logger, moduleReader),
+		bufimagebuild.NewBuilder(logger, moduleReader, bufmoduleref.NewModulePinResolver(clientConfig)),
```
bufdev, line 407: Can you match the style in the codebase and either have all args on one line, or all args separated by newlines?

private/bufpkg/bufmodule/bufmoduleref/bufmoduleref.go
```diff
@@ -327,6 +329,46 @@ func NewProtoModulePinsForModulePins(modulePins ...ModulePin) []*modulev1alpha1.
 	return protoModulePins
 }
 
+// ModulePinResolver is a resolver for module pins.
+type ModulePinResolver interface {
```
bufdev, line 333: We should not be adding this to `bufmoduleref` - if anywhere, it would go in `bufmodule`, and probably just be another method on `ModuleResolver`.

saquibmian:
I see you rebased off of main and now `TestFormatEquivalence`, I'm not entire sure why though, at a first glance it's not clear why my changes would affect that test. I'll need to dig into this.

saquibmian:
Figured it out; in the past we weren't properly adding workspace deps, but not we are. The "pre-formatted" image has a workspace (because it's in the repo), whereas the "formatted" image does not have a workspace (because it's in a temp dir). So naturally, the images are different. I'll fix the test, but all the testdata implicitly having a workspace is bad...

saquibmian:
@doriable @oliversun09 @jchadwick-buf requested review from you all, @bufdev is busy for the rest of the week. Take a look whenever you get a chance, please :)

Don't worry about the Windows test, I'll fix it up, it's some strange path issue.

private/bufpkg/bufmodule/bufmodulebuild/module_file_set_builder.go
```diff
@@ -56,56 +60,153 @@ func (m *moduleFileSetBuilder) build(
...
+	ctx context.Context,
+	module bufmodule.Module,
+	workspace bufmodule.Workspace,
+) ([]bufmodule.Module, error) {
+	// We do a first pass to collect all module pins that we are going to resolve
+	// through the module reader. If there are any conflicts, we will need to go
+	// through dependency resolution.
+	var (
+		dependencyModulePins              []bufmoduleref.ModulePin
+		seenDependencyModulePinIdentities = make(map[string]bufmoduleref.ModulePin)
```
oliversun9, line 94: nit: perhaps something like `seenDependencyIdentityStringToPin` to indicate it's a map?

oliversun9 (approved):

saquibmian:
@oliversun9 any idea why the windows test is failing? I've spent some time looking and I can't seem to track it down. Not being able to run it locally makes it challenging.

oliversun9:
> @oliversun9 any idea why the windows test is failing? I've spent some time looking and I can't seem to track it down. Not being able to run it locally makes it challenging.

I notice that in these tests, a workspace is always built -- this is because the buf.work.yaml at the root of bufbuild/buf repository. This is so far the only way I can think of in which this PR can affect the failing test. I will keep looking, but just post this update in case it rings a bell.

saquibmian:
Declining this PR in favour of bufmod, which re-architects how modules are constructed and images are built.

PR #2532, opened 2023-11-01
Add remaining RPCs needed for sync

saquibmian:
All of these RPCs are congruent with v1beta1.


proto/buf/alpha/registry/v1alpha1/reference.proto
```diff
@@ -63,7 +63,7 @@ message RepositoryVCSCommit {
   // immutable
   google.protobuf.Timestamp create_time = 2;
   // The name of the VCS commit, e.g. for Git, it would be the Git hash.
-  string hash = 3;
+  string name = 3;
```
saquibmian, line 66: This was initially name, we changed it to hash because that better represented what it was, but I changed it back to `name` because upon integration in `core` I found out that the frontend expects all references to have a `name`. I'd rather we go through and refactor that as part of the v1beta1 effort, if we need to, so changing it back for now.

proto/buf/alpha/registry/v1alpha1/repository_branch.proto
```diff
@@ -67,6 +87,27 @@ message ListRepositoryBranchesResponse {
...
 }
 
+message ListRepositoryBranchesByReferenceRequest {
+  // The ID of the repository whose branches should be listed.
+  string repository_id = 1;
+  uint32 page_size = 2;
+  // The first page is returned if this is empty.
+  string page_token = 3;
+  // The reference by which to filter the returned branches.
+  oneof reference {
```
saquibmian, line 97: I explicitly didn't make this `string reference` because we need a guarantee that the reference we are requesting is a VCSCommit

doriable (approved):

bufdev (approved):

unmultimedio (approved):

PR #2534, opened 2023-11-02
Upgrade to protoc 25.0

bufdev:
We haven't done a `protoc` upgrade in a while. This also upgrades the windows dependencies.

@jhump I don't know if this just works, let me know if you think there's any issues with doing this.


private/bufpkg/bufimage/bufimageutil/testdata/options/Files.txtar
```diff
@@ -155,7 +245,7 @@ message FileDescriptorProto {
   repeated int32 public_dependency = 10;
   repeated int32 weak_dependency = 11;
   optional string syntax = 12;
-  optional string edition = 13;
+  optional Edition edition = 14;
```
jhump, line 248: This seems awkward to pull in w/out also pulling in c0928378910332de79950200c163ee07b02792c3.

(That commit is from a branch I sent you last week, just as a preview, since it breaks source-compatibility and forces a change in `image.proto`. This branch: https://github.com/bufbuild/buf/compare/jh/strip-source-only-options

I don't know if it's a good idea to pull in protoc and this descriptor.proto change w/out the corresponding Go runtime changes 🤷 

I had hoped to bring in the runtime changes along with the latest protocompile, all at the same time. If we don't have the latest protocompile, the current version doesn't know to _disallow_ the use of the `features` field in options when the syntax is proto2 or proto3, so it means people could inadvertently play around with features in a way that will be broken whenever we _do_ upgrade protocompile.

bufdev: I'm going to pull in that commit from protobuf-go here :-)

make/buf/all.mk
```diff
@@ -34,7 +35,7 @@ LICENSE_HEADER_YEAR_RANGE := 2020-2023
 LICENSE_HEADER_IGNORES := \/testdata enterprise
 PROTOVALIDATE_VERSION := v0.5.1
 # Comment out to use released buf
-BUF_GO_INSTALL_PATH := ./cmd/buf
+#BUF_GO_INSTALL_PATH := ./cmd/buf
```
jhump, line 38: What's this about?

jhump (approved):
👍

make/buf/all.mk
bufdev: Whoops! Fixed.

proto/buf/alpha/image/v1/image.proto
```diff
@@ -48,7 +48,7 @@ message ImageFile {
   optional google.protobuf.FileOptions options = 8;
   optional google.protobuf.SourceCodeInfo source_code_info = 9;
   optional string syntax = 12;
-  optional string edition = 13;
+  optional google.protobuf.Edition edition = 14;
```
rdesgroppes, line 51: Shouldn't be `13` be now `reserved`?

PR #2535, opened 2023-11-02
Change `buf alpha package` -> `buf alpha sdk`

stefanvanburen:
We've recently updated the language around this concept in the BSR - update it on the CLI as well.

Because this is an alpha command, I didn't bother to forward the old invocation to the new one, but I'm happy to add that. Also happy to add a CHANGELOG entry, but it doesn't look like we've typically mentioned changes to `alpha` commands in the past, with the exception of the removal of `alpha registry token create`.

There's a few other spots where we mention remote packages in comments or as field names in protos, but I don't think we want to include updates to those in the scope of this work?

### DIFF
--- a/private/buf/bufcli/bufcli.go
+++ b/private/buf/bufcli/bufcli.go
@@ -881,20 +881,20 @@ func ValidateErrorFormatFlagLint(errorFormatString string, errorFormatFlagName s
 
 // PackageVersionShortDescription returns the long description for the <package>-version command.
 func PackageVersionShortDescription(name string) string {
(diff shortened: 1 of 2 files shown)


pkwarren:
> There's a few other spots where we mention remote packages in comments or as field names in protos, but I don't think we want to include updates to those in the scope of this work?

We can make those in another PR - it appears to be contained in plugin_curation.proto and bufpluginconfig.go docs so still good to update.

pkwarren (approved):
Verified commands locally - looks good.

bufdev (approved):

PR #2536, opened 2023-11-02
Add instruction on how to regenerate convert test data files

cyinma:
these files shouldn't be regenerated to often as they are meant to be the wire format, hence not adding this as part of `make generate`, but adding an instruction in case we do need to regenerate / others want to know how these are generated


private/buf/cmd/buf/command/convert/testdata/convert/bin_json/duration.txtpb
```diff
@@ -1 +1 @@
-seconds:  3600
+seconds: 3600
```
cyinma, line 1: for some reason there are diff for txtpb format re: spaces, but shouldn't has any actual affect

bufdev (approved):

PR #2538, opened 2023-11-03
Add audit log event proto for user auto merged event

cyinma:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -343,6 +345,8 @@ message PayloadUserLoggedIn {}
 
 message PayloadUserLoggedOut {}
 
+message PayloadUserMerged {}
```
unmultimedio, line 348: Could you share some context of what is the event about? Is a user merging with another user?

bufdev: Let's get in the habit of adding documentation for any new type as well, even though the previous payload types dont have it.

```
// This message represents ... (3-5 sentences on what this is)
```

cyinma: chatted offline with @unmultimedio , this audit event is for the case when a new user from a new IdP login, and we automatically merge it to an existing user from the old IdP. renamed the event and message to clarify a bit and added docs

unmultimedio: When looking at this event, we'll see that another user from _some_ IdP got merged into, with no additional info... I know subject is probably not needed here, but what about at least the IdP provider id or name? So at least we know that for user John Doe we recognized and automerged his IdP user from his ID in whatever IdP provider he used?

cyinma: I don't think we need to log those information? basically we wouldn't know which IdP created the existing user, and logging the current IdP doesn't seems very useful

bufdev (approved):

unmultimedio (approved):

saquibmian (approved):

PR #2544, opened 2023-11-06
Add initial Python Repository config support

stefanvanburen:
I think the only additional key we may need here is to specify the base plugin for `stub-only` packages to satisfy:

> In addition, stub-only distributions SHOULD indicate which version(s) of the runtime package are supported by indicating the runtime distribution’s version(s) through normal dependency data.

Ref: https://peps.python.org/pep-0561/#stub-only-packages

But we can do that additively.


private/bufpkg/bufplugin/bufplugin.go
```diff
@@ -250,10 +258,56 @@ func ProtoRegistryConfigToPluginRegistry(config *registryv1alpha1.RegistryConfig
...
+	}
+	switch protoPythonConfig.GetPackageType() {
+	case registryv1alpha1.PythonPackageType_PYTHON_PACKAGE_TYPE_RUNTIME:
+		pythonConfig.PackageType = "runtime"
+	case registryv1alpha1.PythonPackageType_PYTHON_PACKAGE_TYPE_STUB_ONLY:
+		pythonConfig.PackageType = "stub-only"
+	default:
+		return nil, fmt.Errorf("unknown package type: %v", protoPythonConfig.GetPackageType())
+	}
+	if runtimeLibraries := protoPythonConfig.GetRuntimeLibraries(); runtimeLibraries != nil {
```
pkwarren, line 283: We could probably simplify this to just range over `protoPythonConfig.GetRuntimeLibraries()`.

private/bufpkg/bufplugin/bufpluginconfig/config.go
```diff
@@ -323,6 +334,29 @@ func swiftExternalDependencyToDependencyConfig(externalDep ExternalSwiftRegistry
...
 }
 
+func newPythonRegistryConfig(externalPythonRegistryConfig *ExternalPythonRegistryConfig) (*PythonRegistryConfig, error) {
+	if externalPythonRegistryConfig == nil {
+		return nil, nil
+	}
+	var dependencySpecifications []string
+	for _, externalDependencySpecification := range externalPythonRegistryConfig.Deps {
+		if externalDependencySpecification == "" {
+			return nil, fmt.Errorf("python registry config cannot have empty dependency specification")
```
pkwarren, line 344:
```suggestion
			return nil, fmt.Errorf("python registry config cannot have an empty dependency specification")
```

pkwarren (approved):

private/bufpkg/bufplugin/bufplugin.go
stefanvanburen: good catch, simplified this in the other case but missed it here. https://github.com/bufbuild/buf/pull/2544/commits/9d9709603e7ea4df7f7a9b94c1e6b259c40e1781

pkwarren (approved):

PR #2545, opened 2023-11-06
Expose formatting of ast.FileNode

Alfus:
So individual files can be formatted in memory.

### DIFF
--- a/private/buf/bufformat/bufformat.go
+++ b/private/buf/bufformat/bufformat.go
@@ -16,11 +16,13 @@ package bufformat
 
 import (
 	"context"
+	"io"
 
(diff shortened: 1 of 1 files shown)


doriable (approved):

private/buf/bufformat/bufformat.go
```diff
@@ -66,3 +68,8 @@ func Format(ctx context.Context, module bufmodule.Module) (_ storage.ReadBucket,
 	}
 	return readWriteBucket, nil
 }
+
+func FormatFileNode(fileNode *ast.FileNode, dest io.Writer) error {
```
bufdev, line 72: No godoc, reverses order of arguments without an explanation, PR has no context as to why this is needed

PR #2547, opened 2023-11-06
Allow formatting of an individual ast.FileNode

Alfus:
Avoiding the need to materializing the .proto file being formatted, or a module to contain the file (or a readbucket to contain the result). If a file can parse, it can be formatted. For example, the current vscode-buf extension writes a temp file to format the in memory contents: https://github.com/bufbuild/vscode-buf/blob/6a083e4c4d21235643da65ddbed585d7a391da55/src/formatter.ts#L47


private/buf/bufformat/bufformat.go
```diff
@@ -66,3 +68,9 @@ func Format(ctx context.Context, module bufmodule.Module) (_ storage.ReadBucket,
 	}
 	return readWriteBucket, nil
 }
+
+// Formats the given file node, writing to dest.
+func FormatFileNode(dest io.Writer, fileNode *ast.FileNode) error {
```
bufdev, line 73: In your usage of `FormatFileNode`, are you effectively duplicating this logic elsewhere?

```
		fileNode, err := parser.Parse(moduleFile.ExternalPath(), moduleFile, reporter.NewHandler(nil))
			if err != nil {
				return err
			}
```

If so, it would seem to make sense to make this signature `FormatFile(output io.Writer, file io.Reader, filePath string) error` and put the parsing within the function.

Regardless, we'll want to update `Format` to use this function directly.

We also likely want to rename `Format` to `FormatModule`.

Alfus: My usage uses protocompile library directly to re-parse the file on every ~keystroke without going through module related code (to provide live parse feedback), so I always have the fileNode, but not necessarily a full module.

bufdev: Are you using the `*ast.FileNode` outside of invoking this function? As in is this the only place you use the `*ast.FileNode`, or do you have other uses for it outside of format?

Alfus: I have lots of other use cases for *ast.FileNode, its the basis for for most of the LSP

bufdev (approved):

PR #2552, opened 2023-11-07
Update release workflow

oliversun9:
This updates the CLI release workflow. Releasing CLI itself (not including downstream repos) now has three separate actions, each requiring some level of manual approval.
1. Manually trigger the 'Create Release PR' action and the action will create a PR that updates the changelog and version in `bufcli.go`.
2. Once the the first PR is merged, the second action is triggered. This action builds and signs the new CLI and creates a draft release.
3. Once the draft release is published (approved), the third action is triggered and creates a go-back-to-development PR that updates the version and changelog.

This also removes third party actions in the release workflow.


.github/actions/create-pr-action/action.yaml
```diff
@@ -0,0 +1,42 @@
+name: Create a PR
```
oliversun9, line 1: This replaces https://github.com/peter-evans/create-pull-request

.github/workflows/back-to-development.yaml
```diff
@@ -0,0 +1,54 @@
...
+          - No changes yet.\\
+          
+          }" CHANGELOG.md
+          sed -i "/^Initial beta release.$/ {
+          N;
+          a\
+          [Unreleased]: https://github.com/bufbuild/buf/compare/${{ env.RELEASED_VERSION_WITH_V }}...HEAD
+          }" CHANGELOG.md
+      - name: update home brew badge
+        run: make updatehomebrewbadge VERSION=${{ env.RELEASED_VERSION_WITH_V }}
```
oliversun9, line 42: I didn't test this in my own repo, but it runs in the same relative order as before. @doriable does this look correct?

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-      - v*
-  release:
-    types: [created]
-
-env:
-  VERSION: ${{ github.event.inputs.version }}
-  APP_ID: 251311
-
-jobs:
-  prepare:
```
oliversun9, line 24: This is now `create-release-pr.yaml`

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-    runs-on: ubuntu-latest
-    if: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.version != '' }}
-    steps:
-      - name: Generate token
-        id: generate_token
-        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a
```
oliversun9, line 30: This isn't always needed, we can simply use `secrets.GITHUB_TOKEN` in most cases. One exception is `create-release-pr.yaml`, where _the action_ creates a PR and another checks needs to be triggered by this PR. In this case, we can use the official github action for this purpose.

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-            {"contents": "write", "pull_requests": "write"}
-      - name: Checkout repository code
-        uses: actions/checkout@v4
-        with:
-          token: ${{ steps.generate_token.outputs.token }}
-      - uses: actions/setup-go@v4
-        with:
-          go-version: "1.21.x"
-      - name: Install Buf
-        run: make installbuf
```
oliversun9, line 44: Do we _really_ need to install buf here? I suppose it checks whether it builds, but regular ci.yaml should cover that.

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-        run: |
-          files=`(git fetch origin main:main) && (git diff --name-only main)`
-          for file in $files; do
-            if [ "$file" = "CHANGELOG.md" ]; then
-              exit 0
-            fi
-          done
-          echo ERROR: CHANGELOG has not been updated
-          exit 1
-  tag:
```
oliversun9, line 84: This isn't needed anymore since you pass a tag in `gh release create`

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-          body: |
-            Release prepared for ${{env.VERSION}}
-
-            Reminder: Update the changelog
-          token: ${{ steps.generate_token.outputs.token }}
-      - name: Slack Notification
-        run: |
-          jq --null-input '{ text: "BufCLI Release v${{env.VERSION}} has started: ${{ steps.cpr.outputs.pull-request-url }}" }' \
-          | curl -sSL -X POST -H 'Content-Type: application/json' -d @- '${{ secrets.SLACK_RELEASE_NOTIFICATION_WEBHOOK }}'
-  verify:
```
oliversun9, line 65: This is now `verify-changelog.yaml`

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-          token: ${{ steps.generate_token.outputs.token }}
-          fetch-depth: 0
-      - name: Tag Release
-        run: |
-          git config --global user.password ${{ steps.generate_token.outputs.token }}
-          git tag -d ${{env.VERSION}} 2> /dev/null || echo 'local ref does not exist'
-          git push origin :${{env.VERSION}} 2> /dev/null || echo 'remote ref does not exist'
-          git tag ${{env.VERSION}}
-          git push origin ${{env.VERSION}}
-  perform:
```
oliversun9, line 114: This is now `build-and-draft-release.yaml`

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-          body: Release complete for ${{github.ref_name}}
-          token: ${{ steps.generate_token.outputs.token }}
-      - name: Output Version
-        id: output_version
-        uses: actions/github-script@v6
-        with:
-          script:
-            core.setOutput('version', "${{github.ref_name}}".replace("v", ""));
-  trigger-maven-update:
-    runs-on: ubuntu-latest
```
oliversun9, line 209: These triggers are removed and we will manually released these repos

.github/workflows/back-to-development.yaml
```diff
@@ -0,0 +1,54 @@
...
+          echo "RELEASED_VERSION_WITH_V=v${RELEASED_VERSION_WITH_V##Version*=*\"}" >> $GITHUB_ENV
+      - name: Set NEXT VERSION variable from tag
+        run: |
+          NEXT_VERSION=$(echo ${{ env.RELEASED_VERSION_WITH_V }} | awk -F. -v OFS=. '{$NF += 1 ; print}')
+          echo "NEXT_VERSION=${NEXT_VERSION:1}-dev" >> $GITHUB_ENV
+      - name: Update version
+        run: make updateversion VERSION=${{ env.NEXT_VERSION }}
+      - name: Unrelease changelog
+        run: |
+          sed -i "/^# Changelog/ {
```
oliversun9, line 28: This find the line that says `# Changelog` and inserts content two lines below this line. This isn't a complex expression, but newline and whitespace behavior seem to be different between gnu-sed and mac sed. Therefore, it is not added as a make target.

I wish we can update makego to make `SED_I` `gsed -i` on mac, that way we can guarantee consistent behavior across mac and linux. However, I suspect gsed isn't built-in to GitHub's mac runners and if a CI workflow running on mac needs to use gsed it would have to install it first.

.github/workflows/build-and-draft-release.yaml
```diff
@@ -0,0 +1,57 @@
+name: Build and Draft Release
+on:
+  pull_request:
+    types: [closed]
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: The released version without 'v'. For example, 1.0.0.
```
oliversun9, line 9: If this fails, it can be restarted with manual trigger

.github/workflows/back-to-development.yaml
```diff
@@ -0,0 +1,54 @@
+name: Go back to Development
+on:
+  workflow_dispatch:
```
oliversun9, line 3: If this fails, it can be restarted with manual trigger. However, no input is needed because in a later step, version is read from bufcli.go

.github/actions/changelog-action/entrypoint.sh
```diff
@@ -1,21 +0,0 @@
-#!/usr/bin/env bash
```
oliversun9, line 1: This is now done in `draftrelease.bash`

.github/workflows/build-and-draft-release.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  draft_release:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release')) }}
+    steps:
+      - name: Set VERSION variable
+        # The head ref looks like release/v1.0.0, and we need to trim the string up to the `/v`.
+        run: |
+          VERSION="${{ github.event.inputs.version || github.head_ref}}"
```
oliversun9, line 18: This is handled in action so that the script does not need to know about what triggers it

.github/workflows/back-to-development.yaml
```diff
@@ -0,0 +1,30 @@
...
+  APP_ID: 251311
+
+jobs:
+  post-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository code
+        uses: actions/checkout@v4
+      - name: Get GitHub App Token
+        uses: actions/create-github-app-token@v1
```
oliversun9, line 17: This is needed because we want a CI check to check that this PR has changes in CHANGELOG.md. To trigger that check, this PR needs to be opened with an App token.

.github/workflows/release.yaml
```diff
@@ -1,375 +0,0 @@
...
-      - "release/**"
-      - "next/**"
-    tags:
-      - v*
-  release:
-    types: [created]
-
-env:
-  VERSION: ${{ github.event.inputs.version }}
-  APP_ID: 251311
```
oliversun9, line 21: We should perhaps store this ID in a secret to be safer? It's been here for probably a year by now, but it can be done in a follow up.

make/buf/all.mk
```diff
@@ -135,13 +135,6 @@ bufgeneratesteps:: \
 bufrelease: $(MINISIGN)
 	DOCKER_IMAGE=golang:1.21-bullseye bash make/buf/scripts/release.bash
 
-# We have to manually set the Homebrew version on the Homebrew badge as there
-# is no badge on shields.io for Homebrew packages outside of homebrew-core
-
```
oliversun9, line 140: We don't manually set the badge version anymore. It's now dynamic, see https://github.com/bufbuild/buf/blob/6befbe40183d85d55677092107ff1f1f1ad56a4e/README.md?plain=1#L9

make/buf/scripts/homebrewversion.bash
```diff
@@ -1,8 +0,0 @@
-#!/usr/bin/env bash
```
oliversun9, line 1: The make target is removed in the same PR

.github/workflows/build-and-draft-release.yaml
```diff
@@ -0,0 +1,41 @@
...
+    types: [closed]
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+        description: The released version without 'v'. For example, 1.0.0.
+jobs:
+  draft_release:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'workflow_dispatch' || (github.event_name == 'pull_request' && github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release')) }}
```
oliversun9, line 13: `github.head_ref` is potentially untrusted input, and is used later on. It's a potential vector for attacks since we expand it in-line. However, if the PR has been merged (approved by one of us at buf), the branch name is probably safe.

doriable (approved):

PR #2553, opened 2023-11-08
Add `GetClusterUsage` to AdminService

gilwong00:
This add a new endpoint `GetClusterUsage` to the `AdminService` This will populate the data for the new on prem usage page


proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -70,6 +70,22 @@ message CollisionSource {
   string path = 3;
 }
 
+// MonthlyUsage contains the total amount of message, enum and services types
+// for every organizations repositories within a single tenant BSR instance
+// for a given month and year.
+message MonthlyUsage {
+  // the numerical value of a month
```
bufdev, line 77: Use complete sentences - `The month, between 1 and 12.` Same whole way down.

bufdev (approved):

PR #2554, opened 2023-11-09
Extract all `bufsync.*Func`s into a `bufsync.Handler`

saquibmian:
This also improves the testing to be blackbox; we invoke sync and assert on the end-state.

### DIFF
--- a/private/buf/bufsync/bufsync_test.go
+++ b/private/buf/bufsync/bufsync_test.go
@@ -17,6 +17,7 @@ package bufsync_test
 import (
 	"bytes"
 	"context"
+	"errors"
 	"io"
 	"os"
 	"path"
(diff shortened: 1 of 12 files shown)


private/buf/bufsync/commits_to_sync_test.go
```diff
@@ -66,101 +64,37 @@ func TestCommitsToSyncWithNoPreviousSyncPoints(t *testing.T) {
...
-						moduleIdentity = moduleIdentityInHEAD
+						opts = append(opts, bufsync.SyncerWithModule(moduleDir, nil))
 					}
-					syncableCommits, err := testSyncer.branchSyncableCommits(
-						context.Background(),
-						moduleDir,
-						moduleIdentity,
-						tc.branch,
-						"", // no expected git sync point
+					handler := newMockSyncHandler()
```
unmultimedio, line 79: I'm confused why this test is not broken. The mock BSR sync point checker should be shared across test cases (ie out of the `for _, tc := range testCases` loop) so when syncing main, the other branches will pick up new sync points.

private/buf/bufsync/prepare_sync_test.go
```diff
@@ -76,19 +77,20 @@ func TestPrepareSyncDuplicateIdentities(t *testing.T) {
...
+				syncer, err := bufsync.NewSyncer(
+					zaptest.NewLogger(t),
+					bufsync.NewRealClock(),
+					repo,
+					storagegit.NewProvider(repo.Objects()),
+					newMockSyncHandler(),
+					opts...,
+				)
+				require.NoError(t, err)
+				prepareErr := syncer.Sync(context.Background())
```
unmultimedio, line 93:
```suggestion
				syncErr := syncer.Sync(context.Background())
```

unmultimedio (commented):
Thanks for this clean up!

private/buf/bufsync/commits_to_sync_test.go
saquibmian: This is totally fine. The repo looks like this:
```
| o-o----------o-----------------o (master)
|   └o-o (foo) └o--------o (bar)
|               └o (baz)
```
So `master` syncs 4 commits, then `foo` syncs 2, `bar` syncs 2, and `baz` syncs one.

If we change the order we'll get different counts, of course, but because I'm using `WithAllBranches()`, it's alphabetical order with git default first, so it syncs in that order.

aside: I will be changing this test in my logic refactor to actually explode out pairwise all the branch orders and assert that way. That would more correctly capture all the cases.

private/buf/bufsync/bufsync_test.go
```diff
@@ -89,3 +92,111 @@ func writeFiles(t *testing.T, directoryPath string, pathToContents map[string]st
...
+
+func (c *mockSyncHandler) BackfillTags(
+	ctx context.Context,
+	module bufmoduleref.ModuleIdentity,
+	alreadySyncedHash git.Hash,
+	author git.Ident,
+	committer git.Ident,
+	tags []string,
+) (string, error) {
+	c.tagsByHash[alreadySyncedHash.Hex()] = tags
```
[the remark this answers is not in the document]
saquibmian: I updated the logic to be more "correct" take a look.

private/buf/bufsync/commits_to_sync_test.go
saquibmian: Technically I only need to sync once, but I'm going to change this soon anyway, so the subsequent calls are all no-ops for now.

unmultimedio: If you're using `AllBranches` then you actually don't need to wrap this into multiple test cases, that's why it's not failing. For each test cases it's syncing all branches, and each time you're just asserting one branch. If you don't pass the `AllBranches` option and move the `handler := newMockSyncHandler()` declaration up top to where it was, the test should keep working as it was originally intended, which is sync once branch at a time and assert.

saquibmian: Yes, but see my comment above, I chose to keep it that way for now. Can change if you feel strongly, but I mainly wanted to minimize changes here to "prove" that nothing is broken.

saquibmian: I'll move the syncer down, actually, I'll need that anyway.

unmultimedio: The way it is written now it's redundant:

- sync all
- assert main
- sync all
- assert foo
- sync all
- assert bar
- sync all
- assert baz

That's why it was confusing to me, so what I suggest is option 1:

- sync main
- assert main
- sync foo
- assert foo
- sync bar
- assert bar
- sync baz
- assert baz

or option 2:

- sync all
- assert main
- assert foo
- assert bar
- assert baz

unmultimedio (approved):

PR #2558, opened 2023-11-10
Fix permission issue from checkout action

oliversun9:
The checkout action needs a token for `git push` to work.

Also reverts the `git config` change.


doriable (approved):

oliversun9:
This successfully triggered https://github.com/bufbuild/buf/pull/2557

PR #2559, opened 2023-11-10
Remove duplicated code for scaffolding git repository

saquibmian:


private/pkg/git/gittest/gittest.go
```diff
@@ -164,11 +201,16 @@ func runInDir(t *testing.T, runner command.Runner, dir string, cmd string, args
 		command.RunWithStderr(stderr),
 	)
 	if err != nil {
-		t.Logf("run %q", strings.Join(append([]string{cmd}, args...), " "))
-		_, err := io.Copy(os.Stderr, stderr)
-		require.NoError(t, err)
+		require.FailNow(
```
saquibmian, line 204: This improves the error message so that it's right in the test failures as opposed to stderr.
```
--- FAIL: TestCommitsToSyncWithNoPreviousSyncPoints (0.28s)
    gittest.go:204: 
                Error Trace:    /Users/saquib/dev/buf/private/pkg/git/gittest/gittest.go:204
                                                        /Users/saquib/dev/buf/private/pkg/git/gittest/repository.go:68
                                                        /Users/saquib/dev/buf/private/buf/bufsync/commits_to_sync_test.go:131
                                                        /Users/saquib/dev/buf/private/buf/bufsync/commits_to_sync_test.go:39
                Error:          git command failed: "git checkout bar": exit status 1
                Test:           TestCommitsToSyncWithNoPreviousSyncPoints
                Messages:       error: pathspec 'bar' did not match any file(s) known to git
```

private/pkg/git/gittest/gittest.go
```diff
@@ -34,9 +34,33 @@ const (
...
+	withOnlyInitialCommit bool
+}
+
+type ScaffoldGitRepositoryOption func(*scaffoldOptions)
+
+func ScaffoldGitRepositoryWithOnlyInitialCommit() ScaffoldGitRepositoryOption {
+	return func(opts *scaffoldOptions) {
+		opts.withOnlyInitialCommit = true
+	}
+}
```
unmultimedio, line 55: I think this should be the default option. What about just receiving the repo and then passing it to another func that adds all the extra stuff, and that's invoked by only the tests that need it?

saquibmian: I toyed with this, but the main thing is that the default scaffolding right now does things I don't want to expose yet, like `Push` and `AddWithX`. I could invert this to `WithSampleCommits`, but I'd still place the initial README commit there. What do you think?

unmultimedio: My main issue is that `WithSampleCommits` will only make sense to a bunch of func callers, not potentially all of them, and that's the reason why you wanted to remove the duplicated func, so everyone who needs a base git repo to play around can call this func.

What about having a runner in the caller side that can do the custom things you're thinking at the repo dir, while `Push` and `AddWithX` are ready to be exposed?

saquibmian: I could do `runCommand`, but I wanted to prevent access to the raw dir (a.,.a., write file). 

aside: storagegit also uses this sample data to assert on modules that were read.

unmultimedio: While the repo doesn't have all exported methods to let the callers do their thing (ie commit, push, write files, etc), then I think it's safe to pass the raw directory to the caller. It's a testing pkg anyway.

Non blocking on this, but I really think this should be handled by each caller and not by the scaffolding itself.

unmultimedio (approved):

private/pkg/git/gittest/gittest.go
saquibmian: Done!

private/buf/bufsync/commits_to_sync_test.go
```diff
@@ -105,42 +105,35 @@ func TestCommitsToSyncWithNoPreviousSyncPoints(t *testing.T) {
 func prepareGitRepoSyncWithNoPreviousSyncPoints(
 	t *testing.T,
 	runner command.Runner,
```
unmultimedio, line 107: You don't need this anymore then?

unmultimedio (approved):

private/pkg/git/gittest/gittest.go
```diff
@@ -34,9 +34,48 @@ const (
...
+
+func CommitWithExecutableFile(path string) CommitOption {
+	return func(opts *commitOpts) {
+		opts.executablePaths = append(opts.executablePaths, path)
+	}
+}
+
+type scaffoldOptions struct {
+	withOnlyInitialCommit bool
+}
```
unmultimedio, line 62: Same here, I don't think this option is used anywhere now.

saquibmian: Done

private/buf/bufsync/commits_to_sync_test.go
saquibmian: Done

unmultimedio (approved):
Thank you!!

PR #2563, opened 2023-11-10
Update `IsGitCommitSynced`

saquibmian:
Instead of looking for a label, we now look for a VCSCommit reference. Practically, this means `buf push --tag <commit-ish>` will no longer count as synced.


unmultimedio (approved):
🎉

PR #2569, opened 2023-11-10
Move protected branch determination to handler

saquibmian:
This is preparation for protected branches being asserted server side.

### DIFF
--- a/private/buf/bufsync/bufsync.go
+++ b/private/buf/bufsync/bufsync.go
@@ -175,13 +175,13 @@ type Handler interface {
 		hash git.Hash,
 	) (bool, error)
 
-	// GetModuleReleaseBranch is invoked before syncing, to gather release branch names for all the
-	// modules that are about to be synced. If the BSR module does not exist, the implementation should
-	// return `ModuleDoesNotExistErr` error.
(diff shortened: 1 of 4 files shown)


private/buf/bufsync/bufsync.go
```diff
@@ -175,13 +175,13 @@ type Handler interface {
 		hash git.Hash,
 	) (bool, error)
 
-	// GetModuleReleaseBranch is invoked before syncing, to gather release branch names for all the
-	// modules that are about to be synced. If the BSR module does not exist, the implementation should
-	// return `ModuleDoesNotExistErr` error.
-	GetModuleReleaseBranch(
+	// IsProtectedBranch is invoked when syncing branches to know if a branch's history is protected
+	// and must not diverge since the last sync. If an error is returned, sync will abort.
+	IsProtectedBranch(
```
unmultimedio, line 180: Nit: we could invoke just once per module identity and get `protectedBranches []string`

saquibmian: We could, but the flow isn't like that right now. We cache anyway, so this is fine.

unmultimedio (approved):

PR #2572, opened 2023-11-12
Update bufcas and slicesextended

bufdev:
This PR does a few things:

- Adds `Reduce` and `Count` functions to `slicesextended`.
- Introduces `ParseError` to `bufcas`. An `error` of type `*bufcas.ParseError` is returned from `bufcas.Parse.*` functions on unsuccessful parsing. This will allow us to understand the nature of an `error` returned from the new `ModuleKey.Digest()`, and handle situations such as https://github.com/bufbuild/buf/pull/2111. Instead of blanking out the error from a `buf.lock` file, we can instead return an error saying we did not understand the `DigestType`, and handle that accordingly in places where we don't support it.
- Adds some extra functions to `bufcas` with respect to conversion to/from repeated Protobuf types.
- Updates dependencies as well.


smallsamantha (approved):

Alfus (approved):

private/bufpkg/bufcas/digest.go
```diff
@@ -66,10 +67,16 @@ func (d DigestType) String() string {
 // ParseDigestType parses a DigestType from its string representation.
 //
 // This reverses DigestType.String().
+//
+// Returns an error of type *ParseError if thie string could not be parsed.
```
emcfarlane, line 71: Typo `thie` -> `this`|`the`

PR #2579, opened 2023-11-13
Upgrade OTEL to v1.20.0

pkwarren:
The latest version of OTEL requires embedding the
embedded.TracerProvider interface going forward in any implementations. Instead of implementing/extending TracerProvider, instead return a separate Closer.


private/pkg/observabilityzap/tracer_provider_closer.go
```diff
@@ -20,12 +20,15 @@ import (
...
 	sdktrace "go.opentelemetry.io/otel/sdk/trace"
 	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/embedded"
 )
 
 var _ trace.TracerProvider = &tracerProviderCloser{}
 var _ io.Closer = &tracerProviderCloser{}
 
 type tracerProviderCloser struct {
+	// https://pkg.go.dev/go.opentelemetry.io/otel/trace#hdr-API_Implementations
```
saquibmian, line 30: Otel is the strangest package. How is this interface implemented? It contains an unexported method, so my understanding is that it can _never_ be implemented outside of `otel/trace/embedded`.

pkwarren: I agree - pushed a follow up which will get us out of the business of implementing our own `TracerProvider` and instead just return a separate `io.Closer` to allow shutting down the tracer provider.

saquibmian (approved):

r-n-o:
@pkwarren thanks for merging this so quickly! This PR unblocks a security upgrade for projects using buf: https://github.com/advisories/GHSA-8pgv-569h-w5rw 🎉 

Do you have a sense for when the next release of `bufbuild/buf` will be?

pkwarren:
> Do you have a sense for when the next release of `bufbuild/buf` will be?

We haven't planned the next release yet - it is usually on demand as needed.

How are you consuming buf? If via a go module, you can just do a `go get ...@main` to pull the latest version (with this fix). The CLI binary itself has no dependencies on opentelemetry-go-contrib, so doesn't need to be rebuilt with a fix for https://github.com/advisories/GHSA-8pgv-569h-w5rw.

bufdev:
Of note, installing from sources is not recommended, we recommend you install from the binaries for exactly this reason.

r-n-o:
@pkwarren that's right, we're consuming via go module. `go get github.com/bufbuild/buf/cmd/buf@main` did the trick, thank you!

PR #2583, opened 2023-11-13
Upgrade go dependencies

pkwarren:
Update to the latest go dependencies. Pull in the latest protovalidate-go and remove pinned dependencies which are now updated upstream.


make/buf/all.mk
```diff
@@ -1,8 +1,5 @@
 GO_ALL_REPO_PKGS := ./cmd/... ./private/...
-# Need to pin until https://github.com/google/cel-go/pull/724 is resolved
```
pkwarren, line 2: Marked fixed in latest cel-go release.

bufdev (approved):

PR #2585, opened 2023-11-13
Remove `HandleReadModuleError`

saquibmian:
The logic is instead inlined because it's a core decision that Syncer should make.


private/buf/bufsync/syncer.go
```diff
@@ -651,16 +631,6 @@ func (s *syncer) backfillTags(
 	return nil
 }
 
-type readModuleOpts struct {
-	expectedModuleIdentity string
-}
-
-type readModuleAtOption func(*readModuleOpts)
-
-func readModuleAtWithExpectedModuleIdentity(moduleIdentity string) readModuleAtOption {
```
unmultimedio, line 660: All except this option lgtm, let's chat about that. We don't want any buildable module to sync to the module identity in HEAD.

private/buf/bufsync/bufsync.go
```diff
@@ -31,111 +31,8 @@ import (
...
-	//
-	// If this func returns `LookbackDecisionCodeSkip` for any `ReadModuleErrorCode`, then the syncer
-	// will stop looking when reaching the commit `r` because it already exists in the BSR, select `s`
-	// as the start sync point, and the synced commits into the BSR will be [s, t, x, y, z].
-	//
-	// If this func returns `LookbackDecisionCodeStop` for `ReadModuleErrorCodeModuleNotFound`, the
-	// syncer will stop looking when reaching the commit `u`, will select `v` as the start sync point,
-	// and the synced commits into the BSR will be [x, y, z].
-	HandleReadModuleError(err *ReadModuleError) LookbackDecisionCode
 	// InvalidBSRSyncPoint is invoked by Syncer upon encountering a module's branch sync point that is
```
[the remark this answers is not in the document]
saquibmian: Nah I want to kill this, too. Initially all of this came because we were logging a warning in the CLI and the suggestion was made that Syncer shouldn't do this, but this is a core part of Syncer now.

private/buf/bufsync/syncer.go
```diff
@@ -437,34 +451,29 @@ func (s *syncer) branchSyncableCommits(
...
-			logger.Debug("read module at commit failed, stop looking back in branch", zap.Error(readErr))
-			return git.ErrStopForEach
-		case LookbackDecisionCodeOverride:
-			logger.Debug("read module at commit failed, overriding module identity in commit", zap.Error(readErr))
-			if builtModule == nil {
-				return fmt.Errorf("cannot override commit, no built module: %w", readErr)
+		}
+		if builtModule.ModuleIdentity() == nil {
+			if _, hasOverride := s.modulesDirsToIdentityOverrideForSync[moduleDir]; !hasOverride {
+				logger.Debug("unnamed module, no override, skipping commit")
```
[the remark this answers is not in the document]
saquibmian: The logger already has it :)

private/buf/bufsync/syncer.go
saquibmian: Fixed

private/buf/bufsync/syncer.go
```diff
@@ -512,49 +527,30 @@ func (s *syncer) isGitCommitSynced(ctx context.Context, moduleIdentity bufmodule
 	return commitSynced, nil
 }
 
-// readModuleAt returns a module that has a name and builds correctly given a commit and a module
-// directory, or a read module error. If the module builds, it might be returned alongside a not-nil
-// error.
+// readModule returns a module that has a name and builds correctly given a commit and a module
+// directory. If the module builds, it might be returned alongside a non-nil error.
```
unmultimedio, line 531:
```suggestion
// directory.
```

That's no longer true.

private/buf/bufsync/syncer.go
```diff
@@ -437,34 +451,35 @@ func (s *syncer) branchSyncableCommits(
...
-			if builtModule == nil {
-				return fmt.Errorf("cannot override commit, no built module: %w", readErr)
+		}
+		if builtModule.ModuleIdentity() == nil {
+			if _, hasOverride := s.modulesDirsToIdentityOverrideForSync[moduleDir]; !hasOverride {
+				logger.Debug("unnamed module, no override, skipping commit")
+				return nil
+			}
+		}
+		if builtModule.ModuleIdentity().IdentityString() != targetModuleIdentity {
```
unmultimedio, line 472: This could panic if the built module has no identity but there's an override, because it'd never go into L469. You need to convert this into an `else if`.

private/buf/bufsync/syncer.go
```diff
@@ -619,14 +597,27 @@ func (s *syncer) backfillTags(
...
+			// no module, skip commit
 			return nil
+		} else if builtModule.ModuleIdentity() == nil {
+			if _, hasOverride := s.modulesDirsToIdentityOverrideForSync[moduleDir]; !hasOverride {
+				// no override for this module
+				return nil
+			}
+		} else if builtModule.ModuleIdentity().IdentityString() != moduleIdentity.IdentityString() {
+			if _, hasOverride := s.modulesDirsToIdentityOverrideForSync[moduleDir]; !hasOverride {
+				logger.Debug("module name doesn't match HEAD, no override, skipping commit")
```
unmultimedio, line 618: you could add a similar debug in L613 for unnamed with no override.

unmultimedio (changes requested):
Only one blocker comment

unmultimedio (approved):

PR #2593, opened 2023-11-15
Sync all tags on every run

saquibmian:


private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
```diff
@@ -104,39 +104,65 @@ func (h *syncHandler) IsGitCommitSynced(ctx context.Context, module bufmoduleref
...
+	if repoRes, err := repoService.GetRepositoryByFullName(ctx, connect.NewRequest(&registryv1alpha1.GetRepositoryByFullNameRequest{
+		FullName: moduleIdentity.Owner() + "/" + moduleIdentity.Repository(),
+	})); err != nil {
 		if connect.CodeOf(err) == connect.CodeNotFound {
 			// Repo is not created
-			return "", bufsync.ErrModuleDoesNotExist
+			return nil
 		}
-		return "", fmt.Errorf("attach git tags to module %q: %w", module.IdentityString(), err)
+		return fmt.Errorf("get repository for module identity: %w", err)
```
unmultimedio, line 121: This should never happen, When invoking SyncModuleTags we should have already attempted to create the repo. I'd throw the error here unconditionally.

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
```diff
@@ -104,39 +104,65 @@ func (h *syncHandler) IsGitCommitSynced(ctx context.Context, module bufmoduleref
...
+	} else {
```
unmultimedio, line 122: No need for this `else`

proto/buf/alpha/registry/v1alpha1/repository_tag.proto
```diff
@@ -65,6 +84,19 @@ message CreateRepositoryTagResponse {
   RepositoryTag repository_tag = 1;
 }
 
+message UpdateRepositoryTagRequest {
+  // The ID of the repository this tag belongs to.
+  string repository_id = 1;
+  // The name of the repository tag, to update e.g. "6e2e7f24718a76caa32a80d0e2b1841ef2c61403".
+  string name = 2;
+  // The name of the commit this tag should reference.
+  optional string commit_name = 3;
```
unmultimedio, line 93: In which scenario is this optional?

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
saquibmian: You'd prefer that I return `bufsync.ErrModuleDoesNotExist`? Sure, I can do that.

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
saquibmian: I need it because I don't hold on the `repoRes`.

proto/buf/alpha/registry/v1alpha1/repository_tag.proto
saquibmian: This is so that we're consistent with other `Update*` RPCs, where thing should be optional.
There could be other things to update here, concievably.

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
saquibmian: I got rid of that error, because Syncer doesn't need it, but I returned a different one.

unmultimedio: Sure, there's no need for a specific err but that's ok

proto/buf/alpha/registry/v1alpha1/repository_tag.proto
unmultimedio: Currently the only thing that can be updated in a tag is the commit its pointing to. Having it optional is super confusing. If we're leaving it optional, I guess in the backend if this is not passed this is a no-op?

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
unmultimedio: You're right, looks weird though, that the first conditional branch is always returning and still have an else. I guess it's because you don't want to have `repoRes` to have a big scope?

unmultimedio (approved):

private/buf/cmd/buf/command/alpha/repo/reposync/sync_handler.go
saquibmian: Yup

proto/buf/alpha/registry/v1alpha1/repository_tag.proto
saquibmian: Yes. This is also consistent with v1beta1: https://github.com/bufbuild/registry-proto/blob/d73fe14dfdc908ab9dffde2447f13b1c570098a1/buf/registry/module/v1beta1/tag_service.proto#L135

PR #2596, opened 2023-11-15
Clean up storage.PutOptions

bufdev:
Use an interface so these can only be built by the `storage` package, clean up docs, clean up meaning of 0.

### DIFF
--- a/private/pkg/storage/bucket.go
+++ b/private/pkg/storage/bucket.go
@@ -104,11 +104,15 @@ type PutOption func(*putOptions)
 // Some implementations of Put allow multi-part upload, and allow customizing the
 // chunk size of each part upload, or even disabling multi-part upload.
 //
-// # Setting a suggestedChunkSize of 0 says to suggest disable chunking
+// Setting a suggestedChunkSize of 0 says to suggest disable chunking
+// Negative values will be ignored.
 //
 // This is a suggestion, implementations may choose to ignore this option.
 func PutWithSuggestedChunkSize(suggestedChunkSize int) PutOption {
 	return func(putOptions *putOptions) {
+		if suggestedChunkSize < 0 {
+			return
+		}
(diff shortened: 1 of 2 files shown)


private/pkg/storage/bucket.go
```diff
@@ -58,24 +58,58 @@ type ReadBucket interface {
...
+// size to use by the Bucket implementation for this Put call.
+// Some implementations of Put allow multi-part upload, and allow customizing the
+// chunk size of each part upload, or even disabling multi-part upload.
+//
+// # Setting a suggestedChunkSize of 0 says to suggest disable chunking
+//
+// This is a suggestion, implementations may choose to ignore this option.
+func PutWithSuggestedChunkSize(suggestedChunkSize int) PutOption {
+	return func(putOptions *putOptions) {
+		putOptions.suggestedChunkSize = &suggestedChunkSize
```
pkwarren, line 112: Should we ignore negative values here?

private/pkg/storage/bucket.go
```diff
@@ -252,3 +290,29 @@ type nopReadWriteBucketCloser struct {
...
+func newPutOptions() *putOptions {
+	return &putOptions{}
+}
+
+func (p *putOptions) Atomic() bool {
+	return p.atomic
+}
+
+func (p *putOptions) SuggestedDisableChunking() bool {
+	return p.suggestedChunkSize != nil && *p.suggestedChunkSize == 0
```
pkwarren, line 308: If we don't prevent negative values here we should change the conditional from == to <= here.

private/pkg/storage/bucket.go
bufdev: Done

private/pkg/storage/bucket.go
bufdev: Done

pkwarren (approved):

PR #2600, opened 2023-11-17
Improve CI performance for Windows

jchadwick-buf:
Sometimes the existing workflow [runs quickly](https://github.com/bufbuild/buf/actions/runs/6896243827/job/18761936996) - sometimes it runs [incredibly slowly](https://github.com/bufbuild/buf/actions/runs/6905941659/job/18789829770).

This seems to come down to a disk I/O issue. The cache tarball will get downloaded very quickly, but then it will take minutes to untar. It's unclear why. I suspect there is something in particular wrong with the home directory in the Windows runner. In addition, `go test` itself seems to have unreliable performance characteristics.

Merely just moving the Go cache and path into the GitHub Actions `_temp` dir seems to vastly improve consistency. I ran the workflow multiple times to collect some runs:

- [**Attempt #1**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18792364497): 3m 2s, _no cache_
- [**Attempt #2**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18792644471): 3m 1s, cached
- [**Attempt #3**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18792750339): 2m 45s, cached
- [**Attempt #4**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18792871341): 4m 1s, cached
- [**Attempt #5**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18793063059): 2m 44s, cached
- [**Attempt #6**](https://github.com/bufbuild/buf/actions/runs/6906720017/job/18793181885): 2m 45s, cached

The cache restore times are reliably counted in seconds. There is still some run-to-run variance, but pathological behavior seems to have disappeared, both for `go test` and cache restore. While this doesn't bring performance quite up to par with the other workflows, it does seem to help quite a bit.

I'm really hoping I didn't miss anything here. Any reason why I may have tricked myself into thinking I solved the problem? :)

### DIFF
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -21,9 +21,11 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-go-
       - name: setup-go
-        uses: actions/setup-go@v4 # this contains a fix for Windows file extraction
+        uses: actions/setup-go@v4
         with:
           go-version: '1.21.x'
+          # This runs very slowly and inconsistently on Windows for undetermined reasons.
+          # We manually cache in the GitHub Actions _temp directory instead.
           cache: false
       - name: windows-cache
         uses: actions/cache@v3
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -7,10 +7,11 @@ jobs:
   test:
     env:
       DOWNLOAD_CACHE: 'd:\downloadcache'
-      # Improve performance by using GHA _temp directory.
-      # %USERPROFILE% seems to be really slow, especially for cache restores.
-      GOCACHE: 'd:\a\_temp\go\cache'
-      GOMODCACHE: 'd:\a\_temp\go\modcache'
+      # Improve performance by using D: drive.
+      # C: seems to be really slow, especially for cache restores.
+      GOPATH: 'd:\go\path'
+      GOCACHE: 'd:\go\cache'
+      GOMODCACHE: 'd:\go\modcache'
     runs-on: windows-latest
     steps:
       - name: checkout
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -7,26 +7,19 @@ jobs:
   test:
     env:
       DOWNLOAD_CACHE: 'd:\downloadcache'
-      GOPATH: 'd:\a\_temp\go\work'
+      # Improve performance by using GHA _temp directory.
+      # %USERPROFILE% seems to be really slow, especially for cache restores.
       GOCACHE: 'd:\a\_temp\go\cache'
+      GOMODCACHE: 'd:\a\_temp\go\modcache'
     runs-on: windows-latest
     steps:
       - name: checkout
         uses: actions/checkout@v4
-      - name: go-cache
-        uses: actions/cache@v3
-        with:
-          path: D:/a/_temp/go
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-v2
-          restore-keys: |
-            ${{ runner.os }}-go-
       - name: setup-go
         uses: actions/setup-go@v4
         with:
           go-version: '1.21.x'
-          # This runs very slowly and inconsistently on Windows for undetermined reasons.
-          # We manually cache in the GitHub Actions _temp directory instead.
-          cache: false
+          cache: true
       - name: windows-cache
         uses: actions/cache@v3
         with:
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -9,15 +9,17 @@ jobs:
       DOWNLOAD_CACHE: 'd:\downloadcache'
     runs-on: windows-latest
     steps:
+      - shell: pwsh
+        run: |
+          echo "GOPATH=D:/a/_temp/go/work" >> $env:GITHUB_ENV
+          echo "GOCACHE=D:/a/_temp/go/cache" >> $env:GITHUB_ENV
       - name: checkout
         uses: actions/checkout@v4
       - name: go-cache
         uses: actions/cache@v3
         with:
-          path: |
-            ~\AppData\Local\go-build
-            ~\go\pkg\mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          path: D:/a/_temp/go
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}-v2
           restore-keys: |
             ${{ runner.os }}-go-
       - name: setup-go
--- a/.github/workflows/windows.yaml
+++ b/.github/workflows/windows.yaml
@@ -7,12 +7,10 @@ jobs:
   test:
     env:
       DOWNLOAD_CACHE: 'd:\downloadcache'
+      GOPATH: 'd:\a\_temp\go\work'
+      GOCACHE: 'd:\a\_temp\go\cache'
     runs-on: windows-latest
     steps:
-      - shell: pwsh
-        run: |
-          echo "GOPATH=D:/a/_temp/go/work" >> $env:GITHUB_ENV
-          echo "GOCACHE=D:/a/_temp/go/cache" >> $env:GITHUB_ENV
       - name: checkout
         uses: actions/checkout@v4
       - name: go-cache


.github/workflows/windows.yaml
```diff
@@ -9,15 +9,17 @@ jobs:
...
       - name: checkout
         uses: actions/checkout@v4
       - name: go-cache
         uses: actions/cache@v3
         with:
-          path: |
-            ~\AppData\Local\go-build
-            ~\go\pkg\mod
-          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          path: D:/a/_temp/go
```
pkwarren, line 21: FYI - we can remove the separate actions/cache caching - that was related to https://github.com/actions/setup-go/issues/372 which is now fixed, so we should be able to use setup-go built-in caching instead of a separate step.

.github/workflows/windows.yaml
```diff
@@ -9,15 +9,17 @@ jobs:
       DOWNLOAD_CACHE: 'd:\downloadcache'
     runs-on: windows-latest
     steps:
+      - shell: pwsh
+        run: |
+          echo "GOPATH=D:/a/_temp/go/work" >> $env:GITHUB_ENV
+          echo "GOCACHE=D:/a/_temp/go/cache" >> $env:GITHUB_ENV
```
doriable, line 15: nit: Would it be possible to use the `env` config key for this step to define these env vars instead?

doriable (commented):
This seems reasonable to me, just added a nit :) Thanks for looking into this!

.github/workflows/windows.yaml
jchadwick-buf: The main reason I did this is, don't really like the way setup-go handles caching. I figure packing the stuff we want to cache into a single directory is a lot simpler, especially since we are special casing cache behavior just for Windows.

However, I am more interested in getting a good consensus here. What do you think out? The way I see it, we have basically three good options:
- Switch to setup-go@v4, try to get the automatic caching to work and not produce a pathological case.
- Switch to setup-go@v4 but set `cache: false` explicitly so that it's clear this is intentional.
- Keep as-is for now.

pkwarren: I'm ok with keeping it as-is - it does give us more control if there are additional tweaks that could be made for actions/cache (since they may take a while to make it into actions/setup-go or it may not expose the same levels of control over caching).

emcfarlane: It'd be nice to try get the setup-go@v4 working with caching just to prove the pathological case still exists.

jchadwick-buf: Fair enough, will do.

.github/workflows/windows.yaml
jchadwick-buf: Good call, I tunnel-visioned to hard here.

.github/workflows/windows.yaml
emcfarlane: It probably won't help: https://github.com/actions/runner-images/issues/7320 😬 .

jchadwick-buf: https://github.com/bufbuild/buf/actions/runs/6907562675/job/18795193819 If you want to follow along, here's my attempt at it. Verdict: Immediately, it is clear that we are on to something here: the setup-go@v4 cache is dramatically worse. Hell, even the go test times are worse.

emcfarlane: Shouldn't actions/cache be removed?

jchadwick-buf: D'oh, I just realized we already do use `setup-go@v4` with `cache: false`. I must've read `cache@v4` by accident.

So I can confirm the issue is _not_ fixed using just `setup-go@v4` with caching enabled. I added a comment for why I am leaving cache disabled for now. Since `setup-go@v4` uses the `GOCACHE` and `GOMODCACHE` directories, it should be possible to make this fix work with it too, but since I'm in a stop-the-bleeding mood, I'll defer that to later unless someone would prefer I blocked on this. (More than happy to if someone feels strongly.)

jchadwick-buf:
(Re-requesting your reviews to make sure I handled your comments satisfactory.)

.github/workflows/windows.yaml
```diff
@@ -7,23 +7,25 @@ jobs:
   test:
     env:
       DOWNLOAD_CACHE: 'd:\downloadcache'
+      GOPATH: 'd:\a\_temp\go\work'
```
pkwarren, line 10: I think what we're after is putting both the go build/test cache and mod cache in this directory. I think I'd rather we use `GOMODCACHE` instead of `GOPATH` here then (just in case anything else sneaks in under `GOPATH`).

pkwarren (commented):
Looks good - just one small suggestion.

.github/workflows/windows.yaml
jchadwick-buf: Alright, sounds good. Since this basically involves breaking the cache again, I'm going to go ahead and give a shot at making the setup-go@v4 path work; these are the two directories that setup-go@v4 caches, so it ought to be at least similar.

pkwarren (approved):

jchadwick-buf:
OK, now with `setup-go@v4` caching. Hopefully this will be a lot more pleasing to people.

Performance so far:

- [**Attempt #1**](https://github.com/bufbuild/buf/actions/runs/6907830096/job/18795801282): 4m 12s, uncached
- [**Attempt #2**](https://github.com/bufbuild/buf/actions/runs/6907830096/job/18795920401): 3m 9s, uncached (go1.21.4 was released between these two runs, making the first cache not hit)
- [**Attempt #3**](https://github.com/bufbuild/buf/actions/runs/6907830096/job/18796035967): 3m 8s, cached
- [**Attempt #4**](https://github.com/bufbuild/buf/actions/runs/6907830096/job/18796137543): 2m 43s, cached
- [**Attempt #5**](https://github.com/bufbuild/buf/actions/runs/6907830096/job/18796260115): 3m 12s, cached

OK, still looks good. Nice.

The one last thing would be to check to see if we *really* need the silly _temp dir. @pkwarren has been instrumental in finding underlying issues relating to this problem and found https://github.com/actions/setup-go/pull/393 which suggests that it's not that the temp directory is fast, it's that C: is slow. So in theory, we can move the cache dirs to just D: and it will be fast still.

pkwarren (approved):

jchadwick-buf:
And the verdict is in: no weird `_temp` directory needed.

For more information on what's going on, see [https://github.com/actions/runner-images/issues/8755](this issue). Although it's not said explicitly, I have a hunch at what the problem might be. C: drive obviously needs to contain the operating system and basically all of the runner image defaults, whereas D: is just a fresh, new drive. Internally this is probably implemented as creating a disk from a snapshot. And when creating a disk from a snapshot in a VM, it's pretty typical for the performance to be slow. In fact, even the Azure [documentation references this](https://learn.microsoft.com/en-us/azure/virtual-machines/scripts/create-managed-disk-from-snapshot#check-disk-status). It's not clear that this is exactly how GitHub Actions works internally, but it would suffice to say that the constraint that C: drive must contain all of the runner image contents probably makes it a lot slower due to needing copy-on-write behavior.

So now, and hopefully for the last time...
- [**Attempt #1**](https://github.com/bufbuild/buf/actions/runs/6908057125/job/18796534908): 5m 15s, uncached
- [**Attempt #2**](https://github.com/bufbuild/buf/actions/runs/6908057125/job/18796697517): 3m 43s, cached
- [**Attempt #3**](https://github.com/bufbuild/buf/actions/runs/6908057125/job/18796822676): 2m 51s, cached
- [**Attempt #4**](https://github.com/bufbuild/buf/actions/runs/6908057125/job/18796978724): 3m 40s, uncached (unknown reason; maybe we're near the cache limit)
- [**Attempt #5**](https://github.com/bufbuild/buf/actions/runs/6908057125/job/18797147371): 2m 55s, cached

Phew. This one was really tricky and had a lot of unexpected twists and turns. Thanks @pkwarren for the help figuring things out, and I hope everyone feels I addressed their comments sufficiently. I'll leave this open for a few more minutes to make sure nobody else has anything to say.

PR #2602, opened 2023-11-17
Return early if map type mismatch in buf lint protovalidate

oliversun9:
This PR fixes an issue where `buf lint` crashes when map rules are defined on a non-map field:

```
  int32 not_a_map = 14 [
    (buf.validate.field).map.values.int32.lt = 10,
    (buf.validate.field).map.values.int32.gt = 1
  ];
```

The fix is to return early if type mismatches.

### DIFF
--- a/private/bufpkg/bufcheck/buflint/buflint_test.go
+++ b/private/bufpkg/bufcheck/buflint/buflint_test.go
@@ -614,6 +614,7 @@ func TestRunProtovalidateRules(t *testing.T) {
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 44, 5, 44, 47, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 46, 5, 46, 47, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 50, 5, 50, 57, "PROTOVALIDATE"),
+		bufanalysistesting.NewFileAnnotation(t, "map.proto", 53, 5, 53, 50, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "message.proto", 20, 3, 20, 49, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "message.proto", 27, 5, 27, 51, "PROTOVALIDATE"),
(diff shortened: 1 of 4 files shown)


rodaine:
Are we returning a lint warning/error for this otherwise?

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -614,6 +614,7 @@ func TestRunProtovalidateRules(t *testing.T) {
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 44, 5, 44, 47, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 46, 5, 46, 47, "PROTOVALIDATE"),
 		bufanalysistesting.NewFileAnnotation(t, "map.proto", 50, 5, 50, 57, "PROTOVALIDATE"),
+		bufanalysistesting.NewFileAnnotation(t, "map.proto", 53, 5, 53, 50, "PROTOVALIDATE"),
```
bufdev, line 617: Should this produce two lint errors l, one for each of the two mismatched annotations?

oliversun9:
> Are we returning a lint warning/error for this otherwise?

Now we return a lint warning saying you have map rules for a non-map field, and we don't further check whether a the map rule's key and value sub rules.

private/bufpkg/bufcheck/buflint/buflint_test.go
oliversun9: The two lines belong to the same rule, `(buf.validate.field).map`. 

When we pass the path `(buf.validate.field).map` to `OptionExtensionLocation`, the location we get back is just the first line.

I think it's due to the way `sourceCodeLocation`s are generated by the compiler -- there isn't a location that spans both lines exactly.

oliversun9: We are also not able to tell whether the option is specified in two lines or in the aggregate form.

On a high level, the linter cannot tell

```
  int32 not_a_map = 14 [
    (buf.validate.field).map.values.int32.lt = 10,
    (buf.validate.field).map.values.int32.gt = 1
  ];
```

from

```
  int32 not_a_map = 14 [(buf.validate.field).map.values.int32 = {lt: 10, gt: 1}];
```

bufdev (approved):
Approved assuming @rodaine comment addressed

bufdev (approved):

PR #2618, opened 2023-11-21
Rename stdlib extension packages to use ext prefix instead of extended

bufdev:
This has been an extended annoyance of mine. I chose "extended" when I started because I was worried that "ext" could overlap with "external", when it was meant to signify "this is an extension to the similarly-named stdlib package". Instead, it just caused all of us to type a bunch more characters. Even if you do interpret it as "external", it kind of is - it's external to the stdlib package of similar name.

We did a vote between myself, @mfridman @emcfarlane @oliversun9 and "ext" was the winner.


oliversun9 (approved):

mfridman (approved):

emcfarlane (approved):

proto/buf/alpha/image/v1/image.proto
```diff
@@ -21,14 +21,14 @@ import "google/protobuf/descriptor.proto";
...
 
-// Image is an extended FileDescriptorSet.
+// Image is an ext FileDescriptorSet.
 //
 // See https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto
 message Image {
   repeated ImageFile file = 1;
 }
 
-// ImageFile is an extended FileDescriptorProto.
```
oliversun9, line 31: nit: proto files

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -392,7 +392,7 @@ message MapEntry {
 
 // FileExtension provides the information for the documentation of a file extension.
 message FileExtension {
-  // extension_type is the string representation of the type being extended.
```
oliversun9, line 395: Here it should be `extended` as well. I think these are the only places.

private/pkg/syncext/syncext.go
```diff
@@ -22,7 +22,7 @@ import "sync"
 //
 // If f panics, the returned function will panic with the same value on every call.
 //
-// This is copied from sync.OnceValues and extended to for three values.
+// This is copied from sync.OnceValues and ext to for three values.
```
oliversun9, line 25:
```suggestion
// This is copied from sync.OnceValues and extended to for three values.
```

private/pkg/storage/storagearchive/storagearchive.go
```diff
@@ -229,7 +229,7 @@ func Unzip(
 
 func isAppleExtendedAttributesFile(fileInfo fs.FileInfo) bool {
 	// On macOS, .tar archives created with libarchive will contain additional
-	// files with a prefix of "._" if there are files with extended attributes
+	// files with a prefix of "._" if there are files with ext attributes
```
oliversun9, line 232:
```suggestion
	// files with a prefix of "._" if there are files with extended attributes
```

proto/buf/alpha/image/v1/image.proto
bufdev: fixed

private/pkg/protosource/protosource.go
```diff
@@ -155,7 +155,7 @@ type OptionExtensionDescriptor interface {
 
 // Location defines source code info location information.
 //
-// May be extended in the future to include comments.
+// May be ext in the future to include comments.
```
oliversun9, line 158:
```suggestion
// May be extended in the future to include comments.
```

proto/buf/alpha/registry/v1alpha1/doc.proto
bufdev: fixed

private/pkg/protosource/field.go
```diff
@@ -25,7 +25,7 @@ type field struct {
 	label         descriptorpb.FieldDescriptorProto_Label
 	typ           descriptorpb.FieldDescriptorProto_Type
 	typeName      string
-	// if the field is an extension, this is the type being extended
+	// if the field is an extension, this is the type being ext
```
oliversun9, line 28:
```suggestion
	// if the field is an extension, this is the type being extended
```

private/pkg/licenseheader/cmd/license-header/main.go
```diff
@@ -109,7 +109,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		ignoreFlagShortName,
 		nil,
 		`File paths to ignore.
-These are extended regexes in the style of egrep.
+These are ext regexes in the style of egrep.
```
oliversun9, line 112:
```suggestion
These are extended regexes in the style of egrep.
```

PR #2624, opened 2023-11-24
Split create repo and sync module into different steps

unmultimedio:


saquibmian (approved):
I think the original idea was that we could avoid an RPC in the common case, but I agree this is saner. We make so many RPC calls in Sync anyway.

Sucks that we can't hoist this sooner, to make the other methods more hardened.

private/buf/bufsync/bufsyncapi/sync_handler.go
```diff
@@ -54,6 +54,7 @@ type syncHandler struct {
 
 	moduleIdentityToRepositoryIDCache  map[string]string
 	moduleIdentityToDefaultBranchCache map[string]string
+	existingRepositoriesCache          map[string]struct{}
```
bufdev, line 57: What is this a map of? It's not a map of `Repository`, as that type doesn't exist.

The other maps in this package used `moduleIdentity` as a part of their name, which would lead me to conclude this should be `existingModuleIdentityCache` is there a reason you decided to deviate from what everything else here does?

private/buf/bufsync/bufsyncapi/sync_handler.go
```diff
@@ -75,6 +76,7 @@ func newSyncHandler(
 		createWithVisibility:                 createWithVisibility,
 		moduleIdentityToRepositoryIDCache:    make(map[string]string),
 		moduleIdentityToDefaultBranchCache:   make(map[string]string),
+		existingRepositoriesCache:            make(map[string]struct{}),
```
bufdev, line 79: A general note that none of this is thread-safe, which may or may not be a problem, but we've generally strived to make map mutations thread-safe (although don't everywhere).

private/buf/bufsync/bufsyncapi/sync_handler.go
unmultimedio: You're right, the key is the module identity, I used `repo` wording is because conceptually what we're doing is invoking the `CreateRepository` RPC. I'll rename it 👍🏼

private/buf/bufsync/bufsyncapi/sync_handler.go
unmultimedio: Yeah this doesn't need to be thread safe becase we're not doing sync concurrently in the same CLI job. We loop over modules -> branches -> commits one by one, so this won't ever get called by more than one goroutine at once. Same as the other caches in here.

PR #2626, opened 2023-11-27
Add support for C language plugins

pkwarren:


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -56,6 +56,7 @@ enum PluginLanguage {
   PLUGIN_LANGUAGE_PHP = 13;
   PLUGIN_LANGUAGE_CSHARP = 14;
   PLUGIN_LANGUAGE_SCALA = 15;
+  PLUGIN_LANGUAGE_C = 16;
```
pkwarren, line 59: Don't need this yet but we may in a bit (trying to get ahead on CLI side).

mfridman (approved):

bufdev (approved):

PR #2631, opened 2023-11-30
On login check for an existing valid token before prompting

emcfarlane:
With this change, when users attempt to log in, the system will first access the machine's stored credentials and validate them, if available. This validation process is aimed at preventing user confusion regarding the validity of their existing token.

This now allows users to verify if they have a valid token before prompting for a new one. To login with a new token simply remove the old one by calling `buf registry logout` prior to initiating a new login.

Login with a specified repository and username: 
```bash
$ buf registry login bufbuild.internal --username 'localbuf'
Valid credentials found in /Users/edward/.netrc.
```

Login without specifics:
```bash
$ buf registry login
Log in with your Buf Schema Registry username. If you don't have a username, create one at https://buf.build/.

Username: emcfarlane
Valid credentials found in /Users/edward/.netrc.
```

Aims to streamline the login experience and reduce ambiguity of token validity for users accessing the BSR.

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -157,6 +157,10 @@ func inner(
 			return err
 		}
 	}
+	netrcFilePath, err := netrc.GetFilePath(container)
+	if err != nil {
+		return err
+	}
 	var token string
 	if flags.TokenStdin {
 		data, err := io.ReadAll(container.Stdin())
@@ -165,6 +169,17 @@ func inner(
 		}
 		token = string(data)
 	} else {
+		// Check if we already have a valid token for this remote and username.
+		if machine, err := netrc.GetMachineForName(container, remote); machine != nil && err == nil {
+			currentToken := machine.Password()
+			if err := isValidToken(ctx, container, remote, username, currentToken); err == nil {
+				alreadyLoggedInMessage := "Valid credentials found in %s.\n"
+				if _, err := fmt.Fprintf(container.Stdout(), alreadyLoggedInMessage, netrcFilePath); err != nil {
+					return err
+				}
+				return nil
+			}
+		}
 		var err error
 		token, err = bufcli.PromptUserForPassword(container, "Token: ")
 		if err != nil {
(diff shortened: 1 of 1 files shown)


bufdev:
Let's talk at the desks tomorrow, I am worried this might actually confuse me a bit. Perhaps there's a bit of a different flow here, or a new command we need to add and reference on authentication error.

> To login with a new token simply remove the old one by calling buf registry logout prior to initiating a new login.

This wouldn't be obvious to me as something to do :-) There's also a very valid usecase to wanting to login even if there's existing credentials - I want to replace my token because either (1) I'm going to imminently delete the old one or (2) I want to get a token with a longer expiration date.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -165,6 +169,17 @@ func inner(
 		}
 		token = string(data)
 	} else {
+		// Check if we already have a valid token for this remote and username.
+		if machine, err := netrc.GetMachineForName(container, remote); machine != nil && err == nil {
```
bufdev, line 173: Note that this is not actually getting a token for the given remote/username pair, by definition the API is just getting machine based on remote. Then, in `isValidToken`, I error if the username is different. It's common to want to login with a different user, and the general pattern I'd expect to do that is to just call `buf registry login` again. With this PR, I'd now get an error with that (and that error wouldn't tell me to first logout, although that's fixable).

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -165,6 +169,17 @@ func inner(
...
+			currentToken := machine.Password()
+			if err := isValidToken(ctx, container, remote, username, currentToken); err == nil {
+				alreadyLoggedInMessage := "Valid credentials found in %s.\n"
```
bufdev, line 176: Assuming we want to keep this PR largely as-is, I would want this error message to tell me something like `Valid credentials already exist at %s, if you'd like to change them, please run `buf registry logout`.

Come to think of it though, by forcing users to call `logout`, what you're forcing them to do is delete *all of their BSR credentials, including for any Pro/Enterprise instances. So now, instead of being able to overwrite my token for a given instance with `buf registry login`, I get an error, and I have no way to individually overwrite a token for a single BSR instance as my only option is to delete all tokens.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
emcfarlane: On a different username the error would be raised as not matching and continue the normal flow of prompting the user for their token. No error would be raised. It's behaviour is meant to be analogous to a website style "already logged in" no need to re-enter credentials.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
emcfarlane: Oh, logout deleting all is probably not desired. Will close, can reopen when there's a better case for it.

PR #2634, opened 2023-12-01
Add `buf alpha sdk python-version`

stefanvanburen:
Effectively the same implementation as the others, with just the flag name / examples changed.


bufdev (approved):
Can you do a merge PR against `bufmod` and put that up as well? Or if you don't imminently need this command, instead do it against `bufmod`? `bufmod` will be merged in <1 month and is the future

PR #2636, opened 2023-12-01
Update `bufmod` from `main`

stefanvanburen:
Basically:

```console
$ git checkout bufmod
$ git switch -c svanburen/bufmod-update
$ git merge main
```

Didn't cleanly apply, so had to make some edits in the last two commits:

* The [error message commit didn't apply](https://github.com/bufbuild/buf/commit/d3dc2fa23ee4aa459a5f359e5f1e89bad31be5b4), so added the [error message to the new location](https://github.com/bufbuild/buf/commit/f3d9fce31ea7f198608d4fb50830874531ff553f).
* We made a few updates to the packages used in `buf alpha sdk python-version`, so I made edits based on the other commands in https://github.com/bufbuild/buf/commit/a33642729a246b9f27c9b38c44092ca6a7621294.

`make testbufnew` passes.


private/buf/cmd/buf/buf.go
```diff
@@ -287,7 +289,8 @@ func wrapError(err error) error {
 				"visit https://docs.buf.build/bsr/authentication`, authErr.TokenEnvKey())
 			}
 			return errors.New(`Failure: you are not authenticated. Create a new entry in your netrc, " +
-			"using a Buf API Key as the password. For details, visit https://docs.buf.build/bsr/authentication`)
+				"using a Buf API Key as the password. If you already have an entry in your netrc, check " +
+				"to see that your token is not expired. For details, visit https://docs.buf.build/bsr/authentication`)
```
stefanvanburen, line 293: maybe dumb question, but given that we're using raw strings here, wouldn't they include the raw `"` and `+` characters here? (Going to check locally.)

stefanvanburen: ya, let me fix that, heh:

```console
$ buf beta registry repository get buf.build/svanburen/test-private
Failure: you are not authenticated. Create a new entry in your netrc, " +
				"using a Buf API Key as the password. If you already have an entry in your netrc, check " +
				"to see that your token is not expired. For details, visit https://docs.buf.build/bsr/authentication
```

stefanvanburen: Fixed in https://github.com/bufbuild/buf/pull/2636/commits/cdad00cc70d50f750d61881213d174a8ad2c80a2.

PR #2644, opened 2023-12-04
Update bufmod proto/generate setup

jchadwick-buf:
- Fix `wkt-go-data` for new bufmod API
    - Can now do partial `make generate` using `make generate BUF_BIN=$(which buf)`.
- Includes fixes for existing issues that were not caught due to non-working `make generate`.
- Move `proto` -> `proto/buf` in preparation for incoming `proto/buftest` module.


jchadwick-buf:
Going to merge these 3 PRs into just one, this is too cumbersome I think.

jchadwick-buf:
This can't work anymore since buf.yaml is now v2 in bufmod branch. Closing.

PR #2646, opened 2023-12-04
bufmod: Move proto -> proto/buf

jchadwick-buf:
Moves proto -> proto/buf in preparation for proto/buftest.

Includes an octomerge of `jchadwick/bufmod-partial-generate` and `jchadwick/bufmod-fix-wktdata`, so I set the base to `jchadwick/bufmod-partial-generate` to make the diff as free of noise as possible.


jchadwick-buf:
Going to merge these 3 PRs into just one, this is unnecessarily cumbersome I think.

PR #2650, opened 2023-12-07
Fix formatter issue with message literals with any syntax

jhump:
This was recently pointed out to me by @jacobbuf. The formatter simply didn't handle the special Any syntax in message literals.


bufdev (approved):

jhump:
@bufdev, I just added a note to the changelog about this. PTAL

bufdev (approved):

PR #2661, opened 2023-12-13
Parse lint and breaking ignore paths in buf.yaml v2

oliversun9:
In buf.yaml v2, we want paths for `ignore` and `ignore_only` in `lint` and `breaking` to be relative to the workspace root, but paths in a `CheckConfig` should be relative to module roots, i.e. they are import paths. This PR adds this transformation for read and write.

### DIFF
--- a/private/bufpkg/bufconfig/buf_yaml_file.go
+++ b/private/bufpkg/bufconfig/buf_yaml_file.go
@@ -538,8 +538,9 @@ func writeBufYAMLFile(writer io.Writer, bufYAMLFile BufYAMLFile) error {
 			},
 		)
 		for _, moduleConfig := range bufYAMLFile.ModuleConfigs() {
+			moduleDirPath := moduleConfig.DirPath()
 			externalModule := externalBufYAMLFileModuleV2{
-				Directory: moduleConfig.DirPath(),
+				Directory: moduleDirPath,
 			}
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -306,20 +306,30 @@ func readBufYAMLFile(reader io.Reader, allowJSON bool) (BufYAMLFile, error) {
...
+		// TODO: we do no validation of paths now
+		lintConfig, err := getLintConfigForExternalLint(
+			fileVersion,
+			externalBufYAMLFile.Lint,
+			normalpath.NormalizeAndValidate,
+		)
+		if err != nil {
+			return nil, err
+		}
+		// TODO: we do no validation of paths now
```
oliversun9, line 318: This TODO is left as-is: in `bufcheck/internal`, `newConfigForRuleBuilders` does validation for these paths, which is also the behavior on main. Note that `newConfigForRuleBuilders` is called only for `buf lint` or `buf breaking`.

One interesting implication is that a user can have a semi-invalid buf.yaml (a file with the correct shape but invalid ignore paths) and they can successfully run `buf build` but not `buf lint`. I verified that this is the case with main. (The easiest way to try this out is to add `"."` to `ignore` and run `buf build` and `buf lint`)

bufdev (approved):

PR #2669, opened 2023-12-15
Add buf migrate command

oliversun9:
This PR makes update to the migrate command:
* moves it from `buf beta migrate` to `buf migrate`
* Update flags to take `--module` and `--workspace` instead of paths to those files.
* The destination for the buf.yaml and buf.lock is `.`, unless `--workspace` is specified but `--module` is not specified, in which case the destination is the directory for that `buf.yaml`.
  - If the user runs `buf migrate --workspace foo/bar`, then the new `buf.yaml` will be written at `foo/bar/buf.yaml`.
  - Otherwise, `./buf.yaml` will be written.
* Only allows one workspace to be specified. (This isn't new in this PR, but want to highlight this).
* Add flag `--template`, allowing multiple paths to buf.gen.yamls.
* Fixes the bug where the command crashes when only one directory without a buf.yaml is specified.
* Update lint/breaking rule translation to produce equivalent lists of rules with minimal diff.

Note: with the recent change on [reading a lock file without digest](https://github.com/bufbuild/buf/pull/2681). Right now `buf migrate` fails with `Failure: unimplemented` if there is a buf.lock to write. It shouldn't fail once commit service has been implemented.


private/buf/bufmigrate/migrator.go
```diff
@@ -15,374 +15,302 @@
...
+		m.warnf("%s is a v2 file, no migration required", bufGenYAMLPath)
+		return nil
 	}
+	if typeConfig := bufGenYAML.GenerateConfig().GenerateTypeConfig(); typeConfig != nil && len(typeConfig.IncludeTypes()) > 0 {
+		// TODO: what does this sentence mean? Get someone else to read it and understand it without any explanation.
+		m.warnf(
+			"%s is a v1 generation template with a top-level 'types' section including %s. In a v2 generation template, 'types' can"+
+				" only exist within an input in the 'inputs' section. Since the migration command does not have information"+
+				" on inputs, the migrated generation will not have an 'inputs' section. To add these types in the migrated file, you can"+
+				" first add an input to 'inputs' and then add these types to the input.",
```
oliversun9, line 112: Do you think this make sense? @doriable

bufdev (approved):

PR #2677, opened 2023-12-19
Add flag presence for include imports and include wkt in buf generate

oliversun9:
Update flags `include_imports` and `include_wkt` in `buf generate` so that there is a difference between not setting the flag at all (`buf generate`) and setting the flag explicitly to false (`buf genenerate --inlcude-imports`). The latter overrides any `include_imports` in buf.gen.yaml v2.

Some code related to `buf generate --migrate` is also deleted in this PR.


private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -524,67 +480,42 @@ func getInputImages(
...
+		usage,
 	)
-	if err != nil {
-		return nil, err
+	flag.NoOptDefVal = "true"
+}
+
+// Implements pflag.Value.
+type boolPointerValue struct {
+	// This must not be nil at construction time.
```
oliversun9, line 499: This exists because I couldn't find an out-of-the-box function in `pflags` library (something like `BindBoolVarPointer`).

private/buf/bufgen/bufgen.go
```diff
@@ -121,22 +121,24 @@ func GenerateWithBaseOutDirPath(baseOutDirPath string) GenerateOption {
 	}
 }
 
-// GenerateWithIncludeImports says to also generate imports.
+// GenerateWithIncludeImportsOverride says to always generate imports (or never
```
bufdev, line 124: Add a comment explicitly calling out that the given bool has presence, i.e. it's a pointer and this is a strict override.

bufdev (approved):

PR #2680, opened 2023-12-19
Delete git code from Sync

unmultimedio:
- Delete the whole `storagegit` pkg
- Delete most of the `pkg/git` pkg, since some of it previously existed, and it's used to clone  and checkout repos in `buffetch`


private/pkg/git/git.go
```diff
@@ -173,276 +135,3 @@ type ListFilesAndUnstagedFilesOptions struct {
 	// is being applied to.
 	IgnorePathRegexps []*regexp.Regexp
 }
-
-// Hash represents the hash of a Git object (tree, blob, or commit).
-type Hash interface {
```
bufdev, line 178: There's one place in core that uses `Hash`, we should see if we can remove it?

unmultimedio: You're right, here a PR that removes it: https://github.com/bufbuild/core/pull/13382

bufdev (approved):

PR #2681, opened 2023-12-20
Update callers of GetBufLockFileForPrefix to pass a resolve function

oliversun9:
This is a followup to https://github.com/bufbuild/buf/commit/39f05103f35911b76c5b5eabf98fd2ce9114c797, passing the `BufLockFileWithDigestResolver` option whenever possible.
* Add helper to `bufmoduleapi`:
```
func CommitIDToDigest(
	ctx context.Context,
	clientProvider bufapi.ClientProvider,
	remote string,
	commitID string,
) (bufcas.Digest, error)
```
* Pass a client provider to callers of `GetBufLockFileForPrefix`. These callers can then pass
```
func(ctx context.Context, remote string, commitID string) (bufcas.Digest, error) {
	return bufmoduleapi.CommitIDToDigest(ctx, p.clientProvider, remote, commitID)
},
```
as the resolve function.

In the future, to make fewer network calls, we could have a `CommitResolver` and create it with
```
NewCommitResolver(..., clientProvider, hints)
```
where `hints` are potential commit IDs to resolve.


private/buf/bufworkspace/workspace_test.go
```diff
@@ -65,6 +65,7 @@ func testBasic(t *testing.T, subDirPath string) {
 		zap.NewNop(),
 		tracing.NopTracer,
 		bucket,
+		nil,
```
oliversun9, line 68: It feels awkward to just pass a `nil`, should I create some `NewNopClientProvider` helper?

bufdev: yea

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -69,20 +72,23 @@ func ModuleDataStoreWithTar() ModuleDataStoreOption {
...
+	logger         *zap.Logger
+	clientProvider bufapi.ClientProvider
+	bucket         storage.ReadWriteBucket
 
 	tar bool
 }
 
 func newModuleDataStore(
 	logger *zap.Logger,
+	clientProvider bufapi.ClientProvider,
```
bufdev, line 84: I would actually not pass it to the moduleDataStore, and instead add a comment that we expect all lock files in the moduleDataStore to have digests. We control this.

bufdev (commented):
Mostly looks good, this was less ugly than I thought, see two comments (add a no-op ClientProvider, do not pass to ModuleDataStore)

private/bufpkg/bufapi/bufapi.go
```diff
@@ -37,6 +37,13 @@ func NewClientProvider(clientConfig *connectclient.Config) ClientProvider {
 	return newClientProvider(clientConfig)
 }
 
+// NewNopClientProvider returns a new ClientProvider that provides unimplemented services.
+//
+// Should be used for testing only.
+func NewNopClientProvider() ClientProvider {
+	return newNopClientProvider()
```
bufdev, line 44: Nit: In other places, we just do `var NopClientProvider ClientProvider = nopClientProvider{}` as a global variable, and then you don't have any constructors, and you implement the functions directly on the struct, not the struct pointer, ie `func (nopClientProvider) BranchServiceClient...`

bufdev (commented):
I would add a comment in ModuleDataStore about why we are not passing the option as well.

bufdev (approved):

PR #2682, opened 2023-12-20
Add `v2` modules to integration tests

doriable:
This adds `v2` modules to all `buf/cmd/buf` workspace and lint integration tests.
It also covers a couple of a bit of clean-up around error messages.


bufdev (commented):
I would love some tests added to probably `buflint_test.go` that test the new ignore_only paths being relative to the buf.yaml instead of being root-relative, and some tests that test excludes per directory in these integration tests, does that make sense/could you add those?

bufdev:
Also, are there any tests checking how `v2` integrates with `--config`? I already can't remember the design decision I made regarding  `--config` with inline `v2` values, but would love us to lock in that behavior.

doriable:
> Also, are there any tests checking how `v2` integrates with `--config`? I already can't remember the design decision I made regarding `--config` with inline `v2` values, but would love us to lock in that behavior.

So there currently is a test that inlines `v2` config with `--config`. It seems that the expectation right now is that it returns a non-zero status code (1) but does not provide a helpful error string.

https://github.com/bufbuild/buf/blob/4b7b26d54eab16ead054b42276fd4c24c4a128ee/private/buf/cmd/buf/workspace_test.go#L189-L207

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -1109,30 +1168,40 @@ func testLintWithOptions(
 		storageos.ReadWriteBucketWithSymlinksIfSupported(),
 	)
 	require.NoError(t, err)
+	var options []bufworkspace.WorkspaceBucketOption
```
bufdev, line 1171: This isnt used

private/bufpkg/bufcheck/buflint/buflint_test.go
```diff
@@ -1109,30 +1168,40 @@ func testLintWithOptions(
...
 		zap.NewNop(),
 		tracing.NopTracer,
 		readWriteBucket,
 		bufapi.NopClientProvider,
 		bufmodule.NopModuleDataProvider,
+		options...,
 	)
 	require.NoError(t, err)
 
+	if moduleFullNameString == "" {
```
bufdev, line 1183: change to:

```
opaqueID := moduleFullNameString
if opaqueID == "" {
  opaqueID = "."
}
```

private/bufpkg/bufmodule/module_set.go
```diff
@@ -190,6 +190,40 @@ func ModuleSetToDAG(moduleSet ModuleSet) (*dag.Graph[string], error) {
 	return graph, nil
 }
 
+// ModuleToSelfContainedModuleReadBucketWithOnlyProtoFiles converts the Module to a
```
bufdev, line 193: Why did this move?

private/bufpkg/bufcheck/buflint/buflint_test.go
doriable: Ah yes, I ended up going with targeting the opaque IDs instead

private/bufpkg/bufmodule/module_set.go
doriable: Oh, I think I moved this by accident while traversing some of the helpers. Moved it back.

bufdev (approved):

PR #2696, opened 2024-01-02
Re-introduce managed mode enabled warning

doriable:
We had a warning message where if the user has
managed mode options set but `enabled: false`: #1552.
This warning was originally implemented when reading
the config, which required the config reader to have access
to a logger.

Instead of passing around a logger to the config reader, we can
instead print this message when the `Generate` is called, since
the `generator` has access to a logger.


doriable:
I explored another option to check this in `bufimagemodify` instead,
but this would also require passing the logger down to a library that
should not otherwise log messages.

bufdev (approved):

PR #2704, opened 2024-01-09
Fix some path logic to work on Windows

jchadwick-buf:
This does not fix everything, so Windows CI is still not passing. However, this should still be closer.

I think ExternalPath is supposed to be unnormalized, so the tests that test for it are fixed to use unnormalized paths.


private/buf/buffetch/internal/reader.go
```diff
@@ -572,8 +572,9 @@ func getReadWriteBucketForOS(
 	if err != nil {
 		return nil, err
 	}
+	inputDirPathComponents := normalpath.Components(absInputDirPath)
 	osRootBucket, err := storageosProvider.NewReadWriteBucket(
-		string(os.PathSeparator),
+		inputDirPathComponents[0],
```
bufdev, line 577: This isn't right/isn't what the intended behavior is for unix, can comment later

bufdev (changes requested):
Blocking as this isn't what the fix should be unless I'm missing something

private/buf/buffetch/internal/reader.go
jchadwick-buf: For sake of posterity this behavior is identical on UNIX to the previous behavior because Components is called on the absolute path and looks like this:
```go
[]string{"/", "Users", "john", "Code", "buf", "private", "buf", "cmd", "buf", "testdata", "fail2"}
```
So the bucket on UNIX is always "/" but on Windows could be a volume name (drive letter, UNC path.)
I'm not really opposed to refactoring it anyways, just worth noting.

doriable: So I think using `normalpath.Components` in this way is somewhat reasonable, since it's parsing an absolute path, and so on UNIX, the behaviour is very safe. However, as you noted, the Windows path could be either fully-qualified or UNC, and I think this creates some instability. We do handle the first component in `normalpath` to set the volume:

https://github.com/bufbuild/buf/blob/21f4b652b3dfd500058f314c68c13cafd57d0735/private/pkg/normalpath/normalpath_windows.go#L129-L146

I think the concern is actually when we're given a UNC path. So I think the better route to go is to explicitly parse the drive and pass that through as our path, instead of deriving it from the arbitrary path structure.

doriable:
I put this in draft because I have no resolved all the issues yet, but would like to push up some changes. Will un-draft and request reviews once these issues are resolved.

private/pkg/filepathext/filepathext_windows.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"os"
+	"path/filepath"
+)
+
+var (
+	// USERPROFILE shows the current user's profile folder. It is being used to parse the volume
+	// for the current user and this is used as the process's context.
+	// https://learn.microsoft.com/en-us/windows/deployment/usmt/usmt-recognized-environment-variables#variables-that-are-processed-for-the-operating-system-and-in-the-context-of-each-user
+	FSRoot = filepath.VolumeName(os.Getenv("USERPROFILE")) + string(os.PathSeparator)
```
doriable, line 29: I retract my comment earlier, I think it isn't feasible to parse this through the system env vars -- the process could just be running in an entirely separate drive, which is what is happening in the actions. We should use `normalpath.Components` when parsing the absolute paths and document clearly the behaviours of how `reader.go` is constructing the OS root buckets.

Pearlminimic99 (approved):

private/bufpkg/bufconfig/file.go
```diff
@@ -205,13 +206,15 @@ func newDecodeError(fileName string, err error) error {
 		fileName = "config file"
 	}
 	// We intercept PathErrors in buffetch to deal with fixing of paths.
-	return &fs.PathError{Op: "decode", Path: fileName, Err: err}
+	// We return a cleaned, unnormalized path in the error for clarity with user's filesystem.
+	return &fs.PathError{Op: "decode", Path: filepath.Clean(normalpath.Unnormalize(fileName)), Err: err}
```
bufdev, line 210: `fileName` isn't necessarily a path I believe

doriable: In the case where it isn't a path (e.g. `fileName = "config file"` or some other case), `normalpath.Unnormalize` and `filepath.Clean` would be no-ops.

bufdev: We need to make a decision as to whether paths should be unnormalized in `fs.PathErrors` - I've generally taken the position that they should not be in the rest of the codebase, I'm worried that we're special-casing here, any thoughts?

doriable: I'm fine with either -- the unnormalized error was mostly for UX reasons (matching file paths, and we did show unnormalised file paths in the past, etc.) but as noted, they're not always a file path, and consistency is kind of the key. So I'm totally okay with keeping them as normalised paths if this is the pattern throughout the new libraries, in which case, I can adjust the tests to reflect that also.

doriable:
Turns out a large portion of our issues were due to a `.gitattributes` change that did not allow `git` to convert line endings on checkout for `testdata`. In the past, we set `**/testdata**` files to `binary`, which forces `git` to treat them as binary.
The change set `-crlf` instead, but this assumes that `git` does not check them out as `text` files.
So now, we are setting `-text` explicitly so that they are not being handled by `text=auto` setting, even if they are being checked out as text files. This fixes many of the issues caused by carriage returns.

There are still a couple of remaining TODO's to investigate:
- duplicated path separators
- archive test concurrency issues

PR #2708, opened 2024-01-17
Create a script to collect all bufmod TODOs

doriable:
This is a script for collecting all `TODO`s on `bufmod` that are
not on `main`. Sample output:

```
private/bufpkg/bufmodule/remote_dep.go: 44 // TODO: This needs a LOT of testing.
private/pkg/connectclient/connectclient.go: 21 // TODO: eliminate config, just make the below MakeOptions, add httpClient as a parameter to Make
private/pkg/dag/graph.go: 45 // TODO: It really stinks that we have to use the constructor. We have what amounts
private/pkg/slicesext/slicesext.go: 146 // TODO: Delete this in favor of slices.Clone.
private/pkg/slicesext/slicesext.go: 166 // TODO: Make ToStructMap use this logic, remove ToStructMapOmitEmpty, to match other functions.
private/pkg/slicesext/slicesext.go: 315 // TODO: Replace with slices.Sort when we only support Go versions >= 1.21.
private/pkg/syncext/syncext.go: 25 // TODO: This is directly copied from 1.21 source, remove when no longer need <1.20.
private/pkg/syncext/syncext.go: 57 // TODO: This is directly copied from 1.21 source, remove when no longer need <1.20.
```

### DIFF
--- a/make/buf/scripts/newtodos.bash
+++ b/make/buf/scripts/newtodos.bash
@@ -18,13 +18,12 @@ TODOS="$(git diff "${BASE_BRANCH}" | awk '
 
 FILENAME=""
 while read -r line; do
-  if [[ "${line}" =~ [a-z,0-9]+.go ]]; then
+  if [[ "${line}" =~ [a-z,A-Z,0-9]+\.[a-z]{2,3}$ ]]; then
     FILENAME="${line}"
-  elif [[ "${line}" =~ PACKAGES.md ]]; then
-    # ignore the contents of PACKAGES.md
+  elif [[ "${line}" == *"make/buf/scripts/newtodos.bash"* ]]; then
     FILENAME="no print"
   else
-    if [[ "${FILENAME}" != "no print" ]]; then
+    if [[ "${FILENAME}" != "no print" ]]; then 
       LINENUMBER="${line%%:*}"
       TODO="${line#*:}"
       echo "${FILENAME}":"${LINENUMBER}":"${TODO#"${TODO%%[![:space:]]*}"}"
--- a/make/buf/scripts/newtodos.bash
+++ b/make/buf/scripts/newtodos.bash
@@ -27,7 +27,7 @@ while read -r line; do
     if [[ "${FILENAME}" != "no print" ]]; then
       LINENUMBER="${line%%:*}"
       TODO="${line#*:}"
-      echo "${FILENAME}": "${LINENUMBER}" "${TODO#"${TODO%%[![:space:]]*}"}"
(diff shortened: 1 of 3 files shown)


doriable:
@bufdev I wasn't sure if we wanted to put this as a part of `make generate`, I didn't want to inadvertently cause CI issues by doing that. It seems reasonable to generate and/use on an ad-hoc basis for now but yeah, let me know what you think.

make/buf/scripts/newtodos.bash
```diff
@@ -0,0 +1,33 @@
...
+  f=="?" {if (/^\+\+\+ /) f=substr($0, 7)"\n"; next}
+  /^@@/ {n=$3; sub(/,.*/,"",n); n=0+$3; next}
+  /^\+.*TODO/ {print f n ":" substr($0,2); f=""}
+  substr($0,1,1)~/[ +]/ {n++}')"
+
+FILENAME=""
+while read -r line; do
+  if [[ "${line}" =~ [a-z,0-9]+.go ]]; then
+    FILENAME="${line}"
+  elif [[ "${line}" =~ PACKAGES.md ]]; then
```
bufdev, line 23: There's no reason to ignore `PACKAGES.md`, if someone wants to ignore this, you can grep it out after the fact

make/buf/scripts/newtodos.bash
```diff
@@ -0,0 +1,33 @@
...
+TODOS="$(git diff "${BASE_BRANCH}" | awk '
+  /^diff / {f="?"; next}
+  f=="?" {if (/^\+\+\+ /) f=substr($0, 7)"\n"; next}
+  /^@@/ {n=$3; sub(/,.*/,"",n); n=0+$3; next}
+  /^\+.*TODO/ {print f n ":" substr($0,2); f=""}
+  substr($0,1,1)~/[ +]/ {n++}')"
+
+FILENAME=""
+while read -r line; do
+  if [[ "${line}" =~ [a-z,0-9]+.go ]]; then
```
bufdev, line 21: Optimally, we want all TODOS, not just Golang

bufdev:
This script still has bugs, example:

```
586:					// TODO: sweep and check paths gone: 593 // TODO: add default values
586:					// TODO: sweep and check paths gone: 773 // TODO in v2
```

make/buf/scripts/newtodos.bash
doriable: Fixed this up, I did exclude this script file though.

bufdev (approved):

PR #2711, opened 2024-01-17
Fix panic in `buf curl` that can happen with some map fields

jhump:
When `-v` is used, the command shows the amount of unrecognized data in the response. When it counts that data, a bug in one of the `case` expressions allowed it to fall into the wrong case (for a map whose values are _not_ messages). This would then result in a panic:
```
panic: type mismatch: cannot convert map to message
```

The panic actually happens whether or not the user includes a `-v` option, since this part of the code calculates and prints the value unconditionally. (If `-v` is _not_ specified, the "print" step is effectively a no-op.)

This branch fixes that panic and adds a test. The test uses a message with fields of every possible type (including groups) to make sure that all types are correctly handled without panic. The test reproduces the panic issue, and confirms the fix.


mattrobenolt:
🙏

bufdev (approved):

PR #2728, opened 2024-01-24
Fix InvalidRemoteErr panic

### LINKED ISSUE #2725: mod update panics when dependencies are from multiple registries
## Expected Behavior
Given the following `buf.yaml` file:
```yaml
version: v1
deps:
  - buf.build/bufbuild/protovalidate
  - private.registry/fetch/dependency # removed actual private dependency
breaking:
  use:
    - FILE
lint:
  use:
    - DEFAULT
```
When running `buf mod update`, I would expect my dependencies to be properly versioned in my `buf.lock` file.  If it's intentional to not support multiple registries, then I would assume an error stating that would be provided.

## Current Behavior
The cli panics with a nil pointer dereference.

```bash
pop-os% buf mod update
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x11458b4]

goroutine 1 [running]:
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End.func1()
        go.opentelemetry.io/otel/sdk@v1.20.0/trace/span.go:388 +0x25
go.opentelemetry.io/otel/sdk/trace.(*recordingSpan).End(0xc00045c600, {0x0, 0x0, 0x0?})
        go.opentelemetry.io/otel/sdk@v1.20.0/trace/span.go:426 +0xa3b
panic({0x12b58e0?, 0x2688bf0?})
        runtime/panic.go:920 +0x270
github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate.getDependencies({0x1800f10, 0xc00047a070}, 0x1?, {0x181f2e8, 0xc0000d9800}, 0xc000123db8, 0xc0000d9d00, {0x180ac40?, 0xc00003f0b0}, {0x160365c, ...})
        github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate/modupdate.go:272 +0x5d4
github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate.run({0x1800f10, 0xc00047a070}, {0x181f2e8?, 0xc0000d9800}, 0x0?)
        github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate/modupdate.go:119 +0x272
github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate.NewCommand.func1({0x1800f10?, 0xc00047a070?}, {0x181f2e8?, 0xc0000d9800?})
        github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate/modupdate.go:61 +0x2b
github.com/bufbuild/buf/private/buf/cmd/buf/command/mod/modupdate.NewCommand.NewErrorInterceptor.func3.1({0x1800f10?, 0xc00047a070?}, {0x181f2e8?, 0xc0000d9800?})
        github.com/bufbuild/buf/private/buf/bufcli/errors.go:91 +0x2b
github.com/bufbuild/buf/private/pkg/app/appflag.(*builder).run(0xc000466680, {0x1800ea0, 0xc000370000}, {0x180a870, 0xc000370190}, 0xc00005a680)
        github.com/bufbuild/buf/private/pkg/app/appflag/builder.go:138 +0x525
github.com/bufbuild/buf/private/pkg/app/appflag.(*builder).NewRunFunc.func1({0x1800ea0, 0xc000370000}, {0x180a870, 0xc000370190})
        github.com/bufbuild/buf/private/pkg/app/appflag/builder.go:95 +0x7c
github.com/bufbuild/buf/private/pkg/app/appcmd.commandToCobra.func2(0xc000401100?, {0x26ee1c0?, 0x0, 0x15fc4ba?})
        github.com/bufbuild/buf/private/pkg/app/appcmd/appcmd.go:303 +0x222
github.com/spf13/cobra.(*Command).execute(0xc0000fcf00, {0x26ee1c0, 0x0, 0x0})
        github.com/spf13/cobra@v1.8.0/command.go:987 +0xaa3
github.com/spf13/cobra.(*Command).ExecuteC(0xc000004300)
        github.com/spf13/cobra@v1.8.0/command.go:1115 +0x3ff
github.com/spf13/cobra.(*Command).Execute(...)
        github.com/spf13/cobra@v1.8.0/command.go:1039
github.com/bufbuild/buf/private/pkg/app/appcmd.run({0x1800ea0, 0xc000370000}, {0x180a870?, 0xc0000ce050?}, 0xc000695080)
        github.com/bufbuild/buf/private/pkg/app/appcmd/appcmd.go:251 +0x8fc
github.com/bufbuild/buf/private/buf/cmd/buf.Main.Main.newRunFunc.func1({0x1800ea0?, 0xc000370000?}, {0x180a870?, 0xc0000ce050?})
        github.com/bufbuild/buf/private/pkg/app/appcmd/appcmd.go:110 +0x2b
github.com/bufbuild/buf/private/pkg/app.Run({0x1800e30?, 0x26ee1c0?}, {0x180a870?, 0xc0000ce050}, 0xc00061ff00)
        github.com/bufbuild/buf/private/pkg/app/app.go:317 +0x74
github.com/bufbuild/buf/private/pkg/app.Main({0x1800e30, 0x26ee1c0}, 0x0?)
        github.com/bufbuild/buf/private/pkg/app/app.go:307 +0x9d
github.com/bufbuild/buf/private/pkg/app/appcmd.Main(...)
        github.com/bufbuild/buf/private/pkg/app/appcmd/appcmd.go:91
github.com/bufbuild/buf/private/buf/cmd/buf.Main({0x15fbbae?, 0xc0000061a0?})
        github.com/bufbuild/buf/private/buf/cmd/buf/buf.go:83 +0x47
main.main()
        github.com/bufbuild/buf/cmd/buf/main.go:20 +0x1f
```

## Steps to Reproduce
1. add dependencies from multiple registries to your `buf.yaml` file
2. run `buf mod update`

emcfarlane:
On an unknown remote ModuleIdentity may be nil. Use the fallback logic to select the correct dependency and derive the related module identity. Fixes #2725

A dial error for an invalid host will now correctly return the error formatted with a prompt to check the registry:
```
Failure: dial tcp: lookup private.registry: no such host. Are you sure "private.registry" (derived from module name "private.registry/org/dependency") is a Buf Schema Registry?
```

### DIFF
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -216,12 +216,12 @@ func getDependencies(
 		return nil, nil
 	}
 	var (
-		remote         string
-		moduleIdentity string
+		remote               string
+		moduleIdentityString string
 	)
 	if moduleConfig.ModuleIdentity != nil && moduleConfig.ModuleIdentity.Remote() != "" {
 		remote = moduleConfig.ModuleIdentity.Remote()
-		moduleIdentity = moduleConfig.ModuleIdentity.IdentityString()
+		moduleIdentityString = moduleConfig.ModuleIdentity.IdentityString()
 	} else {
 		// At this point we know there's at least one dependency. If it's an unnamed module, select
 		// the right remote from the list of dependencies.
@@ -230,11 +230,11 @@ func getDependencies(
 			return nil, fmt.Errorf(`File %q has invalid "deps" references`, existingConfigFilePath)
 		}
 		remote = selectedRef.Remote()
-		moduleIdentity = selectedRef.IdentityString()
+		moduleIdentityString = selectedRef.IdentityString()
 		container.Logger().Debug(fmt.Sprintf(
 			`File %q does not specify the "name" field. Based on the dependency %q, it appears that you are using a BSR instance at %q. Did you mean to specify "name: %s/..." within %q?`,
 			existingConfigFilePath,
-			moduleIdentity,
+			moduleIdentityString,
 			remote,
 			remote,
 			existingConfigFilePath,
@@ -274,7 +274,7 @@ func getDependencies(
 	)
 	if err != nil {
 		if !connect.IsWireError(err) && remote != bufconnect.DefaultRemote {
-			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleIdentity)
+			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleIdentityString)
 		}
 		return nil, err
 	}
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -273,7 +273,7 @@ func getDependencies(
 		}),
 	)
 	if err != nil {
-		if remote != bufconnect.DefaultRemote {
+		if !connect.IsWireError(err) && remote != bufconnect.DefaultRemote {
 			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleIdentity)
 		}
 		return nil, err
--- a/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
+++ b/private/buf/cmd/buf/command/mod/modupdate/modupdate.go
@@ -215,9 +215,13 @@ func getDependencies(
 	if len(moduleConfig.Build.DependencyModuleReferences) == 0 {
 		return nil, nil
 	}
-	var remote string
+	var (
+		remote         string
+		moduleIdentity string
+	)
 	if moduleConfig.ModuleIdentity != nil && moduleConfig.ModuleIdentity.Remote() != "" {
 		remote = moduleConfig.ModuleIdentity.Remote()
+		moduleIdentity = moduleConfig.ModuleIdentity.IdentityString()
 	} else {
 		// At this point we know there's at least one dependency. If it's an unnamed module, select
 		// the right remote from the list of dependencies.
@@ -226,10 +230,11 @@ func getDependencies(
 			return nil, fmt.Errorf(`File %q has invalid "deps" references`, existingConfigFilePath)
 		}
 		remote = selectedRef.Remote()
+		moduleIdentity = selectedRef.IdentityString()
 		container.Logger().Debug(fmt.Sprintf(
 			`File %q does not specify the "name" field. Based on the dependency %q, it appears that you are using a BSR instance at %q. Did you mean to specify "name: %s/..." within %q?`,
 			existingConfigFilePath,
-			selectedRef.IdentityString(),
+			moduleIdentity,
 			remote,
 			remote,
 			existingConfigFilePath,
@@ -269,7 +274,7 @@ func getDependencies(
 	)
 	if err != nil {
 		if remote != bufconnect.DefaultRemote {
-			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleConfig.ModuleIdentity.IdentityString())
+			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleIdentity)
 		}
 		return nil, err
 	}
--- a/private/buf/cmd/buf/command/mod/modprune/modprune.go
+++ b/private/buf/cmd/buf/command/mod/modprune/modprune.go
@@ -93,9 +93,13 @@ func run(
 	}
 	var dependencyModulePins []bufmoduleref.ModulePin
 	if len(requestReferences) > 0 {
-		var remote string
+		var (
+			remote         string
+			moduleIdentity string
+		)
 		if config.ModuleIdentity != nil && config.ModuleIdentity.Remote() != "" {
 			remote = config.ModuleIdentity.Remote()
+			moduleIdentity = config.ModuleIdentity.IdentityString()
 		} else {
 			// At this point we know there's at least one dependency. If it's an unnamed module, select
 			// the right remote from the list of dependencies.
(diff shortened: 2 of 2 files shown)


private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -269,7 +274,7 @@ func getDependencies(
 	)
 	if err != nil {
 		if remote != bufconnect.DefaultRemote {
-			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleConfig.ModuleIdentity.IdentityString())
+			return nil, bufcli.NewInvalidRemoteError(err, remote, moduleIdentity)
```
nicksnyder, line 277: This fixes the nil panic, but I am worried this doesn't result in a sane error message in the case where the server actually replied with an error instead of just being unavailable. I think you would get a double error like this `Failure: invalid remote "buf.build" for dependency "buf.build/bufbuild/protovalidate", did you mean "multitenant.staging.buf.team"?. Are you sure "multitenant.staging.buf.team" is a Buf Schema Registry?`

nicksnyder: The `Are you sure "multitenant.staging.buf.team" is a Buf Schema Registry?` should only happen if the server was actually unavailable.

doriable: I think what @nicksnyder is alluding to as the the regression from #2465 is the change that removed the unimplemented error check:

```
if connect.CodeOf(err) == connect.CodeUnimplemented && remote != bufconnect.DefaultRemote {
	return nil, bufcli.NewUnimplementedRemoteError(err, remote, moduleConfig.ModuleIdentity.IdentityString())
}
```

We can keep the fallback logic as written here, but we should also re-introduce the check for `connect.CodeUnimplemented`.

doriable: We should also add this (and reintroduce the `connect.CodeUnimplemented` check) for `mod prune` as well.

emcfarlane: For instance the current error from the issue has the code `unavailable`. Could look at making a more general error message that only identifies the module to help debug?

doriable: Had a conversation with @emcfarlane regarding the context of the original change, which is a user received a `403 forbidden`, since they were attempting to access a valid server (github.com) that is not a BSR instance, and the server tried to check their auth (their BSR token) and returned `403`.

By checking the remote explicitly against a known valid remote, we are surfacing an error message that encourages the user to check their dependency definitions to ensure that they are all valid BSR remotes, but if we isolate it to specific error codes, the user will lose that context (per the previous scenario). We can rework the error message to be a little bit more generic or we can also discuss whether this kind of error is useful at all (I think there is a merit to point users to check their dependencies, so I think it's reasonable to keep the error message this way).

nicksnyder: I agree the previous connect.CodeUnimplemented check was too specific. What I am wondering is if it is possible to get an `err` here that passes the `remote != bufconnect.DefaultRemote` check, but doesn't make sense to append the `did you mean "multitenant.staging.buf.team"` to. Maybe this makes sense in all practical cases, but not clear just by looking at the code.

For example, with this buf.yaml
```
version: v1
deps:
  - multitenant.staging.buf.team/org/dep
  - buf.build/bufbuild/protovalidate
  #- a.registry/org/dependency # removed actual private dependency
breaking:
  use:
    - FILE
lint:
  use:
    - DEFAULT
```

I get 
```
Failure: invalid remote "buf.build" for dependency "buf.build/bufbuild/protovalidate", did you mean "multitenant.staging.buf.team"?
```

I could see this being a confusing error message. No, I didn't mean `multitenant.staging.buf.team` because I want the module on the BSR (but this error message doesn't indicate that is probably the problem).

This diff is probably fine as-is, but I also think the error message(specifically the part added by `NewInvalidRemoteError`) could potentially be improved here, e.g.:
```
Failure: invalid remote "buf.build" for dependency "buf.build/bufbuild/protovalidate", did you mean "multitenant.staging.buf.team"? Depending on modules from different registries is not supported.
```

doriable: I am onboard with rephrasing the error message in this way :D

emcfarlane: We can fix the internal error message to include more details and correctly ignore server messages by checking for a `connect.IsWireErr`.

emcfarlane: Resolving, now will return the error message from the server:
```
Failure: invalid remote "buf.build" for dependency "buf.build/bufbuild/protovalidat", did you mean "multitenant.staging.buf.team"?
```
And annotate if not like the original issue.

nicksnyder (approved):

doriable (approved):

private/buf/cmd/buf/command/mod/modupdate/modupdate.go
```diff
@@ -215,9 +215,13 @@ func getDependencies(
 	if len(moduleConfig.Build.DependencyModuleReferences) == 0 {
 		return nil, nil
 	}
-	var remote string
+	var (
+		remote         string
+		moduleIdentity string
```
bufdev, line 220: same

bufdev (approved):

PR #2730, opened 2024-01-24
Enable govet nilness check and fix infractions (#2729)

stefanvanburen:
Merges latest `main` into `bufmod` and enforces the checks introduced in #2729 on this branch.


stefanvanburen:
going to do this differently because squashing the commit here will ignore the merge commit and cause issues on `bufmod`

PR #2736, opened 2024-01-26
Simplify uuidutil.ToDashless to not return an error

pkwarren:
Update the ToDashless function to not return an error, simplifying call sites. The error checking will only fail if there is a major issue in the uuid module and would be caught by existing unit tests in uuidutil_test.go.

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -208,19 +208,16 @@ func run(
 			},
 		)
 	} else {
-		lines, err = slicesext.MapError(
-			commits,
-			// Printing dashless for historical reasons.
-			func(commit bufmodule.Commit) (string, error) {
-				return uuidutil.ToDashless(commit.ModuleKey().CommitID())
-			},
-		)
-		if err != nil {
-			return err
-		}
 		if len(commits) > 1 {
 			linesErr = syserror.Newf("Received multiple commits back for a v1 module. We should only ever have created a single commit for a v1 module.")
 		}
+		lines = slicesext.Map(
+			commits,
+			// Printing dashless for historical reasons.
+			func(commit bufmodule.Commit) string {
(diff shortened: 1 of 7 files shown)


private/buf/cmd/buf/command/push/push.go
```diff
@@ -208,19 +208,16 @@ func run(
...
 			// Printing dashless for historical reasons.
-			func(commit bufmodule.Commit) (string, error) {
+			func(commit bufmodule.Commit) string {
 				return uuidutil.ToDashless(commit.ModuleKey().CommitID())
 			},
 		)
-		if err != nil {
-			return err
-		}
-		if len(commits) > 1 {
```
bufdev, line 221: This shouldn't have been deleted

pkwarren: It moved above - this wasn't dependent on the lines variable so I figured it could go above.

bufdev: ah, got it

bufdev (approved):

private/pkg/uuidutil/uuidutil.go
```diff
@@ -30,21 +30,9 @@ func New() (uuid.UUID, error) {
...
-		return "", fmt.Errorf("expected - at char 14: %q", s)
-	}
-	if s[18] != '-' {
-		return "", fmt.Errorf("expected - at char 19: %q", s)
-	}
-	if s[23] != '-' {
-		return "", fmt.Errorf("expected - at char 24: %q", s)
-	}
-	return s[0:8] + s[9:13] + s[14:18] + s[19:23] + s[24:], nil
+	return s[0:8] + s[9:13] + s[14:18] + s[19:23] + s[24:]
```
emcfarlane, line 35:
```
	var buf [32]byte
	hex.Encode(buf[:], id[:])
```
I think should work too.

PR #2744, opened 2024-01-30
Improvements to Docker tagging set up

jchadwick-buf:
Uses the official Docker action for generating tag names based on GitHub metadata. This should do the following:

- `latest` tag: Now should only be updated when there is a release tag. This is handled by the `flavor` argument.
- Added `edge` tag that matches the latest commit on main, similar to official Docker images.
- Added semver tags. A tag of v1.2.3 should result in pushes to `1`, `1.2` and `1.2.3` allowing users to track the latest v1 or latest patch release of a minor release.

~~I'm not sure how to actually test this, right now. I may need to set up a separate Docker repo on my fork.~~ Done, works.

P.S.: Merging into `main` first because we want to probably stop updating the `latest` tag ASAP.

P.P.S.: Just so it is clear, the way that Docker handles it indeed expects the semver version to look like `v1.2.3` and then pushes something like `1.2.3`. I checked this with [the docs](https://github.com/docker/metadata-action?tab=readme-ov-file#typesemver) and also [with my own test repo](https://github.com/jchadwick-buf/buf/actions/runs/7728812242/job/21070560056#step:5:104).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -94,25 +94,23 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_TOKEN }}
-      - name: set-tags
-        # See the following links for setting job outputs
-        # https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idoutputs
-        # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-output-parameter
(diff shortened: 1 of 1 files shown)


jchadwick-buf:
No need to rush getting this one in or anything, but added a couple more reviewers that I felt would be relevant (people who recently released and/or have a lot of knowledge in the release process)

Karlitos88:
The

bufdev (approved):

PR #2752, opened 2024-02-05
Add options to doc service

gilwong00:
Adds `PresentFields` method in `options_extensions_descriptor.go` and a new `Option` message in the `doc.proto` file.  
These will allows will enable custom options(extensions) to be consumed from a proto file and return in the doc service used in core.


private/pkg/protosource/field.go
```diff
@@ -207,3 +210,7 @@ func (f *field) PackedLocation() Location {
 func (f *field) ExtendeeLocation() Location {
 	return f.getLocation(f.extendeePath)
 }
+
+func (f *field) PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value {
+	return f.optionExtensionDescriptor.PresentFields()
+}
```
jhump, line 216: You should be able to remove this. `optionExtensionDescriptor` is embedded, so this method is present automatically by virtue of adding it to `optionExtensionDescriptor`.

private/pkg/protosource/protosource.go
```diff
@@ -151,6 +151,8 @@ type OptionExtensionDescriptor interface {
 	// PresentExtensionNumbers returns field numbers for all options that
 	// have a set value on this descriptor.
 	PresentExtensionNumbers() []int32
+
+	PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value
```
jhump, line 155: Needs doc comment. Here's a suggestion:
```suggestion
	// PresentFields returns a map of field descriptor to value for all
	// option fields present. This includes both normal option fields
	// and extension fields (aka custom options).
	PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
   int32 jstype = 4;
 }
 
+// Option provides information for a descriptor option that can be show in documentations.
```
jhump, line 390:
```suggestion
// Option provides information for a descriptor option that can be rendered in documentation.
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
...
+message Option {
```
jhump, line 391: Given the package this is in (registry wide), the name `Option` might be a bit too vague. How about `DescriptorOption`?

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
...
+    string string_value = 2;
+    int64 int_value = 3;
+    uint64 uint_value = 4;
+    double double_value = 5;
+    bool bool_value = 6;
+    string enum_value = 7;
+    ArrayLiteral array_value = 8;
+    MessageLiteral message_value = 9;
+  }
+}
```
jhump, line 422: I'm not sure about our stance on nesting messages. But, just to not add to the already broad `buf.alpha.registry.v1alpha1` namespace, consider nesting these inside of `DescriptorOption`.

private/pkg/protosource/option_extension_descriptor.go
```diff
@@ -126,3 +126,16 @@ func isDescendantPath(descendant, ancestor []int32) bool {
 	}
 	return true
 }
+
+func (o *optionExtensionDescriptor) PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value {
+	// Note: This does not bother to handle unrecognized extensions.
```
jhump, line 131: Since this returns all fields, not just extensions:
```suggestion
	// Note: This does not bother to handle unrecognized fields.
```

jhump (commented):
I think the `protosource` changes should probably be in their own PR. That way you can go ahead and merge that without impacting the BSR's API, and then pull those changes into core.

That would allow you to then open a PR in core that has the implementation of the new service, concurrently with the service definition PR. Maybe that's not how we usually do things, but it seems like a nice property to be able to merge the service API change and the implementation as close together as possible. Otherwise, you may want a comment on the new API `options` fields that they are new and may not yet be populated in any API responses. 🤷

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -201,6 +201,8 @@ message Service {
   ServiceOptions service_options = 8;
   // implicitly_deprecated is true if its enclosing file is deprecated.
   bool implicitly_deprecated = 9;
+
+  repeated Option options = 10;
```
jhump, line 205: I see that many of the other fields also don't have doc comments, including the `service_options` field. But it might be nice to add something to each of these to clarify their purpose, and also to move them so they're adjacent. Something like so:
```suggestion
  // implicitly_deprecated is true if its enclosing file is deprecated.
  bool implicitly_deprecated = 9;

  // A select sub-set of service options.
  ServiceOptions service_options = 8;
  // All options that are present on the service. This is a super-set of
  // service_options and uses a dynamic representation so it can also
  // accommodate custom options with arbitrary types.
  repeated Option options = 10;
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
...
+}
+
+// MessageLiteral represents an option value for a type that is a message.
+message MessageLiteral {
+  map<string, LiteralValue> fields = 1;
+}
+
+// ArrayLiteral represents an option value for a type that is a repeated field.
+// Maps are also represented as ArrayLiterals where each elements are a MessageLiteral
+// that contains a key and value field.
```
jhump, line 405:
```suggestion
// Maps are also represented using an ArrayLiteral, where each map entry is a MessageLiteral
// that contains a key and value field.
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
   int32 jstype = 4;
 }
 
+// Option provides information for a descriptor option that can be show in documentations.
+message Option {
+  // The name of the option.
```
jhump, line 392:
```suggestion
  // The name of the option. Custom options will be in the form:
  //     [fully.qualified.extension.Name]
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +387,40 @@ message FieldOptions {
...
+  string name = 1;
+  // The value of an option.
+  LiteralValue value = 2;
+}
+
+// MessageLiteral represents an option value for a type that is a message.
+message MessageLiteral {
+  map<string, LiteralValue> fields = 1;
```
jhump, line 400: You may want a note here, similar to the comment above about option names, about the convention of the key. Something like:
```suggestion
  // Map of field names to value. Extension field names will be in the form:
  //     [fully.qualified.extension.Name]
  map<string, LiteralValue> fields = 1;
```

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: An alternative to this convention, mainly the use of leading `[` and `]` in the string, is to have another field here that indicates whether this is a custom option or not.
```proto
  bool is_custom = 2; // would need to renumber value to 3
```

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: I just realized that this will end up with no information about how to order the fields when rendering other than their name. It may be sufficient to simply sort the options by name (like with normal options first and custom options after). But it might be even better to sort them by their tag number, which would require a slightly different representation here.

Also, if we wanted more structural information about which fields were extensions and which weren't (like a separate `bool` field, like the comment suggests above), then you'd need to use something other than a map here. Maybe instead something like so:
```proto
message MessageLiteral {
  repeated FieldLiteral fields = 1;
}

message FieldLiteral {
  // When is_extension is true, name is the fully-qualified name of the
  // extension. Otherwise, it is the simple name of the field.
  string name = 1;
  int32 tag = 2;
  bool is_extension = 3;
  LiteralValue value = 4;
}
```
And at that point, it looks like `Option` (or `DescriptorOption`) maybe should just be `FieldLiteral`. So then the fields added to the various element types above would look like so:
```proto
  repeated FieldLiteral options = 10;
```

proto/buf/alpha/registry/v1alpha1/doc.proto
gilwong00: Great idea, I included `is_custom` to the message

proto/buf/alpha/registry/v1alpha1/doc.proto
gilwong00: Good call, renamed to `DescriptorOption`

proto/buf/alpha/registry/v1alpha1/doc.proto
gilwong00: Replaced `DescriptiorOption` with `FieldLiteral`

gilwong00:
> I think the `protosource` changes should probably be in their own PR. That way you can go ahead and merge that without impacting the BSR's API, and then pull those changes into core.
> 
> That would allow you to then open a PR in core that has the implementation of the new service, concurrently with the service definition PR. Maybe that's not how we usually do things, but it seems like a nice property to be able to merge the service API change and the implementation as close together as possible. Otherwise, you may want a comment on the new API `options` fields that they are new and may not yet be populated in any API responses. 🤷

@jhump Good call out. I removed the usage of the `FieldLiteral` field in the messages I initially had them. This way we can still merge all the changes to the `doc.proto` without altering the service. I'll make a follow up PR to reimplement the usage of `FieldLiteral`.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -387,21 +378,28 @@ message FieldOptions {
...
+// FieldLiteral provides information for a descriptor option that can be rendered in documentation.
+message FieldLiteral {
+  // When is_extension is true, name is the fully-qualified name of the
+  // extension. Otherwise, it is the simple name of the field.
   string name = 1;
+  // The unique field number associated with the field.
+  int32 tag = 2;
+  // Indicates whether this is an extension or not.
+  bool is_extension = 3;
   // The value of an option.
```
jhump, line 390:
```suggestion
  // The field's value.
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -387,21 +378,28 @@ message FieldOptions {
   int32 jstype = 4;
 }
 
-// Option provides information for a descriptor option that can be show in documentations.
-message Option {
-  // The name of the option.
+// FieldLiteral provides information for a descriptor option that can be rendered in documentation.
```
jhump, line 381:
```suggestion
// FieldLiteral represents a field and its value. It can be used to model descriptor options
// and to model complex option values.
```

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -387,21 +378,28 @@ message FieldOptions {
...
   // The value of an option.
-  LiteralValue value = 2;
+  LiteralValue value = 4;
 }
 
 // MessageLiteral represents an option value for a type that is a message.
 message MessageLiteral {
+  // Map of field names to value. Extension field names will be in the form:
+  // [fully.qualified.extension.Name].
   map<string, LiteralValue> fields = 1;
```
jhump, line 398: I think you also want to change this type to use `FieldLiteral`.
```suggestion
  repeated FieldLiteral fields = 1;
```

jhump (commented):
Nice. Just a few more tiny things.

private/pkg/protosource/field.go
```diff
@@ -14,7 +14,9 @@
 
 package protosource
 
-import "google.golang.org/protobuf/types/descriptorpb"
```
fyockm, line 17: can probably just remove this change now too

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,47 @@ message FieldOptions {
   int32 jstype = 4;
 }
 
+// FieldLiteral represents a field and its value. It can be used to model descriptor options.
+message FieldLiteral {
```
bufdev, line 382: I don't actually see this used anywhere, where it is used?

Also how was the API created/designed?

private/pkg/protosource/option_extension_descriptor.go
```diff
@@ -126,3 +126,16 @@ func isDescendantPath(descendant, ancestor []int32) bool {
 	}
 	return true
 }
+
+func (o *optionExtensionDescriptor) PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value {
+	// Note: This does not bother to handle unrecognized fields.
+	// Should not be a problem since descriptors models in the buf CLI codebase should have them
+	// all correctly parsed and known.
+	msg := o.message.ProtoReflect()
+	fields := map[protoreflect.FieldDescriptor]protoreflect.Value{}
```
bufdev, line 135: This might happen to work, but `protoreflect.FieldDescriptor` is an interface, and (1) isn't guaranteed to be comparable (2) is likely just comparing struct pointers. If this method needs to be added, something like `ForEachPresentField(func(protoreflect.FieldDescriptor, protoreflect.Value))` may be more appropriate.

jhump: I really like this idea, to provide a callback-based range function instead of building and returning a map. 👍 

But also know that, technically, this _is_ specified to work in the package docs. Search for "Each descriptor is comparable." in the docs: https://pkg.go.dev/google.golang.org/protobuf/reflect/protoreflect. While it is possible that you could have two distinct values represent the same semantic field, that's not possible here because the message can only have a single `protoreflect.FieldDescriptor` and value per tag number.

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: The idea is to supplement the fields that look like `FieldOptions field_options = 10` (across all of the descriptor types) with a field that instead looks like `repeated FieldLiteral options = 11`. This representation can model any option and is specifically intended for modeling custom options in a way that the frontend can render into the docs.

It's not used anywhere yet at [my suggestion](https://github.com/bufbuild/buf/pull/2752#pullrequestreview-1863908719). As it is, this can be pulled into the other repo and the transformation from the `protosource.OptionExtensionDescriptor` model to the `[]*FieldLiteral` model can be implemented and tested there w/out (yet) impacting the BSR API.

But I'm not super-familiar with the typical process of implementing BSR APIs. Is it common to update the API first and then leave this stuff in the API, but unimplemented, for however long it takes to actually implement? It seemed better to me to try to implement the server-side stuff alongside the API change and have them reviewed simultaneously (or concurrently?), to eliminate useless churn (or introducing compatibility issues) in the APIs if implementation reveals that changes are required.

private/pkg/protosource/option_extension_descriptor.go
bufdev: Ugh I hate it regardless - I'm on mobile, I can't find it, but I believe you. But still ugh.

If you're ok with the callback approach, I'd prefer that, so that we can keep "no interfaces as map keys" as a good mantra to maintain in general.

jhump: Yes, I do like the idea. Let's switch to the callback approach.

private/pkg/protosource/protosource.go
```diff
@@ -152,10 +152,9 @@ type OptionExtensionDescriptor interface {
 	// have a set value on this descriptor.
 	PresentExtensionNumbers() []int32
 
-	// PresentFields returns a map of field descriptor to value for all
-	// option fields present. This includes both normal option fields
-	// and extension fields (aka custom options).
-	PresentFields() map[protoreflect.FieldDescriptor]protoreflect.Value
+	// Range implements the same interface as protoreflect.Message. This takes
+	// in a callback that iterates over a protoreflect message field descriptors and values.
```
jhump, line 156: I'd focus more on what the method does vs. comparing to protoreflect.Message.
```suggestion
	// Range iterates through all fields present in the options message invoking
	// the callback for each one. If the callback returns false, it terminates the
	// iteration and returns immediately.
```

bufdev: +1, see comment below

private/pkg/protosource/protosource.go
```diff
@@ -151,6 +151,10 @@ type OptionExtensionDescriptor interface {
 	// PresentExtensionNumbers returns field numbers for all options that
 	// have a set value on this descriptor.
 	PresentExtensionNumbers() []int32
+
+	// Range implements the same interface as protoreflect.Message. This takes
+	// in a callback that iterates over a protoreflect message field descriptors and values.
+	Range(fn func(protoreflect.FieldDescriptor, protoreflect.Value) bool)
```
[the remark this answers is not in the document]
jhump: It is intended on anything with options. I guess the name makes it a bit unclear.

I had suggested naming it `Range` so it has the same signature as `protoreflect.Message`, and thus both sources of fields implement the same interface. But now I see the name is too vague in this context. How about `RangeOptions`?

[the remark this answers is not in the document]
gilwong00: I could be off base here but I think we would want it on enums if there was a situation like this:

```
extend google.protobuf.EnumValueOptions {
  string country_code_abbrv = 2000;
}

enum COUNTRY_CODE {
  COUNTRY_CODE_UNKNOWN = 0;
  COUNTRY_CODE_UNITED_STATE_OF_AMERICA = 1 [(country_code_abbrv) =  "USA"];
}
```
In this case, I think we would want to show `[(country_code_abbrv) =  "USA"]` in the docs for `COUNTRY_CODE` however if this isn't intended then I can move this over to the `Message` interface instead.

private/pkg/protosource/option_extension_descriptor.go
```diff
@@ -127,7 +127,7 @@ func isDescendantPath(descendant, ancestor []int32) bool {
 	return true
 }
 
-func (o *optionExtensionDescriptor) Range(fn func(protoreflect.FieldDescriptor, protoreflect.Value) bool) {
+func (o *optionExtensionDescriptor) ForEachFieldDescriptor(fn func(protoreflect.FieldDescriptor, protoreflect.Value) bool) {
```
jhump, line 130: This name doesn't seem any clearer than `Range` to me. I think `RangeOptions` or maybe `RangePresentOptions` is better, since it more clearly indicates that you are ranging over _option_ fields.

gilwong00: how does `ForEachPresentOption` sound?

proto/buf/alpha/registry/v1alpha1/doc.proto
bufdev: I'm mostly interested how the specific messages were designed - are these net-new, or were they pulled from somewhere? If new-new, I need to review them more carefully. If not net-new, where did they derive from?

bufdev:
To shortchange the back and forth, I pushed up what I think this should be in https://github.com/bufbuild/buf/pull/2752/commits/89c0eeccb96f35f840e8e7c50d68aea290b3844f. I updated the function name to `ForEachPresentOption`, and updated the docs to more closely match the style of `PresentExtensionNumbers`.

However, one issue - I think there's a bit of a discrepancy here in how we are referencing options vs extensions. The docs for `PresentExtensionNumbers`:

```go
	// PresentExtensionNumbers returns field numbers for all options that
	// have a set value on this descriptor.
```

The (now updated) docs for `ForEachPresentOption` (which match the substance of what was there pre-my-edits):

```go
	// ForEachPresentOption iterates through all options that have a set value on this
	// descriptor, invoking fn for each present option.
```

The discrepency being in one case, we're saying "extension numbers", and in the other, we're saying "options", for what (doc-wise) is obstensibly the same item. What should we do here? I *think* the answer is that the docs for `PresentExtensionNumbers` are incorrect, and need to be scoped (which matches what is going on in the function, namely a `fieldDescriptor.IsExtension()` call), but I'm not entirely sure.

A note: when this is merged, `main` will have to be merged into `bufmod`, and `protosource` has moved to `bufprotosource`. @gilwong00 you'll be on hook for this. https://github.com/bufbuild/buf/tree/bufmod/private/bufpkg/bufprotosource

private/pkg/protosource/protosource.go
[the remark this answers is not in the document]
bufdev: got it - this should be on `OptionExtensionDescriptor` then.

private/pkg/protosource/option_extension_descriptor.go
bufdev: yep thats what I got to as well, see below

proto/buf/alpha/registry/v1alpha1/doc.proto
gilwong00: These are net new

jhump:
@bufdev, the discrepancy is intentional and kind of unavoidable because they do different things. `PresentExtensionNumbers` is just for extensions/custom options. This was used in combination with the other two methods I think, for getting source code info for custom options (like to do linting on protovalidate stuff I think?).

But the new method gets all options, not just extensions. This is because I think it will be best to treat all options uniformly instead of having to special-case normal options vs custom options. (Especially with editions: special-casing the handling of the `FeatureSet` message doesn't make sense, and leaning on the same code that renders custom options makes more sense.)

bufdev:
That makes sense - can we update the comments on `PresentExtensionNumbers` to reflect this?

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,47 @@ message FieldOptions {
...
+// that contains a key and value field.
+message ArrayLiteral {
+  repeated LiteralValue elements = 1;
+}
+
+// LiteralValue represents the actual value of the extension.
+message LiteralValue {
+  oneof value {
+    string string_value = 2;
+    int64 int_value = 3;
```
bufdev, line 412: Why does this not contain all the primitive types? This API shouldn't be frontend-specific.

gilwong00: updated this to include all the primitive types

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: Using `FieldOptions` (or other option types) makes consuming the value much more complicated since it will have extensions. So a consumer would need to have the schemata for all custom options in order to interpret the data.

Basically same goes for well-known types -- the closest would be `google.protobuf.Value`, but it represents JSON data (so does not distinguish between any of the numeric types, does not support enums except as string or int values, and supports null which isn't appropriate here). Aside from that, the only thing that can dynamically represent any kind of message is `google.protobuf.Any` and that has the same drawbacks as extensions: the consumer must have the schemata for everything in order to interpret this.

This is _kind of like_ `google.protobuf.Value`, except it is for modeling arbitrary Protobuf values instead of JSON values. It's intended to provide a very easy way to work with the data without the schema.

timostamm: While we could use the `google.protobuf.*Options` messages here, we would also have to ship the descriptors for the extensions and types they use. This only makes sense to me if we use descriptors for the entire package docs, replacing the current layer. This is feasible and has advantages, but it requires extra steps to split descriptor sets by package... and would obviously be a major pivot. 

`google.protobuf.Value` cannot store the values NaN and Infinity for `DOUBLE`. It's unlikely that they appear in option values, but they can. It also cannot represent `BYTES`. This means that the data will contain strings with special meaning - base64 encoded `BYTES`, `"NaN"`, `"Infinity"`, and `"-Infinity"` - that the client cannot distinguish from other strings without knowing the schema, which leads to the same problem as with `google.protobuf.*Options`.

I agree with Josh here.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,55 @@ message FieldOptions {
...
+    fixed64 fixed64_value = 8;
+    sfixed32 sfixed32_value = 9;
+    sfixed64 sfixed64_value = 10;
+    float float_value = 11;
+    double double_value = 12;
+    bool bool_value = 13;
+    string enum_value = 14;
+    bytes bytes_value = 15;
+    ArrayLiteral array_value = 16;
+    MessageLiteral message_value = 17;
```
timostamm, line 426: If I'm not mistaken, this allows us to represent repeated values like they appear in the source, retaining whether they were specified as separate fields or a joined repeated field.

To elaborate - the following two options are identical:

```proto
option (google.api.http) = {
  get: "/v1/{name=messages/*}"
  additional_bindings {
    get: "xxx"
  }
  additional_bindings {
    get: "yyy"
  }
};
```

```proto
option (google.api.http) = {
  get: "/v1/{name=messages/*}"
  additional_bindings: [
    {
      get: "xxx"
    },
    {
      get: "yyy"
    }
  ]
};
```

I think it would be confusing to users if they use one form, but the docs show a different form.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,55 @@ message FieldOptions {
...
+    int32 int32_value = 3;
+    int64 int64_value = 4;
+    uint32 uint32_value = 5;
+    uint64 uint43_value = 6;
+    fixed32 fixed32_value = 7;
+    fixed64 fixed64_value = 8;
+    sfixed32 sfixed32_value = 9;
+    sfixed64 sfixed64_value = 10;
+    float float_value = 11;
+    double double_value = 12;
```
timostamm, line 421: It's likely rare in practice, but I want to point out that we'll loose exponential or hexadecimal notation.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,55 @@ message FieldOptions {
...
+  bool is_extension = 3;
+  // The field's value.
+  LiteralValue value = 4;
+}
+
+// MessageLiteral represents an option value for a type that is a message.
+message MessageLiteral {
+  // Map of field names to value. Extension field names will be in the form:
+  // [fully.qualified.extension.Name].
+  repeated FieldLiteral fields = 1;
```
timostamm, line 398: I suppose Any will have a "synthetic" field with the type URL here?

https://protobuf.com/docs/language-spec#any-messages

proto/buf/alpha/registry/v1alpha1/doc.proto
timostamm: @bufdev, is this API really not frontend specific? The comment on the service says otherwise: https://github.com/bufbuild/buf/blob/17b4060d969e99b151d125ff0d5ff007aa271250/proto/buf/alpha/registry/v1alpha1/doc.proto#L19

Either way, the API is undoubtedly specific to docs, and we ultimately want to present the option value as text. I cannot think of a use case where I would want to distinguish even `DOUBLE` from `FLOAT` in the context of docs, and I'd prefer if we use the widest types only. I don't think it's an issue to include all types here, just that I don't see the need.

timostamm (commented):
IMO `FieldLiteral` is appropriate, although we should be aware that we cannot retain some details of the original form regarding formatting, numeric notation, and comments inside the option value ([practical example](https://github.com/bufbuild/protovalidate/blob/8a3d2c522bcc0d71d126cd55b2a1406d576f7bbe/examples/option_string_len.proto#L20)).

I also want to point out that by modeling option values agnostically, clients that want to render them in the protobuf text format have to implement formatting for non-infinite float values and proper escaping of string and bytes. An opinionated alternative would be to provide all values as already formatted text.

proto/buf/alpha/registry/v1alpha1/doc.proto
bufdev: In general, APIs like this (and APIs in general) really shouldn't be designed with a specific caller in mind, but in this specific instance I don't know how much the field difference matters.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,47 @@ message FieldOptions {
...
+
+// LiteralValue represents the actual value of the extension.
+message LiteralValue {
+  oneof value {
+    string string_value = 2;
+    int64 int_value = 3;
+    uint64 uint_value = 4;
+    double double_value = 5;
+    bool bool_value = 6;
+    string enum_value = 7;
```
bufdev, line 416: Why is this `string`?

gilwong00: We going to be showing this in docs so the name should be more useful than the number value to users.

bufdev: @jhump should this semantically be ie `enum_value_name_value`, if that makes sense? It's not an `enum`, its an enum value, and its the name of the enum value...I don't know.

bufdev (approved):
Overall, for long-term maintainability here, I wish we could lean less into this as a frontend-specific API, and put more of the computation for docs into the FE going forward, i.e. I wish all that the generated documentation needed to do was query the reflection endpoint, get a `FileDescriptorSet`, and the frontend uses `protobuf-es` to do the rest. I recognize that's a major change and not in scope, so approving, but I do wonder if we're adding more pain for us down the road.

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: I think it will be okay to render them uniformly instead of trying to re-create the source. And I'd suggest always doing the second form (array-literal style syntax, which will be more intuitive for most readers).

For one, I think it will look nicer if the APIs are presented consistently in documentation, instead of trying to preserve possibly idiosyncratic syntax in some sources. And two, re-creating the original source from just the descriptor (i.e. trying to re-construct formatting and such from just the source code info) is not really feasible and trying to merge actual sources with the descriptor into something that is more faithful to the original source would also be a real bear to implement.

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: Great catch!

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -377,6 +378,55 @@ message FieldOptions {
   int32 jstype = 4;
 }
 
+// FieldLiteral represents a field and its value. It can be used to model descriptor options.
+message FieldLiteral {
+  // When is_extension is true, name is the fully-qualified name of the
+  // extension. Otherwise, it is the simple name of the field.
+  string name = 1;
+  // The unique field number associated with the field.
+  int32 tag = 2;
```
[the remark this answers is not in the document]
jhump: Yes, good point. For hyper-linking to the docs for an extension (or message type in an `Any`), the module ref will be needed.

proto/buf/alpha/registry/v1alpha1/doc.proto
jhump: That seems like a lot of ceremony. It represents an enum value, so seems like a doc comment clarifying that this is the enum value name feels like it suffices. Another option is to extract out another type -- like maybe an `EnumLiteral` that has both a name and a number. But, honestly, that felts a little overkill 🤷

PR #2753, opened 2024-02-06
Fix deadlock in `NewCommit`

saquibmian:
When `CommitWithExpectedDigest` is passed in, the `ModuleKey` is wrapped, but the wrapper ends up calling itself.

### DIFF
--- a/private/bufpkg/bufmodule/commit.go
+++ b/private/bufpkg/bufmodule/commit.go
@@ -76,18 +76,19 @@ func newCommit(
 		option(commitOptions)
 	}
 	if commitOptions.expectedDigest != nil {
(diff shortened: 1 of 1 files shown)


saquibmian:
Windows tests are failing, but it looks like they're also failing on `bufmod`, and it's the same failure.

bufdev:
Point of order: this isn't a deadlock, this is just a recursive infinite loop https://www.geeksforgeeks.org/conditions-for-deadlock-in-operating-system/ (this just threw me off a little bit when reading heh) - I fixed this and added a comment in https://github.com/bufbuild/buf/commit/bddd1d3495f7a37df9e1cc895baa1266dea3582c so you should be good to go.

PR #2754, opened 2024-02-07
Update to use Golang 1.22 and drop Golang 1.19 support

bufdev:


pkwarren:
I think we can also clean up these now:
```
./private/buf/bufcli/errors_tls_go120.go
./private/buf/bufcli/errors_tls_go119.go
```

bufdev:
Done

pkwarren (approved):

PR #2770, opened 2024-02-15
Add options to doc response

gilwong00:
Part 2 of https://github.com/bufbuild/buf/pull/2752. 

Adding `FieldLiteral` to `Message`, `Enum`, `Service`, `Method` and `EnumValue` responses


proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -320,6 +336,10 @@ message Message {
   MessageOptions message_options = 10;
   // implicitly_deprecated is true if its enclosing file or parent element is deprecated.
   bool implicitly_deprecated = 11;
+  // All options that are present on the message. This is a super-set of
+  // message_options and uses a dynamic representation so it can also
+  // accommodate custom options with arbitrary types.
+  repeated FieldLiteral options = 12;
```
bufdev, line 342: It's weird that we have `message_options` and then `options` where `options` is defined as "all options on the message". Do you instead want to deprecate `message_options` and say that this supersedes it:?

gilwong00: Yea we can do that. Should we also do the same with `enum_options` , `method_options` and 'service_options` as well?

jhump: Yes. In all cases, the new `repeated FieldLiteral options` field deprecates the old `*_options` field.

bufdev: Perfect

gilwong00: Added `[deprecated = true]` to all `*_options` fields

gilwong00: I can completely remove `*_options` fields from the messages in a follow up PR after the changes in core ships.

proto/buf/alpha/registry/v1alpha1/doc.proto
```diff
@@ -198,9 +198,14 @@ message Service {
   string file_path = 5;
   Location location = 6;
   repeated Method methods = 7;
-  ServiceOptions service_options = 8;
+  ServiceOptions service_options = 8 [deprecated = true];
```
bufdev, line 201: Can you move this to the bottom of the message definition, and add a comment "Use options instead."

```
  ...
  repeated FieldLiteral options = 10;
  // Use options instead.
}
```

Also, is `ServiceOptions` used elsewhere, or is this the only use? If not used elsewhere, also deprecate it:

```
message ServiceOptions {
  option deprecated = true;
  ...
}
```

Apply this across all of the below. Sorry for the back and forth.

bufdev (approved):

PR #2780, opened 2024-02-21
Prevent pushing with empty Label/Tag names

saquibmian:
`flags.Draft` and `flags.Branch` may be empty string, and they were being added to the labels name. Instead, I conditionally add them.

I also chose to detect empty strings in `flags.Labels` and `flags.Tags`, via `--label ''` or `--tag ''`.

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -140,7 +140,7 @@ func run(
 	container appext.Container,
 	flags *flags,
 ) (retErr error) {
-	if err := validateCreateFlags(flags); err != nil {
+	if err := validateFlags(flags); err != nil {
 		return err
 	}
 
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
 	return nil
 }
 
+func validateFlags(flags *flags) error {
+	if err := validateCreateFlags(flags); err != nil {
+		return err
+	}
+	if err := validateLabelFlags(flags); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateLabelFlags(flags *flags) error {
+	for _, label := range flags.Labels {
+		if label == "" {
+			return appcmd.NewInvalidArgumentErrorf(
+				"--%s requires a non-empty string.",
+				labelFlagName,
+			)
+		}
+	}
+	for _, tag := range flags.Tags {
+		if tag == "" {
+			return appcmd.NewInvalidArgumentErrorf(
+				"%s|--%s requires a non-empty string.",
+				tagFlagShortName,
+				tagFlagName,
+			)
+		}
+	}
+	return nil
+}
+
 func validateCreateFlags(flags *flags) error {
 	if flags.Create {
 		if flags.CreateVisibility == "" {
@@ -371,16 +402,16 @@ func validateCreateFlags(flags *flags) error {
 }
 
 func combineLabelLikeFlags(flags *flags) []string {
-	return slicesext.ToUniqueSorted(
-		append(
-			flags.Labels,
-			append(
-				flags.Tags,
-				flags.Draft,
-				flags.Branch,
-			)...,
-		),
-	)
+	labels := make([]string, 0, len(flags.Labels)+len(flags.Tags))
+	labels = append(labels, flags.Labels...)
+	labels = append(labels, flags.Tags...)
+	if flags.Draft != "" {
+		labels = append(labels, flags.Draft)
+	}
+	if flags.Branch != "" {
+		labels = append(labels, flags.Branch)
+	}
+	return slicesext.ToUniqueSorted(labels)
 }
 
 func newRequireModuleFullNameOnUploadError(module bufmodule.Module) error {
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -360,7 +360,7 @@ func validateCreateFlags(flags *flags) error {
 	if flags.Create {
 		if flags.CreateVisibility == "" {
 			return appcmd.NewInvalidArgumentErrorf(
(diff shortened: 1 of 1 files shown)


unmultimedio (approved):
lgtm!

private/buf/cmd/buf/command/push/push.go
```diff
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
...
+			return appcmd.NewInvalidArgumentErrorf(
+				"--%s requires a non-empty string.",
+				labelFlagName,
+			)
+		}
+	}
+	for _, tag := range flags.Tags {
+		if tag == "" {
+			return appcmd.NewInvalidArgumentErrorf(
+				"%s|--%s requires a non-empty string.",
```
stefanvanburen, line 371: drive by, but should the short flag have a `-` in front?

```suggestion
				"-%s|--%s requires a non-empty string.",
```

saquibmian: Ah I `x`'d that by accident. Let me push that up.

unmultimedio (approved):

private/buf/cmd/buf/command/push/push.go
```diff
@@ -371,16 +402,16 @@ func validateCreateFlags(flags *flags) error {
...
-		append(
-			flags.Labels,
-			append(
-				flags.Tags,
-				flags.Draft,
-				flags.Branch,
-			)...,
-		),
-	)
+	labels := make([]string, 0, len(flags.Labels)+len(flags.Tags))
```
bufdev, line 405: Why not continue to use previous style, `labels := append(flag.Labels, flags.Tags...)`?

private/buf/cmd/buf/command/push/push.go
```diff
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
...
+			return appcmd.NewInvalidArgumentErrorf(
+				"--%s requires a non-empty string.",
+				labelFlagName,
+			)
+		}
+	}
+	for _, tag := range flags.Tags {
+		if tag == "" {
+			return appcmd.NewInvalidArgumentErrorf(
+				"-%s|--%s requires a non-empty string.",
```
bufdev, line 371: All other usages of `appcmd.NewInvalidArgumentErrorf` (save one) in the codebase do not display short names, and do not use punctuation - update to `--%s requires a non-empty string`

private/buf/cmd/buf/command/push/push.go
```diff
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
...
+	if err := validateCreateFlags(flags); err != nil {
+		return err
+	}
+	if err := validateLabelFlags(flags); err != nil {
+		return err
+	}
+	return nil
+}
+
+func validateLabelFlags(flags *flags) error {
```
bufdev, line 359: Put below `validateCreateFlags` since this is the call order and it is more natural

private/buf/cmd/buf/command/push/push.go
```diff
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
...
+	for _, label := range flags.Labels {
+		if label == "" {
+			return appcmd.NewInvalidArgumentErrorf(
```
bufdev, line 362: No need for newlines

private/buf/cmd/buf/command/push/push.go
```diff
@@ -346,6 +346,37 @@ func createUploadModulesIfNotExist(
...
+				"--%s requires a non-empty string.",
+				labelFlagName,
+			)
+		}
+	}
+	for _, tag := range flags.Tags {
+		if tag == "" {
+			return appcmd.NewInvalidArgumentErrorf(
```
bufdev, line 370: No need for newlines

private/buf/cmd/buf/command/push/push.go
saquibmian: `append` [can mutate the underlying array of its first argument](https://trstringer.com/golang-append/) (`flag.Label`). It shouldn't matter in this case because we're not overwriting indexes, but in general I avoid using `append` to copy a slice I don't own. Happy to change this back if you think it's a non-issue.

private/buf/cmd/buf/command/push/push.go
saquibmian: Done!

private/buf/cmd/buf/command/push/push.go
saquibmian: Done!

private/buf/cmd/buf/command/push/push.go
saquibmian: Did you want me to remove the trailing period on the errors returned by `validateCreateFlags`?

private/buf/cmd/buf/command/push/push.go
bufdev: Well every day's a school day I guess, I don't know how I've gone this long without running into some bug based on this. Sounds good. A slight nit, I might clean it up to `append(slicesext.Copy(flags.Labels), flags.Tags...)` but doesn't matter much.

private/buf/cmd/buf/command/push/push.go
bufdev: Yea for sure, that was a miss on my part

saquibmian: Done!

private/buf/cmd/buf/command/push/push.go
saquibmian: I cleaned it up using `slicesext.Copy`

bufdev: @akshayjshah on my case about how could I be so naive

PR #2781, opened 2024-02-21
Update protovalidate, connect-go, and other go deps

pkwarren:


private/bufpkg/bufcheck/buflint/internal/buflintvalidate/field.go
```diff
@@ -258,7 +258,7 @@ func checkFieldFlags(
 			adder.getFieldRuleName(),
 		)
 	}
-	if fieldConstraints.GetRequired() && fieldConstraints.GetIgnoreEmpty() {
+	if fieldConstraints.GetRequired() && (fieldConstraints.GetIgnoreEmpty() || fieldConstraints.GetIgnore() == validate.Ignore_IGNORE_EMPTY) {
```
pkwarren, line 261: Kept using the previous deprecated field and also added the new enum with the same semantics. We can remove this when protovalidate reaches v1 and the previous ignore empty boolean is removed. See https://github.com/bufbuild/protovalidate/pull/137 for more details.

cc @rodaine

bufdev: https://github.com/bufbuild/registry-proto/pull/75 means we probably need to update this further to use the non-deprecated name

.golangci.yml
```diff
@@ -186,6 +186,9 @@ issues:
     - linters:
         - staticcheck
       text: "GetDeprecatedLegacyJsonFieldConflicts is deprecated"
+    - linters:
+        - staticcheck
+      text: "GetIgnoreEmpty is deprecated"
```
bufdev, line 191: Would optimally scope this to the specific file (the one above should be too, but doesn't need to be in this PR)

pkwarren: Fixed.

private/bufpkg/bufcheck/buflint/internal/buflintvalidate/field.go
pkwarren: Updated to the latest protovalidate.

rodaine (approved):

bufdev (approved):

PR #2782, opened 2024-02-22
Skip WKT files when exporting

unmultimedio:


private/buf/cmd/buf/command/export/export.go
```diff
@@ -176,6 +177,10 @@ func run(
 		return errors.New("no .proto target files found")
 	}
 	for _, imageFile := range image.Files() {
+		if datawkt.Exists(imageFile.Path()) {
```
bufdev, line 180: This isn't what you want - this will ignore a file if it has the same name as a WKT, regardless of if that file was actually either (1) in your module or (2) an actual dependency. Examples from old CLI that will export the file, but will not after this change:

- A module with a dep on `buf.build/protocolbuffers/wellknowntypes`.
- A module that has a file named `google/protobuf/timestamp.proto` locally.

bufdev (changes requested):

private/buf/cmd/buf/command/export/export.go
unmultimedio: you're right, I'll fix

bufdev: Looked into it, fixed in https://github.com/bufbuild/buf/commit/3c4812b9a484aa64d4f4b2c5618c7032369eea7a

PR #2792, opened 2024-02-28
Support recently added "source_file_descriptors" field of CodeGeneratorRequest

jhump:
I've also made sure this branch can apply to `bufmod`. Here's my tentative cherry-pick:
https://github.com/bufbuild/buf/compare/bufmod...jh/strip-source-only-on-bufmod


bufdev (commented):
An overall thing: I think we might want to make it so that we can select source-retention options or not at the `ImageFile` level. A proposal for how I'd do this:

```
type ImageFile interface {
  ...
  FileDescriptorProto() *descriptorpb.FileDescriptorProto
  FileDescriptorProtoWithoutSourceRetentionOptions() (*descriptorpb.FileDescriptorProto, error)
  ...
}
```

Then, in `imageFile`, I would do:

```
type imageFile struct {
  ...
  getFileDescriptorProtoWithoutSourceRetentionOptions func() (*descriptorpb.FileDescriptorProto, error)
}
```

And on construction:

```
func newImageFileNoValidate(...) {
  imageFile := &imageFile{...}
  imageFile.getFileDescriptorProtoWithoutSourceRetentionOptions =sync.OnceValues(imageFile.getFileDescriptorProtoWithoutSourceRetentionOptionsUncached)
  return imageFile
}
```

And at the bottom of `image_file.go`, before `isImageFileInfo/isImageFile`:

```
func (f *imageFile) getFileDescriptorProtoWithoutSourceRetentionOptionsUncached() (*descriptorpb.FileDescriptorProto, error) {
  return stripSourceRetentionOptions(f.fileDescriptorProto)
}
```

private/bufpkg/bufimage/util.go
```diff
@@ -424,7 +424,12 @@ func imageToCodeGeneratorRequest(
 		request.Parameter = proto.String(parameter)
 	}
 	for i, imageFile := range imageFiles {
-		request.ProtoFile[i] = imageFile.FileDescriptorProto()
+		// ProtoFile should include runtime-retained options only. So strip source-only options.
```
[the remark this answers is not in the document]
bufdev: If you implement `ImageFile.FileDescriptorProtoWithoutSourceRetentionOptions()`, then this becomes `if in file_to_generate { use FileDescriptorProto() } else { use FileDescriptorProtoWithoutSourceRetentionOptions() }`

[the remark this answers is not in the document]
jhump: Good catch! Silly me for assuming this would be more consistent/coherent. What a terrible API :/

I guess they moved out source-only options for `files_to_generate` just so that existing plugins (not updated for these new fields) that embed descriptors wouldn't need any updates to respect the source-only retention (and inadvertently _not_ embed them). But the fact that `proto_file` behavior is now bimodal feels... dumb.

private/bufpkg/bufimage/strip_source_only.go
```diff
@@ -0,0 +1,377 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+func stripSourceOnlyOptions[M proto.Message](options M) (M, error) {
```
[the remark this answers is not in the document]
jhump: Fixed in 2d6960a

jhump:
> An overall thing: I think we might want to make it so that we can select source-retention options or not at the ImageFile level.

This can expose the functionality for `*descriptorpb.FileDescriptorProto`. But we'll also need this functionality for `*imagev1.ImageFile` (so we can optionally exclude these options from the output of `buf build`). So, in ca03cd0, I instead opted to export `StripSourceRetentionOptionsFromFile` and make it generic so that it can handle either representation of a file. WDYT?

Also, I tackled the TODO for more validation of the `CodeGeneratorRequest` in 6e767e4. WDYT? Too much? I can easily back it out if so.

private/pkg/protodescriptor/protodescriptor.go
```diff
@@ -161,6 +166,28 @@ func ValidateCodeGeneratorRequestExceptFileDescriptorProtos(request *pluginpb.Co
 	if err := ValidateProtoPaths("CodeGeneratorRequest.FileToGenerate", request.FileToGenerate); err != nil {
 		return err
 	}
+	if len(request.SourceFileDescriptors) != len(request.FileToGenerate) {
```
bufdev, line 169: We can't be guaranteed in this package that the input will be generated by a recent compiler, it is valid for the length of `source_file_descriptors` to be 0 here.

Re: PR comment, I am likely just going to have all this validation live in `protoplugin` and import it here, so we don't need to redo it anyways (although I will add the duplicate check over there).

private/bufpkg/bufimage/source_retention_options.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+	if changed {
+		dirty = true
+	}
+
+	if !dirty {
+		return file, nil
+	}
+
+	newFile, err := shallowCopy(file)
```
bufdev, line 78: What is shallow about the copy exactly? The method documentation implies that I'm getting a full copy back, which would say to me I can do whatever I want with the return value of the method, but this would imply otherwise I think

private/bufpkg/bufimage/source_retention_options.go
```diff
@@ -0,0 +1,400 @@
...
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// StripSourceRetentionOptionsFromFile returns a file descriptor proto that omits any
+// options that in file that are defined to be retained only in source. If file has no
+// such options, then it is returned as is. If it does have such options, a copy is
+// made; the given file will not be mutated.
+//
+// The given file must be either a *descriptorpb.FileDescriptorProto or a
+func StripSourceRetentionOptionsFromFile[F protodescriptor.FileDescriptor](
```
bufdev, line 33: I'm already finding usage for this to construct `CodeGeneratorRequests` outside of this repository (specifically for `protoplugin` testing) - what do you think about moving a `FileDescriptorProto`-specific version of this to say `protocompile` in some capacity? I'm not sure if `protocompile` is the right place (`protobuf-go` is really the right place), but just somewhere else that we can consume it.

Then, I would have a private function here that takes a `*imagev1.Image`, calls the public function for `FileDescriptorProto`, and then uses `fileDescriptorProtoToProtoImageFile` (which is in this package).

private/pkg/protodescriptor/protodescriptor.go
jhump: Ah, right. My bad. I wasn't entirely sure where this was used. But it makes sense that it could be used in the recipient of a request, not just the producer, which means it must be able to handle an older form of the request.

private/bufpkg/bufimage/source_retention_options.go
jhump: I can update the doc comments. It definitely wasn't meant to suggest that you can do whatever you want the value but rather that it does not mutate the incoming value.

It copies only the elements that changed, which are leaf messages that were mutated and their ancestors in the hierarchy. (So if any leaf is changed, you'll get back a different `*descriptorpb.FileDescriptorProto`, but it can still share structure with the input.)

private/bufpkg/bufimage/source_retention_options.go
jhump: Yeah, I think `protocompile` makes sense. Computing this stuff is the job of a compiler anyhow (though it's usually the code-gen part of the compiler, whereas protocompile only produces intermediate representations).

It probably makes the most sense inside the options sub-package there.

jhump: https://github.com/bufbuild/protocompile/pull/250

private/pkg/protodescriptor/protodescriptor.go
bufdev: The issue hasn't been fixed, we'll still fail for old payloads.

private/bufpkg/bufimage/source_retention_options.go
bufdev: This function needs to be updated post-250, but I assume thats just coming up next

jhump: Yeah, there now.

private/pkg/protodescriptor/protodescriptor.go
jhump: > The issue hasn't been fixed, we'll still fail for old payloads.

Because I hadn't pushed any updates 😛 

Fixed now. I just removed most of the extra validation. I realized that the simple check that was here wasn't right anyway because `source_file_descriptors` is it not _specified_ to be in the same order as `file_to_generate`. And it's also not in the same order in practice, neither from the code I just wrote for `buf generate` nor for `protoc`. Both (`protoc` and `buf`) actually build it in the same order as `proto_file`, but pruned down to only the paths in `file_to_generate`.

So doing the _right_ check is a little more complicated and seemed like it warranted a separate PR with a bevy of new tests to make sure it's right. So I can push a PR up for that later.

private/bufpkg/bufimage/util.go
```diff
@@ -433,9 +435,18 @@ func imageToCodeGeneratorRequest(
 			includeWellKnownTypes,
 		) {
 			request.FileToGenerate = append(request.FileToGenerate, imageFile.Path())
+			// Source-retention options for items in FileToGenerate are provided in SourceFileDescriptors.
+			request.SourceFileDescriptors = append(request.SourceFileDescriptors, fileDescriptorProto)
+			// And the corresponding descriptor in ProtoFile will have source-retention options stripped.
+			var err error
+			fileDescriptorProto, err = options.StripSourceRetentionOptionsFromFile(fileDescriptorProto)
+			if err != nil {
+				return nil, fmt.Errorf("failed to process %q for code generator request: %w", imageFile.Path(), err)
```
bufdev, line 444: `failed to strip source-retention options for file %q when constructing a CodeGeneratorRequest`

jhump, line 444:
```suggestion
				return nil, fmt.Errorf("failed to strip source-retention options for file %q when constructing a CodeGeneratorRequest: %w", imageFile.Path(), err)
```

jhump: Done

bufdev (approved):
Thanks for getting through this! Merge into `bufmod` after we merge here.

PR #2796, opened 2024-03-05
Fix wrong v1/v1beta proto file in `newUniversalProtoContentForV1Beta1 `

saquibmian:
We were using the `buf.yaml` file instead of the `buf.lock` file, resulting in duplicate file errors when calculating the B4 digest.


bufdev (approved):
Fun/whoops!

PR #2797, opened 2024-03-05
Add archive registry support for curated plugin

cyinma:


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -232,9 +235,10 @@ message RegistryConfig {
     MavenConfig maven_config = 3;
     SwiftConfig swift_config = 4;
     PythonConfig python_config = 5;
+    ArtifactConfig artifact_config = 6;
```
mfridman, line 238: I could see this going both ways, but I was thinking this would be called `archive_config`.

However, I could see a world where we want to add an `archive_config` to a plugin to modify it's behavior within the archive registry (for example, propagating a plugin option), but not necessarily show that plugin as part of the bundled artifact feature.

I wonder if the config should be named `archive_config` and have an "include in bundled artifacts" option. This would allow us to set _some_ plugins to `true` and retain the flexibility to have ad-hoc `archive_config` without affecting the bundled artifact feature.

cc @pkwarren @stefanvanburen if you have some thoughts let us know.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -232,9 +235,10 @@ message RegistryConfig {
...
   }
   // Reserved for future remote registry types.
-  reserved 6 to 9;
+  reserved 7 to 9;
```
bufdev, line 241: Just a point of order: this is sort of an abuse of the `reserved` keyword, `reserved` is really meant to say "these were previously used", outside of hyper-optimization theres no need to reserve these fields ahead of time.

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
bufdev: I would +1 on archive

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
bufdev: You can still group them together in the source code, regardless of tag - tag is kind of irrelevant unless you're trying to save one byte (you're not)

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
stefanvanburen: yeah, archive makes sense to me.

cyinma: chatted with Mike, renamed this to archive and we can start from there

mfridman (approved):

PR #2799, opened 2024-03-05
Add buftarget package

doriable:
This PR adds a `buftarget` package, which provides path targeting information
for a storage bucket:

- `BucketTargeting` provides information on a `ControllingWorkspace` if one was
found in the bucket, the input path, target paths, and target exclude paths
- `TerminateAtControllingWorkspace` has been consolidated to this package

The way the package is used is that when `bufctl.Controller` calls `buffetch/internal.Reader`
to get the storage bucket , `buffetch/internal.Reader` now also returns `buftarget.BucketTargeting`
alongside the bucket. `bufworkspace.WorkspaceProvider` then handles the workspace and module
targeting using `buftarget.BucketTargeting`. This removes the need to set input path, target paths,
and target exclude paths as workspace provider options, and is instead just passed through
`workspaceTargeting` and `moduleTargeting`.

This change also introduces some fallback logic in `workspaceTargeting` for inputs that did not
resolve to a controlling workspace.

Lastly, tests that were skipped previously for path/exclude-path behavior issues have been
unskipped and resolved accordingly.


bufdev (approved):
Approving to get this into `bufmod` as this is clearly an improvement over the status quo - we're going to do a line-by-line review as we go, but I've QA'ed the logic and done a basic pass, and we're in a much better state after this.

PR #2801, opened 2024-03-05
Bump google.golang.org/protobuf to v1.33.0

stefanvanburen:
Actually pulls in the changes to go.mod - followup to #2800.


bufdev (approved):

stefanvanburen:
We'll need to update to take into account https://github.com/protocolbuffers/protobuf/pull/15164.

private/pkg/protosource/file.go
```diff
@@ -143,7 +143,9 @@ func (f *file) PyGenericServices() bool {
 }
 
 func (f *file) PhpGenericServices() bool {
-	return f.fileDescriptor.GetOptions().GetPhpGenericServices()
+	// Support for PhpGenericServices was removed in
```
bufdev, line 146: This isn't really the "right" fix, really  we should be removing all of this. Give me a minute.

bufdev (approved):

PR #2802, opened 2024-03-06
Fix `WorkspaceDepManager` to handle v1 module default behavior and rename `InputDirPath` to `SubDirPath`

doriable:
We were not defaulting to `v1` module behavior in `WorkspaceDepManager`
when we do not find a `v2` workspace before this. Now, if no `v2` workspace
was found for `WorkspaceDepManager`, we treat the `subDirPath` as a `v1` module
by default.
Also, this renamed `InputDirPath` to `SubDirPath` for clarity.
Lastly, this puts in a quick fix for the a `buf config init` test, since the shape of the
default config has changed.

### DIFF
--- a/private/buf/bufworkspace/workspace_dep_manager_provider.go
+++ b/private/buf/bufworkspace/workspace_dep_manager_provider.go
@@ -90,5 +90,5 @@ func (w *workspaceDepManagerProvider) GetWorkspaceDepManager(
 	}
 	// Otherwise we simply ignore any buf.work.yaml that was found and attempt to build
 	// a v1 module at the SubDirPath
-	return newWorkspaceDepManager(bucket, bucketTargeting.InputDirPath(), false), nil
+	return newWorkspaceDepManager(bucket, bucketTargeting.SubDirPath(), false), nil
 }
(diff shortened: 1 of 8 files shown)


private/buf/bufworkspace/workspace_dep_manager_provider.go
```diff
@@ -82,12 +81,14 @@ func (w *workspaceDepManagerProvider) GetWorkspaceDepManager(
...
+		}
 	}
-	bufYAMLFile := controllingWorkspace.BufYAMLFile()
-	if bufYAMLFile.FileVersion() == bufconfig.FileVersionV2 {
-		return newWorkspaceDepManager(bucket, controllingWorkspace.Path(), true), nil
-	}
-	return newWorkspaceDepManager(bucket, bucketTargeting.InputDirPath(), false), nil
+	// Otherwise we simply ignore any buf.work.yaml that was found and attempt to build
+	// a v1 module at the SubDirPath
+	return newWorkspaceDepManager(bucket, bucketTargeting.SubDirPath(), false), nil
```
saquibmian, line 93: This is because `GetWorkspaceDepManager` only works on v2 workspaces or v1 modules (i.e., not v1 workspaces)?

saquibmian (approved):
Pulled this in core, all tests pass.

private/buf/bufworkspace/workspace_dep_manager_provider.go
doriable: Correct -- yeah, so we gate certain actions (`mod/dep update`, `push`, etc.) from being performed on `v1` workspaces, but `v1` modules are valid. This is done through an option.

https://github.com/bufbuild/buf/blob/c9ad23f6bc66b22f0d5683bea5c92e69e097c7a4/private/buf/bufworkspace/workspace_dep_manager_provider.go#L34-L41

PR #2803, opened 2024-03-06
Allow extra option locations in source info when building an image

jhump:
This is a feature I added to `protocompile` last year because it was easy to do because of the way that `protocompile` parses the option values.

(It can't easily be done in `protoc` because it doesn't actually parse message literals: it scans through the source to the end of the value, looking for a matching `{` that closes the opening `}`, and just stores the tokens in a [string value](https://github.com/protocolbuffers/protobuf/blob/v25.3/src/google/protobuf/descriptor.proto#L921). Later, the value is interpreted by using the Protobuf text format to unmarshal that string of tokens into a message.)

Apparently, I forgot to enable it in `buf` last year. I did this _specifically_ for things like linting custom options (like the protovalidate options), and we can see it pays off a bit in the updated test.

The motivation for me remembering to add this today, so many months later, is so that we can capture more of the comments when rendering custom options in generated docs.


jhump:
cc @gilwong00

bufdev (approved):
Can you run buf build -o tmp.bin on a sufficiently large repo before/after to see how big the diff is in size?

jhump:
> Can you run buf build -o tmp.bin on a sufficiently large repo before/after to see how big the diff is in size?

@bufdev, here's `buf.build/googleapis/googleapis`.
```
-rw-r--r--  1 jhumphries  staff  344112 Mar  6 16:53 after.bin
-rw-r--r--  1 jhumphries  staff  330527 Mar  6 16:52 before.bin
```

That's about 4% difference. I'm happy to run this on any other repos, if you can think of other good examples I should try.

bufdev:
This is fine. Make sure this is merged into `bufmod` when you're done.

jhump:
@mfridman suggested `buf.build/beta/googleapis` and `buf.build/envoyproxy/envoy`, too. Here are the updated numbers:
```
-rw-r--r--  1 jhumphries  staff   3271769 Mar  6 18:10 envoy-after.bin
-rw-r--r--  1 jhumphries  staff   3222085 Mar  6 18:12 envoy-before.bin
-rw-r--r--  1 jhumphries  staff  25166115 Mar  6 18:13 full-googleapis-after.bin
-rw-r--r--  1 jhumphries  staff  24874137 Mar  6 18:13 full-googleapis-before.bin
-rw-r--r--  1 jhumphries  staff    344112 Mar  6 16:53 slim-googleapis-after.bin
-rw-r--r--  1 jhumphries  staff    330527 Mar  6 16:52 slim-googleapis-before.bin
```
So 1.5% growth, 1.1%, and 4.1% respectively. Interestingly, the slimmed-down googleapis actually had the biggest delta. Average of 1.2% increase.

jhump:
> Make sure this is merged into bufmod when you're done.

Aye, aye!

PR #2820, opened 2024-03-18
Port buf to use protoplugin

bufdev:
Description and motivation to come soon.


private/buf/cmd/protoc-gen-buf-lint/lint.go
```diff
@@ -26,48 +26,29 @@ import (
...
-	)
+	protoplugin.Main(protoplugin.HandlerFunc(handle))
 }
 
 func handle(
 	ctx context.Context,
-	container app.EnvStderrContainer,
-	responseWriter protoplugin.ResponseBuilder,
-	request *pluginpb.CodeGeneratorRequest,
+	pluginEnv protoplugin.PluginEnv,
```
mfridman, line 44: I do wonder if it weren't for this refactor, if we would have excluded passing "plugin env" in Handle, i.e.,

```diff
	Handle(
		ctx context.Context,
-		pluginEnv PluginEnv,
		responseWriter ResponseWriter,
		request Request,
	) error
```

Since most implementations would call `os.Environ()` or `os.Stderr` directly, and protoplugin is just proxying these values at this point.

Although on second thought, maybe it's a nice property the caller's function contains all the necessary bits without having to call into `os` env.

mfridman (approved):
nice to have this logic consolidated in one place.

did we need to address the TODO re. source-retention options in the protoc exec?

bufdev:
Yes we do, caught up with @jhump offline and I have a path forward, will do some testing later

bufdev:
Merging so we can test with this today, will follow up

bufdev:
Verified with @jhump that source-retention options are propagated to `protoc` as we expect, we did manual QA (this would be incredibly hard to add an automated test for).

PR #2822, opened 2024-03-18
Dedupe configured dep module refs and keys in migrator

doriable:
For `buf config migrate`, users can have duplicated dependencies
declared across multiple modules. We need to dedupe the module refs
and module keys for these dependencies before we resolve them through
`ModuleKeyProvider` and `CommitProvider`.


private/buf/bufmigrate/migrator.go
```diff
@@ -486,6 +491,10 @@ func (m *migrator) getCommitIDToCommit(
 	ctx context.Context,
 	moduleKeys []bufmodule.ModuleKey,
 ) (map[uuid.UUID]bufmodule.Commit, error) {
+	// We need to dedupe all the module keys of the configured dependencies, since in `v1`
+	// workspaces, each module will have their own configured dependencies, which may be
+	// duplicated across modules.
+	moduleKeys = dedupeStringerSlice(moduleKeys)
```
bufdev, line 497: We're deduping based on previously-seen string value for `ModuleKeys`? We have extensive logic in the `ModuleSetBuilder` relating to what `ModuleKeys` take priority based on timestamp, why is this sufficient here?

private/buf/bufmigrate/migrator.go
```diff
@@ -766,3 +775,17 @@ func filterFileSamePhpGenericServicesMap(m map[string][]string) map[string][]str
 	}
 	return c
 }
+
+// This is a helper function used to deduplicate ModuleKey and ModuleRef, both of
```
bufdev, line 779: If we wanted to do this, this could be accomplished with `slicesext.ToUniqueSorted(slicesext.Map(moduleKeys, bufmodule.ModuleKey.String))`. However, I have doubts that this is what we want.

private/buf/bufmigrate/migrator.go
```diff
@@ -456,6 +457,10 @@ func (m *migrator) getModuleToRefToCommit(
 	ctx context.Context,
 	moduleRefs []bufmodule.ModuleRef,
 ) (map[string]map[string]bufmodule.Commit, error) {
+	// We need to dedupe all the module refs of the configured dependencies, since in `v1`
+	// workspaces, each module will have their own configured dependencies, which may be
+	// duplicated across modules.
+	moduleRefs = dedupeStringerSlice(moduleRefs)
```
bufdev, line 463: What if I have `buf.build/foo/bar:label1` and `buf.build/foo/bar:label2`, what do you want the behavior to be?

doriable: We have logic that follows up this that takes the latest version if there are multiple pins: https://github.com/bufbuild/buf/blob/f4d2a422f9d64cc4820200457fa07fb1d48ce275/private/buf/bufmigrate/migrator.go#L559

The dedupe logic is purely so that we are not resolving the same module key and and module ref multiple times, since this causes issues with our providers.

private/buf/bufmigrate/migrator.go
doriable: See comment below - but basically to build the configured deps list in our `buf.yaml` configuration, we then have additional logic to handle which to choose (re: the `label1` vs `label2` example). This is only to ensure that we do not call the exact same configured reference multiple times using our providers.

private/buf/bufmigrate/migrator.go
doriable: This would give us the slice of strings, we still need `ModuleRef`/`ModuleKey`.

bufdev: Got it - I just added `slicesext.Deduplicate` and `slicesext.DeduplicateAny`, you can do `slicesext.DeduplicateAny(moduleKeys, bufmodule.ModuleKey.String)` now

private/buf/bufmigrate/migrator.go
bufdev: Let's chat about this live - I'm inclined to just merge since it's just migrate, but (1) providers shouldn't be wonky if we provide duplicates (2) want to make sure this logic is clear to others, optimally

doriable: Chatted live, just commenting for posterity: added context around why `bufmodule.ModuleKeyProvider` and `bufmodule.CommitProvider` expects implementations to error for duplicate keys (avoiding unnecessary expensive operations like API calls to resolve these entities).

bufdev (approved):

PR #2827, opened 2024-03-20
Write out string instead of enum value for buf.gen.yaml

doriable:
When writing out `buf.gen.yaml`, we should be writing out the string
representation of enum file/field options instead of the enum value.
When reading `buf.gen.yaml`, we expect to read the string representation,
(references below) so this is a bug when we write out `buf.gen.yaml`.

https://github.com/bufbuild/buf/blob/046fa218e4e1c78d1f561e99bbd7847c7df5d21a/private/bufpkg/bufconfig/generate_managed_option.go#L204-L214

https://github.com/bufbuild/buf/blob/046fa218e4e1c78d1f561e99bbd7847c7df5d21a/private/bufpkg/bufconfig/generate_managed_option.go#L216-L226


private/bufpkg/bufconfig/generate_managed_config.go
```diff
@@ -804,6 +804,13 @@ func newExternalManagedConfigV2FromGenerateManagedConfig(
 		if override.FieldOption() != FieldOptionUnspecified {
 			fieldOptionName = override.FieldOption().String()
 		}
+		if fileOptionName != "" && fieldOptionName != "" {
```
bufdev, line 807: I would move this into the function

bufdev (approved):

PR #2828, opened 2024-03-20
Clean up protoc-gen plugin error message

doriable:
With `v2` `buf.gen.yaml` the plugin name is already parsed with
`protoc-gen-<plugin name>` when this error is returned, so we no
longer need to prefix the plugin name with `protoc-gen` in the
error message.


bufdev (approved):

bufdev:
Overriding github action failure, not actually part of the build.

PR #2829, opened 2024-03-20
Call detrand.Disable once and for all

bufdev:
I've officially had it. The final straw was that prototext was borked and there's no `json.Compact` equivalent for prototext. Making `detrand.Disable` an internal-only function was always a bad idea, and now even Google is working around it in major libraries: https://github.com/google/starlark-go/blob/ee8ed142361c69d52fe8e9fb5e311d2a0a7c02de/lib/proto/proto.go#L774

I'm not linking to the issue on protobuf-go as to not link this issue in, but ask internally if you want to know the issue here.

If `detrand.Disable` is renamed, we will get a build-time error:

```
github.com/bufbuild/buf/private/pkg/protoencoding.init.0: relocation target google.golang.org/protobuf/internal/detrand.DisableMisspelled not defined
```

And if they rename it, it will break Google's own code. I think this is relatively safe long-term, and the worst-case is we blow up at build-time (unless Google maliciously changes the function behavior, but of course they could do this for any function in an OSS library, so I think that is not in-bounds).

### DIFF
--- a/private/pkg/protoencoding/txtpb_marshaler.go
+++ b/private/pkg/protoencoding/txtpb_marshaler.go
@@ -32,9 +32,7 @@ func newTxtpbMarshaler(resolver Resolver) Marshaler {
 func (m *txtpbMarshaler) Marshal(message proto.Message) ([]byte, error) {
 	options := prototext.MarshalOptions{
 		Resolver: m.resolver,
-		// TODO: make this an option
-		Multiline: true,
-		Indent:    "  ",
+		Indent:   "  ",
 	}
 	return options.Marshal(message)
 }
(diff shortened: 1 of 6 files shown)


private/pkg/protoencoding/txtpb_marshaler.go
```diff
@@ -32,9 +32,7 @@ func newTxtpbMarshaler(resolver Resolver) Marshaler {
 func (m *txtpbMarshaler) Marshal(message proto.Message) ([]byte, error) {
 	options := prototext.MarshalOptions{
 		Resolver: m.resolver,
-		// TODO: make this an option
-		Multiline: true,
```
bufdev, line 36: Note that `Indent != ""` implies `Multiline` per the docs, this is just a cleanup.

akshayjshah (approved):

PR #2834, opened 2024-03-21
Fix memory read bucket mapping when no controlling workspace is found

doriable:
In order to treat archive and git refs the same way we treat dir
and proto file refs, we are handling `inputSubDirPath` already
in workspace targeting. This means that we should not actually
be remapping archive and git memory buckets if no controlling
workspace is found.

This was found in our bug bash when testing against an archive
with a `v1` module rooted at the input sub dir path. Added a test
case for this.

### DIFF
--- a/private/buf/buffetch/internal/reader.go
+++ b/private/buf/buffetch/internal/reader.go
@@ -585,26 +585,28 @@ func getReadBucketCloserForBucket(
 	if err != nil {
 		return nil, nil, err
 	}
-	var bucketPath string
-	if bucketTargeting.ControllingWorkspace() != nil {
+	// If a controlling workspace is found, then we map the bucket on the controlling
+	// workspace path. We only need to remap the bucket in the case where a controlling
+	// workspace is found. In the case where no controlling workspace is found, bufworkspace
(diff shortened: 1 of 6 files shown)


private/buf/buffetch/internal/reader.go
```diff
@@ -585,26 +585,28 @@ func getReadBucketCloserForBucket(
...
-	if bucketTargeting.ControllingWorkspace() != nil {
+	// If a controlling workspace is found, then we map the bucket on the controlling
+	// workspace path. We only need to remap the bucket in the case where a controlling
+	// workspace is found. In the case where no controlling workspace is found, bufworkspace
+	// will handle the SubDirPath through workspace targeting given the bucket and BucketTargeting.
+	//
+	// We return the same bucket targeting information, since BucketTargeting
+	// maps the target paths and target exclude paths to the controlling workspace path when
+	// one is found.
+	bucketPath := "."
```
doriable, line 596: `bucketPath` is set here to keep the debug log output below. I think it's reasonable, but it is a bit strange looking.

bufdev (approved):

PR #2837, opened 2024-03-22
Allow disabling lint or breaking for an entire module via a configuration ignore

bufdev:
Up until now, we have made it an error to do the following in your `buf.yaml` configuration:

```yaml
# buf.yaml
version: v1
lint:
  ignore:
    - .
```

This being an error made sense - this is saying that you'd like to ignore lint for all paths in the module. If you didn't want to lint the module, then...don't lint it.

However, this did not take into account workspaces. We have never provided a way for someone to ignore a module within a workspace for linting. Even pre-v2-configuration, this might be something you'd want to do:

```yaml
# buf.work.yaml
version: v1
directories:
- proto/module1
- proto/module2
- vendor
```

You may commonly want to not lint `vendor` while still being able to run `buf lint` for the entire workspace.

In v2, this really comes into the fore. We should provide a way for people to ignore a single module. Instinctively, I would do this:

```yaml
# buf.yaml
version: v2
modules:
  - path: module1
  - path: module2
  - path: vendor
lint:
  ignore:
    - vendor
```

Or I might do this:

```yaml
# buf.yaml
version: v2
modules:
  - path: module1
  - path: module2
  - path: vendor
    lint:
      ignore:
        - vendor
```

However, before this PR, this would result in the following error:

```
Failure: decode buf.yaml: lint.ignore: path "vendor" is equal to module directory "vendor"
```

This PR makes this a valid configuration. If you specify an ignore that is also a module directory, lint (or breaking) checks will be disabled for this entire module.

`ignore` seems like the most logical place to do this. We did consider adding a special `disabled` key, however this feels like it just clutters the configuration. If we want to do this in the future, we can easily add it in in a backwards-compatible manner.

The actual code here isn't the prettiest, but it gets it in. Tests have been added to verify the behavior now matches what we expect.

### DIFF
--- a/private/bufpkg/bufconfig/check_config.go
+++ b/private/bufpkg/bufconfig/check_config.go
@@ -76,7 +76,7 @@ type CheckConfig interface {
 	//     ignore:
 	//       - .
 	//
-	// We could make it so that ignore == moduleDirPath is only allowed for v1, however
+	// We could make it so that ignore == moduleDirPath is only allowed for v2, however
 	// this feels like overkill, so we're just going to keep this consistent for v1
 	// and v2.
 	Disabled() bool
--- a/private/bufpkg/bufconfig/check_config.go
+++ b/private/bufpkg/bufconfig/check_config.go
@@ -19,14 +19,14 @@ import (
 )
 
 var (
-	defaultCheckConfigV1 = newCheckConfigNoValidate(
+	defaultCheckConfigV1 = newEnabledCheckConfigNoValidate(
 		FileVersionV1,
 		nil,
 		nil,
 		nil,
 		nil,
 	)
-	defaultCheckConfigV2 = newCheckConfigNoValidate(
+	defaultCheckConfigV2 = newEnabledCheckConfigNoValidate(
 		FileVersionV2,
 		nil,
 		nil,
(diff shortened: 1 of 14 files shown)


private/bufpkg/bufconfig/check_config.go
```diff
@@ -45,6 +45,41 @@ type CheckConfig interface {
...
+	//
+	// We no longer produce an error in this case. Instead, we set Disabled(), and
+	// do not run checks. This means that the following is no longer an error:
+	//
+	//   version: v1
+	//   lint:
+	//     ignore:
+	//       - .
+	//
+	// We could make it so that ignore == moduleDirPath is only allowed for v2, however
```
pkwarren, line 79:
```suggestion
	// We could make it so that ignore == moduleDirPath is only allowed for v2, however
```
Did you mean v2 here?

pkwarren (approved):

private/bufpkg/bufconfig/check_config.go
bufdev: yea i updated it

doriable (approved):

private/bufpkg/bufconfig/check_config.go
pkwarren: Ah ok was reviewing prior version.

PR #2842, opened 2024-03-25
Update exclude-dirs-use-default setting

pkwarren:
A previous commit in makego had this defined in the wrong place - it needs to be created under 'issues', not 'run'.


bufdev (approved):

bufdev:
This is on main as well - I would fix this on main and then merge into bufmod

PR #2848, opened 2024-03-26
Move buf alpha sdk to buf registry sdk

bufdev:
Newlines also added when printing out the versions.


bufdev:
A few TODOS:

- Discuss changing from `FOO-version` to `version FOO`.
- Add to changelog.

private/buf/bufcli/flags_args.go
```diff
@@ -277,11 +277,11 @@ func PackageVersionLongDescription(registryName, commandName, examplePlugin stri
 Examples:
 
 Get the version of the eliza module and the %s plugin for use with %s.
-    $ buf alpha sdk %s --module=buf.build/connectrpc/eliza --plugin=%s
+    $ buf registry sdk %s --module=buf.build/connectrpc/eliza --plugin=%s
         v1.7.0-20230609151053-e682db0d9918.1
 
 Use a specific module version and plugin version.
-    $ buf alpha sdk %s --module=buf.build/connectrpc/eliza:233fca715f49425581ec0a1b660be886 --plugin=%s:v1.0.0
+    $ buf registry sdk %s --module=buf.build/connectrpc/eliza:233fca715f49425581ec0a1b660be886 --plugin=%s:v1.0.0
```
pkwarren, line 284: If we want the examples to be realistic, we'd want to pass down a plugin version here too as not every plugin has a v1.0.0 release.

private/buf/cmd/buf/buf.go
```diff
@@ -170,6 +170,17 @@ func NewRootCommand(name string) *appcmd.Command {
...
 					registrycc.NewCommand("cc", builder, ``, false),
+					{
+						Use:   "sdk",
+						Short: "Manage Generated SDKs",
+						SubCommands: []*appcmd.Command{
+							goversion.NewCommand("go-version", builder),
+							mavenversion.NewCommand("maven-version", builder),
+							npmversion.NewCommand("npm-version", builder),
+							swiftversion.NewCommand("swift-version", builder),
+							pythonversion.NewCommand("python-version", builder),
```
pkwarren, line 181: Having separate largely identical commands doesn't seem great long term. Ideally to support a new SDK, we'd just allow a new value for `registry` and not have to define a new command.

I checked all the implementations and they are all identical except for the example plugin shown (sometimes with an invalid version as mentioned in previous comment) and the swift CLI takes a `appext.SubCommandBuilder` instead of an `appext.Builder` like all the other ones.

mfridman: To answer your question, I think yes, the versioning scheme depends on the registry type and we want the flexibility for this to change across SDK implementations.

But, as a user, it's quite nice to have a single command because the shape of the response is the same between all of them. They all return a version whose schema may or may not change depending on the plugin's registry type. I agree @pkwarren's proposal to collapse these.

If we add a `--registry-type=npm,go,maven` and make it mandatory then at least it's clear which version output you'll get, although it's a bit redundant since a plugin technically has this information.

mfridman: The server knows whether a plugin is "generated SDK compatible", so for most users it should be sufficient to:

```
buf registry sdk version --plugin=protocolbuffers/go:v1.30.0 --module=bufbuild/buf:latest
```

and get back a version, or an error saying "This plugin is incompatible with generated sdks".

In the JSON payload, we could also return a "registry type" if users wanted to programmatically parse the versioning schema based on the response.

I think simplifying this makes it easier for users to grok and us to maintain?

private/buf/cmd/buf/command/registry/sdk/version/version.go
```diff
@@ -0,0 +1,225 @@
...
+		moduleRef.ModuleFullName().Registry(),
+		registryv1alpha1connect.NewResolveServiceClient,
+	)
+	getLatestCuratedPluginResponse, err := pluginCurationServiceClient.GetLatestCuratedPlugin(
+		ctx,
+		connect.NewRequest(
+			&registryv1alpha1.GetLatestCuratedPluginRequest{
+				Owner:   pluginIdentity.Owner(),
+				Name:    pluginIdentity.Plugin(),
+				Version: pluginVersion,
```
mfridman, line 124: @pkwarren on second thought, I don't think we should support the revision here. Given an explicit version we should always return the latest revision no matter what. WDYT?

mfridman (approved):

private/buf/cmd/buf/command/registry/sdk/version/version.go
pkwarren: In the API or in this command? I looked some more and getting the plugin revision would be more complex if we wanted to support it in this command. We explicitly broke it out into a separate yaml field so that users would get the latest revision.

I don't mind it in the API - the UI has a drop down for revision and may need to display different data based on the revision.

mfridman: Sorry, I meant in this command. I thought you mentioned wanting to support revision, and my only comment was to advise against it specifically in this command.

pkwarren: Agree - I don't think we should support it here.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 - Move `buf mod {clear-cache,cc}` to `buf registry cc`. `buf mod {clear-cache,cc}` is now deprecated.
 - Deprecate `buf mod open`.
 - Delete `buf beta migrate-v1beta1`.
+- Promote `buf alpha sdk <go-version|maven-version|npm-version|swift-version|python-version>` to `buf registry sdk version`.
```
bufdev, line 17: Just say "Add `buf registry sdk version` to get the version of a Generated SDK for a module and plugin." or the like. As far as users are concerned, `alpha` doesn't exist, and we don't support it, so I wouldn't think of this as a promotion.

PR #2849, opened 2024-03-27
Delete PLUGIN_REGISTRY_TYPE_ARCHIVE

bufdev:
All plugins can be used with archives.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -30,13 +30,15 @@ enum CuratedPluginVisibility {
...
+  reserved 6;
+  reserved "PLUGIN_REGISTRY_TYPE_ARCHIVE";
+
   PLUGIN_REGISTRY_TYPE_UNSPECIFIED = 0;
   PLUGIN_REGISTRY_TYPE_GO = 1;
   PLUGIN_REGISTRY_TYPE_NPM = 2;
   PLUGIN_REGISTRY_TYPE_MAVEN = 3;
   PLUGIN_REGISTRY_TYPE_SWIFT = 4;
   PLUGIN_REGISTRY_TYPE_PYTHON = 5;
-  PLUGIN_REGISTRY_TYPE_ARCHIVE = 6;
```
mfridman, line 39: The _only_ time we would need this is if we want a default plugin option for a given plugin against the archive registry. Otherwise some plugins may not work out-of-the-box.

If we don't care about this property, then this makes sense.

bufdev: Not sure what you mean here - I think this enum is denoting what package ecosystem something is a part of, is there something else it should represent?

mfridman: Ah, you only removed the enum, not the registry config.

bufdev: I also removed that, just now, but still not sure what you mean

mfridman (approved):

PR #2850, opened 2024-03-28
Add 'commit' as an alias for 'tag' param in inputs that refer to git repos

jhump:
This is intended to fix an issue that is commonly encountered when users try to pin an input to a particular commit of a git repo. The most intuitive parameter to use is `ref`, with a full commit hash. But the way that actually works is to clone the given `branch` (or repo's default branch if no `branch` param specified), only up to 50 commits deep (can be overridden/increased with `depth` param). This is a problem because, as more commits are added to the repo, the `ref` eventually falls out of the 50-commit depth of the branch's history.

It turns out, that one can use a full commit hash with the `tag` param instead. This pins to a commit perfectly. But `tag` is not an intuitive name for a parameter that accepts a commit hash.

So this PR simply allows users to use `commit` as a parameter, and it means the same thing as `tag`. Unlike `ref`, it **must** be a full hash (no short hashes allowed).

After this is merged and released, I will update the documentation to add `commit` and describe it as for use with full commit hashes or tag names (and to also mention that `tag` is an alias for this parameter, for clarity when using tag names as well as for backwards compatibility).

### DIFF
--- a/private/buf/buffetch/internal/internal.go
+++ b/private/buf/buffetch/internal/internal.go
@@ -522,35 +522,41 @@ func NewProtoFileWriter(
 // The Path will be the only value set when the RawRefProcessor is invoked, and is not normalized.
 // After the RawRefProcessor is called, options will be parsed.
 type RawRef struct {
-	// Will always be set
-	// Not normalized yet
+	// Will always be set.
+	// Not normalized yet.
 	Path string
-	// Will always be set
-	// Set via RawRefProcessor if not explicitly set
+	// Will always be set.
+	// Set via RawRefProcessor if not explicitly set.
 	Format string
-	// Only set for single, archive formats
-	// Cannot be set for zip archives
+	// Only set for single, archive formats.
+	// Cannot be set for zip archives.
 	CompressionType CompressionType
-	// Only set for archive, git formats
+	// Only set for archive, git formats.
 	SubDirPath string
-	// Only set for git formats
-	// Only one of GitBranch and GitTag will be set
+	// Only set for git formats.
+	// Only one of GitBranch and GitCommitOrTag will be set.
 	GitBranch string
+	// Only set for git formats.
+	// Only one of GitBranch and GitCommitOrTag will be set.
+	// Should indicate a full commit hash or tag name.
+	// This is defined as anything that can be given to "git fetch".
+	GitCommitOrTag string
 	// Only set for git formats
-	// Only one of GitBranch and GitTag will be set
-	GitTag string
-	// Only set for git formats
-	// Specifies an exact git reference to use with git checkout.
-	// Can be used on its own or with GitBranch. Not allowed with GitTag.
-	// This is defined as anything that can be given to git checkout.
+	// Specifies a git reference to use with "git checkout".
+	// Can be used on its own or with GitBranch. Not allowed with GitCommitOrTag.
+	// This is defined as anything that can be given to "git checkout".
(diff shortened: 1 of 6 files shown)


jhump:
@bufdev, this was something we chatted about briefly a few months ago. cc @timostamm.

private/buf/buffetch/internal/internal.go
```diff
@@ -499,35 +499,41 @@ func NewWriter(
...
-	GitTag string
-	// Only set for git formats
-	// Specifies an exact git reference to use with git checkout.
-	// Can be used on its own or with GitBranch. Not allowed with GitTag.
-	// This is defined as anything that can be given to git checkout.
+	// Only set for git formats.
+	// Only one of GitBranch and GitCommit will be set.
+	// Should indicate a full commit hash or tag name.
+	// This is defined as anything that can be given to "git fetch".
+	GitCommit string
```
bufdev, line 520: This rename seems to just go in the other direction - now, in the code I would be confused as to why a tag can be used here. I would suggest naming this `GitCommitOrTag`, and updating the functions in this PR appropriately per that naming

bufdev:
> After this is merged and released, I will update the documentation to add commit and describe it as for use with full commit hashes or tag names (and to also mention that tag is an alias for this parameter, for clarity when using tag names as well as for backwards compatibility).

I would just document them separately, ie `commit` is for commits, and `tag` is for tags - that they do the same thing is more of an implementation detail. Telling users in the docs that `commit` works for commits or tags seems just as confusing to me as saying `tag` works with commits or tags, so if we're adding this field, I wouldn't describe it as an alias in the docs.

private/buf/buffetch/internal/internal.go
jhump: Renamed

bufdev (approved):

bufdev (changes requested):
Can you do this against bufmod and not main? There will probably be merge conflicts otherwise

bufdev:
Otherwise looks good/ready to go

jhump:
> Can you do this against bufmod and not main? There will probably be merge conflicts otherwise

@bufdev, done. This widened the impact a bit since bufmod now includes yaml configs for the same sort of thing. Let me know if the way I did it looks like how you want it. I basically added "commit" as a new field in the YAML file, too.

private/bufpkg/bufconfig/input_config.go
```diff
@@ -182,7 +184,7 @@ func NewGitRepoInputConfig(
 		location:          location,
 		subDir:            subDir,
 		branch:            branch,
-		tag:               tag,
+		commitOrTag:       tag,
```
bufdev, line 187: I think the parameter should be renamed to `commitOrTag` in this case

private/bufpkg/bufconfig/input_config.go
```diff
@@ -142,11 +144,11 @@ type InputConfig interface {
 	SubDir() string
 	// Branch returns the git branch to checkout out, not empty only if format is git.
 	Branch() string
-	// Tag returns the git tag to checkout, not empty only if format is git.
-	Tag() string
+	// CommitOrTag returns the git tag to checkout, not empty only if format is git.
```
bufdev, line 147: Docs need to be slightly updated here

private/bufpkg/bufconfig/input_config.go
jhump: done

bufdev (approved):

PR #2858, opened 2024-04-01
Add bufmodule.DigestMismatchError with instructions on how to clear the cache

bufdev:


private/bufpkg/bufmodule/errors.go
```diff
@@ -209,3 +212,47 @@ func (n *NoProtoFilesError) Error() string {
...
+		return ""
+	}
+	var builder strings.Builder
+	_, _ = builder.WriteString(`*** Digest verification failed`)
+	if m.ModuleFullName != nil {
+		_, _ = builder.WriteString(` for "`)
+		_, _ = builder.WriteString(m.ModuleFullName.String())
+		if !m.CommitID.IsNil() {
+			_, _ = builder.WriteString(`:`)
+			_, _ = builder.WriteString(uuidutil.ToDashless(m.CommitID))
```
bufdev, line 237: Just doing dashless so we don't have to come back to this re: https://github.com/bufbuild/buf/pull/2854

private/bufpkg/bufmodule/errors.go
```diff
@@ -209,3 +212,47 @@ func (n *NoProtoFilesError) Error() string {
...
+		_, _ = builder.WriteString(`"`)
+		_, _ = builder.WriteString("\n")
+		_, _ = builder.WriteString("\t")
+		_, _ = builder.WriteString(`Actual digest: "`)
+		_, _ = builder.WriteString(m.ActualDigest.String())
+		_, _ = builder.WriteString(`"`)
+		_, _ = builder.WriteString("\n")
+	}
+	_, _ = builder.WriteString("\t")
+	_, _ = builder.WriteString(`This may either be a result of an attack, or a result of your local cache being corrupted. To clear your local cache, run "buf registry cc".`)
```
pkwarren, line 256: Of the two it is more likely to be the second also this could be the result of a hand edited buf.lock file.

We may also want to consider at some point allowing to clear the cache for a single module or commit instead of wiping everything.

private/bufpkg/bufmodule/errors.go
```diff
@@ -209,3 +212,47 @@ func (n *NoProtoFilesError) Error() string {
...
+			_, _ = builder.WriteString(`:`)
+			_, _ = builder.WriteString(uuidutil.ToDashless(m.CommitID))
+		}
+		_, _ = builder.WriteString(`"`)
+	}
+	_, _ = builder.WriteString(` ***`)
+	_, _ = builder.WriteString("\n")
+	if m.ExpectedDigest != nil && m.ActualDigest != nil {
+		_, _ = builder.WriteString("\t")
+		_, _ = builder.WriteString(`Expected digest: "`)
```
pkwarren, line 245:
```suggestion
		_, _ = builder.WriteString(`Expected digest (from buf.lock): "`)
```

private/bufpkg/bufmodule/errors.go
```diff
@@ -209,3 +212,47 @@ func (n *NoProtoFilesError) Error() string {
...
+		_, _ = builder.WriteString(m.ExpectedDigest.String())
+		_, _ = builder.WriteString(`"`)
+		_, _ = builder.WriteString("\n")
+		_, _ = builder.WriteString("\t")
+		_, _ = builder.WriteString(`Actual digest: "`)
```
pkwarren, line 250: Do we have information to differentiate if the actual digest is from the module cache or the BSR here?

If we fetch the digest from the BSR we could perhaps provide a better error message here to the user to choose a better message ('clear cache' vs. 'attack').

pkwarren (approved):
Looks good - a couple suggestions on the underlying error message.

private/bufpkg/bufmodule/errors.go
bufdev: We don't at the moment, we can do this later if we need to

private/bufpkg/bufmodule/errors.go
bufdev: updated

PR #2859, opened 2024-04-01
Omit build roots in v1 buf.yaml serialization

pkwarren:
Update buf.yaml serialization to omit the 'build' -> 'roots' section of for v1. This is only valid for v1beta1 config files.

Add a test case to verify the fix (previous failed to round trip v1 config files with excludes).

### DIFF
--- a/private/bufpkg/bufconfig/buf_yaml_file.go
+++ b/private/bufpkg/bufconfig/buf_yaml_file.go
@@ -496,7 +496,7 @@ func writeBufYAMLFile(writer io.Writer, bufYAMLFile BufYAMLFile) error {
 		externalBufYAMLFile := externalBufYAMLFileV1Beta1V1{
 			Version: fileVersion.String(),
 		}
-		// Alredy sorted.
+		// Already sorted.
 		externalBufYAMLFile.Deps = slicesext.Map(
 			bufYAMLFile.ConfiguredDepModuleRefs(),
 			func(moduleRef bufmodule.ModuleRef) string {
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -512,10 +512,13 @@ func writeBufYAMLFile(writer io.Writer, bufYAMLFile BufYAMLFile) error {
 		if len(rootToExcludes) != 1 || !(ok && len(excludes) == 0) {
 			roots := slicesext.MapKeysToSortedSlice(rootToExcludes)
 			for _, root := range roots {
-				externalBufYAMLFile.Build.Roots = append(
-					externalBufYAMLFile.Build.Roots,
-					root,
-				)
+				// Roots only valid for v1beta1
+				if fileVersion == FileVersionV1Beta1 {
```
bufdev, line 516: This isn't the solution to the issue - this is just silently ignoring `rootToExcludes` if the `FileVersion` happens to not be `FileVersionV1Beta1`. The definition for `RootToExcludes` specifies that if there is a `v1` or `v2` file, there will be a single key `"."` with potentially some excludes. If `len(rootToExcludes) != 1 || !ok` is true, then this shouldn't be a `v1` or `v2` file - something else is borked.

Granted, there isn't the best error handling here, but this is silently ignoring the issue. A solution to the problem will:

- Return `syserror.Newf` if an expectation has failed - there's obviously a bug in here somewhere, it needs to be investigated, and if we have misbehaving code, we'll return a system error to alert on it and produce an error message to file an issue.
- `switch` on all possible `FileVersions`, similar to other parts of the codebase - we rarely do `if fileVersion ==` to make sure we handle all possible `FileVersions`.

bufdev (changes requested):

private/bufpkg/bufconfig/buf_yaml_file.go
pkwarren: Updated to take some error handling from the V2 config file path below, and although unreachable added validation to the loop over roots to ensure we never get an unexpected root for v1 files.

bufdev: This is getting even more complicated and difficult to understand - what you need to do:

switch `fileVersion`:

case `v1`:

- Validate that `RootToExcludes()` is of the shape you expect, namely of size 1, with a single key `.`
- Output excludes if excludes exist on key `.`.

case `v1beta1`

- Output roots and excludes

default:

syserror

The logic here is getting more confusing - on first reading, it's still not clear to me how the `if fileVersion == FileVersionV1Beta1` is required - perhaps there is a reason, but it still appears to be silently ignoring an issue (even if, by virtue of the if/elses, it isnt).

pkwarren: PTAL - I've updated to a switch statement to be more explicit with the differences between the branches.

PR #2860, opened 2024-04-02
Fix generate v2 disable rules exclusions

emcfarlane:
Fix for a disable rule targeting a `field_option` being applied to a `file_option` and vice versa. For example the following `buf.gen.yaml` file:
```yaml
version: v2
managed:
  enabled: true
  disable:
    - module: buf.build/emcfarlane/foo
      field_option: jstype
  override:
    - file_option: go_package_prefix
      module: buf.build/emcfarlane/foo
      value: github.com/emcfarlane/foo
plugins:
  - local: protoc-gen-go
    out: gen
    opt: 
      - paths=source_relative
inputs:
  - module: buf.build/emcfarlane/foo
```
The disable rule should not affect the `file_option: go_package_prefix`.

### DIFF
--- a/private/bufpkg/bufimage/bufimagemodify/field_option.go
+++ b/private/bufpkg/bufimage/bufimagemodify/field_option.go
@@ -54,8 +54,10 @@ func modifyJsType(
 		return nil
 	}
 	// If the entire file is disabled, skip.
-	if isFieldOptionDisabledForFile(imageFile, bufconfig.FieldOptionJSType, config, "") {
-		return nil
+	for _, disableRule := range config.Disables() {
+		if isFieldOptionDisabledForFile(disableRule, imageFile, bufconfig.FieldOptionJSType) {
+			return nil
+		}
 	}
 	if datawkt.Exists(imageFile.Path()) {
 		return nil
@@ -72,8 +74,10 @@ func modifyJsType(
 				return nil
 			}
 			// If the field is disabled, skip.
-			if isFieldOptionDisabledForFile(imageFile, bufconfig.FieldOptionJSType, config, fullName) {
-				return nil
+			for _, disableRule := range config.Disables() {
+				if isFieldOptionDisabledForField(disableRule, imageFile, bufconfig.FieldOptionJSType, fullName) {
+					return nil
+				}
 			}
 			var jsType *descriptorpb.FieldOptions_JSType
 			for _, override := range config.Overrides() {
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufimage/bufimagemodify/field_option.go
```diff
@@ -132,3 +124,27 @@ func isJsTypePermittedForType(fieldType descriptorpb.FieldDescriptorProto_Type)
 		fieldType == descriptorpb.FieldDescriptorProto_TYPE_FIXED64 ||
 		fieldType == descriptorpb.FieldDescriptorProto_TYPE_SFIXED64
 }
+
+func isFieldOptionDisabledForFile(
```
doriable, line 128: I think it's a bit misleading to name this `isFieldOptionDisabledForFile`, for two reasons:

- you are not checking a single field option, you are checking the all disable rules in the config
- you are also not only checking it for the file, you are also checking the specified field

I would probably put the checks in a helper and then use them in the loops above, something like:

```go

func isFieldOptionDisabledForFile(
  disableRule bufconfig.ManagedDisableRule,
  imageFile bufimage.ImageFile,
  fieldOption bufconfig.FieldOption,
)  bool {
  if disableRule.FileOption() != bufconfig.FileOptionUnspecified {
    return false // FileOption specified, not a matching rule.
	}
  if disableRule.FieldOption() != bufconfig.FieldOptionUnspecified && disableRule.FieldOption() != fieldOption {
	return false // FieldOption specified, does not match option.
  }
  return !fileMatchConfig(imageFile, disableRule.Path(), disableRule.ModuleFullName())
}

func isFieldOptionDisabledForField(
  disableRule bufconfig.ManagedDisableRule,
  imageFile bufimage.ImageFile,
  fieldOption bufconfig.FieldOption,
  fullName protoreflect.FullName,
) bool {
  // Check if the field option is disabled for the full file first, if so, we can return here.
  if isFieldOptionDisabledForFile(disableRule, imageFile, fieldOption) {
    return true
  }
  return disableRule.FieldName() != "" && disableRule.FieldName() != string(fullName)
}
```

And then I would restore the `for` loops above to look something like:

```go
for _, disableRule := range config.Disables() {
  if isFieldOptionDisabledForFile(disableRule, imageFile, fieldOption) {
    return nil
  }
}

for _, disable := range config.Disables() {
  if isFieldOptionDisabledForField(disableRule, imageFile, fieldOption, fullName) {
    return nil
  }
}
```

bufdev (approved):

private/bufpkg/bufimage/bufimagemodify/field_option.go
emcfarlane: Fixed. The `ForField` one must have the FieldName exclusion check before the other cases.

doriable (approved):
Nice, lgtm and super quick turnaround time on the fix! 🎉

PR #2869, opened 2024-04-04
Provide mechanism to start implementing and testing editions-specific functionality

jhump:
This allows us to start implementing and testing logic for editions (which is mostly confined to protosource and bufbreaking), before actually having the CLI provide editions support to users. The idea is that users won't be able to use editions with the CLI until everything is updated to work correctly, but we need a way to enable the support until then, for testing any in-progress editions-specific work.

We have to use "shenanigans" with `go:linkname` because support isn't yet complete and enabled in protocompile, but it's complete "enough" for what we need to do in the CLI (or at least it will be after https://github.com/bufbuild/protocompile/pull/264 lands, and a new release is cut). Unfortunately, protocompile won't be "done-done" with editions support until we can get protobuf-go's `descriptorpb` package updated with the [latest editions-related changes](https://github.com/protocolbuffers/protobuf/commit/b09b3e46fe2a111d412251f6d043649334e9cbee#diff-ccd6d1c124956da1d1ef88a291957bd98abf1b0e7472f7e5153f7b70ef20efb2), and that's not likely to happen until we get a release candidate for v27.0 of protoc 😞.

Happy to hold off until cutting that next release of protocompile, but I wanted to open this up for discussion as I suspect the technique here (which some might call "crimes") may be controversial 😂.

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/testdata/options/a.proto
+++ b/private/bufpkg/bufimage/bufimageutil/testdata/options/a.proto
@@ -9,7 +9,7 @@ message Foo {
 	option (message_foo).foo = "str";
 	option (message_baz) = "str";
 
-	optional string foo = 1 [
+	optional uint64 foo = 1 [
 		(field_foo).foo = "str",
 		(field_baz) = "str",
 		jstype = JS_STRING
--- a/private/bufpkg/bufimage/build_image.go
+++ b/private/bufpkg/bufimage/build_image.go
@@ -18,7 +18,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	_ "unsafe" // to support go:linkname below
 
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
 	"github.com/bufbuild/buf/private/bufpkg/bufmodule"
@@ -27,6 +26,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/thread"
 	"github.com/bufbuild/buf/private/pkg/tracing"
 	"github.com/bufbuild/protocompile"
+	"github.com/bufbuild/protocompile/editionstesting"
 	"github.com/bufbuild/protocompile/linker"
 	"github.com/bufbuild/protocompile/parser"
 	"github.com/bufbuild/protocompile/protoutil"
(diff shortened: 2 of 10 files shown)


private/bufpkg/bufimage/build_image.go
```diff
@@ -33,6 +34,14 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
+// This enables editions support in protocompile. Buf doesn't yet support editions, but rejecting
+// such files happens elsewhere (in github.com/bufbuilder/buf/private/pkg/protodescriptor). We
+// need the compiler's support enabled so that we can start implementing and testing editions
+// support in the CLI, before we enable that support for users.
+//
+//go:linkname allowEditionsInCompiler github.com/bufbuild/protocompile/internal.AllowEditions
+var allowEditionsInCompiler = true
```
bufdev, line 43: Can we just have a `protocompile/editions.AllowEditions()` global function that we can call below? protocompile is pre-v1.0 so we can remove this later

jhump: Eh, I suppose 🤷. TBH, I was really hoping to avoid anything in protocompile's exported API about editions until that repo's support is also complete. This seemed like the nicest way to preserve that property.

timostamm: Having run into a similar situation, a public opt-in doesn't seem great. Once editions have landed, the flag needs to be kept forever, but doesn't serve a purpose anymore.

jhump: > the flag needs to be kept forever

This isn't necessarily an issue. As Peter mentioned, protocompile hasn't yet reached v1.0, so we can change the API and remove elements if need be (and I can document it as temporary, too).

I'm trying to think of what we'd name it to make it clear that users should generally _not_ use it as its incomplete. `AllowEditions` seems too innocent... Maybe `AllowEditionsThoughIncomplete`? Or perhaps it just live in a `testing` package, so it's more clear that it's for testing until the implementation is complete?

jhump: @bufdev, @timostamm, how about this? https://github.com/bufbuild/protocompile/pull/267

private/bufpkg/bufimage/bufimageutil/testdata/options/a.proto
```diff
@@ -9,7 +9,7 @@ message Foo {
 	option (message_foo).foo = "str";
 	option (message_baz) = "str";
 
-	optional string foo = 1 [
+	optional uint64 foo = 1 [
```
jhump, line 12: This was not actually valid. `protoc` would complain with the following:
```
a.proto:12:18: jstype is only allowed on int64, uint64, sint64, fixed64 or sfixed64 fields.
```

The latest protocompile adds this check for consistency, which in turn required updating this file for the tests in here to pass.

private/bufpkg/bufimage/build_image.go
jhump: Updated this PR to use new opt-in function.

PR #2871, opened 2024-04-04
Handle errors from new buf versions to old BSR versions

stefanvanburen:
This adds a better error message / suggestions for remedying a situation where it's suspected that a newer version of the CLI is interacting with an older version of the BSR.

The set of cases to cover is:

1. Explicit code unimplemented - it's overwhelmingly likely that a newer CLI made a request to an unimplemented API in this case
2. Code Unknown + "Method Not Allowed" error message - this is an implementation detail of the BSR, but it's how older versions of the BSR will react to requests to new services
3. Code Unimplemented + http "Not Found", for new RPCs on known services. This is a result of the [default connect-go generated handlers][1].

In newer versions of the BSR, we're going to look to change the behavior resulting in 2, which will result in a separate case to handle here.

Here's a few examples of the output, running against an old version of the BSR (specifically, v1.3.0).

Case 2:

```console
$ buf push
Failure: unknown: 405 Method Not Allowed for https://bufbuild.internal/buf.registry.module.v1.ModuleService/GetModules
This version of the buf CLI may require APIs that have not yet been deployed to https://bufbuild.internal
To resolve this failure, you can either:
- Try using an older version of the buf CLI
- Contact the site admin for https://bufbuild.internal to upgrade the instance
```

Case 1:

```console
$ buf dep update
Failure: unimplemented: 404 Not Found for https://bufbuild.internal/buf.registry.module.v1beta1.CommitService/GetCommits
This version of the buf CLI may require APIs that have not yet been deployed to https://bufbuild.internal
To resolve this failure, you can either:
- Try using an older version of the buf CLI
- Contact the site admin for https://bufbuild.internal to upgrade the instance
```

[1]: https://github.com/connectrpc/connect-go/blob/main/cmd/protoc-gen-connect-go/main.go

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,7 +29,6 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
-	go.akshayshah.org/memhttp v0.1.0
 	go.opentelemetry.io/otel v1.24.0
 	go.opentelemetry.io/otel/sdk v1.24.0
 	go.opentelemetry.io/otel/trace v1.24.0
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
+	go.akshayshah.org/memhttp v0.1.0
 	go.opentelemetry.io/otel v1.24.0
 	go.opentelemetry.io/otel/sdk v1.24.0
 	go.opentelemetry.io/otel/trace v1.24.0
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -398,6 +398,8 @@ func isPossibleNewCLIOldBSRError(connectErr *connect.Error) bool {
 	case connect.CodeUnknown:
 		// Older versions of the BSR return errors of this shape
 		// for unrecognized services.
+		// NOTE: This handling can be removed once all BSR instances
+		// are upgraded past v1.7.0.
 		return connectErr.Message() == fmt.Sprintf("%d %s", http.StatusMethodNotAllowed, http.StatusText(http.StatusMethodNotAllowed))
 	case connect.CodeUnimplemented:
 		// RPC was known, but unimplemented in the BSR version.
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -20,6 +20,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"net/http"
 	"time"
 
 	"connectrpc.com/connect"
@@ -304,6 +305,19 @@ func wrapError(err error) error {
 		return err
 	}
 	if isConnectError {
+		var augmentedConnectError *bufconnect.AugmentedConnectError
+		isAugmentedConnectErr := errors.As(err, &augmentedConnectError)
+		if isPossibleNewCLIOldBSRError(connectErr) && isAugmentedConnectErr {
+			return fmt.Errorf("Failure: %[1]s for https://%[2]s%[3]s\n"+
+				"This version of the buf CLI may require APIs that have not yet been deployed to https://%[2]s\n"+
+				"To resolve this failure, you can either:\n"+
+				"- Try using an older version of the buf CLI\n"+
+				"- Contact the site admin for https://%[2]s to upgrade the instance",
+				connectErr,
+				augmentedConnectError.Addr(),
+				augmentedConnectError.Procedure(),
+			)
+		}
 		connectCode := connectErr.Code()
 		switch {
 		case connectCode == connect.CodeUnauthenticated, isEmptyUnknownError(err):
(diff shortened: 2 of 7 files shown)


private/bufpkg/bufconnect/interceptors_test.go
```diff
@@ -147,3 +151,21 @@ func TestCLIWarningInterceptorFromError(t *testing.T) {
...
+		server.Client(),
+		server.URL(),
+		connect.WithInterceptors(NewAugmentedConnectErrorInterceptor()),
+	)
+	_, err := client.GetCommits(context.Background(), connect.NewRequest(&modulev1.GetCommitsRequest{}))
+	assert.Error(t, err)
+	var augmentedConnectError *AugmentedConnectError
+	assert.ErrorAs(t, err, &augmentedConnectError)
+	assert.Equal(t, "example.com", augmentedConnectError.Addr())
+	assert.Equal(t, "/buf.registry.module.v1.CommitService/GetCommits", augmentedConnectError.Procedure())
```
stefanvanburen, line 170: I _think_ I need a "real" server/client to populate the addr and procedure here; not sure if there's a way to do this better.

nicksnyder:
Looks good to me in terms of solving the problem I wanted us to solve. Will leave the code review to others.

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/spf13/pflag v1.0.5
 	github.com/stretchr/testify v1.9.0
+	go.akshayshah.org/memhttp v0.1.0
```
bufdev, line 32: We have a version of this in connect somewhere - we should use that instead, although I'm surprised we need something like this to test anyways

emcfarlane: If we are happy with the workaround jhump suggests in that issue I think this can be done as a unit test:
```go
type testRequest[T any] struct {
	*connect.Request[T]
}

func (r testRequest[_]) Spec() connect.Spec {
	return connect.Spec{
		Procedure: "/service/method",
		Addr: "...",
	}
}

// pass me to the interceptor as `next` arg.
func nextHandler(ctx context.Context, req conntext.AnyRequest) (conntext.AnyResponse, error) {
	return nil, connect.NewError(...)
}
```

private/bufpkg/bufconnect/errors.go
```diff
@@ -47,3 +47,31 @@ func AsAuthError(err error) (*AuthError, bool) {
 	ok := errors.As(err, &authErr)
 	return authErr, ok
 }
+
+// AugmentedConnectError wraps an error, providing the Procedure and Addr of
+// the request for introspection in callers.
+type AugmentedConnectError struct {
+	cause     error
```
emcfarlane, line 54: Could be left as a *connect.Error.

private/buf/cmd/buf/buf.go
```diff
@@ -304,6 +305,19 @@ func wrapError(err error) error {
 		return err
 	}
 	if isConnectError {
+		var augmentedConnectError *bufconnect.AugmentedConnectError
+		isAugmentedConnectErr := errors.As(err, &augmentedConnectError)
+		if isPossibleNewCLIOldBSRError(connectErr) && isAugmentedConnectErr {
```
emcfarlane, line 310: What about leaving this under the switch for `connect.CodeUnknown` below? It wasn't clear that `isPossibleNewCLIOldBSRError` mutates the message to me at first.

private/bufpkg/bufconnect/errors.go
stefanvanburen: I [wanted to use the original error](https://github.com/bufbuild/buf/blob/addef8cd746cd70bb9540adbd57018eb18021009/private/bufpkg/bufconnect/interceptors.go#L39-L42), which I suppose might not necessarily _immediately_ be a *connect.Error, in the case that an interceptor further down the chain replaced it with a newly wrapped error. Not sure if that's overthinking it, though.

private/buf/cmd/buf/buf.go
stefanvanburen: Doesn't mutate the error (unless I'm missing it); it's just introspecting for a particular message that the BSR returns. Felt like this was better handled in one spot so it was clear what we were looking for and we could put the error message there.

I think in the future, when we can drop the `connect.CodeUnknown` check, we could move this to be included in the switch, since it'll only have to check for `CodeUnimplemented`?

go.mod
stefanvanburen: Replaced with the workaround in https://github.com/bufbuild/buf/pull/2871/commits/e6882be9d0abe78ef3e876aba181ec1eccb3ebbe, thanks for the suggestion!

private/bufpkg/bufconnect/errors.go
```diff
@@ -47,3 +47,31 @@ func AsAuthError(err error) (*AuthError, bool) {
...
+	procedure string
+	addr      string
+}
+
+// Error implements the error interface and returns the error message.
+func (e *AugmentedConnectError) Error() string {
+	return e.cause.Error()
```
bufdev, line 61: Doesn't handle the nil case like `AuthError` does above, resulting in a potential panic.

private/buf/cmd/buf/buf.go
```diff
@@ -304,6 +305,19 @@ func wrapError(err error) error {
 		return err
 	}
 	if isConnectError {
+		var augmentedConnectError *bufconnect.AugmentedConnectError
+		isAugmentedConnectErr := errors.As(err, &augmentedConnectError)
```
bufdev, line 309: When will this ever be false in this context, if ever?

private/bufpkg/bufconnect/errors.go
stefanvanburen: https://github.com/bufbuild/buf/pull/2871/commits/6af6dae9ddd9da097671c41b314adfa116b0cbfc

private/buf/cmd/buf/buf.go
stefanvanburen: Don't know the `buf` code well enough to be sure, but if we were returning a `connect.Error` from anywhere that _wasn't_ wrapped in the interceptor (probably not? — but will that always be true?).

Just trying to be cautious, since this is more of a best-effort hint to users that they may be in this situation, rather than an iron-clad "we know you are definitely in this situation."

bufdev (approved):

PR #2882, opened 2024-04-08
Ensure default label is pushed when using `--tag`

doriable:
This PR implements the following:

- Validation that there are no overlaps in `--label`, `--tag`, `--branch`, and `--draft` flags being used with `buf push`
- We only allow setting `--tag`, `--branch`, and `--draft` flags when pushing a single module at a time, otherwise we raise an error to the user
- If labels are set, then we push to those labels
- If branch/draft is set, then we only push to a single label with the branch/draft name
- If tags are set, then we push to labels with tag names set AND the default label name. We used newly created modules and/or validated module content to get the default label name. This preserves the current behavior of the CLI, where we push to the default namespace when pushing tags.
- Fixes a bug where we were calling `CreateModules` on all `contentModules` at once. If one or more of the `contentModules` already exist, then no new modules will be created, since `CreateModules` is atomic. So we must create each module individually.


private/buf/cmd/buf/command/push/push.go
```diff
@@ -185,7 +186,7 @@ func run(
 	case 0:
 		return nil
 	case 1:
-		_, err := container.Stdout().Write([]byte(uuidutil.ToDashless(commits[0].ModuleKey().CommitID()) + "\n"))
+		_, err := container.Stdout().Write([]byte(commits[0].ModuleKey().CommitID().String() + "\n"))
```
unmultimedio, line 189: I think this will print the commit in dashful format, and what we want is print them all in dashless.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -267,6 +268,40 @@ func validateCreateFlags(flags *flags) error {
...
+	}
+	return validateLabelFlagValues(flags)
+}
+
+// We need to validate the combination of flags that set labels:
+//   - do not allow mixing of `--label` and `--tag` flags
+//   - do not allow mixing of `--label` and `--branch`/`--draft` flags
+//   - do not allow mixing of `--tag` and `--branch`/`--draft` flags
+//   - do not allow mixing of `--draft` and `--branch` flags
+func validateLabelFlagCombinations(flags *flags) error {
```
unmultimedio, line 282: Nit: because there are multiple combinations here, what about this approach?

```go
usedFlags := make(map[string]struct{})
if len(flags.Labels) > 0 {
  usedFlags[labelFlagName] = struct{}
}
if len(flags.Tags) > 0 {
  usedFlags[tagFlagName] = struct{}
}
if flags.Branch != "" {
  usedFlags[branchFlagName] = struct{}
}
if flags.Draft != "" {
  usedFlags[draftFlagName] = struct{}
}
if len(usedFlags) > 1 {
  // compose an error message with all the flags that are used
}
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -280,13 +315,21 @@ func validateLabelFlags(flags *flags) error {
...
-		labels = append(labels, flags.Branch)
+		return bufmodule.UploadWithBranchOrDraft(flags.Branch)
+	}
+	if len(flags.Tags) > 0 {
+		return bufmodule.UploadWithTags(slicesext.ToUniqueSorted(flags.Tags)...)
 	}
-	return slicesext.ToUniqueSorted(labels)
+	if len(flags.Labels) > 0 {
+		return bufmodule.UploadWithLabels(slicesext.ToUniqueSorted(flags.Labels)...)
+	}
```
unmultimedio, line 333: Nit: let's have the most expected case first? labels, tags, branch, draft

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -267,48 +325,51 @@ func (a *uploader) Upload(
...
+				},
 			},
 		),
-	); err != nil && connect.CodeOf(err) != connect.CodeAlreadyExists {
-		return err
+	)
+	if err != nil && connect.CodeOf(err) != connect.CodeAlreadyExists {
+		return nil, err
+	}
+	// If a module already existed, then we
```
unmultimedio, line 359: Incomplete comment

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -267,48 +325,51 @@ func (a *uploader) Upload(
...
+	if connect.CodeOf(err) == connect.CodeAlreadyExists {
+		return a.validateContentModulesExist(ctx, primaryRegistry, []bufmodule.Module{contentModule})
 	}
```
unmultimedio, line 362: L360 is assuming `err` is not nil.

```go
if err != nil {
  if connect.CodeOf(err) != connect.CodeAlreadyExists {
    return nil, err
  }
  return a.validateContentModulesExist(ctx, primaryRegistry, []bufmodule.Module{contentModule})
}
```

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -111,51 +111,109 @@ func (a *uploader) Upload(
...
+		// We know that there is only one module we are uploading contents for.
+		v1beta1ProtoUploadRequestContent, err := getV1Beta1ProtoUploadRequestContent(
 			ctx,
+			[]*modulev1beta1.ScopedLabelRef{
+				labelNameToV1Beta1ProtoScopedLabelRef(uploadOptions.BranchOrDraft()),
+			},
 			primaryRegistry,
-			contentModules,
-		); err != nil {
+			contentModules[0],
```
unmultimedio, line 179: Are we validating somewhere that `len(contentModules) > 0`?

private/bufpkg/bufmodule/uploader.go
```diff
@@ -115,6 +161,13 @@ func (u *uploadOptions) validate() error {
 	if u.createIfNotExist && u.createModuleVisibility == 0 {
 		return errors.New("must set a valid ModuleVisibility if CreateIfNotExist was specified")
 	}
+	// We validate that only one of labels, tags, and branchOrDraft is set.
+	// This is enforced at the flag level, so if more than one is set, we return a syserror.
+	if len(u.labels) > 0 && len(u.tags) > 0 ||
+		len(u.labels) > 0 && u.branchOrDraft != "" ||
+		len(u.tags) > 0 && u.branchOrDraft != "" {
```
unmultimedio, line 168: Similar comment about the map approach

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -111,51 +111,109 @@ func (a *uploader) Upload(
 		return nil, err
 	}
 
+	var modules []*modulev1.Module
 	if uploadOptions.CreateIfNotExist() {
-		if err := a.createContentModulesIfNotExist(
+		// We must attempt to create each module one at a time, since CreateModules is atomic.
```
bufdev, line 116: The fact that `CreateModules` is atomic isn't really the reason this doesn't work - in fact, we'd actually prefer to create all the non-existent modules in an atomic manner. The issue is that if any given module already exists, then we get a not-exists error for all modules passed.

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
bufdev: +1

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -111,51 +111,109 @@ func (a *uploader) Upload(
...
+		v1beta1ProtoScopedLabelRefs = append(
+			v1beta1ProtoScopedLabelRefs,
+			labelNameToV1Beta1ProtoScopedLabelRef(modules[0].DefaultLabelName),
+		)
+		// We know that there is only one module we are uploading contents for.
+		v1beta1ProtoUploadRequestContent, err := getV1Beta1ProtoUploadRequestContent(
+			ctx,
+			v1beta1ProtoScopedLabelRefs,
+			primaryRegistry,
+			contentModules[0],
```
bufdev, line 204: Same

private/bufpkg/bufmodule/uploader.go
```diff
@@ -65,6 +91,16 @@ type UploadOptions interface {
 	//
 	// Will always be present if CreateIfNotExist() is true.
 	CreateModuleVisibility() ModuleVisibility
+	// Tags returns unique and sorted set of tags to be added as labels.
```
bufdev, line 94: What are the differences with tags as opposed to labels? This comment implies they are the same.

My comment here is to make sure it is clear to future readers as to what the difference is, and why we have multiple fields.

private/bufpkg/bufmodule/uploader.go
```diff
@@ -65,6 +91,16 @@ type UploadOptions interface {
...
+	//
+	// We disallow the setting of `--tag` when uploading a workspace (e.g. a ModuleSet with
+	// 1+ target Modules), so we need to separate this out to handle in Upload.
+	Tags() []string
+	// BranchOrDraft returns a branch/draft to be set as a label.
```
bufdev, line 99: Same comment re: differences with labels as in above on tag. Additionally, explaining why branch and draft are using the same field here for future readers.

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
doriable: I changed the error condition above to `if len(contentModules) != 1`, since it doesn't really make sense otherwise. There is a case to be made that `Upload` should have an early return `if len(contentModules) == 0 { return nil, nil }` since there isn't anything to upload. This is also a pretty unlikely case/you're more likely to run into `no .proto files` error while building the workspace if that is happening.

private/bufpkg/bufmodule/uploader.go
doriable: Similar to above.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -267,6 +269,40 @@ func validateCreateFlags(flags *flags) error {
 }
 
 func validateLabelFlags(flags *flags) error {
+	if err := validateLabelFlagCombinations(flags); err != nil {
+		return err
+	}
+	return validateLabelFlagValues(flags)
+}
+
+// We need to validate the combination of flags that set labels:
```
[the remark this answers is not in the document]
doriable: Ah yeah, I based it off initially on the chart of UX things in the task and then cleaned it up for the PR description. Should've done that here also.

bufdev (approved):

PR #2887, opened 2024-04-10
Provide access to reflection resolver from bufimage.Image and bufprotosource

jhump:
This consolidates all code that derives a resolver for an image.

This was previously done in a few places:
1. The result from _building_ an image with `protocompile` just so happens to basically return a resolver already. But it was never used: any code that needed a resolver would separately re-build one.
2. It's mainly needed to parse image protos, to re-parse unrecognized fields to make sure all custom options are recognized and usable (requires creating `protoreflect.ExtensionType` for all custom options, which requires creating a resolver from the image contents).
3. It was also done in the lint rules for validation, to get `protoreflect.Descriptor` instances that can be used with `protovalidate` to resolve the constraints that needed to be checked.
4. Finally, in an internal repo, we have code that does this for generating package documentation in the BSR. (I have a separate PR for that repo, to verify that this works there, too.)

The main motivation for this change is to add support for Editions to `buf breaking`. I want to lean on other things already implemented by `protoreflect.EnumDescriptor` and `protoreflect.FieldDescriptor` without having to reinvent them in `bufprotosource`. Also, one of the things I want is a simple `IsMap()` accessor on a field, but that's a serious pain to do without a resolver and would require building an index of all types in `bufprotosource.NewFiles` in order to locate the message type of a field and inspect its options. So this makes things generally more efficient (never having to re-compute a resolver more than once), simplifies some code, and makes what I want to do next in `bufprotosource` much easier.

I broke up the changes so that it would be easiest to review this branch by going commit-by-commit.

The only commit that may warrant more context is the second one: that one prefactors the codebase to remove the use of `bufprotosource.NewFile` (singular), since a single `ImageFile` won't have a reference to the image resolver. That function was only called from one place, for computing free tag ranges for a message; and it turns out that _none_ of that algorithm actually needed anything in `bufprotosource`, so it was easy forklift over into `bufimageutil`. (There is also one call to `NewFile` in the BSR, but it is trivial to fix, and I have a separate PR for that.)


private/bufpkg/bufimage/bufimage.go
```diff
@@ -340,49 +343,6 @@ func CloneImageFile(imageFile ImageFile) (ImageFile, error) {
 	)
 }
 
-// MergeImages returns a new Image for the given Images. ImageFiles
```
jhump, line 343: The merge functionality was completely unused (I also checked in our internal repo with BSR code); it is only ever invoked from test code.

So I removed it. (I came across it because it has a call to `newImage` that would need to be updated.)

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -268,12 +268,7 @@ func checkEnumZeroValueSuffix(add addFunc, enumValue bufprotosource.EnumValue, s
 var CheckFieldLowerSnakeCase = newFieldCheckFunc(checkFieldLowerSnakeCase)
 
 func checkFieldLowerSnakeCase(add addFunc, field bufprotosource.Field) error {
-	message := field.ParentMessage()
-	if message == nil {
-		// just a sanity check
-		return errors.New("field.Message() was nil")
-	}
```
jhump, line 275: `message` will be nil for top-level extensions. (See other comment about bug-fix below.)

private/bufpkg/bufcheck/buflint/internal/buflintcheck/util.go
```diff
@@ -246,21 +246,32 @@ func newMessageCheckFunc(
...
+		newFileCheckFunc(
+			func(add addFunc, file bufprotosource.File) error {
+				for _, field := range file.Extensions() {
+					if err := f(add, field); err != nil {
+						return err
+					}
+				}
+				return nil
+			},
+		),
```
jhump, line 274: Doh! We previously weren't applying lint rules to top-level extensions. Fixed in this branch.

private/bufpkg/bufimage/build_image.go
```diff
@@ -428,3 +425,19 @@ type buildImageOptions struct {
 func newBuildImageOptions() *buildImageOptions {
 	return &buildImageOptions{}
 }
+
+type resolverFromBuildResult struct {
```
bufdev, line 429: `resolverForBuildResult`

Needs constructor `newResolverForBuildResult`, which should be called above

private/bufpkg/bufimage/bufimage.go
```diff
@@ -238,6 +238,9 @@ type Image interface {
 	// The path is expected to be normalized and validated.
 	// Note that all values of GetDependency() can be used here.
 	GetFile(path string) ImageFile
+	// Resolver returns a resolver backed by this image.
+	Resolver() protoencoding.Resolver
```
bufdev, line 242: I'm really not a fan of this being added as an accessor on `Image` - an `Image` is merely meant to be the wrapper for what amounts to `[]FileDescriptorProto`, in the form of this codebase this feels like something you wouldn't actually include on an `Image`. Is there any way to just return this from `BuildImage`, or does that not accomplish the goal?

Perhaps this is required, I don't know.

private/bufpkg/bufimage/build_image.go
```diff
@@ -428,3 +425,19 @@ type buildImageOptions struct {
...
+}
+
+func (r *resolverFromBuildResult) FindEnumByName(enum protoreflect.FullName) (protoreflect.EnumType, error) {
+	descriptor, err := r.Resolver.FindDescriptorByName(enum)
+	if err != nil {
+		return nil, err
+	}
+	enumDescriptor, ok := descriptor.(protoreflect.EnumDescriptor)
+	if !ok {
+		return nil, fmt.Errorf("%s is not an enum", enum)
```
bufdev, line 440: `%T is not a protoreflect.EnumDescriptor`

bufdev (commented):
Doing review commit-by-commit, here's comments from first commit

private/bufpkg/bufprotosource/bufprotosource.go
```diff
@@ -358,6 +358,11 @@ type Enum interface {
 
 	// Will return nil if this is a top-level Enum
 	Parent() Message
+
+	// AsDescriptor returns a [protoreflect.Descriptor] that
```
bufdev, line 362: On face value, I don't like that we are doing this - `bufprotosource` exposing `protoreflect` types effectively just means we're exposing two versions of a substantially-similar API to the `bufcheck/...` packages. It just feels like it may devolve in the future - I'd prefer we either (1) get rid of `bufprotosource` entirely (we're not staffing this refactor) or (2) continue to lean into `bufprotosource` (which I think is actually significantly cleaner for the linter/breaking change detector use case anyways, having just attempted to write a lint plugin against `protoreflect` in https://github.com/bufbuild/buf/tree/bufmod-lint-plugins).

I may discover why this is necessary in this commit or the next one, however.

private/bufpkg/bufprotosource/descriptor.go
```diff
@@ -46,3 +52,16 @@ func (d *descriptor) getLocationByPathKey(pathKey string) Location {
...
+
+func asDescriptor[T protoreflect.Descriptor](d *descriptor, fullName string, kind string) (T, error) {
+	var zero T
+	desc, err := d.file.resolver.FindDescriptorByName(protoreflect.FullName(fullName))
+	if err != nil {
+		return zero, err
+	}
+	typedDesc, ok := desc.(T)
+	if !ok {
+		return zero, fmt.Errorf("%s is not %s", fullName, kind)
```
bufdev, line 64: `"%s is not a %T", fullName, zero` - not sure how kind plays into this one, looks to be unused otherwise

private/bufpkg/bufcheck/buflint/internal/buflintcheck/util.go
```diff
@@ -308,3 +319,19 @@ func newMethodCheckFunc(
 		},
 	)
 }
+
+func combine(
```
bufdev, line 323: Is there a reason we did this refactor? I'm assuming it has future value?

private/bufpkg/bufprotosource/bufprotosource.go
bufdev: Talked offline, fine with this - would be nice SomedayTM to have all types inherit from the protoreflect types

private/bufpkg/bufimage/bufimage.go
bufdev: Talked offline, this is probably the best of the bad options unfortunately

private/bufpkg/bufimage/build_image.go
jhump: Done

jhump:
@bufdev, hopefully this is good enough to merge now.

I tweaked those messages even though you said F2F not to worry about it... (It literally only took a few seconds and seems like it will improve clarity in the off chance anyone ever encounters them 🤷)

private/bufpkg/bufimage/build_image.go
```diff
@@ -426,18 +427,25 @@ func newBuildImageOptions() *buildImageOptions {
...
 }
 
+// resolverFromBuildResult implements protoencoding.Resolver and is backed
+// by a linker.Resolver which is the result of a protocompile operation.
+// The linker.Resolver provides all necessary methods except FindEnumByName.
 type resolverFromBuildResult struct {
 	linker.Resolver
 }
 
+func newResolverFromBuildResult(result linker.Files) protoencoding.Resolver {
```
bufdev, line 437: `s/From/For` :-)

bufdev: for type name too

jhump: Ah, right. Done.

bufdev (approved):

PR #2888, opened 2024-04-11
Pin later protobuf-go for editions support

pkwarren:
This version is required since #2869 landed. This will resolve issues seen when running 'make upgradegodeps' in the future on this branch.


jhump:
Can we actually pull in v1.33.1-0.20240408130810-98873a205002 instead? It includes some improvements we'll want, and in particular [a fix we'll need](https://github.com/protocolbuffers/protobuf-go/commit/a18684df74fa71694a001549d6b1dc5e3d40b2ca).

jhump (approved):

PR #2889, opened 2024-04-11
Fix push with no flags

pkwarren:
Update to ensure that push sends content even if no labels are provided with --draft/--tag/--branch/--label.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
+++ b/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
@@ -116,8 +116,9 @@ func (a *uploader) Upload(
 		// We must attempt to create each module one at a time, since CreateModules will return
 		// an `AlreadyExists` if any of the modules we are attempting to create already exists,
 		// and no new modules will be created.
-		for _, contentModule := range contentModules {
-			modulesResponse, err := a.createContentModuleIfNotExist(
+		modules = make([]*modulev1.Module, len(contentModules))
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -329,7 +331,7 @@ func (a *uploader) createContentModuleIfNotExist(
 	primaryRegistry string,
 	contentModule bufmodule.Module,
 	createModuleVisibility bufmodule.ModuleVisibility,
-) ([]*modulev1.Module, error) {
+) (*modulev1.Module, error) {
```
pkwarren, line 334: This is just cleanup but I can leave it out if desired. It seems nicer as an api signature to return a single module from `createContentModuleIfNotExist` instead of a slice.

bufdev (approved):

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -140,30 +141,6 @@ func (a *uploader) Upload(
 	}
 
 	var v1beta1ProtoUploadRequestContents []*modulev1beta1.UploadRequest_Content
-	if len(uploadOptions.Labels()) > 0 {
```
pkwarren, line 143: Moved this last (and refactored to use `else` since Labels/BranchOrDraft/Tags are mutually exclusive.

PR #2890, opened 2024-04-12
Fix b5 digest calculation for remote modules

pkwarren:
When a remote module is added to the workspace, the b5 digest of it should not be calculated based on the state of other modules in the workspace (which may point to later commits). Instead, the digest should be used from the ModuleKey for the remote module (if the key references a b5 digest) or it should be fetched from the CommitProvider.

### DIFF
--- a/private/bufpkg/bufmodule/added_module.go
+++ b/private/bufpkg/bufmodule/added_module.go
@@ -282,7 +282,10 @@ func (a *addedModule) ToModule(
 		}
 		switch digestType {
 		case DigestTypeB4:
-			// Convert B4 ModuleKeys to B5 by fetching the commits from the commit provider.
+			// The declared ModuleKey dependencies for a commit may be stored in v1 buf.lock file,
+			// in which case they will use B4 digests. B4 digests aren't allowed to be used as
+			// input to the B5 digest calculation, so we perform a call to convert all ModuleKeys
+			// from B4 to B5 by using the commit provider.
 			commitKeysToFetch := make([]CommitKey, len(declaredDepModuleKeys))
 			for i, declaredDepModuleKey := range declaredDepModuleKeys {
 				commitKey, err := NewCommitKey(declaredDepModuleKey.ModuleFullName().Registry(), declaredDepModuleKey.CommitID(), DigestTypeB5)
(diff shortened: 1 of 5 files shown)


private/bufpkg/bufmodule/added_module.go
```diff
@@ -160,6 +161,28 @@ func (a *addedModule) ToModule(
...
+	getB5DigestForRemoteModule := func() (Digest, error) {
+		moduleKeyDigest, err := a.remoteModuleKey.Digest()
+		if err != nil {
+			return nil, err
+		}
+		if moduleKeyDigest.Type() == DigestTypeB5 {
+			return moduleKeyDigest, nil
+		}
+		// Our module key was created with a b4 digest - we have to look up the b5 digest for the commit.
+		commitKey, err := NewCommitKey(a.remoteModuleKey.ModuleFullName().Registry(), a.remoteModuleKey.CommitID(), DigestTypeB5)
```
pkwarren, line 173: This is the kinda thing that would be good to cache so we don't have to look it up on the fly when we're dealing with b4 module keys.

bufdev:
Let's talk offline about this tomorrow - I need to understand it a bit better

saquibmian (commented):
I think I understand the problem: the remote module's b5 digest is not calculated here with it's exact set of deps, it's calculated with the deps in the workspace, which may override some of the deps that were used when it's digest was initially calculated and stored. This would change the b5 digest for the remote module. Would this not result in a tamper proofing error, though? If so, have we seen this in the wild? We could repro this.

The solution proposed will side-step tamper proofing entirely, won't it?

pkwarren:
> I think I understand the problem: the remote module's b5 digest is not calculated here with it's exact set of deps, it's calculated with the deps in the workspace, which may override some of the deps that were used when it's digest was initially calculated and stored. This would change the b5 digest for the remote module. Would this not result in a tamper proofing error, though? If so, have we seen this in the wild? We could repro this.

Yes - we are seeing this manifest as a tamper proofing error.

> The solution proposed will side-step tamper proofing entirely, won't it?

I think if we augment the solution to tamper proof modules that are part of the workspace but look up the B5 digest for declared dep module keys from the CommitProvider if they don't exist in the workspace, this will work and not break tamper proofing.

PR #2895, opened 2024-04-15
Temp revert to v1 buf.yaml

emcfarlane:


bufdev:
This will need more than this to pull it off - you'll have to edit the Makefiles etc. Let me know why this is needed.

PR #2897, opened 2024-04-16
Update code generation to enforce supported features for editions

jhump:
This adds support for the new "supports editions" feature in `plugin.proto` and also updates code generation to use the plugin's min and max supported edition (when that new feature is supported).

This also makes the result, if the plugin does not support a feature, an _error_ instead of just a _warning_.

Currently, the structure of the code is that it writes all of the details about inadequate support to `stderr` and then returns an error. Perhaps it should pack all of those details about inadequate support _into_ the error that it returns? Looking for feedback on that, as well as on the format of the output when a plugin has inadequate support.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -285,5 +285,5 @@ issues:
     - linters:
         - gochecknoinits
       # must use init to safely enable editions support for testing
-      # (temporary)
+      # TODO: this is temporary; remove it when editions support is ready
       path: private/buf/bufgen/features_test.go
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -282,3 +282,8 @@ issues:
       # Only safe way to opt-in to protocompile editions support. Temporary and
       # will be removed once opt-in is not necessary.
       path: private/bufpkg/bufimage/build_image.go
+    - linters:
+        - gochecknoinits
+      # must use init to safely enable editions support for testing
+      # (temporary)
+      path: private/buf/bufgen/features_test.go
--- a/private/buf/bufgen/features.go
+++ b/private/buf/bufgen/features.go
@@ -15,14 +15,13 @@
 package bufgen
 
 import (
-	"bytes"
-	"errors"
 	"fmt"
 	"sort"
 	"strings"
 
 	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"go.uber.org/multierr"
 	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/pluginpb"
@@ -38,16 +37,16 @@ var featureToFeatureChecker = map[pluginpb.CodeGeneratorResponse_Feature]feature
 // requiredFeatures maps a feature to the set of files in an image that
 // make use of that feature.
 type requiredFeatures struct {
-	featureToFilename map[pluginpb.CodeGeneratorResponse_Feature][]string
-	editionToFilename map[descriptorpb.Edition][]string
-	minEdition        descriptorpb.Edition
-	maxEdition        descriptorpb.Edition
+	featureToFilenames map[pluginpb.CodeGeneratorResponse_Feature][]string
+	editionToFilenames map[descriptorpb.Edition][]string
+	minEdition         descriptorpb.Edition
+	maxEdition         descriptorpb.Edition
 }
 
 func newRequiredFeatures() *requiredFeatures {
 	return &requiredFeatures{
-		featureToFilename: map[pluginpb.CodeGeneratorResponse_Feature][]string{},
-		editionToFilename: map[descriptorpb.Edition][]string{},
+		featureToFilenames: map[pluginpb.CodeGeneratorResponse_Feature][]string{},
+		editionToFilenames: map[descriptorpb.Edition][]string{},
 	}
 }
 
@@ -66,7 +65,7 @@ func computeRequiredFeatures(image bufimage.Image) *requiredFeatures {
 		// Collect all required feature enum values.
 		for feature, checker := range featureToFeatureChecker {
 			if checker(imageFile.FileDescriptorProto()) {
-				features.featureToFilename[feature] = append(features.featureToFilename[feature], imageFile.Path())
+				features.featureToFilenames[feature] = append(features.featureToFilenames[feature], imageFile.Path())
 			}
 		}
 		// We also collect the range of required editions.
@@ -74,7 +73,7 @@ func computeRequiredFeatures(image bufimage.Image) *requiredFeatures {
 			continue
 		}
 		edition := imageFile.FileDescriptorProto().GetEdition()
-		features.editionToFilename[edition] = append(features.editionToFilename[edition], imageFile.Path())
+		features.editionToFilenames[edition] = append(features.editionToFilenames[edition], imageFile.Path())
 		if features.minEdition == 0 || edition < features.minEdition {
 			features.minEdition = edition
 		}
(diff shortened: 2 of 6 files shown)


.golangci.yml
```diff
@@ -281,4 +281,9 @@ issues:
         - gochecknoinits
       # Only safe way to opt-in to protocompile editions support. Temporary and
       # will be removed once opt-in is not necessary.
-      path: private/bufpkg/bufimage/build_image.go
\ No newline at end of file
+      path: private/bufpkg/bufimage/build_image.go
+    - linters:
+        - gochecknoinits
+      # must use init to safely enable editions support for testing
+      # (temporary)
```
bufdev, line 288: Can you add a TODO here in the comments to remove this in the future

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+	"google.golang.org/protobuf/reflect/protoreflect"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/pluginpb"
 )
 
 // requiredFeatures maps a feature to the set of files in an image that
 // make use of that feature.
-type requiredFeatures map[pluginpb.CodeGeneratorResponse_Feature][]string
+type requiredFeatures struct {
+	flags                  map[pluginpb.CodeGeneratorResponse_Feature][]string
```
bufdev, line 33: I don't know what flags actually are - regardless, this variable should be named `featureToXs`, where `X` is what the values are

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+	editions               map[descriptorpb.Edition][]string
```
bufdev, line 34: `editionToXs`, where `X` is what the values are (not sure from reading)

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+	minEdition, maxEdition descriptorpb.Edition
```
bufdev, line 35: nit: split into two lines, just to match rest of style in bufbuild/buf

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+		flags:    map[pluginpb.CodeGeneratorResponse_Feature][]string{},
+		editions: map[descriptorpb.Edition][]string{},
+	}
+}
 
 type featureChecker func(options *descriptorpb.FileDescriptorProto) bool
 
 // Map of all known features to functions that can check whether a given file
 // uses said feature.
 var allFeatures = map[pluginpb.CodeGeneratorResponse_Feature]featureChecker{
```
bufdev, line 49: global vars should always go at the top, only below global consts

Also rename to `featureToFeatureChecker`

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+					failed.editions[requiredEdition] = required.editions[requiredEdition]
+					failedEditions = append(failedEditions, requiredEdition)
+				}
+			}
+		}
+
+		pluginName := configs[responseIndex].Name()
+		if len(failedFeatures) > 0 {
+			_, _ = fmt.Fprintf(
+				container.Stderr(),
```
bufdev, line 133: Can you explain why we're printing to stderr here (and everywhere else) instead of returning an error?

private/buf/bufgen/features.go
```diff
@@ -22,86 +22,189 @@ import (
...
+		return fmt.Errorf("plugins [%v] are unable to generate code for all input files", strings.Join(failedPlugins, ","))
 	}
 }
 
 func featureName(feature pluginpb.CodeGeneratorResponse_Feature) string {
 	// FEATURE_PROTO3_OPTIONAL -> "proto3 optional"
+	return enumReadableName(feature, "FEATURE")
+}
+
+func editionName(edition descriptorpb.Edition) string {
```
bufdev, line 192: What's the purpose of this?

private/pkg/protodescriptor/protodescriptor.go
```diff
@@ -138,6 +138,14 @@ func ValidateFileDescriptor(fileDescriptor FileDescriptor) error {
 		return fmt.Errorf("%s uses edition %s, but editions are not yet supported in buf",
 			fileDescriptor.GetName(), strings.TrimPrefix(fileDescriptor.GetEdition().String(), "EDITION_"))
 	}
+	if fileDescriptor.GetSyntax() == "editions" && fileDescriptor.GetEdition() != descriptorpb.Edition_EDITION_2023 {
+		// Currently, we could only support edition 2023.
+		// TODO: For maximum safety, we probably want to ask protocompile if it supports
+		//       the edition, too. It's a little unlikely that we'd update buf CLI to
+		//       support an edition before protocompile, but just in case.
+		return fmt.Errorf("%s uses unsupported edition %s",
+			fileDescriptor.GetName(), strings.TrimPrefix(fileDescriptor.GetEdition().String(), "EDITION_"))
```
bufdev, line 147: I don't think we should be trimming `EDITION_` here

private/buf/bufgen/features.go
jhump: I called this out in the PR description -- about whether to log details to stderr or just stuff everything into the returned error. Based on this comment, it sounds like you'd prefer to see it all stuffed into the error.

private/buf/bufgen/features.go
jhump: It gets a human-readable name for the edition, to show in error messages. Was that unclear? Maybe I don't understand the question.

.golangci.yml
jhump: done

private/buf/bufgen/features.go
jhump: Changed

private/buf/bufgen/features.go
jhump: done

jhump:
@bufdev, ping

private/buf/bufgen/features.go
```diff
@@ -15,93 +15,189 @@
...
+					"  Edition %q is required by %d file(s):\n",
+					editionName(edition), len(files))
+				_, _ = fmt.Fprintf(
+					&errorDetails,
+					"    %s\n",
+					strings.Join(files, ","))
 			}
 		}
 	}
+	if errorDetails.Len() > 0 {
```
bufdev, line 174: Just use []error and multierr.Join, I haven't seen fmt.Printf to a buffer for error messages, and would argue it's strictly less unwindable than a multierror

private/buf/bufgen/features.go
bufdev: I would expect/think it's more on point to print out EDITION_2023 instead of printing our own/custom name for the edition, I wouldn't strip the prefix

private/buf/bufgen/features.go
```diff
@@ -15,93 +15,189 @@
...
+	pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS: fileHasEditions,
+}
+
 // requiredFeatures maps a feature to the set of files in an image that
 // make use of that feature.
-type requiredFeatures map[pluginpb.CodeGeneratorResponse_Feature][]string
-
-type featureChecker func(options *descriptorpb.FileDescriptorProto) bool
+type requiredFeatures struct {
+	featureToFilename map[pluginpb.CodeGeneratorResponse_Feature][]string
```
bufdev, line 41: featureToFilenames, same below (plural)

private/buf/bufgen/features.go
jhump: So a separate error for every instance? Would you expect the output then to change from this:
```
Plugin "test" does not support required feature(s).
  Feature "proto3 optional" is required by 3 file(s):
    foo.proto, bar.proto, baz.proto
  Feature "supports editions" is required by 1 file(s):
    editions.proto
```
to this?
```
plugin "test" does not support feature "proto3 optional" which is required by "foo.proto"
plugin "test" does not support feature "proto3 optional" which is required by "bar.proto"
plugin "test" does not support feature "proto3 optional" which is required by "baz.proto"
plugin "test" does not support feature "supports editions" which is required by "editions.proto"
```

jhump: Done. Output changed to above, hopefully that's what you were looking for.

private/buf/bufgen/features.go
jhump: done

bufdev (approved):

PR #2898, opened 2024-04-16
Update `user_id` to reserved in UserPluginPreferences 

gilwong00:
Update `user_id` from `UserPluginPreferences` message  and `GetUserPluginPreferencesRequest` message. We can infer the user id from the cookie so cleaning up our request payload and response.


proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -59,12 +59,10 @@ message OrganizationUser {
 // These preferences are stored in the database so the language and plugins
 // are pre-selected when the user navigates back to the Generated SDKs page from any device or browser.
 message UserPluginPreference {
-  // The id of the user.
-  string user_id = 1;
   // The language that the user has indicidated as their preferred language for plugins.
-  PluginLanguage selected_language = 2;
+  PluginLanguage selected_language = 1;
```
bufdev, line 63: As a general thing, you shouldn't do this - this is backwards-incompatible. There's no downside to using `reserved 1` and `reserved "user_id"`

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -222,9 +221,7 @@ message UpdateUserSettingsRequest {
 
 message UpdateUserSettingsResponse {}
 
-message GetUserPluginPreferencesRequest {
-  string user_id = 1;
```
fyockm, line 226: i think you need `reserved user_id` here also

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -59,8 +59,7 @@ message OrganizationUser {
 // These preferences are stored in the database so the language and plugins
 // are pre-selected when the user navigates back to the Generated SDKs page from any device or browser.
 message UserPluginPreference {
-  // The id of the user.
-  string user_id = 1;
+  reserved 1;
```
mfridman, line 62: We should probably be consistent and reserve both field number and name. E.g.,

https://github.com/bufbuild/buf/blob/0b24de2535e571ea33a0dbb280b1407c0fbb764e/proto/buf/alpha/registry/v1alpha1/organization.proto#L234-L235

bufdev (approved):

PR #2899, opened 2024-04-17
Add new breaking checks, minor tweaks to some existing ones

jhump:
This starts updating `buf breaking` to work with editions. I haven't added tests yet because I still need to wire the new checks up to the actual configs/rule sets, which I plan to do in a future branch.

This also sussed out some issues with my previous change (#2887), so those are fixed in here as well.

This change also augments the `bufprotosource.OptionExtensionDescriptor` type to return location information about options and features (instead of just about custom options/extensions).

The changes to existing checks are:
1. Use `FieldDescriptor.Kind()` instead of `*FieldDescriptorProto.GetType()` for type checks. The `FieldDescriptor` implementation is _almost_ the same as the using the `type` field of the descriptor proto, except it incorporates resolved features in Editions files to distinguish between "length-prefixed" encoded messages (standard encoding) and "delimited" encoded messages (aka "group" encoding). When a message field resolves to using delimited encoding, it actually reports `GroupKind` instead of `MessageKind`. And we need to distinguish since they are not compatible on the wire.
2. Ignore synthetic oneofs in oneof checks. In a past chat, you suggested I look at other downstream checks, but I can't find anything else that would be impacted by this outside of these two checks. Having said that, it would not be safe to cut a release with the change as is because we haven't yet added the new `FIELD_SAME_PRESENCE` check to a rule set, which is the new check that would detect synthetic-oneof related changes that these existing two checks will no longer be able to detect. Is that an issue? (I can back those changes out of this PR and add back in later when safer.)

BTW, regarding oneofs, there seems to be a check we are missing: something like `ONEOF_NO_NEW_FIELDS`. As is, if someone adds a brand new field to a message and puts it in an existing oneof, there's not a check that would catch it. Was that intentional: do we consider that a compatible change? I could understand if we did, but it does have some subtle issues with consumers of the message that have the old schema but consume newer data. (For them, the new field will go in the "unrecognized fields" set and the oneof will appear to have no value; even though it really does have a value, just not a recognized one.)

The new checks are:
1. `FIELD_SAME_CARDINALITY` and `FIELD_SAME_PRESENCE`, which will effectively _replace_ the `FIELD_SAME_LABEL` rule. These rules can handle the use of features to set field presence and also now distinguish between normal repeated fields and map fields (which is technically a cardinality change, since maps are limited to unique keys).
2. `ENUM_SAME_TYPE`, for changing an open from open to closed. This will eventually be one of the rules that replaces `FILE_SAME_SYNTAX`, since one of the semantic differences between proto2 and proto3 syntax was enums being open vs closed.


private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -55,6 +56,24 @@ func checkEnumNoDelete(add addFunc, corpus *corpus, previousFile bufprotosource.
...
+	previousDescriptor, err := previousEnum.AsDescriptor()
+	if err != nil {
+		return err
+	}
+	descriptor, err := enum.AsDescriptor()
+	if err != nil {
+		return err
+	}
+	if !previousDescriptor.IsClosed() && descriptor.IsClosed() {
+		add(enum, nil, withBackupLocation(enum.Features().EnumTypeLocation(), enum.Location()), `Enum %q changed from open to closed.`, enum.Name())
```
bufdev, line 72: Do we know that enum.Features() != nil? Can figure out myself later but thought would be quicker to ask

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -220,6 +239,33 @@ func isDeletedFieldAllowedWithRules(previousField bufprotosource.Field, message
...
+	previousDescriptor protoreflect.FieldDescriptor,
+	field bufprotosource.Field,
+	descriptor protoreflect.FieldDescriptor,
+) error {
+	previousCardinality := getCardinality(previousDescriptor)
+	currentCardinality := getCardinality(descriptor)
+	if previousCardinality != currentCardinality {
+		// otherwise prints as hex
+		numberString := strconv.FormatInt(int64(field.Number()), 10)
+		// TODO: specific label location
```
bufdev, line 258: Addressing?

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
```diff
@@ -313,6 +383,33 @@ func checkFieldSameOneof(add addFunc, corpus *corpus, previousField bufprotosour
...
+) error {
+	previousPresence := getFieldPresence(previousDescriptor)
+	currentPresence := getFieldPresence(descriptor)
+	if previousPresence != currentPresence {
+		// otherwise prints as hex
+		numberString := strconv.FormatInt(int64(field.Number()), 10)
+		// TODO: specific label location
+		add(field, nil, field.Location(),
+			`Field %q on message %q changed presence from %q to %q.`,
+			numberString, field.ParentMessage().Name(),
```
bufdev, line 405: Same question about nil (will be whole way down for such call types)

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
```diff
@@ -107,6 +108,80 @@ func newCorpus(
 	}
 }
 
+type cardinality int
+
+const (
+	cardinalityUnknown cardinality = iota
```
bufdev, line 114: Global consts always go at top, always start with iota + 1, would put in separate file

private/bufpkg/bufimage/build_image.go
```diff
@@ -449,3 +458,15 @@ func (r *resolverForBuildResult) FindEnumByName(enum protoreflect.FullName) (pro
 	}
 	return dynamicpb.NewEnumType(enumDescriptor), nil
 }
+
+func addFileToMap(allFiles map[string]linker.File, file linker.File) {
```
bufdev, line 462: keyToValue for map names - will always comment this for bufbuild/buf PRs

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
jhump: Yes.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
jhump: Sorry, this was copied from existing label check and applies here as well. Do you want me to address it now? The label is actually always the first token of a field definition, so even without a `field.LabelLocation()`, this reports the correct location. Maybe I should just remove these TODOs?

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
jhump: nil for... the `ParentMessage()`? I guess I didn't think about it since this was basically copied (nearly every field check in here is formulated exactly this way).

I guess they _all_ need to be revised to behave correctly (and avoid possible nil-dereference panics) in the face of file-level extensions.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
jhump: > always start with iota + 1

So I can skip the "unknown" value then?

> would put in separate file

Just this type? (Like each type in its own file?) Or both of these new enum-like types in the same file?

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
jhump: I've just removed the TODOs. I think it reports the correct location, even without a `field.LabelLocation()` accessor.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/bufbreakingcheck.go
jhump: Refactored this to use a helper and avoid potentially de-referencing `ParentMessage()` for top-level extensions.

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
jhump: Moved constants to top of file. If you want these types in a separate file (or files), let me know.

private/bufpkg/bufimage/build_image.go
jhump: done

jhump:
I'll rebase this against dev and re-open. I actually want to extract a few things out of this branch and into a separate one anyway.

PR #2903, opened 2024-04-19
Deprecate `is_bsr_head` field

unmultimedio:


doriable (approved):

saquibmian (approved):

saquibmian (approved):

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -120,7 +120,7 @@ message LocalModuleResolveResult {
   ResolvedReferenceType resolved_reference_type = 3;
   // Indicates whether this reference is resolvable through BSR_HEAD.
   // False indicates a non-default branch (formerly draft) commit.
-  bool is_bsr_head = 4;
+  bool is_bsr_head = 4 [deprecated = true];
```
bufdev, line 123: Are we still using this field?

unmultimedio: We have a PR open where we stop using it. The RPC still returns it though.

paul-sachs: Nope, nothing in the UI is using this field.

bufdev: Can we delete and reserve it, similar to other changes we've made?

unmultimedio: Done

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -118,9 +118,8 @@ message LocalModuleResolveResult {
   buf.alpha.registry.v1alpha1.LocalModulePin pin = 2;
   // The type the reference resolved as.
   ResolvedReferenceType resolved_reference_type = 3;
-  // Indicates whether this reference is resolvable through BSR_HEAD.
-  // False indicates a non-default branch (formerly draft) commit.
-  bool is_bsr_head = 4;
+  reserved 4;
```
bufdev, line 121: Every other declaration of `reserved` in this repository save one puts the `reserved` declarations at the top of the message, can you follow this convention?

bufdev (approved):
I'm approving this so there's no more back and forth, but please fix the comment up

PR #2913, opened 2024-04-23
Strip legacy options so they don't cause issues reparsing an image

jhump:
The Protobuf-Go runtime is unhappy if it is asked to create a `protoreflect.FileDescriptor` for a file with any references to certain legacy options: `MessageOptions.message_set_wire_format` and `FieldOptions.weak`.

This updates `protoencoding.NewResolver` so that it can modify the schema before creating the descriptors with the `protodesc` package. It does **not** mutate the actual descriptor protos that are part of the schema/image. So the schema is untouched, and the modifications are just a way to "trick" the runtime into allowing a resolver. The resolver is typically used for serialization and de-serialization (like looking up extensions/custom options and message types in an Any) so these slight differences from the actual underlying schema should not matter.


private/pkg/protoencoding/strip_legacy_options.go
```diff
@@ -0,0 +1,298 @@
...
+
+import (
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// stripLegacyOptions modifies the given slice so that it does not contain legacy
+// elements that the Go protobuf runtime does not support -- namely messages that use
```
doriable, line 25: nit: `%s/modifies the given slice so that it does not contain.../modifies the given slice of file descriptors so that the slice does not contain...`

Just for clarity that `FileDescriptorProto`s is the structure being modified (similar to how subsequent functions make it clear we are modifying file/message/field descriptors).

doriable (commented):
Had a conversation with @jhump outside of the PR about this: basically `buf convert` uses the resolver to unmarshal the message, and this change may impact deserialising/serialising messages with `MessageOptions.message_set_wire_format` set. But since this is not supported by the protobuf-go runtime, we need to make sure that this option is not set on messages being deserialised/serialised by `buf convert`, so we need to handle that there.

jhump:
> Had a conversation with @jhump outside of the PR about this: basically `buf convert` uses the resolver to unmarshal the message, and this change may impact deserialising/serialising messages with `MessageOptions.message_set_wire_format` set. But since this is not supported by the protobuf-go runtime, we need to make sure that this option is not set on messages being deserialised/serialised by buf convert, so we need to handle that there.

FYI, @doriable, I just pushed a commit that prevents use of message sets with `buf convert`. It also pulls in some of the changes from #2915, so the whole path through `buf convert` can lean on the schema's resolver. Without those changes, converting a single message would end up calling `protodesc.NewFiles` _three times_. But beyond making things more efficient, the fix in that commit relies on using the schema's resolver as the chokepoint where we can catch attempts to potentially use message sets. We do that, instead of a simple check just for the message type being converted, in the off chance that the input has references to extensions or messages inside of `Any` instances that refer to message sets.

private/pkg/protoencoding/strip_legacy_options.go
jhump: done

private/buf/cmd/buf/command/convert/no_resolve_message_set.go
```diff
@@ -0,0 +1,209 @@
...
+	"google.golang.org/protobuf/types/descriptorpb"
+)
+
+// noResolveMessageSetWireFormat returns an image with the same contents
+// as the given image, but whose resolver refuses to resolve elements that
+// refer to messages that use the message-set wire format. We do this
+// because the protobuf-go runtime does not support message-set wire format
+// (unless built with a "protolegacy" build tag). So if such a message were
+// used, we would not be able to correctly serialize/de-serialize it.
+func noResolveMessageSetWireFormat(image bufimage.Image) bufimage.Image {
```
bufdev, line 33: `imageWithoutMessageSetWireFormatResolution`

private/buf/cmd/buf/command/convert/no_resolve_message_set.go
```diff
@@ -0,0 +1,209 @@
+// Copyright 2020-2024 Buf Technologies, Inc.
```
bufdev, line 1: filename doesn't match the type name

private/buf/cmd/buf/command/convert/no_resolve_message_set.go
```diff
@@ -0,0 +1,209 @@
...
+//
```
bufdev, line 2: this type of logic should not live in `private/buf/cmd/buf` - as much as possible, very little logic should live here. this needs to find the appropriate package to live in, its a component of the library

private/buf/cmd/buf/command/convert/no_resolve_message_set.go
```diff
@@ -0,0 +1,209 @@
...
+// used, we would not be able to correctly serialize/de-serialize it.
+func noResolveMessageSetWireFormat(image bufimage.Image) bufimage.Image {
+	return noResolveMessageSetWireFormatImage{image}
+}
+
+type noResolveMessageSetWireFormatImage struct {
+	bufimage.Image
+}
+
+func (n noResolveMessageSetWireFormatImage) Resolver() protoencoding.Resolver {
```
bufdev, line 41: everything should be on the pointer

bufdev (commented):
overall logic seems fine, assuming we've QA'ed that there's no perf hit here

jhump:
>  assuming we've QA'ed that there's no perf hit here

Not a meaningful one. 99.999% of the time  there is nothing to strip, so it's just the cost of examining all extensions, extension ranges, message, and field options. I benchmarked this using the _full_ `googleapis` (nearly 5000 files).

**before**:
```
BenchmarkNewResolver-10    	      52	 203,764,690 ns/op	493,371,083 B/op	  940,700 allocs/op
```

**after**:
```
BenchmarkNewResolver-10    	      57	 206,037,626 ns/op	493,374,574 B/op	  940,710 allocs/op
```

So a roughly 1% increase in the observed duration to construct the resolver.

private/buf/cmd/buf/command/convert/no_resolve_message_set.go
jhump: done

jhump:
@bufdev, I think I address your feedback. Whenever you have a moment, this is ready for another pass.

doriable (approved):

PR #2915, opened 2024-04-24
Improve efficiency of resolver backed by protocompile output

jhump:
For another arc of work (related to #2913), I was benchmarking the creation of resolvers and their use in re-parsing unrecognized fields (i.e. custom options) when loading images from a binary-encoded descriptor protos.

I happened to be using the _full_ `googleapis/googleapis` for the benchmark (nearly 5000 files), which really tickles the inefficiency of this resolver, since its look-ups were linear _with the number of files_.

So now, we use the compiler's symbol table as an index, to dead-reckon which file a symbol or extension is in. That improved the benchmarks drastically; using this resolver is now more efficient than creating and using a `*protoregistry.Types` (and less memory needed, too, since there are no additional data structures to create beyond the compiler's data structures).

I've also updated numerous calls to `protoencoding.NewResolver` to instead use `Image.Resolver()` since, in many cases, that's the resolver they want and were re-creating.

### DIFF
--- a/private/bufpkg/bufimage/bufimage.go
+++ b/private/bufpkg/bufimage/bufimage.go
@@ -357,10 +357,11 @@ func NewImageForProto(protoImage *imagev1.Image, options ...NewImageForProtoOpti
 	if newImageOptions.noReparse && newImageOptions.computeUnusedImports {
 		return nil, fmt.Errorf("cannot use both WithNoReparse and WithComputeUnusedImports options; they are mutually exclusive")
 	}
-	var resolver protoencoding.Resolver
+	// TODO FUTURE: right now, NewResolver sets AllowUnresolvable to true all the time
+	// we want to make this into a check, and we verify if we need this for the individual command
+	resolver := protoencoding.NewLazyResolver(protoImage.File...)
 	if !newImageOptions.noReparse {
-		var err error
-		if resolver, err = reparseImageProto(protoImage, newImageOptions.computeUnusedImports); err != nil {
+		if err := reparseImageProto(protoImage, resolver, newImageOptions.computeUnusedImports); err != nil {
 			return nil, err
 		}
 	}
@@ -480,7 +481,7 @@ func ImageWithoutImports(image Image) Image {
 			newImageFiles = append(newImageFiles, imageFile)
 		}
 	}
-	return newImageNoValidate(newImageFiles)
+	return newImageNoValidate(newImageFiles, image.Resolver())
 }
 
 // ImageWithOnlyPaths returns a copy of the Image that only includes the files
@@ -669,12 +670,9 @@ type newImageForProtoOptions struct {
 	computeUnusedImports bool
 }
 
-func reparseImageProto(protoImage *imagev1.Image, computeUnusedImports bool) (protoencoding.Resolver, error) {
-	// TODO FUTURE: right now, NewResolver sets AllowUnresolvable to true all the time
-	// we want to make this into a check, and we verify if we need this for the individual command
-	resolver := protoencoding.NewLazyResolver(protoImage.File...)
+func reparseImageProto(protoImage *imagev1.Image, resolver protoencoding.Resolver, computeUnusedImports bool) error {
 	if err := protoencoding.ReparseUnrecognized(resolver, protoImage.ProtoReflect()); err != nil {
-		return nil, fmt.Errorf("could not reparse image: %v", err)
+		return fmt.Errorf("could not reparse image: %v", err)
 	}
 	if computeUnusedImports {
 		tracker := &importTracker{
@@ -708,7 +706,7 @@ func reparseImageProto(protoImage *imagev1.Image, computeUnusedImports bool) (pr
 			}
 		}
 	}
-	return resolver, nil
+	return nil
 }
 
 // We pass in the pathToImageFileInfo here because we also call this in
--- a/private/bufpkg/bufimage/build_image.go
+++ b/private/bufpkg/bufimage/build_image.go
@@ -494,27 +494,6 @@ func (r *resolverForFiles) FindExtensionByNumber(message protoreflect.FullName, 
 	return dynamicpb.NewExtensionType(ext), nil
 }
 
-type container interface {
-	Messages() protoreflect.MessageDescriptors
-	Extensions() protoreflect.ExtensionDescriptors
-}
-
-func findExtension(d container, message protoreflect.FullName, field protoreflect.FieldNumber) protoreflect.ExtensionDescriptor {
-	exts := d.Extensions()
-	for i, length := 0, exts.Len(); i < length; i++ {
-		ext := exts.Get(i)
-		if ext.Number() == field && ext.ContainingMessage().FullName() == message {
-			return ext
-		}
-	}
-	for i := 0; i < d.Messages().Len(); i++ {
-		if ext := findExtension(d.Messages().Get(i), message, field); ext != nil {
-			return ext
-		}
-	}
-	return nil // could not be found
-}
-
 func (r *resolverForFiles) FindMessageByName(message protoreflect.FullName) (protoreflect.MessageType, error) {
 	descriptor, err := r.FindDescriptorByName(message)
 	if err != nil {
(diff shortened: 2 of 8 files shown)


private/bufpkg/bufimage/bufimage.go
```diff
@@ -480,7 +481,7 @@ func ImageWithoutImports(image Image) Image {
 			newImageFiles = append(newImageFiles, imageFile)
 		}
 	}
-	return newImageNoValidate(newImageFiles)
+	return newImageNoValidate(newImageFiles, image.Resolver())
```
jhump, line 484: This was a separate bug: this function (`ImageWithoutImports`) was creating an image that had no resolver.

private/bufpkg/bufimage/bufimage.go
```diff
@@ -357,10 +357,11 @@ func NewImageForProto(protoImage *imagev1.Image, options ...NewImageForProtoOpti
 	if newImageOptions.noReparse && newImageOptions.computeUnusedImports {
 		return nil, fmt.Errorf("cannot use both WithNoReparse and WithComputeUnusedImports options; they are mutually exclusive")
 	}
-	var resolver protoencoding.Resolver
+	// TODO FUTURE: right now, NewResolver sets AllowUnresolvable to true all the time
+	// we want to make this into a check, and we verify if we need this for the individual command
+	resolver := protoencoding.NewLazyResolver(protoImage.File...)
```
jhump, line 362: This was a readability improvement (IMO). In previous iterations, `reparseImageProto` create the resolver and used it to reparse unrecognized fields. In a prior PR, I changed it to return the resolver since we want to use it from here. But I think it's easier to read to have the caller here create the resolver and pass it in to `reparseImageProto`.

private/bufpkg/bufimage/build_image.go
```diff
@@ -80,14 +82,15 @@ func buildImage(
 	if buildResult.Err != nil {
 		return nil, buildResult.Err
 	}
-	fileDescriptors, err := checkAndSortFileDescriptors(buildResult.FileDescriptors, paths)
+	sortedFiles, err := checkAndSortFiles(buildResult.Files, paths)
```
jhump, line 85: I tweaked the names of a few things in here for better consistency and readability. In an earlier PR, I had changes some of the types from `protodescriptor.FileDescriptor` to `linker.File`, but didn't change the corresponding "FileDescriptor" names. This fixes that omission.

private/bufpkg/bufimage/build_image.go
```diff
@@ -137,10 +142,7 @@ func getBuildResult(
 	if err != nil {
 		if err == reporter.ErrInvalidSource {
 			if len(errorsWithPos) == 0 {
-				return newBuildResult(
-					nil,
-					nil,
-					nil,
+				return newFailedBuildResult(
```
jhump, line 145: I introduced this new helper/constructor to aid readability and lessen the `nil` verbosity in this function.

private/bufpkg/bufimage/build_image.go
```diff
@@ -427,25 +426,132 @@ func newBuildImageOptions() *buildImageOptions {
...
-type resolverForBuildResult struct {
-	linker.Resolver
+// resolverForFiles implements protoencoding.Resolver and is backed
+// by a linker.Files and the *linker.Symbols symbol table produced
+// when compiling the files. The symbol table is used as an index
+// for more efficient lookup.
+type resolverForFiles struct {
+	pathToFile map[string]linker.File
+	symbols    *linker.Symbols
+}
```
jhump, line 436: This is the meat of the change. This new resolver implementation is subtantially more efficient than the previous implementation (which leaned on a fairly naive implementation in protocompile's `linker` package).

private/bufpkg/bufimage/build_image.go
```diff
@@ -427,25 +426,132 @@ func newBuildImageOptions() *buildImageOptions {
...
+	msg, ok := descriptor.(protoreflect.MessageDescriptor)
 	if !ok {
-		return nil, fmt.Errorf("%s is a %T, not a protoreflect.EnumDescriptor", enum, descriptor)
+		return nil, fmt.Errorf("element %q is not a message", message)
+	}
+	return dynamicpb.NewMessageType(msg), nil
+}
+
+func (r *resolverForFiles) FindMessageByURL(url string) (protoreflect.MessageType, error) {
+	pos := strings.IndexByte(url, '/')
```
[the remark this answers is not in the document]
jhump: 🤦, yes!

private/bufpkg/bufimage/bufimage.go
jhump: > this just optimistically creates the resolver

While true, it's a **lazy** resolver. That means it really is just allocating a placeholder and doing no other work. Only when the resolver is first used will it actually do the work to create the resolver. If creating the resolver fails at that deferred step, then all `Find...` operations will fail with the error that prevented its construction. So we aren't really doing any meaningful work eagerly -- it only happens if/when we need to _use_ the resolver.

private/bufpkg/bufimage/build_image.go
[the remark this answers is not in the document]
jhump: fixed

private/bufpkg/bufimage/build_image.go
```diff
@@ -427,25 +426,132 @@ func newBuildImageOptions() *buildImageOptions {
...
 }
 
-func newResolverForBuildResult(result linker.Files) protoencoding.Resolver {
-	return &resolverForBuildResult{Resolver: result.AsResolver()}
+type container interface {
+	Messages() protoreflect.MessageDescriptors
+	Extensions() protoreflect.ExtensionDescriptors
+}
+
+func findExtension(d container, message protoreflect.FullName, field protoreflect.FieldNumber) protoreflect.ExtensionDescriptor {
```
bufdev, line 502: move below all method receivers

bufdev (commented):
there seems to be changes duplicated across here and #2913, is that right?

overall seems fine

jhump:
> there seems to be changes duplicated across here and https://github.com/bufbuild/buf/pull/2913, is that right?

Indeed: https://github.com/bufbuild/buf/pull/2913#issuecomment-2076057793

I ended up needing some of what's here to address the `buf convert` issue that @doriable raised in the other PR. My plan was, after one of them is merged, to reconcile the other.

private/bufpkg/bufimage/build_image.go
jhump: done

private/bufpkg/bufimage/build_image.go
```diff
@@ -427,25 +426,132 @@ func newBuildImageOptions() *buildImageOptions {
...
 
-func (r *resolverForBuildResult) FindEnumByName(enum protoreflect.FullName) (protoreflect.EnumType, error) {
-	descriptor, err := r.Resolver.FindDescriptorByName(enum)
+func (r *resolverForFiles) FindEnumByName(enum protoreflect.FullName) (protoreflect.EnumType, error) {
+	descriptor, err := r.FindDescriptorByName(enum)
 	if err != nil {
 		return nil, err
 	}
-	enumDescriptor, ok := descriptor.(protoreflect.EnumDescriptor)
+	en, ok := descriptor.(protoreflect.EnumDescriptor)
```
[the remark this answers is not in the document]
jhump: Oops, this was unintentional; will definitely revert. (This whole resolver was originally in a different package in another work-in-progress branch, and I copied it over wholesale without realizing that it was using different names; that's also why the error message below is different, which I'll change back, too.)

As to other variable names in this PR, are you just referring to new variables added above, like `msg` and `ext`? I'll go ahead and flesh those out...  Other changes to existing variable names, though, were intentional, hoping to in fact better align with consistency in this repo (see [this comment](https://github.com/bufbuild/buf/pull/2915/files#r1577938368)). Was that incorrect to do so?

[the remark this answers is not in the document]
jhump: Fixed (assuming the other changes referenced in that comment are okay to keep).

jhump:
I am merging this since it's been approved. I think I fixed the naming issues you mentioned in that last commit. But if there are any others I missed, please leave a comment on them, and I'll follow up with another branch to fix them.

PR #2916, opened 2024-04-25
Allow reconfiguration of extension and reserved ranges as long as no numbers are removed

jhump:
Previously, the breaking checks for extension and reserved ranges required exactly the same ranges to be preserved. So the following would produce an error, even though it is technically not a breaking change since all tag numbers that are extensions in the first schema are still extensions in the second:
```proto
// first
message Foo {
  extensions 100 to 200;
  extensions 1000 to max;
}

// second
message Foo {
  extensions 100 to 150, 151, 152 to 225;
  extensions 999 to max;
}
```

This PR changes that so it does more analysis to identify the actual coverage of the ranges. It is only an error if the new ranges do not cover all tag numbers that the previous ranges covered.

One motivation for this is if anyone adopts "extension declarations", the suggested approach involves breaking up ranges as extensions are declared incrementally. For example:
```proto
// first
message Foo {
  extensions 100 to 200;
}

// second
message Foo {
  extensions 101 to 103 [
    verification = DECLARATION,
    declaration = {
      number: 101
      reserved: true
    }
    declaration = {
      number: 102
      full_name: ".foo.bar.buzz"
      type: ".foo.bar.Buzz"
    }
    declaration = {
      number: 103
      full_name: ".foo.bar.baz"
      type: "string"
      repeated: true
    }
  ];
  extensions 104 to 200; // no declarations for these yet
}
```


private/bufpkg/bufcheck/bufbreaking/bufbreaking_test.go
```diff
@@ -114,9 +114,10 @@ func TestRunBreakingExtensionMessageNoDelete(t *testing.T) {
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 30, 3, 36, 4, "EXTENSION_MESSAGE_NO_DELETE"),
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 30, 3, 36, 4, "EXTENSION_MESSAGE_NO_DELETE"),
 		bufanalysistesting.NewFileAnnotation(t, "1.proto", 30, 3, 36, 4, "EXTENSION_MESSAGE_NO_DELETE"),
-		bufanalysistesting.NewFileAnnotation(t, "2.proto", 73, 1, 79, 2, "EXTENSION_MESSAGE_NO_DELETE"),
-		bufanalysistesting.NewFileAnnotation(t, "2.proto", 73, 1, 79, 2, "EXTENSION_MESSAGE_NO_DELETE"),
-		bufanalysistesting.NewFileAnnotation(t, "2.proto", 73, 1, 79, 2, "EXTENSION_MESSAGE_NO_DELETE"),
+		bufanalysistesting.NewFileAnnotation(t, "1.proto", 49, 1, 53, 2, "EXTENSION_MESSAGE_NO_DELETE"),
```
jhump, line 117: Each of these adds a new problem, for the newly added `Ten` message. The other changes here are just shifting line numbers, since I added two lines towards the top in the `One` message.

The actual messages look like so:
```
Previously present extension range "[4,5]" on message "Seven" is missing values: [4,5] were removed
Previously present extension range "[1,max]" on message "Ten" is missing values: [1001,1999], [5001,9999], [50001,99999] were removed

```

private/bufpkg/bufprotosource/message_range.go
```diff
@@ -14,8 +14,14 @@
 
 package bufprotosource
 
+import "math"
+
 const (
-	messageRangeInclusiveMax = 536870911
+	// MessageRangeInclusiveMax is the maximum allowed tag for a message field.
+	MessageRangeInclusiveMax = 536870911 // 2^29 - 1
```
jhump, line 21: Note to self, if we're okay making this exported, I should reference it in #2913 (instead of [redefining it](https://github.com/bufbuild/buf/pull/2913/files#diff-59d79b7b8b1bea1b73c878e3b8749e826378a94595cc236c57913f54ead46d97R27))

bufdev (approved):
Approving assuming comment is addressed

private/bufpkg/bufprotosource/message_range.go
jhump: Actually, we don't want `protoencoding` to import `bufprotosource`. So I'll leave this constant duplicated for now. Maybe we can consolidate them into a different package later...

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/tag_ranges.go
```diff
@@ -0,0 +1,166 @@
...
+			elementKind, maxTag, err := classifyElementRange(element)
+			if err != nil {
+				return err
+			}
+			previousString := bufprotosource.TagRangeString(previousRange)
+			removedString := missingRangesString(maxTag, missingRanges)
+			add(element,
+				nil,
+				element.Location(),
+				`Previously present %s range %q on %s %q is missing values: %s were removed`,
```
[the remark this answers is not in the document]
jhump: done

PR #2917, opened 2024-04-25
Update breaking checks for field types to consider message encoding feature

jhump:
This is for Editions support: with editions, just looking at the field descriptor's `type` field won't tell you if the message is length-prefixed (normal) or delimited (a la groups). But they are not compatible. The protobuf-go runtime _does_ return `GroupKind` when the field uses delimited encoding, even though the underlying descriptor proto would report "MESSAGE" as its type. So we leverage that, so we don't have to reinvent anything regarding feature resolution.

The tests are updated to make sure that only the messages with changed encoding are flagged. The `FIELD_SAME_TYPE` test even includes a group in "proto2" being moved to a delimited message in editions, and it (correctly) does _not_ trigger a breaking change error.

### DIFF
--- a/private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
+++ b/private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
@@ -22,65 +22,65 @@ import (
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
 	"github.com/bufbuild/buf/private/bufpkg/bufcheck/internal"
 	"github.com/bufbuild/buf/private/bufpkg/bufprotosource"
-	"google.golang.org/protobuf/types/descriptorpb"
+	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
 var (
 	// https://developers.google.com/protocol-buffers/docs/proto3#updating
-	fieldDescriptorProtoTypeToWireCompatiblityGroup = map[descriptorpb.FieldDescriptorProto_Type]int{
-		descriptorpb.FieldDescriptorProto_TYPE_INT32:  1,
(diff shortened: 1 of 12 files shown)


private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
```diff
@@ -22,65 +22,65 @@ import (
...
 	// https://developers.google.com/protocol-buffers/docs/proto3#updating
-	fieldDescriptorProtoTypeToWireCompatiblityGroup = map[descriptorpb.FieldDescriptorProto_Type]int{
-		descriptorpb.FieldDescriptorProto_TYPE_INT32:  1,
-		descriptorpb.FieldDescriptorProto_TYPE_INT64:  1,
-		descriptorpb.FieldDescriptorProto_TYPE_UINT32: 1,
-		descriptorpb.FieldDescriptorProto_TYPE_UINT64: 1,
-		descriptorpb.FieldDescriptorProto_TYPE_BOOL:   1,
-		descriptorpb.FieldDescriptorProto_TYPE_SINT32: 2,
-		descriptorpb.FieldDescriptorProto_TYPE_SINT64: 2,
+	fieldDescriptorProtoTypeToWireCompatiblityGroup = map[protoreflect.Kind]int{
```
bufdev, line 30: The name of this map should change, right?

bufdev (approved):

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/util.go
jhump: Ah, yes. Done. Changed `fieldDescriptorProtoType` to `fieldKind` in this map name and the other one below.

PR #2919, opened 2024-04-26
fix: close webpages config

testwill:


bufdev:
Thanks for pointing this out! We'll get it in an update soon.

PR #2920, opened 2024-04-26
Make "buf lint" work correctly with top-level extensions

jhump:
In a previous PR, I tried to fix "buf lint" so it would work with top-level extensions ([here](https://github.com/bufbuild/buf/pull/2887/files#r1559894008)). However, I only fixed the potential panic in one of the field rules and failed to address a similar issue in another field linter rule.

So this fixes the other rule and adds tests for both rules that cover extensions -- both nested extensions inside messages as well as top-level extensions.


private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -275,14 +275,16 @@ func checkFieldLowerSnakeCase(add addFunc, field bufprotosource.Field) error {
 	name := field.Name()
 	expectedName := fieldToLowerSnakeCase(name)
 	if name != expectedName {
+		var otherLocs []bufprotosource.Location
+		if field.Extendee() == "" {
+			// also check the message for this comment ignore
+			// this allows users to set this "globally" for a message
+			otherLocs = []bufprotosource.Location{field.ParentMessage().Location()}
+		}
```
jhump, line 283: This is really an open question: **Should** a message-level ignore comment apply to both regular fields _and_ extensions inside the message?

This branch currently does _not_ apply the comment to extensions at all. But I could easily change `if field.Extendee() == ""` to instead be `if field.ParentMessage() != nil`, and then the message-level ignores also apply to nested extensions.

Thoughts?

bufdev: I don't think message ignores should apply to extensions - they don't apply currently right?

jhump: > they don't apply currently right?

Oh, that's a great point: they actually _do_ apply to extensions nested inside the message. So we should preserve that property for backwards compatibility.

PR #2921, opened 2024-04-26
Add start and end time params to GetClusterUsageRequest message

gilwong00:
Update `GetClusterUsageRequest` message to take in a `start_time` and `end_time` date range.


bufdev (approved):

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -253,7 +254,16 @@ message UpdateReviewFlowGracePeriodPolicyRequest {
...
 message UpdateReviewFlowGracePeriodPolicyResponse {}
 
-message GetClusterUsageRequest {}
+message GetClusterUsageRequest {
+  // The timestamp of the beginning time of a date range.
+  google.protobuf.Timestamp start_time = 1;
+  // The timestamp of the ending time of a date range.
+  google.protobuf.Timestamp end_time = 2;
+
+  optional string organization_id = 3;
```
bufdev, line 263: Why are these being marked optional?

bufdev (changes requested):

proto/buf/alpha/registry/v1alpha1/admin.proto
bufdev: Why were they marked optional though? Was there a reason as opposed to the default concept of optional in proto3?

bufdev (approved):

PR #2931, opened 2024-04-29
Add `--create-default-label` flag to `buf push`

doriable:
This adds the `--create-default-label` flag to `buf push`, which allows
the user to set a default label name for a module they are creating with `--create`.
We do not allow this to be set if `--create` is not set, however, the user does
not need to set this if they are using `--create`. If this is not set during `--create`,
then the module will have a default label of "main".

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -34,12 +34,13 @@ import (
 )
 
 const (
-	labelFlagName            = "label"
-	errorFormatFlagName      = "error-format"
-	disableSymlinksFlagName  = "disable-symlinks"
-	createFlagName           = "create"
-	createVisibilityFlagName = "create-visibility"
-	sourceControlURLFlagName = "source-control-url"
+	labelFlagName              = "label"
+	errorFormatFlagName        = "error-format"
+	disableSymlinksFlagName    = "disable-symlinks"
+	createFlagName             = "create"
+	createVisibilityFlagName   = "create-visibility"
+	createDefaultLabelFlagName = "create-default-label"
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufmodule/uploader.go
```diff
@@ -60,11 +60,13 @@ func UploadWithTags(tags ...string) UploadOption {
 }
 
 // UploadWithCreateIfNotExist returns a new UploadOption that will result in the
-// Modules being created on the registry with the given visibility if they do not exist.
-func UploadWithCreateIfNotExist(createModuleVisibility ModuleVisibility) UploadOption {
+// Modules being created on the registry with the given visibility and default label if they do not exist.
+// If not default label name is provided, the module will be created with the default label "main".
```
pkwarren, line 64:
```suggestion
// If the default label name is not provided, the module will be created with the default label "main".
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -275,6 +286,13 @@ func validateCreateFlags(flags *flags) error {
 				createFlagName,
 			)
 		}
+		if flags.CreateDefaultLabel != "" {
```
pkwarren, line 289: Is there any client side validation we should perform on the label name in the if statement above or are we going to defer to the BSR?

doriable: There isn't really any client-side validation that I personally want to add -- I feel like the general rule we want to follow is minimal client-side validation. The types of validation that come to mind are mostly around restricted words, etc. which I think we would want to defer that to the server-side. But open to suggestions on this :D

pkwarren (approved):

PR #2939, opened 2024-05-02
Upgrade protoc version to v27.0-rc1

bufdev:


jhump:
I ran into some issues that popped up due to having `datawkt` at a newer version of `descriptor.proto` than the protobuf-go `descriptorpb` package. Some new fields were added in v27.0 (all Editions-related), but the new fields are used inside of `descriptor.proto` itself (metadata on features). So these new fields are represented as unrecognized fields in the `descriptorpb` message types, which get dropped when serializing to JSON. So this change causes `TestImageConvertRoundtripBinaryJSONBinary` (in `private/buf/cmd/buf/buf_test.go`) to fail.

Having to go a different route shouldn't be too bad since it's temporary: once v27.0 of protoc is released and the protobuf-go runtime is updated to use v27 of `descriptor.proto`, we can change the approach in my other arc of work to use code gen based on `buf.build/protocolbuffers/wellknowntypes`, which will be cleaner than using the sources in `datawkt` anyway I think.

PR #2940, opened 2024-05-02
Migrating to v2 will translate deprecated rules into their replacements

jhump:
I can't find any tests for this stuff. Should I just be manually verifying that this is working as expected?


bufdev (approved):
Approved pending manual QA and adding test using `buf config ls-breaking-rules --format=json --configured_only` and validating `id` list when doing things like `use: PACKAGE except: FIELD_SAME_LABEL`

jhump:
@bufdev, I added a new test in 4bcf0d2.

Maybe take a look and see if that is more or less what you were expecting?

Also, I've manually verified this works (but one bug I had to fix, in 62614b3). So I think this is good to go. I know you already approved, but I was hoping you could look at that new test and let me know if you want something else/more.

bufdev:
Yea, roughly. I was imagining a helper function that would be reused by all these new tests, that took a list of `use/except` as parameters, translated this info `--config`, called `ls-breaking-rules`, and that effectivly took the result of `--format=json` and parsed out just the IDs, and then made it easy to compare to a list of IDs...but either works

PR #2941, opened 2024-05-02
Add buf breaking check to make sure default value doesn't change

jhump:
This is a new rule for the v2 config.


bufdev (approved):

private/bufpkg/bufcheck/bufbreaking/internal/bufbreakingcheck/field_default.go
```diff
@@ -0,0 +1,171 @@
...
+	"math/big"
+	"reflect"
+	"strings"
+
+	"github.com/bufbuild/buf/private/pkg/slicesext"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type fieldDefault struct {
+	comparable any
```
jhump, line 29:
```suggestion
	comparable comparable
```
Will need to update `getDefault` to return an error if it can't type-assert the value to `comparable`, too.

jhump: @bufdev, doh, this isn't possible. The `comparable` type uses a "type term", which means it can only be used as a constraint for a type parameter, and not as a type for a variable, parameter, or struct element. And I can't make this generic (and use a type parameter) because the types aren't known at compile-time: the value is coming from an `any` (returned by `protoreflect.Value.Interface()`).

PR #2942, opened 2024-05-03
Replace FIELD_SAME_CTYPE and FILE_SAME_JAVA_STRING_CHECK_UTF8 checks with Editions-aware versions

jhump:
This one's a bit of a doozy for a few reasons:
1. We need extension descriptors for the custom features, but can't do code gen. The biggest hurdle to code gen has to do with the fact that `datawkt` has an older version of `cpp_features.proto` and `descriptor.proto`. And we can't update it because then the mismatch of the version in `datawkt` and the version used to generate `descriptorpb` causes [other issues](https://github.com/bufbuild/buf/pull/2939#issuecomment-2092874849).

   FWIW, I tried exploring a route where we use an alternate `descriptor.proto`, just for the `buf generate` step, but there's no easy way to get it to generate code for `cpp_features.proto` and _not_ for `descriptor.proto`, since they are both recognized as part of the well-known types. If we could rely on `datawkt` for `descriptor.proto`, then I could make something work, but alas...
   
   ~The end result is using the newer `cpp_features.proto` and alternate `descriptor.proto`, but without code gen. Instead, it just runs `buf build` and we embed the resulting descriptor in a Go package.~

   **Edit**: On Friday, the Go team at Google merged a commit that updates `descriptorpb` to v27.0-rc1. So I've updated this PR so that it updates this module to that newer version of protoc and pins that new commit in protobuf-go. I was then able to do everything with code gen instead of embedding binary descriptors. This is particularly nice as it means we don't need to hand-maintain copies of the enum values in Go constants.

2. The logic for interacting with custom extensions is non-trivial. I chose to have this _re-parse_ the options to make sure all parsed extension values are associated with the correct extension descriptor. The alternative would require doing everything dynamically, solely with strings and number for type names, field names, and field numbers. That logic (a) would have been hairy, and (b) would require inventing the logic for resolving a feature instead of relying on the implementation already provided by protocompile. (It must implement this in order to resolve features for its implementations of `protoreflect.Descriptor`, so I just exposed that logic via its `protoutil` package and then use that from this code.)

3. The logic for computing whether a string is validated to be UTF8 for Java is non-trivial. The logic isn't too many lines of code, but it's subtle enough that this resulted in a very large set of files and messages in the tests... so a large amount of test code for the new FIELD_SAME_JAVA_UTF8_VALIDATION check.

~The use of the embedded descriptors is kinda ugly, but it's only temporary. Once the protobuf-go's `descriptorpb` package is updated to use v27 of `descriptor.proto`, we can update the `datawkt` package and then clean this up a bit. I think we can also figure out a way to do code-gen at that point (so the `customfeatures` logic would interact with a generated extension type in Go code instead of having to parse descriptors and create a dynamic extension type).~


jhump:
BTW, here was an attempt to use the sources in the `datawkt` package instead of the embedded binary descriptors.
https://github.com/bufbuild/buf/pull/2944

I ran into issues so decided to not pursue that route further. But once `descriptorpb` and `datawkt` are both updated to v27, I can go that route instead. (Though I might also be able to use code-gen at that point, and use generated Go code with the extension descriptors instead.)

bufdev (approved):
- File code in linear about getting rid of built-in descriptors
- See if there's a way to generate the enums
- Add comments to upgrading protoc doc
- More comments on how to update the enums if we can't generate them

bufdev (approved):

PR #2943, opened 2024-05-03
Add ValidatePluginResponse

stefanvanburen:
In the case that you have only an individual plugin response, you don't need to check for conflicting files across plugins - you just need the newly added edition validation.

We have a use-case internally where we know that we have an individual plugin response, and don't need to check for conflicting files across the only plugin.

Happy to close this if we don't like this approach, though.


jhump (commented):
This validation would also catch the case where a single plugin response tried to define the same file more than once. So I think you still need it.

stefanvanburen:
@jhump 😖 you're right, I overlooked that. Good thought, going to close this out.

PR #2951, opened 2024-05-06
Add new ListRepositoriesUsage rpc and messages

gilwong00:
Part 1 to: GRW-1804


proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -259,8 +277,27 @@ message GetClusterUsageRequest {
...
+
+message GetAllOrganizationsUsagesRequest {
+  // The timestamp of the beginning time of a date range.
+  google.protobuf.Timestamp start_time = 1;
+  // The timestamp of the ending time of a date range.
+  google.protobuf.Timestamp end_time = 2;
+}
+
+message GetAllOrganizationsUsagesResponse {
+  repeated RepositoryUsage repository_usages = 1;
```
bufdev, line 302: What is repeated here? What is unique and what is not? All needs to be documented

fyockm: // The list of types usage for all repositories in all orgs on a single BSR instance.

fyockm (commented):
generally, i think we should replace all instance of "amount" with "number"

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -87,6 +87,21 @@ message MonthlyUsage {
...
+  // The id of the repository.
+  string repository_id = 1;
+  // The name of the repository.
+  string repository_name = 2;
+  // The id of the repository owner.
+  string owner_id = 3;
+  // // The name of the respository owner.
+  string owner_name = 4;
+  // The total number of message, enum and method types usage for the repository.
+  MonthlyUsage usage = 5;
```
cyinma, line 102: is this supposed to be monthly (i.e. having a specific month and year)?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -142,6 +157,9 @@ service AdminService {
   // GetClusterUsage returns the summation of total message, enum and service types usage
   // for every repository in each organization within a single tenant BSR instance.
   rpc GetClusterUsage(GetClusterUsageRequest) returns (GetClusterUsageResponse);
+  // GetAllOrganizationsUsages returns the usage metrics for all repositories for every organization
+  // within a single tenant BSR instance for a given time range.
+  rpc GetAllOrganizationsUsages(GetAllOrganizationsUsagesRequest) returns (GetAllOrganizationsUsagesResponse);
```
cyinma, line 162: `All` feels like unnecessary for me, maybe something to mention in comment instead? also is this just organization instead of all owners?

gilwong00: Its organization and owners. Maybe this can be called something `GetRepositoryUsages` since this just returns all the repositories and we do the groupings on the FE. wdyt?

proto/buf/alpha/registry/v1alpha1/admin.proto
cyinma: then I will suggest to flatten this to just include what you need

gilwong00: ok, replaced `MonthlyUsage` with 
```
  // The total amount of message types for the month and year.
  uint32 total_message_types = 5;
  // The total amount of enum types for the month and year.
  uint32 total_enum_types = 6;
  // The total amount of method types for the month and year.
  uint32 total_method_types = 7;
  ```

File: proto/buf/alpha/registry/v1alpha1/admin.proto (hunk 3 above)
Comment (gilwong00) [reply proto/buf/alpha/registry/v1alpha1/admin.proto:162]: spoke offline, going to rename the rpc to `ListRepositoriesUsage`

File: proto/buf/alpha/registry/v1alpha1/admin.proto [hunk 4]
```diff
@@ -259,8 +281,27 @@ message GetClusterUsageRequest {
   google.protobuf.Timestamp start_time = 1;
   // The timestamp of the ending time of a date range.
   google.protobuf.Timestamp end_time = 2;
+  // Association id is an optional owner or repository id used to filter usage.
+  oneof association_id {
```
gilwong00, line 285: @cyinma I renamed this to `association_id` instead of entity from our convo. Might not be the best term so open to any suggestions here.

fyockm: are there cases that we'd want both org id _AND_ repository id?

proto/buf/alpha/registry/v1alpha1/admin.proto
bufdev: This doesn't answer the original question - is this monthly? What does the list contain? What should I expect a valid list to include based on the timestamp?

proto/buf/alpha/registry/v1alpha1/admin.proto
gilwong00: Given the design of the chart I don't think were would be a situation where we would want both

gilwong00: renamed to `resource_id` instead

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -87,6 +87,25 @@ message MonthlyUsage {
   uint32 total_method_types = 5;
 }
 
+// RepositoryUsage contains the total amount of message, enum and method types
+// for an repository within an organization.
```
cyinma, line 91:
```suggestion
// for an repository.
```
there is no grouping by owner etc, and its not for org only, so this seems confusing

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -87,6 +87,25 @@ message MonthlyUsage {
...
+message RepositoryUsage {
+  // The id of the repository.
+  string repository_id = 1;
+  // The name of the repository.
+  string repository_name = 2;
+  // The id of the repository owner.
+  string owner_id = 3;
+  // // The name of the respository owner.
+  string owner_name = 4;
+  // The total amount of message types for the month and year.
```
cyinma, line 101:
```suggestion
  // The total amount of message types.
```
I don't think there is any month/year relationship here

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -142,6 +161,9 @@ service AdminService {
   // GetClusterUsage returns the summation of total message, enum and service types usage
   // for every repository in each organization within a single tenant BSR instance.
   rpc GetClusterUsage(GetClusterUsageRequest) returns (GetClusterUsageResponse);
+  // ListRepositoriesUsage returns the usage metrics for all repositories for every organization and owner
```
cyinma, line 164:
```suggestion
  // ListRepositoriesUsage returns the usage metrics for all repositories
```
doesn't seems necessary to me

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -142,6 +161,9 @@ service AdminService {
...
+  // within a single tenant BSR instance for a given time range.
```
cyinma, line 165: is it really single tenant only?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -259,8 +281,28 @@ message GetClusterUsageRequest {
   google.protobuf.Timestamp start_time = 1;
   // The timestamp of the ending time of a date range.
   google.protobuf.Timestamp end_time = 2;
+  // Resource id is an optional owner or repository id used to filter usage.
+  oneof resource_id {
```
cyinma, line 285: while I don't have a better suggestion, we do use "resource" to refer to "repository or plugin" in [our API](https://github.com/bufbuild/buf/blob/main/proto/buf/alpha/registry/v1alpha1/resource.proto), which is different with the meaning here

proto/buf/alpha/registry/v1alpha1/admin.proto
gilwong00: yea this is single tenant only

proto/buf/alpha/registry/v1alpha1/admin.proto
gilwong00: right good catch, removed

proto/buf/alpha/registry/v1alpha1/admin.proto
gilwong00: thanks for calling that out, i named this to `source_id` instead

proto/buf/alpha/registry/v1alpha1/admin.proto
fyockm: `// The list contains types usage info and counts for all repositories in all orgs on a single BSR instance, based on a snapshot single point in time within the requested time range where the aggregate types usage is at a maximum.`

does this make more sense?


> is this monthly? 

no, it's for any point within the requested date range

> What does the list contain? 

repository information and types usage for all repositories on the instance

> What should I expect a valid list to include based on the timestamp?

the types count for the point when types usage was at a maximum in the time range

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -142,6 +161,9 @@ service AdminService {
   // GetClusterUsage returns the summation of total message, enum and service types usage
   // for every repository in each organization within a single tenant BSR instance.
   rpc GetClusterUsage(GetClusterUsageRequest) returns (GetClusterUsageResponse);
+  // ListRepositoriesUsage returns the usage metrics for all repositories within
+  // a single tenant BSR instance for a given time range.
```
cyinma, line 165: I think this will be the "highest point in time usage" or something similar

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -259,8 +281,29 @@ message GetClusterUsageRequest {
   google.protobuf.Timestamp start_time = 1;
   // The timestamp of the ending time of a date range.
   google.protobuf.Timestamp end_time = 2;
+  // Source id is an optional owner or repository id used to filter usage.
+  oneof source_id {
```
cyinma, line 285: how about `usage_filter`?

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -259,8 +281,29 @@ message GetClusterUsageRequest {
...
+message ListRepositoriesUsageRequest {
+  // The timestamp of the beginning time of a date range.
+  google.protobuf.Timestamp start_time = 1;
+  // The timestamp of the ending time of a date range.
+  google.protobuf.Timestamp end_time = 2;
+}
+
+message ListRepositoriesUsageResponse {
+  // The list contains types usage info and counts for all repositories in all
+  // orgs on a single BSR instance, based on a snapshot single point in time within
```
cyinma, line 306:
```suggestion
  // owners on a single BSR instance, based on a snapshot single point in time within
```

cyinma (commented):
please also update the PR title

cyinma (approved):
lgtm, please make sure to update the PR title

fyockm (approved):

proto/buf/alpha/registry/v1alpha1/admin.proto
```diff
@@ -259,8 +281,29 @@ message GetClusterUsageRequest {
   google.protobuf.Timestamp start_time = 1;
   // The timestamp of the ending time of a date range.
   google.protobuf.Timestamp end_time = 2;
+  // Usage filter is an optional owner or repository id used to filter usage.
+  oneof usage_filter {
+    // The id of the owner.
+    string owner_id = 3;
+    // The id of the repository.
+    string repository_id = 4;
+  }
```
nicksnyder, line 290: Can you explain why this is a oneof?

proto/buf/alpha/registry/v1alpha1/admin.proto
nicksnyder: Between the documentation of this method, and the documentation/name for RepositoryUsage, some things are very unclear
* How many items will be in this list? Do we need to have paging?
* What is the timeframe associated with each item? Is this configurable? Should it be?
* How do I map items in this list to a particular timeframe?

Other thought:
* If this method is really specialized toward returning max usage within some time horizon, maybe the names should more explicitly represent that.

nicksnyder (changes requested):
Generally we want our internal APIs to be as clearly named, designed, and documented as our external APIs. Please read through https://github.com/bufbuild/registry-proto to get a sense of what a good standard looks like moving forward, and to understand the patterns we have established (e.g. like how we do pagination). It would also be good to add validation rules to all fields as appropriate (lmk if there is a reason we haven't been doing that already).

PR #2953, opened 2024-05-07
Add `--git-metadata` flag to `buf push`

doriable:
This PR adds the `--git-metadata` flag to `buf push`:

- this flag is a "meta-flag" that uses Git source control state to help set `--label` flag(s), `--source-control-url`, and `--create-default-label` to the HEAD branch of the default Git remote
- this flag is only compatible with checkouts of Git source repositories
- this flag does not allow you to set `--source-control-url`, `--create-default-label`, `--label`, `--tag`, `--branch`, or `--draft` labels alongside

This PR also changes the default visibility for `--create-visibility` to "private". This means that users are no longer required to specify the `--create-visibility` flag when calling `buf push --create` -- it will default to creating a private repository if one does not already exist.

Also, this PR includes a changelog entry for all the changes above, as well as the `--label`, `--create-default-label`, and `--source-control-url` flags.


private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+
+func getGitMetadataUploadOptions(
+	ctx context.Context,
+	container appext.Container,
+	flags *flags,
+) ([]bufmodule.UploadOption, error) {
+	input, err := bufcli.GetInputValue(container, flags.InputHashtag, ".")
+	if err != nil {
+		return nil, err
+	}
```
doriable, line 440: Attempting to figure out a graceful way of surfacing an error if this is not a checkout of a Git repository (e.g. a source ref).

doriable: We should validate for unchecked references here, so we can also use this to handle a `128`.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+		// Bitbucket is the only URL that uses the "/commits" route, both Github and Gitlab
+		// use "/commit", so we default to that if the remote URL does not point at Bitbucket.
+		if strings.HasPrefix(gitDefaultRemoteURL, bitBucketURL) {
+			return bufmodule.UploadWithSourceControlURL(fmt.Sprintf(
+				bitBucketRemoteURLFormat,
+				gitDefaultRemoteURL,
+				currentHEADCommit,
+			)), nil
+		}
+		return bufmodule.UploadWithSourceControlURL(fmt.Sprintf(
```
pkwarren, line 601: Won't this break if someone is using a git remote instead of a https one? We should properly handle rewriting origins like `git@github.com:git/git.git` into appropriate URLs too.

Also I think we need to validate that the remote matches well known formats for GitHub/GitLab and if it doesn't match don't build a URL.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+			defaultGitRemoteURLFormat,
+			gitDefaultRemoteURL,
+			currentHEADCommit,
+		)), nil
+	}
+	// Otherwise we sort by alphabetical order and return the first URL
```
pkwarren, line 607: This is only getting the URL to the remote - not building the full URL to the commit like above. I also think we should allow the user some control over which origin is used instead of returning the first sorted one or origin.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+			ctx,
+			gitCommand,
+			command.RunWithArgs("config", "--get", fmt.Sprintf("remote.%s.url", remote)),
+			command.RunWithStdout(buffer),
+			command.RunWithStderr(buffer),
+			command.RunWithDir(input),
+		); err != nil {
+			return nil, err
+		}
+		remoteToURL[remote] = strings.TrimSpace(buffer.String())
```
pkwarren, line 581: Just to check, I ran `git remote set-url origin ""` and it worked somehow. We might want to only add to the map if it is non-empty here.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+		return remoteToHEADBranch[sortedRemotes[0]], nil
+	}
+	return "", nil
+}
+
+// This checks for the HEAD branch from the output of `git remote show <remote>`.
+// sed -n '/HEAD branch/s/.*: //p'
+func getHEADBranchFromGitRemoteOutput(output []byte) (string, error) {
+	// First find where "HEAD branch"
+	index := bytes.Index(output, []byte("HEAD branch: "))
```
pkwarren, line 677: Instead of dealing in bytes, I think this would by much easier to read/maintain with a bufio.Scanner, strings.Cut, etc.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+		if err != nil {
+			return "", err
+		}
+		remoteToHEADBranch[remote] = branch
+	}
+	// We prioritze the Git default remote, "origin", HEAD branch as the default branch
+	if gitDefaultRemoteHEADBranch, ok := remoteToHEADBranch[gitDefaultRemote]; ok {
+		return gitDefaultRemoteHEADBranch, nil
+	}
+	// Otherwise we sort by alphabetical order and return the first URL
```
pkwarren, line 665:
```suggestion
	// Otherwise we sort by alphabetical order and return the first origin's branch
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
 	return nil
 }
 
+// We do not allow users to set --source-control-url, --create-default-label, and --label
+// flags if the --git-metadata flag is set.
+func validateGitMetadataFlags(flags *flags) error {
+	if flags.GitMetadata {
+		var usedFlags []string
+		if flags.SourceControlURL != "" {
```
pkwarren, line 396: Not blocking but I feel like if we couldn't detect a source code URL, we should allow the user to specify one (or fail push with an error). Otherwise they might not realize that git-metadata is failing to discover information.

Alternatively, we could warn if we fail to detect a URL or a HEAD branch (if --create is specified).

Additionally, labels should be additive - so we'll add labels by default for git metadata but it may be that users want to specify additional labels on commits and we should just take the intersection of them both.

private/pkg/git/git.go
```diff
@@ -135,3 +139,22 @@ type ListFilesAndUnstagedFilesOptions struct {
 	// is being applied to.
 	IgnorePathRegexps []*regexp.Regexp
 }
+
+// GetDefaultBranch returns the repository's default branch name. It attempts to read it from the
+// `.git/refs/remotes/origin/HEAD` file, and expects it to be pointing to a branch also in the
+// `origin` remote.
+func GetDefaultBranch(gitDirPath string) (string, error) {
```
pkwarren, line 146: This appears unused?

pkwarren (commented):
Overall looks good. I think we could benefit from some tests on the git logic (we could easily run `git init` in a temp dir and test various scenarios).

private/buf/cmd/buf/command/push/push.go
```diff
@@ -132,6 +144,33 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
...
+		gitMetadataFlagName,
+		false,
+		fmt.Sprintf(
+			`Uses the Git source control state to set flag values. If this flag is set, we will use the following values for your flags:
+
+	--%s to <git remote URL>/<repository name>/commit/<commit sha> (e.g. https://github.com/acme/weather/commit/ffac537e6cbbf934b08745a378932722df287a53)
+	--%s for each Git tag and branch for this Git commit
+	--%s to the Git default branch (e.g. main) - this is only in effect if --%s is also set
+
+The source control URL and default branch is based on the default Git remote, %q. If you do not have this remote, we take the URL for the first remote based on alphabetical ordering.
```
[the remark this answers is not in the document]
saquibmian: I've worked at places where `origin` was not the default remote. I would instead suggest that we take any remote, provided it's the only one. If there is more than one remote, require the user to specify one? I don't think we should arbitrarily pick one, though. Either there's one, or origin, or fail.

Question: how do fork builds work? If someone forks bufbuild/buf and sends a PR, and we have a GitHub action that runs `buf push --git-metadata` for the PR, what happens?

private/buf/cmd/buf/command/push/push.go
saquibmian: > if we couldn't detect a source code URL, we should allow the user to specify one

I think we should fail if we can't detect a source control URL or default branch.

> we should just take the intersection of them both

We discussed this and decided that, for now, it's one or the other. `--git-metadata` opines that Git is the source of truth for labels. If there are more labels to associate, associate them in Git and we'll push them. If you need something outside of that, you can't use `--git-metadata`.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+	input string,
+) (bufmodule.UploadOption, error) {
+	tags, err := getGitTagsOnCurrentCommit(ctx, runner, input)
+	if err != nil {
+		return nil, err
+	}
+	branch, err := getCurrentGitBranch(ctx, runner, input)
+	if err != nil {
+		return nil, err
+	}
```
saquibmian, line 513: Interesting, I would expect this to be the set of all branches that point to this commit, not just the current branch. Wouldn't that get us closer to what is in Git, and also pushes on other branches no-ops due to squashing? Otherwise a single git commit that is on multiple branches will not have a corresponding commit in the BSR with multiple branches.

I can see the value in being context-centric though -- this isn't `sync`, after all. What do we think?

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+			command.RunWithDir(input),
+		); err != nil {
+			return nil, err
+		}
+		remoteToURL[remote] = strings.TrimSpace(buffer.String())
+	}
+	// We prioritize the Git default remote, "origin", URL
+	if gitDefaultRemoteURL, ok := remoteToURL[gitDefaultRemote]; ok {
+		// First we need to trim the `.git` suffix if there is one.
+		gitDefaultRemoteURL = strings.TrimSuffix(gitDefaultRemoteURL, ".git")
```
saquibmian, line 586: I think we need to parse this URL. We will likely want to special-case based on hostname. This URL can be in a few different formats, see here: https://git-scm.com/docs/git-clone#_git_urls

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+) (bufmodule.UploadOption, error) {
+	tags, err := getGitTagsOnCurrentCommit(ctx, runner, input)
+	if err != nil {
+		return nil, err
+	}
+	branch, err := getCurrentGitBranch(ctx, runner, input)
+	if err != nil {
+		return nil, err
+	}
+	labels := append(tags, branch)
```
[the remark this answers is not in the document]
saquibmian: I was just about to say, this may be empty. Additionally, what would happen if CI runs on a commit that is in a detached HEAD such that `git branch --show-current` yields no results, and there are no tags associated with the commit? I suspect you will end up with no labels, which will effectively push to the default label?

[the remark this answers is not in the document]
pkwarren: That seems dangerous as a default - perhaps we require that a branch is discovered?

private/buf/bufcli/flags_args.go
```diff
@@ -148,8 +148,8 @@ func BindCreateVisibility(flagSet *pflag.FlagSet, addr *string, flagName string,
 	flagSet.StringVar(
 		addr,
 		flagName,
-		"",
-		fmt.Sprintf(`The repository's visibility setting, if created. Must be set if --%s is set. Must be one of %s`, createFlagName, stringutil.SliceToString(allVisibiltyStrings)),
+		privateVisibility,
+		fmt.Sprintf(`The repository's visibility setting, if created. Can only be set if --%s is set. Defaults to %s if not set with --%s. Must be one of %s`, createFlagName, createFlagName, privateVisibility, stringutil.SliceToString(allVisibiltyStrings)),
```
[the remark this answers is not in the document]
doriable: Ah yes, and I need to swap the two. Sorry, I rewrote this message a few times ^^". It should be:

```
The repository's visibility setting, if created. Can only be set if --create is set. Defaults to private if this flag is not set with --create. Must be one of [public,private] (default "private")
```

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
doriable: Yeah, so we had a conversation about this, and we will be erroring if `origin` doesn't exist, and printing a warning if the checkout has multiple remotes.

private/pkg/git/git.go
doriable: Oh! I pulled this in and no longer needed it, and didn't realise i didn't clear this file before committing. Yes, that is correct, will be deleting.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -132,6 +144,33 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
...
+			labelFlagName,
+			createDefaultLabelFlagName,
+			createFlagName,
+			gitDefaultRemote,
+			sourceControlURLFlagName,
+			labelFlagName,
+			createDefaultLabelFlagName,
+			tagFlagName,
+			branchFlagName,
+			draftFlagName,
```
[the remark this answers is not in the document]
doriable: Yeah, I went back and forth on this one. I think you are right, given the flags are hidden, we should not expose them, but still check for them.

CHANGELOG.md
```diff
@@ -2,7 +2,18 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Change the default visibiity of `--create-visibility` flag for `buf push` to `private`
```
unmultimedio, line 5:
```suggestion
- Change the default visibility of `--create-visibility` flag for `buf push` to `private`
```

private/buf/cmd/buf/command/push/push.go
doriable: Ah yes, it's a lot better that way!

private/buf/cmd/buf/command/push/push.go
doriable: > Additionally, labels should be additive - so we'll add labels by default for git metadata but it may be that users want to specify additional labels on commits and we should just take the intersection of them both.

Yes, I had a conversation with folks this morning, and we're definitely allowing users to add `--label`s to this.

For now, we are disallowing `--source-control-url` and `--create-default-label`, but this may change.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -132,6 +144,33 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
...
 		"The URL for viewing the source code of the pushed modules (e.g. the specific commit in source control).",
 	)
+	flagSet.BoolVar(
+		&f.GitMetadata,
+		gitMetadataFlagName,
+		false,
+		fmt.Sprintf(
+			`Uses the Git source control state to set flag values. If this flag is set, we will use the following values for your flags:
+
+	--%s to <git remote URL>/<repository name>/commit/<commit sha> (e.g. https://github.com/acme/weather/commit/ffac537e6cbbf934b08745a378932722df287a53)
```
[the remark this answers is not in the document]
doriable: > Technically, the URL is dependent on the VCS type (bitbucket is commits).

Yeah, that is handled in the code, I can specify that it's `<route>` in the template.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +388,306 @@ func validateLabelFlagValues(flags *flags) error {
...
+	remotes []string,
+	input string,
+) (bufmodule.UploadOption, error) {
+	remoteToURL := map[string]string{}
+	for _, remote := range remotes {
+		buffer := bytes.NewBuffer(nil)
+		if err := runner.Run(
+			ctx,
+			gitCommand,
+			command.RunWithArgs("config", "--get", fmt.Sprintf("remote.%s.url", remote)),
```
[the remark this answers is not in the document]
doriable: Yeah -- this is something I went back and forth on. And these values may be different. You could have `git config --get origin.remote.url` return the SSH vs. HTTPS. This is related to @saquibmian's comment below and also your other comments on empty URL's and also the SSH URL. Let me collate the comments below and I'll propose something.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
doriable: I don't think that's necessarily dangerous -- I think that the implication should be that if you have a detached `HEAD` checkout, you would be pushing to default label, and if that is a problem, then your default label should be protected by policy checks. It's unlikely that if your CI that does this for `buf push` doesn't also have similar workflows for other things -- therefore it must  be safe to operate against detached checkouts, so I think this is reasonable default behaviour. A warning for empty labels is reasonable though, I think.

private/buf/cmd/buf/command/push/push.go
doriable: Just updating the comment here based on the conversations we've had external to this PR: it would be all branches that point directly to the commit at the time of the command.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
doriable: Consolidating URL conversation here: so I think we should make a best effort on URL. The approach I took is:
- check that the URL has a valid scheme, if not, print a warning and set no source control URL
- if it has a valid scheme, parse the hostname. if the hostname is one of known hostnames, then we construct a `https` url for the user with the commit sha. otherwise we return just the URL to the source control repository, since there is no way to ascertain the route to the specific commit.

private/buf/cmd/buf/command/push/push.go
doriable: Yeah, that's actually a good point, sorry, missed this earlier! Updating.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+	// url.Path already has a "/" prefix.
+	if strings.Contains(url.Hostname(), bitBucketHostname) {
+		return fmt.Sprintf(bitBucketRemoteURLFormat, scheme, url.Hostname(), strings.TrimSuffix(url.Path, ".git"), commitSHA), nil
+	}
+	// If the hostname contains "github" (e.g. github.foo.com or github.com) or "gitlab"
+	// (e.g. gitlab.foo.com or gitlab.com), we use the route "commit" to construct a source
+	// control URL to the commit.
+	// url.Path already has a "/" prefix.
+	if strings.Contains(url.Hostname(), githubHostname) || strings.Contains(url.Hostname(), gitlabHostname) {
+		return fmt.Sprintf(defaultGitRemoteURLFormat, scheme, url.Hostname(), strings.TrimSuffix(url.Path, ".git"), commitSHA), nil
```
saquibmian, line 722: Same idea here, there may be a non-standard port here.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+			return "", err
+		}
+		branch, err := getHEADBranchFromGitRemoteOutput(buffer)
+		if err != nil {
+			return "", err
+		}
+		remoteToHEADBranch[remote] = branch
+	}
+	// We always use the Git remote, "origin", to check for the `HEAD` branch as the
+	// "default" branch.
```
saquibmian, line 776: Why fetch the others, then?

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+		remotes = append(remotes, remote)
+	}
+	if !gitOriginRemoteFound {
+		return nil, fmt.Errorf("Git remote %q not found", gitOriginRemote)
+	}
+	if len(remotes) > 1 {
+		logger.Warn(
+			fmt.Sprintf("More than one Git remote found, %q will be used for Git metadata: %v", gitOriginRemote, remotes),
+		)
+	}
```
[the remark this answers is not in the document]
doriable: So the decision was that we wouldn't fail if there is more than one remote (basically exactly because of the forked upstreams situation -- you may get additional remotes from forks), but we do fail if the remote we care about, `origin` isn't present, and we basically only care about `origin`.

I think if we are sticking to this, then the only relevant remote is only ever `origin`, so we may not even need to return any here and simply use this to validate remotes instead.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+	commitSHA string,
+	logger *zap.Logger,
+) (string, error) {
+	url, err := url.Parse(rawSourceControlURL)
+	if err != nil {
+		return "", err
+	}
+	// Docs: https://git-scm.com/docs/git-clone#_git_urls
+	// Validate URL scheme
+	if !slices.Contains([]string{"ssh", "git", "http", "https", "ftp", "ftps"}, url.Scheme) {
```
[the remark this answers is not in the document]
doriable: So I think the issue here is that the current checkout may be using `ssh`, even though `https` is also available, so as long as the scheme conforms to valid Git schemes, I think we should accept it. And then you make an attempt to point at a valid URL based on the hostname.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+	// We default to https unless the scheme is specfically "http"
+	scheme := "https"
+	if url.Scheme == "http" {
+		scheme = "http"
+	}
+	// If the hostname contains "bitbucket" (e.g. bitbucket.foo.com or bitbucket.org), we
+	// use the Bitbucket route "commits" to construct a source control URL to the commit
+	// url.Path already has a "/" prefix.
+	if strings.Contains(url.Hostname(), bitBucketHostname) {
+		return fmt.Sprintf(bitBucketRemoteURLFormat, scheme, url.Hostname(), strings.TrimSuffix(url.Path, ".git"), commitSHA), nil
```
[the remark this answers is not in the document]
doriable: I'm not sure how the port factors in here ^^" We're making an attempt at constructing a `https` url based on the hostname, which drops the port...

But yes, I can add some unit tests for this.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
saquibmian: > may not even need to return any here and simply use this to validate remotes instead

Did you continue to return all remotes in the event that we revert this decision? If so, maybe a comment to that effect would be helpful -- a number of my comments were about this.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
saquibmian: How do we know what the URL of the UI is? It could be on a port? With self-hosted it's hard to know.

private/buf/cmd/buf/command/push/push.go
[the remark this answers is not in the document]
doriable: It wasn't so much in case we revert so much as things were a bit in-flight as code was being written. I think if that's the UX we're going for, it's totally fine to have the code more closely reflect that :) So i'm onboard with making these changes.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+		command.RunWithDir(input),
+	); err != nil {
+		return nil, err
+	}
+
+	modifiedFiles = append(modifiedFiles, getAllTrimmedLinesFromBuffer(stdout)...)
+	return modifiedFiles, nil
+}
+
+func getGitRemotes(
```
bufdev, line 529: Same

bufdev, line 529: Why are we returning multiple remotes? There's comments on this below as well

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+		return nil, err
+	}
+	runner := command.NewRunner()
+	uncommittedFiles, err := checkForUncommittedGitChanges(ctx, runner, input)
+	if err != nil {
+		// We surface additional information here for the user to ensure they are using this flag
+		// with a git checkout.
+		return nil, fmt.Errorf("unable to check input %q, please ensure this is a Git repository checkout: %w", input, err)
+	}
+	if len(uncommittedFiles) > 0 {
```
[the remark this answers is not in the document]
doriable: This has always been a part of the product spec -- we do not allow users to push unchecked/uncommitted references. This makes sense, since we are tagging this with Git commit information through the VCS -- if there are uncommitted changes being pushed, that commit information would not make sense.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +394,415 @@ func validateLabelFlagValues(flags *flags) error {
...
+		return nil, err
+	}
+	if branch != "" {
+		// We do not want to add an empty string to the list of labels
+		labels = append(labels, branch)
+	}
+	if len(labels) > 0 {
+		return bufmodule.UploadWithLabels(labels...), nil
+	}
+	logger.Warn("No tags or branch found for HEAD, contents will be pushed to the default label")
```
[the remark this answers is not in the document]
doriable: Based on today's conversation, we will now be erroring if no refs are found, so going to resolve this :)

doriable:
Resolved old conversations around URL parsing and erroring behaviours (based on today's conversation), refactored git commands for metadata to `pkg/git`, and added some tests around source control url parsing.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +384,158 @@ func validateLabelFlagValues(flags *flags) error {
...
+	env map[string]string,
+	input string,
+	gitCommitSha string,
+) (bufmodule.UploadOption, error) {
+	refs, err := git.GetRefsForGitCommitAndRemote(ctx, runner, env, input, gitOriginRemote, gitCommitSha)
+	if err != nil {
+		return nil, err
+	}
+	if len(refs) == 0 {
+		return nil, fmt.Errorf("not tags or branch found for HEAD, %s", gitCommitSha)
```
saquibmian, line 518:
```suggestion
		return nil, fmt.Errorf("not tags or branches found for HEAD, %s", gitCommitSha)
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +384,158 @@ func validateLabelFlagValues(flags *flags) error {
...
+	runner command.Runner,
+	input string,
+	gitCommitSha string,
+) (bufmodule.UploadOption, error) {
+	sourceControlURL, err := git.GetSourceControlURL(ctx, runner, input, gitOriginRemote, gitCommitSha)
+	if err != nil {
+		return nil, err
+	}
+	if sourceControlURL == "" {
+		return nil, fmt.Errorf("no valid source control URL parsed for git commit sha: %s", gitCommitSha)
```
saquibmian, line 534: > unable to determine source control URL for this repository; only GitHub/GitLab/BitBucket are supported

CHANGELOG.md
```diff
@@ -58,6 +58,20 @@
   `features.field_presence = LEGACY_REQUIRED` in Editions files. This new rule is not active by
   default in existing `v1` and `v1beta1` configurations, for backwards-compatibility reasons.
   Migrate your config to `v2` to use them.
+- Add `buf beta registry label archive` and `buf beta registry label unarchive` commands
```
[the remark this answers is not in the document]
doriable: Oh sure, yeah -- I just adjusted the changelog in the chang I'm making for those commands, but yeah.

private/pkg/git/url.go
```diff
@@ -0,0 +1,111 @@
...
+	gitlabHostname              = "gitlab"
+)
+
+// GetSourceControlURL gets the parses a user-facing source control URL based on the URL
+// from git config based on the given remote, git commit sha, and directory.
+//
+// It checks the hostname of repository remote URL -- if it contains a known hostname,
+// github, gitlab, and bitbucket, we construct a URL. Otherwise, we return an empty source
+// control URL.
+func GetSourceControlURL(
```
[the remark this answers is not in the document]
doriable: So I waffled a little on this one -- but since these functions are not encapsulated by an interface, I ended up not adding it to `git.go` because I didn't want to conflate the two styles... I can move them to `git.go` if they are confusing.

private/pkg/git/url.go
```diff
@@ -0,0 +1,111 @@
...
+// If the hostname contains bitbucket (e.g. bitbucket.mycompany.com or bitbucket.org),
+// then it uses the route /commits for the git commit sha.
+// If the hostname contains github (e.g. github.mycompany.com or github.com) or gitlab
+// (e.g. gitlab.mycompany.com or gitlab.com) then it uses the route /commit for the git
+// commit sha.
+func parseSourceControlURL(
+	rawRepositoryURL string,
+	gitCommitSha string,
+) (string, error) {
+	repositoryURL, err := url.Parse(rawRepositoryURL)
```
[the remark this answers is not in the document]
doriable: Yeah -- this is reasonable. I'll implement some custom handling for the scp-like URLs.

private/pkg/git/ur_test.go
```diff
@@ -0,0 +1,62 @@
...
+	gitCommitSha := "007bdc4ddc7e4854b4bf3ff6c1f61eda"
+	t.Run("ssh, bitbucket", func(t *testing.T) {
+		t.Parallel()
+		sourceControlURL, err := parseSourceControlURL("ssh://user@bitbucket.org:1234/user/repo.git", gitCommitSha)
+		require.NoError(t, err)
+		require.Equal(t, "https://bitbucket.org/user/repo/commits/007bdc4ddc7e4854b4bf3ff6c1f61eda", sourceControlURL)
+	})
+	t.Run("ssh, github", func(t *testing.T) {
+		t.Parallel()
+		sourceControlURL, err := parseSourceControlURL("ssh://git@github.com/user/repo.git", gitCommitSha)
```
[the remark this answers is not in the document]
doriable: That is what they show on the UI, but you can also clone in this format. But either way, I can add handling for the scp-style URL and test it here.

private/pkg/git/url.go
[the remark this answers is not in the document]
doriable: I like this -- this also works well with the style of the this package is in, so let me move some stuff around and implement the scp-style URL handling logic alongside.

private/pkg/git/git.go
```diff
@@ -135,3 +135,46 @@ type ListFilesAndUnstagedFilesOptions struct {
...
+
+// Remote represents a Git remote and provides associated metadata.
+type Remote interface {
+	// Name of the remote (e.g. "origin")
+	Name() string
+	// HEADBranch is the name of the HEAD branch of the remote.
+	HEADBranch() string
+	// Kind of remote (e.g. GitHub, GitLab, BitBucket, etc.).
+	Kind() RemoteKind
+	// Hostname is the host name parsed from the remote URL. If the remote is not an unknown
```
saquibmian, line 161:
```suggestion
	// Hostname is the host name parsed from the remote URL. If the remote is an unknown
```

private/pkg/git/remote.go
```diff
@@ -0,0 +1,231 @@
...
+
+const (
+	bitBucketHostname = "bitbucket"
+	githubHostname    = "github"
+	gitlabHostname    = "gitlab"
+	gitSuffix         = ".git"
+)
+
+var (
+	ErrRemoteNotFound = errors.New("not found")
```
saquibmian, line 39:
```suggestion
	ErrRemoteNotFound = errors.New("git remote not found")
```
In case the error bubbles up unhandled

private/pkg/git/remote.go
```diff
@@ -0,0 +1,231 @@
...
+)
+
+type remote struct {
+	name           string
+	kind           RemoteKind
+	hostname       string
+	repositoryPath string
+	headBranch     string
+}
```
saquibmian, line 48: There should be a `newRemote` right after this, rather than constructing the remote in various places.

private/pkg/git/remote.go
```diff
@@ -0,0 +1,231 @@
...
+		return &url.URL{
+			Host: match[2],
+			Path: "/" + match[3], // scp-like URLs do not have the leading slash
+		}
+	}
+	return nil
+}
+
+// getRemoteHEADBranch returns the HEAD branch based on the given remote and given
+// directory. Querying the remote for the HEAD branch requires the passing the
```
saquibmian, line 188:
```suggestion
// directory. Querying the remote for the HEAD branch requires passing the
```

saquibmian (commented):
Looks good, just a couple stylistic things.

CHANGELOG.md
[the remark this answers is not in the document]
unmultimedio: Unresolving, this item was duped, we need to remove it from here too.

private/buf/bufcli/flags_args.go
```diff
@@ -148,8 +148,8 @@ func BindCreateVisibility(flagSet *pflag.FlagSet, addr *string, flagName string,
 	flagSet.StringVar(
 		addr,
 		flagName,
-		"",
-		fmt.Sprintf(`The repository's visibility setting, if created. Must be set if --%s is set. Must be one of %s`, createFlagName, stringutil.SliceToString(allVisibiltyStrings)),
+		privateVisibility,
+		fmt.Sprintf(`The repository's visibility setting, if created. Can only be set if --%s is set. Defaults to %s if this flag is not set with --%s. Must be one of %s`, createFlagName, privateVisibility, createFlagName, stringutil.SliceToString(allVisibiltyStrings)),
```
pkwarren, line 152:
```suggestion
		fmt.Sprintf(`The repository's visibility setting, if created. Can only be set with --%s. Must be one of %s`, createFlagName, stringutil.SliceToString(allVisibiltyStrings)),
```

This will replace:
```
The repository's visibility setting, if created. Can only be set if --create is set. Defaults to private if this flag is not set with --create. Must be one of [public,private] (default "private")
```
with:
```
The repository's visibility setting, if created. Can only be set with --create. Must be one of [public,private] (default "private")
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -132,6 +142,29 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
...
+		gitMetadataFlagName,
+		false,
+		fmt.Sprintf(
+			`Uses the Git source control state to set flag values. If this flag is set, we will use the following values for your flags:
+
+	--%s to <git remote URL>/<repository name>/<route>/<checked out commit sha> (e.g. https://github.com/acme/weather/commit/ffac537e6cbbf934b08745a378932722df287a53)
+	--%s for each Git tag and branch pointing to the currently checked out commit
+	--%s to the Git default branch (e.g. main) - this is only in effect if --%s is also set
+
+The source control URL and default branch is based on the Git remote, %q, and requires you to have this remote.
```
pkwarren, line 156:
```suggestion
The source control URL and default branch is based on the required Git remote %q.
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -132,6 +142,29 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
...
+This flag is only compatible with checkouts of Git source repositories.
+This flag does not allow you to set any of the following flags yourself: --%s, --%s.`,
```
pkwarren, line 158: Do we need to mention anything about how any additional `--label` parameters will be added in addition to the default labels detected from Git?

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +386,155 @@ func validateLabelFlagValues(flags *flags) error {
...
+// getGitMetadataSourceControlURLUploadOption takes a remote and git commit sha and makes
+// the best effort to construct a user-facing URL based on the hostname.
+//
+// If the remote hostname contains bitbucket (e.g. bitbucket.mycompany.com or bitbucket.org),
+// then it uses the route /commits for the git commit sha.
+//
+// If the remote hostname contains github (e.g. github.mycompany.com or github.com) or gitlab
+// (e.g. gitlab.mycompany.com or gitlab.com) then it uses the route /commit for the git
+// commit sha.
+func getGitMetadataSourceControlURLUploadOption(
```
pkwarren, line 511: I think it would be good to move the building of source control URLs to the git package in order to keep the logic in here purely around upload. Perhaps a `SourceControlURL(commitSHA string)` on the `git.Remote` interface?

private/buf/cmd/buf/command/push/push.go
doriable: Hmm, it might not be a bad idea to do so -- the more information the better. I'll add something here for that :)

private/buf/cmd/buf/command/push/push.go
doriable: Hmm, yeah, I went back and forth a little bit on where specifically this logic should live, since it's kind of "business-y" (rather than "generically git"), but also, the `RemoteKind` type already breaks that a little bit. So given that, it might make sense to put there.

It is a little strange to have this function take `commitSHA`, but I think that is reasonable. I'll make it clear that it just accepts the `commitSHA` as a string but doesn't do any validation against that string.

doriable: Actually, in this case, we wouldn't need to expose `RemoteKind` at all. I think I'm going to unexport the enum and keep it since it is useful for keeping the parsing logic clean (and independent from the initial parsing of the URL).

private/pkg/git/remote.go
```diff
@@ -51,7 +67,7 @@ func (r *remote) Name() string {
 	return r.name
 }
 
-func (r *remote) Kind() RemoteKind {
+func (r *remote) Kind() remoteKind {
```
pkwarren, line 70: We can remove this method now if we want.

pkwarren (approved):

pkwarren (approved):

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +385,117 @@ func validateLabelFlagValues(flags *flags) error {
...
+	ctx context.Context,
+	container appext.Container,
+	flags *flags,
+) ([]bufmodule.UploadOption, error) {
+	input, err := bufcli.GetInputValue(container, flags.InputHashtag, ".")
+	if err != nil {
+		return nil, err
+	}
+	runner := command.NewRunner()
+	uncommittedFiles, err := git.CheckForUncommittedGitChanges(ctx, runner, input)
```
bufdev, line 436: To what I can see, there is no verification here that `input` in in fact a directory, however it is expected by `git.CheckForUncommitedGitChanges` that `input` is a path to a local directory. This validation needs to be done, and should be obvious in the context of this function.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +385,117 @@ func validateLabelFlagValues(flags *flags) error {
...
+	if err != nil {
+		// We surface additional information here for the user to ensure they are using this flag
+		// with a git checkout.
+		return nil, fmt.Errorf("unable to check input %q, please ensure this is a Git repository checkout: %w", input, err)
+	}
+	if len(uncommittedFiles) > 0 {
+		return nil, fmt.Errorf("uncommitted changes found in the following files: %v", uncommittedFiles)
+	}
+	originRemote, err := git.GetRemote(ctx, runner, app.EnvironMap(container), input, gitOriginRemote)
```
bufdev, line 445: Same comment re: input validation.

bufdev, line 445: Other functions in the `git` package (and others) take containers, this function should continue to follow this pattern by taking an `app.EnvContainer` instead of passing a `map[string]string` via `app.EnvironMap`.

private/pkg/git/remote.go
```diff
@@ -0,0 +1,282 @@
...
+	bitBucketHostname           = "bitbucket"
+	githubHostname              = "github"
+	gitlabHostname              = "gitlab"
+	gitSuffix                   = ".git"
+	githubGitlabRemoteURLFormat = "https://%s%s/commit/%s"
+	bitBucketRemoteURLFormat    = "https://%s%s/commits/%s"
+)
+
+var (
+	ErrRemoteNotFound = errors.New("git remote not found")
```
bufdev, line 41: No Godoc, and should be in `git.go` - `remote.go` should be limited to `remote` and supporting functions called by `remote` methods

private/pkg/git/remote.go
```diff
@@ -0,0 +1,282 @@
...
+)
+
+// remoteKind is the kind of remote based on Git source (e.g. GitHub, GitLab, BitBucket, etc.)
+type remoteKind int
+
+const (
```
bufdev, line 47: Global consts always come first, then global vars, then types.

private/pkg/git/metadata.go
```diff
@@ -0,0 +1,144 @@
...
+const (
+	gitCommand      = "git"
+	gitOriginRemote = "origin"
+	tagsPrefix      = "refs/tags/"
+	headsPrefix     = "refs/heads/"
+)
+
+// CheckForUncommittedGitChanges checks if there are any uncommitted and/or unchecked
+// changes from git based on the given directory.
+func CheckForUncommittedGitChanges(
```
bufdev, line 35: Why were these functions split out into a separate file than `git.go`, but `Remote` was not?

doriable: Oh yeah, I put this in a separate file initially because these didn't come with an additional interface abstraction. But I've moved them to `git.go`, since it does conform with the old Go package styles.

doriable: Oh yeah, I put this in a separate file initially because these didn't come with an additional interface abstraction. But I've moved them to `git.go`, since it does conform with the old Go package styles.

private/buf/cmd/buf/command/push/push.go
doriable: Added validation here.
It is worth noting that `git.CheckForUncommittedGitChanges` would return an error that indicates if the provided input is not a valid directory and was captured in the error handling below, but it's safer to more explicit with the validation here.

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +386,120 @@ func validateLabelFlagValues(flags *flags) error {
...
+	ctx context.Context,
+	container appext.Container,
+	flags *flags,
+) ([]bufmodule.UploadOption, error) {
+	input, err := bufcli.GetInputValue(container, flags.InputHashtag, ".")
+	if err != nil {
+		return nil, err
+	}
+	// validate that input is a local directory
+	fileInfo, err := os.Stat(input)
```
bufdev, line 437: This is not sufficient to validate the input type. You need to parse this in the same manner as we do in the format command, to validate the input would be parsed into a DirRef, and you additionally need to do this validation

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +386,120 @@ func validateLabelFlagValues(flags *flags) error {
...
+	if err != nil || !fileInfo.IsDir() {
+		return nil, appcmd.NewInvalidArgumentErrorf("input, %s, must be a local Git repository checkout: %w", input, err)
+	}
+	runner := command.NewRunner()
+	uncommittedFiles, err := git.CheckForUncommittedGitChanges(ctx, runner, input)
+	if err != nil {
+		return nil, err
+	}
+	if len(uncommittedFiles) > 0 {
+		return nil, fmt.Errorf("uncommitted changes found in the following files: %v", uncommittedFiles)
```
bufdev, line 447: You should specify to the user that git-metadata requires there to be no uncommitted changes

private/buf/cmd/buf/command/push/push.go
```diff
@@ -347,6 +386,120 @@ func validateLabelFlagValues(flags *flags) error {
...
+	gitLabelsUploadOption, err := getGitMetadataLabelsUploadOptions(ctx, runner, container, input, currentGitCommit)
+	if err != nil {
+		return nil, err
+	}
+	if gitLabelsUploadOption != nil {
+		gitMetadataUploadOptions = append(gitMetadataUploadOptions, gitLabelsUploadOption)
+	}
+	sourceControlURL := originRemote.SourceControlURL(currentGitCommit)
+	if sourceControlURL == "" {
+		return nil, appcmd.NewInvalidArgumentError("unable to determine source control URL for this repository; only GitHub/GitLab/BitBucket are supported")
```
bufdev, line 470: Why does it matter? This is a severe limitation, and excludes any github enterprise customer. We should be able to parse a source control url for any http/https/ssh endpoint

private/pkg/git/git.go
```diff
@@ -135,3 +152,161 @@ type ListFilesAndUnstagedFilesOptions struct {
...
+	// HEADBranch is the name of the HEAD branch of the remote.
+	HEADBranch() string
+	// Hostname is the host name parsed from the remote URL. If the remote is an unknown
+	// kind, then this may be an empty string.
+	Hostname() string
+	// RepositoryPath is the path to the repository based on the remote URL. If the remote
+	// is an unknown kind, then this may be an empty string.
+	RepositoryPath() string
+	// SourceControlURL makes the best effort to construct a user-facing source control url
+	// given a commit sha string based on the remoteKind and available hostname and
```
nicksnyder, line 169: Seems weird to me to reference a private implementation detail like `remoteKind` in a public interface.

private/pkg/git/git.go
```diff
@@ -135,3 +152,161 @@ type ListFilesAndUnstagedFilesOptions struct {
...
+	// repository path information.
+	//
+	// If the remote hostname contains bitbucket (e.g. bitbucket.mycompany.com or bitbucket.org),
+	// then it uses the route /commits for the git commit sha.
+	//
+	// If the remote hostname contains github (e.g. github.mycompany.com or github.com) or gitlab
+	// (e.g. gitlab.mycompany.com or gitlab.com) then it uses the route /commit for the git
```
nicksnyder, line 176: I think it would be clearer if you documented the full route returned, like:

`https://<hostname><repository-path>/commit/<git-commit-sha>`


(and same for bitbucket)

private/pkg/git/remote.go
```diff
@@ -0,0 +1,277 @@
...
+	"github.com/bufbuild/buf/private/pkg/app"
+	"github.com/bufbuild/buf/private/pkg/command"
+)
+
+const (
+	bitBucketHostname           = "bitbucket"
+	githubHostname              = "github"
+	gitlabHostname              = "gitlab"
+	gitSuffix                   = ".git"
+	githubGitlabRemoteURLFormat = "https://%s%s/commit/%s"
```
nicksnyder, line 37: is it really worth it to smash these two cases together rather than having a separate variable for each, even if the value is the same?

private/buf/cmd/buf/command/push/push.go
nicksnyder: There are two potential questions here so will answer both.

**Why is it an error if we can't infer a source control url?**
If the user provided `--git-metadata` then they are expecting us to automatically set some values. If we can't set the values they are expecting, we should error. This means during setup, they will catch the problem early, and then later if, due to a code host migration, it breaks, they will know instantly.

**Why are we limited to GitHub/Bitbucket/GitLab?**
The specification for Remote.SourceControlURL() (which is in the interface documentation in this PR) is designed to work for enterprise customers as long as the name of their code host is in the remote name. If it isn't, we don't know what code host we are dealing with and therefore we don't know how to correctly construct the user facing url because the routes are different depending on the code host.

On GitHub and GitLab, routes look like `https://<host><repo-path>/commit/<commit-sha>`
and on BitBucket, routes look like `https://<host><repo-path>/commits/<commit-sha>`
(note `/commit/` vs `/commits`).

bufdev: The code quite literally checks for if the hostname includes `github, gitlab, bitbucket`, and just with a `strings.Contains` on the hostname. This both requires hostnames to include github, for example (this isn't a hard requirement), and it would also accept things like `foogithubbar`, which doesn't feel like what this filter wants to do in the first place.

This limitation isn't acceptable for our enterprise customers. Additionally, my impression was that "source control URL" wasn't mean to link to a specific point-in-time on the repository, it was just meant to link to the repository itself.

We can discuss deferring this until after release, but I consider this a bug - limiting acceptable hostnames to `strings.Contains` on one of `github, gitlab, bitbucket` isn't acceptable

bufdev (approved):

PR #2956, opened 2024-05-08
Add Cargo Registry support to `buf.plugin.yaml`

stefanvanburen:
The plan is to land this on `main` (for the plugins repo) and then merge it over to `dev`. Draft until I PR over on plugins.


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -227,6 +228,29 @@ message PythonConfig {
...
+  // It is the subset of a full Cargo dependency specification, which contains
+  // fields that are irrelevant for Generated SDKs.
+  // Ref: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html
+  message RuntimeLibrary {
+    // name specifies the name of the dependency.
+    string name = 1;
+    // version_requirement specifies the version requirement of the dependency.
+    string version_requirement = 2;
+    // default_features specifies whether or not default features will be enabled for the dependency.
+    bool default_features = 3;
```
mfridman, line 243: Can you remind me, what happens if this is `true` and we also specify features, it's additive?

mfridman (approved):

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
stefanvanburen: Had to check on this myself — best reference is https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature. My understanding is that it's additive — `default-features = true, features = ['xyz']` results in both the `'xyz'` and any features in the `default` feature to be enabled.

mfridman: Got it, really doesn't matter for our purposes since it'll be a property of the plugin itself.

mfridman (approved):

PR #2962, opened 2024-05-09
Update changelog to reflect recent work in buf breaking and buf lint

jhump:
These are change-log worthy, but I failed to make note of them in recent PRs.

Is this too wordy?

(Some of this is duplicated in change log updates in #2963. I was hoping to merge this first, just to update the change-log to reflect changes already merged into the dev branch. After that, I'll reconcile that other branch with whatever wordsmithing we land on here.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,35 @@
 
 ## [Unreleased]
 
-- No changes yet.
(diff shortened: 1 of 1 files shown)


bufdev (approved):
Just get it in, I'll do a pass before v1.32.0

PR #2966, opened 2024-05-10
Avoid pruning unused dependencies for `buf dep update`

doriable:
When running `buf mod update`, we want to include in the `buf.lock` deps
that have been configured in `buf.yaml` and may not be used yet. This is to
support workflows where a user can configure a dependency and `buf dep update`
before they use the newly configured dep in their proto definitions.

`Prune` does the following things right now:
- builds the workspace
- remove unused dependencies from the `buf.lock`
- check for dependencies not configured in `buf.yaml` that are present in `buf.lock`

The first step is necessary for `buf dep update` for tamper-proofing and validating
the build. However, the latter two steps are unnecessary -- we want to keep
unused dependencies and their transitive dependencies, and since `dep update`
gets the dependencies from `buf.yaml` and writes them into `buf.lock`, there
will not be dependencies present not from the `buf.yaml`, so that step is a
no-op.

So instead of running `Prune`, we validate the build after the logic for `dep udpate`
and we log out the unused dependencies as a warning to the users (but we do not
remove them).


private/buf/cmd/buf/command/dep/internal/internal.go
```diff
@@ -114,6 +120,23 @@ func Prune(
...
+		}
+		unusedModuleKeys, err := moduleKeyProvider.GetModuleKeysForModuleRefs(
+			ctx,
+			unusedModuleRefs,
+			workspaceDepManager.BufLockFileDigestType(),
+		)
+		if err != nil {
+			return err
+		}
+		depModuleKeys = append(depModuleKeys, unusedModuleKeys...)
```
bufdev, line 137: We should make sure that the union of `depModuleKeys` and  `unusedModuleKeys` are unique by `ModuleFullName`. Do we expect that `unusedModuleKeys` will never have overlap with `depModuleKeys`? If so, error if not unique, otherwise we need to dedupe.

doriable: Yeah, I think our expectation is that they should not overlap, since `depModuleKeys` is pulled from what is used and `unusedModuleKeys` should only be things that have never been used, so agree, we should error here.

doriable:
Found something in testing today: when pulling in a new dependency that is not used yet, we do not pull in its transitive dependencies... I think this is expected/fine since you may not end up needing all transitive dependencies from the new dependency. But I wanted to note this.

private/buf/cmd/buf/command/dep/internal/internal.go
```diff
@@ -114,6 +120,34 @@ func Prune(
...
+		unusedModuleKeys, err := moduleKeyProvider.GetModuleKeysForModuleRefs(
+			ctx,
+			unusedModuleRefs,
+			workspaceDepManager.BufLockFileDigestType(),
+		)
+		if err != nil {
+			return err
+		}
+		foundDepModuleKeys := copy(make([]bufmodule.ModuleKey, len(depModuleKeys)), depModuleKeys)
+		depModuleKeys = append(depModuleKeys, unusedModuleKeys...)
```
saquibmian, line 138: I don't think this is right. You need to use `GraphProvider` to pull in the full dependency graph including those references (i.e., `internal.ModuleKeysAndTransitiveDepModuleKeysForModuleRefs`, which is what `dep update` does anyway). Without this, as soon as the user goes to import a file from the dependency, if that file's transitive dependencies are not found, the module will fail to build and the user will have to run `dep update` again. If you are using GraphProvider, you need to coordinate with `depModuleKeys` as well -- it needs to be entire enclosure.

Question: why do we prune the lock file at all in `dep update`? I think we want to remove unused deps, but isn't the core issue that the user just pulled it and is about to use it?

saquibmian (commented):
I don't think this is right.

private/buf/cmd/buf/command/dep/internal/internal.go
doriable: > if that file's transitive dependencies are not found, the module will fail to build and the user will have to run dep update again.

Yeah, hence top-level comment around how the transitive dependencies wouldn't be pulled in. This seems somewhat reasonable because you don't know which transitive dependencies you need until you actually use the dependency.

> Question: why do we prune the lock file at all in dep update? I think we want to remove unused deps, but isn't the core issue that the user just pulled it and is about to use it?

Yeah, but there are also dependencies that have been removed from configuration (`buf.yaml`) that are being pruned from `buf.lock` at the time of `dep update`.

saquibmian: I would be frustrated as a user if I had to run `dep update` twice, especially if I'm in an IDE and the tooling is complaining.

> dependencies that have been removed from configuration

Aren't this already pruned earlier in `dep update`? `dep update` is basically `ModuleKeysAndTransitiveDepModuleKeysForModuleRefs` + write to lock file, which should remove entries from lock file that are stale.

doriable: By running `Prune` again, we are verifying the build, which does have some implications on tamper-proofing. But perhaps running the full `Prune` is not the way to go, as you are correct, the first write should already capture that (there is a clarifying comment also, my mistake on this one ^^").

doriable: I made a change where we do not run `Prune` at all in `dep update`, but instead just verify the build and log unused deps. This way, we can ensure the transitive dependencies of the unused deps are included, I think you might be right, this might be the more elegant user experience.

private/buf/cmd/buf/command/dep/internal/internal.go
```diff
@@ -69,6 +70,8 @@ func Prune(
 	bufYAMLBasedDepModuleKeys []bufmodule.ModuleKey,
 	workspaceDepManager bufworkspace.WorkspaceDepManager,
 	dirPath string,
+	// For some use cases, such as dep/mod update, we want to keep unused declared dependencies
+	// in buf.lock (and only prune dependencies that are no longer declared) when set to true.
```
saquibmian, line 74: This comment is not needed anymore.

private/buf/cmd/buf/command/dep/internal/internal.go
```diff
@@ -120,6 +111,30 @@ func Prune(
 	return workspaceDepManager.UpdateBufLockFile(ctx, depModuleKeys)
 }
 
+// LogUnusedConfiugredDepsForWorkspace takes a workspace and logs the unused configured
+// dependneices as warnings to the user.
```
saquibmian, line 115:
```suggestion
// dependencies as warnings to the user.
```

saquibmian (commented):
Minor comments

private/buf/cmd/buf/command/dep/depupdate/depupdate.go
```diff
@@ -143,11 +143,11 @@ func run(
 	if err := workspaceDepManager.UpdateBufLockFile(ctx, configuredDepModuleKeys); err != nil {
 		return err
 	}
-	// Prune the buf.lock. This also verifies the workspace builds again.
-	// Building also has the side effect of doing tamper-proofing.
```
bufdev, line 147: This comment should remain and continue to be true.

However, based on the latest changes, it is no longer true. The workspace is no longer built.

doriable: The workspace is still being built on the next line and then we are still logging out the unused configured deps for the user.

bufdev: `GetWorkspace` does not build the workspace.

nicksnyder: @doriable where does it get build? Previous prune codepath explicitly calls `controller.GetImageForWorkspace`

nicksnyder: ah, nm, right there in the diff, lol

nicksnyder: wait, no, that is just `GetWorkspace` not `GetImageForWorkspace`

bufdev: Correct - the previous code called `GetImageForWorkspace`, which does the build, this code does not

doriable: Ah my apologies, I misconstrued that as compiling the `ModuleSet` rather than building the image. Yes, I should add that step back.

private/buf/cmd/buf/command/dep/depprune/depprune.go
```diff
@@ -75,5 +75,13 @@ func run(
 	if err != nil {
 		return err
 	}
-	return internal.Prune(ctx, container.Logger(), controller, configuredDepModuleKeys, workspaceDepManager, dirPath)
+	return internal.Prune(
+		ctx,
+		container.Logger(),
+		controller,
+		container,
```
pkwarren, line 82: It looks like the container parameter didn't end up being used.

private/buf/cmd/buf/command/dep/internal/internal.go
```diff
@@ -53,11 +53,12 @@ func ModuleKeysAndTransitiveDepModuleKeysForModuleRefs(
 
 // Prune prunes the buf.lock.
 //
-// Used by both mod prune and mod update.
+// Used by both dep/mod prune and dep/mod update.
```
pkwarren, line 56:
```suggestion
// Used by dep/mod prune.
```

It appears that update is no longer using this, so we can tweak the comment.

pkwarren (commented):
Changes look good - two minor comments.

pkwarren (approved):

saquibmian (approved):

bufdev (approved):

PR #2967, opened 2024-05-10
Simplify handling of custom features in buf breaking

jhump:
This is for the logic for querying the `(pb.cpp)` and `(pb.java)` custom features ("custom" because they are extension fields in `FeatureSet` instead of regular fields).

The gist is that, with some improved handling over in protocompile for custom features and potentially mismatching descriptors, we can get rid of a bunch of special logic in here. (Yay!)

This is just a draft PR right now because it's pointing to a commit in protocompile that hasn't yet merged to main (here: https://github.com/bufbuild/protocompile/pull/306). When that PR is merged, I'll update the `go.mod` pin and mark this as ready for review.


jhump:
Hmm, never mind. This broke tests in a serious way, so it clearly isn't working. I'll revisit/re-open once I get it all working...

PR #2969, opened 2024-05-10
Update buf format to work correctly with editions

jhump:
Thanks @stefanvanburen for noticing this!

The formatter needed to be updated to be aware of the `edition` AST node (which appears instead of `syntax` in Editions files) and also needed updates to reserved names, which use identifiers instead of string literals in Editions.

### DIFF
--- a/private/buf/bufformat/formatter.go
+++ b/private/buf/bufformat/formatter.go
@@ -243,11 +243,16 @@ func (f *formatter) writeFileHeader() {
 			continue
 		}
 	}
-	if f.fileNode.Syntax == nil && packageNode == nil && importNodes == nil && optionNodes == nil {
+	if f.fileNode.Syntax == nil && f.fileNode.Edition == nil &&
+		packageNode == nil && importNodes == nil && optionNodes == nil {
 		// There aren't any header values, so we can return early.
 		return
 	}
-	if syntaxNode := f.fileNode.Syntax; syntaxNode != nil {
+	editionNode := f.fileNode.Edition
+	if editionNode != nil {
+		f.writeEdition(editionNode)
+	}
+	if syntaxNode := f.fileNode.Syntax; syntaxNode != nil && editionNode == nil {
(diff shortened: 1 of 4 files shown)


private/buf/bufformat/formatter.go
```diff
@@ -243,11 +243,16 @@ func (f *formatter) writeFileHeader() {
...
+		packageNode == nil && importNodes == nil && optionNodes == nil {
 		// There aren't any header values, so we can return early.
 		return
 	}
-	if syntaxNode := f.fileNode.Syntax; syntaxNode != nil {
+	editionNode := f.fileNode.Edition
+	if editionNode != nil {
+		f.writeEdition(editionNode)
+	}
+	if syntaxNode := f.fileNode.Syntax; syntaxNode != nil && editionNode == nil {
```
stefanvanburen, line 255: This would effectively wipe a `syntax` line from the file if an `edition` line exists, but that seems reasonable, right?

stefanvanburen (approved):

private/buf/bufformat/formatter.go
jhump: They cannot both exist. You either specify syntax proto2 or proto3 **or** you specify an edition.

stefanvanburen: makes sense — just wondering if we should leave it up to a linter or a `buf build` failing to point that out to the user, rather than removing it altogether. But I don't think this is _unreasonable_ behavior, and should be a niche case anyway.

jhump: It's an invariant of the AST. This logic is really here just to try to produce a valid source file even if there's something unexpectedly wrong with the AST. FWIW, we already do similar in the reserved nodes, where we prefer names, then identifiers, and then ranges. If a node had more than one set, the others would be ignored (which is correct to produce a valid source file since only one is allowed).

bufdev (approved):

PR #2980, opened 2024-05-14
Change `buf beta registry label archive/unarchive` to `buf beta registry archive/unarchive`

doriable:
This PR changes `buf beta registry label archive/unarchive` to
`buf beta registry archive/unarchive`, which is able to take a
workspace input and archive/unarchive all labels for modules
in the workspace.
This UX change is needed to support uses-cases such as
GitHub Actions.


emcfarlane (approved):

CHANGELOG.md
```diff
@@ -58,6 +56,8 @@
   `features.field_presence = LEGACY_REQUIRED` in Editions files. This new rule is not active by
   default in existing `v1` and `v1beta1` configurations, for backwards-compatibility reasons.
   Migrate your config to `v2` to use them.
+- Add `buf beta registry archive` and `buf beta registry unarchive` commands for managing
+  labels on the BSR.
```
saquibmian, line 60:
```suggestion
  archivable assets (like labels) on the BSR.
```
The command is not strictly associated with labels.

private/buf/cmd/buf/command/beta/registry/archive/archive.go
```diff
@@ -0,0 +1,133 @@
...
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appext.SubCommandBuilder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <input> --label <label>",
+		Short: "Archive one or more labels for the given input",
```
saquibmian, line 44: Same idea here... this command is not specific to labels.

private/buf/cmd/buf/command/beta/registry/unarchive/unarchive.go
```diff
@@ -0,0 +1,133 @@
...
+
+// NewCommand returns a new Command.
+func NewCommand(
+	name string,
+	builder appext.SubCommandBuilder,
+) *appcmd.Command {
+	flags := newFlags()
+	return &appcmd.Command{
+		Use:   name + " <input> --label <label>",
+		Short: "Unarchive one or more labels for the given input",
```
saquibmian, line 44: Same idea here

saquibmian (commented):
My comments are mostly about the command's docs; otherwise looks good.

CHANGELOG.md
doriable: Yeah -- I debated about how to present the docs, the command is not specific to labels, but labels are the only thing that you can interact with, and it's a required flag. I didn't want "archive" to be misconstrued with "deprecate", for example, but maybe I'm digging too deep ^^"

nicksnyder: While we have structured this command to be a generic archive command, but today we can only archive labels so I think it is clearer to users if we document it as a command for archiving labels. If/when we introduce other resource types we can update it then.

I would make a slight tweak to this wording to say this isn't for "managing" labels, but for archiving and unarchiving labels.

saquibmian:
I wonder if we should consider target modules in the workspace as well. The `Archive/Unarchive` RPCs currently does not allow archiving/unarchiving labels that don't exist. Does that pose a problem where the user would like to archive a label that only exists for some modules in the workspace but not for others?

doriable:
> Does that pose a problem where the user would like to archive a label that only exists for some modules in the workspace but not for others?

I tried to bring this up yesterday, but I think it got a little lost. Right now the behaviour would be that this would error, since it's a single RPC call (per registry, but everything _should_ be from the same registry), and I think this is a pretty reasonable behaviour. We are disallowing this for `v1` workspaces, and so users with `v1` workspaces would still need to continue with the workflows they already have to manage each individual module. This will start coming up for `v2` workspaces, but I am okay with launching with the behaviour of error-ing and then going from there. But open to other's thoughts on this.

nicksnyder:
I think matching the API behavior here is fine since the alternative is just to target the specific modules that have the label you want to archive.

bufdev (approved):

PR #2982, opened 2024-05-16
Return a more user-friendly error in workspace targeting for overlaps

doriable:
In the past, if the an overlap is found between `moduleDirPath` and `inputPath`,
but `inputPath` does not exist (e.g. `moduleDirPath == "."` and `inputPath == "fake-path"`)
then it would return the error:

```
failed to build input "fake-path" because it is contained by module at path "." specified in your configuration, you must provide the workspace or module as the input, and filter to this path using --path"
```

Which isn't particularly clear, since `fake-path` doesn't exist. Added a check to return a
nicer error to the user by checking `inputPath`. Also included test cases for `v1` and `v2`
workspaces.

### DIFF
--- a/private/buf/bufworkspace/workspace_targeting.go
+++ b/private/buf/bufworkspace/workspace_targeting.go
@@ -700,7 +700,7 @@ func checkForOverlap(
 				return err
 			}
 			if empty {
-				return nil, bufmodule.ErrNoTargetProtoFiles
+				return bufmodule.ErrNoTargetProtoFiles
 			}
 			return fmt.Errorf("failed to build input %q because it is contained by module at path %q specified in your configuration, you must provide the workspace or module as the input, and filter to this path using --path", inputPath, moduleDirPath)
 		}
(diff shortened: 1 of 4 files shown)


private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -681,9 +681,26 @@ func checkForControllingWorkspaceOrV1Module(
...
-func checkForOverlap(inputPath string, moduleDirPaths []string) error {
+func checkForOverlap(
+	ctx context.Context,
+	bucket storage.ReadBucket,
+	inputPath string,
+	moduleDirPaths []string,
+) error {
 	for _, moduleDirPath := range moduleDirPaths {
 		if normalpath.ContainsPath(moduleDirPath, inputPath, normalpath.Relative) {
+			// In the case where the inputPath would appear to be relative to moduleDirPath,
```
bufdev, line 692: If the path does exist but is empty though, this will be a confusing error message as well right?

doriable: Yeah, this is true -- ideally I want to check if the directory is a valid prefix for the bucket. Given the `IsEmpty` check, I've tweaked this to return `ErrNoTargetProtoFiles` since the `inputPath` is empty (and I still think this is a valuable check/improvement over just returning the overlap error).

bufdev:
Let's punt this until after the release

bufdev (approved):

PR #2983, opened 2024-05-16
Only archive and unarchive labels for target modules

doriable:
We should only be archiving and unarchiving labels
for target modules.


saquibmian (approved):
Awesome. This is what I was trying to say in [this comment](https://github.com/bufbuild/buf/pull/2980#issuecomment-2112857189), but didn't communicate effectively. Thanks for circling back on this!

PR #2989, opened 2024-05-17
Add `revision` to `GetRemotePackageVersionPlugin`

stefanvanburen:
We're now using these RPCs to power the BSR UI, which need to be able to specify this field. Not currently adding it to the CLI calls, which will continue to leave this unspecified (meaning the latest revision).


paul-sachs (approved):

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -221,4 +221,7 @@ message GetRemotePackageVersionPlugin {
   // If empty, this is a reference to the latest version.
   // example: v1.0.0
   string version = 3;
+  // The revision of the plugin version.
+  // example: 1
+  uint32 revision = 4;
```
stefanvanburen, line 226: FWIW, I believe we originally kept this message different from `CuratedPluginReference` due to the inability to specify this field, but it's now [exactly the same](https://github.com/bufbuild/buf/blob/ff65e5e7572270fcd8f2490dfdfa58235c3100d2/proto/buf/alpha/registry/v1alpha1/plugin_curation.proto#L271-L280), field-wise.

bufdev (approved):

PR #3003, opened 2024-05-21
Map target paths and target exclude paths for archive/git refs

doriable:
In CLI versions pre-1.32.0, for archive and git refs, we would map
the root of the bucket of archive and git refs to the `subDirPath`,
which is inconsistent with the behavior for dir refs (which maps
target paths and target exclude paths to the execution context).

In 1.32.0, we made changes to handle subDirPath, target paths,
and target exclude paths in a unified way, however this is not
backwards compatible with pre-1.32.0 versions. So, in order
to be backwards compatible, we need to remap the target
paths and target exclude paths.

I explored an option to check of the subDirPath contains
the target paths and target exclude paths before remapping,
however this is brittle, since we could hit an edge case with
directories that have the same name nested:

```
.
└── proto
    └── foo
        └── foo
```

Tests have been adjusted to reflect this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 ## [Unreleased]
 
 - Fix archive and git inputs so that `--path` and `--exclude-path` paths are relative to
-  the `#subdir` rather than the root of the input. This fixes a unintended behavior change that was introduced in `v1.32.0`.
+  the `#subdir` rather than the root of the input. This fixes an unintended behavior change
+  that was introduced in `v1.32.0`.
 
 ## [v1.32.0] - 2024-05-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 - Fix archive and git inputs so that `--path` and `--exclude-path` paths are relative to
-  the `#subdir` rather than the root of the input.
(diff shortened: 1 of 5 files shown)


bufdev:
Needs changelog entry for when we do 1.32.1 release

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Fix archive and git inputs so that `--path` and `--exclude-path` paths are relative to
+  the `#subdir` rather than the root of the input.
```
nicksnyder, line 6:
```suggestion
- Fix archive and git inputs so that `--path` and `--exclude-path` paths are relative to
  the `#subdir` rather than the root of the input. This fixes a unintended behavior change that was introduced in `v1.32.0`.
```

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Fix archive and git inputs so that `--path` and `--exclude-path` paths are relative to
+  the `#subdir` rather than the root of the input. This fixes a unintended behavior change that was introduced in `v1.32.0`.
```
nicksnyder, line 6: `a` -> `an` (and wrap)
```suggestion
  the `#subdir` rather than the root of the input. This fixes an unintended behavior change
  that was introduced in `v1.32.0`.
```

bufdev (approved):

PR #3009, opened 2024-05-22
Avoid building image for `buf export` when `--exclude-imports` flag is set

### LINKED ISSUE #3002: buf export --exclude-imports not being applied
I am trying to upgrade from buf v1.25.0 to v1.32.0.  We are using the export command to download protos from a separate git repo.  After upgrading to the latest version of buf the export command now fails with the error:

```
xxxx/xxxx/xxx/xxx/xx/xxxx.proto:22:8:import "google/api/annotations.proto": file does not exist
```

I would expect the exclusion of the imports would prevent this check or is there a different way to prevent this check with the export command in v1.32.0?

doriable:
This PR changes `buf export` so that when `--exclude-imports` is
passed, it does not require the input to be buildable as an image.
This allows a user to provide an input that may not be buildable, and
they can still get the source files back when `--exclude-imports` is
set, since we do not need the image to check for import files.

Fixes #3002

### DIFF
--- a/private/buf/cmd/buf/command/export/export.go
+++ b/private/buf/cmd/buf/command/export/export.go
@@ -150,15 +150,6 @@ func run(
 		return err
 	}
 	moduleReadBucket := bufmodule.ModuleSetToModuleReadBucketWithOnlyProtoFiles(workspace)
-	image, err := controller.GetImageForWorkspace(
-		ctx,
-		workspace,
-		bufctl.WithImageExcludeSourceInfo(true),
-		bufctl.WithImageExcludeImports(flags.ExcludeImports),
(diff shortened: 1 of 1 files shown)


doriable:
Verified this with the repro from #3002:

```
🐱~/src/playground/buf-3002 @ $buf export ssh://git@github.com/rgaskill/buf-test.git#branch=main --exclude-imports --path proto/test/pkg -o .
🐱~/src/playground/buf-3002 @ $ tree .
.
└── pkg
    └── v2
        └── test.proto

3 directories, 1 file
```

bufdev (approved):

Issue #2986, opened 2024-05-16
Add lint rule to prevent stable packages importing unstable packages


noahseger:
> A common mistake we see is forgetting a v1alpha import into a v1 package. Has a lint rule preventing this been considered? If not, what do folks here generally use to lint better on top of Buf?

I think the general community would benefit from baking this into the DEFAULT configuration though, because even changing a field from `v1alpha.MyType` to `v1.MyType` is breaking, and requires the ignore / commit / unignore / commit dance.

bufdev:
I'm not sure I understand the question or proposal, can you explain a bit more what you are looking for?

noahseger:
@bufdev say I have two protobufs like:

```protobuf
syntax = "proto3";

package examples.v1alpha;

message AlphaDependency {
  string foo = 1;
}
```

```protobuf
syntax = "proto3";

import "examples/v1alpha/dependency.proto"

package examples.v1;

message StableExample {
  examples.v1alpha.AlphaDependency dependency = 1;
}
```

There are two primary problems:

1. Buf (by default) allows me to change the v1alpha dependency in incompatible ways, which silently breaks v1
2. If I want to promote the v1alpha dependency to v1, the type change in the existing stable file is breaking

bufdev:
- By default, `buf` does not allow you to make breaking changes to the v1alpha package, that is an option you have to opt into via https://buf.build/docs/configuration/v2/buf-yaml#ignore_unstable_packages. If `ignore_unstable_packages` is not set, then breaking changes to `v1alpha1` will result in errors when running `buf breaking`.
- Changing the type of a field is considered a breaking change, so updating the type of `examples.v1alpha.AlphaDependency` to say `examples.v1.AlphaDependency` will result in an error when running `buf breaking`. There's no concrete way for `buf breaking` to denote that "this was just a package change, the underlying type is exactly the same, do not error", and we wouldn't want to ignore this as a general concern. We could, in theory, do deep type inspection, but this would be source-code-breaking to begin with, and we consider that out of scope. Your mitigation here is to override the breaking change error if you are certain that your change is expected.

noahseger:
Ah yes, understood — we did enable `ignore_unstable_packages` because we need to evolve alpha versions in breaking ways.

But to be clear, the feature request isn't to inspect the alpha dependency types (although I think this would work with `PACKAGE` as the breaking configuration type?).

The feature request is:

- when I have a stable proto file
- and I import an unstable proto file into it
[...]

PR #3010, opened 2024-05-23
Add STABLE_PACKAGE_NO_IMPORT_UNSTABLE lint rule

bufdev:
I'm not yet sure if we want to add this into mainline. Opening this for discussion.

This is re: https://github.com/bufbuild/buf/issues/2986.

This adds a `STABLE_PACKAGE_NO_IMPORT_UNSTABLE` lint rule, with no category, that will produce a lint failure if an stable package imports a file with an unstable package. For example:

1. If `a.proto` imports `b.proto`, and `a.proto` has package `a.v1`:

- If `b.proto` has package `foo`, no failure - cannot determine package stability.
- If `b.proto` has package `foo.v1`, no failure - stable imports stable.
- If `b.proto` has package `foo.v1beta1`, failure - stable imports unstable.

2. If `a.proto` has package `a`:

- No failure ever - cannot determine package stability.

3. If `a.proto` has package `a.v1beta1`:

- No failure ever - package is unstable, does not matter what it imports.

If we do decide to go ahead with this, this PR:

- Needs some discussion as to the rule name.
- Needs testing.


private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
```diff
@@ -1029,6 +1029,32 @@ func checkServiceSuffix(add addFunc, service bufprotosource.Service, suffix stri
...
+		return nil
+	}
+	// Package is stable. Check imports.
+	for _, fileImport := range file.FileImports() {
+		packageVersion, ok := protoversion.NewPackageVersionForPackage(fileImport.File().Package())
+		if !ok {
+			continue
+		}
+		if packageVersion.StabilityLevel() != protoversion.StabilityLevelStable {
+			add(fileImport, fileImport.Location(), nil, `This file has package %q that represents a stable version, but it imports %q, which has package %q that represents an unstable version. Stable packages should not depend on unstable packages.`, file.Package(), fileImport.Import(), fileImport.File().Package())
```
akshayjshah, line 1052: Can we shorten the error text?

`This file is in stable package %q, so it should not depend on %q from unstable package %q.`

akshayjshah (commented):
This feels like a valuable check to me!

Organizations with many existing violations of this rule may have a hard time adopting it, but that shouldn't stop us from offering it.

private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go
noahseger: 👍 @akshayjshah @bufdev 

I'm just doc golfing but less text is always nice`Stable package %q cannot import unstable file %q.`

private/bufpkg/bufcheck/buflint/testdata/lint_stable_package_import/api/v1/foo.proto
```diff
@@ -0,0 +1,9 @@
+syntax = "proto3";
+
+package api.v1;
+
+import "api/v1alpha/bar.proto";
```
bufdev, line 5: An import on a stable package, and a package with no stability level, should also be added (both should not result in failures)

bufdev (commented):
Changelog needs to be updated

bufdev (commented):
LGTM

PR #3013, opened 2024-05-24
Support annotated tags for push git metadata

emcfarlane:
Fix resolving of annotated tags on push with `--git-metadata` flag by fetching from the remote and then using `--points-at` to discover each ref and tag. Currently only lightweight tags are resolved as the commit sha for annotated tags doesn't directly point to the commit.


saquibmian (approved):
LGTM

bufdev (approved):

PR #3014, opened 2024-05-24
Make `addedModule` to `Module` operation work in batch

unmultimedio:
When looking at `GetGraph` RPC performance we realized the CLI is fetching remote module data sequentially for each dep, when all of them can be done in batch in a single RPC.

This PR allows fetching remote module datas in batch. In summary:

```diff
- modules, err := slicesext.MapError(
-   addedModules,
-   func(a *addedModule) (Module, error) {
-     return a.ToModule() // this loads addedModules' ModuleDatas in sequence one by one
-   },
- )
+ modules, err := addedModulesToModules(addedModules) // this loads all addedModules' datas in batch
```

### DIFF
--- a/private/bufpkg/bufmodule/added_module.go
+++ b/private/bufpkg/bufmodule/added_module.go
@@ -158,7 +158,18 @@ func addedModulesToModules(
 		if a.localModule != nil {
 			continue // local Modules were already loaded.
 		}
-		module, err := a.toModule(ctx, commitProvider, getModuleDatas, inputIdxToRemoteIdx[i])
+		i := i
+		module, err := a.toModule(
+			ctx,
+			commitProvider,
+			func() (ModuleData, error) {
+				moduleDatas, err := getModuleDatas()
+				if err != nil {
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/added_module.go
```diff
@@ -94,72 +94,115 @@ func (a *addedModule) OpaqueID() string {
...
+// addedModulesToModules converts the addedModules to Modules.
 //
-// If the addedModule is a local Module, this is just returned.
-// If the addedModule is a remote Module, the ModuleDataProvider and CommitProvider are queried to get the Module.
-func (a *addedModule) ToModule(
+// If an addedModule is a local Module, this is just returned.
+// If an addedModule is a remote Module, the ModuleDataProvider and CommitProvider are queried to get the Module.
+//
+// The Modules are returned in the same order as requested.
+func addedModulesToModules(
```
pkwarren, line 103: To keep with the previous naming scheme, perhaps call it `toModules`.

PR #3023, opened 2024-05-27
Improve push `--create` behaviour

emcfarlane:
This PR improves the push behaviour when the `--create` flag is applied. It always tries to fetch the repository before attempting to create if not found. This improves the happy path when every module that is being pushed to already exists. It lowers the required permissions as module creation requires greater permissions than fetching.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
+++ b/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
@@ -111,38 +111,17 @@ func (a *uploader) Upload(
 	if err != nil {
 		return nil, err
 	}
-
-	// This must be in the same order as contentModules.
-	var modules []*modulev1.Module
-	if uploadOptions.CreateIfNotExist() {
-		// We must attempt to create each module one at a time, since CreateModules will return
-		// an `AlreadyExists` if any of the modules we are attempting to create already exists,
-		// and no new modules will be created.
-		modules = make([]*modulev1.Module, len(contentModules))
-		for i, contentModule := range contentModules {
-			module, err := a.createContentModuleIfNotExist(
-				ctx,
-				primaryRegistry,
-				contentModule,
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -111,38 +111,17 @@ func (a *uploader) Upload(
...
-		// this matches the order of contentModules.
-		modules, err = a.validateContentModulesExist(
-			ctx,
-			primaryRegistry,
-			contentModules,
-		)
-		if err != nil {
-			return nil, err
-		}
+	// We must validate that the modules exist, and if they do not, create them.
```
bufdev, line 114: This is a tad confusing, as the RPC will return `NotFound` if *any* of the N modules does not exist. If so, the create logic is called. Is this what we want? If so, we should update the comments to reflect this is what is going on and why this is OK

emcfarlane: Yes I believe that is what we want. Get all the modules, which will succeed for the happy path where all modules already exist. If some do not exist a 404 error is returned. We then iterate over each module trying to create it as it's not clear which modules failed the fetch. Will update the comments.

doriable: I commented on the issue in detail, but just responding with my 2c here: our API returns an error that explicitly and clearly handles the case where a module already exists for `CreateModules`, so it feels counter-intuitive for a client that intends to call `CreateModules` to also call `GetModules` beforehand to validate.

mfridman: We can probably close this PR, @doriable brings up some valid points. We chatted on the issue and the consensus was to change the behavior elsewhere, instead of in the CLI code.

PR #3031, opened 2024-05-28
Make message literal output more consistent

jhump:
This canonicalizes the formatting of message literals in option values. In particular, it will make the following changes to the format:
1. Always use `{` and `}`. If existing code uses `<` and `>`, change it.
2. Always include a separator between the field name and value. It is technically not required between a field name and a message value or a list value that contains messages. But it is not possible for the formatter to know this perfectly for all cases (for example, in the case of an empty list, how can the formatter know if its element type is a message or not?). So, for consistency, this always uses a `:` separator. If the existing code omits it, one is added.
3. Always omit optional separators between fields. The separators are never necessary. Either `,` or `;` is accepted (you can even mix in the same message literal, which is kinda confusing!). We were already dropping superfluous separators if we ended reformatting a message into "compact" form (one-liner). Now we also do it for expanded form.

These inconsistencies in the output of `buf format` were point out to me by @pkwarren.


private/buf/bufformat/testdata/proto2/message/v1/message_options.golden.proto
```diff
@@ -37,7 +37,7 @@ message Foo {
     foo: 1
     bar: 2
     baz: 3
-    buzz: <
+    buzz: {
```
pkwarren, line 40: Replacing `<` with `{` feels like a nice cleanup. Are there any inputs where `<` would be preferable?

private/buf/bufformat/testdata/proto2/option/v1/option.golden.proto
```diff
@@ -93,16 +93,16 @@ enum Type {
...
-      foo: 1, // Trailing comment on foo.
+      foo: 1 // Trailing comment on foo.
       // Leading comment on bar.
       bar: 2 // Trailing comment on bar.
       // Leading comment on recursive.
-      recursive: <
+      recursive: {
         // Leading comment on foo.
-        foo: 1,
+        foo: 1
```
pkwarren, line 102: Personally I prefer always including the trailing commas even if they aren't necessary (perhaps a side effect of use of Go structs, JSON objects, etc.). It could be argued though that the least amount of unnecessary whitespace is desirable.

private/buf/bufformat/testdata/proto3/all/v1/all.proto
```diff
@@ -69,7 +69,9 @@ message One {
   int64 one_one = 1 [ (custom.field_option) = true];
   Two one_two = 2[
   (custom.field_option) = false,
-  (custom.field_thing_option) = { foo: 6}
+  (custom.field_thing_option) = {
```
bufdev, line 72: Is the reason this is going to a newline because of the comment? I think sans comment we'd want it on a single line, so just checking.

private/buf/bufformat/testdata/proto3/all/v1/all.proto
```diff
@@ -69,7 +69,9 @@ message One {
...
+     foo: 6, // comment after foo
```
bufdev, line 73: Do we need the trailing comma?

private/buf/bufformat/testdata/proto3/all/v1/all.proto
jhump: Yes, because of the comment.

private/buf/bufformat/testdata/proto3/all/v1/all.proto
jhump: This is the input file. If you look in the golden version of this file, the comma is removed.

As to why the comma is present in the input -- it's to make sure we don't lose the comment, which is attached to the comma which we _are_ removing.

bufdev (approved):

PR #3038, opened 2024-05-31
Allow user to override --source-control-url and --create-default-label when using --git-metadata

doriable:
This allows the user to override the behaviour of `--source-control-url`
and `--create-default-label` with their own values when using `--git-metadata`
flag with `buf push`.
The previous behaviour disallowed the user to set these values themselves
if the `--git-metadata` flag was set.

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -154,8 +154,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 
 The source control URL and default branch is based on the required Git remote %q.
 This flag is only compatible with checkouts of Git source repositories.
-If you set the --%s flag yourself, then this value will be used instead and the source control URL will not be derived from the Git source control state.
-This flag does not allow you to set the --%s flag yourself.`,
+If you set the --%s flag and/or --%s flag yourself, then the value(s) will be used instead and the information will not be derived from the Git source control state.`,
 			sourceControlURLFlagName,
 			labelFlagName,
 			labelFlagName,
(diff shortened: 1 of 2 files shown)


private/buf/cmd/buf/command/push/push.go
```diff
@@ -154,7 +154,8 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 
 The source control URL and default branch is based on the required Git remote %q.
 This flag is only compatible with checkouts of Git source repositories.
-This flag does not allow you to set any of the following flags yourself: --%s, --%s.`,
+If you set the --%s flag yourself, then this value will be used instead and the source control URL will not be derived from the Git source control state.
```
bufdev, line 157: Are there other flags we want to be allowed to be overwritten?

doriable: So the only flag left is `--create-default-label`. So a use-case hasn't come up for that flag yet (we had a use-case come up for `--source-control-url` for testing purposes), and I am comfortable not loosening this pre-emptively, since the general rule is that it's much easier to loosen something than to make something more restrictive down the road. However that isn't a strong opinion for me, if others feel that we should simply allow overrides for both flags (and we already have `--label` be additive to the labels `--git-metadata` is setting) as a simpler mental model, I would be onboard with this.

bufdev: I would provide overrides consistently here, so I would do it too

doriable: Sounds good, I'll adjust!

bufdev (approved):

PR #3039, opened 2024-05-31
Add buf beta registry label commands to replace tag and draft commands

oliversun9:
This PR
* Adds `buf beta registry label [create,get,list]`, with condensed `--help` output:
```
$ buf beta registry label create --help                                                         
Create a label for a specified commit
Usage:
  buf beta registry label create <buf.build/owner/repository:commit> <label> [flags]
Flags:
      --format string     The output format to use. Must be one of [text,json] (default "text")
      --update-existing   Update the label if it already exists
      
$ buf beta registry label get --help   
Get label details
Usage:
  buf beta registry label get <buf.build/owner/repository:label> [flags]
Flags:
      --format string   The output format to use. Must be one of [text,json] (default "text")
      
$ buf beta registry label list --help 
List repository labels
Usage:
  buf beta registry label list <buf.build/owner/repository[:ref]> [flags]
Flags:
      --archive-status string   The archive status of the labels listed. Must be one of [archived,unarchived,all] (default "unarchived")
      --format string           The output format to use. Must be one of [text,json] (default "text")
  -h, --help                    help for list
      --page-size uint32        The page size. (default 10)
      --page-token string       The page token. If more results are available, a "next_page" key is present in the --format=json output
      --reverse                 Reverse the results
```
* Removes `buf beta registry label [tag,draft]` commands and their sub-commands since tags are drafts are deprecated concepts now.

Next steps:
* Move `buf beta registry [archive,unarchive]` under `buf beta registry label`. Some changes are needed since these two commands take an `<input>`, which is different from the UX of other `label` commands (as well as the other commands).
* Perhaps add a `buf beta registry label ls-history` command. I'm hesitant because this would mean there will be both `label list` and `label ls-history`, which could look odd. I also don't think `label get --list-history` is a good idea either, because the `ListLabelHistory` rpc has a paginated response, `label get ... --show-history --page-size ...` also feels odd.


private/buf/bufprint/bufprint.go
```diff
@@ -117,6 +118,18 @@ func NewRepositoryTagPrinter(writer io.Writer) RepositoryTagPrinter {
 	return newRepositoryTagPrinter(writer)
 }
 
+// RepositoryLabelPrinter is a repository label printer.
+// TODO: perhaps rename this to LabelPrinter along with other printers
```
oliversun9, line 122: Is this fine?

private/buf/bufprint/repository_label_printer.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+}
+
+func (p *repositoryLabelPrinter) printRepositoryLabelsText(outputRepositoryLabels []outputRepositoryLabel) error {
+	return WithTabWriter(
+		p.writer,
+		[]string{
+			"Name",
+			"Commit",
+			"Create Time",
```
oliversun9, line 75: The only other printers where `Created` is used as the header name, but if I call this `Created` I will have to use `Archived` as well, which might not be a good header name.

private/buf/bufprint/repository_label_printer.go
```diff
@@ -0,0 +1,125 @@
...
+			"Archive Time",
+		},
+		func(tabWriter TabWriter) error {
+			for _, outputRepositoryLabel := range outputRepositoryLabels {
+				formattedArchiveTime := "N/A"
```
oliversun9, line 80: Unarchived (normal) labels don't have an archived time, so I'm just printing a placeholder here. I could also just print an empty string here, but it could look weird if all labels printed are unarchived.

Another solution is to print this column only when at least one commit has an archived time.

bufdev (approved):

private/buf/bufprint/repository_label_printer.go
bufdev: Would prefer empty string. Even better, would prefer column not be printed in text mode if there are no labels with an archive time.

bufdev (approved):

PR #3051, opened 2024-06-05
Include descriptor's location in set of ignore locations

jhump:
This is a proposal for an alternative to #3050. There are actually numerous places in the buf lint and breaking change rules where a location is provided for some element within a descriptor -- like a name, number, option, etc. But the descriptor's `source_code_info` only retains comments for complete declarations. So those other rules are subject to the same potential issue, and the fix is to include the descriptor's location as an ignore location in each of them.

Instead of updating every such rule, this change instead opts to address it more systematically, by _always_ using the descriptor's location as an ignore location (no need to explicitly pass it from every rule). So this undoes the two changes in the protovalidate checks as they are superseded by the change to the check helper that handles the addition of annotations.

### DIFF
--- a/private/bufpkg/bufcheck/internal/helper.go
+++ b/private/bufpkg/bufcheck/internal/helper.go
@@ -110,10 +110,17 @@ func (h *Helper) addFileAnnotationf(
 	format string,
 	args ...interface{},
 ) {
+	baseLocs := []bufprotosource.Location{location}
+	if locDescriptor, ok := descriptor.(bufprotosource.LocationDescriptor); ok {
+		if descLocation := locDescriptor.Location(); descLocation != location {
+			// Include the location of the descriptor itself as a source of ignores.
+			baseLocs = append(baseLocs, descLocation)
+		}
+	}
 	if h.ignoreFunc != nil && h.ignoreFunc(
 		h.id,
 		append([]bufprotosource.Descriptor{descriptor}, extraIgnoreDescriptors...),
-		append([]bufprotosource.Location{location}, extraIgnoreLocations...),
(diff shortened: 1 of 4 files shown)


private/bufpkg/bufcheck/internal/helper.go
```diff
@@ -110,10 +110,17 @@ func (h *Helper) addFileAnnotationf(
 	format string,
 	args ...interface{},
 ) {
+	baseLocs := []bufprotosource.Location{location}
```
doriable, line 113: nit: would probably name this `ignoreLocations` or just `locations` (since it is being passed to the `ignoreFunc` later anyway). I would probably also adjust the subsequent variables, `%s/locDescriptor/locationDescriptor` and `%s/descLocation/descriptorLocation` to adhere to the fuller naming conventions.

doriable (commented):
This change generally makes a lot of sense to me -- I think it is reasonable to check the descriptor's location for ignores. Just left a very small nit, would be interested to see if there are any other thoughts on this change :) Thank you!

private/bufpkg/bufcheck/internal/helper.go
jhump: done

bufdev:
I actually had gone through each rule and done the very-specific ignores I wanted, so at first glance I wasn't sure of this change, but having reviewed it further, I think this is a pure improvement - I don't think there is a case where we don't want the underlying LocationDescriptor to also be able to provide an ignore location. I've made a slight update to remove duplicates (by pointer comparison unfortunately, best effort) in the extra ignore locations, PTAL, otherwise LGTM.

bufdev (approved):

PR #3058, opened 2024-06-07
Avoid overwriting files that don't actually change in buf format

jhump:
This avoids file system change notifications for files that don't actually change. See thread in Public Slack for more context: https://bufbuild.slack.com/archives/CRZ680FUH/p1717784544902049

### DIFF
--- a/private/buf/cmd/buf/command/format/format.go
+++ b/private/buf/cmd/buf/command/format/format.go
@@ -323,14 +323,15 @@ func run(
 	}
 
 	diffBuffer := bytes.NewBuffer(nil)
-	if err := storage.Diff(
+	changedPaths, err := storage.DiffWithFilenames(
 		ctx,
 		runner,
 		diffBuffer,
 		originalReadBucket,
 		formattedReadBucket,
 		storage.DiffWithExternalPaths(), // No need to set prefixes as the buckets are from the same location.
-	); err != nil {
+	)
+	if err != nil {
 		return err
 	}
 	diffExists := diffBuffer.Len() > 0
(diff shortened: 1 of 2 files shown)


private/buf/cmd/buf/command/format/format.go
```diff
@@ -346,17 +347,25 @@ func run(
...
 			}
 		}
-		// If we haven't overridden the output flag and havent set write, we can stop here.
+		// If we haven't overridden the output flag and haven't set write, we can stop here.
 		if flags.Output == "-" && !flags.Write {
 			return nil
 		}
 	}
 	if flags.Write {
+		changedPathSet := make(map[string]struct{}, len(changedPaths))
```
bufdev, line 356: nit: you can use `slicesext.ToStructMap(changedPaths)`

private/pkg/storage/diff.go
```diff
@@ -238,16 +263,16 @@ func Diff(
...
 			if len(diffData) > 0 {
 				if _, err := writer.Write(diffData); err != nil {
-					return err
+					return nil, err
 				}
 			}
 		}
 	}
-	return nil
+	return changedPaths, nil
```
bufdev, line 275: Can you add a comment that this is sorted? And document that the return value is sorted?

bufdev (approved):
LGTM with two nits

private/pkg/storage/diff.go
jhump: It is not currently sorted. Do you think it should be? The way it is currently used, it does not need to be.

It isn't necessarily sorted because it is constructed in two passes. While each pass is over a sorted slice, the second pass could append items that should be inserted if we need it to be sorted.

private/buf/cmd/buf/command/format/format.go
jhump: done

private/pkg/storage/diff.go
bufdev: I think it should be, to make us really explicit what is being returned. Simple to do anyways

jhump: Okay. Now always sorted and documented, too.

PR #3072, opened 2024-06-11
Fix append calls that should have copied the slice

oliversun9:
There are usages of `append` where the result is not assigned back to the variable. After going through them, most of do not break. In fact, it only breaks things when one slice is stored as a variable or a struct's field, and a prefix of this slice is appended to.

This PR fixes problematic appends in `populateMessage`, which breaks in certain cases (see the test case added).

It also fixes the append in `WalkFileInfos`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,6 @@
 - Update `buf format -w` so that it does not touch files whose contents don't actually
   change. This eliminates noisy notifications to file-system-watcher tools that are
   watching the directory that contains proto sources.
-- Fixes issue where `buf breaking` or `buf lint` might fail when a `.proto` file contains
-  a message or field definition nested deeply in other message definitions.
 
 ## [v1.32.2] - 2024-05-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Update `buf format -w` so that it does not touch files whose contents don't actually
   change. This eliminates noisy notifications to file-system-watcher tools that are
   watching the directory that contains proto sources.
+- Fixes issue where `buf lint` fails when the message is deeply nested.
 
 ## [v1.32.2] - 2024-05-28
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,8 @@
 - Update `buf format -w` so that it does not touch files whose contents don't actually
   change. This eliminates noisy notifications to file-system-watcher tools that are
   watching the directory that contains proto sources.
-- Fixes issue where `buf lint` fails when the message is deeply nested.
+- Fixes issue where `buf breaking` or `buf lint` might fail when a `.proto` file contains
+  a message or field definition nested deeply in other message definitions.
 
 ## [v1.32.2] - 2024-05-28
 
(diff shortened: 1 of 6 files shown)


private/bufpkg/bufmodule/module_read_bucket.go
```diff
@@ -658,7 +658,7 @@ func (t *targetedModuleReadBucket) WalkFileInfos(
 			return fn(fileInfo)
 		},
 		append(
-			options,
+			slicesext.Copy(options),
```
oliversun9, line 661: I don't think this one breaks anything currently. However, I can see how this might break:

```
baseOptions := ...
t.Run("...", func(t, *testing.T) {
  t.Parallel()
  WalkFileInfos(ctx, fn, baseOptions)
})
t.Run("...", func(t, *testing.T) {
  t.Parallel()
  WalkFileInfos(ctx, fn, append(baseOptions, extraOption))
})
```

private/bufpkg/bufprotosource/testdata/nested/foo.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+
+message A {
+  message B {
```
oliversun9, line 4: Hard to break with fewer layers and that is probably because `append` creates a new underlying array more often when the capacity is low.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Update `buf format -w` so that it does not touch files whose contents don't actually
   change. This eliminates noisy notifications to file-system-watcher tools that are
   watching the directory that contains proto sources.
+- Fixes issue where `buf lint` fails when the message is deeply nested.
```
bufdev, line 25: I'm not sure what this means.

oliversun9: Updated in [2440513](https://github.com/bufbuild/buf/pull/3072/commits/24405130b09f52b599c6e03d9039e289104da4a7), I also noticed that it could happen in `buf breaking` as well.

private/pkg/slicesext/slicesext.go
```diff
@@ -150,6 +150,40 @@ func Copy[T any](s []T) []T {
 	return sc
 }
 
+// Grow increases the slice's capacity, if necessary, to guarantee space for
```
oliversun9, line 153: These two functions are copied from https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/slices/slices.go;l=502

pkwarren (commented):
Changes look good to me.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - Update `buf format -w` so that it does not touch files whose contents don't actually
   change. This eliminates noisy notifications to file-system-watcher tools that are
   watching the directory that contains proto sources.
+- Fixes issue where `buf breaking` or `buf lint` might fail when a `.proto` file contains
+  a message or field definition nested deeply in other message definitions.
```
doriable, line 26: I'm not entirely convinced we should have a changelog entry for this, since it doesn't really give users a clear indication of what changed. I think I would remove the changelog entry entirely from this PR.

doriable (commented):
All seems very reasonable to me, left a pretty small comment.

doriable (approved):

pkwarren (approved):

bufdev (approved):

PR #3073, opened 2024-06-11
Only validate min/max editions in code gen response when input uses editions

jhump:
This works around an issue with some pre-v27.0 releases of protoc and the code generators there in (such as Java, C++, Python, etc).

Some of the languages advertised support for features, but in an experimental way, that did _not_ include the minimum and maximum edition fields being set. With v1.32.0+ of buf, code gen always fails because the min/max editions are missing. This changes the validation so that it only occurs when an actual input file uses editions. So if editions aren't used at all, the plugin's output will be used without any error messages.

### DIFF
--- a/private/buf/bufgen/features.go
+++ b/private/buf/bufgen/features.go
@@ -111,9 +111,32 @@ func checkRequiredFeatures(
 				failedFeatures = append(failedFeatures, feature)
 			}
 		}
+		pluginName := configs[responseIndex].Name()
 		if supported&uint64(pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS) != 0 && len(required.editionToFilenames) > 0 {
-			// Plugin supports editions, and files include editions. So make sure
-			// the plugin supports precisely the right editions.
+			// Plugin supports editions, and files include editions.
+			// First, let's make sure that the plugin set the min/max edition fields correctly.
+			if response.MinimumEdition == nil {
+				return fmt.Errorf(
+					"plugin %q advertises that it supports editions but did not indicate a minimum supported edition",
+					pluginName,
+				)
+			}
+			if response.MaximumEdition == nil {
+				return fmt.Errorf(
+					"plugin %q advertises that it supports editions but did not indicate a maximum supported edition",
+					pluginName,
+				)
+			}
+			if response.GetMaximumEdition() < response.GetMinimumEdition() {
+				return fmt.Errorf(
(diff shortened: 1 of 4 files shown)


private/buf/bufgen/features.go
```diff
@@ -111,9 +111,32 @@ func checkRequiredFeatures(
 				failedFeatures = append(failedFeatures, feature)
 			}
 		}
+		pluginName := configs[responseIndex].Name()
 		if supported&uint64(pluginpb.CodeGeneratorResponse_FEATURE_SUPPORTS_EDITIONS) != 0 && len(required.editionToFilenames) > 0 {
-			// Plugin supports editions, and files include editions. So make sure
-			// the plugin supports precisely the right editions.
+			// Plugin supports editions, and files include editions.
```
pkwarren, line 116: What if `len(required.editionToFilenames) > 0` but the plugin response *doesn't* indicate it supports editions?

private/buf/bufgen/features.go
```diff
@@ -111,9 +111,32 @@ func checkRequiredFeatures(
...
+				)
+			}
+			if response.MaximumEdition == nil {
+				return fmt.Errorf(
+					"plugin %q advertises that it supports editions but did not indicate a maximum supported edition",
+					pluginName,
+				)
+			}
+			if response.GetMaximumEdition() < response.GetMinimumEdition() {
+				return fmt.Errorf(
```
pkwarren, line 131: nit: Would we ever want to use `multierr` here to return all of the errors to the user (so they could fix everything wrong w/ the plugin) instead of individually?

private/buf/bufgen/features.go
jhump: That is already covered. This is the test case for it: https://github.com/bufbuild/buf/pull/3073/files#diff-8ab9fea8f32ab6d954c3bcfd1ca0d38d6327c57ca60e571d3d61ea914368b690R128-R134

pkwarren: Missed that when looking through the code - good to know.

pkwarren (approved):
Verified this fixes the issue with older plugins.

bufdev (approved):

doriable (approved):

private/buf/bufgen/features.go
jhump: I don't think so. We only want to validate the actual edition is supported if the response's min and max are valid. And we only want to do this check if both min and max are present. So I don't think they combine in a way that makes a multi error that useful.

PR #3075, opened 2024-06-12
Add exclude unnamed flag for push

emcfarlane:
This PR adds the flag `--exclude-unnamed` for `push`. On `push`, this flag will allow for unnamed modules to be present in the `buf.yaml` config. Modules are filtered by name before pushing to the remote. Local only modules, modules without a name, are excluded. All dependencies for a named module must have a valid name.

As an example the following config will work with `buf push --exclude-unnamed` as long as the module `foo` does not depend on `bar`:
```yaml
version: v2
modules:
  - path: foo
    name: buf.build/example/foo
  - path: bar
```

### DIFF
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -47,7 +47,7 @@ const (
 	createDefaultLabelFlagName = "create-default-label"
 	sourceControlURLFlagName   = "source-control-url"
 	gitMetadataFlagName        = "git-metadata"
-	skipUnnamedFlagName        = "skip-unnamed"
+	excludeUnnamedFlagName     = "exclude-unnamed"
 
 	// All deprecated.
 	tagFlagName      = "tag"
@@ -93,7 +93,7 @@ type flags struct {
 	CreateVisibility   string
 	CreateDefaultLabel string
 	SourceControlURL   string
-	SkipUnnamed        bool
+	ExcludeUnnamed     bool
 	GitMetadata        bool
 	// special
 	InputHashtag string
@@ -168,8 +168,8 @@ If you set the --%s flag and/or --%s flag yourself, then the value(s) will be us
 		),
 	)
 	flagSet.BoolVar(
-		&f.SkipUnnamed,
-		skipUnnamedFlagName,
+		&f.ExcludeUnnamed,
+		excludeUnnamedFlagName,
 		false,
 		"Only push named modules to the BSR. Named modules must not have any unnamed dependencies.",
 	)
@@ -236,8 +236,8 @@ func run(
 	if flags.SourceControlURL != "" {
 		uploadOptions = append(uploadOptions, bufmodule.UploadWithSourceControlURL(flags.SourceControlURL))
 	}
-	if flags.SkipUnnamed {
-		uploadOptions = append(uploadOptions, bufmodule.UploadWithSkipUnnamed())
+	if flags.ExcludeUnnamed {
+		uploadOptions = append(uploadOptions, bufmodule.UploadWithExcludeUnnamed())
 	}
 
 	commits, err := uploader.Upload(ctx, workspace, uploadOptions...)
--- a/private/buf/cmd/buf/command/push/push.go
+++ b/private/buf/cmd/buf/command/push/push.go
@@ -47,7 +47,7 @@ const (
 	createDefaultLabelFlagName = "create-default-label"
 	sourceControlURLFlagName   = "source-control-url"
 	gitMetadataFlagName        = "git-metadata"
-	ignoreUnnamedFlagName      = "ignore-unnamed"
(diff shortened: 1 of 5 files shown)


private/buf/cmd/buf/command/push/push.go
```diff
@@ -165,6 +167,12 @@ If you set the --%s flag and/or --%s flag yourself, then the value(s) will be us
 			createDefaultLabelFlagName,
 		),
 	)
+	flagSet.BoolVar(
+		&f.IgnoredUnnamed,
+		ignoreUnnamedFlagName,
+		false,
+		"Ignore unnamed modules. Unnamed modules must not be depended on by any named modules.",
```
doriable, line 174: > Unnamed modules must not be depended on by any named modules.

Is this true? It seems that modules that are being pushed cannot depend on unnamed modules, but not the other way around?

emcfarlane: I think thats correct but wording is confusing. What about?

> Named modules must not have any unnamed dependencies.

doriable: Oh yes, that is a lot less confusing to me ^^"

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -107,6 +108,38 @@ func (a *uploader) Upload(
 	if err != nil {
 		return nil, err
 	}
+	// Only push named modules to the registry. Any dependencies for named modules must have a name.
+	// Local unnamed modules can be skippedl if the UploadWithSkipUnnamed option is set.
```
nicksnyder, line 112:
```suggestion
	// Local unnamed modules can be skipped if the UploadWithSkipUnnamed option is set.
```

private/buf/cmd/buf/command/push/push.go
```diff
@@ -165,6 +167,12 @@ If you set the --%s flag and/or --%s flag yourself, then the value(s) will be us
 			createDefaultLabelFlagName,
 		),
 	)
+	flagSet.BoolVar(
+		&f.SkipUnnamed,
+		skipUnnamedFlagName,
+		false,
+		"Skip unnamed modules will only push named modules. Named modules must not have any unnamed dependencies.",
```
nicksnyder, line 174: This doc comment is kind of confusing to parse. What about this?
```suggestion
		"Only push named modules to the BSR. Named modules must not have any unnamed dependencies.",
```

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -107,6 +108,38 @@ func (a *uploader) Upload(
...
+		if err != nil {
+			return false, err
+		}
+		if err := slicesext.Reduce(deps, func(err error, dep bufmodule.ModuleDep) error {
+			if moduleName := dep.ModuleFullName(); moduleName == nil {
+				return multierr.Append(err, fmt.Errorf("dependency %q does not have a name", dep.OpaqueID()))
+			}
+			return err
+		}, nil); err != nil {
+			return false, fmt.Errorf("All dependencies for module %s must have a name: %w.", moduleName.String(), err)
```
bufdev, line 132: Using multierr here is going to result in a not-fantastic actual error message for the end-user, it'd be better if we were able to print the string `all dependencies for module %q must be named but modules %s had no name", moduleName.String(), strings.Join(allDepModuleOpaqueIDs, ", "`

private/buf/cmd/buf/command/push/push.go
```diff
@@ -92,6 +93,7 @@ type flags struct {
 	CreateVisibility   string
 	CreateDefaultLabel string
 	SourceControlURL   string
+	SkipUnnamed        bool
```
bufdev, line 96: On second thought, I actually think we should call this `--exclude-unnamed` as we use `exclude` as the verb everywhere else.

emcfarlane: Renamed to exclude.

doriable (approved):
This looks reasonable to me :) Just putting a note here for the future, we are currently using `OpaqueID` in several places now for modules without names -- this is reasonable to me, but at some point we'll need to solidify the semantics for `OpaqueID` and adjust the docs (since the docs imply that it should not be relied upon, but we definitely do in these cases).

bufdev:
> since the docs imply that it should not be relied upon, but we definitely do in these cases

The docs definitely should be tightened up, but just a note that they said its *structure* should not be relied upon, not that it shouldn't be relied upon as a whole. However, we should document that it can be used as a human-readable identifier of the Module.

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
bufdev: This sounds right - thanks for being consistent!

bufdev (approved):

PR #3090, opened 2024-06-17
Fix missing insertion point error handling

emcfarlane:
On missing an insertion point, fail with a user facing error and avoid panic'ing due to an invalid slice when the insertion point file is empty.

Fixes #1649

From the issue, `buf` will now correctly return an error like:
```
Failure: plugin prost-crate: could not find insertion point "features" in "Cargo.toml"
```

PR #3092, opened 2024-06-17
Improve insertion point behaviour

emcfarlane:
Small improvements on creating insertion points. Removes the need to trim the resulting file for newlines. Removes a buffer allocation to build each insertion.

### DIFF
--- a/private/bufpkg/bufprotoplugin/response_writer.go
+++ b/private/bufpkg/bufprotoplugin/response_writer.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"io/fs"
 	"unicode"
@@ -108,6 +109,7 @@ func writeInsertionPoint(
 	// doing 2 full scans of the file (if it is a single line), or implementing
 	// bufio.Scanner.Scan() inline
 	newline := []byte{'\n'}
+	var found bool
 	for targetScanner.Scan() {
 		targetLine := targetScanner.Bytes()
 		if !bytes.Contains(targetLine, match) {
@@ -137,15 +139,19 @@ func writeInsertionPoint(
 		// allocate
 		_, _ = postInsertionContent.Write(targetLine)
 		_, _ = postInsertionContent.Write(newline)
+		found = true
 	}
 
 	if err := targetScanner.Err(); err != nil {
 		return nil, err
 	}
 
+	if !found {
+		return nil, fmt.Errorf("could not find insertion point %q in %q", insertionPointFile.GetInsertionPoint(), insertionPointFile.GetName())
+	}
 	// trim the trailing newline
 	postInsertionBytes := postInsertionContent.Bytes()
-	return postInsertionBytes[:len(postInsertionBytes)-1], nil
+	return bytes.TrimSuffix(postInsertionBytes, newline), nil
 }
 
 // leadingWhitespace iterates through the given string,
--- a/private/bufpkg/bufprotoplugin/response_writer.go
+++ b/private/bufpkg/bufprotoplugin/response_writer.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"strings"
 	"unicode"
 	"unicode/utf8"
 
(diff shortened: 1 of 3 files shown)


private/bufpkg/bufprotoplugin/response_writer.go
```diff
@@ -110,48 +111,43 @@ func writeInsertionPoint(
 	// bufio.Scanner.Scan() inline
 	newline := []byte{'\n'}
 	var found bool
-	for targetScanner.Scan() {
+	for i := 0; targetScanner.Scan(); i++ {
+		if i > 0 {
+			// These writes cannot fail, they will panic if they cannot allocate.
+			_, _ = postInsertionContent.Write(newline)
```
emcfarlane, line 117: Create a newline only if we are going to append more content to the file. This removes the need to trim after the buffer is finalized.

private/bufpkg/bufprotoplugin/response_writer.go
```diff
@@ -110,48 +111,43 @@ func writeInsertionPoint(
...
-		// Create another scanner so that we can seamlessly handle
-		// newlines in a platform-agnostic manner.
-		insertedContentScanner := bufio.NewScanner(bytes.NewBufferString(insertionPointFile.GetContent()))
-		insertedContent := scanWithPrefixAndLineEnding(insertedContentScanner, whitespace, newline)
-		// This write cannot fail, it will panic if it cannot
-		// allocate
-		_, _ = postInsertionContent.Write(insertedContent)
+		// Insert the content from the insertion point file. Handle newlines in
+		// a platform-agnostic manner.
+		insertedContentReader := strings.NewReader(insertionPointFile.GetContent())
```
emcfarlane, line 133: Prefer `strings.NewReader` over `bytes.NewBufferString` as the string is RO.

private/bufpkg/bufprotoplugin/response_writer.go
```diff
@@ -110,48 +111,43 @@ func writeInsertionPoint(
...
+		writeWithPrefixAndLineEnding(postInsertionContent, insertedContentReader, whitespace, newline)
```
emcfarlane, line 134: Write to the buffer instead of allocating a new slice and then writing to the buffer.

private/bufpkg/bufprotoplugin/response_writer.go
```diff
@@ -110,48 +111,43 @@ func writeInsertionPoint(
...
 
 		// Code inserted at this point is placed immediately
 		// above the line containing the insertion point, so
 		// we include it last.
 		// These writes cannot fail, they will panic if they cannot
 		// allocate
 		_, _ = postInsertionContent.Write(targetLine)
-		_, _ = postInsertionContent.Write(newline)
```
doriable, line 141: Do we not need a new line write before the `targetLine`, similar to the non-matching case here?

```
		if i > 0 {
			// These writes cannot fail, they will panic if they cannot allocate.
			_, _ = postInsertionContent.Write(newline)
		}
```

emcfarlane: I don't think so. Instead of always appending a trailing newline so the next content starts at a newline, we reverse that logic to append a newline if not the first line. This removes the need to strip the newline when we finished writing. Target line will be on a newline as the insert point above has a trailing newline.

doriable: Ah I see, we did that outside of the conditional, okay cool 👍🏼

doriable (approved):
Seems reasonable to me, I think these are some nice improvements.

bufdev (commented):
How is this currently tested? Are you confident that the new code is being exercised?

I'm hesitant to change 3-4 year old code whose performance doesn't matter at all, that has been working, just for a perf improvement, unless we're confident that it is fully tested (maybe it is).

emcfarlane:
Tests are [here](https://github.com/bufbuild/buf/blob/fc8e51c70c671f92703c415484ecb0d8a7f5a076/private/bufpkg/bufprotoplugin/bufprotoplugin_test.go#L27) with some benchmarks that show the improvements. #3090 improved the coverage for the error cases captured in that issue. Agree that the benchmarks don't matter for this flow but thought this simplified it and removed the workarounds for both newline printing, that caused the original panic issue, and the need to calculate a buffer size.

bufdev (approved):

PR #3094, opened 2024-06-18
Set Editions-related fields in CodeGeneratorResponse when proxying to protoc

jhump:
v27 of `protoc` fully supports Editions, but when using the buf CLI and proxying to it (for generating Java, C++, Python, etc), it refuses to accept Editions. The reason is that buf CLI's proxying code wasn't setting the Editions-related fields in the code generator response. This fixes that.

This also updates `buf alpha protoc --version` to print `27.0-buf` instead of `5.27.0-buf`. This mirrors `protoc`, which prints `libprotoc 27.0`. (The "5" major version is the backwards-compatibility level of the C++ runtime and `protoc` stopped reporting that initial extra version number starting in v21, which was v3.21 of the C++ runtime.)

### DIFF
--- a/private/buf/bufprotopluginexec/version.go
+++ b/private/buf/bufprotopluginexec/version.go
@@ -21,6 +21,7 @@ import (
 	"strings"
 
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/pluginpb"
 )
 
@@ -118,6 +119,23 @@ func getFeatureProto3OptionalSupported(version *pluginpb.Version) bool {
 	}
 	// version.GetMajor() > 3
 	return true
+}
+
+// Should I notify that I am OK with editions (and, if so, which ones)?
+func getFeatureEditionsSupported(version *pluginpb.Version) (supported bool, min, max descriptorpb.Edition) {
+	if version.GetMajor() > 5 || (version.GetMajor() == 5 && version.GetMinor() >= 27) {
+		// TODO: Update this to include later editions as they are supported in later versions of protoc.
+		return true, descriptorpb.Edition_EDITION_2023, descriptorpb.Edition_EDITION_2023
+	}
+	return false, 0, 0
+	// TODO: We will likely want to add a getSetExperimentalEditionsFlag() in the future.
+	//       But we don't need it now because the versions in which it was available
+	//       (v24.0 - v26.x) are not really suitable for using it since the implementation
+	//       of Editions actually underwent considerable changes between v26 and the final
+	//       product in v27. So the earlier experimental support is not desirable to enable.
+	//       Most of the changes were in Editions itself, not in Edition 2023. So future
+	//       editions may go smoother, so it may be useful to enable experimental editions
+	//       for code gen in future versions of protoc.
 }
 
 // Is kotlin supported as a builtin plugin?
--- a/private/buf/bufprotopluginexec/version.go
+++ b/private/buf/bufprotopluginexec/version.go
@@ -122,12 +122,20 @@ func getFeatureProto3OptionalSupported(version *pluginpb.Version) bool {
 }
 
 // Should I notify that I am OK with editions (and, if so, which ones)?
-func getFeatureEditionsSupported(version *pluginpb.Version) (supported bool, min, max descriptorpb.Edition) {
-	if version.GetMajor() > 5 || (version.GetMajor() == 5 && version.GetMinor() >= 27) {
-		// TODO: Update this to include later editions as they are supported in later versions of protoc.
-		return true, descriptorpb.Edition_EDITION_2023, descriptorpb.Edition_EDITION_2023
+func getFeatureEditionsSupported(version *pluginpb.Version) (supported bool, max descriptorpb.Edition) {
+	majorVersion := version.GetMajor()
+	if version.GetSuffix() == "buf" && majorVersion == 5 {
+		// Buf reported 5.27.0 instead of 27.0 in v1.32 and v1.33.
+		majorVersion = version.GetMinor()
 	}
-	return false, 0, 0
+
+	if majorVersion < 27 {
+		return false, 0
+	}
+
+	// TODO: Update this to include later editions as they are supported in later versions of protoc.
+	return true, descriptorpb.Edition_EDITION_2023
+
 	// TODO: We will likely want to add a getSetExperimentalEditionsFlag() in the future.
 	//       But we don't need it now because the versions in which it was available
 	//       (v24.0 - v26.x) are not really suitable for using it since the implementation
--- a/private/buf/bufprotopluginexec/version_test.go
+++ b/private/buf/bufprotopluginexec/version_test.go
@@ -58,29 +58,27 @@ func TestGetFeatureProto3OptionalSupported(t *testing.T) {
 
 func TestGetFeatureEditionsSupported(t *testing.T) {
 	t.Parallel()
-	assertEditionsNotSupported(t, newVersion(2, 12, 4, ""))
-	assertEditionsNotSupported(t, newVersion(3, 11, 4, ""))
-	assertEditionsNotSupported(t, newVersion(4, 21, 1, ""))
-	assertEditionsNotSupported(t, newVersion(4, 21, 1, "buf"))
-	assertEditionsNotSupported(t, newVersion(5, 25, 0, ""))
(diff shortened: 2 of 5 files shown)


private/buf/bufprotopluginexec/version.go
```diff
@@ -120,6 +121,23 @@ func getFeatureProto3OptionalSupported(version *pluginpb.Version) bool {
 	return true
 }
 
+// Should I notify that I am OK with editions (and, if so, which ones)?
+func getFeatureEditionsSupported(version *pluginpb.Version) (supported bool, min, max descriptorpb.Edition) {
+	if version.GetMajor() > 5 || (version.GetMajor() == 5 && version.GetMinor() >= 27) {
```
pkwarren, line 126: When testing locally with protoc 27.1, I saw the major version set to 27 and the minor version set to 1.

private/buf/bufprotopluginexec/version.go
```diff
@@ -120,6 +121,23 @@ func getFeatureProto3OptionalSupported(version *pluginpb.Version) bool {
...
+		// TODO: Update this to include later editions as they are supported in later versions of protoc.
+		return true, descriptorpb.Edition_EDITION_2023, descriptorpb.Edition_EDITION_2023
```
pkwarren, line 128: Shouldn't the min edition be proto2 to match https://github.com/protocolbuffers/protobuf/blob/a978b75794a6ce4547c9db08a115c458d9190934/src/google/protobuf/compiler/cpp/generator.h#L77-L78?

private/buf/bufprotopluginexec/version_test.go
```diff
@@ -55,6 +56,33 @@ func TestGetFeatureProto3OptionalSupported(t *testing.T) {
 	assert.True(t, getFeatureProto3OptionalSupported(newVersion(21, 0, 0, "")))
 }
 
+func TestGetFeatureEditionsSupported(t *testing.T) {
+	t.Parallel()
+	assertEditionsNotSupported(t, newVersion(2, 12, 4, ""))
```
pkwarren, line 61: nit: Maybe overkill since we'd only really need to test < 27 and >= 27 to start. I imagine these will get more interesting over time as new editions are supported.

private/buf/bufprotopluginexec/version.go
jhump: Oof, you're right. I was also looking at the output of `buf alpha protoc --version`, which prints 5.27 :(

I guess we may want to change that, too.

private/buf/bufprotopluginexec/version.go
jhump: That is super weird. It is not necessary to set a min below 2023 because 2023 was the first edition. But I see that other plugins set it that way, too (even `protoc-gen-go`) 🤔 

I have to admit I find that odd since proto3 and proto2 were _not_ editions. Those entries in the enum, from my understanding, were just ways of attributing semantics of features for pre-editions syntaxes (in feature defaults) and not meant to be used this way.

Neither buf nor protoc actually need it to be set that way. For proto2 and proto3 files, these fields in the response are ignored:
https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/compiler/command_line_interface.cc#L2591-L2594
I can change it, for consistency with the original C++ impl, but TBH I find it counter-intuitive 🤷

private/buf/bufprotopluginexec/version.go
jhump: I've updated the logic to handle both the `buf alpha protoc` output and actual `protoc` output and updated the tests accordingly.

I also changed it so that `buf alpha protoc` going forward will omit the C++ runtime major version. So with this PR it would print `27.0-buf` (compared to actual `protoc` printing `libprotoc 27.0`). Previously, it printed `5.27.0-buf`.

private/buf/bufprotopluginexec/version.go
jhump: Updated

pkwarren (approved):

private/buf/bufprotopluginexec/version_test.go
jhump: I left a couple of extra tests, just to sanity check the output from older versions of protoc and buf. Hopefully it's not too bad overkill.

jhump:
@pkwarren, I just realized that since we _don't_ bother setting `--experimental_editions` when we run `protoc`, it will fail separately (with a message about experimental editions) if the inputs include an edition it refuses to support.

So I think that means we can greatly simplify this and just _always_ report `EDITION_MAX`, to advertise this as supporting _all_ editions and let `protoc` itself generate an error if the input have an unsupported edition.

The downside is that the error message will be a little confusing. Also, there's not currently a way to allow the user to specify an extra arg to `protoc` (like to provide the `--experimental_editions` flag). But we _do_ have a way to do that with normal plugins: the `path` property in the YAML file can be a slice of strings instead of a single string. Maybe we should do the same for the `protoc_path` property, to allow users to set that flag?

private/buf/cmd/buf/command/alpha/protoc/protoc.go
```diff
@@ -73,8 +73,10 @@ Additional flags:
 		BindFlags:     flagsBuilder.Bind,
 		NormalizeFlag: flagsBuilder.Normalize,
 		Version: fmt.Sprintf(
-			"%v.%v.%v-buf",
-			bufprotopluginexec.DefaultVersion.GetMajor(),
+			"%v.%v-buf",
+			// DefaultVersion has an extra major version that corresponds to
+			// backwards-compatibility level of C++ runtime. The actual version
+			// of the compiler is just the minor and patch versions.
```
pkwarren, line 79: I checked historical versions of protoc and everything after 22.0 doesn't use the C++ runtime version. Do we want `bufprotopluginexec.DefaultVersion` to update to match?

```
$ ./protoc-versions.sh                    
libprotoc 3.21.0
libprotoc 3.21.1
libprotoc 3.21.2
libprotoc 3.21.3
libprotoc 3.21.4
libprotoc 3.21.5
libprotoc 3.21.6
libprotoc 3.21.7
libprotoc 3.21.8
libprotoc 3.21.9
libprotoc 3.21.10
libprotoc 3.21.11
libprotoc 3.21.12
libprotoc 22.0
libprotoc 22.1
libprotoc 22.2
libprotoc 22.3
libprotoc 22.4
libprotoc 22.5
libprotoc 23.0
libprotoc 23.1
libprotoc 23.2
libprotoc 23.3
libprotoc 23.4
libprotoc 24.0
libprotoc 24.1
libprotoc 24.2
libprotoc 24.3
libprotoc 24.4
libprotoc 25.0
libprotoc 25.1
libprotoc 25.2
libprotoc 25.3
libprotoc 26.0
libprotoc 26.1
libprotoc 27.0
libprotoc 27.1
```

jhump: > Do we want bufprotopluginexec.DefaultVersion to update to match?

I don't believe so because that version is also provided to plugins from `buf alpha protoc`, and IIUC in that case `protoc` _does_ include the extra major version number.

jhump: But I could be wrong. I'm basing that based on the fact that if you change `DefaultVersion` to not include that, you'll get test failures because the extra digit is expected. IIUC, those tests are actually running `protoc` and expecting the output to match that of `buf alpha protoc`. So I believe that means that the major version 5 is passed to plugins in the version number in the code gen request.

pkwarren: I *believe* I've seen the major version number change based on the plugin. For example, Java generated code now uses `4.27.1`. C++ and Python use `5.27.1`.

pkwarren: Regardless - let's leave this for now in this PR. Resolving.

jhump: That's kind of funny that 3.21 is reported with the 3 major version since the GitHub release does _not_ have it: it's just [v21.0](https://github.com/protocolbuffers/protobuf/releases/tag/v21.0). I guess they forgot to update the command-line version output until v22. Doh!

jhump:
Take a look at #3098 as a possible alternative.

bufdev:
Approved #3098, assuming that usurps this PR.

jhump:
Closing this. Will merge the alternative in #3098 instead.

Issue #2463, opened 2023-09-25
Buf breaking: fatal: .git/index: index file open failed: Not a directory


jalaziz:
I'm trying to use the buf breaking hook with pre-commit and I am running into an odd problem that I have not been able to debug.

My `pre-commit` config looks like:
```
  - repo: https://github.com/bufbuild/buf
    rev: v1.26.1
    hooks:
      - id: buf-generate
        args: ['--template', 'api/proto/buf.gen.yaml', '--output', 'api/proto/']
        files: api/proto/
      - id: buf-breaking
        args: ['--against', '.git#branch=main,recurse_submodules=true']
      - id: buf-lint
      - id: buf-format
      - id: buf-mod-update
        args: ['api/proto']
        files: api/proto/
      - id: buf-mod-prune
        args: ['api/proto']
        files: api/proto/
```

When run with `pre-commit run --all-files` everything works fine.

However, when I attempt to commit changes that affect proto files, buf breaking fails with:
```
buf breaking.............................................................Failed
- hook id: buf-breaking
- exit code: 1

Failure: could not clone file:///Users/jalaziz/dev/foo/.git: exit status 128
Preparing worktree (detached HEAD 10f9f8b)
fatal: .git/index: index file open failed: Not a directory
```

ddadonJoin:
I have the same problem. @jalaziz Did you find a workaroud?

jalaziz:
> I have the same problem. @jalaziz Did you find a workaroud?

@ddadonJoin Unfortunately not 😞. I had to disable it for now.

berendjan:
i have the same issue, i think its because the repo is cloned with --bare.  So there is no .git directory, instead the index file is in de root. Havent tested it though

strowk:
I sort of figured why this happens. The culprit is env variable GIT_INDEX_FILE which is normally set to ".git/index" when hooks are running.

See in https://git-scm.com/docs/githooks :
> Environment variables, such as GIT_DIR, GIT_WORK_TREE, etc., are exported so that Git commands run by the hook can correctly locate the repository. If your hook needs to invoke Git commands in a foreign repository or in a different working tree of the same repository, then it should clear these environment variables so they do not interfere with Git operations at the foreign location

This "If your hook needs to invoke Git commands in a foreign repository" applies to buf, as it creates temp bare repo and adds there a worktree, which is where it currently breaks for me due to GIT_INDEX_FILE pointing to that relative path ".git/index".
[...]

Issue #419, opened 2021-08-13
Error when checking breaking change against previous ref of current branch


chronitis:
This was encountered using `buf-breaking-action` (using the recommended syntax as of 0.4.0 for checking against a previous commit on the same branch), but can be replicated locally, so I suspect is a regression in `buf`, not the action.

When trying to check breaking changes against a previous revision of the current branch, as of 0.50.0 I get an error:

```bash
// on branch master
buf breaking --against 'https://github.com/org/repo.git#branch=master,ref=HEAD~1'
> Failure: could not clone https://github.com/org/repo.git: exit status 128
> fatal: Refusing to fetch into current branch refs/heads/master of non-bare repository.
```

This also happens trying to refer to the same branch in the local repository:

```bash
buf breaking --against '.git#branch=master,ref=HEAD~1'
> Failure: could not clone file:///path/to/repo/.git: exit status 128
> fatal: Refusing to fetch into current branch refs/heads/master of non-bare repository
> fatal: the remote end hung up unexpectedly.
```

Checking previous versions, this regression is first seen with 0.47.0; presumably due to

> Rewrite the git cloner to use git init && git fetch rather than git clone. git clone is limited to local branches on the remote, whereas git fetch we can fetch any references on the remote including remote branches.

Git version 2.25.1, in case that is relevant.

bufdev:
This is fixed in #421. This will go out in the next release. Until then, you can use `go get github.com/bufbuild/buf/cmd/buf@master`. Thanks for the bug report!

bufdev:
This is released on v0.51.0.

PR #3095, opened 2024-06-18
Fix git ref to allow branches and tags

### LINKED ISSUE #2932: Git inputs cannot reference tags with the ref option
As far as I understand, git branches and tags are types of git refs (they are stored under `.git/refs`).

IMO it should therefore be possible to use the `ref` option as a generic way for referencing commits, branches, and tags.

This is actually stated in the documentation:

>Any ref that is a valid input to git checkout is accepted.

https://buf.build/docs/reference/inputs#other-options

However, while `git checkout remote-tag` is working, `buf generate "ssh://git@github.com/path-to-repo.git#ref=remote-tag"` is not.

IMO this feature would make parameterizing CI pipelines easier, e.g.: 

- `ssh://git@github.com/path-to-repo.git#ref=develop` - build against latest changes on the develop branch
- `ssh://git@github.com/path-to-repo.git#ref=v1.0.0` - build against a stable release tag

Note: this is also how kubectl kustomize is implementing the `ref`option:

> ref - a [git fetch-able ref](https://git-scm.com/docs/git-fetch), typically a branch, tag, or full commit hash (short hashes are not supported)

https://github.com/kubernetes-sigs/kustomize/blob/master/examples/remoteBuild.md#remote-directories

emcfarlane:
This changes the git ref input to align with the git notion of a ref as a tag or branch. We now always try to fetch a given ref before falling back to fetching the HEAD to resolve partial refs. If both a ref and branch are provided, the branch is fetched directly before resolving the ref. The inputs `tag` and `commit` can still be used but can now be replaced by setting the `ref`.

For example:
 - Branch `ref=main` will fetch and checkout the `main` branch
 - Tag `ref=v1.0.0` will fetch and checkout the `v1.0.0` tag
 - Partial `ref=123456` will fetch `HEAD` and checkout `123456`
 - Branch with partial `ref=123456,branch=feature` will fetch `feature` and checkout `123456`
 - Commit `ref=<sha>` will fetch `<sha>` and checkout `<sha>`
 - Branch with commit `ref=<sha>,branch=feature` with fetch `feature` and checkout `<sha>`

Fixes #2932


private/pkg/git/cloner.go
```diff
@@ -97,51 +88,38 @@ func (c *cloner) CloneToBucket(
...
-	defer func() {
-		retErr = multierr.Append(retErr, worktreeDir.Close())
+		retErr = multierr.Append(retErr, baseDir.Close())
 	}()
 
 	buffer := bytes.NewBuffer(nil)
 	if err := c.runner.Run(
 		ctx,
 		"git",
-		command.RunWithArgs("init", "--bare"),
```
emcfarlane, line 121: Removed `--bare` and the below `worktree` in favour of simply checking out the repository into this directory. Are there any requirements for the worktree flow?

private/pkg/git/cloner.go
```diff
@@ -154,97 +132,102 @@ func (c *cloner) CloneToBucket(
...
-		"worktree",
-		"add",
-		worktreeDir.AbsPath(),
-		worktreeRef,
-	)
 	if err := c.runner.Run(
 		ctx,
 		"git",
-		command.RunWithArgs(args...),
+		command.RunWithArgs("checkout", "--force", "FETCH_HEAD"),
```
emcfarlane, line 185: `--force` was added as a safeguard following github actions checkout.

private/pkg/git/cloner.go
```diff
@@ -344,41 +327,39 @@ func getSSHKnownHostsFilePaths(sshKnownHostsFiles string) []string {
...
-		// After fetch we won't have checked out any refs. This
-		// will cause `refs=` containing "HEAD" to fail, as HEAD
-		// is a special case that is not fetched into a ref but
-		// instead refers to the current commit checked out. By
-		// checking out "FETCH_HEAD" before checking out the
-		// user supplied ref, we behave similarly to git clone.
-		return "HEAD", "FETCH_HEAD", gitName.checkout()
-	} else {
-		return "HEAD", "FETCH_HEAD", ""
+	checkout, cloneBranch := gitName.checkout(), gitName.cloneBranch()
```
emcfarlane, line 339: `cloneBranch` is misnamed, this can be a branch, tag or commit.

private/pkg/git/cloner.go
emcfarlane: Need to check this is supported: https://github.com/bufbuild/buf/pull/421

bufdev (approved):
I am very out of date on this package, so assume I approve, and get a review from someone else more familiar with git.

private/pkg/git/cloner.go
doriable: Just tested this behaviour, it seems that this does break #419. Let's chat about this and figure something out.

doriable:
We cannot currently merge this since it breaks the behaviour for #419 -- we'll need to investigate a solution that continues to work with that.

private/pkg/git/cloner.go
emcfarlane: I tested on this repo with the following and the commands work as intended:
```bash
$ buf breaking --against 'https://github.com/bufbuild/buf.git#branch=main,ref=HEAD~1'
$ buf breaking --against '.git#branch=main,ref=HEAD~1'
```
For the issue #2463 I created the following `pre-commit` hook:
```
repos:
  - repo: https://github.com/bufbuild/buf
    rev: ed/gitRefResolve
    hooks:
      - id: buf-breaking
        args: ['--against', '.git#branch=main,recurse_submodules=true']
```
which when creating a breaking change seems to be correct:
```
$ git commit
[WARNING] The 'rev' field of repo 'https://github.com/bufbuild/buf' appears to be a mutable reference (moving tag / branch).  Mutable references are never updated after first install and are not supported.  See https://pre-commit.com/#using-the-latest-version-for-a-repository for more details.  Hint: `pre-commit autoupdate` often fixes this.
buf breaking.............................................................Failed
- hook id: buf-breaking
- exit code: 100

proto/acme/example/v1/book.proto:8:1:Previously present field "4" with name "read_status" on message "Book" was deleted.
```
@doriable what was the failure cases you saw?

doriable: So when running the case with `.git#branch=main,ref=HEAD~1`, I get the following (scrubbed the full path of the `/private` directory`):

```
$ buf breaking --against '.git#branch=main,ref=HEAD~1'
Failure: could not clone file:///Users/dkeung/src/buf/buf/.git: exit status 128
fatal: refusing to fetch into branch 'refs/heads/main' checked out at '/private/var/folders/...'
```

doriable: Commenting to keep track of offline discussion: seems like this has to do with the version of Git I had installed, `git version 2.39.3 (Apple Git-146)`. When I swapped to a later version installed through Homebrew, `git version 2.45.2`, it works. We'll need to investigate this for versions of Git.

emcfarlane: Workaround to add `--update-head-ok`. This looks only to be needed by Apple versions of `git`. Testing against `git` v2.39.3 installed from source is fine. But the corresponding `git version 2.39.3 (Apple Git-146)` causes this failure.

emcfarlane: Reason for failure and noticing it only with apple git versions is due to the refspec. Apple git defaults to `main` while normal `git` init branch is master. The refspec of `+main:main` will cause a failure if the current branch is `main` on `git fetch` due to updating the `HEAD` to main, which fetch intentionally disallows. To fix this we now add the `--update-head-ok`, this behaviour is desired and intentional.

doriable (approved):

marcus-griep-simplisafe:
Related to #3265.

emcfarlane:
@marcus-griep-simplisafe updated in the other issue. Working on a fix.

PR #3098, opened 2024-06-18
Allow `protoc_path` to be a slice so user can specify extra args

jhump:
With this, the user can specify `--experimental_editions`. When proxying to `protoc`, we can synthesize support for all editions since `protoc` will actually handle the validation and fail if it can't handle an edition.

This is an alternate to #3094.


doriable:
This seems like a nicer UX to me in the position of a user dealing with experimental editions, since I would already be opting myself into managing something finicky (likely because I need some feature that is coming up with experimental editions, etc.). And this decouples our code paths from managing editions.

private/bufpkg/bufconfig/buf_gen_yaml_file.go
```diff
@@ -496,11 +496,12 @@ type externalGeneratePluginConfigV2 struct {
 	// Local is the local path (either relative or absolute) to a binary or other runnable program which
 	// implements the protoc plugin interface. This can be one string (the program) or multiple (remaining
 	// strings are arguments to the program).
-	Local interface{} `json:"local,omitempty" yaml:"local,omitempty"`
+	Local any `json:"local,omitempty" yaml:"local,omitempty"`
```
pkwarren, line 499: Should we change `Opt` to `any` as well while we're here?

pkwarren (approved):
I like that this approach means we don't have to keep up with every editions version range used by protoc. I don't think we store the CodeGeneratorResponse anywhere either so it should be ok to do this.

private/bufpkg/bufconfig/buf_gen_yaml_file.go
jhump: yeah, good catch

pkwarren (approved):

pkwarren (commented):
Do we need a changelog entry for supporting arguments to protoc_path?

bufdev (approved):

PR #3100, opened 2024-06-19
Wrap commit not found error to display module:commit not found on push

oliversun9:
In `buf push`, only the dependency commit IDs (not the dependency module names) are included in `UploadRequest`. This means if one of the dependency commit IDs is not found, the server can only return an error saying this ID is not found, instead of module:ID not found.

This PR converts ID-not-found errors to module:ID-not-found errors.

Before:
```
Failure: a dependency was not found: shake256/buf.build:2f6ada1cf40c462793a1bc9094e68b66
```
After:
```
Failure: dependency buf.build/foo/bar:2f6ada1cf40c462793a1bc9094e68b66 does not exist
```

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
+++ b/private/bufpkg/bufmodule/bufmoduleapi/uploader.go
@@ -282,7 +282,7 @@ func (a *uploader) Upload(
 			),
 		)
 		if err != nil {
-			return nil, err
+			return nil, asDependencyNotFoundError(err, remoteDeps)
 		}
 		universalProtoCommits, err = slicesext.MapError(response.Msg.Commits, newUniversalProtoCommitForV1Beta1)
 		if err != nil {
@@ -313,7 +313,7 @@ func (a *uploader) Upload(
 			),
 		)
 		if err != nil {
-			return nil, err
+			return nil, asDependencyNotFoundError(err, remoteDeps)
 		}
 		universalProtoCommits, err = slicesext.MapError(response.Msg.Commits, newUniversalProtoCommitForV1)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -494,3 +494,38 @@ func remoteDepToV1Beta1ProtoUploadRequestDepRef(
...
 }
+
+// In UploadRequest, only dependency commit IDs are included (as opposed to module + commit ID), this means
+// when the BSR fails to resolve the commit ID, it can only return an error reporting this ID is not found,
+// without saying which module the commit ID belongs to.
+//
+// asDependencyNotFoundError returns an error reporting moduleFullName:CommitID not found if err
+// is an error reporting the commitID was not found. Otherwise, it returns the err as is.
+func asDependencyNotFoundError(err error, remoteDeps []bufmodule.RemoteDep) error {
+	if connect.CodeOf(err) != connect.CodeInvalidArgument {
```
oliversun9, line 505: Leaving this PR as draft because I'm not sure if the server should return `CodeInvalidArgument` when one of the dependencies is not found.

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -494,3 +494,38 @@ func remoteDepToV1Beta1ProtoUploadRequestDepRef(
...
+		return err
+	}
+	for _, remoteDep := range remoteDeps {
```
bufdev, line 508: I don't quite understand the logic here - this `for` loop exists as soon as it finds the first `RemoteDep` that has both a `ModuleFullName` and a `CommitID`. All `RemoteDeps` have `ModuleFullNames` per the documentation (meaning the `nil` check below is redundant), and I believe that all `RemoteDeps` also have `CommitIDs`. Regardless, this will return an error message for the first `RemoteDep` almost all, or all, of the time. How do we know that the first `RemoteDep` is the reason the not found error was returned? I believe the answer is we do not know that, and this could end up being even more confusing for the user.

If we want to do this, we're going to have to use error details to return what entity was not found.

oliversun9: > exits as soon as it finds the first RemoteDep that has both a ModuleFullName and a CommitID

Actually, on top of this, it also checks whether the error from the server contains this dep's remote name and its commit ID, so

If the remote deps are:
```
buf.build/foo/bar:abcde...
buf.build/bar/baz:12345...
```

and the server error is `a dependency was not found: shake256/buf.build:12345...`

It would not match the first remote dep but the second one.

> we're going to have to use error details to return what entity was not found

I _think_ what I described above uses error details, but I could be wrong.

bufdev: It's not using error details - see https://pkg.go.dev/connectrpc.com/connect#ErrorDetail. Reading the logic again, it seems like you're relying on the string value of the error message - which perhaps is an OK heuristic for now, and still an improvement, but it's not great. You're basically relying on the backend having a human-readable error message that you can parse in the way you want. However, the backend can change this error message at any time - it's not meant to be parsed, and its structure isn't part of the contract.

Even if we do rely on this for now, though, we should get *all* the dependencies not found (ie all found in the string), not just the first one (I'm not sure if the error string from the backend includes all the not-found dependencies, but again, this is the problem with relying on someting without a contract).

Cc @pkwarren @mfridman @doriable who should be able to provide you some more detail here.

mfridman: In retrospect, it would have been nice if the client sent a "ref" that could be used directly in error messages on the server, instead of having to stitch the various bits back on the client.

I.e., instead of sending a bag of `tuuid`'s the client always sent full module references like `buf.build/foo/bar:ttuid`.

But, that's likely irrelevant, and given the batched nature of the API we're probably going to lean more and more into `ErrorDetail`s.


In hindsight, it would have been more efficient if the client sent a complete "ref" that could be used directly in server-side error messages, rather than requiring the client to stitch back together the various components.

For example, instead of transmitting a collection of `tuuid`s, the client could have sent full module references such as `buf.build/foo/bar:ttuid`.

However, this is probably irrelevant now. Given the batched nature of the API, we're likely to lean more on `ErrorDetail`s to avoid string parsing. I mention this, because historically we "banned" error details from escaping externally, so we'll need to remove those filters.

mfridman: @bufdev just confirming, you're okay with BSR returning `ErrorDetail`? Historically we tried to avoid using them, but I think we'll need to lean on them more. Which is a better option than raw string parsing (which you mentioned above).

PR #3102, opened 2024-06-20
Make upgrade

bufdev:
@jhump, there's a build issue relating to features (I ran it locally, it'll appear in CI), can you take a look?


go.mod
```diff
@@ -3,30 +3,30 @@ module github.com/bufbuild/buf
...
-	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.34.1-20240508200655-46a4cf4ba109.1
-	buf.build/gen/go/bufbuild/registry/connectrpc/go v1.16.2-20240606161333-696c2cfeae8c.1
-	buf.build/gen/go/bufbuild/registry/protocolbuffers/go v1.34.1-20240606161333-696c2cfeae8c.1
+	buf.build/gen/go/bufbuild/protovalidate/protocolbuffers/go v1.34.2-20240508200655-46a4cf4ba109.2
+	buf.build/gen/go/bufbuild/registry/connectrpc/go v1.16.2-20240610164129-660609bc46d3.1
+	buf.build/gen/go/bufbuild/registry/protocolbuffers/go v1.34.2-20240610164129-660609bc46d3.2
 	connectrpc.com/connect v1.16.2
 	connectrpc.com/otelconnect v0.7.0
-	github.com/bufbuild/protocompile v0.14.0
+	github.com/bufbuild/protocompile v0.13.1-0.20240510201809-752249dfc37f
```
pkwarren, line 11: Do we need to remove the pinned version of protocompile in make/buf/all.mk?

bufdev: Probably

jhump: For posterity, re-iterating what was said in Slack, this is in fact the issue for the panic that was seen in tests.

This line was reverting protocompile to a point before [this fix](https://github.com/bufbuild/protocompile/pull/306). But, as of #2999, that fix is necessary for the CLI codebase.

pkwarren (approved):

PR #3111, opened 2024-06-25
Set top-level `LintConfig` and `BreakingConfig` and use for image inputs

### LINKED ISSUE #3080: buf 1.32.2 breaking doesn't respect v2 config file when using binpb input
### What's up?

Sorry for lack of test case, but this should be pretty easy to reproduce:

I was trying to set up a check using `buf breaking` like this:

```
buf breaking image1.bin --against image2.bin --config buf.yaml
```

The image files were not generated by buf, they were generated by `protoc -o`

I tried using buf 1.32.2 with a config like:

```
version: v2
breaking:
  use:
    - WIRE
```

This did the breaking check using the default `FILE` level, the `WIRE` level was not respected. I also tried adding `ignore` settings that were not respected.

I downgraded to buf 1.6.0 and used the same command line with the same config except `version: v1`, and it respected the breaking settings as expected.

doriable:
This PR exposes the top-level lint and breaking configs for `buf.yaml` files.

For `v1` `buf.yaml` files, there is only ever one `ModuleConfig`, so the
corresponding lint and breaking configs are used.

For `v2` `buf.yaml` files, if a top-level lint and/or breaking config exists, then
we return those.
Otherwise, `TopLevelLintConfig()`/`TopLevelBreakingConfig` will return `nil`.

We use the top-level lint/breaking configs, if possible, with image inputs, since we
cannot determine which specific `v2` module config to use with images. In
the case where there is no top-level config, we fall back to defaults.

Fixes #3080

### DIFF
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/fs"
 	"net/http"
+	"path/filepath"
 	"sort"
 
 	"github.com/bufbuild/buf/private/buf/buffetch"
@@ -393,23 +394,49 @@ func (c *controller) GetTargetImageWithConfigs(
 			// We did not find a buf.yaml in our current directory, and there was no config override.
 			// Use the defaults.
 		} else {
-			switch fileVersion := bufYAMLFile.FileVersion(); fileVersion {
-			case bufconfig.FileVersionV1Beta1, bufconfig.FileVersionV1:
-				moduleConfigs := bufYAMLFile.ModuleConfigs()
-				if len(moduleConfigs) != 1 {
-					return nil, fmt.Errorf("expected 1 ModuleConfig for FileVersion %v, got %d", len(moduleConfigs), fileVersion)
+			// We take the top-level lint and breaking config from buf.yaml.
+			//
+			// In the case of v1 buf.yaml files, there is only ever one top-level config, so it
+			// should be populated and we use that.
+			// In the case of v2 buf.yaml files, this may be empty, since there may not be a
+			// top-level config. If that happens, we print a warning to the user and let them
+			// know that we are using v2 defaults.
+			// In the case where there is a top-level config and individual module configs, we
+			// print out a message to the user to let them know we are using the top-level config.
+			var configInfo string
+			// If configuration is read from an override, we determine if it was a raw config or
+			// a path to a valid override file.
+			if override := functionOptions.configOverride; override != "" {
+				switch filepath.Ext(override) {
+				case ".json", ".yaml", ".yml":
+					configInfo = functionOptions.configOverride
+				default:
+					configInfo = fmt.Sprintf("from --config")
 				}
-				lintConfig = moduleConfigs[0].LintConfig()
-				breakingConfig = moduleConfigs[0].BreakingConfig()
-			case bufconfig.FileVersionV2:
-				// Use the default LintConfig and BreakingConfig from the file. This is
-				// the top-level lint and/or breaking config(s) if any are set or the default v2
-				// configs. v2 buf.yamls may contain multiple modules, we don't know what lint or
-				// breaking config to apply.
-				lintConfig = bufYAMLFile.DefaultLintConfig()
-				breakingConfig = bufYAMLFile.DefaultBreakingConfig()
-			default:
-				return nil, syserror.Newf("unknown FileVersion: %v", fileVersion)
+			} else {
+				// Otherwise we use the file path
+				configInfo = bufYAMLFile.ObjectData().Name()
+			}
+			if topLevelModuleConfig := bufYAMLFile.TopLevelModuleConfig(); topLevelModuleConfig != nil {
+				if bufYAMLFile.FileVersion() == bufconfig.FileVersionV2 && len(bufYAMLFile.ModuleConfigs()) > 1 {
+					c.logger.Sugar().Warnf(
+						`Configuration %s has multiple modules. The top-level lint/breaking configuration is used with your image, since Buf cannot deduce which specific module configuration to use otherwise.`,
+						configInfo,
+					)
+				}
+				lintConfig = topLevelModuleConfig.LintConfig()
+				breakingConfig = topLevelModuleConfig.BreakingConfig()
+			} else {
+				// Ensure that this is a v2 config
+				if fileVersion := bufYAMLFile.FileVersion(); fileVersion != bufconfig.FileVersionV2 {
+					return nil, syserror.Newf("non-v2 version with no top-level module config: %s", fileVersion)
+				}
+				c.logger.Sugar().Warnf(
+					`Configuration %s does not have a top-level lint/breaking configuration. Buf cannot deduce which specific module configuration to use with your image, so the default configuration is used instead.`,
+					configInfo,
+				)
+				lintConfig = bufconfig.DefaultLintConfigV2
+				breakingConfig = bufconfig.DefaultBreakingConfigV2
 			}
 		}
 		return []ImageWithConfig{
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -378,8 +378,8 @@ func (c *controller) GetTargetImageWithConfigs(
 		if err != nil {
 			return nil, err
 		}
-		lintConfig := bufconfig.DefaultLintConfigV1
-		breakingConfig := bufconfig.DefaultBreakingConfigV1
+		lintConfig := bufconfig.DefaultLintConfigV2
+		breakingConfig := bufconfig.DefaultBreakingConfigV2
 		bufYAMLFile, err := bufconfig.GetBufYAMLFileForPrefixOrOverride(
 			ctx,
 			bucket,
@@ -402,9 +402,12 @@ func (c *controller) GetTargetImageWithConfigs(
 				lintConfig = moduleConfigs[0].LintConfig()
 				breakingConfig = moduleConfigs[0].BreakingConfig()
 			case bufconfig.FileVersionV2:
-				// Do nothing. Use the default LintConfig and BreakingConfig. With
-				// the new buf.yamls with multiple modules, we don't know what lint or
+				// Use the default LintConfig and BreakingConfig from the file. This is
+				// the top-level lint and/or breaking config(s) if any are set or the default v2
+				// configs. v2 buf.yamls may contain multiple modules, we don't know what lint or
 				// breaking config to apply.
+				lintConfig = bufYAMLFile.DefaultLintConfig()
+				breakingConfig = bufYAMLFile.DefaultBreakingConfig()
 			default:
 				return nil, syserror.Newf("unknown FileVersion: %v", fileVersion)
 			}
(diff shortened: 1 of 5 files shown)


doriable:
I've also tested this against the user repro in the issue.

private/buf/bufctl/controller.go
```diff
@@ -378,8 +378,8 @@ func (c *controller) GetTargetImageWithConfigs(
 		if err != nil {
 			return nil, err
 		}
-		lintConfig := bufconfig.DefaultLintConfigV1
-		breakingConfig := bufconfig.DefaultBreakingConfigV1
+		lintConfig := bufconfig.DefaultLintConfigV2
```
pkwarren, line 381: We might want to leave this as V1 here for backwards compatibility (if no config file is found).

I also noticed that it appears that `DefaultLintConfigV2` isn't passing the right value for `allowCommentIgnores` - it should be `true` I think to match the new behavior when parsing from the external config file?

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -275,6 +300,14 @@ func (c *bufYAMLFile) ModuleConfigs() []ModuleConfig {
 	return slicesext.Copy(c.moduleConfigs)
 }
 
+func (c *bufYAMLFile) DefaultLintConfig() LintConfig {
+	return c.defaultLintConfig
+}
+
+func (c *bufYAMLFile) DefaultBreakingConfig() BreakingConfig {
+	return c.defaultBreakingConfig
```
pkwarren, line 308: If these are nil, shouldn't we return the default configs from moduleConfigs[0] as documented in the method docs? I don't think we want to return nil here ever.

private/bufpkg/bufconfig/lint_config.go
```diff
@@ -34,7 +34,7 @@ var (
 		false,
 		false,
 		"",
-		false,
+		true, // We default to allowing comment ignores in v2
```
doriable, line 37: @pkwarren caught an issue with the default `v2` lint config, which should default to allowing comment ignores (https://buf.build/docs/configuration/v2/buf-yaml)

>       # v1 configurations had an allow_comment_ignores option to opt-in to comment ignores.
>       #
>       # In v2, we allow comment ignores by default, and allow opt-out from comment ignores
>       # with the disallow_comment_ignores option.
>       disallow_comment_ignores: false

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -74,6 +74,18 @@ type BufYAMLFile interface {
 	// All ModuleConfigs will have unique ModuleFullNames.
 	// Sorted by DirPath.
 	ModuleConfigs() []ModuleConfig
+	// DefaultLintConfig returns the DefaultLintConfig for the File.
+	//
+	// For v1 buf.yaml files, this will be the default v1 lint config.
```
pkwarren, line 79: Why we wouldn't return the top-level config always for v1 buf.yaml files to match v2 behavior?

private/bufpkg/bufconfig/buf_yaml_file.go
doriable: Hmm, you're right -- yeah, if the constructor is used and `nil` is set, we don't want to return `nil`.

private/buf/bufctl/controller.go
doriable: Yeah, it's fair for back compat to set this back to `v1`. And yes, made an adjustment to `v2` defaults :)

private/bufpkg/bufconfig/buf_yaml_file.go
doriable: Hmm... yeah, I suppose that is correct -- since for `v1` `buf.yaml`'s, there really only is "one top-level" config, and if that isn't set, it just returns default v1, and we should stick to that. Going to adjust the docs and the code accordingly.

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -76,13 +76,13 @@ type BufYAMLFile interface {
 	ModuleConfigs() []ModuleConfig
 	// DefaultLintConfig returns the DefaultLintConfig for the File.
 	//
-	// For v1 buf.yaml files, this will be the default v1 lint config.
+	// For v1 buf.yaml files, there is only ever a single ModuleConfig, so that is returned.
```
pkwarren, line 79:
```suggestion
	// For v1 buf.yaml files, there is only ever a single ModuleConfig, so its LintConfig is returned.
```

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -76,13 +76,13 @@ type BufYAMLFile interface {
...
 	// For v2 buf.yaml files, if a top-level lint config exists, then it will be the top-level
 	// lint config. Otherwise it is the default v2 lint config.
 	DefaultLintConfig() LintConfig
 	// DefaultBreakingConfig returns the DefaultBreakingConfig for the File.
 	//
-	// For v1 buf.yaml, this will be the default v1 breaking config.
+	// For v1 buf.yaml files, there is only ever a single ModuleConfig, so that is returned.
```
pkwarren, line 85:
```suggestion
	// For v1 buf.yaml files, there is only ever a single ModuleConfig, so its BreakingConfig is returned.
```

pkwarren (approved):

oliversun9 (approved):

pkwarren (approved):

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -275,6 +300,20 @@ func (c *bufYAMLFile) ModuleConfigs() []ModuleConfig {
...
+func (c *bufYAMLFile) DefaultLintConfig() LintConfig {
+	if c.defaultLintConfig == nil {
+		return c.moduleConfigs[0].LintConfig()
+	}
+	return c.defaultLintConfig
+}
+
+func (c *bufYAMLFile) DefaultBreakingConfig() BreakingConfig {
+	if c.defaultBreakingConfig == nil {
+		return c.moduleConfigs[0].BreakingConfig()
```
bufdev, line 312: This isn't actually correct - this is just choosing a single configuration among potentially many and calling it the default, which is what we've tried to avoid pretty consistently with v2. I think what you're actually looking for in this PR is some concept of "if there is a single configuration, give it to me, otherwise tell me it is not present". What this logic looks like:

```func
SingleLintConfig() (LintConfig, bool)
```

- If v1, return the v1 configuration.
- If v2:
  - If there is one module with attached configuration, return the attached configuration. Else, return the top-level configuation.
  - If there is more than one module, return false

If you get false, do nothing, as we do right now.

doriable: So if a user passes the following file to `--config`:

```
version: v2
lint:
  use:
    - WIRE
breaking:
  use:
    - FILE
modules:
  - path: foo
  - path: bar
    lint:
      use:
        - DEFAULT
      except:
        - IMPORT_USED
  - path: vendor
```

Wouldn't we want them to pick up the top-level configs for the default?

bufdev: We'd probably want to error, or at least print a warning along the lines of:

```
WARN Configuration {from flag,at $PATH was used} but this configuration had multiple modules. Buf cannot deduce which {lint,breaking} configuration to use with your image. Ignoring your configuration and using the default configuration instead.
```

I could see an argument to also modify the logic to "if >1 modules, return the top-level module", but that may just leave us open to even more support stuff in the long-term.

Going a different direction, we could take the position that for image inputs only, we always take the top-level configuration, and never look into specific module configs. This would mean that if you had the following:

```
version: v2
modules:
  - path: .
     lint: ...
```

The `modules[0].LintConfig` would not be picked up and propagated. The position here would be that module-specific configuration never applies to an image, and we treat configuration for an image as if it were a "new" module in the `modules` list.

This may be what `moduleConfigs[0]` does in practice for v1, but not for v2, but I'd also be comfortable with adopting this position, it seems intellectually consistent.

doriable: I am a fan of the latter approach for images. For `v1`, the behaviour will stay the same since there should only ever be a single lint/breaking config. For `v2`,  we check for a top-level config, and if nothing is there, then we would use whatever the version defaults are. And yes, in the case you provided, where there is a single module, `path: .`, that would not be considered the default configuration. Changes incoming.

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -74,6 +74,13 @@ type BufYAMLFile interface {
 	// All ModuleConfigs will have unique ModuleFullNames.
 	// Sorted by DirPath.
 	ModuleConfigs() []ModuleConfig
+	// TopLevelModuleConfig returns the top-level ModuleConfig for the File.
+	//
+	// For v1 buf.yaml files, there is only ever a single ModuleConfig, so this is returned.
+	// For v2 buf.yaml files, if a top-level module config exists, then it will be the top-level
+	// module config. Otherwise, this will return nil, so callers should be aware this may be
+	// empty.
+	TopLevelModuleConfig() ModuleConfig
```
bufdev, line 83: Why is this a `ModuleConfig`? We only want `LintConfig` and `BreakingConfig`, I don't know what a top-level `ModuleConfig` is in terms of v2

private/buf/bufctl/controller.go
```diff
@@ -393,20 +394,49 @@ func (c *controller) GetTargetImageWithConfigs(
...
-				// breaking config to apply.
-			default:
-				return nil, syserror.Newf("unknown FileVersion: %v", fileVersion)
+			} else {
+				// Otherwise we use the file path
+				configInfo = bufYAMLFile.ObjectData().Name()
+			}
+			if topLevelModuleConfig := bufYAMLFile.TopLevelModuleConfig(); topLevelModuleConfig != nil {
+				if bufYAMLFile.FileVersion() == bufconfig.FileVersionV2 && len(bufYAMLFile.ModuleConfigs()) > 1 {
+					c.logger.Sugar().Warnf(
```
bufdev, line 422: I don't think we need to print either of these warnings, I just think we take the position that doing lint or breaking change detection for an image means we treat it like any other module - if it doesn't have a configuration, we use the top-level configuration, which may be empty and then we just go to default.

private/bufpkg/bufconfig/buf_yaml_file.go
doriable: I used `ModuleConfig` because it worked better for the warnings and also to account for situations like:

```
version: v2
name: <remote>/<owner>/<module>
```

But in that case, it would be the default anyway, so it doesn't super matter, and also since we're not printing out the warnings, I can swap it back to `LintConfig` and `BreakingConfig`.

doriable: Updated back to `LintConfig` and `BreakingConfig`, and updated the PR title + description.

bufdev (approved):

PR #3113, opened 2024-06-26
Fix studio agent flaky test

oliversun9:
Tests `TestPlainPostHandlerTLS/invalid_upstream` and `TestPlainPostHandlerH2C/invalid_upstream` are [flaky](https://github.com/bufbuild/buf/actions/runs/9599996622/job/26475130640?pr=3102), because the listener only calls `Accept` once, and when it accepts a connection from something other than the agent, the listener calls it a day, leaving the agent hanging.

This PR makes the listener repeatedly accept new connections until it has accepted a connection from the agent.

Running `go test -count=1 -timeout=20s -race github.com/bufbuild/buf/private/buf/bufstudioagent` on main on my laptop would fail (with timeout) this test within the first 5 times, and trying it on this branch, it doesn't fail (tried 100 times and it hasn't failed).

TLDR: random ports can make connection with the listener and we need to account for that.

### DIFF
--- a/private/buf/bufstudioagent/bufstudioagent_test.go
+++ b/private/buf/bufstudioagent/bufstudioagent_test.go
@@ -20,6 +20,7 @@ import (
 	"crypto/x509"
 	"encoding/base64"
 	"errors"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
@@ -196,9 +197,30 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
 		listener, err := net.Listen("tcp", "127.0.0.1:")
 		require.NoError(t, err)
 		go func() {
-			conn, err := listener.Accept()
-			require.NoError(t, err)
-			require.NoError(t, conn.Close())
+			for i := 0; i < 10; i++ {
+				fmt.Println("before accept")
+				conn, err := listener.Accept()
+				if err != nil {
+					if opErr, ok := err.(*net.OpError); ok {
+						fmt.Println("op---:", opErr)
+						continue
+					}
+					fmt.Println("err---:", err)
+					continue
+				}
+				fmt.Println("local", conn.LocalAddr())
+				fmt.Println("remote", conn.RemoteAddr())
+				b := make([]byte, 0, 1000)
+				x, err := conn.Read(b)
+				if err != nil {
+					fmt.Println("fail to read", err)
+					continue
+				}
+				fmt.Printf("read %d bytes\n", x)
+				require.NoError(t, conn.Close())
+				fmt.Println("after accept")
+			}
+			t.Fail()
 		}()
 		defer listener.Close()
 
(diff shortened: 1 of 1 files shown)


private/buf/bufstudioagent/bufstudioagent_test.go
```diff
@@ -193,14 +194,31 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
...
-			conn, err := listener.Accept()
-			require.NoError(t, err)
-			require.NoError(t, conn.Close())
+			for {
+				select {
+				case <-done:
+					listener.Close()
+					return
+				default:
+					err := listener.SetDeadline(time.Now().Add(time.Second))
```
oliversun9, line 212: `time.Second` is an arbitrary number, but the nice thing about it is that when you run `go test -count=1 -timeout=20s -race github.com/bufbuild/buf/private/buf/bufstudioagent` you can estimate how many connections did it accept before accepting the one coming from the studio agent.

doriable: I'm okay with leaving this deadline for now, but for testing, I think it's fine to retry pretty aggressively (e.g. `time.Milisecond`), since we want the test to run as quickly as possible.

doriable (approved):
Looks good to me, thanks for fixing this up!

private/buf/bufstudioagent/bufstudioagent_test.go
```diff
@@ -193,14 +194,31 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
...
+			done <- struct{}{}
+		})
 		go func() {
-			conn, err := listener.Accept()
-			require.NoError(t, err)
-			require.NoError(t, conn.Close())
+			for {
+				select {
+				case <-done:
+					listener.Close()
```
emcfarlane, line 209: Why the channel? It's equivalent to call `listener.Close()` in the Cleanup and you already must handle the error on accept.

private/buf/bufstudioagent/bufstudioagent_test.go
```diff
@@ -193,14 +194,31 @@ func testPlainPostHandlerErrors(t *testing.T, upstreamServer *httptest.Server) {
...
+					return
+				default:
+					err := listener.SetDeadline(time.Now().Add(time.Second))
+					require.NoError(t, err)
+					conn, err := listener.Accept()
+					if err != nil {
+						continue
+					}
+					require.NoError(t, conn.Close())
```
emcfarlane, line 218: `require` is not safe from go routines: https://pkg.go.dev/testing#T.FailNow

oliversun9: Good catch!

private/buf/bufstudioagent/bufstudioagent_test.go
oliversun9: If I call `listener.Close()` from the parent goroutine, the listener wouldn't know it's closed (plz correct me if I'm wrong) and then would just keep calling `Accept` even though it's already closed.

emcfarlane: Ah the continue would need to do error handling. Maybe this is fine.

private/buf/bufstudioagent/bufstudioagent_test.go
oliversun9: Updated them to `continue`

private/buf/bufstudioagent/bufstudioagent_test.go
oliversun9: Updated to 10ms: [dde47d6](https://github.com/bufbuild/buf/pull/3113/commits/dde47d68dced33e354b66dcbb698f4cf2f0f22d0), I'm not sure how what would happen with 1ms when GitHub action runners' performance is degraded.

private/buf/bufstudioagent/bufstudioagent_test.go
emcfarlane: You can use `t.Error` safely as that will call fail not failnow.

oliversun9: Fixed in [4852e27](https://github.com/bufbuild/buf/pull/3113/commits/4852e27e51e59999c54730a4eb23f841423474c4)

bufdev (approved):

Issue #2714, opened 2024-01-18
Feature request: Option on buf convert to set `DiscardUnknown` to false when unmarshalling input


lewchuk-span:
I'd like to put a feature request for [this todo](https://github.com/bufbuild/buf/blob/main/private/pkg/protoencoding/txtpb_unmarshaler.go#L35) to be implemented and exposed as a command line parameter. Specifically, as a parameter on `buf convert`.

My use case is that I am using `buf convert` to validate textpb files against my protobuf definitions. We have found that having samples of valid instances of protos along with the schema definitions helps our team understand the semantic intention of the schemas. To support this process we have implemented validation of those samples using buf convert which helps capture:
Basic invalid formats
Type errors on fields
In addition we want to reject samples which have extra fields on them. In our use case those extra fields are almost always caused by spelling mistakes on a field that does exist in the schema. In addition if the field with the spelling error is a message type all child fields get no validation at all potentially allowing type errors to slip through.

We have implemented a hacky solution using convert to output `textpb` and then using `diff` but this is a brittle solution (subject to the detrand randomness and requiring our samples to match the formatting of the proto textpb marshaller which might change with future versions). If we had a --error-on-unknown flag we could simplify our system making it more reliable and flexible.

PR #3115, opened 2024-06-28
Add buf convert --disallow-unknown

bufdev:
Begins https://github.com/bufbuild/buf/issues/2714. See https://github.com/bufbuild/protoyaml-go/issues/37 for the remaining work for YAML that once fixed, can resolve the aforementioned issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,13 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf convert --disallow-unknown` flag to error if unknown fields or enum name values
+  are contained within the input to `buf convert`.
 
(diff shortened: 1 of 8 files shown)


bufdev:
This was just a drive-by PR and it's actually wrong. This adjusts the default behavior for binary and text formats to discard unknown values, which breaks a lot of things in unexpected ways. For some reason, we had discard unknown default to true for JSON, false for wire and txtpb, and implicitly false for yaml. This was likely a decision I made a long time ago.

This PR will need to be significantly adjusted to not introduce breaking changes.

bufdev:
Closing this for now - this PR is seriously off.

PR #3128, opened 2024-07-04
Remove username prompt on buf registry login

emcfarlane:
This removes the requirement to provide your username when calling `buf registry login`. On login the token is validated by fetching the current user. The username is used only to validate that the current token is associated with the user but is otherwise redundant. The flag has now been marked deprecated and is hidden.

The output of the login command is also updated to be more user friendly prompting the user to visit the web UI to create a token if they do not have one. On successful login the user is informed that they are logged in as the current user.

Token via prompt:
```
$ buf registry login
Enter the BSR token created at https://buf.build/settings/user.

Token:
Logged in as emcfarlane. Credentials saved to /Users/edward/.netrc.
```

Token via stdin:
```
$ echo ${BUF_TOKEN} | buf registry login buf.build --token-stdin
Logged in as emcfarlane. Credentials saved to /Users/edward/.netrc.
```

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -75,7 +75,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		"The username to use.",
 	)
-	_ = flagSet.MarkDeprecated(usernameFlagName, "This flag no longer has any effect")
+	_ = flagSet.MarkDeprecated(usernameFlagName, "This flag is no longer needed as the username is automatically derived from the token")
 	_ = flagSet.MarkHidden(usernameFlagName)
 	flagSet.BoolVar(
 		&f.TokenStdin,
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -47,7 +47,7 @@ func NewCommand(
 	return &appcmd.Command{
 		Use:   name + " <domain>",
 		Short: `Log in to the Buf Schema Registry`,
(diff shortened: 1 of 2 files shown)


doriable (approved):
This makes a lot of sense to me. I like that `--username` continues to be available -- I mulled over whether we should consider marking the flag as deprecated, but I think the fact that it continues to provide validation makes a lot of sense.

doriable (commented):
We should probably add a changelog entry to this.

doriable (approved):

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -73,7 +73,7 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		&f.Username,
 		usernameFlagName,
 		"",
-		"The username to use. This command prompts for a username by default",
+		"The username to use. If provided, the command will validate that the token is associated with this username",
```
bufdev, line 76: I would hide and deprecate this flag.

emcfarlane: Hidden and reverted text to just "The user to use".

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -73,8 +73,10 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		&f.Username,
 		usernameFlagName,
 		"",
-		"The username to use. This command prompts for a username by default",
+		"The username to use.",
 	)
+	_ = flagSet.MarkDeprecated(usernameFlagName, "This flag no longer has any effect")
```
bufdev, line 78: Change to "This flag is no longer needed as the username is automatically derived from the token"

bufdev (approved):

PR #3130, opened 2024-07-04
Add `clean` to `buf.gen.yaml`

doriable:
As a quick follow-up to #3124 and in response to https://github.com/bufbuild/buf/pull/3124#issuecomment-2208430946,
this adds a `clean` config key to `v2` `buf.gen.yaml`, for example:

```yaml
version: v2
clean: true
plugins:
  - local: custom-gen-go
    out: gen/go
    opt: paths=source_relative
    strategy: directory
  - protoc_builtin: java
    out: gen/java
```

When running `buf generate` with the above configs, the outs set to each plugin
(e.g. `gen/go` and `gen/java`) will be removed before code generation is run.

If `buf generate --clean` flag is set, then that will always take precedence, even if
`clean: false` in the configuration. And likewise, if `buf generate --clean=false`,
and `clean: true` in the configuration, then we would not delete the out directories.


doriable:
> I'm not sure why I'd want this - why would I want per-plugin config of cleanup?

This is based on @timostamm's comment, which I agree with

> I think it would also be nice to have the flag available in buf.gen.yaml next to [out](https://buf.build/docs/configuration/v2/buf-gen-yaml#out). Then users don't have to specify the flag all the time. And specifying the flag next to the out directory path makes it more obvious what is going to be deleted.

It's convenient not having to specify the flag all the time/have it explicitly set in configurations so users can easily reference it. And it does make the association with `out` very clear.

bufdev:
Why would I want to set it on a per-plugin basis? What's an example of where I'd want to clean for one plugin, but not for another?

Perhaps I could want to do this globally, would be interested what that'd look like from a config perspective.

bufdev:
Updated to `clean` instead of `clean_plugin_outs` - I think we should just match the flag.

bufdev (approved):

bufdev:
buf.build/docs needs to be updated

PR #3132, opened 2024-07-05
Add --list-services and --list-methods flags to buf curl

jhump:
This adds `--list-services` and `--list-methods` flags which change the behavior of `buf curl`, causing it to list known services or methods, instead of invoking an RPC.

### DIFF
--- a/private/buf/cmd/buf/command/curl/curl.go
+++ b/private/buf/cmd/buf/command/curl/curl.go
@@ -27,6 +27,7 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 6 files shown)


private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -999,3 +1113,16 @@ func makeHTTPClient(f *flags, isSecure bool, authority string, printer verbose.P
 func secondsToDuration(secs float64) time.Duration {
 	return time.Duration(float64(time.Second) * secs)
 }
+
+func memoizeFunc(action func() (connect.HTTPClient, error)) func() (connect.HTTPClient, error) {
```
bufdev, line 1117: This is just syncext.OnceValues

bufdev (approved):

PR #3133, opened 2024-07-05
Add limited_write to DisplayRepositoryElementsResponse

gilwong00:
We want to only allow and display certain actions and elements on the BSR if the base repository role is `limited_write`.


proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -92,6 +92,8 @@ message DisplayRepositoryElementsResponse {
   bool delete = 2;
   // Display write repository element if true.
   bool write = 3;
+  // Similar to write but displays a more restrictive set of repository elements if true.
+  bool limited_write = 4;
```
unmultimedio, line 96:
```suggestion
  // Display write repository elements if true. It includes all elements from limited_write.
  bool write = 3;
  // Display limited_write repository elements if true. This is a subset of write elements.
  bool limited_write = 4;
```

Although that's a UI concern, may be it's not even necessary to be specific in these docs.

unmultimedio (approved):

unmultimedio (approved):

bufdev (approved):

PR #3139, opened 2024-07-08
File locking module data store

doriable:
This adds file locking to the cache implementation of `ModuleDataStore`, using a
`module.lock` file. This fixes a race condition we had with concurrent
reads/writes for module data by concurrent invocations of the `buf` CLI.

In this implementation, we are taking a shared lock on `<module_key>/module.lock`
when reading from the cache, and taking an exclusive lock for writes.

When clearing out the module directory, the `module.lock` file is deleted last
to ensure that we hold the lock while everything else is being cleared out.

When writing data to the cache, we first take a shared lock and check `module.yaml`. If
`module.yaml` contains valid data, then we do not need to do an extraneous write. If not,
then we can upgrade to an exclusive lock, check `module.yaml` again, and if it is valid, then
we do not write, otherwise we proceed to writing the data.

A separate `module.lock` file is used instead of locking on an existing file, e.g. `module.yaml`
since that is the intended use of the `flock` package, since the file handler is not exposed
from the lock and Windows prevents the same process from accessing a region on a file it
has taken an exclusive lock on: (https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-lockfileex)

> If the locking process opens the file a second time, it cannot access the specified region through this second handle until it unlocks the region.

This PR also adds support for setting a default lock timeout and retry delay on `filelock.Locker` using `LockerOption`s when constructing the `Locker`. If any timeouts and/or retry delays are passed to `Lock` through `LockOption`s, then those will take precedence.


.golangci.yml
```diff
@@ -281,4 +281,7 @@ issues:
       path: private/buf/cmd/buf/buf_test.go
       # The LsModules tests call chdir and cannot be parallelized.
       text: "LsModules"
-      
+    - linters:
+        - forbidigo
+      # We are using errgroup in a test for concurrent reads/writes to the cache.
+      path: private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```
pkwarren, line 287: We might be able to refactor the test to use thread.Parallelize - I don't think we're relying on any specific behavior from errgroup there?

private/buf/bufcli/cache.go
```diff
@@ -166,12 +167,17 @@ func newModuleDataProvider(
...
+	if err != nil {
+		return nil, err
+	}
 	return bufmodulecache.NewModuleDataProvider(
 		container.Logger(),
 		delegateModuleDataProvider,
 		bufmodulestore.NewModuleDataStore(
 			container.Logger(),
 			cacheBucket,
+			bufmodulestore.ModuleDataStoreWithFileLocker(filelocker),
```
pkwarren, line 180: Since this is required for correctness, do we want to make it a required argument instead of an option?

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -214,6 +220,57 @@ func TestModuleDataProviderBasic(t *testing.T) {
...
+	cacheDir := filepath.Join(tempDir, "cache")
+
+	for i := 0; i < 20; i++ {
+		require.NoError(t, os.MkdirAll(cacheDir, 0755))
+		errs, ctx := errgroup.WithContext(context.Background())
+
+		for j := 0; j < 5; j++ {
+			bucket, err := storageos.NewProvider().NewReadWriteBucket(cacheDir)
+			require.NoError(t, err)
+			filelocker, err := filelock.NewLocker(cacheDir)
```
pkwarren, line 238: Can we lift this out of the inner loop?

private/bufpkg/bufmodule/bufmodulecache/module_data_provider.go
```diff
@@ -62,6 +62,6 @@ func newModuleDataProvider(
 func (p *moduleDataProvider) GetModuleDatasForModuleKeys(
 	ctx context.Context,
 	moduleKeys []bufmodule.ModuleKey,
-) ([]bufmodule.ModuleData, error) {
+) (_ []bufmodule.ModuleData, retErr error) {
```
pkwarren, line 65: We can revert this.

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -214,6 +220,57 @@ func TestModuleDataProviderBasic(t *testing.T) {
...
+			require.NoError(t, err)
+
+			cacheProvider := newModuleDataProvider(
+				zap.NewNop(),
```
pkwarren, line 242: We should configure a zaptest logger here - I'm seeing errors when running it with logging enabled.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -297,6 +327,15 @@ func (p *moduleDataStore) deleteInvalidModuleData(
 	if err != nil {
 		return err
 	}
+	if p.filelocker != nil {
+		unlocker, err := p.filelocker.Lock(ctx, dirPath+"/"+externalModuleDataFileName)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			retErr = multierr.Append(retErr, unlocker.Unlock())
```
pkwarren, line 336: There may still be a race here as we're calling `DeleteAll` below which also deletes the module.yaml file (which we're locking to prevent concurrent access). We need to rethink this - perhaps by deleting all files other than the module.yaml and then deleting it at the end.

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -214,6 +220,57 @@ func TestModuleDataProviderBasic(t *testing.T) {
 	)
 }
 
+func TestConcurrentCacheReadWrite(t *testing.T) {
+	t.Skip("skipping expensive test for concurrent cache reads/writes")
```
pkwarren, line 224: I would expect it to finish a lot faster than this - the first one to acquire should get the write lock and complete writing everything and then the other read locks should read cached data. Are we acquiring the locks in order here?

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Yeah, I think this makes sense -- deleting all contents of the module dir first before deleting `module.yaml`.

private/buf/bufcli/cache.go
doriable: I waffled a little bit on this, since the cache could technically be a storage bucket and also the file locker is ignored when `tar` is set. So it's not exactly mandatory for the cache, so I went with an option here, I'm open to making this permanent though (and still ignoring for `tar`).

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
doriable: Seems like the major gain/improvement on the test was not calling the delete all on non-existent errors, which makes sense since that is an expensive operation.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -167,14 +183,33 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
 		}
+		p.logDebugModuleKey(
+			moduleKey,
+			"module data store dir read write bucket",
+			zap.String("dirPath", dirPath),
+		)
+		moduleCacheBucket = storage.MapReadWriteBucket(p.bucket, storage.MapOnPrefix(dirPath))
+		// Only attempt to get a file lock when storing individual files
+		if p.filelocker != nil {
+			unlocker, err := p.filelocker.RLock(ctx, dirPath+"/"+externalModuleDataLockFileName)
```
pkwarren, line 198: Use `normalpath.Join` here?

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -167,14 +183,33 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+				if err := unlocker.Unlock(); err != nil {
+					p.logger.Debug("failed_get_runlock", zap.Error(err))
+					retErr = multierr.Append(retErr, err)
+				}
+			}()
+		}
 	}
 	defer func() {
-		if retErr != nil {
+		if retErr != nil && !errors.Is(err, fs.ErrNotExist) {
```
pkwarren, line 212: I don't know if `fs.ErrNotExist` can be returned by any other calls other than the `storage.ReadPath` below - if so we might want to tweak this.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -297,15 +332,56 @@ func (p *moduleDataStore) deleteInvalidModuleData(
...
+			return err
+		}
+		defer func() {
+			if err := unlocker.Unlock(); err != nil {
+				p.logger.Debug("failed_delete_unlock", zap.Error(err))
+				retErr = multierr.Append(retErr, err)
+			}
+		}()
+	}
+	moduleDir, err := os.Open(dirPath)
```
pkwarren, line 348: As discussed this should be updated to use storage APIs.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -317,11 +393,101 @@ func (p *moduleDataStore) putModuleData(
...
+		var readLocked bool
+		if p.filelocker != nil {
+			readUnlocker, err = p.filelocker.RLock(ctx, dirPath+"/"+externalModuleDataLockFileName)
+			if err != nil {
+				p.logger.Debug("failed_put_rlock", zap.Error(err))
+				return err
+			}
+			readLocked = true
+			defer func() {
+				if readLocked && readUnlocker != nil {
```
pkwarren, line 419: I think we can remove `readLocked` and just use `readLocker != nil` here (and assign it to nil below after unlocking it).

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -232,12 +234,16 @@ func TestConcurrentCacheReadWrite(t *testing.T) {
...
+			func(i int) (func(ctx context.Context) error, error) {
+				logger := logger.Named(fmt.Sprintf("job-%d", i))
 				bucket, err := storageos.NewProvider().NewReadWriteBucket(cacheDir)
 				if err != nil {
 					return nil, err
 				}
-				filelocker, err := filelock.NewLocker(cacheDir)
+				filelocker, err := filelock.NewLocker(
+					cacheDir,
+					filelock.LockerWithLockRetryDelay(10*time.Millisecond),
```
pkwarren, line 245: It would be good to add a comment why we're doing this here (drops test runtime from ~16s down to ~1s).

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -345,35 +344,29 @@ func (p *moduleDataStore) deleteInvalidModuleData(
...
-	moduleDir, err := os.Open(dirPath)
-	if err != nil {
-		return err
-	}
-	defer func() {
-		// If the moduleDir was already closed, that is fine
-		if err := moduleDir.Close(); err != nil && err != fs.ErrClosed {
-			p.logger.Debug("failed_delete_close_module_dir", zap.Error(err))
-			retErr = multierr.Append(retErr, err)
+	var deleted map[string]struct{}
```
pkwarren, line 347: It would be good to add a comment here that we're going to delete everything but the lock file (while the lock is held).

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -345,35 +344,29 @@ func (p *moduleDataStore) deleteInvalidModuleData(
...
+		}
+		deleted[topLevelPath] = struct{}{}
+		return nil
+	}); err != nil {
+		return err
 	}
-	// Delete the lock file
-	return p.bucket.Delete(ctx, dirPath+"/"+externalModuleDataLockFileName)
+	// Delete the entire dirPath
+	return p.bucket.DeleteAll(ctx, dirPath)
```
pkwarren, line 369: Does the deferred Unlock fail here if the underlying file has been deleted?

private/pkg/filelock/locker.go
```diff
@@ -56,10 +72,17 @@ func (l *locker) RLock(ctx context.Context, path string, options ...LockOption)
...
+			LockWithTimeout(l.lockTimeout),
+			LockWithRetryDelay(l.lockRetryDelay),
+		},
+		options..., // Any additional options set will be applied last
+	)
 	return rlock(
 		ctx,
 		normalpath.Unnormalize(normalpath.Join(l.rootDirPath, path)),
-		options...,
+		options..., // Any additional options set will be applied last
```
pkwarren, line 85:
```suggestion
		options...,
```

nit: Remove duplicate comment - makes more sense above on the append call.

private/pkg/filelock/locker.go
```diff
@@ -35,16 +38,29 @@ func newLocker(rootDirPath string) (*locker, error) {
...
+		lockTimeout:    lockerOptions.lockTimeout,
+		lockRetryDelay: lockerOptions.lockRetryDelay,
 	}, nil
 }
 
 func (l *locker) Lock(ctx context.Context, path string, options ...LockOption) (Unlocker, error) {
 	if err := validatePath(path); err != nil {
 		return nil, err
 	}
+	options = append(
```
pkwarren, line 57: nit: Can use `slicesext.Concat` here (avoiding resizing slice).

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Based on my understanding, Unix file locks are advisory and the `Unlock` call is made to the file descriptor of the handler from the original `Lock` call, and so if this operation was invalid, I would expect to see a `EINVAL`. To validate this, I wrote a small test locally where I deleted a file I held a lock on and called unlock (no error) and I ran the test ~2K+ iterations.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -440,13 +430,13 @@ func (p *moduleDataStore) putModuleData(
 		// Otherwise, release shared lock and upgrade to an exclusive lock for writes.
 		if readUnlocker != nil {
 			err := readUnlocker.Unlock()
-			readLocked = false
 			if err != nil {
 				return err
 			}
+			readUnlocker = nil // unset the readUnlocker since we are upgrading the lock
```
pkwarren, line 436: I think we want to move this after line 432 so we don't try to unlock two times.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Yes, those calls can return `fs.ErrNotExist`, but we only attempt to read those paths if they are parsed from a valid `module.yaml`. Might still make sense to call `deleteInvalidModuleData` in those cases, so let me rework this.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -167,14 +182,37 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+			}()
+		}
 	}
+	// We want to avoid calling deleteInvalidModuleData if no module.yaml file was found,
+	// since it is an expensive operation and there is no valid module data, so the cache
+	// will be overwritten.
+	var externalModuleDataFileNotFound bool
 	defer func() {
-		if retErr != nil {
+		if retErr != nil && !externalModuleDataFileNotFound {
```
pkwarren, line 215: I think we should rethink deleting module data at all for directory (not tar) module stores in the Get path. We are deciding to delete data while a read lock is held - by the time a write lock is acquired it is possible that the underlying data is now valid.

We probably want to have a test that covers this too - so far I think we only test writing invalid data with a storagemem backed bucket - we should update `module_data_store_test.go` to also exercise those tests against a storageos bucket w/ t.Tempdir() too.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -391,22 +546,6 @@ func (p *moduleDataStore) putModuleData(
 	return storage.PutPath(ctx, moduleCacheBucket, externalModuleDataFileName, data)
```
pkwarren, line 546: We should write this file atomically - that way if we're ever killed (i.e. kill -9) we never have the chance to read an invalid module.yaml file.

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -214,6 +222,64 @@ func TestModuleDataProviderBasic(t *testing.T) {
 	)
 }
 
+func TestConcurrentCacheReadWrite(t *testing.T) {
+	t.Parallel()
+
+	bsrProvider, moduleKeys := testGetBSRProviderAndModuleKeys(t, context.Background())
+	tempDir := t.TempDir()
+	cacheDir := filepath.Join(tempDir, "cache")
+	logger := zaptest.NewLogger(t)
```
pkwarren, line 231: We probably want to set the log level to info here before merging as this test is really chatty in debug mode.

pkwarren:
We might consider landing #3123 before this so we can update to the latest filelock version. It requires Go 1.21+ in all of the more recent releases.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: This makes sense to me -- we also cannot upgrade the lock, so grabbing the exclusive lock would require some messing around. Also, `GetModuleDatasForModuleKeys` is basically relying on `deleteInvalidModuleData` to return the `fs.NotExist` error, when in effect, we should be treating all errors in the "get" as a cache miss.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -129,9 +138,7 @@ func (p *moduleDataStore) GetModuleDatasForModuleKeys(
 	for _, moduleKey := range moduleKeys {
 		moduleData, err := p.getModuleDataForModuleKey(ctx, moduleKey)
 		if err != nil {
-			if !errors.Is(err, fs.ErrNotExist) {
-				return nil, nil, err
-			}
+			// Treat any error returned as a cache miss
```
bufdev, line 141: This seems wrong - I'd want to understand why we are changing this

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -162,22 +169,46 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+				// If there is an error fetching the tar bucket that is not because the path does
+				// not exist, we assume this is corrupted and delete the tar.
+				tarPath, err := getModuleDataStoreTarPath(moduleKey)
+				if err != nil {
+					return nil, err
+				}
+				if err := p.bucket.Delete(ctx, tarPath); err != nil {
+					return nil, err
+				}
+				// Return a path error indicating the module data was not found
```
bufdev, line 181: This needs more explanation in comments as to why we're doing this - I was confused on the first pass, and would be in a year+. I am a little confused overall though - we're blanket saying that *any* error means that we assume it's corrupted, so deep system errors will be swallowed here

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -162,22 +169,46 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+		// Only attempt to get a file lock when storing individual files
+		if p.filelocker != nil {
+			unlocker, err := p.filelocker.RLock(ctx, normalpath.Join(dirPath, externalModuleDataLockFileName))
+			if err != nil {
+				p.logger.Debug("failed to get rlock for getModuleDataForModuleKey", zap.Error(err))
+				return nil, err
+			}
+			defer func() {
+				if err := unlocker.Unlock(); err != nil {
+					p.logger.Debug("failed to runlock for getModuleDataForModuleKey", zap.Error(err))
```
bufdev, line 206: You're not RUnlocking though, although I think you should be

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -317,11 +311,99 @@ func (p *moduleDataStore) putModuleData(
...
+		var readUnlocker filelock.Unlocker
+		if p.filelocker != nil {
+			readUnlocker, err = p.filelocker.RLock(ctx, normalpath.Join(dirPath, externalModuleDataLockFileName))
+			if err != nil {
+				p.logger.Debug("failed to get rlock for putModuleData", zap.Error(err))
+				return err
+			}
+			defer func() {
+				if readUnlocker != nil {
+					if err := readUnlocker.Unlock(); err != nil {
```
bufdev, line 336: Same question about Unlock vs RUnlock

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -317,11 +311,99 @@ func (p *moduleDataStore) putModuleData(
...
+		p.logDebugModuleKey(
+			moduleKey,
+			"module data store dir read write bucket",
+			zap.String("dirPath", dirPath),
+		)
+		moduleCacheBucket = storage.MapReadWriteBucket(p.bucket, storage.MapOnPrefix(dirPath))
+
+		// Only attempt to get a file locks when storing individual files.
+		// Before writing to the module directory, first get a shared lock and check module.yaml
+		var readUnlocker filelock.Unlocker
```
bufdev, line 327: This isn't a readUnlocker, its an unlocker

private/buf/bufcli/cache.go
```diff
@@ -166,12 +167,17 @@ func newModuleDataProvider(
 	if err != nil {
 		return nil, err
 	}
+	filelocker, err := filelock.NewLocker(fullCacheDirPath)
```
bufdev, line 170: The locker should not be on the module cache path, you should have a separate lock cache path that doesn't mix lock files into where we store our module data.

bufdev (commented):
I'm going to have to take another pass, I'm having a bit of trouble following the logic - the logic is spelled out more in the PR comment than in the code, so might be worth explaining more what is going on in the code as well, but I'll work to take another look later

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: The `filelock.Locker` API provides an `Unlocker` interface, which only has `Unlock`, regardless if you're taking a shared lock or an exclusive lock. This maps to the `flock` package, and makes sense, since on most systems you call a single Unlock (e.g. `LOCK_UN`) regardless of the type of lock you are holding (and although we don't expose it through our `filelock` wrapper layer, you may replace shared locks with exclusive locks , and calling `Unlock` would on the shared lock would also release the exclusive lock).

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Addressed in a previous comment.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Similarly to the comment above, the interface returned from `RLock` is a `filelock.Unlocker`, but this is the unlocker for this specific shared lock. Our `filelock` API does not expose the underlying `*flock.Flock`, so we cannot replace the shared lock with an exclusive lock, so we must handle the shared lock with its own unlocker here.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: In the previous implementation, we are burying the errors even deeper:
- We coerce a call on `deleteInvalidModuleData` on every error https://github.com/bufbuild/buf/blob/8996858e1193b1918fdc328d73805eb35a6acd8b/private/bufpkg/bufmodule/bufmodulestore/module_data_store.go#L176-L180
- And we only surface an error from the `deleteInvalidModuleData` process, otherwise we then coerce all of those errors to be a `PathError` https://github.com/bufbuild/buf/blob/8996858e1193b1918fdc328d73805eb35a6acd8b/private/bufpkg/bufmodule/bufmodulestore/module_data_store.go#L277-L286
- And then all of those path errors are then considered a cache miss

So there are a couple of things about this change:
- We are simply returning the error here and transparently treating it as a cache miss -- if we want, we can log these errors
- We removed `deleteInvalidModuleData`, which causes cache corruption because it was creating a race

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Related to the comment above -- we were already swallowing the system errors. This is transparently pulling that logic to this layer, since in the case for `tar`, we would still need to delete the invalid module data tar in this case and return a relevant error.

private/buf/bufcli/cache.go
doriable: That is fair -- added a new directory for this.

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -235,7 +302,10 @@ func testGetBSRProviderAndModuleKeys(t *testing.T, ctx context.Context) (bufmodu
 		bufmoduletesting.ModuleData{
 			Name: "buf.build/foo/mod3",
 			PathToData: map[string][]byte{
-				"mod3.proto": []byte(
+				"mod3a.proto": []byte(
```
pkwarren, line 305: Did we end up needing to make this change for any reason for the tests? If not we could revert.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -99,20 +105,23 @@ func ModuleDataStoreWithTar() ModuleDataStoreOption {
...
+	bucket     storage.ReadWriteBucket
+	filelocker filelock.Locker
 
 	tar bool
 }
 
 func newModuleDataStore(
 	logger *zap.Logger,
 	bucket storage.ReadWriteBucket,
+	filelocker filelock.Locker,
```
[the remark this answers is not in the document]
doriable: I opted to go with setting a `NopLocker` when passed `nil`, since i want to keep the constructor as simple as possible and preserve the current signature. Even though there are not many callsites in the codebase, I think it's good practice and this would be representative of the behaviour we want anyway.

pkwarren (approved):

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -162,22 +173,44 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
-		if retErr != nil {
-			retErr = p.deleteInvalidModuleData(ctx, moduleKey, retErr)
+		p.logDebugModuleKey(
+			moduleKey,
+			"module data store dir read write bucket",
+			zap.String("dirPath", dirPath),
+		)
+		moduleCacheBucket = storage.MapReadWriteBucket(p.bucket, storage.MapOnPrefix(dirPath))
+		// Only attempt to get a file lock when storing individual files
+		unlocker, err := p.filelocker.RLock(ctx, dirPath+externalModuleDataLockFileExt)
```
bufdev, line 202: I would drastically prefer if we added a `getModuleDataStoreDirLockPath(moduleKey)` function that returned `getModuleDataStoreDirPath(moduleKey) + externalModuleDataLockFileExt`

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -162,22 +173,44 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+		if err != nil {
+			p.logger.Debug("failed to get rlock for getModuleDataForModuleKey", zap.Error(err))
```
bufdev, line 204: We are returning these errors up the stack, is there a specific reason we need debug logger lines around locking and unlocking? Were these helpful in development?

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -317,11 +313,95 @@ func (p *moduleDataStore) putModuleData(
...
+		p.logDebugModuleKey(
+			moduleKey,
+			"module data store dir read write bucket",
+			zap.String("dirPath", dirPath),
+		)
+		moduleCacheBucket = storage.MapReadWriteBucket(p.bucket, storage.MapOnPrefix(dirPath))
+
+		// Only attempt to get a file locks when storing individual files.
+		// Before writing to the module directory, first get a shared lock and check module.yaml
+		var readUnlocker filelock.Unlocker
```
bufdev, line 329: Why are you declaring the variable instead of just doing `readUnlocker, err :=` on the next line?

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -317,11 +313,95 @@ func (p *moduleDataStore) putModuleData(
...
+				if err := readUnlocker.Unlock(); err != nil {
+					p.logger.Debug("failed to runlock for putModuleData", zap.Error(err))
+					retErr = multierr.Append(retErr, err)
+				}
+			}
+		}()
+		data, err := storage.ReadPath(ctx, moduleCacheBucket, externalModuleDataFileName)
+		p.logDebugModuleKey(
+			moduleKey,
+			fmt.Sprintf("module data store put read check %s", externalModuleDataFileName),
```
bufdev, line 346: Was this copied from previous code? I don't understand what this is, perhaps it was my fault though.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -263,49 +296,12 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
-		return p.bucket.Delete(ctx, tarPath)
-	}
-	dirPath, err := getModuleDataStoreDirPath(moduleKey)
-	if err != nil {
-		return err
-	}
-	return p.bucket.DeleteAll(ctx, dirPath)
-}
-
 func (p *moduleDataStore) putModuleData(
```
bufdev, line 299: It's really difficult to understand what the code in here is trying to accomplish now. We need specific documentation as we go, and comment within explaining what this code block is trying to accomplish. Something like "we're going to first check for X, if that's not the case, we're going to do this overwrite like this, etc etc".

bufdev, line 299: I don't see anywhere where invalid data is cleaned up anymore. Perhaps I missed it. What happens if I have invalid  data, say file "a.proto", and the valid data is only the file "b.proto", both written to same cache directory...will "a.proto" not be there anymore? Basically, I don't understand how the deleteInvalidModuleData case is now handled. It appears to only be handled for tar files (`getModuleDataForModuleKey` calls `bucket.Delete`).

private/bufpkg/bufmodule/bufmodulecache/bufmodulecache_test.go
```diff
@@ -214,6 +223,64 @@ func TestModuleDataProviderBasic(t *testing.T) {
...
+							return err
+						}
+					}
+					return nil
+				}, nil
+			},
+		)
+		require.NoError(t, err)
+		require.NoError(t, thread.Parallelize(context.Background(), jobs))
+		require.NoError(t, os.RemoveAll(cacheDir))
```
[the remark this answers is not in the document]
pkwarren: This is a subdirectory of the temp dir created by the test and we want to clear it after each attempt (to start fresh). This exposed race conditions that wouldn't be hit once we managed to successfully cache everything during an attempt.

private/buf/bufcli/cache.go
```diff
@@ -96,6 +97,12 @@ var (
 	//
 	// Normalized.
 	v3CacheWKTRelDirPath = normalpath.Join("v3", "wellknowntypes")
+	// v3CacheModuleLockRelDirPath is the relative path to the lock files directory for module data. This path
+	// is relative to the files cache directory for module data and is used to store lock files for
+	// synchronizing reading and writing module data from the cache.
+	//
+	// Normalized.
+	v3CacheModuleLockRelDirPath = "lock"
```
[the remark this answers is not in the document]
doriable: It is scoped to `v3`, I call `normalpath.Join(fullCacheDirPath, v3CacheModuleLockRelDirPath)` below, but it is probably better to set this explicitly here.

private/buf/bufcli/cache.go
```diff
@@ -166,12 +173,21 @@ func newModuleDataProvider(
 	if err != nil {
 		return nil, err
 	}
+	// Create filelocker directory and filelocker
+	if err := createCacheDir(container.CacheDirPath(), normalpath.Join(v3CacheModuleRelDirPath, v3CacheModuleLockRelDirPath)); err != nil {
+		return nil, err
+	}
+	filelocker, err := filelock.NewLocker(normalpath.Join(fullCacheDirPath, v3CacheModuleLockRelDirPath))
```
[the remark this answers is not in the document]
doriable: This joins on the `fullCacheDirPath`, which includes `v3CacheModuleRelDirPath`, but yes, the constant being clearer would make this more readable.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -89,6 +92,9 @@ type ModuleDataStoreOption func(*moduleDataStore)
 // ModuleDataStoreWithTar returns a new ModuleDataStoreOption that reads and stores
 // tar files instead of storing individual files in a directory in the bucket.
 //
+// The filelocker will be ignored when this is set, since we are storing all files associated
```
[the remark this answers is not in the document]
doriable: Yeah, that's fair, definitely an implementation detail that `tar` would not require the locking for synchornization.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: Yes, this was helpful for diagnosing the cache behaviours during development, but we can remove them.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: So the validity of the contents of `v3/modules/<digest_type>/<module_key>` is determined by the presence of a valid `module.yaml` file. As documented at the end of `getModuleDataForModuleKey`, we rely on `module.yaml` to be written last.

So the case you're describing, where there is an invalid `a.proto` but a valid `b.proto`, this occurs because of some interruption to the process while writing `a.proto`. In which case, no `module.yaml` would be written, this would not be considered valid, and then new data is fetched, etc.

doriable: So `getModuleDataForModuleKey` currently only does read operations, and each step of that read is documented... we can add more documentation, but I need some clarification on which parts.

As the last comment in `getModuleDataForModuleKey` addresses, validity of content is always determined by `module.yaml`, and that is always expected to be written last.

Any error and or invalid `module.yaml` is returned as an error, and then handled by fetching new data and putting it into the cache. There is nothing being overwritten/changed by `getModuleDataForModuleKey`, it is only reading the cache.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
doriable: This was not -- this is part of the comment above:

```
		// Before writing to the module directory, first get a shared lock and check module.yaml
```

We are reading the `module.yaml` preemptively before attempting to write new data to the cache.

private/buf/bufcli/cache.go
```diff
@@ -96,6 +98,11 @@ var (
 	//
 	// Normalized.
 	v3CacheWKTRelDirPath = normalpath.Join("v3", "wellknowntypes")
+	// v3CacheModuleLockRelDirPath is the relative path to the lock files directory for module data.
+	// This directory is used to store lock files for synchronizing reading and writing module data from the cache.
+	//
+	// Normalized.
+	v3CacheModuleLockRelDirPath = normalpath.Join("v3", "module_locks")
```
bufdev, line 105: `wellknowntypes` doesn't use underscores, this shouldn't either, for consistency - `modulelocks`

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
bufdev: This shouldn't be a PR comment - this should explained in the code for future readers who may question it. It should be clear to a reader at first glance why we are returning a `fs.PathError` instead of the error itself from reading the code comments - perhaps this wasn't the case in the past,  but it needs to be going forward.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
```diff
@@ -162,22 +166,46 @@ func (p *moduleDataStore) getModuleDataForModuleKey(
...
+			"module data store dir read write bucket",
+			zap.String("dirPath", dirPath),
+		)
+		moduleCacheBucket = storage.MapReadWriteBucket(p.bucket, storage.MapOnPrefix(dirPath))
+		moduleDataStoreDirLockPath, err := getModuleDataStoreDirLockPath(moduleKey)
+		if err != nil {
+			return nil, err
 		}
-	}()
+		// Only attempt to get a file lock when storing individual files
```
bufdev, line 198: I don't know what this means - what is an "individual file"? It should be clear from the code comments as to what this code does. This may take a paragraph or two of comments.

private/bufpkg/bufmodule/bufmodulestore/module_data_store.go
bufdev: Can you point me to the documentation? I can't follow what is going on here.

bufdev (approved):

PR #3140, opened 2024-07-08
File locking module data provider

doriable:
This adds file locking on `module.yaml` file to the cache module
data provider. It takes a lock on all requested `<module_key>/module.yaml` files
before cache operations (instead of taking a lock on each read/write operation
using the `ModuleDataStore`, like in #3139).

This implementation is less expensive in that it takes less file locks, however
each concurrent `buf` CLI invocation that uses the cache for overlapping
module data would be blocked.

Personally, I prefer the implementation in #3139 even though testing is a bit
more expensive (due to the increased number of locks taken and released).

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -281,4 +281,7 @@ issues:
       path: private/buf/cmd/buf/buf_test.go
       # The LsModules tests call chdir and cannot be parallelized.
       text: "LsModules"
-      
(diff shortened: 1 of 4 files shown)


doriable:
Going to close this for now, since based on discussions, the implementation in #3139 is preferable.

PR #3143, opened 2024-07-09
Add OAuth2 client to support device flow

emcfarlane:
Client implementation for the OAuth2 Device Authorization Grant flow defined by RFC 8628 with a registration endpoint defined by OpenID Connect Dynamic Client Registration.

https://datatracker.ietf.org/doc/html/rfc8628
https://openid.net/specs/openid-connect-registration-1_0.html


doriable (approved):

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,181 @@
...
+// and poll for the device access token.
+type Client struct {
+	baseURL string
+	client  *http.Client
+}
+
+// NewClient returns a new Client with the given base URL and HTTP client.
+func NewClient(baseURL string, client *http.Client) *Client {
+	return &Client{
+		baseURL: baseURL,
```
mfridman, line 39: Nit, could strip a potential `/` from the suffix to avoid `c.baseURL+DeviceRegistrationPath` failing later on.

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	rsp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer rsp.Body.Close()
```
mfridman, line 61: Nit, we typically catch this error in other places

```go
	defer func() {
		retErr = multierr.Append(retErr, resp.Body.Close())
	}()
```

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,181 @@
...
+		return nil, err
+	}
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Header.Set("Accept", "application/json")
+
+	rsp, err := c.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer rsp.Body.Close()
```
mfridman, line 93: Same here.

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+	if code := rsp.StatusCode; code != http.StatusOK {
+		return nil, fmt.Errorf("oauth2: invalid status: %v", code)
+	}
+	return &payload.DeviceAuthorizationResponse, nil
+}
+
+// AccessDeviceToken polls the authorization server for the device access token. The interval
+// parameter specifies the polling interval in seconds.
+func (c *Client) AccessDeviceToken(ctx context.Context, interval int, args *DeviceAccessTokenRequest) (*DeviceAccessTokenResponse, error) {
```
mfridman, line 113: Do you know if this `ctx` has a deadline when called? ~Should we have one to avoid this running indefinitely?~

Unrelated, but looking forward to go1.23 timer behavior, https://tip.golang.org/doc/go1.23#timer-changes

private/pkg/oauth2/client_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestRegisterDevice(t *testing.T) {
```
mfridman, line 29: I think the one test that would have been nice to have is the `ErrorCodeExpiredToken` and/or `ErrorCodeAccessDenied`, so we can test the fallthrough and return.

mfridman (approved):
Mainly nits, but otherwise lgtm. Awesome work.

private/pkg/oauth2/client.go
emcfarlane: It will be cancellable, but the context is tied tot the cli close too.

mfridman (approved):

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+	if code := resp.StatusCode; code != http.StatusOK {
+		return nil, fmt.Errorf("oauth2: invalid status: %v", code)
+	}
+	return &payload.DeviceAuthorizationResponse, nil
+}
+
+// AccessDeviceToken polls the authorization server for the device access token. The interval
+// parameter specifies the polling interval in seconds.
+func (c *Client) AccessDeviceToken(ctx context.Context, interval int, args *DeviceAccessTokenRequest) (*DeviceAccessTokenResponse, error) {
```
bufdev, line 119: `interval` should be a functional option, like:

```go
// AccessDeviceTokenWithPollingInterval returns a new AccessDeviceTokenOption that sets the polling interval.
//
// The default is 5 seconds. Polling may not be longer than 30 seconds.
func AccessDeviceTokenWithPollingInterval(pollingInterval time.Duration) AccessDeviceTokenOption {
  return func(accessDeviceTokenOptions *accessDeviceTokenOptions) {
    accessDeviceTokenOptions.pollingInterval = pollingInterval
  }
}
```

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+	encodedValues := args.ToValues().Encode()
+	if interval <= 0 {
+		interval = 5 // Default polling interval of 5 seconds.
```
bufdev, line 122: Make `5` and `30` private global consts at top of file

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+				return nil, &payload.Error
+			default:
+				return nil, &payload.Error
+			}
+		}
+	}
+}
+
+func (c *Client) parseJSONResponse(rsp *http.Response, payload any) error {
+	body, err := io.ReadAll(io.LimitReader(rsp.Body, 1<<20))
```
bufdev, line 179: `1<<20` should be a const at top of file

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+				// If the user has denied the device or the token has expired, return the error.
+				return nil, &payload.Error
+			default:
+				return nil, &payload.Error
+			}
+		}
+	}
+}
+
+func (c *Client) parseJSONResponse(rsp *http.Response, payload any) error {
```
bufdev, line 178: `rsp, resp, etc` should be `response`, same comment for request-like things should be `request` and not `req, args, etc`

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+	body, err := io.ReadAll(io.LimitReader(rsp.Body, 1<<20))
+	if err != nil {
+		return err
+	}
+	contentType, _, _ := mime.ParseMediaType(rsp.Header.Get("Content-Type"))
```
bufdev, line 183: Can scope within `if`

private/pkg/oauth2/client.go
```diff
@@ -0,0 +1,191 @@
...
+	if contentType != "application/json" {
+		return fmt.Errorf("%d %s", rsp.StatusCode, body)
```
bufdev, line 185: Couldn't this potentially be a 200 status code? This error feels off. At the least we need a code comment here.

private/pkg/oauth2/client.go
bufdev, line 178: Why does this need to be on `*Client`? It doesn't appear to use anything from `*Client` and should likely be a static function

private/pkg/oauth2/client_test.go
```diff
@@ -0,0 +1,310 @@
...
+	}
+}
+
+type roundTripFunc func(r *http.Request) (*http.Response, error)
+
+func (s roundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return s(r)
+}
+
+func newJSONResponse(t *testing.T, statusCode int, body string) *http.Response {
```
bufdev, line 286: `testNewJSONResponse` (prefix all test functions with `test`, just for consistency with the rest of the codebase)

private/pkg/oauth2/client.go
emcfarlane: Added a comment, included the content type and body. Most likely an error unrelated to oauth flow.

private/pkg/oauth2/client_test.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+}
+
+type roundTripFunc func(r *http.Request) (*http.Response, error)
+
+func (s roundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
+	return s(r)
+}
+
+func testNewJSONResponse(t *testing.T, statusCode int, body string) *http.Response {
```
bufdev, line 287: comment here was to do this for all the private functions, ie `testAssert.*`

bufdev (approved):

PR #3149, opened 2024-07-10
Add configuration for nuget plugins

pkwarren:
Add proto messages to support the Nuget repository (C# plugins).


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -183,6 +203,24 @@ message MavenConfig {
   repeated RuntimeConfig additional_runtimes = 3;
 }
 
+// NugetConfig is the configuration for a nuget C# plugin.
+message NugetConfig {
+  // RuntimeLibrary describes a runtime dependency of the generated code.
+  message RuntimeLibrary {
+    // The name of the nuget package.
```
stefanvanburen, line 210: Not sure if we want to call this out as a "Package ID".

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -89,6 +90,25 @@ enum PythonPackageType {
   PYTHON_PACKAGE_TYPE_STUB_ONLY = 2;
 }
 
+// DotnetTargetFramework specifies the target framework for dotnet plugins.
+// Reference: https://learn.microsoft.com/en-us/dotnet/standard/frameworks
```
stefanvanburen, line 94: do we want to add a comment as to why we've selected this subset of target frameworks?

stefanvanburen (approved):
looks good to me, mostly nitpicking on the docs / naming.

stefanvanburen (approved):

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -89,6 +90,27 @@ enum PythonPackageType {
   PYTHON_PACKAGE_TYPE_STUB_ONLY = 2;
 }
 
+// DotnetTargetFramework specifies the target framework for dotnet plugins.
+// This is not meant to be an exhaustive list - most plugins should target a
+// netstandard version.
+// Reference: https://learn.microsoft.com/en-us/dotnet/standard/frameworks
+enum DotnetTargetFramework {
```
bufdev, line 97: How often is this going to change? This might be better as a dynamic string field if it will change often

pkwarren: I don't think we'll need to change it very often at all - most plugins will require a pretty old version of this for compatibility with the broadest number of .NET runtimes.

Going forward we'd only add new values if they're required by a plugin.

bufdev (approved):

PR #3156, opened 2024-07-15
Add new cmake registry type

smaye81:
This adds a new registry type and config for cmake to the protobuf definitions. It also adds the plumbing for using this config for remote plugins.

Note that the registry config struct is intentionally empty since Cmake does not necessitate any sort of registry config (at least currently).


pkwarren (commented):
I think within the current constraints we should add an empty `CmakeConfig` message and add a field to the `oneof registry_config`. Even if we don't have options now, that provides a place for them to go and also makes the cmake support behave like other registries.

bufdev (approved):

PR #3158, opened 2024-07-15
Refactor bufprint package

oliversun9:
Printing stuff in `buf beta registry ...` can be put into three categories:
* Printing a page (with a page token)
* Printing the info of an entity (table form in `text` and detailed form in `json`)
* Printing the name of entities (full name in `text` and detailed from in `json`)

This PR replaces `[Commit,Label,Module,Organization]Printer` with three print functions.

This also adds `bufprint` struct tag to the `output[Commit/Label/Module/Organization]` types, but this tag is limited within the `bufprint` package.

`CuratedPluginPrinter` and `TokenPrinter` are left alone -- they can be updated/removed when we work on commands that use these printers.


oliversun9:
This will stay un-mergable for a while, until the commands promotion PRs are merged. Closing for now.

Issue #3161, opened 2024-07-16
`buf lint` Incorrect type of 'this' on repeated fields when using protovalidate with custom CEL expressions


finnhartshorn-anz:
### GitHub Repository

https://github.com/finnhartshorn-anz/buf-lint-protovalidate-bug

### Commands

```shell
buf lint
```

### Output

```shell
test/v1/bug.proto:9:31:(buf.validate.field).cel.expression on field "field_1" fails to compile: found no matching overload for 'size' applied to '(int)'
 | size(this) != 0
 | ....^
test/v1/bug.proto:14:31:(buf.validate.field).cel.expression on field "field_2" fails to compile: expression of type 'int' cannot be range of a comprehension (must be list, map, or dynamic)
 | this.filter(r, r > 4).unique()
 | ^
```

### Expected Output

Expected no lint failures, the CEL expressions work perfectly when compiled with client libraries (protovalidate-go in our case).

### Anything else?

Message that fails linting, `this` is interpreted as an int by the linter but is actually a list because the field is repeated
```
message Bug {
  repeated int32 field_1 = 1 [(buf.validate.field).cel = {
    id: "bug"
    message: "bug"
    expression: "size(this) != 0"
  }];
}
```

oliversun9:
Hi there, thank you for filing this bug! I am able to reproduce it and will look into it.

oliversun9:
This is fixed in https://github.com/bufbuild/buf/pull/3162 and will go out in the next release. Thank you.

PR #3162, opened 2024-07-16
Fix repeated field resolution in CEL for buf lint protovalidate rule

oliversun9:
This fixes https://github.com/bufbuild/buf/issues/3161.

`buf lint` for protovalidate should work correctly for both

```
repeated int32 ints = 1 [(buf.validate.field).repeated.items.cel = {
  id: "..."
  message: "..."
  expression: "this > 2"
}];
```

and 

```
repeated int32 ints = 1 [(buf.validate.field).cel = {
  id: "..."
  message: "..."
  expression: "size(this) > 2"
}];
```

This PR fixes the bug in second case where `buf lint` thought `this` is an `int32`, while it should be a _list_ of `int32`.

### DIFF
--- a/private/bufpkg/bufcheck/buflint/testdata/protovalidate/proto/cel_field.proto
+++ b/private/bufpkg/bufcheck/buflint/testdata/protovalidate/proto/cel_field.proto
@@ -8,140 +8,155 @@ import "google/protobuf/timestamp.proto";
 
 message Foo {
   required double compile_fail = 1 [(buf.validate.field).cel = {
-    id: "Foo.compile_fail",
-    message: "compile_fail_message",
+    id: "Foo.compile_fail"
+    message: "compile_fail_message"
     expression: "this > '1'"
   }];
   optional string some_expressions_fail = 2 [
     // this one fails
     (buf.validate.field).cel = {
-      id: "Foo.some_expressions_fail_1",
-      message: "some_expressions_fail_1_message",
+      id: "Foo.some_expressions_fail_1"
+      message: "some_expressions_fail_1_message"
       expression: "this > 5"
     },
     // this one succeeds
     (buf.validate.field).cel = {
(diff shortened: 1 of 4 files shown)


private/bufpkg/bufcheck/buflint/testdata/protovalidate/proto/cel_field.proto
```diff
@@ -8,140 +8,155 @@ import "google/protobuf/timestamp.proto";
 
 message Foo {
   required double compile_fail = 1 [(buf.validate.field).cel = {
-    id: "Foo.compile_fail",
-    message: "compile_fail_message",
+    id: "Foo.compile_fail"
+    message: "compile_fail_message"
```
oliversun9, line 12: I think `buf format` has been updated? Or my auto-format wasn't enabled this whole time. The former is more likely.

oliversun9: Also happy to revert these changes but I think they look nicer without the commas.

private/bufpkg/bufcheck/buflint/testdata/protovalidate/proto/cel_field.proto
```diff
@@ -8,140 +8,155 @@ import "google/protobuf/timestamp.proto";
...
+      message: "must have at least one value and all values must be greater than 2"
+      expression: "size(this) > 0 && this.all(x, x >= 2)"
+    },
+    // This is incorrect because "this" is not an int32 but a list of int32.
+    (buf.validate.field).cel = {
+      id: "min_value"
+      message: "each value must be greater than 2"
+      expression: "this > 2"
+    }
+  ];
```
oliversun9, line 161: This is the new test case.

bufdev (approved):

PR #3167, opened 2024-07-17
Update buf registry login to use browser by default

emcfarlane:
Updates the command `buf registry login` to use the brower login flow. On login, the users browser will be opened prompting the user to approve or deny access to the device. This simplifies the login flow avoiding the steps previously required to create an access token before providing it to the CLI. 

```
$ buf registry login
Opening your browser to complete authorization process.

If your browser doesn't open automatically,
please open this URL in a browser to complete the process:
        https://buf.build/settings/user/device?code=Z8YJ-8FGD

Logged in as emcfarlane. Credentials saved to /Users/edward/.netrc.
```

The flag `--no-browser` may be provided to disable the browser flow and prompt the user for the access token. The flag `--token-stdin` behaviour remains the same as before. Providing both flags is an error.
```
$ buf registry login --no-browser
Enter the BSR token created at https://buf.build/settings/user.

Token:
Logged in as emcfarlane. Credentials saved to /Users/edward/.netrc.
```


private/pkg/open/exec_windows.go
```diff
@@ -0,0 +1,50 @@
...
+	stderr := bytes.NewBuffer(nil)
+	if err := runner.Run(
+		ctx,
+		"cmd",
+		command.RunWithArgs("/c", "start"),
+		command.RunWithStdout(stdout),
+		command.RunWithStderr(stderr),
+		command.RunWithEnv(app.EnvironMap(envContainer)),
+	); err != nil {
+		return fmt.Errorf("could not open %q: %w, %s", input, err, stderr.String())
```
doriable, line 47: We should have this as a consistent error string/format in `open.go`, shared across all OS implementations.

private/pkg/open/exec_windows.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func open(
+	ctx context.Context,
+	runner command.Runner,
+	envContainer app.EnvContainer,
+	input string,
+) error {
+	// Replace characters that are not allowed by cmd/bash.
+	input = strings.ReplaceAll(input, "&", `^&`)
```
doriable, line 36: We should have these as constants in this file.

doriable (commented):
I tested the workflow on Windows, in Powershell, and it does not seem to behave as intended, instead of getting the browser open, it opened a new Command Prompt window, and then when I went to the link directly, an error occurred.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -211,11 +236,102 @@ func inner(
...
+		return "", err
+	}
+	appConfig, err := bufapp.NewConfig(container, externalConfig)
+	if err != nil {
+		return "", err
+	}
+	client := httpclient.NewClient(appConfig.TLS)
+	oauth2Client := oauth2.NewClient(baseURL, client)
+	// Register the device.
+	deviceRegistration, err := oauth2Client.RegisterDevice(ctx, &oauth2.DeviceRegistrationRequest{
```
emcfarlane, line 270: Could inspect the error case here to fallback to the `--no-browser` flow for a 501 not implemented error. This would allow pointing at older BSR instances that don't have the required auth endpoints.

doriable: I like this idea -- I think not breaking for older BSR versions is pretty nice. I wonder if we should have a note about this in the `--help` string, something along the lines of "for old versions of the Buf Schema Registry (BSR), buf registry login defaults to --no-browser".

emcfarlane: Theres `errors.ErrUnsupported` which seems nice to trigger on content type failures, or any other unsupported response types.

private/pkg/oauth2/client.go
```diff
@@ -28,6 +29,13 @@ import (
 	"go.uber.org/multierr"
 )
 
+var (
+	// ErrUnsupported is returned when we receive an unsupported response from the server.
+	//
+	// TODO(go1.23): replace by errors.ErrUnsupported once it is available.
```
mfridman, line 35: I thought this was added in go1.21?

> The new [ErrUnsupported](https://go.dev/pkg/errors/#ErrUnsupported) error provides a standardized way to indicate that a requested operation may not be performed because it is unsupported.

https://go.dev/doc/go1.21#errorspkgerrors

emcfarlane: Ops yep, depends on https://github.com/bufbuild/buf/pull/3123

private/pkg/oauth2/client_test.go
```diff
@@ -87,7 +87,7 @@ func TestRegisterDevice(t *testing.T) {
 				Body:       io.NopCloser(strings.NewReader(`server error`)),
 			}, nil
 		},
-		err: fmt.Errorf("oauth2: invalid response: 500 server error"),
+		err: fmt.Errorf("oauth2: %w: 500 server error", ErrUnsupported),
```
mfridman, line 90: Nit, should these set of tests be referencing HTTP 501 - Not Implemented?

emcfarlane: Yep, will be a 501 when intentionally disabling it, but I think old servers would 404 too. This test is just showing the error message is captured in the error.

doriable (approved):

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -140,31 +157,33 @@ func inner(
...
-	if !flags.TokenStdin {
-		if _, err := fmt.Fprintf(
-			container.Stdout(),
-			"Enter the BSR token created at https://%s/settings/user.\n\n",
-			remote,
-		); err != nil {
-			return err
-		}
+	if flags.TokenStdin && flags.NoBrowser {
+		return fmt.Errorf("cannot use both --%s and --%s flags", tokenStdinFlagName, noBrowserFlagName)
```
bufdev, line 161: This doesn't make sense to me - wouldn't you *exclusively* want to use `--token-stdin` when you are *not* using the browser?

I think this makes more sense if you name the flag for what it will do, not what it won't - ie `--prompt`

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -63,6 +70,7 @@ The <domain> argument will default to buf.build if not specified.`, netrc.Filena
 type flags struct {
 	Username   string
 	TokenStdin bool
+	NoBrowser  bool
```
bufdev, line 73: I would rename this to `--prompt`, `promptFlagName`, `Prompt`

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -211,11 +230,126 @@ func inner(
...
+	return deviceToken.AccessToken, nil
+}
+
+// getClientName returns the client name for the device registration.
+func getClientName() (string, error) {
+	hostname, err := os.Hostname()
+	if err != nil {
+		return "", err
+	}
+	switch runtime.GOOS {
```
bufdev, line 349: nit: generally better pattern for this is using a variable and then having separate files for different OS's. See ie https://github.com/bufbuild/buf/tree/main/private/pkg/app app_unix.go app_windows.go

private/pkg/oauth2/client.go
```diff
@@ -28,6 +29,13 @@ import (
 	"go.uber.org/multierr"
 )
 
+var (
+	// ErrUnsupported is returned when we receive an unsupported response from the server.
+	//
+	// TODO(go1.21): replace by errors.ErrUnsupported once it is available.
```
bufdev, line 35: Once this PR is merged, make sure to merge main into https://github.com/bufbuild/buf/pull/3123 and delete this

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
bufdev: This was not fully addressed. Please address it before resolving.

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
emcfarlane: Is there a best practise for mentioning the flag? `--%s` or `%q`?

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -140,31 +158,33 @@ func inner(
...
-	if !flags.TokenStdin {
-		if _, err := fmt.Fprintf(
-			container.Stdout(),
-			"Enter the BSR token created at https://%s/settings/user.\n\n",
-			remote,
-		); err != nil {
-			return err
-		}
+	if flags.TokenStdin && flags.Prompt {
+		return fmt.Errorf("cannot use both --%s and --%s flags", tokenStdinFlagName, promptFlagName)
```
bufdev, line 162: One more thing - this should be a `appcmd.NewInvalidArgumentErrorf`

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
emcfarlane: Changing to `--%s`.

bufdev (approved):

private/pkg/oauth2/client.go
bufdev: I did this FYI

PR #3175, opened 2024-07-22
Install zip before using it in the release process

oliversun9:
This installs `zip` with `apt-get` in `release.bash`, when it's run inside a docker image.

### DIFF
--- a/make/buf/scripts/release.bash
+++ b/make/buf/scripts/release.bash
@@ -140,6 +140,11 @@ for os in Darwin Linux; do
   done
 done
 
+apt-get -y update
+apt-get -y autoremove
+apt-get clean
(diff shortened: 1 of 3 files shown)


make/buf/scripts/release.bash
```diff
@@ -140,6 +140,11 @@ for os in Darwin Linux; do
   done
 done
 
+apt-get -y update
+apt-get -y autoremove
+apt-get clean
+apt-get install -y zip
+
```
doriable, line 147: This is my fault, lol, but yeah, given that we want to make sure `release.bash` runs locally on all machines, instead of installing here, we should create a Dockerfile instead.

doriable (approved):

PR #3181, opened 2024-07-23
Add a comment with link to docs for `buf config init`

### LINKED ISSUE #3180: Add link to docs in `buf config init`
### Feature

`buf config init` currently generates:

```yaml
version: v2
lint:
  use:
    - DEFAULT
breaking:
  use:
    - FILE
```

I effectively always need to reference https://buf.build/docs/configuration/v2/buf-yaml in order to specify more configuration. It would be nice if the generated config could directly link to that page, perhaps in a comment, so that it's easier to get started editing the configuration.

Something like:

```yaml
# https://buf.build/docs/configuration/v2/buf-yaml
version: v2
lint:
  use:
    - DEFAULT
breaking:
  use:
    - FILE
```

I don't think `buf config init` can generate non-`v2` buf.yaml files, but the URL should correspond to the version of the file being created.

doriable:
This adds a comment at the top of the `buf.yaml` file generated from
`buf config init` with a link to our public docs to make it easier for users
who are onboarding to get more information on `buf.yaml` configs.

Running `buf config init` generates a `buf.yaml` with the following contents:

```
# For details on buf.yaml configuration, visit https://buf.build/docs/configuration/v2/buf-yaml
version: v2
lint:
  use:
    - DEFAULT
breaking:
  use:
    - FILE

```

The comment is preserved when round-tripping `buf.yaml`.

Fixes #3180

### DIFF
--- a/private/bufpkg/bufconfig/buf_yaml_file.go
+++ b/private/bufpkg/bufconfig/buf_yaml_file.go
@@ -131,11 +131,11 @@ func NewBufYAMLFile(
 type BufYAMLFileOption func(*bufYAMLFileOptions)
 
 // BufYAMLFileWithIncludeDocsLink returns a new BufYAMLFileOption that specifies including
-// a comment with a link to the public docs at the top of the buf.yaml file. The comment
-// is included when set to true.
-func BufYAMLFileWithIncludeDocsLink(includeDocsLink bool) BufYAMLFileOption {
+// a comment with a link to the public docs for the appropriate buf.yaml version at the top
+// of the buf.yaml file.
+func BufYAMLFileWithIncludeDocsLink() BufYAMLFileOption {
 	return func(bufYAMLFileOptions *bufYAMLFileOptions) {
-		bufYAMLFileOptions.includeDocsLink = includeDocsLink
+		bufYAMLFileOptions.includeDocsLink = true
 	}
 }
 
--- a/private/bufpkg/bufconfig/buf_yaml_file.go
+++ b/private/bufpkg/bufconfig/buf_yaml_file.go
@@ -44,6 +44,7 @@ const (
 	// we released with it, however.
 	oldBufYAMLFileName        = "buf.mod"
 	defaultBufYAMLFileVersion = FileVersionV1Beta1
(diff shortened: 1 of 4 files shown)


private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -106,17 +110,35 @@ func NewBufYAMLFile(
...
 	)
 }
 
+// BufYAMLFileOption is an option for a new BufYAMLFile
+type BufYAMLFileOption func(*bufYAMLFileOptions)
+
+// BufYAMLFileWithIncludeDocsLink returns a new BufYAMLFileOption that specifies including
+// a comment with a link to the public docs at the top of the buf.yaml file. The comment
+// is included when set to true.
+func BufYAMLFileWithIncludeDocsLink(includeDocsLink bool) BufYAMLFileOption {
```
oliversun9, line 136: nit: would we prefer `BufYAMLFileWithIncludeDocsLink()` since it's false by default?

doriable: Yes, that makes a lot of sense to me.

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -328,6 +365,9 @@ func readBufYAMLFile(
 	if err != nil {
 		return nil, err
 	}
+	// Check if this file has the docs link comment from "buf config init" as its first line
+	// so we can preserve the comment in our round trip.
+	includeDocsLink := bytes.HasPrefix(data, []byte(fmt.Sprintf(docsLinkComment, fileVersion.String())))
```
oliversun9, line 370: I think if we decide to change `docsLinkComment`'s value in the future, this won't pick up the header. Are we ok with that?

oliversun9 (approved):
LGTM! I left two non-blocking comments

private/bufpkg/bufconfig/buf_yaml_file.go
doriable: I think that if we change the shape of the config file (e.g. the contents of the comment and/or where it is in the file, etc.), we'll need to decide on what we do and don't preserve on round trips, and we would change the parsing logic accordingly.

bufdev (approved):

PR #3182, opened 2024-07-23
Validate domain in buf registry login and logout

pkwarren:
We don't do explicit validation of the domain in 'buf registry login', which leads to some confusing error messages. Update the command to validate the domain with netext.ValidateHostname and update ValidateHostname to not accept URLs (it previously didn't look up the port number so it didn't perform proper validation).

### DIFF
--- a/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
+++ b/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
@@ -29,6 +29,7 @@ import (
 	"github.com/bufbuild/buf/private/pkg/app/appcmd"
 	"github.com/bufbuild/buf/private/pkg/app/appext"
 	"github.com/bufbuild/buf/private/pkg/connectclient"
+	"github.com/bufbuild/buf/private/pkg/netext"
 	"github.com/bufbuild/buf/private/pkg/netrc"
 	"github.com/spf13/pflag"
 )
@@ -135,6 +136,9 @@ func inner(
 	remote := bufconnect.DefaultRemote
 	if container.NumArgs() == 1 {
 		remote = container.Arg(0)
+		if _, err := netext.ValidateHostname(remote); err != nil {
+			return err
+		}
 	}
 	// Do not print unless we are prompting
 	if !flags.TokenStdin {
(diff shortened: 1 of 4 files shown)


private/pkg/netext/netext.go
```diff
@@ -37,8 +38,12 @@ func ValidateHostname(hostname string) (string, error) {
 	}
 
 	parsedHost := hostname
-	if host, _, err := net.SplitHostPort(hostname); err == nil {
-		parsedHost = host
+	if host, port, err := net.SplitHostPort(hostname); err == nil {
+		// net.SplitHostPort performs very lax validation of ports (allowing to be resolved from /etc/services).
+		// Only accept numeric ports here.
+		if portNum, err := strconv.ParseUint(port, 10, 32); err == nil && portNum <= 65535 {
```
pkwarren, line 44: There were two options here:
- Use strconv and validation (this matches protovalidate-go).
- Allow non-numeric ports and use `net.LookupPort("ip", port)` here - https://pkg.go.dev/net#LookupPort. This requires calling out to the name resolver and felt like overkill here.

pkwarren: Ref: https://github.com/bufbuild/protovalidate-go/blob/ec23f34e22ff982547206a8f7c86bc0bf2ad4091/celext/lib.go#L537-L540

private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -135,6 +136,9 @@ func inner(
 	remote := bufconnect.DefaultRemote
 	if container.NumArgs() == 1 {
 		remote = container.Arg(0)
+		if _, err := netext.ValidateHostname(remote); err != nil {
```
emcfarlane, line 139: Should do in `registrylogout.go` too.

pkwarren: Good catch - fixed.

emcfarlane (approved):

doriable (approved):

bufdev (approved):

PR #3184, opened 2024-07-24
Only check directories in workspace targeting fallback logic

### LINKED ISSUE #3183: `buf build` thinks .proto file is directory that should contain Buf workspace file
### GitHub Repository

https://github.com/bufbuild/buf

### Commands

```shell
// Given a filename, return a DescFile.
export function readProtoFile(
  filename: string,
  protoRoot: string
): protobuf.DescFile {
  // Create a temporary file.
  const tmpDir = tmp.dirSync({ unsafeCleanup: true });

  const imagePath = path.join(tmpDir.name, "image.bin");

  // Use buf to generate an image of the file descriptor.
  const args = ["build", "-o", imagePath, protoRoot, "--path", filename];
  const result = spawnSync("buf", args);
  let error;
  if (result.error) {
    error = result.error;
  } else if (result.status !== 0) {
    error = result.stderr.toString();
  }

  if (error) {
    throw new Error(
      `buf build failed: ${error}\n\ncommand: buf ${args.join(" ")}`
    );
  }

  // ...
}
```

### Output

```shell
buf build failed: Failure: stat /tmp/vnJL2j/path/to/my.proto/buf.work.yaml: not a directory


    command: buf build -o /tmp/tmp-545545-9JZ9IzSiyrH2/image.bin /tmp/vnJL2j --path /tmp/vnJL2j/path/to/my.proto
```

### Expected Output

`buf build` should succeed

### Anything else?

This is happening in the unit tests of an NPM package with a dependency that runs buf build to compile .proto files and then do stuff with the FileDescriptorSet.  The issue occurs on GHA Linux workers, but I can't reproduce it locally on my Windows laptop (where the tests pass).

doriable:
When calling our fallback logic checking for workspaces, we
check in the target paths to see if a valid workspace/module
is found. However, we should only be checking directories and
not target paths, such as `--path path/to/my.proto` (we want to
check `path/to`).

Tests have been added for both the relative and absolute directory
inputs without `buf.yaml` files.

Fixes #3183

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Fix the git input parameter `ref` to align with the `git` notion of a ref. This allows for the use 
   of branch names, tag names, and commit hashes.
+- Fix bug to check the directory of target proto file paths in fallback logic when scanning
+  for workspace and/or module config files
 
 ## [v1.35.0] - 2024-07-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@
 
 - Fix the git input parameter `ref` to align with the `git` notion of a ref. This allows for the use 
   of branch names, tag names, and commit hashes.
-- Fix bug to check the directory of target proto file paths in fallback logic when scanning
-  for workspace and/or module config files
+- Fix unexpected `buf build` errors with absolute path directory inputs without workspace and/or
+  module configurations (e.g. `buf.yaml`, `buf.work.yaml`) and proto file paths set to the `--path` flag.
 
 ## [v1.35.0] - 2024-07-22
 
--- a/private/buf/bufworkspace/workspace_targeting.go
+++ b/private/buf/bufworkspace/workspace_targeting.go
@@ -625,6 +625,14 @@ func checkForControllingWorkspaceOrV1Module(
 	ignoreWorkspaceCheck bool,
 ) (buftarget.ControllingWorkspace, error) {
 	path = normalpath.Normalize(path)
+	// We attempt to check that the provided target path is not a file by checking the extension.
+	// Any valid proto file provided as a target would have the .proto extension, so we treat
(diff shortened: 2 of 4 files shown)


private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -625,6 +625,14 @@ func checkForControllingWorkspaceOrV1Module(
 	ignoreWorkspaceCheck bool,
 ) (buftarget.ControllingWorkspace, error) {
 	path = normalpath.Normalize(path)
+	// We attempt to check that the provided target path is not a file by checking the extension.
```
bufdev, line 628: Any chance that `LICENSE` could mess this up somehow?

bufdev (commented):
Also do a changelog entry so we can do a patch release if we want

private/buf/bufworkspace/workspace_targeting.go
doriable: If the user provides `--path LICENSE` or `--path text_file_no_extension`, then it would be an invalid when we attempt to check for a workspace (we would expect to get the `Is not a directory` error) and that would be a valid error. We basically want to make sure that when the user is providing a valid target path that is a proto file, we check the directory it is in for a buf config file.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 - Fix the git input parameter `ref` to align with the `git` notion of a ref. This allows for the use 
   of branch names, tag names, and commit hashes.
+- Fix bug to check the directory of target proto file paths in fallback logic when scanning
```
bufdev, line 7: This says what the logic is doing, not what the bug was and what the user impact of it was. I'd expect something more along the lines of "Fix issue where [whatever was happening in 3183]", does that make sense? Also complete sentences (ie period at end)

bufdev (approved):

Issue #3185, opened 2024-07-24
Storage bucket with different behaviours


doriable:
### GitHub Repository

https://github.com/bufbuild/buf

### Commands

```shell
buf build -o out.bin path/to/input --path path/to/input/my.proto
buf build -o out.bin /abs/path/to/input --path /abs/path/to/input/my.proto
```

### Output

```shell
Failure: stat /abs/path/to/input/my.proto/buf.work.yaml: not a directory
```

### Expected Output

When investigating #3183, it seems that in the case of a relative input (and subsequently a storage bucket with a relative root), the error was not surfaced, and only surfaced with the absolute input (which has a storage bucket with an absolute root path).

I would expect the `stat` error in both cases before the fix in #3184.

PR #3188, opened 2024-07-24
Fix storage OS bucket not returning ErrNotExist correctly

oliversun9:
This PR fixes a bug of the storage bucket where it does not return an `fs.ErrNotExist` error when it should.

It happens when both conditions met:
* The bucket is created with an absolute path, i.e. `provider.NewReadWriteBucket("/absPath")`
* `bucket.Get(ctx, "existing.file/foo")` is called, where `existing.file` exists as a regular file.

Fixes https://github.com/bufbuild/buf/issues/3185

### DIFF
--- a/private/pkg/storage/storageos/bucket.go
+++ b/private/pkg/storage/storageos/bucket.go
@@ -20,7 +20,6 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"github.com/bufbuild/buf/private/pkg/filepathext"
 	"github.com/bufbuild/buf/private/pkg/normalpath"
@@ -284,12 +283,6 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
 		// It's important that we detect these cases so that
 		// multi buckets don't unnecessarily fail when one of
 		// its delegates actually defines the path.
-		elements := strings.Split(normalpath.Normalize(externalPath), "/")
-		if len(elements) == 1 {
-			// The path is a single element, so there aren't
-			// any other files to check.
-			return err
-		}
 		lastParentPath := externalPath
 		parentPath := filepath.Dir(externalPath)
 		for ; parentPath != lastParentPath; lastParentPath, parentPath = parentPath, filepath.Dir(parentPath) {
--- a/private/pkg/storage/storageos/bucket.go
+++ b/private/pkg/storage/storageos/bucket.go
@@ -290,8 +290,10 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
 			// any other files to check.
 			return err
 		}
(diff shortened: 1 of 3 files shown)


private/pkg/storage/storageos/bucket.go
```diff
@@ -284,14 +283,10 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
 		// It's important that we detect these cases so that
 		// multi buckets don't unnecessarily fail when one of
 		// its delegates actually defines the path.
-		elements := strings.Split(normalpath.Normalize(externalPath), "/")
```
oliversun9, line 287: Note if the path is absolute, the first element will be `""`.

private/pkg/storage/storageos/bucket.go
```diff
@@ -284,14 +283,10 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
...
-		if len(elements) == 1 {
-			// The path is a single element, so there aren't
-			// any other files to check.
-			return err
-		}
-		for i := len(elements) - 1; i >= 0; i-- {
-			parentFileInfo, err := os.Stat(filepath.Join(elements[:i]...))
```
oliversun9, line 294: According to [Join's doc](https://pkg.go.dev/path/filepath#Join), empty elements are ignored. This means if external path is `/foo/bar/baz`, it would check these paths here:

* `foo/bar`
* `foo`

but it should check:

* `/foo/bar`
* `/foo`

private/pkg/storage/storageos/bucket.go
```diff
@@ -284,14 +283,10 @@ func (b *bucket) validateExternalPath(path string, externalPath string) error {
...
+		lastParentPath := externalPath
+		parentPath := filepath.Dir(externalPath)
+		for ; parentPath != lastParentPath; lastParentPath, parentPath = parentPath, filepath.Dir(parentPath) {
+			parentFileInfo, err := os.Stat(parentPath)
```
emcfarlane, line 289: Could keep the file normalization and use `path.Dir`.
```
currentDir := path.Dir(externalPath)
for currentDir != "." && currentDir != "/" {
        //...
}
```

emcfarlane (approved):

doriable (approved):
Thanks for finding this, was a tricky one with `strings.Split` on absolute paths. If this merges, we can actually remove the file extension heuristic introduced in #3184 (but still keep the tests):
https://github.com/bufbuild/buf/blob/5406aa0e2a4ec4eb18964c5ced3217be34c899d9/private/buf/bufworkspace/workspace_targeting.go#L628-L635

bufdev (approved):

doriable (approved):

doriable (approved):

PR #3190, opened 2024-07-25
Fix b5 digest calculation for remote modules

saquibmian:
See the comment in `added_module.go` for details.


bufdev (commented):
I'm extremely uncomfortable with this solution, and it's going to require some deep in-person conversations to resolve. Overall concerns:

- For remote modules, we're just completely bypassing any local calculation of what the digest is, based on how we expect digests to be calculated, and relying on the remote calculation. This feels subtly iffy in a few ways.
- I'm not confident that tamper-proofing will actually occur after this is done - I would need to spend more time on this, but while it "happens at the ModuleData" level, I don't understand why we can effecvtively do the old calculation method at the ModuleData level with no issue, but we need to do this bypass at the Module level.

With more time, I could parse this out, but it's not clear to me, and I'm concerned that this is going to introduce a nasty long-term bug. Let's talk about this next week.

saquibmian:
Closed in favor of #3631.

PR #3201, opened 2024-07-31
Fix deprecated flag casing for flags

emcfarlane:
Continuation of https://github.com/bufbuild/buf/pull/3200, lowercases the remaining deprecated flags.


oliversun9 (approved):
LGTM, also verified that this covers all usages of `flagSet.MarkDeprecated`

stefanvanburen (approved):

Issue #3056, opened 2024-06-06
include_imports does not respect 'strategy: directory'


abhinav:
### GitHub Repository

https://github.com/abhinav/buf-include_imports-strategy-repro

### Commands

```shell
buf generate
```

### Output

```shell
protoc: common/v1/value.proto, keyvalue/v1/service.proto
```

### Expected Output

protoc: common/v1/value.proto
protoc: keyvalue/v1/service.proto

### Anything else?

I wrote a minimal plugin in the reproduction repo that just logs the files to generate to demonstrate the problem.

I'm using the latest release of Buf as of writing this.

```
❯ buf --version
1.32.2
```

The only workaround right now is to manually list imports in the `inputs` section.
Without this, old plugins like Twirp [explode during package name resolution](https://github.com/twitchtv/twirp/blob/v8.1.3/protoc-gen-twirp/generator.go#L192-L208).

doriable:
Thank you for providing a clear example. In the use-case of Twirp, specifically, is there a reason why you would need `include_imports` for the `protoc-gen-twirp` plugin? For that particular use-case, it seems `protoc-gen-twirp` should not need to generate additional code for the import file(s).

bufdev:
Thanks for the detailed issue here.

I'll try to explain what is happening here, and why I think this may be a documentation issue on our part - in effect, `include_imports` for stub generation isn't actually a feature available in `protoc`, and there is no equivalent to `include_imports: true` with `strategy: directory` in protoc.

`strategy: directory` is an easy way to split up your generation invocation into multiple independent protoc plugin requests, by directory, which is as you mentioned a common need in the Golang ecosystem for many plugins (including Twirp). The logic in a nutshell (a little more obscured under the hood but):

- Take your non-import input files and split them by directory.
- For each directory of non-import files:
  - Create a new `CodeGeneratorRequest`
  - Add the directory's non-import file names into `file_to_generate`.
  - Add the directory's non-import files AND all of those files' imports into `proto_file`. `CodeGeneratorRequests` need to be self-contained, so anything that the non-import files import need to be within `proto_file`, but won't have code generated for them.
  - Send this `CodeGeneratorRequest` to your plugin and get back the result.

`include_imports: true` modifies this `CodeGeneratorRequest` by just adding all the imports also into `file_to_generate`.

[...]

PR #3203, opened 2024-08-01
[WIP] Split imports as well for directory strategy in buf generate

oliversun9:
This is still WIP, but it seems to work with the base case.

TODOs:
* testing
* cleanup the code
* there are still places (`*_test.go` and `.../alpha/protoc/plugin.go`) that call `ImageByDir`, they should stop calling it.

Note, this cannot be achieved by simply removing the import check [here](https://github.com/bufbuild/buf/blob/429c3acea420a3e3577494860a3fb0f47806d899/private/bufpkg/bufimage/bufimage.go#L527), because the split images will conflate `IsImport` with _whether we want to generate a file if permitted by include-imports_. The latter is introduced to a new type `ImageFileForGeneration`.

Fixes https://github.com/bufbuild/buf/issues/3056

### DIFF
--- a/private/buf/bufgen/generator.go
+++ b/private/buf/bufgen/generator.go
@@ -324,6 +324,13 @@ func (g *generator) execLocalPlugin(
 	if err != nil {
 		return nil, err
 	}
+	// fmt.Println("IMAGES:")
+	// for _, image := range pluginImages {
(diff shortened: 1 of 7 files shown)


bufdev:
@doriable didn't I have a partial PR for this? Is this using that?

bufdev:
Should we close this now?

oliversun9:
Yes, chatted with @doriable, we can close this one now.

PR #3205, opened 2024-08-01
Add Set type to slicesext

emcfarlane:
Adds a new type `Set` that can replace the use of `map[T]struct{}` or be cast to and from to gain the helper methods as needed. Converted the two methods `ToStructMap` and `ToStructMapOmitEmpty` to return the new type, with the respective name change `ToSet` and `ToSetOmitEmpty`. Converted some of the use cases to show usability of the new methods, but not exhaustive. Intended to be initialized with make `make(slicesext.Set[string], size)` to allow setting size or not. `Delete` and `Len` helpers are added for convenience but can simply call the map functions.

### DIFF
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -1107,10 +1107,10 @@ func (c *controller) warnUnconfiguredTransitiveImports(
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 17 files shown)


bufdev:
Why do we need this? I think generally in my experience with go, making a concrete `Set` type has been quasi-frowned-upon, and `map[T]struct{}` is a pretty well-known standard.

emcfarlane:
Closing for now, not useful enough.

PR #3206, opened 2024-08-02
Update buf push to include top level license and doc file

oliversun9:
This updates module-building behavior. 

If a module pointed by a `buf.yaml` v2 does not have its own license file or doc file (e.g. `README.md`) but there is a license file or doc file next to `buf.yaml`, `buf push` will use those instead.


doriable:
Just leaving a quick note here: since this change affects the way `v2` workspaces are built, fundamentally, we should run our integration tests (deps, tamper-proofing, etc.) against this change, along with the tests included here.

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -576,6 +599,20 @@ func getMappedModuleBucketAndModuleTargeting(
 	return mappedModuleBucket, moduleTargeting, nil
 }
 
+func isBucketEmpty(ctx context.Context, bucket storage.ReadBucket) (bool, error) {
```
emcfarlane, line 602: This can be replaced with [`storage.IsEmpty`](https://github.com/bufbuild/buf/blob/777f4fa5f17f1eb2343fbb85021384f87daaffec/private/pkg/storage/util.go#L161)

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -557,10 +562,28 @@ func getMappedModuleBucketAndModuleTargeting(
 			),
 		)
 	}
+	docBucket := bufmodule.GetDocStorageReadBucket(ctx, moduleBucket)
+	licenseBucket := bufmodule.GetLicenseStorageReadBucket(moduleBucket)
+	if useTopLevelLicenseDoc {
+		docBucketEmpty, err := isBucketEmpty(ctx, docBucket)
```
emcfarlane, line 568: I think this logic could be moved to the `bufmodule` package, with a func signature of `GetDocStorageReadBucket(ctx, bucket, moduleDir, useTopLevelLicenseDoc)`. Then instead of building an empty bucket and checking for emptiness, we can stat the files in order and return a mapped bucket.

emcfarlane: Do you know if we need to have two separate methods for the buckets and then combine them? Would be nice to use a single set of Matchers with an OR for the doc or license file.

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -165,6 +165,7 @@ func v2WorkspaceTargeting(
 	bucket storage.ReadBucket,
 	bucketTargeting buftarget.BucketTargeting,
 	bufYAMLFile bufconfig.BufYAMLFile,
+	defaultToTopLevelLicenseDoc bool,
```
bufdev, line 168: Needs code comment as to why this is called `defaultToTopLevelLicenceDoc` and the other one is called `useTopLevelLicenseDoc` - regardless, needs code comments as to what the behavior is depending on the function of this variable

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -557,10 +562,28 @@ func getMappedModuleBucketAndModuleTargeting(
 			),
 		)
 	}
+	docBucket := bufmodule.GetDocStorageReadBucket(ctx, moduleBucket)
```
bufdev, line 565: This whole section needs comments, almost above every line - what is the exact logic here? Explain this code as if you were explaining it out loud to someone, but in the comments.

bufdev:
Note to self: this is reviewed and is waiting for updates from @oliversun9

bufdev (approved):

PR #3218, opened 2024-08-07
Allow use of pre-compiled descriptors as dependencies/inputs for an image

jhump:
This is needed to work with CSR clients, which may provide elements of the schema in the form of a descriptor proto instead of source. So we need something like this to be able to compile the provided schema, coupled with sources for other files/references.

DM me in Slack if you need more context.
cc @pkwarren

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -174,7 +174,7 @@ issues:
     - linters:
         - containedctx
       # we actually want to embed a context here
-      path: private/bufpkg/bufimage/parser_accessor_handler.go
+      path: private/bufpkg/bufimage/file_resolver.go
     - linters:
         - containedctx
       # we actually want to embed a context here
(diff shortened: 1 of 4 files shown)


bufdev (approved):
Yea seems reasonable. If `buf alpha protoc` was also updated to add `--descriptor_set_in`, I'd dole out gold stars.

jhump:
@pkwarren discovered that this isn't quite sufficient. And instead of augmenting this (and making it more complicated to satisfy the issue he encountered), he has a rather elegant work-around that doesn't require any new options in `bufimage`. So closing this out.

> Yea seems reasonable. If `buf alpha protoc` was also updated to add `--descriptor_set_in`, I'd dole out gold stars.

I think that'll have to wait for another day!

PR #3225, opened 2024-08-11
Fix close on write storage mappers

emcfarlane:
Param is missing on constructor, close isn't called.


bufdev (approved):
Wow, not fantastic on my part from a long time ago

PR #3226, opened 2024-08-11
Flatten normalpath EqualsOrContainsPath check

emcfarlane:
This is a small optimization around the `EqualsOrContainsPath` method. It's called repeatedly over a filesystem when walking and using mappers or matchers. The check is now constant time for a given length. Added a micro benchmark to showcase on unix. Didn't adjust the windows method as theres some changes needed for volume handling.
```
                             │    sec/op     │    sec/op     vs base                 │
EqualsOrContainsPath/Short-8    45.36n ± ∞ ¹   15.56n ± ∞ ¹        ~ (p=1.000 n=1) ²
EqualsOrContainsPath/Long-8    458.20n ± ∞ ¹   15.56n ± ∞ ¹        ~ (p=1.000 n=1) ²
geomean                         144.2n         15.56n        -89.21%
```

### DIFF
--- a/private/pkg/normalpath/normalpath_unix_test.go
+++ b/private/pkg/normalpath/normalpath_unix_test.go
@@ -319,6 +319,23 @@ func TestEqualsOrContainsPath(t *testing.T) {
 	testEqualsOrContainsPath(t, true, "b", "b/a/c")
 }
 
+func BenchmarkEqualsOrContainsPath(b *testing.B) {
+	b.Run("Short", func(b *testing.B) {
+		var ok bool
+		for i := 0; i < b.N; i++ {
+			ok = EqualsOrContainsPath("a/b/c", "a/b/c/d", Relative)
+		}
+		assert.True(b, ok)
+	})
+	b.Run("Long", func(b *testing.B) {
+		var ok bool
+		for i := 0; i < b.N; i++ {
+			ok = EqualsOrContainsPath("a/b/c", "a/b/c/d/e/f/g/h/i/j", Relative)
+		}
+		assert.True(b, ok)
+	})
+}
+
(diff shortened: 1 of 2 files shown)


private/pkg/normalpath/normalpath_unix_test.go
```diff
@@ -319,6 +319,23 @@ func TestEqualsOrContainsPath(t *testing.T) {
...
+	})
+	b.Run("Long", func(b *testing.B) {
+		var ok bool
+		for i := 0; i < b.N; i++ {
+			ok = EqualsOrContainsPath("a/b/c", "a/b/c/d/e/f/g/h/i/j", Relative)
+		}
+		assert.True(b, ok)
+	})
+}
+
```
emcfarlane, line 338: +       pathRoot := "/"
+       if pathType == Relative {
+               pathRoot = "."
        }
-       return false
+       // If the value is the root, it contains everything.
+       return value == pathRoot ||
+               // If the value is the path, it contains itself.
+               strings.EqualFold(value, path) ||
+               // If the value is a directory and the path is in the directory.
+               // Check has prefix and the next character is a path separator.
+               (len(path) > len(value) &&
+                       strings.EqualFold(path[:len(value)], value) &&
+                       (value[len(value)-1] == '/' || path[len(value)] == '/'))

bufdev (commented):
Generally not in favor of these types of changes - unless there's a real-world measurable reason to do this cleanup, we're changing the logic of one of the functions that backs a lot of our code.

This particular one might be OK, but there's also an overhead for the CLI team to consider edge cases here, which we don't have a lot of time to do. Overall, not sure it's worth the review and risk burden.

emcfarlane:
Can close, no real world diff as it's dominated by io.

PR #3229, opened 2024-08-12
Separate matchers as a bucket filter

emcfarlane:
Adds a new read bucket Filter that works exclusively on Matchers. This allows for further optimizations by using the Matchers to constrain the bucket walk path (see #3230). Buckets that use Matchers as Mappers have been split as a Mapper and Filter bucket. The method MapPrefix on Mappers has been removed. Matchers no longer implement Mappers.


private/pkg/storage/mapper.go
```diff
@@ -28,12 +28,6 @@ type Mapper interface {
 	// The returned path is expected to be normalized and validated.
 	// If the path cannot be mapped, this returns false.
 	MapPath(path string) (string, bool)
-	// Map maps the prefix to the full prefix.
-	//
-	// The path is expected to be normalized and validated.
-	// The returned path is expected to be normalized and validated.
-	// If the path cannot be mapped, this returns false.
-	MapPrefix(prefix string) (string, bool)
```
emcfarlane, line 36: `MapPrefix` can be removed and replaced with calls to `MapPath` now that `Matcher`s won't restrict the mappers.

private/pkg/storage/matcher.go
```diff
@@ -82,123 +85,47 @@ func MatchNot(matcher Matcher) Matcher {
 
 // We limit or/and/not to Matchers as composite logic must assume
 // the input path is not modified, so that we can always return it
-//
-// We might want to just remove Matcher implementing Mapper for simplification,
-// and just have a Matches function, then handle chaining them separately.
```
emcfarlane, line 87: Removed, `Matcher`s no longer implement `Mapper`s.

private/pkg/storage/filter.go
```diff
@@ -0,0 +1,108 @@
...
+	"io/fs"
+
+	"github.com/bufbuild/buf/private/pkg/normalpath"
+)
+
+// FilterReadBucket filters the ReadBucket.
+//
+// If the Matchers are empty, the original ReadBucket is returned.
+// If there is more than one Matcher, the Matchers are anded together.
+func FilterReadBucket(readBucket ReadBucket, matchers ...Matcher) ReadBucket {
```
emcfarlane, line 28: Chose `FilterReadBucket` over `MatchReadBucket`, the naming of filter with the filter funcs being called matchers I think is still clear.

bufdev (approved):

Issue #3219, opened 2024-08-07
Buf v1.32.0 and newer walks the entire repository tree when using `--path`


det:
### What's up?

When running `buf lint --path path/to/foo.proto`, it will time out after 2 minutes on a large monorepo. Running the command under `strace -f` reveals that it is walking the entire repository looking at files, including extremely large git-ignored trees like `node_modules` or `bazel-out`. This behavior started in v1.32.0 and still exists in v1.36.0. Previous versions run nearly instantly.

Example failure message: `Failure: timed out after 267189 files: context deadline exceeded`

emcfarlane:
Hi @det , looking into the issue. A workaround for now is to limit the module to a subdir in the repository with the `path` parameter.

doriable:
Tagging a related issue: #2713

emcfarlane:
Fix has been release now in v1.38.0

PR #3230, opened 2024-08-12
Constrain filtered walks on prefix paths

emcfarlane:
Adds a new method MatchPrefix to matchers. This is used on Walk to constrain the walked paths to only included paths declared by matchers. If the prefix is a child of an excluded path the bucket isn't walked.

This is a partial fix for #3219 . On construction of a new module the documentation and license files are joined as a [multi bucket](https://github.com/bufbuild/buf/blob/dd3cb80308df1ce9cb381a344aa2f5e1a7f4606d/private/buf/bufworkspace/workspace_targeting.go#L560-L565). When walking a multi bucket each bucket is walked completely. This can be slow for large repositories, which is then multiplied at least 3 times (once for the module, once for the documentation file, and once for the license file). With the walk constraints applied the underlying bucket will only walk the single file for the documentation and license buckets, as a Walk with a prefix of `.` will be mapped to the delegate bucket as a Walk starting at the filename as that is the only valid child.

### DIFF
--- a/private/pkg/storage/filter.go
+++ b/private/pkg/storage/filter.go
@@ -88,16 +88,22 @@ func (r *filterReadBucketCloser) Walk(ctx context.Context, prefix string, f func
 	if err != nil {
 		return err
 	}
-	return r.delegate.Walk(
-		ctx,
-		prefix,
-		func(objectInfo ObjectInfo) error {
-			if !r.matcher.MatchPath(objectInfo.Path()) {
-				return nil
-			}
-			return f(objectInfo)
-		},
-	)
+	prefixRoots := r.matcher.MatchPrefix(prefix)
+	for _, prefixRoot := range prefixRoots {
+		if err := r.delegate.Walk(
+			ctx,
+			prefixRoot,
+			func(objectInfo ObjectInfo) error {
+				if !r.matcher.MatchPath(objectInfo.Path()) {
+					return nil
+				}
+				return f(objectInfo)
+			},
+		); err != nil {
+			return err
+		}
+	}
+	return err
 }
 
 func (r *filterReadBucketCloser) Close() error {
--- a/private/pkg/storage/matcher.go
+++ b/private/pkg/storage/matcher.go
@@ -26,6 +26,13 @@ type Matcher interface {
 	//
 	// The path is expected to be normalized and validated.
 	MatchPath(string) bool
+	// MatchPrefix returns the matching prefixes.
+	//
+	// The prefix is expected to be normalized and validated.
+	// If the prefix is only partially matched, the valid child prefixes are returned.
+	// If the prefix is fully matched, the prefix is returned.
+	// If the prefix is not matched, nil is returned.
+	MatchPrefix(string) []string
 	isMatcher()
 }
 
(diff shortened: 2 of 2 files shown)


private/pkg/storage/matcher.go
```diff
@@ -45,25 +52,19 @@ func MatchPathBase(base string) Matcher {
 
 // MatchPathEqual returns a Matcher for the path.
 func MatchPathEqual(equalPath string) Matcher {
-	return pathMatcherFunc(func(path string) bool {
-		return path == equalPath
```
emcfarlane, line 49: On windows should this be [`strings.EqualFold`](https://pkg.go.dev/strings#EqualFold)?

private/pkg/storage/filter.go
```diff
@@ -88,16 +88,22 @@ func (r *filterReadBucketCloser) Walk(ctx context.Context, prefix string, f func
...
-		},
-	)
+	prefixRoots := r.matcher.MatchPrefix(prefix)
+	for _, prefixRoot := range prefixRoots {
+		if err := r.delegate.Walk(
+			ctx,
+			prefixRoot,
+			func(objectInfo ObjectInfo) error {
+				if !r.matcher.MatchPath(objectInfo.Path()) {
+					return nil
```
emcfarlane, line 98: Ideally we would be able to skip subdirectories if the prefix is excluded as this Walk will still iterate over excluded child paths.

private/pkg/storage/matcher.go
```diff
@@ -26,6 +26,13 @@ type Matcher interface {
 	//
 	// The path is expected to be normalized and validated.
 	MatchPath(string) bool
+	// MatchPrefix returns the matching prefixes.
+	//
+	// The prefix is expected to be normalized and validated.
+	// If the prefix is only partially matched, the valid child prefixes are returned.
+	// If the prefix is fully matched, the prefix is returned.
+	// If the prefix is not matched, nil is returned.
+	MatchPrefix(string) []string
```
bufdev, line 35: If `MatchPath` returns `bool`, I would expect all `Match.*` functions to return `bool`. A different naming convention will need to  be used.

I don't understand how this interacts with `MatchPath`. Is it valid to have a situation where a value returned from `MatchPrefix` would return `false` for `MatchPatch`? Is this validated in some way, and are the creation of `Matchers` restricted to this package/should they be? Why or why not would a caller outside of this package need `MatchPrefix`? Is the returned slice a copy, or should it not be mutated?

This PR is editing a foundational interfact, so all these questions need to be answered as cleanly as possible, and documented.

emcfarlane: Will fix naming.

`MatchPrefix` returns valid child prefixes. If a bucket is constructed with `MatchPathEqualOrContained("a/b/c")` and the caller walks at `.` we want to ignore all paths outside `a/b/c` so should only walk at the prefix `a/b/c`. In this case `MatchPath(".")` would return false. If we walk `a/b/c/d` we have a parent include so want `MatchPrefix("a/b/c/d")` to return the same path `a/b/c/d` and `MatchPath("a/b/c/d")` would return true.

`Matchers` are restricted to this package with the `isMatchers` method, this is the same as `Mappers` with the `isMappers`. Callers outside wouldn't need either `MatchPath` or `MatchPrefix`, from grepping I don't think any code uses Mappers to use `MapPath` or `UnmapFullPath` outside of this package. Returned slice is a copy, can be mutated.

emcfarlane:
Closing, most of the benefits have already been merged by caching the document and license storage buckets in https://github.com/bufbuild/buf/pull/3232 .

PR #3232, opened 2024-08-12
Cache doc and license storage buckets

emcfarlane:
For large repositories, buckets can be slow to iterate over. Use memory backed buckets to reduce file access for both doc and license files. This can speed up builds by up to 2/3 on large repositories by avoiding iterating over os storage buckets. As an example running build of a single proto file in a large workspace of 96k files takes 1.60s with `v1.36.0`, 1.06s with `v1.31.0` and 0.56s with this change.

This is a simpler approach to #3230 but can be used in conjunction with it.

### DIFF
--- a/private/buf/bufworkspace/workspace_targeting.go
+++ b/private/buf/bufworkspace/workspace_targeting.go
@@ -557,10 +557,18 @@ func getMappedModuleBucketAndModuleTargeting(
 			),
 		)
 	}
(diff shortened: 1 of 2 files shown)


doriable (approved):
I think this is a reasonable, even if we do not end up taking the approach of restricting the walks in the storage layers (#3230). There is just no reason to need to access the doc and/or license files on disk within a single execution context of the CLI. However, will have others chime in on their thoughts.

bufdev (approved):

PR #3233, opened 2024-08-12
Fix problem with input dir not detecting controlling workspace in certain cases

oliversun9:
This fixes the problem where controlling workspace is not detected when the input directory is strictly contained in (not equal to) the workspace directory (where `buf.yaml` v2 is) and but also strictly contains one or more module DirPaths:

With `./buf.yaml`:

```
version: v2
  - path: dir/foo # has dir/foo/foo.proto
  - path: dir/bar # has dir/bar/bar.proto
```

Run `buf build dir -o=-#format=json | jq '.file[].name'`

Output before: (`dir/foo` and `dir/bar` are not built as modules)

```
"bar/bar.proto"
"foo/foo.proto"
```

Output now:

```
"bar.proto"
"foo.proto"
```

The intended behavior has been mentioned [here](https://github.com/bufbuild/buf/blob/b712a4f7589fc8e48c1334f962ed79a817d10419/private/buf/bufworkspace/workspace_targeting.go#L186)


doriable:
This is a tricky change because what it is saying is that if the `input` is a directory that contains one or more modules from a controlling workspace that has been found we respect the controlling workspace and build that. The tricky part is what if we have the following tree and configuration:

```
$ tree .
.
├── buf.yaml
└── foo
    ├── bar
    │   └── bar.proto
    └── baz
        └── baz.proto
```

```
version: v2
  - path: foo/bar
```

```
$ buf build foo
```

If we respect the controlling workspace and only build the input with that context, `foo/baz` would not be included...

doriable (approved):
This makes sense to me, we should add a changelog for this :)

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Fix incorrect builds for directory input when it is a subdirectory in a v2 workspace and contains
```
bufdev, line 5: I can't understand what this means as a user, can you clarify a bit more? cc @doriable

bufdev: Done

bufdev (approved):

PR #3235, opened 2024-08-13
Fix constraints of paths on target images

### LINKED ISSUE #2713: Specifying Folders In `buf generate --path` Arguments Causes Massive Performance Hit
Hi, I've been chatting through this problem on the [buf slack in a thread here](https://bufbuild.slack.com/archives/CRZ680FUH/p1704916350808539). I've been discussing it with @jhump for the most part.

The problem is arising when we're calling `buf generate` with a folder in the `--path` arguments versus calling it with files in the `--path` arguments.

In simple terms:
A) `buf generate --config=buf.yaml --path=/foo/bar/directory` takes ~1 minute to generate `.ts` files for 11 `.proto` files nested in that directory.
B) `buf generate --config=buf.yaml --path=/foo/bar/directory/file1.proto,/foo/bar/directory/file2.proto,...` takes <1 second to generated .ts files for 11 `.proto` files nested in the same directory.

The root folder we're calling `buf generate` from is a _very large monorepo_ with hundreds of thousands of files. _If we do not recursively expand all `.proto` files and inject them into that one `--path` argument (or specify them as 11 separate `--path` arugments), `buf generate` becomes 60+x slower._

If I can provide any more context let me know. I verified this by running `buf generate` a bunch of different times without expanding the files and specifying the folder to make sure it's the folder that's causing it.

### LINKED ISSUE #3219: Buf v1.32.0 and newer walks the entire repository tree when using `--path`

emcfarlane:
This changes the build image for lint and breaking commands to constraint the paths within the module to only targeted paths. Currently, module deps are resolved for the entire module resulting in a full walk of the files. This caused performance issues but could also cause issues from excluded proto files now being included in the build. For example, given a module with files `a/valid.proto` and `b/invalid.proto` at the root of the module. The current behaviour would allow for building the subset as `buf build --path=a` but lint could fail `buf lint --path=a` if `b/invalid.proto` has an unknown import. This is a common issue with `node_modules` or other build folders containing `.proto` files with unresolvable imports.

Fixes #3219, fixes #2713


private/bufpkg/bufmodule/module.go
```diff
@@ -196,32 +196,6 @@ func ModuleToModuleKey(module Module, digestType DigestType) (ModuleKey, error)
...
-//
-// *** THIS IS PROBABLY NOT THE FUNCTION YOU ARE LOOKING FOR. *** You probably want
-// ModuleSetToModuleReadBucketWithOnlyProtoFiles to convert a ModuleSet/Workspace to a
-// ModuleReadBucket. This function is used for cases where we want to create an Image
-// specifically for one Module, such as when we need to associate LintConfig and BreakingConfig
-// on a per-Module basis for buf lint and buf breaking. See bufctl.Controller, which is likely
-// the only place this should be used outside of testing.
-func ModuleToSelfContainedModuleReadBucketWithOnlyProtoFiles(module Module) (ModuleReadBucket, error) {
-	modules := []Module{module}
-	moduleDeps, err := module.ModuleDeps()
```
emcfarlane, line 215: Resolving the module deps requires walking the full tree. To avoid the performance issue on large repos we no longer resolve all dependencies.

private/buf/bufctl/controller.go
```diff
@@ -1029,14 +1029,7 @@ func (c *controller) buildTargetImageWithConfigs(
...
 		}
-		module := moduleSet.GetModuleForOpaqueID(opaqueID)
-		if module == nil {
-			return nil, syserror.Newf("new ModuleSet from WithTargetOpaqueIDs did not have opaqueID %q", opaqueID)
-		}
-		moduleReadBucket, err := bufmodule.ModuleToSelfContainedModuleReadBucketWithOnlyProtoFiles(module)
-		if err != nil {
-			return nil, err
-		}
+		moduleReadBucket := bufmodule.ModuleSetToModuleReadBucketWithOnlyProtoFiles(moduleSet)
```
emcfarlane, line 1032: The `moduleSet` is build from the workspace targeted to the current module by `workspace.WithTargetOpaqueIDs(opaqueID)`.

private/buf/cmd/buf/buf_test.go
```diff
@@ -294,10 +294,13 @@ func TestFail7(t *testing.T) {
...
 		nil,
 		bufctl.ExitCodeFileAnnotation,
-		"", // stdout is empty
+		filepath.FromSlash(`testdata/fail/buf/buf.proto:3:1:Files with package "other" must be within a directory "other" relative to root but were in directory "fail/buf".
+testdata/fail/buf/buf.proto:6:9:Field name "oneTwo" should be lower_snake_case, such as "one_two".`),
 		// This is new behavior we introduced. When setting a config override, we no longer do
 		// a search for the controlling workspace. See bufctl/option.go for additional details.
-		filepath.FromSlash(`Failure: testdata/export/other/proto/unimported.proto: import "another.proto": file does not exist`),
+		// Only the paths specified in the command are considered.
+		// This avoids build failures from other proto files under testdata.
```
emcfarlane, line 302: This test showcases the failure which is no longer hit with the new pathing restriction. The file `testdata/export/other/proto/unimported.proto` imports are no longer resolved.

bufdev: This feels strictly wrong - if `buf build` would not pass, `buf lint` should not pass either. We probably should talk about this live.

doriable: `buf build` does pass for this case, `buf lint` did not pass before Ed's change because of the full imports check. We can chat about it live, for sure. It might be worth clarifying in the test comment that `buf build --path testdata/fail/buf/buf.proto` does pass.

emcfarlane: Yep will clarify, build failure is on linting not running build. `buf build` passes.

private/buf/bufctl/controller.go
bufdev: Unless I'm missing something, this is incorrect. I do not want to build the entire `ModuleSet`, I only want to build the `Module` and its `ModuleDeps`. These ware not necessarily the same thing. A `ModuleSet` may have `Modules` A, B, C, D, E, but `Module` A may only depend on B, C, and I only want to send `A, B, C` to `BuildImage`.

bufdev: This is in the comment you deleted:

```
// This function is used for cases where we want to create an Image
// specifically for one Module, such as when we need to associate LintConfig and BreakingConfig
// on a per-Module basis for buf lint and buf breaking. See bufctl.Controller, which is likely
// the only place this should be used outside of testing.
```

bufdev: If I am missing something (you hinted above with `workspace.WithTargetOpaqueIDs`), then this section needs a few paragraphs of code comments to explain why this is correct.

emcfarlane: We loop over each module set above `for _, module := range modules {...}` and then run `workspace.WithTargetOpaqueIDs(module.OpaqueID())` which would then take module `A` and only include it's deps `B` and `C`. So for multiple modules I believe from the comment is still kept.

emcfarlane: My comment is wrong, this will include more modules.

bufdev: OK, then my assumption is this needs to be reworked, correct?

emcfarlane: The `moduleReadBucket` may contain more modules then is needed by build image. This has changed where before the moduleSet was filtered to contain only the target and its' dependencies. However, the image must filter the `moduleReadBucket` to only target paths. So the final image will be the same as before.

The removed comment behaviour is still maintained, I've moved it to this function. An image is built for each module to associated the LintConfig and BreakingConfig on a per-module basis.

bufdev (approved):

PR #3238, opened 2024-08-15
Ensure that `buf dep update` does not create an empty buf.lock when there are no new deps or existing deps to update

doriable:
`buf dep update` currently goes through the entire process
of updating deps, even if there are no new deps found, and no existing deps
in a `buf.lock` are found.

This PR returns early when there are no new configured deps and
no existing deps, to ensure that we do not create an empty `buf.lock`
when one may not have existed in the first place and it prints a warning
for the user.


private/buf/cmd/buf/command/dep/depupdate/depupdate.go
```diff
@@ -128,6 +128,16 @@ func run(
 	if err != nil {
 		return err
 	}
+	if configuredDepModuleKeys == nil && existingDepModuleKeys == nil {
+		// No new configured deps were found, and no existing buf.lock deps were found, so there
+		// is nothing to update, we can return here.
+		// This ensures we do not create an empty buf.lock when one did not exist in the first
+		// place and we do not need to go through the entire operation of updating non-existent
+		// deps and building the image for tamper-proofing.
+		logger.Sugar().Warnf("No configured dependencies were found to update in %q.", dirPath)
```
stefanvanburen, line 137: I might say:

- no buf.yaml: `No buf.yaml was found in %q, no dependencies to update.`
- buf.yaml with no deps: `No configured dependencies were found to update in %q/buf.yaml.`

Maybe? You all are more familiar with the CLI's error messages than me :)

stefanvanburen (approved):
makes sense to me, just spitballing error messages

private/buf/cmd/buf/command/dep/depupdate/depupdate.go
stefanvanburen: (and it's very probable I'm missing context here, so feel free to discard these suggestions!)

doriable: So I thought about the case where no `buf.yaml` was found, was originally going to surface that error, and then handle it here. But if there was an existing `buf.lock`, for example, and then someone deleted their `buf.yaml` or emptied out the `deps` key in `buf.yaml`, you want to handle both of those cases and interact with `buf.lock`.

This current logic basically allows that to continue to go through, and so the error message does not pertain specifically to the existence of `buf.yaml`, but more the fact that there are no new or existing configured deps.

bufdev (approved):

PR #3239, opened 2024-08-15
Parse docker image digest from Status if Aux not available

mfridman:
This PR adds logic to retrieve the docker image digest from the status field following a successful push. There are some cases where the `Aux` field is not used, and instead, we have to parse it out of the `Status` field.

I've mentioned it in the comment, but adding it here for brevity, the string we're parsing from is:

https://github.com/moby/moby/blob/1c282d1f1b90ff188a1b46f48548ac3151ca2ddf/daemon/containerd/image_push.go#L126-L133

For a bit more context, I've opened an issue upstream https://github.com/moby/moby/issues/48331, but afaics there isn't a better way.

Without this change, I was getting failures pushing a plugin against images built with [containerd image store](https://docs.docker.com/desktop/containerd/) feature enabled. The plugin was pushed, but we failed to parse out the final image digest.

> Failure: failed to determine image digest after push

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add `STABLE_PACKAGE_NO_IMPORT_UNSTABLE` lint rule which disallows files from stable packages
   to import files from unstable packages.
+- Fix plugin push failures when pushing an image built with containerd image store.
 
 ## [v1.36.0] - 2024-08-06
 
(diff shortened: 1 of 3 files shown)


pkwarren (approved):
Should we add an entry to the changelog?

emcfarlane (approved):

PR #3243, opened 2024-08-16
[WIP] Port lint rules to bufcheck package

oliversun9:
Still in the process of checking my own work, but open this draft PR so I can start leaving PR comments


private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -190,6 +197,1022 @@ func handleLintDirectorySamePackage(
...
+	) {
+		responseWriter.AddProtosourceAnnotation(
+			location,
+			nil,
+			format,
+			args...,
+		)
+	}
+	// TODO: make buflintvalidate.CheckMessage's signature the same as this function,
+	// instead of accepting a legacy addFunc as an argument. Same comment for CheckField.
```
oliversun9, line 937: Wrapping the response writer in an addFunc for now, so that we can be certain that the `buflintvalidate` package isn't changed at all, with `diff -r  private/buf/bufcheck/internal/bufcheckserver/internal/buflintvalidate  private/bufpkg/bufcheck/buflint/internal/buflintvalidate` showing no diff.

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverutil/bufcheckserverutil.go
```diff
@@ -78,6 +78,22 @@ func NewRuleHandler(
 	)
 }
 
+// NewMultiHandler returns a new check.RuleHandler combined from multiple handlers.
```
oliversun9, line 81: This is the old [combine function](https://github.com/bufbuild/buf/blob/c11801c352ea5c693faa3b8f43ec886e1339d765/private/bufpkg/bufcheck/buflint/internal/buflintcheck/util.go#L335)

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -190,6 +197,1022 @@ func handleLintDirectorySamePackage(
...
+func handleLintEnumFirstValueZero(
+	responseWriter bufcheckserverutil.ResponseWriter,
+	_ bufcheckserverutil.Request,
+	enum bufprotosource.Enum,
+) error {
+	if values := enum.Values(); len(values) > 0 {
+		if firstEnumValue := values[0]; firstEnumValue.Number() != 0 {
+			// proto3 compilation references the number
+			responseWriter.AddProtosourceAnnotation(
+				firstEnumValue.NumberLocation(),
```
oliversun9, line 253: The [original implementation passes extra ignore locations](https://github.com/bufbuild/buf/blob/c11801c352ea5c693faa3b8f43ec886e1339d765/private/bufpkg/bufcheck/buflint/internal/buflintcheck/buflintcheck.go#L192), but here it does not because it relies on the caller/CLI to handle extra ignore locations correctly. 

Other rules that do this are, `EnumValue*`, `EnumZeroValueSuffix`, `Field*`, `OneofLowerSnakeCase`, `Protovalidate`, `RPC*`.

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -190,6 +197,1022 @@ func handleLintDirectorySamePackage(
...
+					)
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// HandleLintPackageSameDirectory is a handle function.
+var HandleLintPackageSameDirectory = bufcheckserverutil.NewLintPackageToFilesRuleHandler(handleLintPackageSameDirectory)
```
oliversun9, line 686: The original function calls [`newPackageToFilesCheckFunc`](https://github.com/bufbuild/buf/blob/4f697079bebbe88dc6b780bf8773f9046eaa90d8/private/bufpkg/bufcheck/buflint/internal/buflintcheck/util.go#L147), which includes imports, but `NewLintPackageToFilesRuleHandler` does not include import. So this check's behavior has deviated. Or we could add a `NewLintPackageToFilesIncludeImportRuleHandler`.

Same comment for directory to files.

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
oliversun9: On second thought, what if the location here is `nil`? Then the caller/CLI will not be able to figure out the associated paths.

Actually, the statement ^ is not even correct. If the location is `nil`, currently `AddProtosourceAnnotation` would just panic. Just thinking out loud here.

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -178,18 +186,1073 @@ func handleLintDirectorySamePackage(
...
+				)
+			} else {
+				message = fmt.Sprintf(
+					"Files in package %q have multiple values %q for option %q and all values must be equal.",
+					pkg,
+					strings.Join(optionValues, ","),
+					name,
+				)
+			}
+			// TODO: investigate the case where noOptionValue is false but fileOptionLocation is nil.
```
oliversun9, line 908: Yes this is strange. Perhaps the handlers should not assume these locations are not `nil`.

private/buf/bufcheck/internal/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -178,18 +186,1073 @@ func handleLintDirectorySamePackage(
...
+	pkgFiles []bufprotosource.File,
+) error {
+	dirMap := make(map[string]struct{})
+	for _, file := range pkgFiles {
+		dirMap[normalpath.Dir(file.Path())] = struct{}{}
+	}
+	if len(dirMap) > 1 {
+		dirs := slicesext.MapKeysToSortedSlice(dirMap)
+		for _, file := range pkgFiles {
+			if packageLocation := file.PackageLocation(); packageLocation != nil {
```
bufdev, line 721: Overall thing: The way I would handle this is have:

```go
var sourcePath protoreflect.SourcePath
if packageLocation := file.PackageLocation(); packageLocation != nil {
  sourcePath = packageLocation.SourcePath()
}
responseWriter.AddAnnotation(
  check.WithFileName(file.Path()),
  check.WithSourcePath(sourcePath),
  check.WithMessagef(
    "Multiple directories %q contain files with package %q.",
    strings.Join(dirs, ","),
    pkg,
  ),
)
```

This means you aren't duplicating as much. This applies across this whole file.

PR #3260, opened 2024-08-21
Add includes to v2 buf.yaml module config and allow duplicate module paths in v2 buf.yaml

oliversun9:
This PR
- adds `includes` key to module configurations in v2 `buf.yaml`, accepting a list of directories, relative to workspace root
- allows multiple module configurations in the same v2 `buf.yaml` to have the same directory path
  - this is done by updating `bucketID` construction to account for modules with duplicate paths

The change to bucketID affects some user facing errors, for example:
```
Failure: foo/bar/baz/v1/baz.proto is contained in multiple modules: proto//(the 2nd duplicate), proto//(the 3rd duplicate)
```

and some command outputs:
```
digraph {

  "proto"
  "proto//(the 2nd duplicate)"
  "proto//(the 3rd duplicate)"

}
```

Note: a module's `bucketID` is only surfaced when it does not have a full name (e.g. `buf.build/foo/bar`) specified.


private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -201,10 +202,40 @@ func v2WorkspaceTargeting(
 	moduleDirPaths := make([]string, 0, len(bufYAMLFile.ModuleConfigs()))
 	bucketIDToModuleConfig := make(map[string]bufconfig.ModuleConfig)
 	moduleBucketsAndTargeting := make([]*moduleBucketAndModuleTargeting, 0, len(bufYAMLFile.ModuleConfigs()))
+	// With in a v2 bufYAMLFile, multiple module configs may have the same DirPath. To make sure each module
+	// has a unique BucketID, we cannot use their DirPaths as BucketIDs directly, but instead we append
+	// an index (1-indexed) to the path as the bucketID to distinguish modules at the same DirPath. More
+	// specifically, if multiple modules share the same path, the first one's bucketID is just the path,
+	// but starting at the second module, bucketID is "<path>//(the nth duplicate)".
```
doriable, line 209: `<path>//(the nth duplicate)` feels to me like it is optimising for surfacing to the user, which should not be a concern at the `BucketID` layer. The `BucketID` is only surfaced to the user in errors if a module does not have a module name associated with it, I think having a terser ID format is reasonable here (e.g. `<path>//(n)`).

CHANGELOG.md
```diff
@@ -12,6 +12,14 @@
   module's directory.
 - Fix constraints of `--path` flag for lint and breaking rules to avoid resolving all files
   within a module. This change can result in a performance improvement for large workspaces.
+- Add `includes` key to module configurations in v2 `buf.yaml`, accepting a list of directories.
+  * If `includes` is specified, a proto file is considered in the module only if it is in one of the
+    directories specified.
+  * If both `includes` and `excludes` keys are specified for a module, a proto file is considered
+    part of this module if it is contained in any of the include paths and not in any of the exclude
+    paths.
+  * This key does not affect license files or documentation files, e.g. `README.md`.
```
bufdev, line 21: Is this also the case with excludes?

oliversun9: I don't think it's applicable with `excludes`. If a module is at `proto`, its exclude paths must be in the form of `proto/some_dir` and cannot exclude `proto/LICENSE`.

bufdev: OK so are both includes and excludes directories only, or do they also allow specifying `.proto` files?

oliversun9: They are both directories only.

private/bufpkg/bufconfig/module_config.go
```diff
@@ -78,6 +80,19 @@ type ModuleConfig interface {
 	//
 	// This may be nil.
 	ModuleFullName() bufmodule.ModuleFullName
+	// Includes says the module only includes these paths. The paths are relative to DirPath.
```
bufdev, line 83: The assymetry between `Includes` being a `[]string` and `RootToExcludes` including roots and being a `map[string][]string`, with the excludes in the map being root-relative and the includes in the slice being dir-path-relative (is there a difference?) is confusing, and isn't explained in the documentation. As a user/reader of this API, I don't immediately understand, which is the burden of proof going forward. The documentation should address:

- Why includes are relative to a different directory than excludes, if they are.
- Why there is the difference in signature between includes and excludes.

I bet the answer will be some form of "includes are v2 only", which I don't think is a good answer here - these configs are meant to be version-agnostic as much as possible. I think you'll likely end up changing the signature to `RootToIncludes` to match the excludes signature.

CHANGELOG.md
bufdev: I don't understand the changelog then - why are license and doc files being referenced?

oliversun9: I thought specifying `moduleDirPath/subdir` as an `includes:` path might give the impression that `moduleDirPath/LICENSE` is left out (which is not the case with `excludes:`).

private/buf/cmd/buf/command/config/internal/internal.go
```diff
@@ -259,10 +259,20 @@ func lsRun(
...
 	}
-	return nil, fmt.Errorf("no module found for path %q", modulePath)
+	switch len(moduleConfigsFound) {
+	case 0:
+		return nil, fmt.Errorf("no module found for path %q", modulePath)
+	case 1:
+		return moduleConfigsFound[0], nil
+	default:
+		// TODO: add --module-name flag to allow differentiation
+		return nil, fmt.Errorf("multiple modules found for %q", modulePath)
```
oliversun9, line 276: This is the helper used by `ls-[lint/breaking]-rules`, if multiple module matches the `--module-path` flag, should those commands just run once for each matching module?

private/bufpkg/bufconfig/buf_yaml_file.go
```diff
@@ -697,14 +759,23 @@ func writeBufYAMLFile(writer io.Writer, bufYAMLFile BufYAMLFile) error {
 		for _, moduleConfig := range bufYAMLFile.ModuleConfigs() {
 			moduleDirPath := moduleConfig.DirPath()
 			joinDirPath := func(importPath string) string {
-				return filepath.Join(moduleDirPath, importPath)
+				return normalpath.Join(moduleDirPath, importPath)
```
oliversun9, line 762: This is a fix, because other paths, such as those in lint/breaking configs are handled using `normalpath.Join`.

private/bufpkg/bufconfig/module_config.go
doriable: The disconnect here from my perspective is that we are using `map[string][]string` to handle the `v1beta1` keys, `roots` and `excludes` on a single `ModuleConfig`. As the documentation for `RootsToExcludes` states, we should only ever expect a single root for `v1` and `v2` configs:

https://github.com/bufbuild/buf/blob/8063ca9cced372fb8c288121083b107f88540cd3/private/bufpkg/bufconfig/module_config.go#L105

This is because a `Module` should only ever have one root and multiple roots are handled as a `ModuleSet` -- but `v1beta1` breaks that concept, because the `roots` key in the `v1beta1` configuration is more akin to defining a `ModuleSet` rather than a single `Module`. The proper fix to have it so that a single `ModuleConfig` and `Module` has only one root and still supporting `v1beta1` configurations should be that `v1beta1` configs would have one `ModuleConfig` per root, and then at the targeting layer only return a single `ModuleReadBucket` for all of these configs, since they should be built and used as one module (and they should be resolved as a single `OpaqueID`).

However, this feels a little bit out of scope for this PR, and maybe should be a fast follow-up, but I am not sure that we should lean into the fact that at the `ModuleConfig` layer, we have a construct that supports multiple roots for a single `Module` when that is not the intended structure on configurations going forward.

It is also worth noting that `v1beta1` enforces rules on its `roots` that we are breaking in in `v2`:

https://github.com/bufbuild/buf/blob/8063ca9cced372fb8c288121083b107f88540cd3/private/bufpkg/bufconfig/module_config.go#L85-L88

bufdev:
I did a basic QA and this didn't appear to work:

```
# there are no imports across files
$ tree
.
├── ]
├── buf.yaml
├── one
│   └── a
│       ├── a1.proto
│       └── a2.proto
├── three
│   └── c
│       ├── c1.proto
│       └── c2.proto
└── two
    └── b
        ├── bone
        │   ├── bone1.proto
        │   └── bone2.proto
        └── btwo
            ├── btwo1.proto
            └── btwo2.proto

$ cat buf.yaml
version: v2
modules:
  - path: one
  - path: two
    includes:
      - two/b/bone
  - path: two
    includes:
      - two/b/btwo
  - path: two
  - path: three

$ buf build
Failure: b/bone/bone1.proto is contained in multiple modules: two, two//(the 3rd duplicate)
```

Not sure what the error message means either.

doriable:
That configuration wouldn't work, since you have a module defined with `two` as the root with no includes or excludes... 
```
  - path: two
```

The error message needs to be cleaned up, since the bucket ID for that module is using its root and `nth duplicate` (we need to come up with a clearer name/ID to surface, probably -- in the case if that module was named, we would surface the module name).

oliversun9:
> I did a basic QA and this didn't appear to work:
> 
> ```
> # there are no imports across files
> $ tree
> .
> ├── ]
> ├── buf.yaml
> ├── one
> │   └── a
> │       ├── a1.proto
> │       └── a2.proto
> ├── three
> │   └── c
> │       ├── c1.proto
> │       └── c2.proto
> └── two
>     └── b
>         ├── bone
>         │   ├── bone1.proto
>         │   └── bone2.proto
>         └── btwo
>             ├── btwo1.proto
>             └── btwo2.proto
> 
> $ cat buf.yaml
> version: v2
> modules:
>   - path: one
>   - path: two
>     includes:
>       - two/b/bone
>   - path: two
>     includes:
>       - two/b/btwo
>   - path: two
>   - path: three
> 
> $ buf build
> Failure: b/bone/bone1.proto is contained in multiple modules: two, two//(the 3rd duplicate)
> ```
> 
> Not sure what the error message means either.

I think that's the expected behavior, as there are files existing multiple modules.
> two, two//(the 3rd duplicate)

says the the first module named `two` and the module named `two` have `b/bone/bone1.proto` in common.

As for why it's formatted as `two//(the 3rd duplicate)`, instead of something nicer, such as `two(the 3rd duplicate)`, see comment [here](https://github.com/bufbuild/buf/blob/9586b3a0b32a81b65ef61027a9d4b2741db3b1ae/private/buf/bufworkspace/workspace_targeting.go#L862).

CHANGELOG.md
bufdev: This should be removed - it confuses me that we're referencing license and readme at all here.

private/bufpkg/bufconfig/module_config.go
bufdev: I don't understand most of the above to be honest, perhaps need to talk offline. This, however:

> It is also worth noting that v1beta1 enforces rules on its roots that we are breaking in v2:

This is not clear to me from our discussions/this PR/the changelog/etc that it is now intended to allow `foo/bar` and `foo`. Why is this required? I was just about to do QA to make sure this had not changed, and if it needs to, we need to talk about it.

private/buf/bufworkspace/testdata/basic/workspacev2_duplicate_path/buf.yaml
```diff
@@ -0,0 +1,18 @@
+version: v2
+modules:
+  - path: ./proto/shared
+    excludes:
+      - proto/shared/prefix/foo
+  - path: proto//shared1
+    includes:
+      - ./proto//shared1/prefix/x/
```
bufdev, line 8: Why do these test files have spurious `/`s at the end? This would imply to me (without further comments) that the spurious slashes are what we're intending to test here. It seems like this is a false flag.

private/buf/bufworkspace/workspace_provider.go
```diff
@@ -313,15 +313,15 @@ func (w *workspaceProvider) getWorkspaceForBucketAndModuleDirPathsV1Beta1OrV1(
 				return nil, err
 			}
 		}
-		moduleConfig, ok := v1WorkspaceTargeting.bucketIDToModuleConfig[moduleTargeting.moduleDirPath]
```
bufdev, line 316: What changed here and why? This likely needs comments.

private/buf/bufworkspace/workspace_targeting.go
bufdev: This should all be changed courtesy of the other PR, and this comment needs to be updated

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -783,3 +840,31 @@ func checkForOverlap(
 	}
 	return fmt.Errorf("input %q did not contain modules found in workspace %v", inputPath, moduleDirPaths)
 }
+
+func formatBucketIDForDuplicateModulePath(moduleDirPath string, nthDuplicate int) string {
```
bufdev, line 844: Why are we doing all this pretty-print work for bucket IDs?

private/buf/bufworkspace/workspace_targeting.go
```diff
@@ -783,3 +840,31 @@ func checkForOverlap(
...
+		}
+	}
+	// We are appending the 1-based index (i.e. nth) to the moduleDirPath, but we also want to avoid
+	// the case where another module's DirPath is equal to formatBucketIDForDuplicateModulePath(modulePath, n).
+	// In other words, to avoid collision, the return of this function should not be a valid module DirPath.
+	// We use "//" as a separator to make collision impossible: No module's DirPath contains "//"
+	// because it must already have been normalized and must be a relative path, i.e. "foo//bar//" is
+	// normalized to "foo/bar" before being stored as a ModuleConfig's DirPath. For this reason we
+	// choose it over "foo/bar(the 2nd duplicate)", even though it does not look as nice.
+	return fmt.Sprintf("%s//(the %d%s duplicate)", moduleDirPath, nthDuplicate, ordinalSuffix)
```
bufdev, line 869: Sort order will be broken when n >=10

bufdev:
More QA:

```
Failure: "two" had no .proto files
```

This happens when a set of includes/excludes results in no .proto files, however just the path is referencered. We'll need to probably use the description from the chained PR.

private/buf/bufworkspace/workspace_targeting.go
oliversun9: This is updated in [499a77b](https://github.com/bufbuild/buf/pull/3260/commits/499a77b62faf79786243dd77a25085155bb57164)

private/buf/bufworkspace/workspace_targeting.go
oliversun9: With the PR adding `description()`, this no longer needs to be pretty-printed, but before that PR this is meant to be surfaced in error messages.

No longer pretty-formatting, fixed in [499a77b](https://github.com/bufbuild/buf/pull/3260/commits/499a77b62faf79786243dd77a25085155bb57164)

private/buf/bufworkspace/workspace_provider.go
oliversun9: Added comments in [b2f2a40](https://github.com/bufbuild/buf/pull/3260/commits/b2f2a405372c228ec8b7a6e063d8740bb888fd9b)

private/buf/bufworkspace/testdata/basic/workspacev2_duplicate_path/buf.yaml
oliversun9: Removed in [9aaeb3b](https://github.com/bufbuild/buf/pull/3260/commits/9aaeb3be49cc95eae929ff8eaa318b94aeeb8a12) and [613b27b](https://github.com/bufbuild/buf/pull/3260/commits/613b27b6b3cd5da3a0a26c4fe45b1564e6c4be5c)

private/buf/bufworkspace/workspace_targeting.go
oliversun9: Formatting bucketID is updated in [499a77b](https://github.com/bufbuild/buf/pull/3260/commits/499a77b62faf79786243dd77a25085155bb57164), but the problem still exists. I think that's fine because the format of bucketID should not be relied upon.

doriable (approved):

bufdev (approved):

CHANGELOG.md
```diff
@@ -16,6 +16,13 @@
   module's directory.
 - Fix constraints of `--path` flag for lint and breaking rules to avoid resolving all files
   within a module. This change can result in a performance improvement for large workspaces.
+- Add `includes` key to module configurations in v2 `buf.yaml`, accepting a list of directories.
+  * If `includes` is specified, a proto file is considered in the module only if it is in one of the
+    directories specified.
+  * If both `includes` and `excludes` keys are specified for a module, a proto file is considered
+    part of this module if it is contained in any of the include paths and not in any of the exclude
+    paths.
+- Allow multiple module configurations in the same v2 `buf.yaml` to have the same directory path.
```
emcfarlane, line 25: Looks like the CHANGELOG needs to be bumped to the above section.

oliversun9: Good catch!

private/buf/cmd/buf/command/config/internal/internal.go
doriable: I think this is fine -- we should do a follow-up to add the `--module-name` flag for this.

bufdev (approved):

PR #3263, opened 2024-08-23
Add mapping from module opaque ID to display name for better error messages

oliversun9:
This PR adds a new option `LocalModuleWithPrettyPrintName` for `AddLocalModule` so that when constructing a v2 module set, a more descriptive name can be passed in. The values of this option for local modules are then gathered in a map, mapping an opaqueID to a pretty print name. This map is stored in the ModuleSet and can be accessed through a private function within the `bufmodule` package.


bufdev:
@oliversun9 I made a bunch of changes in https://github.com/bufbuild/buf/pull/3263/commits/3d5a5de4629de8506377df8452e9efccf10ae419, check them out, LGTM

bufdev (approved):

private/buf/bufworkspace/workspace_provider.go
```diff
@@ -451,3 +466,38 @@ func (w *workspaceProvider) getWorkspaceForBucketModuleSet(
...
+
+// This formats a module name based on its module config entry in the v2 buf.yaml:
+// `path: foo, includes: ["foo/v1, "foo/v2"], excludes: "foo/v1/internal"`.
+func getLocalModuleDescription(bucketID string, moduleConfig bufconfig.ModuleConfig) string {
+	// We use bucketID instead of moduleConfig.DirPath() as the latter will always
+	// be "." for v1 Modules.
+	//
+	// Specifying this as "path" is somewhat breaking the API contract here, however
+	// we know we're only doing this for local modules.
+	description := fmt.Sprintf("path: %q", bucketID)
```
bufdev, line 478: I think this assumption is immediately broken by the parent PR

doriable: Yeah, this is -- it should be `moduleConfig.DirPath()` here.

doriable (approved):
Approving this, but added to the comment around the description.

private/buf/bufworkspace/workspace_provider.go
oliversun9: I think it should be `bucketID` for v1/v1beta1 and `moduleConfig.DirPath()` for v2

doriable: Ah I see, yes because in the case of `v1`/`v1beta1` we would want to use the `bucketID` since it would be relative to the controlling workspace... I think we can merge this, and in the parent PR, we would add a check here for the `bucketID` -- if it has an index (duplicated root `bucketID`) we would use the `moduleConfig.DirPath()`.

bufdev: I don't think parsing the format of `bucketID` is a good idea, that is the definition of an implementation detail at this level. I made this slightly better, still relying on impl details, but not quite as much, and added some comments.

Issue #3265, opened 2024-08-26
HEAD~1 is no longer treated as a valid ref


neoeinstein:
### GitHub repository with your minimal reproducible example

https://github.com/bufbuild/buf/

### Commands

```shell
buf breaking <URL>#ref=HEAD~1
```

### Output

```shell
fatal: invalid refspec 'HEAD~1'
```

### Expected Output

No failures. This worked correctly and as expected in 1.34. Appears related to #3197 and #3215, but with a different appearance.

### Anything else?

Is there an alternate formulation we should be using to check against the prior commit on a branch?

emcfarlane:
This syntax should work. Fixing. An alternative for now is to specify the branch `buf breaking --against '.git#branch=main,ref=HEAD~'`.

emcfarlane:
The fix has been released in [v1.39.0](https://github.com/bufbuild/buf/releases/tag/v1.39.0).

PR #3198, opened 2024-07-30
Fix git checkout for annotated tags

emcfarlane:
Remove refspec annotation for tags. Annotated tags cannot be mapped to a local ref.

Fixes #3197

PR #3266, opened 2024-08-26
Fix git checkout for relative HEAD refs

emcfarlane:
Annotated tag [fix](https://github.com/bufbuild/buf/pull/3198) caused a regression on `ref=HEAD~` and similar git refs. We need to be more lenient with error handling to try the fallback checkout, before finally resolving the checkout. 

Fixes https://github.com/bufbuild/buf/issues/3265

### DIFF
--- a/private/pkg/git/cloner.go
+++ b/private/pkg/git/cloner.go
@@ -142,6 +142,7 @@ func (c *cloner) CloneToBucket(
 	// First, try to fetch the fetchRef directly. If the ref is not found, we
 	// will try to fetch the fallback ref with a depth to allow resolving partial
 	// refs locally. If the fetch fails, we will return an error.
+	var usedFallback bool
 	fetchRef, fallbackRef, checkoutRef := getRefspecsForName(options.Name)
 	buffer.Reset()
 	if err := c.runner.Run(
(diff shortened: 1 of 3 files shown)


private/pkg/git/cloner.go
```diff
@@ -354,7 +358,7 @@ func getRefspecsForName(gitName Name) (fetchRef string, fallbackRef string, chec
 	} else if checkout != "" && checkout != "HEAD" {
 		// If a checkout ref is specified, we fetch the ref directly.
 		// We fallback to fetching the HEAD to resolve partial refs.
-		return checkout, "HEAD", ""
+		return checkout, "HEAD", checkout
```
emcfarlane, line 361: This was invalid before, we should never have a fallback ref without a checkout.

private/pkg/git/cloner.go
```diff
@@ -160,10 +161,11 @@ func (c *cloner) CloneToBucket(
 		command.RunWithDir(baseDir.AbsPath()),
 	); err != nil {
 		// If the ref fetch failed, without a fallback, return the error.
-		if fallbackRef == "" || !strings.Contains(buffer.String(), "couldn't find remote ref") {
+		if fallbackRef == "" {
```
emcfarlane, line 164: This error check works for branches but fails on `HEAD~`. I've removed the error check and now always try the fallback if available.

bufdev (approved):

bufdev:
Needs changelog

bufdev (approved):

PR #3277, opened 2024-08-29
Add tests for lint and breaking plugins

doriable:
This PR adds test cases for lint and breaking plugins:
- Add integration tests to `buf_test.go` for `ls-{lint, breaking}-rules`, `buf lint`, and `buf breaking` with plugins
- Add integration test to `buf_test.go` for a plugin that panics
- Adds tests to `lint_test.go` and `breaking_test.go` with plugins
- Adds tests to `multi_client_test.go` to check for duplicated rules and categories between custom and built-in plugins


.dockerignore
```diff
@@ -14,6 +14,12 @@ private/buf/cmd/buf/command/alpha/protoc/internal/protoc-gen-insertion-point-rec
 private/buf/cmd/buf/command/alpha/protoc/internal/protoc-gen-insertion-point-writer/protoc-gen-insertion-point-writer
 private/buf/cmd/buf/command/alpha/protoc/test.txt
 private/buf/cmd/buf/command/generate/internal/protoc-gen-top-level-type-names-yaml/protoc-gen-top-level-type-names-yaml
+private/buf/cmd/buf/internal/buf-plugin-panic/cmd/buf-plugin-panic/buf-plugin-panic
```
bufdev, line 17: - Some of the plugins are used by `bufcheck`, which is a dependency of `private/buf/cmd/buf`
- I don't see why `duplicate-plugins` should be grouped in their own subdirectory while the othders shouldnt
- The `Specs` in each test plugin aren't actually used outside of the `main` (I think you patterned this off of the examples, which I just changed)

I would:

- Move these to `private/bufpkg/bufcheck/internal/cmd/buf-plugin-.*`
- Flatten the duplicate plugins to the same level as the others
- Make them each just be a `main` package, ie instead of doing `internal/buf-plugin-panic/cmd/buf-plugin-panic`, just do `internal/cmd/buf-plugin-panic`

private/buf/cmd/buf/buf_test.go
```diff
@@ -612,6 +612,207 @@ func TestCheckLsLintRulesFromConfig(t *testing.T) {
 		"--config",
 		filepath.Join("testdata", "small_list_rules", "buf.yaml"),
 	)
+	// defaults only, built-ins and plugins.
+	testRunStdout(
+		t,
+		nil,
+		0,
+		`
+ID                                CATEGORIES                DEFAULT  PURPOSE
```
bufdev, line 621: Just a general note: these tests are going to be an absolute bear to maintain. The existing `ls-lint-rules` tests across three config versions is OK, and has saved me a few times, but having ~10 of these is going to be tough. I don't know if there's a time-efficient way to refactor, perhaps with `contains`-type tests, to test just what you want. Also `--format-json` flag might be helpful. But perhaps not worth it.

private/buf/cmd/buf/buf_test.go
```diff
@@ -2557,6 +3049,218 @@ func TestLintWithPaths(t *testing.T) {
...
 
+func TestLintWithPlugins(t *testing.T) {
+	t.Parallel()
+	// defaults only, comment ignores on.
+	testRunStdout(
+		t,
+		nil,
+		bufctl.ExitCodeFileAnnotation,
+		filepath.FromSlash(`
+testdata/check_plugins/current/proto/api/v1/service.proto:11:1:Service name "api.v1.FooServiceMock" has banned suffix "Mock". (buf-plugin-suffix)
```
bufdev, line 3060: Same thing here - I know we already have a bunch of these already though, and don't have any good options. Perhaps something like `--error-format=json` and then parse the JSON and make all the messages empty before comparison. That might be a lot of work, and we just want to get these across.

.dockerignore
doriable: Plugins have been moved and collapsed accordingly, and did a merge with `bufplugin` to pick-up latest updates.

private/buf/cmd/buf/buf_test.go
doriable: Yeah... so @oliversun9 and I had a chat this morning about how to refactor some of these tests, because maintenance is definitely a concern. We like the idea of parsing the `json` output for the `ls-{lint, breaking}-rules` tests, but these lint and breaking one, we couldn't really think of a much nicer solution, since you need to structure the expected output somehow. We are proposing keeping these as-is for now unless we think of something really nice for them.

private/buf/cmd/buf/buf_test.go
oliversun9: This is addressed in [84df89c](https://github.com/bufbuild/buf/pull/3277/commits/84df89c5ffa7e2ab3906dcf406eddb4c46b986f9), making the new tests json.

Note: there are 9 tests from the past testing the table output.

bufdev (approved):

PR #3279, opened 2024-08-30
Add custom lint and breaking change plugins

bufdev:


go.mod
```diff
@@ -1,15 +1,15 @@
 module github.com/bufbuild/buf
 
-go 1.21.0
-
-toolchain go1.23.0
+go 1.22
```
bufdev, line 3: We need to figure out if this is OK

rodaine: This should be fine if we're requiring Go 1.22+:

`go`: minimum version required for the language features in the module (that are not behind a build flag)
`toolchain`: recommended stdlib and compiler to use when building the module (ignored transitively), must be greater than or equal to the `go` directive (and elided if equal)

It's dense, but the [docs](https://go.dev/doc/toolchain) on toolchains are helpful.

doriable (approved):

NathanBaulch:
Starting in buf v1.40 my generated .pb.go file headers changed from:
```
// 	protoc-gen-go v1.34.2
```
to
```
// 	protoc-gen-go v1.34.2-devel
```
It looks like this PR upgraded `google.golang.org/protobuf` to the unreleased development version `v1.34.3-0.20240816073751-94ecbc261689`. Was this intentional?

bufdev:
You've commented this in three separate places - as explained in the issue, this has no effect on your generated code. Yes, this was intentional.

go.mod
```diff
@@ -1,16 +1,18 @@
 module github.com/bufbuild/buf
 
-go 1.21.0
+go 1.22.0
```
jhump, line 3: I thought we needed to support the latest three versions of Go, not just latest two, for customers that build from source? Does this change here mean that we have changed our policy?

I actually noticed the other week that the Google Cloud Go client libraries are also changing to only support two latest versions, [but not starting until January](https://github.com/googleapis/google-cloud-go?tab=readme-ov-file#go-versions-supported). Maybe we are following suit?

PR #3286, opened 2024-09-03
Chore fix make upgrade go version

emcfarlane:
Routine make upgrades are failing. See https://github.com/bufbuild/buf/actions/runs/10665104712/job/29621728404 . Looks related to the golangci-lint go version. Force go version to be 1.23, avoid caching of deps to avoid poising caches on upgrades.

### DIFF
--- a/.github/workflows/make-upgrade.yaml
+++ b/.github/workflows/make-upgrade.yaml
@@ -19,6 +19,11 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ steps.app_token.outputs.token }}
+      - name: setup-go
+        uses: actions/setup-go@v5
+        with:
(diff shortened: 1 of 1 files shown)


bufdev:
I think this might be the opposite of the problem - I think the problem is that its timing out BECAUSE theres no cache, and golangci-lint takes forever in that case. Are you sure this will resolve it, or do you just want to try it and see if it does? Either is fine.

emcfarlane:
Locally I had issues with the linter being on the incorrect version when upgrading to go1.23. It would get stuck throttling away at 100% cpu. Don't think the cache is an issue as I would have thought we'd have seen more failures, but could be wrong that might be a go1.23 issue too. Happy to merge this and try?

bufdev (approved):

PR #3288, opened 2024-09-03
Add BUF_BETA_PLUGINS_ENABLED env var gating custom lint and breaking plugins

oliversun9:
This adds `BUF_BETA_PLUGINS_ENABLED` and `buf lint/breaking` would fail is it's not set and there are plugins specified in `buf.yaml`.

```
$ buf lint
Failure: there are plugins set in the buf.yaml, however this feature is in beta and not yet officially released. In order to use plugins, set the environment variable BUF_BETA_PLUGINS_ENABLED=1

$ echo $? 
1

"BUF_BETA_PLUGINS_ENABLED" is not set
$ BUF_BETA_PLUGINS_ENABLED=1 buf lint

echo $? 
0
```

The lint error report is beyond this PR and will be fixed in https://github.com/bufbuild/buf/pull/3289


private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -219,6 +219,11 @@ func run(
 		if flags.ExcludeImports {
 			breakingOptions = append(breakingOptions, bufcheck.BreakingWithExcludeImports())
 		}
+		if len(imageWithConfig.PluginConfigs()) > 0 {
+			if err := bufcli.CheckBetaPluginEnabled(container); err != nil {
+				return fmt.Errorf("plugin configs are set but custom breaking plugins are not enabled: %w", err)
+			}
```
doriable, line 225: We may need to surface a more detailed error message, something like:

```
There are plugins set for breaking change detection, however this feature is in beta and not yet officially released. In order to use plugins, set the environment variable BUF_BETA_PLUGINS_ENABLED=1.
```

In which case, you may need to change the way the env var is checked (the function returns a `bool` and then you construct a detailed error here).

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -219,6 +219,11 @@ func run(
 		if flags.ExcludeImports {
 			breakingOptions = append(breakingOptions, bufcheck.BreakingWithExcludeImports())
 		}
+		if len(imageWithConfig.PluginConfigs()) > 0 {
+			if err := bufcli.CheckBetaPluginEnabled(container); err != nil {
```
doriable, line 223: There needs to be a `TODO` to remove this check here and in `lint`.

private/buf/cmd/buf/command/lint/lint.go
```diff
@@ -138,6 +138,11 @@ func run(
 		if err != nil {
 			return err
 		}
+		if len(imageWithConfig.PluginConfigs()) > 0 {
+			if err := bufcli.CheckBetaPluginEnabled(container); err != nil {
+				return fmt.Errorf("plugin configs are set but custom lint plugins are not enabled: %w", err)
+			}
+		}
```
doriable, line 145: Similar comments to breaking.

private/buf/cmd/buf/command/breaking/breaking.go
oliversun9: I think [6c614dc](https://github.com/bufbuild/buf/pull/3288/commits/6c614dc6cdc5b10cd21f4b1ab478a034912825df) addresses it, but it takes a slightly different approach

private/buf/cmd/buf/command/breaking/breaking.go
oliversun9: I thought this is fine as there is a TODO at the definition for `CheckBetaPluginEnabled`. Once the function is deleted, we will be reminded to remove it because it won't build otherwise, but now I think it's easier to spot the TODO here than in `bufcli/env.go`, going to change.

oliversun9: addressed in [ba3180b](https://github.com/bufbuild/buf/pull/3288/commits/ba3180b3c61d5e51d7f10d83e8721e736d187997)

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -219,6 +219,12 @@ func run(
 		if flags.ExcludeImports {
 			breakingOptions = append(breakingOptions, bufcheck.BreakingWithExcludeImports())
 		}
+		if len(imageWithConfig.PluginConfigs()) > 0 {
+			// TODO: remove this as part of publicly releasing lint/breaking plugins
+			if err := bufcli.CheckBetaPluginEnabled(container); err != nil {
```
bufdev, line 224: You've only gated two places that call a `bufcheck.Client`, there's 6-7 in this codebase.

private/buf/bufcli/env.go
```diff
@@ -84,3 +86,18 @@ func WarnBetaCommand(_ context.Context, container appext.Container) {
 		container.Logger().Warn("This command is in beta. It is unstable and likely to change. To suppress this warning, set " + betaSuppressWarningsEnvKey + "=1")
 	}
 }
+
+// CheckBetaPluginEnabled checks whether beta lint/breaking plugin is enabled through an env var.
+// This returns nil if it is enabled and an error otherwise.
+//
+// TODO: remove this as part of publicly releasing lint/breaking plugins
+func CheckBetaPluginEnabled(container appext.Container) error {
```
bufdev, line 94: IsPluginsEnabled, per the spec

bufdev: This is also supposed to be a bool. You are returning an error saying that there are plugins set, but this function does not actually check if plugins are set or not.

private/buf/cmd/buf/buf_test.go
```diff
@@ -4379,6 +4426,36 @@ func testModInit(t *testing.T, expectedData string, document bool, name string,
 	require.Equal(t, expectedData, string(data))
 }
 
+// TODO: remove this as part of publicly releasing lint/breaking plugins
+func testRunStdoutWithBetaPluginEnabled(t *testing.T, stdin io.Reader, expectedExitCode int, expectedStdout string, args ...string) {
```
bufdev, line 4430: Don't do this, just enable it for every test

private/bufpkg/bufcheck/breaking_test.go
```diff
@@ -1356,6 +1356,7 @@ func testBreaking(
 		previousImage,
 		bufcheck.BreakingWithExcludeImports(),
 		bufcheck.WithPluginConfigs(workspace.PluginConfigs()...),
+		bufcheck.WithPluginsEnabled(),
```
oliversun9, line 1359: This is needed for tests to pass

private/bufpkg/bufcheck/client.go
```diff
@@ -352,6 +367,14 @@ func (c *client) getMultiClient(
 	return newMultiClient(c.logger, checkClientSpecs), nil
 }
 
+// TODO: remove this as part of publicly releasing lint/breaking plugins
+func validatePluginConfigs(pluginConfigs []bufconfig.PluginConfig, isPluginEnabled bool) error {
+	if len(pluginConfigs) > 0 && !isPluginEnabled {
+		return fmt.Errorf("custom plugins are not publicly supported. For more information, please contact us at https://buf.build/docs/contact")
```
bufdev, line 373: S/publicly/yet/

bufdev (approved):

PR #3307, opened 2024-09-09
Fix for protoc plugin extension reparse issue

### LINKED ISSUE #3306: Unused import detection is broken in `protoc-gen-buf-lint` when the image contains an incompatible version of protovalidate than the one buf was compiled with

https://github.com/jchadwick-buf/protoc-gen-buf-lint-unused-ext-import-issue

### Commands

```shell
protoc -I . --buf-lint_out=. $(find ./test -name '*.proto')
```

### Output

```shell
--buf-lint_out: test/c.proto:5:1:Import "test/a.proto" is unused.
```

### Expected Output

I expect no lint failures.

### Anything else?

This is caused because `protoplugin` uses `proto.Unmarshal`, using the default `proto.UnmarshalOptions`. This causes the `Resolver` to be `protoregistry.GlobalTypes`, resulting in all of the extensions from gencode imported by the Buf CLI code being prematurely resolved before the reparse phase. In this case, it happens due to `protovalidate`.

This is blocking https://github.com/bufbuild/protovalidate/pull/246. When the import tracker for `protoc-gen-buf-lint` tries to mark imports as used, it tries to mark all of the extensions for options as used, but it hits an [unexpected case](https://github.com/bufbuild/buf/blob/a9a86caae8734d20dc2b96700803eb0d109e5251/private/bufpkg/bufimage/import_tracker.go#L37-L41) when it sees the extensions for the shared protovalidate constraint, because it resolves to the fullname `buf.validate.priv.field`, but the PR renames the `priv` package to `shared`.

Fixing this will require some kind of modification to `protoplugin`, but a change that will fix this issue is very likely to change the behavior in other cases, so I think this warrants the creation of a new option in `protoplugin`.

jchadwick-buf:
This is a solution to the bug described in #3306. Paired with protoplugin changes that allow it, this ensures that gencode shipped with buf CLI does not interfere with the reparse process by prematurely resolving extensions to the gencode versions.

(Draft until protoplugin has an appropriate interface we can use upstream. This interim version uses a `replace` directive just to demonstrate that it does indeed fix the problem.)


private/buf/cmd/protoc-gen-buf-breaking/breaking.go
```diff
@@ -43,7 +45,12 @@ const (
 
 // Main is the main.
 func Main() {
-	protoplugin.Main(protoplugin.HandlerFunc(handle))
+	protoplugin.Main(
+		protoplugin.HandlerFunc(handle),
+		protoplugin.WithUnmarshalOptions(proto.UnmarshalOptions{
+			Resolver: (*protoregistry.Types)(nil),
```
bufdev, line 51: I don't understand what this is doing and why - needs extensive code comments

jchadwick-buf: The code has been made clearer, no more typed nil.

private/buf/cmd/protoc-gen-buf-breaking/breaking.go
```diff
@@ -43,7 +44,10 @@ const (
 
 // Main is the main.
 func Main() {
-	protoplugin.Main(protoplugin.HandlerFunc(handle))
+	protoplugin.Main(
+		protoplugin.HandlerFunc(handle),
+		protoplugin.WithExtensionTypeResolver(protoencoding.EmptyResolver),
```
doriable, line 49: I think we want some kind of comment here (and for lint) that explains why we are passing in `EmptyResolver` here, e.g.

```
// Passing in an `EmptyResolver` to protoplugin for unmarshalling instead of
// defaulting to the protoregistry.GlobalTypes. This prevents marking unused extensions
// that are resolved through protoregistry.GlobalTypes as used.
protoplugin.WithExtensionTypeResolver(protoencoding.EmptyResolver)
```

(Feel free the massage the wording of the comment ^^")

jchadwick-buf: Done, I reworded it a little bit and linked the issue as well.

bufdev (approved):

PR #3308, opened 2024-09-09
Remove dep github.com/jhump/protoreflect

emcfarlane:
Refactors the bufimageutil testing to remove the dep on `github.com/jhump/protoreflect` (and indirect on `github.com/golang/protobuf`) . Removes `desc` and `protoprint` libraries for testdata by binary encoding them as `.binpb` files for comparisons. Comparisons still provide human readable diffs through the `go-cmp` lib. Files aren't able to use protojson or prototext due to requiring resolvers for `Any` and supporting unknown fields. The testdata is now opaque but the output diff will be reported and can be updated using the same mechanism as previously: `make bufimageutilupdateexpectations`.

Fixes `make upgrade` (see [CI failures](https://github.com/bufbuild/buf/actions/runs/10771047330/job/29865724166)).


private/bufpkg/bufimage/bufimageutil/bufimageutil_test.go
```diff
@@ -52,114 +48,114 @@ func TestOptions(t *testing.T) {
...
-	runSourceCodeInfoTest(t, "foo.bar", "all.txtar")
+	runSourceCodeInfoTest(t, "foo.bar.Foo", "Foo.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Foo", "Foo+Ext.json")
+	runSourceCodeInfoTest(t, "foo.bar.Foo.NestedFoo", "NestedFoo.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Bar", "Bar.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Baz", "Baz.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Quz", "Quz.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Svc", "Svc.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar.Svc.Do", "Do.json", noExts...)
+	runSourceCodeInfoTest(t, "foo.bar", "all.json")
```
emcfarlane, line 158: These tests were already using json files, renamed to `.json`.

private/bufpkg/bufimage/bufimageutil/bufimageutil_test.go
```diff
@@ -184,7 +180,7 @@ func TestTransitivePublic(t *testing.T) {
 	filteredImage, err := ImageFilteredByTypes(image, "c.Baz")
 	require.NoError(t, err)
 
-	_, err = desc.CreateFileDescriptorsFromSet(bufimage.ImageToFileDescriptorSet(filteredImage))
+	_, err = proto.Marshal(bufimage.ImageToFileDescriptorSet(filteredImage))
```
emcfarlane, line 183: Not sure if this is a suitable replacement for what was being exercised.

bufdev: Use the `protoencoding` package

private/bufpkg/bufimage/bufimageutil/bufimageutil_test.go
```diff
@@ -276,44 +272,28 @@ func runDiffTest(t *testing.T, testdataDir string, typenames []string, expectedF
...
-		)
-	}
-	sort.SliceStable(archive.Files, func(i, j int) bool {
-		return archive.Files[i].Name < archive.Files[j].Name
-	})
-	generated := txtar.Format(archive)
-	checkExpectation(t, ctx, generated, bucket, expectedFile)
+	checkExpectation(
+		t, ctx, fileDescriptorSet, bucket, expectedFile,
+		proto.MarshalOptions{Deterministic: true}.Marshal,
```
bufdev, line 277: Use `protoencoding` wherever possible for marshal/unmarshal

bufdev (commented):
Will defer to @jhump if this is acceptable

jhump:
Oof, this would make it a serious pain when changing the tree-shaking code in bufimageutil. The value of them being text isn't just for test assertion comparisons but also for code reviewers -- when you change any of the logic (including fixes), with a text format, it's trivial to see exactly _how_ your branch changed the output in the PR diff. If we merge this, that will make these golden test files basically useless for code reviews. 

If the desire is simply to get the old API v1 usage of the CLI codebase, you could instead `go get github.com/jhump/protoreflect/v2@bd8d71d0d99e0821a2a8437b1e9baad3c2716596` and change to use the v2 [protoprint](https://pkg.go.dev/github.com/jhump/protoreflect/v2@v2.0.0-20240613195607-6a462d90c19f/protoprint), which uses the newer Protobuf API v2 runtime. (There is no release yet, which is why you'd have to get a particular commit.)

jhump:
> but still not as legible as before.

I suppose this is marginally better, but it's still very low value for a reviewer.

> I do think not relying on the output of the printer is beneficial. 

Why?

emcfarlane:
> Why?

Doesn't require any dependencies. No requirement for stable output from protoprint package. Diff is done on the message level, so if any information was lost through the translation of messages to the protoprint format the current tests wouldn't capture it.

jhump:
> Doesn't require any dependencies.

Why is this important? This is only test code. This does not change the compilation time or size of the `buf` binary. Also, this dependency is quite small compared to the other 40+ deps (and another ~70 indirect), so I'm unclear why there is value in prosecuting this one.

> No requirement for stable output from protoprint package.

It already produces stable output.

> if any information was lost through the translation of messages to the protoprint format the current tests wouldn't capture it.

I don't think this is much of a risk. But this is the only argument that really makes sense to me. And, honestly, just use JSON instead of binary. Your PR summary said this about JSON:

> Files aren't able to use protojson or prototext due to requiring resolvers for Any and supporting unknown fields

But this isn't true. We already handle json format for descriptors -- you just have to parse them 2x. I'm pretty sure we even already have code in the buf CLI for precisely this.

But, having said that, this will still be a huge pain for reviewers to actually see semantic differences because source code info in a JSON descriptor is very hard to decipher...

The tradeoffs to achieve the ostensible benefits seem too great and outweigh any benefit IMO.

emcfarlane:
> But this isn't true. We already handle json format for descriptors -- you just have to parse them 2x. I'm pretty sure we even already have code in the buf CLI for precisely this.

Maybe I'm missing the package that handles this? JSON currently fails on this `e.proto` file [here](https://github.com/bufbuild/buf/blob/1af1c50bc2b662cbb6e48073c217305926d23c21/private/bufpkg/bufimage/bufimageutil/testdata/any/e.proto#L8). Can't resolve the Any type `"type.googleapis.com/blah.blah.Blah"`, returning a dynamic type could work but the value is invalid wire data.

> Why is this important? 

To fix the `make upgrade` failure due the package now being deprecated. We can upgrade to the unreleased package as a fix, but less deps seemed better and avoids any breaks in the future.

bufdev:
FWIW:

- I already put the exception in `.golangci.yml` https://github.com/bufbuild/buf/blob/main/.golangci.yml#L290
- All things considered, I'd love to only have a dep on bufbuild/protocompile + google.golang.org/protobuf and kill the dep on jhump/protoreflect
- However, I don't care that much :-)

If this is going to be a pain, we can close this if you want. I would like to get it in, but it's not a big deal.

jhump:
> Can't resolve the Any type "type.googleapis.com/blah.blah.Blah", returning a dynamic type could work but the value is invalid wire data.

FWIW, you would use a dynamic type of an empty message and ignore unknown fields via `protojson.UnmarshalOptions` for the first pass at parsing. Looks like this is not exposed via a useful helper. It's in the guts of the buf CLI controller, for loading an image from JSON: https://github.com/bufbuild/buf/blob/1af1c50bc2b662cbb6e48073c217305926d23c21/private/buf/bufctl/controller.go#L948-L957
And it also fails to handle unrecognized `Any` values. Doh! It was likely implemented just to handle preserving custom options, which just get dropped on the floor in the first pass of parsing since they aren't recognized/known extensions yet. I can open a PR to fix that code to do that (though it's still not really import-able/re-usable, sadly).

emcfarlane:
@jhump tested moving to protoprint v2 API [here](https://github.com/bufbuild/buf/compare/ed/v2Protoprint). Some differences with the v1 output are causing test failures.

jhump:
> Some differences with the v1 output are causing test failures.

@emcfarlane, that's unfortunate. I think the exception in golangci config is adequate for now then.

> kill the dep on jhump/protoreflect

@bufdev, it definitely provides value for these tests. It's just a test dependency and isn't very big, so I'm a bit curious about the motivation here.

jhump:
> JSON currently fails on this `e.proto` file [here](https://github.com/bufbuild/buf/blob/1af1c50bc2b662cbb6e48073c217305926d23c21/private/bufpkg/bufimage/bufimageutil/testdata/any/e.proto#L8).

Oof, I was wrong. We could _unmarshal_ such an image using a technique like I described. But this won't even _serialize_ to create the JSON because it can't resolve the type. A case where the JSON format for `Any` is total trash 😢

bufdev:
Motivation would just be cleanup - would be nice to have one codepath that does protoreflect-like things.

jhump:
@emcfarlane, I dug into your protoprintv2 branch and pushed a couple of commits.

The changes in expected output are actually correct: while it is acceptable to include the "optional" keyword for optional extensions in proto3, it is not necessary and has no effect. And since it is _disallowed_ in editions, the newer version of protoprint elides it for proto3 as well. So only proto2 syntax files will have the explicit optional keyword (because it is required in proto2).

PR #3316, opened 2024-09-12
Add an LSP server command

mcy:
This PR adds a new command, buf beta lsp, which serves an LSP for Protobuf files.

Most of the code lives under `buflsp`. Currently, the LSP implements:

- Parse and diagnose.
- Format on save.
- Go to definition, inc. go to definition of Buf dependencies.
- Semantic highlighting.
- Symbol tooltips.


private/buf/buflsp/file_manager.go
```diff
@@ -0,0 +1,64 @@
...
+
+// fileManager tracks all files the LSP is currently handling, whether read from disk or opened
+// by the editor.
+type fileManager struct {
+	lsp   *lsp
+	table refcount.Map[protocol.URI, file]
+}
+
+// newFiles creates a new file manager.
+func newFiles(lsp *lsp) *fileManager {
```
bufdev, line 34: `newFileManager`

private/buf/buflsp/file.go
```diff
@@ -0,0 +1,377 @@
...
+	// has acquired b.proto's lock already, and attempts to acquire a lock to a.proto,
+	// again because of a pointer in importMap. This will deadlock, and it will
+	// deadlock in such a way that will be undetectable to the Go scheduler, so the
+	// LSP will hang forever.
+	//
+	// This seems like a contrived scenario, but it can happen if a user creates two
+	// mutually-recursive Protobuf files. Although this is not permitted by Protobuf,
+	// the LSP must handle this invalid state gracefully.
+	//
+	// TODO(mcy): enforce this somehow.
```
doriable, line 66: Is this already addressed?

private/buf/buflsp/buflsp.go
```diff
@@ -0,0 +1,280 @@
...
+
+			retErr = actual(ctx, replier, req)
+		}()
+
+		<-ctx.Done()
+		if ctx.Err() == context.DeadlineExceeded {
+			// Don't return this error; that will kill the whole server!
+			l.logger.Sugar().Errorf("timed out while handling %s; this is likely a bug", req.Method())
+		}
+		return retErr
```
emcfarlane, line 196: This looks like a data race on `retErr`. I'm not sure you need the go routine here as passing down the context should be enough to ensure the work is bounded.

private/buf/buflsp/buflsp.go
```diff
@@ -0,0 +1,280 @@
...
+	"github.com/bufbuild/buf/private/pkg/tracing"
+	"go.lsp.dev/jsonrpc2"
+	"go.lsp.dev/protocol"
+	"go.opentelemetry.io/otel/attribute"
+	"go.uber.org/zap"
+)
+
+const (
+	// TODO: make configurable
+	handlerTimeout = 3 * time.Second
```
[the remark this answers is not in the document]
mcy: This shouldn't be made configurable until it becomes a problem. It's basically a relief-valve for potential latency issues.

private/buf/buflsp/buflsp.go
mcy: No, this is necessary because jsonrpc2 does not parallelize requests by default.

emcfarlane: This go routine doesn't parallelize the processing of requests as it blocks and waits for the completion. The same blocking can be done without the go routine and rely on the context to cancel the work. There still looks like a race condition on retErr as the above return on context cancel doesn't guarantee the write happens before the read.

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,168 @@
...
+// mutex is a sync.Mutex with some extra features.
+//
+// The main feature is reentrancy-checking. Within the LSP, we need to lock-protect many structures,
+// and it is very easy to deadlock if the same request tries to lock something multiple times.
+// To achieve this, Lock() takes a context, which must be modified by withRequestID().
+type mutex struct {
+	lock sync.Mutex
+	// This is the id of the context currently holding the lock.
+	who  atomic.Uint64
+	pool *mutexPool
```
emcfarlane, line 81: Maybe a `sync.Cond` could be used with `Broadcast` to wake all non owners and reschedule the lock and allow checking who owns the current lock:

```go
func (m *mutex) Lock(id uint64) {
    m.lock.Lock()
    for m.owner != 0 {
        if m.owner == id {
            m.lock.Unlock()
            panic("this id locked twice")
        }
        m.cond.Wait()
    }
    m.owner = id
    m.lock.Unlock()
}
func (m *mutex) Unlock(id uint64) {
    m.lock.Lock()
    if m.owner != id {
        m.lock.Unlock()
        panic("invalid")
    }
    m.id = 0
    m.cond.Broadcast()
    m.lock.Unlock()
}
```

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,168 @@
...
+	id := getRequestID(ctx)
+	switch who := mu.who.Load(); {
+	case who == poison:
+		// This lock was locked twice by the same request and should probably never be
+		// unlocked. panic.
+		panic("buflsp/mutex.go: non-reentrant lock locked twice by the same request")
+	case who != id && id > 0:
+		panic("buflsp/mutex.go: lock was locked by one request and unlocked by another")
+	default:
+		mu.who.Store(0)
```
emcfarlane, line 155: `mu.who.Store(0)` must be called with `mu.lock.Lock()` else you can race the `Lock()` function.

bufdev (commented):
Oh my phone but don't forget changelog if we haven't done it

private/buf/buflsp/mutex.go
emcfarlane: Ah, this is guaranteed.

private/buf/buflsp/buflsp.go
mcy: It waits for completion or deadline expiration. Your observation re: relying on the context to cancel the work is only correct if the work is I/O bound, but that is not the case here.

There is no data race; I suggest looking at how context cancelation and synchronization works (writing to a channel is synchronization).

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,168 @@
...
+//
+// The returned unlocker is not thread-safe.
+func (mu *mutex) Lock(ctx context.Context) (unlocker func()) {
+	var unlocked bool
+	unlocker = func() {
+		if unlocked {
+			return
+		}
+		mu.Unlock(ctx)
+		unlocked = true
```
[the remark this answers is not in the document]
mcy: Look for `unlock()` elsewhere in the PR. It is necessary to avoid squirrely control flow.

private/buf/buflsp/mutex.go
mcy: That's not what this is doing. This ensures that if A has locked X, it cannot also lock Y, because an acquire within a critical section is what causes deadlocks in the LSP.

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,168 @@
...
+		// unlocked. panic.
+		fallthrough
+	case who == id && id > 0:
+		// We seem to have tried to lock this lock twice. Panic, and poison the lock.
+		mu.who.Store(poison)
+		panic("buflsp/mutex.go: non-reentrant lock locked twice by the same request")
+	default:
+		// Block until the lock is free.
+		mu.lock.Lock()
+		mu.who.Store(id)
```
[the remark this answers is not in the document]
mcy: Yeah that's just a missing branch.

private/buf/buflsp/mutex.go
emcfarlane: Why is the mutex not shared between X and Y?

private/buf/buflsp/buflsp.go
emcfarlane: It's an unlikely data race, but the context can be canceled by the caller without `context.DeadlineExceeded ` so  it's still racy. CPU bounded can still take ctx and check to close work. Guess for a short lived CPU only this is fine, but most functions should take ctx and check it so it doesn't have to be only I/O bound work.

emcfarlane: I think this could be written with an explicit channel.
```go
errChan := make(chan err, 1)
go func() { errChan <- work(ctx) }()
select {
case err <- errChan:
    return err
case <- ctx.Done():
    return ctx.Err()
}
```

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if mu.who.Load() == id && id > 0 {
+		// We seem to have tried to lock this lock twice. Panic, and poison the lock.
+		mu.who.Store(poison)
+		panic("buflsp/mutex.go: non-reentrant lock locked twice by the same request")
+	}
+
+	if mu.pool != nil {
+		mu.pool.lock.Lock()
+		defer mu.pool.lock.Unlock()
```
emcfarlane, line 120: This now deadlocks on any request waiting for a lock as `pool.Lock()` is taken with `lock.Lock()` meaning all locks are blocked by the single lock waiting as the next locker waits on `pool.Lock()` which is blocked by `lock.Lock()`.

doriable (approved):

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,182 @@
...
+
+	mu.storeWho(0)
+
+	mu.pool.check(id, mu, true)
+	mu.lock.Unlock()
+}
+
+func (mu *mutex) storeWho(id uint64) {
+	for {
+		// This has to be a CAS loop to avoid races with a poisoning p.
```
emcfarlane, line 173: Not really. If you are unable to swap it's confirmed to be poisoned as the lock is always held.
```go
if mu.who.CompareAndSwap(0, id) {
	panic("buflsp/mutex.go: non-reentrant lock locked twice by the same request")
}
```

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,182 @@
...
+func (mu *mutex) Unlock(ctx context.Context) {
+	id := getRequestID(ctx)
+	if mu.who.Load() != id {
+		panic("buflsp/mutex.go: lock was locked by one request and unlocked by another")
+	}
+
+	mu.storeWho(0)
+
+	mu.pool.check(id, mu, true)
+	mu.lock.Unlock()
```
emcfarlane, line 168: The `mu.lock.Unlock()` should be deferred, as this never gets called on `panic` of `storeWho`. So go routines will still deadlock on line `150` as they wait for this Unlock. Although does this matter? Would only matter if using recovers.

bufdev: @mcy can you check over all of your concurrency logic here before we merge?

private/buf/buflsp/file_manager.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/bufbuild/buf/private/pkg/refcount"
+	"go.lsp.dev/protocol"
+)
+
+// fileManager tracks all files the LSP is currently handling, whether read from disk or opened
+// by the editor.
+type fileManager struct {
+	lsp   *lsp
+	table refcount.Map[protocol.URI, file]
```
bufdev, line 30: @doriable can you do a pass on naming here? None of these names match our style

private/buf/buflsp/buflsp.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		return reply(ctx, result, err)
+	}
+}
+
+// connAdapter wraps a connection and logs calls and notifications.
+//
+// By default, the ClientDispatcher does not log the bodies of requests and responses, making
+// for much lower-quality debugging.
```
doriable, line 213: Even though this is pretty small, I think it's reasonable to split this out in a different file, since this is fairly isolated functionality (even though I see the reasoning to keep it side-by-side with `adaptReplier`/`wrapReplier`).

private/buf/buflsp/buflsp.go
```diff
@@ -0,0 +1,269 @@
...
+		if req.Method() != protocol.MethodInitialize && l.initParams.Load() == nil {
+			return replier(ctx, nil, fmt.Errorf("the first call to the server must be the %q method", protocol.MethodInitialize))
+		}
+
+		return actual(ctx, replier, req)
+	}
+}
+
+// adaptReplier wraps a jsonrpc2.Replier, allowing us to inject logging and tracing and so on.
+func (l *lsp) adaptReplier(reply jsonrpc2.Replier, req jsonrpc2.Request) jsonrpc2.Replier {
```
doriable, line 190: I think the preferred naming convention in the code base for wrapping/wrapped types tend to be `wrapReplier` and `wrappedConnection` for the type below.

private/buf/buflsp/file.go
```diff
@@ -0,0 +1,675 @@
...
+	//
+	// This seems like a contrived scenario, but it can happen if a user creates two
+	// mutually-recursive Protobuf files. Although this is not permitted by Protobuf,
+	// the LSP must handle this invalid state gracefully.
+	//
+	// This is enforced by mutex.go.
+	lock mutex
+
+	text    string
+	version int32
```
doriable, line 81: We might want to document what `version` here refers to and how it should be used.
And maybe `%s/text/content`, but I'm low confidence on this one (it makes sense to me either way, since this is the `string` representation of content).

private/buf/buflsp/file_manager.go
```diff
@@ -0,0 +1,66 @@
...
+	pool  mutexPool
```
doriable, line 31: `lsp` seems reasonable to me, but I think `table` should be more explicit, e.g. `fileURIToFile`, and there should be a comment indicating why we are using the `refcount` map rather than built-in map (I know there's a comment already on the `refcount` package about why to use it, but that isn't immediately reflected/translated here).
Similarly, we should be more explicit with `mutexPool` and call it `mutexPool` and add a comment to indicate this is a pool of the mutexes of the files that are managed by the `fileManager` and why it exists (to check for processes attempting to get a lock multiple times).

private/buf/buflsp/mutex.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+	id, ok := ctx.Value(&nextRequestID).(uint64)
+	if !ok {
+		return 0
+	}
+
+	// Make sure we don't return 0. This is the only place where the id is actually
+	// witnessed so doing +1 won't affect anything.
+	return id + 1
+}
```
doriable, line 57: These are both helpers, so they can be at the bottom of the file, even though `withRequestID` is shared by another file. Also, the docs should be `%s/withReentrancy/withRequestID`.

private/buf/buflsp/report.go
```diff
@@ -0,0 +1,83 @@
...
+
+	"github.com/bufbuild/protocompile/linker"
+	"github.com/bufbuild/protocompile/parser"
+	"github.com/bufbuild/protocompile/reporter"
+	"go.lsp.dev/protocol"
+)
+
+// report is a reporter.Reporter that captures diagnostic events as
+// protocol.Diagnostic values.
+type report struct {
```
doriable, line 30: Since this implements `reporter.Reporter`, it should probably be named `reporter`.

private/buf/buflsp/report.go
```diff
@@ -0,0 +1,83 @@
...
+
+		unused[unusedImport.UnusedImport()] = true
+	}
+}
+
+// error2diagnostic converts a protocompile error into a diagnostic.
+//
+// Unfortunately, protocompile's errors are currently too meagre to provide full code
+// spans; that will require a fix in the compiler.
+func newDiagnostic(err reporter.ErrorWithPos, isWarning bool) protocol.Diagnostic {
```
doriable, line 64: Doc update `%s/error2diagnostic/newDiagnostic`.

doriable (commented):
Add some comments around style/readability.

private/buf/buflsp/report.go
doriable: Instead of alias-ing the package, let's leave this as `report` -- it is still a clear name for the implementation, since it is a "report" of the diagnostics, missing syntax, and unused imports.

We should probably rename `syntaxMissing` and `unusedImports` with the same map naming conventions, so in this case `fileNameToSyntaxMissing` and `fileNameToUnusedImports`.

doriable: Chatted in person, `syntaxMissing` is a good name for the set, and we'll rename `unusedImports`.

PR #3317, opened 2024-09-13
Expand buf lint PROTOVALIDATE rule to lint example and predefined rules

oliversun9:
This updates the `PROTOVALIDATE` rule in `buf lint` to provide checks for two new features:

- For the [protovalidate examples feature](https://github.com/bufbuild/protovalidate/pull/239), check:
    - example values satisfy the constraints (builtin + cel)
    - if no constraint is defined on the field, an example value should not be present
- For the [predefined rules feature](https://github.com/bufbuild/protovalidate/pull/246), check:
    - a predefined rule compiles (the expression defined on `buf.validate.*Rules` compiles)
    - example values also satisfy any predefined rules

The `protovalidate` test dependency has also been updated and some previous integration tests have also been updated to reflect this.


private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverhandle/lint.go
```diff
@@ -935,56 +936,90 @@ func handleLintPackageVersionSuffix(
...
-	// NOTE: Oneofs also have protovalidate support, but they only have a "required" field, so nothing to lint.
-	bufcheckserverutil.NewLintMessageRuleHandler(handleLintMessageProtovalidate),
-	bufcheckserverutil.NewLintFieldRuleHandler(handleLintFieldProtovalidate),
+var HandleLintProtovalidate = bufcheckserverutil.NewRuleHandler(
+	func(
+		ctx context.Context,
+		responseWriter bufcheckserverutil.ResponseWriter,
+		request bufcheckserverutil.Request,
+	) error {
+		// TODO: refactor so that add func is no longer needed
```
oliversun9, line 945: This can be done in a separate PR. This PR will maintain the status quo.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -163,16 +168,30 @@ func checkField(
...
 		fieldDescriptor,
 		fieldDescriptor.Cardinality() == protoreflect.Repeated,
+		extensionTypeResolver,
 	)
 }
 
 func checkConstraintsForField(
 	adder *adder,
 	fieldConstraints *validate.FieldConstraints,
+	// TODO: fix wording
```
oliversun9, line 182: Going to fix

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,150 @@
...
+		if err := options.Unmarshal(unknown, reflectMessage.Interface()); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func celTypeForStandardRuleMessageDescriptor(
+	ruleMessageDescriptor protoreflect.MessageDescriptor,
+) *cel.Type {
```
doriable, line 106: @rodaine Just wondering if there is a better way to do this and/or if `protovalidate-go` exposes some mechanism for me to resolve the "type of rule" based on the rule message descriptor (e.g. `StringRules`). The context of this is that, we are parsing predefined rules from our files, and then checking their CEL expressions. And so we need the type for the extension itself (which is easy, we use the `celext` function you and @oliversun9 exported a while ago) and we also need the type for the rule that the extension is extending.

make/buf/all.mk
```diff
@@ -144,6 +144,9 @@ bufgeneratebuflinttestdata:
 	$(BUF_BIN) export \
 		buf.build/bufbuild/protovalidate:$(PROTOVALIDATE_VERSION) \
 		--output private/bufpkg/bufcheck/testdata/lint/protovalidate/vendor/protovalidate
+	$(BUF_BIN) export \
```
bufdev, line 147: Why are we doing this? This appears to be exactly the same thing as in the above directory.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -700,6 +755,284 @@ func checkTimestampRules(adder *adder, timestampRules *validate.TimestampRules)
...
+	fieldConstraints *validate.FieldConstraints,
+	typeRulesMessage protoreflect.Message,
+	containingMessageDescriptor protoreflect.MessageDescriptor,
+	// Not nil only if fieldDescriptor is a synthetic field for map key/value.
+	parentMapFieldDescriptor protoreflect.FieldDescriptor,
+	fieldDescriptor protoreflect.FieldDescriptor,
+	exampleValues []protoreflect.Value,
+	extensionTypeResolver protoencoding.Resolver,
+) error {
+	if err := protoencoding.ReparseExtensions(extensionTypeResolver, typeRulesMessage); err != nil {
```
bufdev, line 770: This feels like a pretty deep level to be calling protoencoding.ReparseExtensions, to do this per field seems unnecessary, can you provide some documentation as to why we're calling this here, and why it can't be called at a higher level?

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func (r *constraintsResolverForTargetField) ResolveFieldConstraints(desc protoreflect.FieldDescriptor) *validate.FieldConstraints {
+	if desc.FullName() != r.targetField.FullName() {
+		return nil
+	}
+	return r.StandardConstraintResolver.ResolveFieldConstraints(desc)
+}
+
+func (r *constraintsResolverForTargetField) ResolveMessageConstraints(desc protoreflect.MessageDescriptor) *validate.MessageConstraints {
```
bufdev, line 42: nit: no need to have `r` name for struct, and no need for `desc` either

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+func (r *constraintsResolverForTargetField) ResolveOneofConstraints(desc protoreflect.OneofDescriptor) *validate.OneofConstraints {
```
bufdev, line 46: same

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,133 @@
...
+	return nil
+}
+
+// This function is copied directly from protovalidate-go, except refactored to use protoencoding
+// for marshalling and unmarshalling. We also added error handling for marshal/unmarshal.
+//
+// This resolves the given extension and returns the constraints for the extension.
+func resolveExt[C proto.Message](
```
bufdev, line 54: other methods called `*Extension`, call this `resolveExtension`

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/util.go
```diff
@@ -0,0 +1,133 @@
...
+	options proto.Message,
+	extType protoreflect.ExtensionType,
+	resolver protoencoding.Resolver,
+) (constraints C, retErr error) {
+	num := extType.TypeDescriptor().Number()
+	var msg proto.Message
```
bufdev, line 60: `message`

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
doriable: Add a comment to clarify this, but basically it's to ensure that predefined rules from imported files are properly resolved.

bufdev: This comment doesn't make sense to me as a reader - I am still confused why I need to call `ReparseExtensions` for every field. Can you make sure this comment would make sense to another reader without context by verifying with someone else what their understanding of it is?

bufdev (approved):

PR #3325, opened 2024-09-16
Fix comments on separators and handling message literals and compound strings in options

### LINKED ISSUE #3322: Buf format deletes inline comments in some cases

https://github.com/anzboi/buf-bugs

### Commands

```shell
buf format --path module/format.proto
```

### Output

```shell
syntax = 'proto3';

package module;

import "google/api/annotations.proto";

service Service {
  rpc Method(Message) returns (Message) {
    option (google.api.http) = {
      additional_bindings: []
      additional_bindings: [] /* This comment is safe */
      additional_bindings: []
      body: "*" // but this one is
    };
  }
}

message Message {}
```

### Expected Output

```shell
syntax = 'proto3';

package module;

import "google/api/annotations.proto";

service Service {
  rpc Method(Message) returns (Message) {
    option (google.api.http) = {
      additional_bindings: [] /* This comment will get deleted */
      additional_bindings: [] /* This comment is safe */
      additional_bindings: [] /* this one is not */
      body: "*" // but this one is
    };
  }
}

message Message {}
```

### Anything else?

This issue seems to occur only under specific circumstances. We noticed it only after empty repeated options with a trailing comma, like `my_option: [], // this comment gets deleted`. If you remove the trailing comma, or the comment is after a different type (string, int), the comment is safe.

We have not tested much further, like what if the repeated field is populated, or after a nested message, or if the option occurs in a message options, field options, service options, file options.

doriable:
This fixes some issues in the formatter for handling trailing comments
on separators (`,` and `;`) in message literals. These comments are added
to the element if the element itself did not have trailing comments. However,
the comments that were added to message and array literal elements are
"left behind" because they are not being handled. This PR adds logic that
handles these comments similar to how comments on separators are handled --
it checks array and message literals for any trailing comments that may have been
added in this manner and adds them to closing brackets, if the closing brackets
did not have comments originally. Tests have been added for these cases.

In the process of testing, we also found an additional bug, where empty message
literals were not being handled as compact messages and compound strings in
options were not being written with their ends in-line for trailing commas.

Fixes #3322


bufdev (approved):

bufdev:
Needs changelog

emcfarlane (approved):

PR #3327, opened 2024-09-17
Run bufcheck WithPlugins option

emcfarlane:
Moves the `bufconfig.PluginConfig` to be provided with the `pluginrpc.Runner` as a `bufcheck.Plugin`. This allows for setting the runner based on the configuration of the plugin. Removes the default runner as this is now provided when invoking a bufcheck check using the new `WithPlugins` client function option that replaces `WithPluginConfigs`. `RunnerProvider` is removed. A helper `NewPluginsForRunner` is provided to map local plugins to a `command.Runner`.

### DIFF
--- a/private/bufpkg/bufcheck/bufcheck.go
+++ b/private/bufpkg/bufcheck/bufcheck.go
@@ -21,6 +21,8 @@ import (
 	"buf.build/go/bufplugin/check"
 	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
+	"github.com/bufbuild/buf/private/pkg/command"
+	"github.com/bufbuild/buf/private/pkg/pluginrpcutil"
 	"github.com/bufbuild/buf/private/pkg/slicesext"
 	"github.com/bufbuild/buf/private/pkg/syserror"
 	"github.com/bufbuild/buf/private/pkg/tracing"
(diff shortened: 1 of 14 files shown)


private/bufpkg/bufcheck/bufcheck.go
```diff
@@ -162,27 +164,44 @@ func WithPluginsEnabled() ClientFunctionOption {
 	return pluginsEnabledOption{}
 }
 
-// RunnerProvider provides pluginrpc.Runners for program names and args.
-type RunnerProvider interface {
-	NewRunner(programName string, programArgs ...string) pluginrpc.Runner
+// Plugin is a configured lint or breaking plugin that wraps a pluginrpc.Runner.
+type Plugin interface {
```
bufdev, line 168: I don't fully understand this interface - it doesn't make sense to me why a Plugin is a Runner. Should I be able to execute arbitrary arguments on a Plugin? Almost for sure no.

emcfarlane: A plugin is run/invoked so I think having a method Run makes sense. Although maybe the interface name shouldn't be `Runner` as that does imply handling multiple.

bufdev: I probably should have been more direct, apologies - I don't think the interface signature makes sense here. I'm not super convinced by this change as a whole, wondering if your RunnerProvider idea might be less invasive.

PR #3332, opened 2024-09-20
Fix flaky symlink test on windows

emcfarlane:
Fix flaky test failing to remove all as file is still open. Forces the close on the opened bucket object. Removes the t.Cleanup to allow testing to handle windows closes more gracefully. See https://github.com/golang/go/blob/c208b913954514ac7cab0fa701fba9c89af70392/src/testing/testing.go#L1271

```
--- FAIL: TestOS (0.00s)
    --- FAIL: TestOS/get_non_existent_file_symlink (1.27s)
        testing.go:1232: TempDir RemoveAll cleanup: remove C:\Users\RUNNER~1\AppData\Local\Temp\TestOSget_non_existent_file_symlink229375574\001\foo.txt: The process cannot access the file because it is being used by another proces
```


oliversun9 (approved):
LGTM, thank you

PR #3341, opened 2024-09-23
Single threaded LSP concurrency model

emcfarlane:
Removes the locking structures from the LSP server. Requests are processed serially. This simplifies the implementation, but may need to be revisited later.


emcfarlane:
Closing. Chatted offline, concurrency primitives will be needed later.

PR #3342, opened 2024-09-24
Pull in new versions from `makego`

doriable:
This pulls in the latest versions from `makego` and runs `make upgrade`.

This adds a new check from `golangci-lint`, `G115`, which checks for overflow
from integer conversions. This PR includes fixes and exemptions for this check.
The check has several false positives that are still being worked on (a link to the 
issue is included in the comments in `.golangci.yml`) so exemptions have been added
to address false positives and/or known safe operations.

### DIFF
--- a/make/go/dep_buf.mk
+++ b/make/go/dep_buf.mk
@@ -7,8 +7,8 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/bufbuild/buf/releases 20240521 checked 20240524
-BUF_VERSION ?= v1.32.1
+# https://github.com/bufbuild/buf/releases 20240911 checked 20240916
+BUF_VERSION ?= v1.41.0
 # Settable
 #
 # If set, this path will be installed every time someone depends on $(BUF)
--- a/make/go/dep_protoc_gen_go.mk
+++ b/make/go/dep_protoc_gen_go.mk
@@ -7,9 +7,8 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/protocolbuffers/protobuf-go/releases 20240611 checked 20240612
-# TODO: Change back to released version once bug is fixed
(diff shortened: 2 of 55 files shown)


make/go/dep_buf.mk
```diff
@@ -7,8 +7,8 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/bufbuild/buf/releases 20240521 checked 20240524
-BUF_VERSION ?= v1.32.1
+# https://github.com/bufbuild/buf/releases 20240911 checked 20240916
+BUF_VERSION ?= v1.41.0
```
bufdev, line 11: FWIW we're on v1.42.0 now, but not that important.

make/go/dep_protoc_gen_go.mk
```diff
@@ -7,9 +7,8 @@ $(call _assert_var,CACHE_VERSIONS)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/protocolbuffers/protobuf-go/releases 20240611 checked 20240612
-# TODO: Change back to released version once bug is fixed
-PROTOC_GEN_GO_VERSION ?= 94ecbc26168965a670a0f7cf86f658131c790a9c
+# https://github.com/protocolbuffers/protobuf-go/releases 20240906 checked 20240916
+PROTOC_GEN_GO_VERSION ?= v1.34.3-0.20240906163944-03df6c145d96
```
bufdev, line 11: Why is this being done? We have always used commits here, not Go pseudo-versions

doriable: Looks like this was approved in https://github.com/bufbuild/makego/pull/153/files

make/go/dep_buf.mk
doriable: Yeah, this won't affect `bufbuild/buf`, but it's a good note to update `makego` for our other projects.

make/go/dep_protoc_gen_go.mk
bufdev: OK it's fine

bufdev (approved):

PR #3344, opened 2024-09-26
Reparse all extensions, not just unrecognized ones

jhump:
When handling user-provided schemas (images, file descriptor sets, etc), we already reparse unrecognized fields, since custom options will usually be unrecognized, but we need all fields recognized in order to serialize deterministically.

_However_, we really need to re-parse *all* extensions (and thus all custom options), because if any are recognized, it means we're using the program's linked-in global registry of extensions, which could have a different definition from the extension as it is defined in the schema we are handling.

So this adds a new `ReparseExtensions` and deprecated `ReparseUnrecognized`. The new function will reparsed unrecognized fields (since they are usually unrecognized extensions), but it will also reparse recognized extensions -- by serializing them to bytes and then de-serializing using the proper resolver.

I left the old one around, but deprecated, since it is used outside of this repo. But if we're okay with _forcing_ the other repos to update all call sites in the same PR that updates the dependency, I can remove it.


private/pkg/protoencoding/reparse_unrecognized.go
```diff
@@ -22,6 +22,8 @@ import (
 // ReparseUnrecognized uses the given resolver to parse any unrecognized fields in the
 // given reflectMessage. It does so recursively, resolving any unrecognized fields in
 // nested messages.
+//
+// Deprecated: Use ReparseExtensions instead.
```
bufdev, line 26: Is there a specific reason we need to keep this around? If it's still used in core, refactor core to not use it

bufdev: I think description answers this - yes, force other repos to refactor

jhump: Done

private/pkg/protoencoding/json_marshaler.go
```diff
@@ -38,7 +38,7 @@ func newJSONMarshaler(resolver Resolver, options ...JSONMarshalerOption) Marshal
 }
 
 func (m *jsonMarshaler) Marshal(message proto.Message) ([]byte, error) {
-	if err := ReparseUnrecognized(m.resolver, message.ProtoReflect()); err != nil {
+	if err := ReparseExtensions(m.resolver, message.ProtoReflect()); err != nil {
```
jhump, line 41: Oops, this was a call site I missed in first commit...

bufdev (approved):

PR #3345, opened 2024-09-26
Default protoencoding resolver to an empty resolver 

emcfarlane:
This changes the behaviour of protoencoding to always set a resolver on init, defaulting to the `protoencoding.EmtpyResolver`. Currently, Marshaller and Unmarshallers will default to the `protoregistry.GlobalTypes` if a nil resolver is set. To keep the same behaviour the caller must pass in the GlobalTypes explicitly.


bufdev (approved):
Fine by me, but I'd verify that this doesn't bork anything in core.

doriable (approved):

bufdev (approved):

PR #3346, opened 2024-09-26
Force usage of private/pkg/protoencoding

bufdev:
This probably should have been done a long time ago. This forces usage of `private/pkg/protoencoding` so that we don't keep on getting marshal/unmarshal problems, and stay consistent. A follow-up should do this in core as well.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -21,6 +21,15 @@ linters-settings:
       - '^log\.'
       - '^print$'
       - '^println$'
+      # Use private/pkg/protoencoding Marshalers and Unmarshalers
+      - '^proto.Marshal$'
+      - '^proto.Unmarshal$'
+      - '^proto.MarshalOptions$'
+      - '^proto.UnmarshalOptions$'
+      - '^protojson.Marshal$'
+      - '^protojson.Unmarshal$'
+      - '^protojson.MarshalOptions$'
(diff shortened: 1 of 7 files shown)


emcfarlane (approved):

doriable (approved):

.golangci.yml
```diff
@@ -21,6 +21,15 @@ linters-settings:
...
       - '^println$'
+      # Use private/pkg/protoencoding Marshalers and Unmarshalers
+      - '^proto.Marshal$'
+      - '^proto.Unmarshal$'
+      - '^proto.MarshalOptions$'
+      - '^proto.UnmarshalOptions$'
+      - '^protojson.Marshal$'
+      - '^protojson.Unmarshal$'
+      - '^protojson.MarshalOptions$'
+      - '^protojson.UnmarshalOptions$'
```
jhump, line 32: Should this include YAML, too?

jhump (approved):

.golangci.yml
bufdev: Ugh probably - @emcfarlane any chance you can commandeer and add?

bufdev:
Actually @emcfarlane just going to merge this as-is, can you add protoyaml as a follow-up?

PR #3348, opened 2024-09-26
Unbreak CI after #3316

mcy:


.golangci.yml
```diff
@@ -110,6 +110,26 @@ issues:
       # G404 checks for use of the ordinary non-CPRNG.
       path: private/buf/buflsp/progress.go
       text: "G404:"
+    - linters:
```
bufdev, line 113: @doriable can you take a look to see if these exceptions make sense?

doriable: Yep -- so truncating conversions is definitely a known false positive in the issue I put for the G115 exemptions. I went through each one and they are all good. We already have a clear comment for the one in `report.go` also for converting the version, so this looks good to me.

doriable (approved):

PR #3363, opened 2024-10-02
Add curated plugin doc and collection fields

mfridman:


proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -112,6 +112,21 @@ enum DotnetTargetFramework {
   DOTNET_TARGET_FRAMEWORK_NET_8_0 = 13;
 }
 
+// The collection a Plugin belongs to. These collections help organize plugins based on their
+// functionality or ecosystem.
+enum PluginCollection {
```
bufdev, line 117: This feels like a completely new entity, not an enum - it should probably be dynamic

mfridman: Sure, updated. I could see us expanding this to include a description, maybe a logo, create_time, etc.

bufdev: This doesn't appear to be an entity - it has no ID as an example.

How does this fit into the new API, why is this going here, etc? Need more details.

mfridman: It's a hard-coded list of plugin mappings to collections. These aren't intended to be user-configurable and there's no mechanism for users to edit these collections (at least for now).

The goal is to display older protoc plugins in the UI under these collections.

This is something we should put more careful thought into when designing the newer `plugin.v1` API in registry-proto, but to unblock certain UI work feels fine to piggyback on the existing primitives?

bufdev (approved):
Approving to unblock, but this shape won't be approved in the API that we will be using in v1

PR #3370, opened 2024-10-04
Temporarily serialize requests in the LSP, and fix the bugs that doing so revealed

### LINKED ISSUE #3359: `buf beta lsp` local go-to-definition issue

https://github.com/bufbuild/registry-proto

### Commands

```shell
nvim buf/registry/module/v1/module_service.proto
# Go to line 60, column 12 (cursor on `ModuleRef`)
# Invoke LSP go-to-definition
```

### Output

Nothing — the definition is not opened.

The logs are _very_ noisy (from `:LspLog`); here's the first couple lines:

```
[START][2024-10-01 10:11:35] LSP logging initiated
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[34mINFO\27[0m\tnew file version: file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module_service.proto, 0 -> 0\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module_service.proto, 0\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module.proto, -1\n"
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/owner/v1/owner.proto, -1\n"
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/.cache/buf/v3/modules/b5/buf.build/bufbuild/protovalidate/b983156c5e994cc9892e0ce3e64e17e0/files/buf/validate/validate.proto, -1\n"
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/.cache/buf/v3/wellknowntypes/27.0/google/protobuf/descriptor.proto, -1\n"
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tcould not find URI for import "buf/validate/expression.proto"\n\27[33mWARN\27[0m\tcould not find URI for import "buf/validate/priv/private.proto"\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/.cache/buf/v3/wellknowntypes/27.0/google/protobuf/duration.proto, -1\n\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/.cache/buf/v3/wellknowntypes/27.0/google/protobuf/timestamp.proto, -1\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tcould not find image for "file:///Users/stefanvanburen/.cache/buf/v3/modules/b5/buf.build/bufbuild/protovalidate/b983156c5e994cc9892e0ce3e64e17e0/files/buf/validate/validate.proto"\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:11:35] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tcould not find image for "file:///Users/stefanvanburen/.cache/buf/v3/modules/b5/buf.build/bufbuild/protovalidate/b983156c5e994cc9892e0ce3e64e17e0/files/buf/validate/validate.proto"\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
```

It then seems to continuously try to find the image for `bufbuild/protovalidate/b9831...` and fail over and over.

### Expected Output

I expected to go to the [definition of `ModuleRef`](https://github.com/bufbuild/registry-proto/blob/9533816e2b0a8f77dde4e07bfb28ca4eba6c4c80/buf/registry/module/v1/module.proto#L103) (within the same package at `./module.proto:103:9`).

### Anything else?

I haven't used the LSP enough to know for sure, but this seems like an issue only with same-package symbols - go-to-definition works with e.g. `buf.registry.owner.v1.OwnerRef` on line 92 in the same file, and all of the `buf.validate` extensions, for example.

### LINKED ISSUE #3361: `buf beta lsp` broken link on WKT docs

https://github.com/bufbuild/registry-proto

### Commands

```shell
nvim buf/registry/module/v1/module_service.proto
# Go to line 29 column 12 (cursor on `i` in `idempotency_level`)
# Invoke LSP `hover`
```

### Output

In the Hover "buffer" you'll see:

````
```proto-decl
field google.protobuf.MethodOptions.idempotency_level


[`google.protobuf.MethodOptions.idempotency_level` on the Buf Schema Registry](https://%!s(&lt;nil&gt;)/docs/main:google.protobuf#google.protobuf.MethodOptions.idempotency_level)

&lt;missing docs&gt;
```
````

### Expected Output

````
```proto-decl
field google.protobuf.MethodOptions.idempotency_level
```

[`google.protobuf.MethodOptions.idempotency_level` on the Buf Schema Registry](https://buf.build/protocolbuffers/wellknowntypes/docs/main:google.protobuf#google.protobuf.MethodOptions)

&lt;missing docs&gt;
````

### Anything else?

It might be an edge case for options, but note that the original link was trying to go directly to the `idempotency_level` field of MethodOptions, which doesn't have a corresponding anchor tag in the BSR. I think going to the `MethodOptions` message tag is Good Enough, otherwise we'll need to figure out how to add anchors to individual fields in the BSR.

### LINKED ISSUE #3362: `buf beta lsp` spins on empty files

https://github.com/bufbuild/registry-proto

### Commands

```shell
echo "syntax = \"proto3\";" > buf/registry/module/v1/commit.proto
nvim buf/registry/module/v1/commit.proto
```

### Output

Blocks the LSP server. Starting from the empty file causes the the `SemanticTokensFull` to [spin](https://github.com/bufbuild/buf/blob/40848114247eee1bc044baebbd10ca97552adef8/private/buf/buflsp/server.go#L326-L334). Adding a message doesn't fix it unless we also restart the LSP server with the now non-empty proto file.

### Expected Output

Return an empty response if the file is empty.

### Anything else?

_No response_

mcy:
Fixes #3362, #3359, #3361.

Previously, the LSP contained a somewhat fussy concurrency model, where files had to be carefully locked and unlocked (with runtime enforcement) to avoid deadlocks. This is necessary so that parsing and resolution does not hang the LSP for particularly long: you want to avoid a situation where a user opens a file that takes several seconds to load, which is a realistic possibility.

This does not hang the editor itself, but prevents go-to-definition from working on already-loaded files while the new file is loading. Already with only needing to go three or four levels deep, as in the protos in `buf-registry`, causes a noticeable pause. I have not profiled it, but I suspect it's because the CLI is not optimized for memoization, so the memoization that I'm doing probably duplicates work other parts of the CLI are doing.

However, the current solution was understood to be temporary, since Josh and I are designing a new query system for incremental compilation, which would replace the current system. Also, the current system is quite painful to debug, despite the many steps I have taken to enforce locking invariants.

Therefore, given that:

1. The current solution is going to be replaced,
2. The current solution is a velocity speed bump, because it's tricky to debug,
3. The LSP is useable with a less-than-ideal user experience until (1),

it makes sense to switch to a global lock to serialize all requests until (1). Removing (2) will make it easier to fix "low hanging" bugs in the LSP, instead of blocking on (1) happening. If we start pushing the LSP to users before (1), we should beware of (3).

As a result of switching to serialized requests, #3362 disappears in a puff of "this problem no longer exists". It also revealed that the way dependency paths were produced by `GetImportableImageFileInfos` was subtly incorrect if the input file path was inside of the cache. Fixing this resolved #3359. Because we now we have crisp information about where imports come from, it is trivial to resolve #3361.


private/buf/bufctl/controller.go
```diff
@@ -129,6 +130,7 @@ type Controller interface {
 		defaultMessageEncoding buffetch.MessageEncoding,
 		options ...FunctionOption,
 	) error
+	GetWKTBucket(ctx context.Context) (storage.ReadBucket, error)
```
bufdev, line 133: This shouldn't be part of the `Controller` really - instead, expose `bufcli.newWKTStore` as public `bufcli.NewWKTStore` (ie just `s/n/N/`) and call this function in `buflsp` to make a `bufwktstore.Store`.

mcy: Done

private/buf/buflsp/file.go
```diff
@@ -357,15 +327,27 @@ func (f *file) IndexImports(ctx context.Context) {
...
 		var imported *file
 		if fileInfo.LocalPath() == f.uri.Filename() {
 			imported = f
 		} else {
 			imported = f.Manager().Open(ctx, protocol.URI("file://"+fileInfo.LocalPath()))
+
+			isLocal := fileInfo.LocalPath() == fileInfo.ExternalPath()
+			if !isLocal && imported.importablePathToObject == nil {
+				// If this is an external file, it will be in the cache and therefore
+				// finding imports via lsp.findImportable() will not be possible.
```
doriable, line 345: I am not entirely sure I follow this logic -- so a file we may be checking may be a dependency file, and therefore the URI would be to a file in the cache. But I'm not sure why that would then mean we would not be able to call `controller.GetImportableFileInfos`? It appears to me that we are basically replicating that logic in `findImportable` (resolving the workspace using the controller, walking the files, and adding WKTs)...

bufdev:
What's the status on this one?

private/buf/buflsp/file.go
mcy: GetImportableFileInfos and findImportable will both produce "truncated" paths when looking up the workspace of a cached file, presumably because it does not have a buf.yaml. I think trying to fix that is beyond the scope of this PR, so I'm working around it. A comment will appear soon explaining it.

mcy:
I'm currently stuck debugging the issue Doria pointed out (insofar that I think there's a way I'm using workspaces wrong but docs are a bit paltry), but unfortunately the logging migration appears to have broken debugging for the LSP (and broken some other things along with it? still triaging). Not sure I can give an ETA on that.

bufdev:
I'm not sure how the logging would have impacted it, but if you can point out specifics happy to help. The changes were largely superficial, however.

private/buf/buflsp/file.go
doriable: Ah, I think I might know what's going on. In the case of the cache, we do not have a way to show the root of the workspace (e.g. there are no configuration files), so when resolving the workspace for the given file path, it is truncating it to that directory. There may be a way to fix this by special-casing paths within the cache, since those are known modules.

mcy:
The biggest issue was that the function https://github.com/bufbuild/buf/pull/3370/files#diff-9fde706d21f072dfdf6f3fc98fc5dd6035d8aca46905dc8d6e819cf3cdde4f5bR335 was deleted with no replacement. slog does not natively know how to log arbitrary types, and passing such values into `slog.Any` does not cause anything to actually be logged. There's a lot of places where symbols are logged to track whether symbol resolution is working correctly and it took me a while to figure out why logging wasn't coming out right.

Also there were several unfortunate merge conflicts because the logging upgrade touched every log statement in the LSP, so merging main introduced additional bugs I had to track down while fixing logging.

doriable (approved):

PR #3373, opened 2024-10-05
Remove CI for Go 1.21

bufdev:
We are up to Go 1.22 now. As of Go 1.21, we no longer need to go back three versions, and we went to 1.22 a while back.

### DIFF
--- a/.github/workflows/previous.yaml
+++ b/.github/workflows/previous.yaml
@@ -9,7 +9,7 @@ jobs:
   test-previous:
     strategy:
       matrix:
-        go-version: ['1.21.x', '1.22.x']
+        go-version: ['1.22.x']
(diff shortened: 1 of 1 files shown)


.github/workflows/previous.yaml
```diff
@@ -9,7 +9,7 @@ jobs:
   test-previous:
     strategy:
       matrix:
-        go-version: ['1.21.x', '1.22.x']
+        go-version: ['1.22.x']
```
mfridman, line 12: Not sure if we want to keep bumping this manually, but setup-go supports `'oldstable'` and `'stable'` [aliases](https://github.com/actions/setup-go?tab=readme-ov-file#using-stableoldstable-aliases) which they keep those up-to-date.

PR #3393, opened 2024-10-08
Add short-circuit to multi read bucket in overlay mode

mfridman:
Small optimization for `multiReadBucket`. The goal is to avoid a `Stat` call when overlay is enabled and the file is found.

### DIFF
--- a/private/pkg/storage/multi.go
+++ b/private/pkg/storage/multi.go
@@ -79,7 +79,9 @@ func (m *multiReadBucket) Get(ctx context.Context, path string) (ReadObjectClose
 	if m.overlay {
 		// If overlay is enabled, attempt a Get operation against the first bucket. If the file is
 		// found, we avoid the potentially expensive Stat call.
-		if readObjectCloser, err := m.delegates[0].Get(ctx, path); err == nil && readObjectCloser != nil {
+		if readObjectCloser, err := m.delegates[0].Get(ctx, path); err != nil && !IsNotExist(err) {
+			return nil, err
+		} else if readObjectCloser != nil {
 			return readObjectCloser, nil
 		}
 	}
--- a/private/pkg/storage/multi.go
+++ b/private/pkg/storage/multi.go
@@ -16,6 +16,7 @@ package storage
 
 import (
 	"context"
+	"errors"
 	"io/fs"
 
 	"github.com/bufbuild/buf/private/pkg/storage/storageutil"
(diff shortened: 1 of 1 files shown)


private/pkg/storage/multi.go
```diff
@@ -76,6 +76,15 @@ func newMultiReadBucket(
 }
 
 func (m *multiReadBucket) Get(ctx context.Context, path string) (ReadObjectCloser, error) {
+	if m.overlay {
+		// If overlay is enabled, attempt a Get operation against the first bucket. If the file is
+		// found, we avoid the potentially expensive Stat call.
+		if readObjectCloser, err := m.delegates[0].Get(ctx, path); err != nil && !IsNotExist(err) {
+			return nil, err
+		} else if readObjectCloser != nil {
```
bufdev, line 84: I don't think this is right, I think you want

if err != nil {
  if !IsNotExist(err) {
    return nil, err
  }
  return nil, nil
}
return readObjectCloser, nil

bufdev: edit: nevermind

bufdev, line 84: Add comment: `// err == nil || IsNotExist(err)`

bufdev, line 84: Also add comment "if we did find a ReadObjectCloser, return it, otherwise do the expensive logic to check every delegate"

You could also optimize this further to skip the first delegate via a `bool` to m.getObjectInfoAndDelegateIndex

mfridman: Scanned the codebase a bit, and updated the logic to match the typical style of handling the error within the error block and adding an else otherwise.

private/pkg/storage/multi.go
```diff
@@ -76,6 +77,20 @@ func newMultiReadBucket(
...
 
 func (m *multiReadBucket) Get(ctx context.Context, path string) (ReadObjectCloser, error) {
+	if m.overlay {
+		// If overlay is enabled, attempt a Get operation against the first bucket. If the file is
+		// found, we avoid the potentially expensive Stat call.
+		readObjectCloser, err := m.delegates[0].Get(ctx, path)
+		if err != nil {
+			if !errors.Is(err, fs.ErrNotExist) {
+				return nil, err
+			}
```
bufdev, line 87: Probably a comment here "fallthrough to expensive logic to check every delegate"

mfridman: Added in https://github.com/bufbuild/buf/pull/3393/commits/386db91e616773a90fc8fb7a1129142cbea3c43d

PR #3404, opened 2024-10-16
Implement breaking lints in the LSP

mcy:
This change adds breaking lints to the LSP as part of the general lint pass.

Breaking lints are interesting because they need an "against". I've chosen to expose two configuration knobs (which are set in the settings for an LSP in the standard way):

1. `buf.against` sets the against strategy. For the LSP, there are three that reasonably make sense: against some remote's HEAD, against your local HEAD, and against whatever is on disk. This means that lints go away when pushing to main, committing working changes, or saving in the editor, respectively.
2. `buf.gitRemote`, the remote to use for the above. "origin" is the default for almost everyone, but it need not be that, so it's a config knob.

Also, in the process of standing up loading config settings from the client, I found a deadlock in the `jsonrpc2` library. To work around it, every request runs on its own goroutine (although they are still serialized wrt to the big LSP lock). This ensures that the single goroutine `jsonrpc2` spawns to actually poll the IPC socket doesn't deadlock. Why do they only spawn one goroutine? Presumably because they expect users to spawn a goroutine to handle each request, but this does not appear to be visibly documented. (Much debugging of deadlocked goroutines happened today.)


private/buf/buflsp/file.go
```diff
@@ -161,6 +198,107 @@ func (f *file) Update(ctx context.Context, version int32, text string) {
...
+		// This is an invalid remote, so it means gitRemote should be
+		// interpreted as a local branch name instead.
+		f.gitRemoteBranch = f.gitRemote
+		f.gitRemote = ""
+		f.lsp.logger.Debug(
+			"found local branch",
+			slog.String("uri", string(f.uri)),
+			slog.String("ref", f.gitRemoteBranch),
+		)
+	} else if err != nil {
```
doriable, line 239: This behaviour is somewhat reasonable, but I wonder if branch + remote should just be different config fields with sane defaults for clarity. Even if this is documented behaviour, the config field being named `gitRemote` but pointing to a local branch might cause some confusion. This would have cascading changes through the `fileOpener`, etc.

private/buf/buflsp/file.go
```diff
@@ -407,132 +542,115 @@ func (f *file) IndexImports(ctx context.Context) {
...
+				data, err = readAtRef(f.gitRemote + "/" + f.gitRemoteBranch)
+			}
+		case againstHead:
+			data, err = readAtRef("HEAD")
 		}
 
-		imageFiles = append(imageFiles, imageFile)
-		f.lsp.logger.Debug(fmt.Sprintf("added image file for %s", descriptor.Path()))
+		if data == nil || errors.Is(err, git.ErrNotARepo) {
+			return os.Open(fileInfo.LocalPath())
```
doriable, line 617: For this case, should we check `f.IsLocal()`?

private/buf/buflsp/file.go
```diff
@@ -407,132 +542,115 @@ func (f *file) IndexImports(ctx context.Context) {
...
+		return ioext.CompositeReadCloser(bytes.NewReader(data), ioext.NopCloser), err
 	}
+}
 
-	if err != nil {
-		f.lsp.logger.Warn("could not build image", slog.String("uri", string(f.uri)), slogext.ErrorAttr(err))
+// BuildImage build Buf Images for this file, to be used with linting
+// routines.
+//
+// This operation requires IndexImports().
```
doriable, line 627: Nit: `%s/BuildImage/BuildImages`

private/pkg/git/git.go
```diff
@@ -352,6 +358,67 @@ func GetRefsForGitCommitAndRemote(
...
+		} else if errors.Is(err, os.ErrNotExist) {
+			parent := filepath.Dir(dir)
+			if parent == dir {
+				return nil, fmt.Errorf("could not find .git directory for %s: %w", orig, ErrNotARepo)
+			}
+			dir = parent
+		} else {
+			return nil, err
+		}
+	}
```
doriable, line 396: We could use `CheckDirectoryIsValidGitCheckout` to check if this is a git repository:
https://github.com/bufbuild/buf/blob/b886118276b22562ead39d3debbffd1774460040/private/pkg/git/git.go#L215

The call-site will need to handle `ErrInvalidGitCheckout` instead.

private/buf/buflsp/file.go
```diff
@@ -31,22 +32,46 @@ import (
...
-	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+const (
+	againstTrunk againstKind = iota + 1
+	againstHead
+	againstSaved
 )
 
 const descriptorPath = "google/protobuf/descriptor.proto"
```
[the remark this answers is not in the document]
mcy: Done

private/buf/buflsp/file.go
mcy: No: the local path is the path on disk, which is the file we need to open for e.g. things that are in cache.

private/buf/buflsp/file.go
mcy: Done

private/pkg/git/git.go
mcy: That's not quite what we want. We want to get the repository root, which requires traversing upwards (vs IsValidCheckout, which just kicks `rev-parse`).

private/buf/buflsp/file.go
mcy: I think I'm going to change the semantics as follows:
1. againstKind will be `againstGit` or `againstSaved`.
2. There will be a single config knob for the ref to use for `againstGit`, with a default of `origin/HEAD`.

private/buf/buflsp/file.go
```diff
@@ -170,6 +198,99 @@ func (f *file) Update(ctx context.Context, version int32, text string) {
...
+			{ScopeURI: f.uri, Section: ConfigBreakingGitRef},
+		},
+	})
+	if err != nil {
+		// We can throw the error away, since the handler logs it for us.
+		return
+	}
+
+	f.againstStrategy = getSetting(f, settings, ConfigBreakingStrategy, 0, againstFromString)
+	f.againstGitRef = getSetting(f, settings, ConfigBreakingGitRef, 1, func(s string) (string, bool) { return s, true })
```
doriable, line 218: Might be worth commenting here the indices follow the ordering from the `ConfigurationParams.Items` from the request above.

private/buf/buflsp/file.go
```diff
@@ -170,6 +198,99 @@ func (f *file) Update(ctx context.Context, version int32, text string) {
...
+
+	switch f.againstStrategy {
+	case againstDisk:
+		f.againstGitRef = ""
+	case againstGit:
+		// Check to see if buf.againstGit is a valid ref.
```
doriable, line 224: Need to update this comment to match the config key.

private/buf/buflsp/file.go
```diff
@@ -188,12 +309,13 @@ func (f *file) Refresh(ctx context.Context) {
 	progress.Report(ctx, "Indexing Imports", 2.0/6)
 	f.IndexImports(ctx)
 
-	progress.Report(ctx, "Detecting Module", 3.0/6)
+	progress.Report(ctx, "Detecting Buf Module", 3.0/6)
 	f.FindModule(ctx)
 
-	progress.Report(ctx, "Linking Descriptors", 4.0/6)
-	f.BuildImage(ctx)
+	progress.Report(ctx, "Running Lints", 4.0/6)
```
doriable, line 315: Similar to the comment `config.go`, this should be `"Running Checks"`.

private/buf/buflsp/config.go
```diff
@@ -0,0 +1,24 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package buflsp
+
+// Configuration keys for the LSP.
+// Keep in sync with bufbuild/vscode-buf/package.json.
+const (
+	ConfigCLIPath          = "buf.commandLine.path"
+	ConfigCLIUpdates       = "buf.commandLine.checkForUpdates"
```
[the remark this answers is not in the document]
mcy: Yeah I don't think those are worth keeping around.

private/buf/buflsp/config.go
```diff
@@ -0,0 +1,24 @@
...
+	ConfigBreakingStrategy = "buf.linting.breakingAgainst.strategy"
+	ConfigBreakingGitRef   = "buf.linting.breakingAgainst.gitRef"
```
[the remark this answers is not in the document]
mcy: Sure.

doriable (commented):
Thanks for putting in these changes -- we should add a changelog entry for this and this is probably good to merge.

doriable (approved):

doriable (approved):

PR #3411, opened 2024-10-20
Improve lsp hover definition comment formatting

hirasawayuki:
This PR introduces a new `formatComment` function to enhance the formatting of comments displayed in hover definitions. The implementation handles both single-line and multi-line comments, ensuring proper formatting while maintaining Markdown compatibility.

## Before
screenshot2

As shown above, hover definitions currently display unnecessary asterisks from the proto file's comment syntax, making the documentation harder to read.

## After

screenshot1

The new implementation removes the redundant asterisks while preserving the comment's structure, resulting in cleaner and more readable documentation in hover definitions.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -493,24 +493,12 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
 	var printed bool
 	for _, comments := range allComments {
 		for i := 0; i < comments.Len(); i++ {
-			comment := comments.Index(i).RawText()
-
 			// The compiler does not currently provide comments without their
 			// delimited removed, so we have to do this ourselves.
-			if strings.HasPrefix(comment, "//") {
-				// NOTE: We do not trim the space here, because indentation is
-				// significant for Markdown code fences, and if every line
-				// starts with a space, Markdown will trim it for us, even off
-				// of code blocks.
-				comment = strings.TrimPrefix(comment, "//")
-			} else {
-				comment = strings.TrimSuffix(strings.TrimPrefix(comment, "/*"), "*/")
-			}
-
+			comment := formatComment(comments.Index(i).RawText())
 			if comment != "" {
 				printed = true
 			}
-
 			// No need to process Markdown in comment; this Just Works!
 			fmt.Fprintln(&tooltip, comment)
 		}
@@ -521,6 +509,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
 	}
 
 	return tooltip.String()
+}
+
+// formatComment takes a raw comment string and formats it by removing comment
+// delimiters and unnecessary whitespace. It handles both single-line (//) and
+// multi-line (/* */) comments.
+func formatComment(comment string) string {
+	comment = strings.TrimSpace(comment)
+
+	if strings.HasPrefix(comment, "//") {
+		// NOTE: We do not trim the space here, because indentation is
+		// significant for Markdown code fences, and if every line
+		// starts with a space, Markdown will trim it for us, even off
+		// of code blocks.
+		return strings.TrimPrefix(comment, "//")
+	}
+
+	if strings.HasPrefix(comment, "/*") && strings.HasSuffix(comment, "*/") {
+		comment = strings.Trim(comment, "/*")
+		// single-line
+		if !strings.Contains(comment, "\n") {
+			return strings.TrimSpace(comment)
+		}
+
+		lines := strings.Split(strings.TrimSpace(comment), "\n")
+		for i, line := range lines {
+			line = strings.TrimSpace(line)
+			line = strings.TrimLeft(line, "*")
+			line = strings.TrimPrefix(line, " ")
+			lines[i] = line
+		}
+
+		return strings.Join(lines, "\n")
+	}
+
+	return comment
 }
 
 // symbolWalker is an AST walker that generates the symbol table for a file in IndexSymbols().
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -495,7 +495,7 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
 		for i := 0; i < comments.Len(); i++ {
 			// The compiler does not currently provide comments without their
 			// delimited removed, so we have to do this ourselves.
-			comment := formatComment(comments.Index(i).RawText())
+			comment := commentToMarkdown(comments.Index(i).RawText())
 			if comment != "" {
 				printed = true
 			}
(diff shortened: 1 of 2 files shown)


private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
 	return tooltip.String()
 }
 
+// formatComment takes a raw comment string and formats it by removing comment
+// delimiters and unnecessary whitespace. It handles both single-line (//) and
+// multi-line (/* */) comments.
+func formatComment(comment string) string {
```
mcy, line 517: I think a better name for this is to rename it to something like commentToMarkdown, since the important thing is that this is emitting a Markdown string.

private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+			return strings.TrimSpace(comment)
+		}
+
+		lines := strings.Split(strings.TrimSpace(comment), "\n")
+		for i, line := range lines {
+			line = strings.TrimSpace(line)
+			line = strings.TrimLeft(line, "*")
+			line = strings.TrimPrefix(line, " ")
+			lines[i] = line
+		}
```
mcy, line 541: I would prefer if this only triggered if the comment starts with `/**` Otherwise, the following will be mangled:

```proto
/*

my things:
* foo
* bar
* baz

*/
```

mcy: I would also like to see a comment that explains that what this is processing is a Doxygen-style comment, with an inline example of what we expect the comment to look like.

private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+		// NOTE: We do not trim the space here, because indentation is
+		// significant for Markdown code fences, and if every line
+		// starts with a space, Markdown will trim it for us, even off
+		// of code blocks.
+		return strings.TrimPrefix(comment, "//")
+	}
+
+	if strings.HasPrefix(comment, "/*") && strings.HasSuffix(comment, "*/") {
+		comment = strings.Trim(comment, "/*")
+		// single-line
```
mcy, line 530: Comments should be complete sentences.

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
+package buflsp
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_formatComment(t *testing.T) {
```
mcy, line 9: Don't use underscores in test names. Just name it TestFormatComment or equivalent.

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
+package buflsp
```
mcy, line 1: Missing license notice.

private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+		if !strings.Contains(comment, "\n") {
+			return strings.TrimSpace(comment)
+		}
+
+		lines := strings.Split(strings.TrimSpace(comment), "\n")
+		for i, line := range lines {
+			line = strings.TrimSpace(line)
+			line = strings.TrimLeft(line, "*")
```
mcy, line 538: This should be TrimPrefix.

private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+			line = strings.TrimPrefix(line, " ")
```
mcy, line 539: This seems incorrect for the same reason as stated in my `//NOTE` comment.

private/buf/buflsp/symbol.go
```diff
@@ -523,6 +511,41 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+	if strings.HasPrefix(comment, "//") {
+		// NOTE: We do not trim the space here, because indentation is
+		// significant for Markdown code fences, and if every line
+		// starts with a space, Markdown will trim it for us, even off
+		// of code blocks.
+		return strings.TrimPrefix(comment, "//")
+	}
+
+	if strings.HasPrefix(comment, "/*") && strings.HasSuffix(comment, "*/") {
+		comment = strings.Trim(comment, "/*")
```
mcy, line 529: This will mangle some comments, such as `/*This is *important**/`. Please include a test case to handle this.

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+		expected string
+	}{
+		{
+			name:     "Single-line comment",
+			input:    "// This is a single-line comment",
+			expected: " This is a single-line comment",
+		},
+		{
+			name:     "Multi-line comment",
+			input:    "/*\n This is a\n multi-line comment\n */",
```
mcy, line 24: Use go raw strings `` `xyz` `` instead, to make this easier to read.

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expected: "This is a\nmulti-line comment",
+		},
+		{
+			name:     "Multi-line comment with mixed indentation",
+			input:    "/*\n  * First line\n  * - Second line\n  *   - Third line\n */",
```
mcy, line 29: This test case is incorrect. You should not trim asterisks if the comment does not begin with `/**`. Alternatively, you could instead only trim asterisks if *every* line starts with one.

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expected: "This is a\nmulti-line comment\nwith multi-asterisks",
+		},
+		{
+			name:     "Single-line multi-line comment",
+			input:    "/* Single-line multi-line comment */",
+			expected: "Single-line multi-line comment",
+		},
+		{
+			name:     "Empty comment",
+			input:    "/**/",
```
mcy, line 44: Can you add a test case for there being a space in this?

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+func Test_formatComment(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name     string
+		input    string
+		expected string
+	}{
+		{
+			name:     "Single-line comment",
```
mcy, line 18: can you make the test names be all-lowercase, separated by dashes?

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+			expected: "Single-line multi-line comment",
+		},
+		{
+			name:     "Empty comment",
+			input:    "/**/",
+			expected: "",
+		},
+	}
+
+	for _, tt := range tests {
```
mcy, line 49: `tt` -> `test`

private/buf/buflsp/symbol_test.go
```diff
@@ -0,0 +1,58 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			result := formatComment(tt.input)
+			if result != tt.expected {
+				assert.Equal(t, tt.input, result)
+			}
```
mcy, line 55: Rewrite this to
```go
assert.Equal(t, text.expected, formatComment(test.input))
```

mcy (commented):
Sorry for the delay. This hasn't been very high on my priority list and I've only just had a chance to look at this.

hirasawayuki:
@mcy 
Thank you for your review! I've addressed your feedback and made the suggested changes.
Could you please review the updated code?

mcy (approved):

doriable (approved):

PR #3416, opened 2024-10-22
Add `buf registry whoami` command

### LINKED ISSUE #3414: Allow a user to determine if they are logged into the BSR.
### Feature

In my local workflow, I want to double check that a user has a netrc with credentials in it so they can be authenticated to the BSR. I use this check to make sure that a docker build does not fail on their machine since we are mounting their netrc in as a secret to do client downloads from BSR.

## Change I'd like to see

The buf cli could probably have something like `buf registry auth status` or some similar commands that could tell you that you are logged in or not. The message that it prints isn't really important (could print the auth json) the more important part would be that it exits 0 if auth'ed and exits 1 if not auth'ed.

## What does my bash code look like without this feature?

```bash
echo "Checking Buf Schema Registry login..."
bsr_auth_http_code=$(curl -o /dev/null -s -w "%{http_code}" -L -n {bsr_module_url}/@v/list)
if [ "$bsr_auth_http_code" -eq 200 ]; then
    echo "Successfully logged into BSR"
else
    echo "Please auth with the buf cli"
    exit 1
fi
echo -e "Done!\n"
```

`buf curl` could also work for this.

doriable:
This adds a `buf registry whoami` command, which checks if you
are logged in to the Buf Schema Registry at a given domain. The domain
defaults `buf.build` if none is provided.

Fixes #3414


private/buf/cmd/buf/command/beta/registry/whoami/whoami.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+	authnService := connectclient.Make(clientConfig, remote, registryv1alpha1connect.NewAuthnServiceClient)
+	currentUserResponse, err := authnService.GetCurrentUser(ctx, connect.NewRequest(&registryv1alpha1.GetCurrentUserRequest{}))
+	if err != nil {
+		if connectErr := new(connect.Error); errors.As(err, &connectErr) && connectErr.Code() == connect.CodeUnauthenticated {
+			return fmt.Errorf("Not currently logged in for %s", remote)
+		}
+		return err
```
doriable, line 84: Right now, we aren't really differentiating between unauthenticated vs. other error states, everything exits `1` unless authenticated (which exits `0`).

The error message will indicate the nature of the error, for example, in the case of an invalid remote:

```
$ buf beta registry whoami fake.address
Failure: the server hosted at that remote is unavailable. Are you sure "fake.address" is a valid remote address?
```

But we do not do any special handling for the errors.

private/buf/cmd/buf/command/beta/registry/whoami/whoami.go
```diff
@@ -0,0 +1,92 @@
...
+			},
+		),
+		BindFlags: flags.Bind,
+	}
+}
+
+type flags struct{}
+
+func newFlags() *flags {
+	return &flags{}
```
emcfarlane, line 57: What do you think about a `--format=json` for accessing the username?

emcfarlane (approved):

private/buf/cmd/buf/command/beta/registry/whoami/whoami.go
doriable: I like this! I can add this.

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `buf beta registry whoami` command, which checks if you are logged in to the Buf Schema
```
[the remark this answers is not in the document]
bufdev: No

bufdev (commented):
Probably good to go

CHANGELOG.md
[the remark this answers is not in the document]
doriable: Moved this out of `beta` and directly to `buf registry whoami`.

private/buf/cmd/buf/command/beta/registry/whoami/whoami.go
doriable: Added `--format=json` (and `--format=text` will print the "logged in as" statement as before).

private/buf/bufprint/bufprint.go
```diff
@@ -248,6 +248,14 @@ func NewOrganizationEntity(organization *ownerv1.Organization, remote string) En
 	}
 }
 
+func NewUserEntity(user *registryv1alpha1.User) Entity {
```
bufdev, line 251: No godoc

private/buf/cmd/buf/command/registry/whoami/whoami.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/bufbuild/buf/private/pkg/connectclient"
+	"github.com/bufbuild/buf/private/pkg/netext"
+	"github.com/spf13/pflag"
+)
+
+const (
+	formatFlagName = "format"
+)
+
+func NewCommand(
```
bufdev, line 39: No godoc

bufdev:
Adjust the title and description

private/buf/cmd/buf/command/registry/whoami/whoami.go
```diff
@@ -0,0 +1,122 @@
...
+	currentUserResponse, err := authnService.GetCurrentUser(ctx, connect.NewRequest(&registryv1alpha1.GetCurrentUserRequest{}))
+	if err != nil {
+		if connectErr := new(connect.Error); errors.As(err, &connectErr) && connectErr.Code() == connect.CodeUnauthenticated {
+			return fmt.Errorf("Not currently logged in for %s", remote)
+		}
+		return err
+	}
+	user := currentUserResponse.Msg.User
+	if user == nil {
+		return errors.New("No user found for provided token")
```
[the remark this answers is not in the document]
doriable: This mirrors the error that is returned from `buf registry login` when we check the token:
https://github.com/bufbuild/buf/blob/81698bbe0296d70ed886338aa943c28398c53f56/private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go#L212-L214

We can prompt the user to get a new token using `buf registry login`. If that makes sense, then I can update the error message on both sides.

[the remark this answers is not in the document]
bufdev: I didn't provide a token though. Like as a user, this command does not accept a token, so it doesn't make sense to tell the user that their provided token didn't result in a user.

I get that this is the error message for the other one, but this just means both need to be updated, and should in no way reference tokens, as the user does not interact with tokens (typically) via either this command of `buf registry login`.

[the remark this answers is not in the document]
doriable: That is fair -- I left the `buf registry login` error as-is, since it's part of a validation flow where the token is provided through the command. For this one, I provided instructions to the user to refresh their credentials and to check the env var if set.

private/buf/cmd/buf/command/registry/whoami/whoami.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		if connectErr := new(connect.Error); errors.As(err, &connectErr) && connectErr.Code() == connect.CodeUnauthenticated {
+			return fmt.Errorf("Not currently logged in for %s.", remote)
+		}
+		return err
+	}
+	user := currentUserResponse.Msg.User
+	if user == nil {
+		return fmt.Errorf(
+			`No valid user found for login credentials. Run %q to refresh your credentials. If you have %s environment variable set, ensure that the token is valid.`,
```
bufdev, line 106: - "If you have BUF_TOKEN environment variable set" isn't proper english, I believe you mean "If you have the BUF_TOKEN environment variable set".
- "Run 'buf registry login' to refresh your credentials isn't correct". If someone runs "buf registry whoami acme.buf.dev", they should be instructed to run "buf registry login acme.buf.dev".
- Change "No valid user found for login credentials" to "No user is logged in to %s", with the hostname being put in there.

emcfarlane: Errors are now reported as:
```
$ buf registry whoami
Failure: No user is logged in to buf.build. Run "buf registry login" to refresh your credentials. If you have the BUF_TOKEN environment variable set, ensure that the token is valid.
exit status 1
$ buf registry whoami bufbuild.internal
Failure: No user is logged in to bufbuild.internal. Run "buf registry login bufbuild.internal" to refresh your credentials. If you have the BUF_TOKEN environment variable set, ensure that the token is valid.
exit status 1
```

private/buf/cmd/buf/command/registry/whoami/whoami.go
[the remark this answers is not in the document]
bufdev: Still waiting on this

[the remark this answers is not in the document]
doriable: For `buf registry login`, the error message comes up after the token is provided through the login process, so it should still make sense... in the case of `buf registry login`, it might make sense to report it as a syserror, since the token is provided by the user.

bufdev (approved):

PR #3424, opened 2024-10-28
Fix LSP leading comment formatting

### LINKED ISSUE #3423: `buf beta lsp` formatting duplicates leading comments in file

https://github.com/stefanvanburen/buf-repros

### Commands

```shell
cd ./lsp-duplicate-leading-comments
nvim test.proto
# format the file with the LSP
```

### Output

Results in a duplicate leading comment, diff:

```diff
diff --git a/lsp-duplicate-leading-comments/test.proto b/lsp-duplicate-leading-comments/test.proto
index f383f8e..8e97aa1 100644
--- a/lsp-duplicate-leading-comments/test.proto
+++ b/lsp-duplicate-leading-comments/test.proto
@@ -1,4 +1,5 @@
 // Leading comment.
+// Leading comment.
 syntax = "proto3";
 
 package TODO;
```

### Expected Output

No diff.

### Anything else?

The issue seems to lie in using the [fileNode in the range returned in the TextEdit](https://github.com/bufbuild/buf/blob/9ec774500bdc9f6e90f5836f15a147168b30a45a/private/buf/buflsp/server.go#L220); the AST doesn't count the leading comments when [calculating `NodeInfo.Start()`](https://github.com/bufbuild/buf/blob/9ec774500bdc9f6e90f5836f15a147168b30a45a/private/buf/buflsp/symbol.go#L875-L878), while the formatter touches the entire file.

stefanvanburen:
Fixes #3423.

### DIFF
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -215,9 +215,22 @@ func (s *server) Formatting(
 		return nil, err
 	}
 
+	nodeInfo := file.fileNode.NodeInfo(file.fileNode)
 	return []protocol.TextEdit{
 		{
-			Range:   infoToRange(file.fileNode.NodeInfo(file.fileNode)),
+			Range: protocol.Range{
+				// The formatter always returns the entire file; make sure we
+				// represent the edit as the entire range (which may not match up to
+				// the file's NodeInfo.Start()).
+				Start: protocol.Position{
+					Line:      0,
+					Character: 0,
+				},
+				End: protocol.Position{
+					Line:      uint32(nodeInfo.End().Line) - 1,
+					Character: uint32(nodeInfo.End().Col) - 1,
+				},
+			},
 			NewText: out.String(),
 		},
 	}, nil
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -215,20 +215,30 @@ func (s *server) Formatting(
 		return nil, err
 	}
 
-	nodeInfo := file.fileNode.NodeInfo(file.fileNode)
+	// XXX: The current compiler does not expose a span for the full file. Instead of
+	// potentially undershooting the correct span (which can cause comments at the
+	// start and end of the file to be duplicated), we instead manually count up the
+	// number of lines in the file. This is comparatively cheap, compared to sending the
+	// entire file over a domain socket.
+	var lastLine, lastLineStart int
+	for i := 0; i < len(file.text); i++ {
+		// NOTE: we are iterating over bytes, not runes.
+		if file.text[i] == '\n' {
+			lastLine++
+			lastLineStart = i + 1 // Skip the \n.
+		}
+	}
+	lastChar := len(file.text[lastLineStart:]) - 1 // Bytes, not runes!
 	return []protocol.TextEdit{
 		{
 			Range: protocol.Range{
-				// The formatter always returns the entire file; make sure we
-				// represent the edit as the entire range (which may not match up to
-				// the file's NodeInfo.Start()).
 				Start: protocol.Position{
 					Line:      0,
 					Character: 0,
 				},
 				End: protocol.Position{
-					Line:      uint32(nodeInfo.End().Line) - 1,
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/server.go
```diff
@@ -215,9 +215,22 @@ func (s *server) Formatting(
...
 	}
 
+	nodeInfo := file.fileNode.NodeInfo(file.fileNode)
 	return []protocol.TextEdit{
 		{
-			Range:   infoToRange(file.fileNode.NodeInfo(file.fileNode)),
+			Range: protocol.Range{
+				// The formatter always returns the entire file; make sure we
+				// represent the edit as the entire range (which may not match up to
+				// the file's NodeInfo.Start()).
```
mcy, line 224: This comment doesn't actually explain what's going on here precisely. The way I would propose you do this is as follows:
```go
range_ := infoToRange(file.FileNode.NodeInfo(file.fileNode))
// NOTE: This is working around a limitation with the current compiler's parser. There are
// cases where leading comments are not included in the overall file node's span. To remedy
// this, we flatten 
range_.Start.Line = 0
range_.Start.Character = 0
```

private/buf/buflsp/server.go
```diff
@@ -215,9 +215,22 @@ func (s *server) Formatting(
...
+				Start: protocol.Position{
+					Line:      0,
+					Character: 0,
+				},
+				End: protocol.Position{
+					Line:      uint32(nodeInfo.End().Line) - 1,
+					Character: uint32(nodeInfo.End().Col) - 1,
+				},
+			},
```
mcy, line 233: This is still potentially incorrect, due to trailing comments.

Instead, given that parsing is a relatively intense operation, I would prefer that we simply count the number of
lines in file.fileNode and the number of characters on the last line, and construct a Range using that. Don't bother with the spans in the AST, since they clearly don't reflect what we actually mean.

Also, include the following comment:
```go
// XXX: The current compiler does not expose a span for the full file. Instead of
// potentially undershooting the correct span (which can cause comments at the
// start and end of the file to be duplicated), we instead manually count up the
// number of lines in the file. This is comparatively cheap, compared to sending the
// entire file over a domain socket.
```

Also, please file a bug against protocompile to include a way to retrieve a span of the whole file.

mcy (commented):
Thanks for catching this. I think have a minor quibble about how this is resolved, though. Please see my comment.

private/buf/buflsp/server.go
stefanvanburen: FWIW, it doesn't seem like the issue extends to trailing comments, maybe because [`NodeInfo.End()` has special handling for EOF](https://pkg.go.dev/github.com/bufbuild/protocompile/ast#NodeInfo.End)?

> I would prefer that we simply count the number of lines in file.fileNode and the number of characters on the last line, and construct a Range using that.

I was looking around at the FileNode APIs and didn't find much exposed that would help with that; mind pointing me to how that would be done / make that change yourself? (Happy for you to take this over; my `protocompile` knowledge is weak 😅.)

mcy: I'm juggling a bunch of PRs already and am about to go on vacation, so I prefer you do it if only to make sure it gets done. Also, good excuse to grow your knowledge!

Anyways, the right way to do this:
```go
var lastLine, lastLineStart int
for i := 0; i < len(file.text); i++ {
  // NOTE: we are iterating over bytes, not runes.
  if file.text[i] == '\n' {
    lastLine++
    lastLineStart = i + 1 // Skip the \n.
  }
}
lastChar := len(file.text[lastLineStart:]) - 1 // Bytes, not runes!
```

stefanvanburen: https://github.com/bufbuild/buf/pull/3424/commits/0057f0db8ca9b51227164c886a6338c50ea2fece, PTAL!

stefanvanburen: > Also, please file a bug against protocompile to include a way to retrieve a span of the whole file.

https://github.com/bufbuild/protocompile/issues/360!

PR #3430, opened 2024-10-29
Add audit logging resource and events for check plugin

cyinma:
this add the necessary audit logging types and events for the check plugin works


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -115,6 +116,8 @@ enum EventType {
   EVENT_TYPE_USER_AUTO_MERGED_FROM_NEW_IDP = 37;
   EVENT_TYPE_DEVICE_AUTHORIZATION_GRANT_APPROVED = 41;
   EVENT_TYPE_DEVICE_AUTHORIZATION_GRANT_DENIED = 42;
+  EVENT_TYPE_CHECK_PLUGIN_CREATED = 43;
```
mfridman, line 119: Why wouldn't this be a "plugin created" and then on the payload plugin created you have a "type" for the plugin?

bufdev (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
cyinma: updated, we will be reusing the resource and event type of alpha plugin then

bufdev (approved):

mfridman (approved):

PR #3431, opened 2024-10-29
Add bufplugin package 

emcfarlane:
This adds a new package, `bufplugin`, which implements types and functions for operating on Buf plugins. It is similar to the `bufmodule` package for modules.

Parsing of full names and refs has been split out into a separate package from `bufmodule` called `bufparse`. This exposes two types `FullName` and `Ref` which replace `ModuleFullName` and `ModuleRef` and is used by both. Methods `ModuleFullName() ModuleFullName` and `ModuleRef() ModuleRef` have been renamed to `FullName() bufparse.FullName` and `Ref() bufparse.Ref`.

The error type `ParseError` returned from `Parse.*` methods have been moved to the single type `bufparse.ParseError` updating the packages `bufcas`, `bufmodule` and `bufplugin`.


private/bufpkg/bufplugin/parse.go
```diff
@@ -0,0 +1,83 @@
+// Copyright 2020-2024 Buf Technologies, Inc.
```
bufdev, line 1: I thought this stuff was being pulled out to a common `bufparse` package? Also would likely result in some of `errors.go` being pulled out (and potentially other stuff).

private/bufpkg/bufplugin/plugin_full_name.go
```diff
@@ -0,0 +1,132 @@
...
+func NewPluginFullName(
+	registry string,
+	owner string,
+	name string,
+) (PluginFullName, error) {
+	return newPluginFullName(registry, owner, name)
+}
+
+// ParsePluginFullName parses a PluginFullName from a string in the form "registry/owner/name".
+func ParsePluginFullName(pluginFullNameString string) (PluginFullName, error) {
```
bufdev, line 50: Assume this would be pulled out to a `ParseFullName` for example

private/bufpkg/bufplugin/parse.go
emcfarlane: I did look at pulling it out into `bufparse` but reverted in 477557f. I noticed we already had another `ParseError` type in `bufcas` so thought copying was more consistent. Will refactor back out to a package.

private/bufpkg/bufplugin/plugin_full_name.go
emcfarlane: This has now been pulled out into `bufparse.ParseFullName` and returns a type of `bufparse.FullName`. Updated the PR docs.

private/buf/buffetch/internal/module_ref.go
```diff
@@ -46,15 +46,15 @@ func newModuleRef(
 	if strings.Contains(path, "://") {
 		return nil, NewInvalidPathError(format, path)
 	}
-	moduleRef, err := bufmodule.ParseModuleRef(path)
+	moduleRef, err := bufparse.ParseRef(path)
 	if err != nil {
 		// TODO: this is dumb
 		return nil, NewInvalidPathError(format, path)
```
doriable, line 52: `bufparse.ParseRef` only returns an error from the util `parseRefComponents`, which only ever returns a `ParseError`. To keep it consistent for now, we should continuing to return the invalid path error, but in the future, we should consider returning `ParseError` where appropriate from `buffetch`.

private/bufpkg/bufplugin/digest.go
```diff
@@ -0,0 +1,214 @@
...
+
+// DigestType is a type of digest.
+type DigestType int
+
+// ParseDigestType parses a DigestType from its string representation.
+//
+// This reverses DigestType.String().
+//
+// Returns an error of type *bufparse.ParseError if the string could not be parsed.
+func ParseDigestType(s string) (DigestType, error) {
```
doriable, line 58: Had an offline conversation about whether `DigestType` should also be a shared package (e.g. `bufpkg/bufdigest`), since this is also duplicated across modules and plugins. The respective keys (e.g. `ModuleKey` and `PluginKey`) types would control for the digest type at the time of construction. Our conclusion is that this makes sense, but should it be done as a follow-up PR given the size of the migration of `Ref`/`FullName`?

private/bufpkg/bufmodule/digest.go
```diff
@@ -75,15 +76,15 @@ func (d DigestType) String() string {
 //
 // This reverses DigestType.String().
 //
-// Returns an error of type *ParseError if thie string could not be parsed.
+// Returns an error of type *bufparse.ParseError if the string could not be parsed.
 func ParseDigestType(s string) (DigestType, error) {
```
doriable, line 80: See comment in `bufplugin/digest.go`: we considered splitting digest into its own shared package. Looking at this, it does make sense, but maybe it should be done in a follow-up.

private/bufpkg/bufplugin/commit.go
```diff
@@ -0,0 +1,72 @@
...
+// NewCommit returns a new Commit.
+func NewCommit(
+	pluginKey PluginKey,
+	getCreateTime func() (time.Time, error),
+) Commit {
+	return newCommit(
+		pluginKey,
+		getCreateTime,
+	)
+}
```
doriable, line 41: Similar to digest, it might make sense to split `Commit` into its own package as well. Same question/concern around whether that should be a follow-up PR.

doriable (commented):
Overall, this change seems very reasonable -- it's more of a refactor to unifying the types and then adding the plugin types. We should discuss whether it makes sense to have digest and commit as shared types instead also. There could even be an argument to be made for `Key` to be a single interface that `bufmodule` and `bufplugin` each implement with its own semantics (e.g. restrictions on digest types, etc.).

doriable (approved):

PR #3435, opened 2024-10-30
Avoid returning TextEdit on empty diff

stefanvanburen:
In the case that the formatted text is the same as the existing text in the document, we shouldn't return a TextEdit. (This manifests in needing multiple undo operations in an editor to get back to a change, because the TextEdit is registered as a change to the document despite not changing the text.)

### DIFF
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -215,6 +215,12 @@ func (s *server) Formatting(
 		return nil, err
 	}
 
+	newText := out.String()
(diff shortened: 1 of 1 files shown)


stefanvanburen:
[Relevant protocol docs](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_formatting) (I think we're on 3.16 from the protocol package).

image

doriable (approved):
This looks good to me!

PR #3436, opened 2024-10-30
Improve LSP diagnostic info

stefanvanburen:
This removes the extraneous information from the diagnostic report (filename, line, column), which is already included in the range.

It also adds a hard-coded `buf` as the source of the diagnostic.

Before:

image

After:

image

### DIFF
--- a/private/buf/buflsp/report.go
+++ b/private/buf/buflsp/report.go
@@ -17,8 +17,6 @@
 package buflsp
 
 import (
-	"fmt"
-
 	"github.com/bufbuild/protocompile/linker"
 	"github.com/bufbuild/protocompile/parser"
 	"github.com/bufbuild/protocompile/reporter"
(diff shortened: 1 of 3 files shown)


private/buf/buflsp/report.go
```diff
@@ -79,7 +77,8 @@ func newDiagnostic(err reporter.ErrorWithPos, isWarning bool) protocol.Diagnosti
 		// essentially a bug that will result in worse diagnostics until fixed.
 		Range:    protocol.Range{Start: pos, End: pos},
 		Severity: protocol.DiagnosticSeverityError,
-		Message:  fmt.Sprintf("%s:%d:%d: %s", err.GetPosition().Filename, err.GetPosition().Line, err.GetPosition().Col, err.Unwrap().Error()),
+		Message:  err.Unwrap().Error(),
+		Source:   "buf",
```
doriable, line 81: I think this change is reasonable, but I believe one of the reasons we print out the position information is because this is also printed into the console of VSCode, for example, so we want more information. But agree that in-line, this is clearer.

stefanvanburen: yeah, I figured different clients might have different displays. Definitely open to figuring out what makes sense for most clients.

I will note though that adding the position information is inconsistent with some of our other diagnostics (e.g. [our `buf lint` diagnostics](https://github.com/bufbuild/buf/blob/4b587f3b86c3a942adf7cce9ad172278bea3032d/private/buf/buflsp/file.go#L605-L618)):

image

emcfarlane: Go lsp doesn't include line diagnostics, so this makes sense to me.

stefanvanburen: given @doriable's 👍 to consistently using `buf-lsp` as a source, did that in https://github.com/bufbuild/buf/pull/3436/commits/d2c6441e105572ca8c08b697a5c4b12ca9b25662.

I've seen different LSPs do different things, e.g. I think `gopls` uses the individual [analyzer name](https://github.com/golang/tools/blob/master/gopls/doc/analyzers.md) as the source, but I think starting consistently with `buf-lsp` is better for now.

emcfarlane (approved):

PR #3441, opened 2024-10-31
Drop go.uber.org/multierr in favor of errors.Join

stefanvanburen:
Fairly mechanical update to use `errors.Join` in all of the situations in which we were previously using multierr.


private/buf/bufgen/generator.go
```diff
@@ -296,8 +295,12 @@ func (g *generator) execPlugins(
 		jobs,
 		thread.ParallelizeWithCancelOnFailure(),
 	); err != nil {
-		if errs := multierr.Errors(err); len(errs) > 0 {
-			return nil, errs[0]
+		// Handle errors joined with `errors.Join`, which is documented to fulfill this interface.
+		if unwrappableErr, ok := err.(interface{ Unwrap() []error }); ok {
```
stefanvanburen, line 299: Think I'm preserving the original intent here, but worth a look.

Ref: https://pkg.go.dev/errors#Join

private/pkg/oauth2/client_test.go
```diff
@@ -106,7 +106,9 @@ func TestRegisterDevice(t *testing.T) {
 			})
 			output, err := c.RegisterDevice(ctx, test.input)
 			assert.Equal(t, test.output, output)
-			assert.Equal(t, err, test.err)
+			if test.err != nil {
+				assert.EqualError(t, err, test.err.Error())
+			}
```
stefanvanburen, line 111: adjusted these tests slightly; using `Equal` on a joined error doesn't work in the same way. It seemed more important for these tests to assert that the output matches.

private/buf/bufgen/generator.go
bufdev: I'm actually not sure what the intent was here at all, I assume `errors.Join(errors.New("foo"))` is equivalent (in type and string value) to `errors.New("foo")`, is this right?

stefanvanburen: string value: yes, type: no; even in the single case, using `errors.Join` [will wrap the error in a `joinError` value](https://cs.opensource.google/go/go/+/refs/tags/go1.23.2:src/errors/join.go;l=29), hence the updates to the tests below.

stefanvanburen: Going back to why this bit of code is grabbing the first error, it looks like we changed from an `errgroup` impl in https://github.com/bufbuild/buf/pull/621/files#diff-7dd1b326e69fcfcd1514d864d17e1d392cc61affac23277063baa058b6dee82dR215; [Wait() returns the first error](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.Wait).

bufdev: Ah. I did two things:

- I made it so that `errors.Join` is not called by `thread.Parallelize` unless there really is more than one error. I am pretty annoyed that `errors.Join` doesn't do this by default, I am sure there was some big-brained reasoning here (probably along the lines of "well you should always have a distinct error type if you call `errors.Join`) but it doesn't make sense to me, and multierr didn't do this. Alas, not our problem. But might as well clean this up for the `thread.Parallelize` case, where it's pretty common that we'll only have one error.
- I deleted the special handling here. I don't see why it's necessary, it's likely just a historical relic.

dylan-bourque:
One gotcha I've run into with `errors.Join()` is that calling `.Error()` on the returned `error` will **always** inject newlines between the joined errors.  It's usually not an issue, but there are things out there that would very much prefer single-line errors.

Consider calling out the potential behavior change.

private/pkg/thread/thread.go
```diff
@@ -91,19 +90,19 @@ func Parallelize(ctx context.Context, jobs []func(context.Context) error, option
...
 		case <-ctx.Done():
 			stop = true
-			retErr = multierr.Append(retErr, ctx.Err())
+			errs = append(errs, ctx.Err())
 		case semaphoreC <- struct{}{}:
 			select {
 			case <-ctx.Done():
 				stop = true
-				retErr = multierr.Append(retErr, ctx.Err())
+				errs = append(errs, ctx.Err())
```
emcfarlane, line 98: Unlikely, but I think `errs` and above require wrapping with the lock to avoid racing with jobs write below.

bufdev: yea this is a pretty bad miss, this was true before this PR too, will change

bufdev:
I know I requested this, but I'm actually starting to wonder if this is strictly worse and whether we should just not do this. I want to standardize on stdlib as much as anyone, but multierr isn't hurting anyone and (to my estimation) is done significantly better than errors.Join (no random wrapping of single errors, no usage of newlines). Thoughts @mfridman @stefanvanburen @doriable

mfridman:
I've made peace with `errors.Join` and for the times when you want precise control over the formatted error, you have options:

- Use `fmt.Errorf("%w: %w", err1, err2)` which supports multiple `%w` verbs
- At the very top of the chain unwrap the errors and format, e.g.,

```go
err1 := sql.ErrNoRows
err2 := errors.New("call 2 failed")
err3 := errors.New("call 3 failed")
var err4 error

all := errors.Join(err1, err2, err3, err4)
if joined, ok := all.(interface{ Unwrap() []error }); ok {
	fmt.Println(formatErrors(joined.Unwrap()))
	// sql: no rows in result set: call 2 failed: call 3 failed
}
```

It's not very convenient, but most of the time, it's probably good enough. I wish in the original proposal they kept the option for the caller to supply the separator, or at the very least allow for customization via `errors.SetSeparator`.

stefanvanburen:
no strong feeling, mostly in favor for stdlib & consistency with our other repos.

dylan-bourque:
I've also just accepted that you have to use `Unwrap() []error` for precise formatting.  It's unfortunate, but we have to use what we have. 🤷‍♂️

fwiw I'm in favor of standardizing on stdlib packages anywhere possible.

private/pkg/thread/thread.go
bufdev: fixed

bufdev:
Alright, let's go for it then. I'll play along :-)

bufdev (approved):

PR #3443, opened 2024-11-01
Drop go.uber.org/atomic in favor of sync/atomic

stefanvanburen:
The primitives in sync/atomic are not as good as those in uber/atomic; happy to drop this.

### DIFF
--- a/private/pkg/storage/limit.go
+++ b/private/pkg/storage/limit.go
@@ -16,8 +16,7 @@ package storage
 
 import (
 	"context"
-
-	"go.uber.org/atomic"
+	"sync/atomic"
 )
 
 // LimitWriteBucket returns a [WriteBucket] that writes to [writeBucket]
@@ -42,7 +41,6 @@ type limitedWriteBucket struct {
 func newLimitedWriteBucket(bucket WriteBucket, limit int64) *limitedWriteBucket {
 	return &limitedWriteBucket{
 		WriteBucket: bucket,
-		currentSize: atomic.NewInt64(0),
 		limit:       limit,
 	}
 }
@@ -78,7 +76,7 @@ func (o *limitedWriteObjectCloser) Write(p []byte) (int, error) {
 	writeSize := int64(len(p))
 	newBucketSize := o.bucketSize.Add(writeSize)
 	if newBucketSize > o.limit {
-		o.bucketSize.Sub(writeSize)
+		o.bucketSize.Add(-writeSize)
 		return 0, &errWriteLimitReached{
 			Limit:       o.limit,
 			ExceedingBy: newBucketSize - o.limit,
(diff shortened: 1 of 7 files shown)


private/pkg/storage/limit.go
```diff
@@ -42,7 +41,6 @@ type limitedWriteBucket struct {
 func newLimitedWriteBucket(bucket WriteBucket, limit int64) *limitedWriteBucket {
 	return &limitedWriteBucket{
 		WriteBucket: bucket,
-		currentSize: atomic.NewInt64(0),
```
stefanvanburen, line 45: zero value is 0: https://pkg.go.dev/sync/atomic#Int64

private/pkg/storage/storageos/bucket.go
```diff
@@ -404,6 +404,25 @@ func (w *writeObjectCloser) Close() error {
...
+}
+
+// Load loads the stored error.
+func (e *onceError) Load() error {
+	err, ok := e.err.Load().(error)
+	if !ok {
+		return nil
+	}
+	return err
+}
```
stefanvanburen, line 424: took this from a bufstream PR; could use a raw `atomic.Value` inline, but this prevents needing to do the type assertion at the callsite. 🤷.

private/pkg/storage/storageos/bucket.go
```diff
@@ -404,6 +404,25 @@ func (w *writeObjectCloser) Close() error {
...
+// Store stores the err.
+func (e *onceError) Store(err error) {
+	e.err.CompareAndSwap(nil, err)
+}
+
+// Load loads the stored error.
+func (e *onceError) Load() error {
+	err, ok := e.err.Load().(error)
+	if !ok {
+		return nil
```
bufdev, line 421: This seems like a really bad situation - I need to review my Golang a bit, but if this is non-nil, this means that the thing that was stored was not an error on `e.err`, which would be a system error in itself.

stefanvanburen: yeah, I think for safety purposes you could move `onceError` to a different package so someone within this package can't muck with the `e.err` value directly (instead of using `Store`)?

bufdev: Maybe, but that's not really what I'm saying - I'm saying that you are returning `nil` when `ok == false`, and `ok` should only be false if `e.err.Load()` is not of type `error`, which would be a really bad scenario (somehow something other than an `error` got stored on `e.err`. `ok might also be `false` if `e.err.Load()` is `nil`, but this is where my Golang is failing me - I assume that's the only thing that is meant to be captured here.

I would imagine that if you want to do something like this, you need to do:

```go
errIface := e.err.Load()
if errIface == nil {
  return nil
}
err, ok := errIface.(error) 
if !ok {
  return syserror.Newf("expected error but got %T", errIface)
}
return err
```

stefanvanburen: yeah, that looks more technically correct to me - if you somehow stored something that isn't type `error`, you'd always just get `nil` currently. I'll update to your approach.

stefanvanburen: https://github.com/bufbuild/buf/pull/3443/commits/1a41b945067818b0536a484cef27290843a9fe2d - feel free to change, just preferred the `atomicValue` var name.

bufdev: lgtm - I would probably update bufstream as well

bufdev (approved):

Issue #3438, opened 2024-10-31
Formatting with the LSP removes invalid syntax instead of erroring


doriable:
### GitHub repository with your minimal reproducible example (do not leave this field blank or fill out this field with "github.com/bufbuild/buf" or we will automatically close your issue, see the instructions above!)

https://github.com/bufbuild/registry-proto

### Commands

```shell
buf beta lsp

Removing the trailing comma of a field option in an array (https://github.com/bufbuild/registry-proto/blob/81aff0449f6c041834c1d498585237bcd1716c30/buf/registry/module/v1/file.proto), for example, deletes the entire contents of the message rather than producing a diagnostic/error.
```

### Output

```shell
Screenshot of output with empty `File` message included below.
```

### Expected Output

The behaviour of `buf format -w` right now, for example, would be to return a file annotation/error.

![image](https://github.com/user-attachments/assets/6ea86e57-4e8e-4c58-87a2-5e2d57ac9dc2)

We should do the same thing with the LSP, where we return a diagnostic rather than a destructive operation for formatting.

### Anything else?

![image](https://github.com/user-attachments/assets/a2cb653f-db09-4711-9533-19b4c99f6edd)

PR #3447, opened 2024-11-04
Do not format a file that contains an error in the LSP

sprsquish:
Fixes: #3438

Unlike the CLI command, parsing and formatting are separate processes in the LSP. The CLI will fail to parse an invalid file before it gets to the formatter. 

This changes checks for errors in the file's diagnostic info before continuing to the formatter. I checked this locally and verified it will not format a file that contains an error.


private/buf/buflsp/server.go
```diff
@@ -208,6 +208,12 @@ func (s *server) Formatting(
 		return nil, fmt.Errorf("received update for file that was not open: %q", params.TextDocument.URI)
 	}
 
+	for _, diagnostic := range file.diagnostics {
+		if diagnostic.Severity == protocol.DiagnosticSeverityError {
+			return nil, fmt.Errorf("file contains errors")
+		}
+	}
```
doriable, line 215:
```suggestion
  // We check the diagnostics on the file, if there are any build errors, we do not want
  // to format an invalid AST, so we skip formatting and return an error for logging.
	for _, diagnostic := range file.diagnostics {
		if diagnostic.Severity == protocol.DiagnosticSeverityError {
			return nil, fmt.Errorf("cannot format file %q: %s", file.uri.Filename(), diagnostic.Message)
		}
	}
```

We should add a quick explanation of why we are checking the error here and also provide the failure from the diagnostic for logging/debugging purposes.

doriable (commented):
This approach looks good, I have pulled locally and tested. I left a quick comment to provide a little more details in the error and context in the code, but otherwise this is good. Thank you for the submission!

private/buf/buflsp/server.go
sprsquish: > We should add a quick explanation of why we are checking the error here

agreed

> also provide the failure from the diagnostic for logging/debugging purposes.

I'm not sure this would always be helpful. The loop exits as soon as an error is found, but there may be multiple errors. This could follow organizeImport's example and show the first error and a count of how many other errors there are.

doriable: I understand the sentiment behind this, and this makes reasonable sense to me. However, I do think the behaviour here is that any error found would prevent formatting from happening, so it seems maybe extraneous to do a count of the rest of the errors.

My perspective is that in practice, formatting is invoked, and it doesn't happen because we found invalid syntax in the file. The user will still have all the in-line diagnostics for the invalid symbols, and if there are multiple, they are shown all of them in-line.

This error appears in the logs and is here to represent the call-site that stopped any formatting from happening at all, and in this case, we prevented the formatting from happening based on the first diagnostic with an error in the file, so this is representative of what happened. The goal of this is to help debug what is expected behaviour, so it seems reasonable to log the first error message to me.

Perhaps a better message could be:

```
			return nil, fmt.Errorf("file %q was not formatted, error found: %s", file.uri.Filename(), diagnostic.Message)
```

sprsquish: I agree that a user will have inline diagnostics and will want to use those more than errors in a log. I think it's important that there's some sort of explanation. I won't belabor my point beyond this last comment.

My issue with only logging one error without indicating that there are more is that I'll see the error, fix it, try to format, and fail because I had more errors. Logging a count at least lets me know that I have more than the one to fix before the formatter will take action.

If you're unconvinced, I'm happy to concede and log the first error only. If I try to format and look to the logs to see why it's not working, having at least the first error message will clue me in to the idea that the formatter is not going to run if there are any errors in my file and I should fix them all.

doriable: Alright, let's log a count -- there shouldn't be any material impact from that work, and you're right that if someone is debugging, they should be aware that they would need to fix _all_ errors in order to proceed with formatting. How does something like:

```
return nil, fmt.Errorf("cannot format file %q, %v errors found: a file must have no errors for formatting.", file.uri.Filename(), errorCount)
```

Feel free to massage the wording :)

sprsquish: Okay. I pushed a change with a small tweak to the error message. Thanks for the review!

doriable (approved):
Thank you for submitting the change!

PR #3449, opened 2024-11-05
Adjust cache `ModuleDataProvider` to re-retrieve delegate values from store after putting in store

### LINKED ISSUE #3360: `buf beta lsp` panic on empty module cache

https://github.com/bufbuild/registry-proto

### Commands

```shell
buf registry cc
nvim buf/registry/module/v1/module_service.proto
```

### Output

From `:messages`:

```
Client bufls quit with exit code 2 and signal 0. Check log for errors: /Users/stefanvanburen/.local/state/nvim/lsp.log                                                                                                                                                                                                                                                                                                                    
13 lines yanked                                                                                                                                                                                                                                                                                                                                                                                                                           
Press ENTER or type command to continue  
```

From `:LspLog`:

```shell
[START][2024-10-01 10:19:29] LSP logging initiated
[ERROR][2024-10-01 10:19:29] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[34mINFO\27[0m\tnew file version: file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module_service.proto, 0 -> 0\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module_service.proto, 0\n'
[ERROR][2024-10-01 10:19:29] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/module/v1/module.proto, -1\n"
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/src/registry-proto/buf/registry/owner/v1/owner.proto, -1\n"
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tcould not load import import "file://" from disk: %!w(*fmt.wrapError=&{could not read file "file://" from disk: open : no such file or directory 0x140007d9380})\n'
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"\27[34mINFO\27[0m\tparsing AST for file:///Users/stefanvanburen/.cache/buf/v3/wellknowntypes/27.0/google/protobuf/descriptor.proto, -1\n"
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	'\27[33mWARN\27[0m\tnotify returned\t{"method": "$/progress"}\n'
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"panic: runtime error: invalid memory address or nil pointer dereference\n[signal SIGSEGV: segmentation violation code=0x2"
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	" addr=0x30 pc=0x104e58848]\n\ngoroutine 12 [running]:\ngithub.com/bufbuild/buf/private/buf/buflsp.(*symbol).ResolveCrossFile(0x14000c56870, {0x1057e4a58, 0x14000b01bd0})\n\tgithub.com/bufbuild/buf/private/buf/buflsp/symbol.go:322 +0x1098\ngithub.com/bufbuild/buf/private/buf/buflsp.(*file).IndexSymbols(0x140002f04b8, {0x1057e4a58, 0x14000b01bd0})\n\tgithub.com/bufbuild/buf/private/buf/buflsp/file.go:"
[ERROR][2024-10-01 10:19:30] .../vim/lsp/rpc.lua:770	"rpc"	"/opt/homebrew/bin/buf"	"stderr"	"641 +0x3b0\ngithub.com/bufbuild/buf/private/buf/buflsp.(*file).Refresh(0x140002f04b8, {0x1057e4a58, 0x14000b01bd0})\n\tgithub.com/bufbuild/buf/private/buf/buflsp/file.go:209 +0x164\ncreated by github.com/bufbuild/buf/private/buf/buflsp.(*server).DidOpen in goroutine 11\n\tgithub.com/bufbuild/buf/private/buf/buflsp/server.go:176 +0xe0\n"
```

### Expected Output

No `panic`.

### Anything else?

Ideally, the LSP should handle downloading required modules to the cache.

doriable:
This PR changes the caching behaviour so that once remote values
have been retrieved and stored in the cache, we re-retrieve these
values from the cache. This ensures that if the cache is an on-disk
store, then all values will have a `LocalPath` value.

This is needed to fix the LSP when there is no cache.  When the LSP
checks for importable files, it builds the workspace.
Previous to this change, an empty cache results in the remotely fetched
imports to not have `LocalPath`, even though they are put in the cache
after being retrieved. The LSP relies on `LocalPath`, so it breaks.

Tested this locally with my own Neovim + LSP setup.

Fixes #3360


stefanvanburen:
tried this out and works for me locally. Without this + an empty cache, the first open of a file with imports shows a ton of diagnostics. (I think some process in the lsp _does_ eventually populate the cache, because after an edit of the file + a save, the diagnostics go away.)

With this + an empty cache, no diagnostics are shown. Seems like a clear improvement to me.

private/buf/buflsp/file.go
```diff
@@ -45,7 +45,13 @@ import (
 	"google.golang.org/protobuf/reflect/protoreflect"
 )
 
-const descriptorPath = "google/protobuf/descriptor.proto"
+const (
+	descriptorPath = "google/protobuf/descriptor.proto"
+)
+
+var (
+	errFileInfoNoLocalPath = errors.New("file info has no local path")
```
bufdev, line 53: Is this error printed for a user? Even if not, I don't actually know what it means - what does it mean for a file info to not have a path? This should be clearer

private/buf/buflsp/file.go
```diff
@@ -736,6 +743,41 @@ func findImportable(
 	return imports, nil
 }
 
+// getWorkspaceWithLocalPaths checks each proto file in a workspace and ensures they all have
+// a LocalPath. If not, then it refreshes the workspace. As outlined on [storage.ObjectInfo],
+// LocalPath will not always be present if the file did not originate from disk. If an import's
+// module data was fetched remotely, then it will not have a path on disk to the cache. By
+// refreshing the module, and rebuilding the workspace, we populate the module data from the cache.
+func getWorkspaceWithLocalPaths(ctx context.Context, fileName string, lsp *lsp) (bufworkspace.Workspace, error) {
```
bufdev, line 751: Is there a reason this isn't a method on `*lsp`?

private/buf/buflsp/file.go
```diff
@@ -736,6 +743,41 @@ func findImportable(
...
+	workspace, err := lsp.controller.GetWorkspace(ctx, fileName)
+	if err != nil {
+		return nil, err
+	}
+	var refreshWorkspace bool
+	for _, module := range workspace.Modules() {
```
bufdev, line 757: This is very expensive - this means that for every call of `getWorkspaceWithLocalPaths`, you are going to iterate over every file in the workspace. This feels like it would be prohibitive.

private/buf/buflsp/file.go
```diff
@@ -736,6 +743,41 @@ func findImportable(
...
+		if err := module.WalkFileInfos(ctx, func(fileInfo bufmodule.FileInfo) error {
+			if fileInfo.FileType() == bufmodule.FileTypeProto && fileInfo.LocalPath() == "" {
+				return fmt.Errorf("file %q no local path: %w", fileInfo.Path(), errFileInfoNoLocalPath)
```
bufdev, line 760: The error message that would be printed here is `file "foo.proto" no local path: file info has no local path`. I don't know what this means or what to do with it as a user, can you make the error clearer?

private/buf/buflsp/file.go
```diff
@@ -736,6 +743,41 @@ func findImportable(
...
+			if errors.Is(err, errFileInfoNoLocalPath) {
+				lsp.logger.Info(fmt.Sprintf("re-creating workspace: %v", err))
+				refreshWorkspace = true
+				break
+			}
+			return nil, err
+		}
+	}
+	if refreshWorkspace {
+		workspace, err = getWorkspaceWithLocalPaths(ctx, fileName, lsp)
```
bufdev, line 773: I don't understand what is going on here (which means either it needs code clarification, or this function needs to be refactored). It appears you're making a recursive call with no clear base case that will result in recursion terminating. Moreover, each recursive call is O(files).

doriable: I can clarify the comment at the top-level of the function and in-line more information, however this is an accurate description of what is going on. We are trying to ensure that the workspace that is built, for all imports, we can resolve a `LocalPath`. And the only way to check for that is to iterate through each file, there are no other guarantees otherwise.

The exit condition is where all files have a `LocalPath`, which should be the state we converge to once all imports are being resolved through the cache.

This is an expensive operation as it is implemented because of the conditions set by the LSP. In practice, we should have a fully populated cache and module data on-disk fairly quickly, since we cache all module data.

The other option we have is to change the semantics of `ModuleDataProvider`, such that if we resolve remote module data and put those values in the cache after, then we return module data with the cached information.

private/buf/buflsp/file.go
doriable: It is printed in logs, but not directly exposed to the user in-line. I can clarify this error, however it is mostly for logging and debugging.

private/buf/buflsp/file.go
doriable: Yes -- covered some of this in the response to a comment below. If we do pursue changing the semantics `ModuleDataProvider` to return file info after writing to the cache, then we can avoid checking each file since we'll have guarantees to the file on-disk in the cache.

doriable:
@bufdev pulled in the changes today for `ModuleDataStore`. Made some small adjustments:
- checking to make sure there are no unfound keys from the store
- adjusted the wording to specify on-disk cache, since `LocalPath` would still be absent for an in-memory cache, for example

Also updated the PR title and description and tested the LSP (and other `buf` operations) locally.

bufdev (approved):

PR #3450, opened 2024-11-06
Add build for armv7 buf binaries

nimish-ks:
This PR adds coss-compiled build support for linux arm LE `linux armv7l` architecture, used by some of the older ARM devices such as the raspberry pi.

```yml
      - name: setup-qemu
        uses: docker/setup-qemu-action@v3
        id: qemu
        with:
          # alpine image doesn't support linux/riscv64
          platforms: linux/386,linux/amd64,linux/arm64,linux/arm/v7,linux/arm/v6,linux/ppc64le,linux/s390x
```
I can see that `linux armv7` support already exists in the [ci.yml](https://github.com/nimish-ks/buf/blob/main/.github/workflows/ci.yaml#L89) file. Do I need to make changes elsewhere?

This will also fix some downstream issues such as: https://github.com/usememos/memos/issues/3306


make/buf/scripts/release.bash
```diff
@@ -27,6 +27,7 @@ goarch() {
     x86_64) echo amd64 ;;
     arm64) echo arm64 ;;
     aarch64) echo arm64 ;;
+    armv7l) echo arm ;;
```
doriable, line 30:
```suggestion
    armv7) echo arm ;;
```

We want to build for `armv7`, which would incorporate `armv7l` (since the architecture we're building for is `armv7`).

make/buf/scripts/release.bash
```diff
@@ -83,7 +84,7 @@ mkdir -p "${RELEASE_DIR}"
 cd "${RELEASE_DIR}"
 
 for os in Darwin Linux Windows; do
-  for arch in x86_64 arm64; do
+  for arch in x86_64 arm64 armv7l; do
```
doriable, line 87:
```suggestion
  for arch in x86_64 arm64 armv7; do
```

Same change as above.

make/buf/scripts/release.bash
```diff
@@ -100,15 +101,20 @@ for os in Darwin Linux Windows; do
       buf \
       protoc-gen-buf-breaking \
       protoc-gen-buf-lint; do
-      CGO_ENABLED=0 GOOS=$(goos "${os}") GOARCH=$(goarch "${arch}") \
-        go build -a -ldflags "-s -w" -trimpath -buildvcs=false -o "${dir}/bin/${binary}${extension}" "${DIR}/cmd/${binary}"
+      if [ "${arch}" == "armv7l" ]; then
+        CGO_ENABLED=0 GOOS=$(goos "${os}") GOARCH=$(goarch "${arch}") GOARM=7 \
+          go build -a -ldflags "-s -w" -trimpath -buildvcs=false -o "${dir}/bin/${binary}${extension}" "${DIR}/cmd/${binary}"
+      else
```
doriable, line 107:
```suggestion
      if [ "${arch}" == "armv7" ]; then
        if [ "${os}" == "Linux" ]; then
          CGO_ENABLED=0 GOOS=$(goos "${os}") GOARCH=$(goarch "${arch}") GOARM=7 \
            go build -a -ldflags "-s -w" -trimpath -buildvcs=false -o "${dir}/bin/${binary}${extension}" "${DIR}/cmd/${binary}"
        else
          continue
        fi 
      else
```

We only want to build `armv7` for Linux.

make/buf/scripts/release.bash
```diff
@@ -141,7 +147,7 @@ for os in Darwin Linux; do
 done
 
 for os in Windows; do
-  for arch in x86_64 arm64; do
+  for arch in x86_64 arm64 armv7l; do
```
doriable, line 150:
```suggestion
  for arch in x86_64 arm64; do
```

Since we're only building for `armv7` for Linux, this isn't necessary.

make/buf/scripts/release.bash
```diff
@@ -100,15 +101,20 @@ for os in Darwin Linux Windows; do
...
+          go build -a -ldflags "-s -w" -trimpath -buildvcs=false -o "${dir}/bin/${binary}${extension}" "${DIR}/cmd/${binary}"
+      fi
       cp "${dir}/bin/${binary}${extension}" "${binary}-${os}-${arch}${extension}"
     done
   done
 done
 
 for os in Darwin Linux Windows; do
-  for arch in x86_64 arm64; do
+  for arch in x86_64 arm64 armv7l; do
```
doriable, line 117:
```suggestion
  for arch in x86_64 arm64 armv7; do
    if [ "${arch}" == "armv7" ] && [ "${os}" != "Linux" ]; then
      continue
    fi
```

Since we're only building for Linux, we'll need to add that condition here also.

make/buf/scripts/release.bash
```diff
@@ -128,7 +134,7 @@ for os in Darwin Linux Windows; do
 done
 
 for os in Darwin Linux; do
-  for arch in x86_64 arm64; do
+  for arch in x86_64 arm64 armv7l; do
```
doriable, line 137:
```suggestion
  for arch in x86_64 arm64 armv7; do
    if [ "${arch}" == "armv7" ] && [ "${os}" != "Linux" ]; then
      continue
    fi
```

Same as above, only building for Linux.

doriable (commented):
Thank you submitting these changes! I left some comments for changes that would be needed. I applied these changes locally and tested out the build, and got `armv7` binaries to build.

If these changes are merged, this will not be in effect until the next release and would also not be applied to old releases, so in order for downstream services like `memos` to pick this up, they would need to update to the release when it becomes available.

And just making sure I've addressed any open questions:

> I can see that linux armv7 support already exists in the [ci.yml](https://github.com/nimish-ks/buf/blob/main/.github/workflows/ci.yaml#L89) file. Do I need to make changes elsewhere?

This is for our Docker images and should not affect your use-case. However, based on what I'm seeing in `memos`, they use `buf` to manage the generated code for their proto API definitions and they manage `buf` through `npm/pnpm`. So we'll also need to update the binaries available to our NPM packages if this merges.

make/buf/scripts/release.bash
nimish-ks: Yes, makes sense. My mistake.

nimish-ks:
> If these changes are merged, this will not be in effect until the next release and would also not be applied to old releases, so in order for downstream services like memos to pick this up, they would need to update to the release when it becomes available.
I understand. Looking at the pace of the last few release cycles (few weeks to a month), I don't believe It would be a long wait. :)

> This is for our Docker images and should not affect your use-case. 
Cool.

> However, based on what I'm seeing in memos, they use buf to manage the generated code for their proto API definitions and they manage buf through npm/pnpm. So we'll also need to update the binaries available to our NPM packages if this merges.
Yes, thought so too. It it this one? https://www.npmjs.com/package/@bufbuild/protobuf

doriable (commented):
> Yes, thought so too. It it this one? https://www.npmjs.com/package/@bufbuild/protobuf

Yep, that's the one! Once we merge this, I'll update the NPM package publishing process to incorporate these changes for the next release, and then `memos` and similar projects should be able to update seamlessly.

Before I approve, could you please pull in the latest `main` and merge to the PR?

nimish-ks:
> Before I approve, could you please pull in the latest main and merge to the PR?

I believe I already have in - [1633130](https://github.com/bufbuild/buf/pull/3450/commits/1633130fa5dbd625c166a4a9e2e416f4a8e71958). Please correct me if I am mistaken.

doriable (approved):
Thank you!

PR #3451, opened 2024-11-06
Add `clean` key to `buf generate` long description

doriable:
This adds the `clean` key to the example `buf.gen.yaml` in
the `buf generate` command long description.

### DIFF
--- a/private/buf/cmd/buf/command/generate/generate.go
+++ b/private/buf/cmd/buf/command/generate/generate.go
@@ -68,6 +68,10 @@ func NewCommand(
     # The valid values are v1beta1, v1 and v2.
     # Required.
     version: v2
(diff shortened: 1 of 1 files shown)


oliversun9 (approved):

private/buf/cmd/buf/command/generate/generate.go
```diff
@@ -68,6 +68,10 @@ func NewCommand(
     # The valid values are v1beta1, v1 and v2.
     # Required.
     version: v2
+    # When clean is set to true, ddelete the directories, zip files, and/or jar files specified in the
```
stefanvanburen, line 71:
```suggestion
    # When clean is set to true, delete the directories, zip files, and/or jar files specified in the
```

PR #3482, opened 2024-11-19
Add `buf plugin update` command

emcfarlane:
This adds a new command `buf plugin update` enabling referencing of remote plugin refs in `buf.yaml` files. Running the command resolves the refs and updates the `buf.lock` file, adding the new section plugins with the resolved digest. Plugins are cached under the `v3/plugins` directory. This will allow for `buf lint` to invoke remote plugins locally enabling sharing of custom plugins, however this isn't currently enabled in this PR.

The trimmed output of `buf plugin update --help`:
```
Update pinned remote plugins in a buf.lock

Fetch the latest digests for the specified plugin references in buf.yaml.

The first argument is the directory of the local module to update.
Defaults to "." if no argument is specified.

Usage:
  buf plugin update <directory> [flags]
```

An example `buf.yaml` showing a module dependency and a remote plugin reference:
```
version: v2
modules:
  - path: proto
deps:
  - buf.build/bufbuild/protovalidate
plugins:
  - plugin: buf.build/myorg/myplugin
```

And the related `buf.lock` file generated when running `buf dep update` and then `buf plugin update`:
```
# Generated by buf. DO NOT EDIT.
version: v2
deps:
  - name: buf.build/bufbuild/protovalidate
    commit: 5a7b106cbb87462d9a8c9ffecdbd2e38
    digest: b5:0f2dc6c9453e9cc9e9f36807aaa5f94022e837d91fef4dcaeed79a35c0843cc64eba28ff077aab24da3b2cb12639ad256246f9f9a36c033b99d5754b19996b7e
plugins:
  - name: buf.build/myorg/myplugin
    commit: 950f0fb649be417db3d7e79720bc32be
    digest: p1:2b767f444522ab7b89fed9ddd960246786dba313e8a0fd612ac75c22a816fde81818500aa132c911df9ede41975bae69dacca876ed967de04adba13d2d3c3ea6
```


doriable (approved):
This all makes sense to me, looks good!

PR #3483, opened 2024-11-20
Fix annotation for FIELD_NOT_REQUIRED rule

doriable:
The annotation for `FIELD_NOT_REQUIRED` did not
pass in the field name.


oliversun9 (approved):
Good catch!

PR #3489, opened 2024-11-22
Add an allowlist of which clients can use the LSP server

mcy:
This PR uses the client's reported identity to allowlist which clients we support, with a breakout flag called `--break-warranty-seal`.

Initially allowlisted are several VSCode variants and nvim, along with comments for where those names came from.

### DIFF
--- a/private/buf/buflsp/buflsp.go
+++ b/private/buf/buflsp/buflsp.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"log/slog"
+	"slices"
 	"sync"
 	"sync/atomic"
 
@@ -40,6 +41,7 @@ import (
 // Returns a context for managing the server.
 func Serve(
 	ctx context.Context,
+	hasWarranty bool,
 	wktBucket storage.ReadBucket,
 	container appext.Container,
 	controller bufctl.Controller,
@@ -65,6 +67,7 @@ func Serve(
 			&connWrapper{Conn: conn, logger: container.Logger()},
 			zap.NewNop(), // The logging from protocol itself isn't very good, we've replaced it with connAdapter here.
 		),
+		hasWarranty: hasWarranty,
 		logger:      container.Logger(),
 		controller:  controller,
 		checkClient: checkClient,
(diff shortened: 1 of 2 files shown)


mcy:
Currently blocked on a decision on the set of clients to include. We intend to merge this but not before we make a decision on this front.

nicksnyder:
I don't think we should touch this PR until we are able to have a convo with the right folks about what we are officially supporting, but when we do come back to this, I think a better name for this flag would be `--allow-unsupported-clients`.

private/buf/buflsp/buflsp.go
```diff
@@ -81,6 +84,27 @@ func Serve(
...
+	// https://github.com/microsoft/vscode/blob/e1de2a458dfb770545489daf499131fd328924e7/product.json#L3
+	"Code - OSS",
+
+	// https://github.com/VSCodium/vscodium/blob/8bcb412ec53f0232e373d9d04b50f804f090ea67/prepare_vscode.sh#L138
+	"VSCodium",
+	"VSCodium - Insiders",
+
+	// https://github.com/neovim/neovim/blob/9a681ad09e2add96d47bf3f39cca8029f3bf09df/runtime/lua/vim/lsp/client.lua#L539
+	"Neovim",
+}
```
jpaju, line 106: Could you also consider whitelisting [Helix editor](https://helix-editor.com/) ?

```suggestion
	// https://github.com/helix-editor/helix/blob/cbbeca6c5227e65bebdbe9abbadbd2202ffc1005/helix-lsp/src/client.rs#L682
	"helix"
}
```

nicksnyder: No matter what we do, we aren't going to try to intentionally break any editor, but our focus is going to be making this language server work with VS Code. This PR just makes it so anyone using a non-VS Code editor has to explicitly understand their editor is not explicitly supported (by explicitly providing a CLI flag). We won't be testing Helix ourselves, so Helix will not be added to this allowlist.

bufdev:
Should we close this now?

doriable:
> Should we close this now?

I think we should close this, I don't think we're intending on restricting the client.

PR #3498, opened 2024-11-27
Add `buf plugin registry commit` commands

emcfarlane:
This adds `buf registry plugin commit {add-label,info,list,resolve}` commands to manage BSR plugin commits.

The trimmed output of `buf plugin registry commit --help`:
```
Manage a plugin's commits

Usage:
  buf registry plugin commit [flags]
  buf registry plugin commit [command]

Available Commands:
  add-label   Add labels to a commit
  info        Get commit information
  list        List plugins commits
  resolve     Resolve commit from a reference
```

### DIFF
--- a/private/buf/bufprint/bufprint.go
+++ b/private/buf/bufprint/bufprint.go
@@ -199,27 +199,37 @@ func PrintEntity(writer io.Writer, format Format, entity Entity) error {
 	}
 }
 
-// NewModuleLabelEntity returns a new label entity to print.
-func NewModuleLabelEntity(label *modulev1.Label, moduleFullName bufparse.FullName) Entity {
+// NewLabelEntity returns a new label entity to print. It takes a label as an
+// interface to allow for modulev1.Label and pluginv1beta1.Label to be passed.
+func NewLabelEntity(label interface {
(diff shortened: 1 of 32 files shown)


private/buf/bufprint/bufprint.go
```diff
@@ -199,8 +199,8 @@ func PrintEntity(writer io.Writer, format Format, entity Entity) error {
 	}
 }
 
-// NewLabelEntity returns a new label entity to print.
-func NewLabelEntity(label *modulev1.Label, moduleFullName bufparse.FullName) Entity {
+// NewModuleLabelEntity returns a new label entity to print.
+func NewModuleLabelEntity(label *modulev1.Label, moduleFullName bufparse.FullName) Entity {
```
doriable, line 203: I know we discussed having something like:

```
func NewLabelEntity(label interface{ GetName() string... etc. }, entityFullName bufparse.FullName)
```

I think this makes sense for both commit and labels to combine module and plugin.

emcfarlane: Switched to common interfaces for entities, same in #3499

doriable (approved):

PR #3507, opened 2024-12-03
Use ordered set for imports to iterate deterministically when filtering an image

### LINKED ISSUE #3506: Using the `type` option results in non-deterministic ordering of file dependencies

https://github.com/haines/buf-dependency-order

### Commands

```shell
# Run `buf build protos --type hub.playground.v1.CerbosService --output descriptors.yaml` until a diff is found
./repro
```

### Output

```diff
diff --git a/descriptors.yaml b/descriptors.yaml
index a9cc4a1..c13d2dd 100644
--- a/descriptors.yaml
+++ b/descriptors.yaml
@@ -54551,8 +54551,8 @@ file:
       - buf/validate/validate.proto
       - cerbos/request/v1/request.proto
       - cerbos/response/v1/response.proto
-      - protoc-gen-openapiv2/options/annotations.proto
       - google/api/field_behavior.proto
+      - protoc-gen-openapiv2/options/annotations.proto
     messageType:
       - name: CheckResourcesRequest
         field:
```

### Expected Output

The file's `dependency` list should be in a deterministic order, so that no diff is produced. This is the case if the `--type` option is omitted.

### Anything else?

Initially reported as https://github.com/bufbuild/protobuf-es/issues/1030

doriable:
Prior to this change, we used a map to keep track of imports
when filtering an image based on type. This resulted in a bug
where iterating through the imports was non-deterministic, and
resulted in non-deterministic images.
This change uses an ordered set for imports instead of a map.

Fixes #3506

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
+++ b/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
@@ -269,23 +269,18 @@ func ImageFilteredByTypesWithOptions(image bufimage.Image, types []string, opts 
 		// Only handle imports and dependencies if there are any.
 		for indexFrom, importPath := range imageFileDescriptor.GetDependency() {
 			path := append(basePath, int32(indexFrom))
-			// If there are no required imports, we mark this as deleted.
-			if importsRequired == nil {
-				sourcePathRemapper.markDeleted(path)
+			// We check if the import path exists among required imports. If yes, we
+			// move and then delete from required imports as we go.
+			if importsRequired != nil && importsRequired.index(importPath) != -1 {
+				sourcePathRemapper.markMoved(path, int32(indexTo))
+				indexFromTo[int32(indexFrom)] = int32(indexTo)
+				imageFileDescriptor.Dependency[indexTo] = importPath
+				indexTo++
+				// delete them as we go, so we know which ones weren't in the list
+				importsRequired.delete(importPath)
 			} else {
-				// Otherwise, we check if the import path exists among required imports. If yes, we
-				// move and then delete from required imports as we go.
-				if i := importsRequired.index(importPath); i != -1 {
-					sourcePathRemapper.markMoved(path, int32(indexTo))
-					indexFromTo[int32(indexFrom)] = int32(indexTo)
-					imageFileDescriptor.Dependency[indexTo] = importPath
-					indexTo++
-					// delete them as we go, so we know which ones weren't in the list
-					importsRequired.delete(importPath)
-				} else {
-					// Path did not exist in required imports, we mark as deleted.
-					sourcePathRemapper.markDeleted(path)
-				}
+				// Path did not exist in required imports, we mark as deleted.
+				sourcePathRemapper.markDeleted(path)
 			}
 		}
 		imageFileDescriptor.Dependency = imageFileDescriptor.Dependency[:indexTo]
@@ -293,6 +288,8 @@ func ImageFilteredByTypesWithOptions(image bufimage.Image, types []string, opts 
 		// Add any other imports (which may not have been in the list because
 		// they were picked up via a public import). The filtered files will not
 		// use public imports.
+		// The imports are added in the order they are encountered when importing
+		// to maintain a deterministic ordering.
 		if importsRequired != nil {
 			imageFileDescriptor.Dependency = append(imageFileDescriptor.Dependency, importsRequired.keys()...)
 		}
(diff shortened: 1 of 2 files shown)


doriable:
Tested this change with the repro provided by the user in the issue filed and it works :)

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -1000,3 +1012,57 @@ func stripSourceRetentionOptionsFromFile(imageFile bufimage.ImageFile) (bufimage
...
+
+// delete removes a key from the index map of ordered imports. Note that the path is not removed
+// from the paths list. If you want to iterate through the paths, use keys() to get all
+// non-deleted keys.
+func (o *orderedImports) delete(path string) error {
+	if _, ok := o.pathToIndex[path]; ok {
+		delete(o.pathToIndex, path)
+		return nil
+	}
+	return fmt.Errorf("no import found for: %s", path)
```
emcfarlane, line 1050: Why error on delete?

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -1000,3 +1012,57 @@ func stripSourceRetentionOptionsFromFile(imageFile bufimage.ImageFile) (bufimage
...
+
+// keys provides all non-deleted keys from the ordered imports.
+func (o *orderedImports) keys() []string {
+	keys := make([]string, 0, len(o.pathToIndex))
+	for _, path := range o.paths {
+		if _, ok := o.pathToIndex[path]; ok {
+			keys = append(keys, path)
+		}
+	}
+	return keys
```
emcfarlane, line 1061: As we always build a new set of keys we could remove the storing the `paths` field in `orderedImports` and build it up with a sort, naming `pathToIndex` as `pathToWeight`, something like:
```go
paths := slices.Collect(maps.Keys(m))
sort.Slice(paths, func(i, j int) bool { return pathWeight[paths[i]] < pathWeight[paths[j]] })
```
Then the delete function doesn't need to be special cased.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -266,27 +266,39 @@ func ImageFilteredByTypesWithOptions(image bufimage.Image, types []string, opts
...
 
 		// Add any other imports (which may not have been in the list because
 		// they were picked up via a public import). The filtered files will not
 		// use public imports.
-		for importPath := range importsRequired {
-			imageFileDescriptor.Dependency = append(imageFileDescriptor.Dependency, importPath)
+		if importsRequired != nil {
+			imageFileDescriptor.Dependency = append(imageFileDescriptor.Dependency, importsRequired.keys()...)
 		}
+
```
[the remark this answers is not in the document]
doriable: It needs to be in order of the dependencies though... not just a sort on the string...

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
doriable: I suppose we don't actually need to error on delete, we can simply have a no-op if attempting to delete on a key that doesn't exist...

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -266,27 +266,39 @@ func ImageFilteredByTypesWithOptions(image bufimage.Image, types []string, opts
...
+					imageFileDescriptor.Dependency[indexTo] = importPath
+					indexTo++
+					// delete them as we go, so we know which ones weren't in the list
+					if err := importsRequired.delete(importPath); err != nil {
+						return nil, err
+					}
+				} else {
+					// Path did not exist in required imports, we mark as deleted.
+					sourcePathRemapper.markDeleted(path)
+				}
```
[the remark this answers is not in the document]
emcfarlane: Still think this could be moved back to:
```go
if importsRequired != nil && importsRequired.index(importPath) != -1 { ... } else { ... }
```
As the branch is the has check but for nil and existence. The value of `i` isn't used. So I think it makes sense to move into one.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -1000,3 +1010,55 @@ func stripSourceRetentionOptionsFromFile(imageFile bufimage.ImageFile) (bufimage
...
+	for _, path := range o.paths {
+		if _, ok := o.pathToIndex[path]; ok {
+			keys = append(keys, path)
+		}
+	}
+	return keys
+}
+
+// newOrderedImports creates a new orderedImports structure.
+func newOrderedImports() *orderedImports {
```
emcfarlane, line 1060: For style the new functions should appear under the struct declaration.

emcfarlane (approved):

PR #3513, opened 2024-12-05
Add configurable base plugin support in buf.plugin.yaml

mfridman:
This PR adds an optional `base_plugin` field to the Go registry config, allowing plugins like gRPC and Connect to use a different base plugin instead of the default `protocolbuffers/go`. This enables Generated SDK scenarios:

1. Multi-plugins to generate correct imports for module dependencies, rather than using the BSR's default `protocolbuffers/go`
2. Custom plugins (like gRPC) to properly integrate with custom multi-plugins for import path generation (both plugin and module dependencies)

Note: A multi-plugin bundles generators like `protoc-gen-go`, `protoc-gen-go-vtproto`, and `protoc-gen-go-json` that output to the same package and are packaged using: https://github.com/bufbuild/tools/tree/main/cmd/protoc-gen-multi

### Example:

```yaml
version: v1
name: remote/org/go-grpc
plugin_version: v1.5.1
output_languages:
  - go
deps:
  - plugin: remote/org/go-multi:v0.1.0
registry:
  go:
    base_plugin: remote/org/go-multi # --- NEW FIELD
    min_version: "1.21"
    deps:
      - module: google.golang.org/grpc
        version: v1.68.1
  opts:
    - paths=source_relative
    - require_unimplemented_servers=false
    - separate_package=true
```


private/bufpkg/bufremoteplugin/bufremotepluginconfig/bufremotepluginconfig.go
```diff
@@ -440,6 +441,7 @@ type ExternalGoRegistryConfig struct {
 		Module  string `json:"module,omitempty" yaml:"module,omitempty"`
 		Version string `json:"version,omitempty" yaml:"version,omitempty"`
 	} `json:"deps,omitempty" yaml:"deps,omitempty"`
+	BasePluginName string `json:"base_plugin_name,omitempty" yaml:"base_plugin_name,omitempty"`
```
mfridman, line 444:
```suggestion
	BasePlugin string `json:"base_plugin,omitempty" yaml:"base_plugin,omitempty"`
```

This might be clearer and more concise. The intended format is remote/owner/plugin

proto/buf/alpha/registry/v1alpha1/plugin_curation.proto
```diff
@@ -125,6 +125,14 @@ message GoConfig {
...
   // The minimum Go version required by the plugin.
   string minimum_version = 2;
+  // Optionally specifies an alternate base plugin like "remote/org/go-multi" instead of the default
+  // "remote/protocolbuffers/go". If provided, the name must match one of the plugin dependencies in
+  // the top-level deps field.
+  //
+  // This affects how imports are resolved - the specified plugin's import path will be used as the
+  // base path for all generated code, replacing the default protocolbuffers/go import paths. Used
+  // when depending on non-default BSR plugins.
+  string base_plugin_name = 3;
```
mfridman, line 135: @srikrsna-buf pointed out this could be a boolean, like `use_current_plugin_as_base`. But that'll only solve item 1. in the description (specifically, module dependencies).

You could also argue if a plugin has no plugin deps, then the default (on the BSR) should be to use the plugin itself as the base (for module deps) instead of `protocolbuffers/go`. Example of petapis -> paymentaips module dep:

```diff
-	v1alpha1 "bufbuild.internal/gen/go/acme/paymentapis/protocolbuffers/go/payment/v1alpha1"
+	v1alpha1 "bufbuild.internal/gen/go/acme/paymentapis/custom-plugins/go/payment/v1alpha1"
```

Either way, I went a step further and made this a configurable string so it satisfies both 1. and 2. Where 2. is module and plugin dependencies.

private/bufpkg/bufremoteplugin/bufremotepluginconfig/bufremotepluginconfig.go
mfridman: Changed in a2edfb7f to "base plugin" terminology.

private/bufpkg/bufremoteplugin/bufremotepluginconfig/bufremotepluginconfig.go
```diff
@@ -440,6 +441,7 @@ type ExternalGoRegistryConfig struct {
 		Module  string `json:"module,omitempty" yaml:"module,omitempty"`
 		Version string `json:"version,omitempty" yaml:"version,omitempty"`
 	} `json:"deps,omitempty" yaml:"deps,omitempty"`
+	BasePlugin string `json:"base_plugin,omitempty" yaml:"base_plugin,omitempty"`
```
mfridman, line 444: An alternative way to configure this might be to expose an `is_base: true` within `deps`, e.g.,

```yaml
deps:
  - plugin: buf.build/library/go:v1.28.0
    is_base: true
```

And allowing exactly one plugin within deps to have this be set to true.

However, I don't like this because it leaks a Generated SDK configuration beyond the registry config.

private/bufpkg/bufremoteplugin/bufremotepluginconfig/config.go
```diff
@@ -260,17 +269,37 @@ func newGoRegistryConfig(externalGoRegistryConfig *ExternalGoRegistryConfig) (*G
...
+		// specified. This ensures there's exactly one base type and it has a known dependency to
+		// generate imports correctly and build a correct Go mod file.
+		if len(pluginDependencies) > 0 {
+			ok := slices.ContainsFunc(pluginDependencies, func(ref bufremotepluginref.PluginReference) bool {
+				return ref.IdentityString() == basePlugin.IdentityString()
+			})
+			if !ok {
+				return nil, fmt.Errorf("base plugin %q not found in plugin dependencies", externalGoRegistryConfig.BasePlugin)
+			}
+		}
```
stefanvanburen, line 297: Is the intention that if there's no plugin dependency, assigning `base_plugin` has no effect? Is there a use-case for assigning it without a matching dependency?

mfridman: Not quite, we still want to support option 1 in the description.

Suppose you have a multi-plugin that slams ALL the generators into a single multi-plugin. In this scenario, there are no plugin deps (all plugins output to the same package), but we still need to generate the correct module imports.

E.g., think of `acme/petapis`, which depends on `acme/paymentapis`. Without specifying the base type, we default to `protocolbuffers/go`

```diff
package petv1

import (
-	v1alpha1 "remote/gen/go/acme/paymentapis/protocolbuffers/go/payment/v1alpha1"
+	v1alpha1 "remote/gen/go/acme/paymentapis/custom-plugins/go/payment/v1alpha1"
```

Not the `.../custom-plugins/go/...`

Likewise we need to build a `go.mod` file that has the correct Go module:

```go
module remote/gen/go/acme/petapis/custom-plugins/go

go 1.21

require (
	remote/gen/go/acme/paymentapis/custom-plugins/go v0.2.0-20241204162507-076bca1034ed.1
```

stefanvanburen: ah, I see, was not thinking about the multi-plugin. That makes sense to me.

stefanvanburen (approved):

private/bufpkg/bufremoteplugin/bufremotepluginconfig/config.go
mfridman: I wonder if the BSR should support option 1 with no further configuration. If there are no plugin dependencies, use the plugin name itself (instead of `protocol buffers/go`) for module dependencies.

Effectively, the same as manually setting just so `protocolbuffers/go` isn't used.

```yaml
version: v1
name: remote/org/go-multi # --- SAME
plugin_version: v1.35.2
output_languages:
  - go
registry:
  go:
    base_plugin: remote/org/go-multi # --- SAME
    min_version: "1.21"
    deps:
      - module: google.golang.org/protobuf
        version: v1.35.2
  opts:
    - paths=source_relative
```

emcfarlane (approved):

srikrsna-buf (approved):

PR #3523, opened 2024-12-09
Add `buf plugin prune` command

doriable:
This PR adds the `buf plugin prune` command, which prunes
plugins that are no longer configured in the `buf.yaml` from the `buf.lock`.

This also fixes a behavior in `buf plugin update`, where we should
only be updating plugin versions and adding new plugins to the `buf.lock`,
and otherwise preserving the contents of the plugins key in the `buf.lock`.

### DIFF
--- a/private/buf/cmd/buf/command/plugin/pluginprune/pluginprune.go
+++ b/private/buf/cmd/buf/command/plugin/pluginprune/pluginprune.go
@@ -1,0 +1,117 @@
+// Copyright 2020-2024 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package pluginprune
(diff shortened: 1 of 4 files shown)


doriable:
Tested with local cluster, module, and WASM check plugins.

private/buf/cmd/buf/command/plugin/pluginupdate/pluginupdate.go
```diff
@@ -131,6 +133,31 @@ func run(
 	if err != nil {
 		return err
 	}
+	// We do not prune the buf.lock during update, we only add update existing plugin keys
+	// or add new plugin keys.
```
emcfarlane, line 137: This doesn't seem consistent with `buf dep update` which will remove items from the lock file if the dep is removed.

doriable: Ah yes -- let me revert change.

private/buf/cmd/buf/command/plugin/pluginprune/pluginprune.go
```diff
@@ -0,0 +1,117 @@
...
+		return err
+	}
+	configuredRemotePluginKeys, err := pluginKeyProvider.GetPluginKeysForPluginRefs(
+		ctx,
+		configuredRemotePluginRefs,
+		bufplugin.DigestTypeP1,
+	)
+	if err != nil {
+		return err
+	}
```
emcfarlane, line 80: Why does this need the API to resolve the keys? I think this should be `workspaceDepManager.ExistingBufLockFileDepModuleKeys(ctx)` compared to `configuredRemotePluginRefs`.

emcfarlane (approved):

PR #3528, opened 2024-12-10
Move registry update commands under settings

emcfarlane:
This moves `buf registry module update` and `buf registry plugin update` commands to `buf registry module settings update` and `buf registry plugin setting update`, respectively. The command `buf registry module update` is marked deprecated. The command `buf registry plugin update` is removed as it was not yet released.


doriable (commented):
All make sense to me, we should probably include this in the changelog :)

doriable (approved):

PR #3530, opened 2024-12-11
Do not set err to unwrapped connectErr in error interceptor

doriable:
This PR removes setting `err = connectErr.Unwrap()` from our
error interceptor after the `connectErr` checks. The reason for this
is that this could create some lossiness with `err`. For example, when
using `thread.ParallelizeWithCancelOnFailure`, the errors are appended
together with `errors.Join`:

https://github.com/bufbuild/buf/blob/7321103f28e0668a434e8370c8bfe4aa9dd2a62b/private/pkg/thread/thread.go#L128

When we call `errors.As(err, connectErr)` in the error interceptor, this matches
the first error in the error tree, in this case the context cancellation, and sets that
as the underlying error. So, when `connectErr.Unwrap()` is called, only the context
cancellation error is returned, and the rest of `err` is lost.

### DIFF
--- a/private/buf/cmd/buf/buf.go
+++ b/private/buf/cmd/buf/buf.go
@@ -471,7 +471,6 @@ func wrapError(err error) error {
 			}
 			return errors.New(msg)
 		}
-		err = connectErr.Unwrap()
 	}
 
 	sysError, isSysError := syserror.As(err)


private/buf/cmd/buf/buf.go
```diff
@@ -471,7 +471,6 @@ func wrapError(err error) error {
 			}
 			return errors.New(msg)
 		}
-		err = connectErr.Unwrap()
```
bufdev, line 474: There's a reason we did this - do you know what that is? I don't want to undo it without knowing the reason. I believe it was related to UX ie the error message printed out in this case, but maybe not.

doriable: So I discussed this with @emcfarlane today, and our thoughts were the reason for this is because `connectErr.Unwrap()` would remove the connect code from the error message and return the underlying error, e.g. `canceled: context canceled` -> `context canceled`. I don't recall any other discussions on this.

The impetus for this change was because there was an issue reported where calling `buf generate` resulted in a `context canceled` error because of `thread.ParallelizeWithCancelOnFailure`. The original `err` would've returned:

```
plugin protoc-gen-es: exec: "protoc-gen-es": executable file not found in $PATH
canceled: context canceled
```

But because we only returned `context canceled`, it was not easy to debug.

bufdev (approved):

PR #3531, opened 2024-12-11
Have `SemanticTokensFull` return `nil` when no symbols are available for the file

doriable:
When a file does not have parsed symbols, we are currently
sending a `nil` slice for the data in semantic tokens.

The specification for `textDocument/semanticTokens/full` outlines that
the expected response is `SemanticTokens | null`, where `SemanticTokens`
is defined as:

```
export interface [SemanticTokens](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokens) {
	/**
	 * An optional result id. If provided and clients support delta updating
	 * the client will include the result id in the next semantic token request.
	 * A server can then instead of computing all semantic tokens again simply
	 * send a delta.
	 */
	resultId?: string;

	/**
	 * The actual tokens.
	 */
	data: [uinteger](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uinteger)[];
}
```

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_semanticTokens

`data` is not an optional field for `SemanticTokens`, so setting it to `nil`
causes failures on clients:

Screenshot 2024-12-11 at 18 23 04

From the LSP logs:
```
DEBUG\tresponding\t{"method": "textDocument/semanticTokens/full", "params": {"data":null}}\n'
```

We should instead be returning `nil` for `SemanticTokensFull` in this case, which
is expected by clients.

LSP logs after the change:

```
DEBUG\tresponding\t{"method": "textDocument/semanticTokens/full", "params": null}\n
```


private/buf/buflsp/server.go
```diff
@@ -411,9 +411,9 @@ func (s *server) SemanticTokensFull(
 	defer progress.Done(ctx)
 
 	var (
-		encoded           []uint32
 		prevLine, prevCol uint32
 	)
+	encoded := []uint32{}
```
bufdev, line 416: Can you point to how this makes a difference? Downstream code should generally not care whether this is 0 or nil. If there's a specific reason, this should be documented here with a code comment so that we understand this choice in the future.

doriable: Updated the PR, title, and description: I did some digging into the specification, and `textDocument/semanticTokens/full` expects `SemanticTokens | null` as the response, but `data` on `SemanticTokens` is not optional. So in this case, if we are not sending any `data`, we should instead be returning `nil` for semantic tokens. I added this explanation as in-line comments in the code change as well.

bufdev (approved):

PR #3533, opened 2024-12-12
Add missing audit events for metadata updates and verification status

unmultimedio:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -208,6 +218,32 @@ message PayloadOrganizationCreated {}
...
+  // The url update that happened to the organization.
+  optional UrlUpdate url_update = 2;
+}
+
+message PayloadOrganizationVerificationStatusChanged {
+  // old_verification_status is the organization verification status before the update.
+  string old_verification_status = 1;
+  // new_verification_status is the organization verification status after the update.
+  string new_verification_status = 2;
+}
```
unmultimedio, line 245: This is in a similar spot like [this other conversation](https://github.com/bufbuild/buf/pull/3522#discussion_r1876499290), the definition of these verification statuses live in `registry-proto` [here](https://github.com/bufbuild/registry-proto/blob/58fb62ad0dc05e274d3d4e85f5c79c1896ed1964/buf/registry/owner/v1/organization.proto#L65), so we don't have a way to reference the right value here.

The options I see are:

1. Log the string value of that `enum` in `registry-proto` here; that doesn't feel right.
2. Add an equivalent `enum` in this pkg, and a parser in the audit log writer that converts from that `registry-proto`'s `enum` to the one in here; it's duplication of types but that doesn't sound too bad I think.
3. Punt auditing this event until we move this pkg to be colocated in `registry-proto`.

This same scenario happens with `PayloadUserVerificationStatusChanged`.

unmultimedio: In general terms we're leaning on auditing any write operation. Even the original proposal was thinking how we could expand on read operations too, but we haven't gone that far yet.

This PR adds metadata (true it may be not super important, just URLs and descriptions), and verification status changes (they sound pretty important to me, and they're not being audited yet).

Let's chat about it.

bufdev (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
bufdev: approving from my side but you both resolve this

unmultimedio:
Chatted internally with the team, closing this PR until/if we decide these events are worth auditing.

PR #3534, opened 2024-12-12
Update protovalidate-go to v0.8.0

jchadwick-buf:
We had a breaking change in protovalidate-go v0.8.0, nothing major. Also updates protoyaml-go to v0.3.0 for the same reason.

Looks like the structured field path API might make it possible to improve this code. That said, it looks pretty carefully constructed, so I chose to do a 1:1 update instead. "First, do no harm."

I'll leave it to the reviewers to decide if I should be more aggressive about trying to refactor the code instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,17 +6,17 @@ toolchain go1.23.3
 
 require (
 	buf.build/gen/go/bufbuild/bufplugin/protocolbuffers/go v1.35.2-20241031151143-70f632351282.1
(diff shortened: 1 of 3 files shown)


doriable (approved):
Thank you for doing this! I've been keeping an eye out on `protovalidate-go` and had this update queued, so this is very helpful!

My original thought was to first do a direct replacement as you've done here, and then consider a refactor after. I think there has been a lot of changes, mostly for the better in terms of ergonomics, since the inception of the lint rule, so there's likely a lot of improvements we could make. I'm going to stamp this first, and I think it's good to merge, and I'll bring this up in our sync today in terms of how we'd like to invest in it :)

PR #3535, opened 2024-12-12
Deprecate default branch changed event

unmultimedio:
Audit log emitters don't use `PayloadRepositoryDefaultBranchChanged` events anymore, they use `PayloadRepositoryDefaultLabelNameChanged` instead.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -382,7 +382,12 @@ message PayloadRepositoryDefaultLabelNameChanged {
   string new_default_label_name = 4;
 }
 
+// PayloadRepositoryDefaultBranchChanged is deprecated.
+//
+// Use PayloadRepositoryDefaultLabelNameChanged instead.
 message PayloadRepositoryDefaultBranchChanged {
+  // deprecated, use PayloadRepositoryDefaultLabelNameChanged instead.
```
bufdev, line 389: Capitalize Deprecated

PR #3537, opened 2024-12-12
Use Workspaces for buf config ls rules

emcfarlane:
This refactors the commands `buf config {ls-lint-rules, ls-breaking-rules}` to utilize Workspaces. Workspaces allow for re-use of the validation for config overrides and provide more information about the source.

### DIFF
--- a/private/buf/bufworkspace/workspace.go
+++ b/private/buf/bufworkspace/workspace.go
@@ -92,6 +92,7 @@ type Workspace interface {
 	// handle v1 vs v2 transparently. Right now, this is only approved to be used in push
 	// when we want to know whether we need to print out only CommitIDs. Any other usages
 	// need to be evaluated.
+	// TODO: Validate the usecase for `buf config ls-lint-rules` and `buf config ls-breaking-rules`.
 	IsV2() bool
 
 	isWorkspace()
--- a/private/buf/cmd/buf/command/config/internal/internal.go
+++ b/private/buf/cmd/buf/command/config/internal/internal.go
@@ -18,15 +18,15 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/fs"
 
 	"buf.build/go/bufplugin/check"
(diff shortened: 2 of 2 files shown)


private/buf/bufworkspace/workspace.go
```diff
@@ -92,6 +92,7 @@ type Workspace interface {
 	// handle v1 vs v2 transparently. Right now, this is only approved to be used in push
 	// when we want to know whether we need to print out only CommitIDs. Any other usages
 	// need to be evaluated.
+	// TODO: Validate the usecase for `buf config ls-lint-rules` and `buf config ls-breaking-rules`.
 	IsV2() bool
```
emcfarlane, line 96: This change uses the `IsV2` method to restrict use of the flag `--module-path` to only work on v2 Workspaces and read the version for use with `bufcheck.Client` `AllRules` method. The current behaviour reads the version directly from the `buf.yaml` file.

private/buf/cmd/buf/command/config/internal/internal.go
```diff
@@ -254,12 +253,16 @@ func lsRun(
 		}
 	} else {
 		allRulesOptions := []bufcheck.AllRulesOption{
-			bufcheck.WithPluginConfigs(bufYAMLFile.PluginConfigs()...),
+			bufcheck.WithPluginConfigs(workspace.PluginConfigs()...),
+		}
+		fileVersion := bufconfig.FileVersionV1
+		if workspace.IsV2() {
```
emcfarlane, line 259: We would need to return the `bufconfig.FileVersion` here, as there is currently no way to tell if the version is v1, v1beta1 or v2.

doriable: Hmm, that doesn't make sense at a `Workspace` level though, since `v1` and `v1beta1` `buf.yaml` versions are module-level concerns... we may need to rethink this refactor a little bit.

PR #3538, opened 2024-12-12
Add proto definition of plugin display RPC

cyinma:


proto/buf/alpha/registry/v1alpha1/display.proto
```diff
@@ -29,6 +29,10 @@ service DisplayService {
   rpc DisplayRepositoryElements(DisplayRepositoryElementsRequest) returns (DisplayRepositoryElementsResponse) {
     option idempotency_level = NO_SIDE_EFFECTS;
   }
+  // DisplayCheckPluginElements returns which check plugin elements should be displayed to the user.
+  rpc DisplayCheckPluginElements(DisplayCheckPluginElementsRequest) returns (DisplayCheckPluginElementsResponse) {
```
mfridman, line 33: I think this should be `DisplayPluginElements` and the terminology should be "plugin" because even though there might be additional types of plugins, the plugin pages (like settings, et.c) should be the same?

cyinma: yeh make sense, updated

PR #3539, opened 2024-12-13
Fix relative targetpath bug

sinwoobang:
# Bug Fix: Path Resolution for Local File System Imports

TL;DR
- Before: When using local files, you were building paths starting from where the command was run
- After: Now you build paths starting from where the files are imported from, just like we already do with git
- The fix: You explicitly resolve the target path using the import path as the base directory

The key change was making sure we use the import location as our "starting point" when figuring out where to put files, regardless of whether we're importing from local files or git.

## Issue
The export command was failing for local filesystem imports while working correctly for git imports. When using local filesystem imports, relative output paths were incorrectly resolved, causing errors when trying to export to paths outside the context directory.

For example, this local filesystem import failed:
```
export -o apivendor --exclude-imports ../local-vendor-protos --path package/project/common
```
With error:
```
Failure: ../cur_dir/package/project/common: is outside the context directory
```

While the equivalent git import worked as expected:
```
export -o apivendor --exclude-imports ssh://git/local-vendor-protos.git --path package/project/common
```

## Root Cause
The path resolution logic for local filesystem imports was not implementing the same correct behavior as git imports. Local filesystem imports were incorrectly using the current directory as the base for resolving relative paths.

## Fix
Updated the local filesystem path resolution logic to match the correct behavior already implemented in git imports. The path is now properly resolved as `../apivendor/package/project/common` for both import types.

### DIFF
--- a/private/buf/buffetch/internal/reader.go
+++ b/private/buf/buffetch/internal/reader.go
@@ -636,7 +636,12 @@ func getReadWriteBucketForOS(
 	}
 	fsRootTargetPaths := make([]string, len(targetPaths))
 	for i, targetPath := range targetPaths {
-		_, fsRootTargetPath, err := fsRootAndFSRelPathForPath(targetPath)
+		resolvedTargetPath := targetPath
+		if !filepath.IsAbs(targetPath) {
+			resolvedTargetPath = fsRoot + fsRootInputSubDirPath + "/" + targetPath
+		}
+
+		_, fsRootTargetPath, err := fsRootAndFSRelPathForPath(resolvedTargetPath)
 		if err != nil {
 			return nil, nil, err
 		}


bufdev:
@doriable can you check this out and make sure its (1) right (2) unix/windows independent?

doriable:
Hello, thanks for submitting this PR! So, this is actually the intended behaviour of paths and `buf` commands.

For local inputs, paths provided to flags are always relative to the execution context of the command. For git inputs, it is considered relative to the input directory, which in your case is the top-level of the repository, but this can be set with the `subdir` (docs: https://buf.build/docs/reference/inputs/#git).

I can help with adjusting your workflows as needed, feel free to file an issue and/or reach out on Slack.

PR #3547, opened 2024-12-16
Stagger lint and breaking checks in LSP Refresh

doriable:
Since lint and breaking change detection are more
expensive, we do not need to run them on every single
file refresh. Instead, we use the top-level LSP lock and a
short stagger (5ms) and only run when the attempt to get
the top-level lock is successful. Otherwise, we no-op and
assume that the checks are handled by another `Refresh`
request.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -295,6 +295,8 @@ func (f *file) Refresh(ctx context.Context) {
 	// lock. It is safe to use because if another LSP call is made, we allow checks to finish
 	// before resolving a subsequent LSP request.
 	go func() {
+		// We stagger the check operation by 5ms and run it for the latest Refresh state.
+		time.Sleep(refreshCheckStagger)
 		// Call TryLock, if unnsuccessful, then another thread holds the lock, so we provide a
 		// debug log and move on.
 		if !f.lsp.lock.TryLock() {
@@ -303,10 +305,8 @@ func (f *file) Refresh(ctx context.Context) {
 			)
 			return
 		}
-		// We have successful obtained the lock, we can now run the checks.
+		// We have successfully obtained the lock, we can now run the checks.
 		defer f.lsp.lock.Unlock()
-		// We stagger the check operation by 5ms and run it for the latest Refresh state.
-		time.Sleep(refreshCheckStagger)
 		f.BuildImages(ctx)
 		f.RunLints(ctx)
 		f.RunBreaking(ctx)
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"slices"
 	"strings"
+	"time"
 
 	"github.com/bufbuild/buf/private/buf/bufworkspace"
 	"github.com/bufbuild/buf/private/bufpkg/bufanalysis"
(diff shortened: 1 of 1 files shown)


bufdev (approved):

private/buf/buflsp/file.go
```diff
@@ -285,9 +289,28 @@ func (f *file) Refresh(ctx context.Context) {
...
-	f.RunBreaking(ctx)
+	// Since checks are a more expensive operation, we do not want to run a check on every
+	// Refresh call. Instead, we can stagger the checks and only run them periodically by
+	// spinning them off into a go routine. Then we attempt to lock using the top-level LSP
+	// lock. It is safe to use because if another LSP call is made, we allow checks to finish
+	// before resolving a subsequent LSP request.
+	go func() {
+		// Call TryLock, if unnsuccessful, then another thread holds the lock, so we provide a
+		// debug log and move on.
+		if !f.lsp.lock.TryLock() {
```
mcy, line 300: This is out of order. You want to sleep first, *then* attempt to grab the lock. Otherwise, this blocks requests from the LSP, which will mean the generated lints won't use the most recent content.

emcfarlane, line 302: This looks like the final run of lint and breaking could be mismatched to the output. I think we'd want to ensure if a change occurred after the job was run but no change happens after its finished it re-runs to reflect the latest. This could also be done by cancelling the previous work and always starting at the latest.

doriable: The way the stagger works is that we do not want to hammer checks when many `Refresh` calls are made in a flurry. I just fixed the ordering, so we should be converging on the final state when the checks are run. The top-level locking would also mean you wouldn't need to cancel the previous work because no additional LSP requests are being handled while the checks are running.

emcfarlane: The number of running jobs should be limited but the final annotations should always catch up to the latest otherwise I think the output could be stale e.g.:

```mermaid
sequenceDiagram
    Client->>Refresh: Change triggered job
    activate Refresh
    Client--xRefresh: Change before job is completed is lost
    Refresh->>Client: Stale annotation
    deactivate Refresh
```
Could mean that the user just has to write a comment or something to trigger the job again? The lock could also mean the job is always skipped if the LSP is busy?

private/buf/buflsp/file.go
```diff
@@ -285,9 +289,28 @@ func (f *file) Refresh(ctx context.Context) {
...
+			f.lsp.logger.Debug(
+				fmt.Sprintf("another thread holds the LSP lock, no new checks started for %v", f.uri),
+			)
+			return
+		}
+		// We have successfully obtained the lock, we can now run the checks.
+		defer f.lsp.lock.Unlock()
+		f.BuildImages(ctx)
```
emcfarlane, line 310: Should `BuildImages` be out of this loop as it's used below in `IndexSymbols`?

private/buf/buflsp/file.go
doriable: It's the top-level LSP lock, so if a job is running, then the next `Refresh` request blocks until the lock is freed up again, so this should not be the case...

private/buf/buflsp/file.go
doriable: `IndexSymbols` uses the AST file nodes, but does not appear to use the image...

emcfarlane (approved):

PR #3548, opened 2024-12-16
Add bufplugin.RunnerProvider for remote plugins

emcfarlane:
This adds a new bufplugin.RunnerProvider to support remote wasm plugins. The constructor `bufcheck.NewRunnerProvider` has been renamed to `NewLocalRunnerProvider` to clarify the runner interacts with the local filesystem when invoking plugins. It now takes a `bufplugin.PluginKeyProvider` and `bufplugin.PluginDataProvider` to resolve and load remote plugin references.

This also changes the PluginConfig to fix arguments for different plugin types. The method Path has been split into Name and Args. This is analogous to the naming in the std libraries `exec.Command` function. Name must not be empty. Args may be nil, but are valid for all plugin types. The name is either the path to the binary, the path to the wasm binary, or the fullname of the plugin in the form remote/owner/name.

Subset of #3524


doriable (approved):
This makes sense to me, and splitting it out to handle the args is helpful!

PR #3550, opened 2024-12-16
Add `WithRelatedCheckConfigs` option to check additional check configs for unused plugins warning

doriable:
We check for unused plugins when running `Lint`,
`Breaking`, and `ConfiguredRules`. However, this is based
on the provided `CheckConfig`, and does not account for
configurations outside of the current check. This results in a
confusing unused plugin warning for users who may not have
configured rules and/or categories configured for all modules
in their workspace.

This PR introduces a `WithRelatedCheckConfigs`, that allows
us to provide additional check configs to check plugins against.
It also allows us to no longer rely on the heuristic to ignore plugins
with multiple rule types -- we can provide all related check configs
and check against those.


private/bufpkg/bufcheck/rules_config.go
```diff
@@ -102,6 +104,10 @@ type rulesConfig struct {
 	// Rules, we don't warn for this plugin, as it may have had rules configured in breaking that
 	// we haven't accounted for.
 	//
+	// The caller can provider additional check configs to check if the plugin has rules configured.
+	// If the plugin has rules configured in any of the additional check configs provided, then
+	// we don't warn.
```
emcfarlane, line 109: The case above (`In client.Lint and client.Breaking, we only have the Lint or Breaking config`) could probably be removed now to say related configs are provided for validation and remove the heuristic workaround based on other rule types.

private/bufpkg/bufcheck/bufcheck.go
```diff
@@ -126,6 +126,24 @@ type ConfiguredRulesOption interface {
...
+	BreakingOption
+	ConfiguredRulesOption
+}
+
+// WithAdditionalCheckConfigs returns a new LintBreakingAndConfiguredRulesOption that allows
+// the caller to provide additional check configs that are being run. This allows the client
+// to check for unused plugins across additional check configs and provide users with a warning.
+//
+// The default is to only check the configs provided to the client for Lint, Breaking, or ConfiguredRules.
+func WithAdditionalCheckConfigs(additionalCheckConfigs ...bufconfig.CheckConfig) LintBreakingAndConfiguredRulesOption {
```
emcfarlane, line 141: Maybe `WithRelatedCheckConfigs` would be a better name here? As additional sounds like it would be in _addition_ to the current config.

private/bufpkg/bufcheck/rules_config.go
doriable: So this is one I was kinda hoping to get opinions on. Currently, I am only providing the respective other check configs (e.g. other lint configs for lint, other breaking configs for breaking), but yes, we could provide both instead, and that would allow us to remove this clause and rely solely on the related check configs. I like that idea, but I wanted to get others' opinions on this.

doriable: I thought a little bit more about this, and the goal of this is to provide users with a clear warning for whenever a plugin does not have any configured rules and/or categories. And so I think it actually doesn't make sense to rely on the heuristic, it seems we are doing this because `RulesConfig` always only ever applies to a single `RuleType`:

https://github.com/bufbuild/buf/blob/b4d78361a823ccad7b204bbd7e320fac4df253cc/private/bufpkg/bufcheck/rules_config.go#L126

However, at the time of the validation, we have access to all rules, and so we can check all related check configs, agnostic to rule type, which would allow us to surface a much clearer warning to users without relying on a heuristic. So I have pushed up a change to reflect that for the warning.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -217,6 +218,22 @@ func run(
...
+				return imageWithConfig.LintConfig()
+			},
+		),
+		slicesext.Map(
+			imageWithConfigs,
+			func(imageWithConfig bufctl.ImageWithConfig) bufconfig.CheckConfig {
+				return imageWithConfig.BreakingConfig()
+			},
+		)...,
+	)
```
emcfarlane, line 236: Maybe a for loop is easier here as it would avoid needing to build the two lists? Or we could maybe add an `AllConfigs() []bufconfig.CheckConfig` method to `ImageWithConfig`.

emcfarlane (approved):

private/buf/cmd/buf/command/breaking/breaking.go
doriable: I think this makes sense, pushed up a change for this :)

private/bufpkg/bufcheck/rules_config.go
```diff
@@ -305,11 +303,31 @@ func newRulesConfig(
...
+	allRuleIDToRule, err := getIDToRuleOrCategory(allRules)
+	if err != nil {
+		return nil, err
+	}
+	allRuleIDToCategoryIDs, err := getRuleIDToCategoryIDs(allRules)
+	if err != nil {
+		return nil, err
+	}
+	allCategoryIDToRuleIDs := getCategoryIDToRuleIDs(allRuleIDToCategoryIDs)
+	for _, checkConfig := range relatedCheckConfigs {
```
bufdev, line 317: It looks like if you specify the WithRelatedCheckConfigs option, these CheckConfigs are not *additionally* checked, rather these CheckConfigs are the *only* CheckConfigs that are checked, and the original CheckConfigs (?) from allRules are not checked. Am I wrong/does this make sense? Make sure it is additional.

doriable: Yep! On lines 294-305, we are still checking the original configurations, we are just checking the additional rules in lieu of checking whether the plugin has rules from other rules types.

emcfarlane (approved):

PR #3560, opened 2024-12-20
Add flag --digest-changes-only to filter commits for modules and plugins

emcfarlane:
This adds a flag `--digest-changes-only` to the commands `buf registry {module,plugin} commit list` to filter out any commits without a changed digest. By default all commits are listed. This is similar to the UI view "Hide commits with no changes" in the BSR.


bufdev:
Need to workshop naming here live. The "no changes" labeling always bothered me a tad, so we can hammer out both at once on Tuesday! Prob will be a 5-10 minute call.

mfridman:
More of a driveby to suggest some names: `--content-only`, `--filter-content-only` or `--source-only`.

Maybe just me, but I've always thought about this behavior as "metadata" (vcs url, deps) vs "content" (proto, license, etc.) changes.

doriable:
I think that maybe incorporating the term digest, e.g. `--digest-changes-only`, might be more clear, since new commits are created for changes in source control URL and we otherwise have commit squashing on the BSR. Otherwise, this all looks reasonable to me!

bufdev:
`--digest-changes-only` sounds good to me!

doriable (approved):

bufdev (approved):

PR #3595, opened 2025-01-16
Add input parameter filter for git commands

emcfarlane:
This adds a new input parameter `filter` to assist in working with large git projects. The param filter sets the git filter-spec argument when fetching repositories. Only one argument can be provided, the combined form can be used to set multiple filters. When using filter combined with a `subdir` the checkout will now also be sparse to further reduce the number of workspace files.

The documentation for filter arguments can be found in the `git rev-list` command: https://git-scm.com/docs/git-rev-list#Documentation/git-rev-list.txt---filterltfilter-specgt
See this blog post on the different filter checkouts that may be used: https://github.blog/open-source/git/get-up-to-speed-with-partial-clone-and-shallow-clone/

As an example using the envoy project as a larger repository, it can be built using the new filter argument:
```
 buf build https://github.com/envoyproxy/envoy.git#subdir=api,branch=main,filter=tree:0
```
Which reduces the fetch of the git repository from 10.6s to 2.6s.


bufdev (approved):
Make sure you update docs too

doriable (approved):

PR #3610, opened 2025-01-27
Remove support for deprecated protovalidate options

jchadwick-buf:
This is related to https://github.com/bufbuild/protovalidate/pull/297, so it should wait until that is merged first.


private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -1037,6 +1014,104 @@ func checkExampleValues(
...
+
+	// Ensure that enum and message types are available to be resolved. It's a
+	// bit tricky to get this right, so we just import everything the original
+	// file containing the root message imports (and that file, too.)
+	for i := 0; i < originalFile.Imports().Len(); i++ {
+		newFileDescriptorProto.Dependency = append(
+			newFileDescriptorProto.Dependency,
+			originalFile.Imports().Get(i).Path(),
+		)
+	}
```
jhump, line 1058: If you just mark the one import above (the original file) as a public import, you don't need any of this.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -1037,6 +1014,104 @@ func checkExampleValues(
...
+	// want to create a synthetic clone of the map field itself.
+	rootFieldDescriptor := fieldDescriptor
+	if parentMapFieldDescriptor != nil {
+		rootFieldDescriptor = parentMapFieldDescriptor
+	}
+	originalFile := rootFieldDescriptor.ContainingMessage().ParentFile()
+	newFileDescriptorProto := &descriptorpb.FileDescriptorProto{
+		Name:    proto.String(":synthetic:"),
+		Package: proto.String("__SYNTHETIC__"),
+		Syntax:  proto.String("proto3"),
```
jhump, line 1044: This file must have the same syntax (and possibly edition) as the original file. Otherwise, it the new field descriptor may have properties that are incompatible with this. For example, if the field has features, those aren't allowed outside of editions. If it has a required label, that's not allowed outside of proto2.

Also, for editions, you need to copy over the original file's features -- to make sure the new synthetic file has all of the same file-wide default feature values as the original. And the new synthetic containing message will need to mirror that of the original. But this step gets more complicated: if the original message was not a top-level message, you need to actually _combine_ the features of the original message and all of its ancestors/enclosing messages, so that the set of message-level features (from which a field might inherit) are identical to those in the original message.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -1037,6 +1014,104 @@ func checkExampleValues(
...
+
+	// The map field and entry message have to be declared in the same scope, so
+	// we need to make a clone of the map entry type itself. This needs to be
+	// nested under the synthetic message.
+	if parentMapFieldDescriptor != nil {
+		// Shouldn't happen, but we don't want to panic if it does.
+		if !parentMapFieldDescriptor.IsMap() {
+			return nil, nil, errors.New("parentMapFieldDescriptor is not a map")
+		}
+		newMapEntryMessageDescriptor := &descriptorpb.DescriptorProto{
```
jhump, line 1078: I think this would be much safer (and less code) to instead use `protodesc` to mechanically translate the entire map entry message to a `*descriptorpb.DescriptorProto`, to make sure you don't miss anything here (such as features in an editions file).

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -1037,6 +1014,104 @@ func checkExampleValues(
...
+			Options: &descriptorpb.MessageOptions{
+				MapEntry: proto.Bool(true),
+			},
+		}
+		newMessageDescriptorProto.NestedType = append(
+			newFileDescriptorProto.MessageType,
+			newMapEntryMessageDescriptor,
+		)
+		newFieldDescriptorProto.Type = descriptorpb.FieldDescriptorProto_TYPE_MESSAGE.Enum()
+		newFieldDescriptorProto.TypeName = proto.String(newMapEntryMessageDescriptor.GetName())
```
jhump, line 1093: This isn't necessary. If it was originally a map field, these are already set this way, from the call to `protodesc.ToFieldDescriptorProto`.

jhump (commented):
Wow, this is complicated.

I imagine this would be _way_ simpler and easier to understand (and less potentially brittle) to just skip the synthetic message and validate the violations solely using a filter on the violation field paths.

The only problem I foresee with that is that if any other field has a configuration error (like a bad CEL expression) then none of the examples can be validated. But we'd still issue a lint warning for that configuration error, right? So seems like it's probably an okay trade-off.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
jhump: Er... never mind, I'm wrong. Ignore this comment. You'd need to instead actually _edit_ the original file to make all of _its_ imports public, which you definitely don't want to hassle with...

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
jhump: You'll also need to handle proto2 groups: in those cases, it should be fine to treat like any other message field, and let it refer to the group definition in the original file. But you'd need to change the field's type from `TYPE_GROUP` to `TYPE_MESSAGE`.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
jhump: Oof, never mind. That is only true of the first one, the type enum. This is correct that it must re-write the type name.

doriable:
Closing this since we are going to update this with the new `Filter` interface (https://github.com/bufbuild/protovalidate-go/pull/185), feel free to re-open and/or continue the discussion here as needed!

PR #3611, opened 2025-01-28
Fix outdated comment

nicksnyder:
This comment is out of sync with the implementation.

Do we also need to check short flag names separately (like `-t`) or is that already handled?


doriable (approved):
Stamping -- and just to make sure I address the question in the description, we don't need a separate check for the short flag names, they are captured with the same `flags` field.

PR #3618, opened 2025-02-03
Make upgrade

doriable:
Upgrade our dependencies, except for `protovalidate-go`.
The API changes affect our lint rule in a fairly significant way, so
I'll put up a separate PR for that upgrade.


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/bufbuild/buf
 
-go 1.22.0
+go 1.23.0
```
doriable, line 3: `go-containerregistry` is on `1.23.0`

bufdev (approved):

PR #3619, opened 2025-02-04
Change code example type

rakuista:
The new docs site uses `console` instead of `terminal` to be compatible with Pygments.

### DIFF
--- a/private/bufpkg/bufcobra/markdown.go
+++ b/private/bufpkg/bufcobra/markdown.go
@@ -183,15 +183,12 @@ func hasSubCommands(cmd *cobra.Command) bool {
 
 // processDescription is used to process description and example text. It does the following:
 //
-// - Converts all code blocks to terminal code blocks  (```terminal)
-// - Unidents code blocks
+// - Converts all code blocks to console code blocks  (```console)
+// - Unindents code blocks
 // - Writes out the text, escaping all HTML characters.
 //
-// This is done because terminal code blocks have special handling in some documentation
-// systems (e.g. not copying the $ sign if the code block has a copy button).
-//
-// Also, Docusaurus treats indented code blocks different and does not escape HTML characters
-// so we do call that here unilaterally.
+// This is done because Pygments (which mkdocs-material uses for syntax highlighting)
+// specifies `console` as the language for bash sessions in code blocks.
 func processDescription(description string) string {
 	out := &bytes.Buffer{}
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufcobra/markdown.go
```diff
@@ -104,7 +104,7 @@ func generateMarkdownPage(
 	p("\n\n")
 	if command.Runnable() {
 		p("### Usage\n")
-		p("```terminal\n$ %s\n```\n\n", command.UseLine())
+		p("```console\n$ %s\n```\n\n", command.UseLine())
```
stefanvanburen, line 107: We'll also want to update this in the `processDescription` function below, and make sure we update the function documentation.

For that function though, I did notice that while we're _typically_ showing usage of `buf` in some console-esque context, we'll occasionally mis-classify some non-console text — e.g. on https://buf.build/docs/reference/cli/buf/generate/, the "template file" YAML is classified as `terminal` and would be changed to `console`. I think it's ok to come back and fix that separately, if at all.

doriable: Yeah, I think it's fine to unify all to `terminal`. I'm not sure there is a better way to deal with different types of code blocks in descriptions unless the description itself passes the code block type. But yeah, definitely something to consider later.

stefanvanburen: `s/terminal/console` and I agree with you :)

private/bufpkg/bufcobra/markdown.go
```diff
@@ -104,15 +104,15 @@ func generateMarkdownPage(
...
 	}
 	if len(command.Long) > 0 {
 		p("### Description\n\n")
 		p("%s \n\n", processDescription(command.Long))
 	}
 	if len(command.Example) > 0 {
 		p("### Examples\n\n")
-		p("```\n%s\n```\n\n", processDescription(command.Example))
+		p("```console\n%s\n```\n\n", processDescription(command.Example))
```
stefanvanburen, line 115: This is fine to do, but I don't think we use `command.Example` — I was more referring in my comment to the function down on line 195; we'll want to update that from `terminal` -> `console` in the code and probably the comments above it. Happy to push a commit if you'd like me to!

rakuista:
Yep, realized that after committing this—working on it.

stefanvanburen (approved):

doriable (approved):

PR #3625, opened 2025-02-07
Document registry logout argument

pkwarren:
Update `buf registry logout` to document the domain optional argument, which if specified will log out of a specific BSR instance.


private/buf/cmd/buf/command/registry/registrylogin/registrylogin.go
```diff
@@ -53,9 +53,9 @@ func NewCommand(
 ) *appcmd.Command {
 	flags := newFlags()
 	return &appcmd.Command{
-		Use:   name + " <domain>",
+		Use:   name + " [domain]",
```
pkwarren, line 56: Since this is an optional argument I think it should be in square brackets, not angle brackets.

emcfarlane (approved):

bufdev (approved):

PR #3629, opened 2025-02-11
Parse app config strictly by default

saquibmian:
In #3376 we switched `ReadConfig` from `UnmarshalYAMLStrict` to `UnmarshalYAMLNonStrict`. This was to enable a use case in an internal application. This PR switches `ReadConfig` back to `UnmarshalYAMLStrict` and adds `ReadConfigNonStrict` to enable the non-strict use case. The internal application will document why it needs the non-strict property.

I verified that nothing in the CLI needs non-strict parsing, everything is using the CLI's `NameContainer` and reads `bufapp.Config`.

### DIFF
--- a/private/pkg/app/appext/appext.go
+++ b/private/pkg/app/appext/appext.go
@@ -174,6 +174,25 @@ func ReadConfig(container NameContainer, value interface{}) error {
 	return nil
 }
 
+// ReadConfigNonStrict reads the configuration from the YAML configuration file config.yaml
+// in the configuration directory, ignoring any unknown properties.
+//
+// If the file does not exist, this is a no-op.
+// The value should be a pointer to unmarshal into.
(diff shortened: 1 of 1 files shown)


saquibmian:
> If we did it in that PR it was definitely intentional

Yup that's why I opened the PR to discuss if this was intentional. I have poked around a little and can't see yet why we would do this, wondering if you can recall why we did this?

mfridman:
I also cannot recall why. Maybe we could add a `ReadConfigStrict` or a functional option to control the default behavior?

In other internal projects, we'd like to use strict unmarshal. It was a property we'd like to maintain to avoid config issues.

mfridman (approved):
Makes sense to me to have `ReadConfig` be strict by default and offer a way to use non-strict.

No opinion whether it should be a separate function or a functional option.

doriable (approved):
Agree with @mfridman's thoughts on the options, stamping!

PR #3631, opened 2025-02-12
Fix b5 module digest calculation with transitive dependencies

emcfarlane:
This fixes the b5 digest calculation for remote Modules when a transitive dependency is no longer a dependency in the parents ModuleSet. The digest calculation on Module.Digest would filter dependencies causing a ModuleData to fail tamper-proofing, as it would not filter it's dependencies. The documentation has been reworded to highlight the error case.

The ModuleData's method `DeclaredDepModuleKeys` has been renamed to `DepModuleKeys`. This clarifies the difference between declared dependencies, those in the `buf.yaml`, and the full dependency list, those in the `buf.lock`. The digest calculation must use the full list of dependencies, not the declared.

This is similar to the fix proposed in #3190 but does not cache the digest calculation in ModuleData for reuse in the Modules digest calculation. A Modules digest will always be calculated locally.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
+++ b/private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
@@ -125,6 +125,8 @@ func (a *moduleDataProvider) getIndexedModuleDatasForRegistryAndIndexedModuleKey
 	indexedModuleKeys []slicesext.Indexed[bufmodule.ModuleKey],
 	digestType bufmodule.DigestType,
 ) ([]slicesext.Indexed[bufmodule.ModuleData], error) {
+	// This returns the graph with pruned dependencies for the given indexedModuleKeys.
+	// We must resolve the declared dependencies for each ModuleKey.
 	graph, err := a.graphProvider.GetGraphForModuleKeys(ctx, slicesext.IndexedToValues(indexedModuleKeys))
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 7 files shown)


private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
```diff
@@ -149,62 +151,45 @@ func (a *moduleDataProvider) getIndexedModuleDatasForRegistryAndIndexedModuleKey
...
+			func(i int, j int) bool {
+				return depModuleKeys[i].FullName().String() < depModuleKeys[j].FullName().String()
+			},
+		)
 
-			universalProtoContent, ok := commitIDToUniversalProtoContent[moduleKey.CommitID()]
-			if !ok {
-				// We only care to get content for a subset of the graph. If we have something
-				// in the graph without content, we just skip it.
-				return nil
```
emcfarlane, line 180: This change avoids walking the dependencies. The TopoSort is only required to be done on the provided `indexedModuleKeys`.

private/bufpkg/bufmodule/module.go
```diff
@@ -452,33 +452,31 @@ func newGetDigestFuncForModuleAndDigestType(module *module, digestType DigestTyp
...
 				}
-				moduleDepFullNames := make(map[string]struct{}, len(moduleDeps))
-				for _, dep := range moduleDeps {
-					fullName := dep.FullName()
-					if fullName == nil {
-						return nil, syserror.Newf("remote module dependencies should have full names")
-					}
-					moduleDepFullNames[fullName.String()] = struct{}{}
-				}
-				prunedDepModuleKeys := make([]ModuleKey, 0, len(declaredDepModuleKeys))
```
emcfarlane, line 475: This is the fix and the cause of the tamper-proof error. Dependencies should **not** be filtered to the context of the parent Module. This is required to avoid the Digest for the remote changing when included in a Module. If the parent Module `A` doesn't include dependency `X` but the remote module `B` depends on `X`, its dependency should still be used in the calculation of the Digest for `B`.

private/bufpkg/bufmodule/module_data.go
```diff
@@ -140,24 +147,15 @@ func newModuleData(
...
-				// tamper-proofing failing.
-				//
-				// This mismatch is a bit weird, however, and also results in us effectively computing
-				// the digest twice for any remote module: once here, and once within Module.Digest,
-				// which does have a slight performance hit.
-				actualDigest, err = getB5DigestForBucketAndDepModuleKeys(ctx, bucket, declaredDepModuleKeys)
+				// The B5 digest is calculated based on the dependencies.
+				// Dependencies are not required to be resolved to a Module to calculate the digest.
+				// Each dependent ModuleKey is a reference to a specific commit of a module
+				// that includes the dependencies expected Digest.
```
saquibmian, line 157: This comment was a little hard for me to parse. Maybe something like this?
> The B5 digest is calculated based on the dependencies. We do not need to construct a Module to calculate a digest as we can trust that `DepModuleKeys` is the complete set of deps for the Module and is exactly the deps that were present in the ModuleSet when this Module was pushed.

emcfarlane: I'm not sure trust is correct, as it is always validated if used. What about:
> // The B5 digest is calculated based on the dependencies.
> // The dependencies are not required to be resolved to a Module to calculate this Digest.
> // Each ModuleKey includes the expected Digest which is validated when loading that dependencies ModuleData, if needed.

private/bufpkg/bufmodule/module_data.go
```diff
@@ -32,16 +32,23 @@ type ModuleData interface {
...
 	//
 	// This bucket is not self-contained - it requires the files from dependencies to be so.
 	//
 	// This bucket will only contain module files - it will be filtered via NewModuleData.
 	Bucket() (storage.ReadBucket, error)
-	// DeclaredDepModuleKeys returns the declared dependencies for this specific Module.
-	DeclaredDepModuleKeys() ([]ModuleKey, error)
+	// DepModuleKeys returns the dependencies for this specific Module.
+	//
+	// The dependencies are the same as that would appear in the buf.lock file.
```
saquibmian, line 45:
```suggestion
	// The dependencies are the same as that would appear in the v1 buf.lock file.
```
(v2 buf.lock is for workspaces, so would contain more deps than this module needs)

private/bufpkg/bufmodule/module_data.go
```diff
@@ -32,16 +32,23 @@ type ModuleData interface {
...
+	// These include all direct and transitive dependencies. A Module constructed
+	// from this ModuleData as the target will require all Modules referenced by
+	// its DepModuleKeys to be present in the ModuleSet.
+	//
+	// This is used for digest calculations. It is not used otherwise.
```
saquibmian, line 50: Within the CLI... internally it is used for other purposes. Consider removing.

saquibmian (approved):
Algorithm is sound and is basically an unrolled variant of #3190.

doriable (approved):

saquibmian (approved):

PR #3632, opened 2025-02-12
Cleanup TopoSort on module data deps

emcfarlane:
This is a small optimization to remove uneeded work when sorting module dependencies. No change in behavior.

### DIFF
--- a/private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
+++ b/private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
@@ -167,7 +167,7 @@ func (a *moduleDataProvider) getIndexedModuleDatasForRegistryAndIndexedModuleKey
 
 		universalProtoContent, ok := commitIDToUniversalProtoContent[moduleKey.CommitID()]
 		if !ok {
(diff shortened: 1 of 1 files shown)


private/bufpkg/bufmodule/bufmoduleapi/module_data_provider.go
```diff
@@ -149,62 +149,44 @@ func (a *moduleDataProvider) getIndexedModuleDatasForRegistryAndIndexedModuleKey
...
-			moduleKey bufmodule.ModuleKey,
-			_ []bufmodule.ModuleKey,
-			_ []bufmodule.ModuleKey,
-		) error {
-			// TopoSort will get us both the direct and transitive dependencies for the key.
-			//
-			// The outgoing edge list is just the direct dependencies.
-			//
-			// There is definitely a better way to do this in one pass for all commits with
-			// memoization - this is algorithmically bad.
```
emcfarlane, line 163: This PR avoids TopoSort on each dep. This function is usually called with one indexed module key, but this comment might be referring to an optimization on multiple module keys.

doriable (approved):

PR #3638, opened 2025-02-18
Upgrade deps

doriable:
Upgrade our deps, excluding `protovalidate-go` and
`bufplugin` (which depends on the new version of `protovalidate-go`).


Dockerfile.buf
```diff
@@ -1,4 +1,4 @@
-FROM --platform=${BUILDPLATFORM} golang:1.23-alpine3.21 as builder
+FROM --platform=${BUILDPLATFORM} golang:1.24-alpine3.21 as builder
```
emcfarlane, line 1: This upgrades go version.

doriable: Only for our Docker image... which I think is fine?

PR #3639, opened 2025-02-18
Remove stdlib api functions now that we're on Go 1.23

pkwarren:


private/bufpkg/bufprotosource/files.go
```diff
@@ -50,8 +51,11 @@ func newFiles[F InputFile](
 		}
 		return files, nil
 	}
-
-	chunks := slicesext.ToChunks(indexedInputFiles, chunkSize)
+	// slices.Chunk panics if given a chunk size less than 1 - allocate a single chunk in this case.
```
pkwarren, line 54: This is one difference between the previous `slicesext.ToChunks` implementation and the stdlib one, although it doesn't seem too bad in practice. If desired, I could re-add the method to slicesext and add this handling there but still re-use the stdlib method.

private/pkg/slicesext/slicesext.go
```diff
@@ -12,36 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// Some implementations are copied from "slices", and hence
```
pkwarren, line 15: It looks like any methods from stdlib slices have been removed now so removed this notice.

saquibmian (approved):
LGTM but I'd probably wait for a CLI code owner.

private/bufpkg/bufprotosource/files.go
doriable: Let's re-add the method to `slicesext` with the handling encapsulated since the `slicesext` package is staying, so I think it makes sense to minimize the impact of this.

pkwarren: Added back in b52fd39bf and re-added tests.

doriable (approved):

PR #3652, opened 2025-02-26
Update protovalidate dependency

jchadwick-buf:
This is a successor to #3610. 

Changed:
- Update protovalidate to latest revision.
- Update example linting code to use `WithFilter`
- Still retains the other changes (removing support for deprecated protovalidate options.)

This uses an unreleased version of protovalidate-go, which was suggested so we could try to finalize the filter API before cutting a release with it.

Open question remains: should we try to warn users if the legacy removed protovalidate options are present, somehow? I assume this could be checked by looking for unknown fields in the options.

(This should unbreak the errors that occur when trying to use buf with `go tool`, although we do not recommend using `go tool`.)


jchadwick-buf:
Seems like we are running into problems with the latest protobuf runtime.

doriable:
> Seems like we are running into problems with the latest protobuf runtime.

This should be resolved on `main`, I'll keep an eye on this PR if there are more issues!

go.mod
```diff
@@ -6,17 +6,17 @@ toolchain go1.23.5
...
-	buf.build/go/bufplugin v0.7.0
+	buf.build/go/bufplugin v0.8.0
 	buf.build/go/protoyaml v0.3.1
 	buf.build/go/spdx v0.2.0
 	connectrpc.com/connect v1.18.1
 	connectrpc.com/otelconnect v0.7.1
 	github.com/bufbuild/protocompile v0.14.1
 	github.com/bufbuild/protoplugin v0.0.0-20250106231243-3a819552c9d9
-	github.com/bufbuild/protovalidate-go v0.8.2
+	github.com/bufbuild/protovalidate-go v0.9.3-0.20250317160558-38a17488914d
```
emcfarlane, line 19: Does this need to be updated to a release version before merge?

private/buf/bufctl/controller.go
```diff
@@ -1287,6 +1287,15 @@ func (c *controller) handleFileAnnotationSetRetError(retErrAddr *error) {
 	}
 }
 
+// Implements [protoyaml.Validator] using [protovalidate.Validator]. This allows us to pass
+// a [protovalidate.Validator] to the [protoencoding.NewYAMLUnmarshaler] and validate types
+// when unmarshalling.
+type yamlValidator struct{ protovalidateValidator protovalidate.Validator }
```
emcfarlane, line 1293: This type doesn't look necessary as the interface is the same: https://pkg.go.dev/buf.build/go/protoyaml#Validator and https://pkg.go.dev/github.com/bufbuild/protovalidate-go#Validator .

emcfarlane (approved):

private/buf/bufctl/controller.go
doriable: The current version of `Validator.Validate` takes variadic `ValidationOption`s: https://github.com/bufbuild/protovalidate-go/blob/38a17488914d10dba0b143496b8d43541e6e3378/validator.go#L33-L41

PR #3665, opened 2025-03-09
Allow zero length for binpb, txtpb, and yaml for `buf convert`

### LINKED ISSUE #3663: Zero length encoding is considered invalid.

https://github.com/daniel-wong-dfinity-org/foo.proto

### Commands

```shell
# I assume that buf 1.50.0 is installed and on your PATH.

echo -n '' | \
buf convert \
  foo.proto \
  --type=M \
  --from=-#format=binpb \
  --to=-#format=txtpb
```

### Output

```shell
Failure: --from: length of data read from "-#format=binpb" was zero
```

### Expected Output

None. Not even a trailing newline. Literally zero bytes.

### Anything else?

`buf convert` itself is able to produce 0 length encoded message in binpb format. Therefore, according to the tool itself, 0 length encoding is valid. For example:

```
$ echo '' | \   
buf convert \
  foo.proto \
  --type=M \
  --from=-#format=txtpb \
  --to=-#format=binpb | \
wc
      0       0       0
```

doriable:
This removes the zero length check for `buf convert`, allowing empty
inputs for `binpb`, `txtpb`, and `yaml` formats. Empty is valid for these
formats. We return the unmarshal error for `json`, since null/empty is not
valid for json (we expect a json object).

Fixes #3663


private/buf/cmd/buf/buf_test.go
```diff
@@ -3596,18 +3595,65 @@ func TestConvertWithImage(t *testing.T) {
...
+	})
+	t.Run("no stdin input from json", func(t *testing.T) {
 		t.Parallel()
 		testRunStdoutStderrNoWarn(
 			t,
 			nil,
 			1,
 			"",
-			`Failure: --from: length of data read from "-" was zero`,
+			`Failure: --from: proto: syntax error (line 1:1): unexpected token`,
```
doriable, line 3650: Not a big fan of the error message returned from the json unmarshaler.
We could wrap the error:

https://github.com/bufbuild/buf/blob/4d05f30e35c0fbee2c384878b55d6adac4477fb6/private/pkg/protoencoding/json_unmarshaler.go#L45

```
func (m *jsonUnmarshaler) Unmarshal(data []byte, message proto.Message) error {
	options := protojson.UnmarshalOptions{
		Resolver:       m.resolver,
		DiscardUnknown: !m.disallowUnknown,
	}
  if err := options.Unmarshal(data, message); err != nil {
    return fmt.Errorf("json unmarshal: %w", err)
  }
  return nil
}
```

PR #3671, opened 2025-03-10
Add validation for verification URI in registry login

doriable:
This adds validation to ensure that the URL scheme
returned from the oauth handler is only `http`/`https`.

### DIFF
--- a/private/pkg/oauth2/client.go
+++ b/private/pkg/oauth2/client.go
@@ -254,7 +254,7 @@ func validateDeviceAuthorizationResponse(deviceAuthorizationResponse DeviceAutho
 func validateURIScheme(rawURI string) error {
 	parsed, err := url.Parse(rawURI)
 	if err != nil {
-		return err
+		return fmt.Errorf("oauth2: invalid verification URI, %s: %w", rawURI, err)
 	}
 	if parsed.Scheme != "https" && parsed.Scheme != "http" {
 		return fmt.Errorf("oauth2: invalid verification URI scheme, %q, received: %s", parsed.Scheme, parsed)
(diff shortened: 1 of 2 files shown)


private/pkg/oauth2/client_test.go
```diff
@@ -153,7 +153,39 @@ func TestAuthorizeDevice(t *testing.T) {
...
+				ClientSecret: "clientSecret",
+			},
+			transport: func(t *testing.T, r *http.Request) (*http.Response, error) {
+				testAssertFormRequest(t, r, url.Values{"client_id": {"clientID"}, "client_secret": {"clientSecret"}})
+				return testNewJSONResponse(t, http.StatusOK, `{"device_code":"deviceCode","user_code":"userCode","verification_uri":"https://example.com","verification_uri_complete":"file:///path/to/file.ext?code=userCode","expires_in":10,"interval":5}`), nil
+			},
+			err: fmt.Errorf(
+				`oauth2: invalid verification URI scheme, %q, received: %s`,
+				"file",
+				"file:///path/to/file.ext?code=userCode",
```
emcfarlane, line 186: Nit: why are both of these tests files and not a `http` one?

private/pkg/oauth2/client.go
```diff
@@ -237,3 +241,23 @@ func parseJSONResponse(response *http.Response, payload any) error {
...
+	if err := validateURIScheme(deviceAuthorizationResponse.VerificationURI); err != nil {
+		return err
+	}
+	return validateURIScheme(deviceAuthorizationResponse.VerificationURIComplete)
+}
+
+func validateURIScheme(rawURI string) error {
+	parsed, err := url.Parse(rawURI)
+	if err != nil {
+		return err
```
emcfarlane, line 257: Error here could return the same invalid verification URI error.

emcfarlane (approved):

bufdev (approved):

emcfarlane (approved):

PR #3673, opened 2025-03-10
Remove user plugin preference protos

smaye81:
This PR removes the v1alpha1 `Get/UpdateUserPluginPreferences` proto definitions. We no longer use these definitions on the BSR.


mfridman (approved):

doriable (approved):

proto/buf/alpha/registry/v1alpha1/user.proto
```diff
@@ -221,20 +213,3 @@ message UpdateUserSettingsRequest {
...
 
 message UpdateUserSettingsResponse {}
-
-message GetUserPluginPreferencesRequest {
-  reserved 1;
-  reserved "user_id";
-}
-
-message GetUserPluginPreferencesResponse {
-  UserPluginPreference preference = 1;
```
mfridman, line 231: Did we remove this as well, the `UserPluginPreference` or is that used?

smaye81: It's not used actually. It can be removed also: https://github.com/bufbuild/buf/pull/3674

PR #3679, opened 2025-03-11
Fix deprecated include types flag for buf generate

emcfarlane:
This fixes the flag `--include-types` for `buf generate`. This is a deprecated flag, replaced by `--type`. Currently this flag won't be applied. The behavior now allows for mixing of flag types, for example the following includes all 3 types:
```bash
buf generate --type=pkg.Foo --include-types=pkg.Bar --type=pkg.Baz
```


doriable (approved):
Thanks for catching this -- despite the small code change, I actually think it's a nice thing to call out in the changelog as a fixed behaviour.

doriable (approved):

PR #3690, opened 2025-03-13
Re-order docker tags to correct image version

pkwarren:
The first tag appears to be used as the
`org.opencontainers.image.version`, so re-order the tags so the full version comes first for release builds.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -108,7 +108,6 @@ jobs:
           tags: |
             type=edge,branch=main
             type=semver,pattern={{version}}
-            type=semver,pattern={{major}}.{{minor}}.{{patch}}
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}}
           flavor: |
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -107,10 +107,10 @@ jobs:
             bufbuild/buf
           tags: |
             type=edge,branch=main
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -107,10 +107,10 @@ jobs:
             bufbuild/buf
           tags: |
             type=edge,branch=main
-            type=semver,pattern={{major}}
-            type=semver,pattern={{major}}.{{minor}}
-            type=semver,pattern={{major}}.{{minor}}.{{patch}}
             type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}.{{patch}}
```
pkwarren, line 111: This looks to match the examples here: https://github.com/docker/metadata-action/tree/902fa8ec7d6ecbf8d84d538b9b233a880e428804?tab=readme-ov-file#semver

emcfarlane (commented):
Nice

emcfarlane (approved):

.github/workflows/ci.yaml
emcfarlane: Do these two get de-dupe? `{{version}}` and `{{major}}.{{minor}}.{{patch}}`?

pkwarren: Yeah - lemme remove. We shouldn't list them both times I don't think.

pkwarren: Fixed in 145ec7d.

emcfarlane (approved):

PR #3696, opened 2025-03-18
Add `--against-registry` flag for `buf breaking`

### LINKED ISSUE #3654: Improvements to running `buf breaking` for workspaces
### Feature

Right now there are some pain points to running `buf breaking` for workspaces:

- [ ] https://github.com/bufbuild/buf/issues/3641 -- when a new module is added to a workspace, `buf breaking` will error because there is an incongruence in the number of images between the input and the `--against`.
- [ ] There is currently no way to run `buf breaking` for a workspace against the registry. The user must run `buf breaking` for each module.
  - Proposing a `buf breaking --against-registry` flag to address this, that would have the following behaviour:
    - When set, this would compare each module in the input to its registry counterpart. This will be based on the `name` key in the module config.
    - If there is no way to determine a module's registry (e.g. no `name` key is set or it is a pre-built image), then the module will be ignored.
    - If there are no commits for the module on the registry (e.g. a new module), then the module will be ignored.

doriable:
This PR adds a `--against-registry` flag to `buf breaking` that can be
set in lieu of `--against`. When `--against-registry` is set, breaking checks
are run against the latest commit on the default label of the remote module
in the registry.

`buf breaking` will error if one of `--against` or `--against-registry` is
not set:

```
$ buf breaking
Failure: Must set --against or --against-registry
```

`buf breaking --against-registry` will error if not all modules in the input
are named (and thus a remote cannot be resolved):

```
$ buf breaking --against-registry
Failure: cannot use --against-registry with unnamed module, proto/local
```

Fixes #3654

### DIFF
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -54,10 +54,13 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
-// ImageWithConfig pairs an Image with lint and breaking configuration.
+// ImageWithConfig pairs an Image with its corresponding [bufmodule.Module] full name
+// (which may be nil), [bufmodule.Module] opaque ID, and lint and breaking configurations.
 type ImageWithConfig interface {
 	bufimage.Image
 
+	ModuleFullName() bufparse.FullName
+	ModuleOpaqueID() string
 	LintConfig() bufconfig.LintConfig
 	BreakingConfig() bufconfig.BreakingConfig
 	PluginConfigs() []bufconfig.PluginConfig
@@ -476,6 +479,8 @@ func (c *controller) GetTargetImageWithConfigsAndCheckClient(
 		imageWithConfigs := []ImageWithConfig{
 			newImageWithConfig(
 				image,
+				nil, // No module name for a single message ref
+				"",  // No module opaque ID for a single message ref
 				lintConfig,
 				breakingConfig,
 				pluginConfigs,
@@ -1170,6 +1175,8 @@ func (c *controller) buildTargetImageWithConfigs(
 			imageWithConfigs,
 			newImageWithConfig(
 				image,
+				module.FullName(),
+				module.OpaqueID(),
 				workspace.GetLintConfigForOpaqueID(module.OpaqueID()),
 				workspace.GetBreakingConfigForOpaqueID(module.OpaqueID()),
 				workspace.PluginConfigs(),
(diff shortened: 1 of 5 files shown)


private/buf/bufctl/controller.go
```diff
@@ -54,10 +54,13 @@ import (
...
 )
 
-// ImageWithConfig pairs an Image with lint and breaking configuration.
+// ImageWithConfig pairs an Image with its corresponding [bufmodule.Module] full name
+// (which may be nil), [bufmodule.Module] opaque ID, and lint and breaking configurations.
 type ImageWithConfig interface {
 	bufimage.Image
 
+	ModuleFullName() bufparse.FullName
+	ModuleOpaqueID() string
```
emcfarlane, line 63: Maybe this could be `Module() bufmodule.Module` and document that it may be nil? From the docs it seems like the `Module.FullName` might be empty (like in Module) not that it isn't a Module,  as the opaque ID can be empty.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -194,19 +208,58 @@ func run(
...
+		}
 	}
-	if len(imageWithConfigs) != len(againstImageWithConfigs) {
+	if flags.AgainstRegistry {
+		for _, imageWithConfig := range imageWithConfigs {
+			if imageWithConfig.ModuleFullName() == nil {
+				return fmt.Errorf(
+					"cannot use --%s with unnamed module, %s",
+					againstRegistryFlagName,
+					imageWithConfig.ModuleOpaqueID(),
```
emcfarlane, line 247: Do you need to check for empty opaque ID here?

doriable: Re: the comment above, the `OpaqueID` should always be present, but I could check for an empty and return a `syserror` here.

private/buf/bufctl/controller.go
doriable: Discussed outside of the PR: this is mostly for the case of `messageRef`, where we resolve the message through the bucket without the `Module`, so there is no `FullName` or `OpaqueID`. In those cases, we are not able to use `--against-registry` flag.

private/buf/cmd/buf/command/breaking/breaking.go
doriable: Adjusted this error for the `buffetch.MessageRef` case -- given that we don't build the module for those cases, `--against-registry` basically cannot be used.

emcfarlane (approved):

PR #3698, opened 2025-03-20
Fix path handling on windows

### LINKED ISSUE #3699: Fix Language server import path handling cross-platform
Uri's are currently being manually constructed and are non-compliant on windows. This results in the LSP incorrectly converting `file://` URIs to a valid file path.

Example: https://github.com/bufbuild/buf/blob/37b5a2a496a2279aa03507be64cba00193644a1e/private/buf/buflsp/file.go#L521.

go.lsp.dev/uri already has cross-platform handling of URIs - #3698 proposes to leverage this to resolve the issue.

christogav:
Ensure go.lsp.dev/uri is used for path handling to correctly handle paths cross-platform e.g. windows.

Closes #3699.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -38,6 +38,7 @@ require (
 	github.com/tetratelabs/wazero v1.9.0
 	go.lsp.dev/jsonrpc2 v0.10.0
 	go.lsp.dev/protocol v0.12.0
+	go.lsp.dev/uri v0.3.0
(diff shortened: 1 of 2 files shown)


doriable:
Hello, thanks for putting this up! Overall looks good to me, you'll need to run `make generate` and push to have lint cleared up.

PR #3706, opened 2025-03-24
LSP: Fix diagnostics not being published

christogav:
Language Server runs checks (lint, breaking) in a go routine which may not finish before diagnostics are published.

Add a PublishDiagnostics at the end of the go routine to ensure diagnostics are published.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -311,6 +311,9 @@ func (f *file) Refresh(ctx context.Context) {
 		f.BuildImages(ctx)
 		f.RunLints(ctx)
 		f.RunBreaking(ctx)
+
+		// Ensure diagnostics are published after checks are run.
+		f.PublishDiagnostics(ctx)
 	}()
 
 	progress.Report(ctx, "Indexing Symbols", 5.0/6)


private/buf/buflsp/file.go
```diff
@@ -311,6 +311,9 @@ func (f *file) Refresh(ctx context.Context) {
 		f.BuildImages(ctx)
 		f.RunLints(ctx)
 		f.RunBreaking(ctx)
+
+		// Ensure diagnostics are published after checks are run.
+		f.PublishDiagnostics(ctx)
```
doriable, line 316: So this is a bit of a tricky thing: if we add this, it basically means that every lint/breaking run will publish its own diagnostics when they finish, which in itself could be racy. Instead, if we rely on `f.PublishDiagnostics` at the base of each `Refresh` call (line 324), each top-level LSP call will lock, and each `Refresh` call will publish the latest lint/breaking diagnostics that ran.

This PR does surface the issue that if no subsequent call is made, then there is a case where diagnostics could simply not be published. The crux of this is that the checks should be synchronized to `Refresh`, and we are currently in the process of reworking the LSP's concurrency model to improve this.

chrisgavinator: I can see #3709 and #3710 raised that looks like a better addressing for this. Thanks! Will close.

christogav: I can see https://github.com/bufbuild/buf/pull/3709 and https://github.com/bufbuild/buf/pull/3710 raised that looks like a better addressing for this. Thanks! Will close.

PR #3707, opened 2025-03-24
Include `gitConfigAuthArgs` on `git checkout` when `filter` is set

CGA1123:
When using `filter` `buf` will fail to run `git checkout` as the credentials aren't passed to `git`.

This results in failures such as:

```
Failure: could not clone https://github.com/foo/bar.git: exit status 128
fatal: could not read Username for 'https://github.com': No such device or address
fatal: could not fetch <SHA> from promisor remote
```

I noticed this while testing bufbuild/buf-action#142 which relies on overriding `credential.helper` for authentication in GitHub Actions.

### DIFF
--- a/private/pkg/git/cloner.go
+++ b/private/pkg/git/cloner.go
@@ -181,7 +181,7 @@ func (c *cloner) CloneToBucket(
 		if err := execext.Run(
 			ctx,
 			"git",
-			execext.WithArgs("sparse-checkout", "set", options.SubDir),
+			execext.WithArgs(append(gitConfigAuthArgs, "sparse-checkout", "set", options.SubDir)...),
 			execext.WithEnv(app.Environ(envContainer)),
 			execext.WithStderr(buffer),
 			execext.WithDir(baseDir.Path()),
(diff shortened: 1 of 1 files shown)


private/pkg/git/cloner.go
```diff
@@ -193,10 +193,18 @@ func (c *cloner) CloneToBucket(
 	// Always checkout the FETCH_HEAD to populate the working directory.
 	// This allows for referencing HEAD in checkouts.
 	buffer.Reset()
+
+	// Include auth when using git-filters, so git can pull the required
+	// objects from the origin.
+	checkoutArgs := []string{"checkout", "--force", "FETCH_HEAD"}
+	if options.Filter != "" {
+		checkoutArgs = append(gitConfigAuthArgs, checkoutArgs...)
+	}
```
doriable, line 202: This approach seems overall reasonable -- we could address this alongside the `sparse-checkout`, for example:

```
if options.Filter != "" {
  checkoutArgs = append(gitConfigAuthArgs, checkoutArgs...)
  if options.SubDir != "" {
    // The sparse checkout of the subdir...
  }
}
```

doriable: Apologies, I should've commented this earlier -- going through it, it actually makes sense to prepend `gitConfigAuthArgs` to all invocations (since this is how we are fetching). So instead of checking this against `options.Filter != ""`, it should be prepended to all the `checkout` and `sparse-checkout` calls (similar to how we're prepending to the submodules call).

doriable (approved):

Issue #3725, opened 2025-03-31
Non-deterministic file dependencies


haines:
### GitHub repository with your minimal reproducible example (do not leave this field blank or fill out this field with "github.com/bufbuild/buf" or we will automatically close your issue, see the instructions above!)

https://github.com/haines/buf-dependencies-flapping

### Commands

```shell
# Runs `buf generate` repeatedly until a diff is produced
./repro
```

### Output

```shell
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
.
diff --git a/src/cerbos/policy/v1/policy_pb.ts b/src/cerbos/policy/v1/policy_pb.ts
index 3d8c9e5..b51cbe6 100644
--- a/src/cerbos/policy/v1/policy_pb.ts
+++ b/src/cerbos/policy/v1/policy_pb.ts
@@ -13,13 +13,14 @@ import type { OutputEntry, OutputEntryJson, Trace, TraceJson } from "../../engin
 import { file_cerbos_engine_v1_engine } from "../../engine/v1/engine_pb.js";
 import type { Value, ValueJson } from "@bufbuild/protobuf/wkt";
 import { file_google_protobuf_struct } from "@bufbuild/protobuf/wkt";
+import { file_protoc_gen_openapiv2_options_annotations } from "../../../protoc-gen-openapiv2/options/annotations_pb.js";
 import type { Message } from "@bufbuild/protobuf";
 
 /**
  * Describes the file cerbos/policy/v1/policy.proto.
  */
 export const file_cerbos_policy_v1_policy: GenFile = /*@__PURE__*/
[...]

PR #3734, opened 2025-04-01
Ensure deterministic order for filtered public imports

emcfarlane:
This ensures that public imports are always added to the image in a deterministic way. It changes the current order by first seen to order by file name. First seen is not valid as when including options we range over them in an [undefined order](https://pkg.go.dev/google.golang.org/protobuf@v1.36.6/reflect/protoreflect#Message) (although from testing it looks to always be in tag order).

Related to https://github.com/bufbuild/buf/issues/3506  and https://github.com/bufbuild/buf/issues/3725 . Both these issues link to the same error of adding unused imports. This was fixed in https://github.com/bufbuild/buf/pull/3727 . This PR fixes an edge case where if 2 or more public imports are resolved for a file, they must be appended to the current imports in a deterministic way.

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/image_filter.go
+++ b/private/bufpkg/bufimage/bufimageutil/image_filter.go
@@ -17,6 +17,7 @@ package bufimageutil
 import (
 	"fmt"
 	"slices"
+	"sort"
 	"strings"
 
 	"github.com/bufbuild/buf/private/bufpkg/bufimage"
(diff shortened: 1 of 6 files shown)


private/bufpkg/bufimage/bufimageutil/image_filter.go
```diff
@@ -266,26 +264,31 @@ func (b *sourcePathsBuilder) remapDependencies(
...
 	}
-	if importsRequired != nil {
-		newDependencies = append(newDependencies, importsRequired.keys()...)
+	// Add imports picked up via a public import. The filtered files do not use public imports.
+	if publicImportCount := len(importsRequired); publicImportCount > 0 {
+		for importPath := range importsRequired {
+			newDependencies = append(newDependencies, importPath)
+		}
+		// Sort the public imports to ensure the output is deterministic.
+		sort.Strings(newDependencies[len(newDependencies)-publicImportCount:])
```
doriable, line 288: This ordering isn't consistent with the current ordering... which means that we could be creating a breaking behaviour here, right?

emcfarlane: The current ordering is not deterministic. The issue https://github.com/bufbuild/buf/issues/3725 was re-raised as the unused imports (which should be just public imports) were being added in an undefined order. The issue brought up would still affect any file that has 2 or more public imports (which should be hopefully no-one), but this aims to solve that edge case.

doriable (approved):

PR #3736, opened 2025-04-02
Add file path to file annotations without source code info

### LINKED ISSUE #3715: Emit file name (when possible) even without source code info for lint/breaking file annotations
Prior to our migration to `bufplugin`, we would populate the file name for a file annotation emitted by lint/breaking checks, even if source code info was not sent:

```
foo/v1/foo.proto:1:1:RPC "Echo" has the same type "foo.v1.Message" for the request and response.
```

This is because we populated the file name from the image file rather than the descriptor's location.

Right now, we are using `bufcheckserverutil.ResponseWriter.AddProtosourceAnnotation`, which relies on the presence of `protosource.Location` to populate the file annotation: https://github.com/bufbuild/buf/blob/aee9553e6628c848cdafc30240f9644b0f44d314/private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverutil/response_writer.go#L22-L33

To fix this behaviour, we can rework this to attempt to populate the file name from the image file we're checking using `check.WithFileName` when locations are not available (https://github.com/bufbuild/bufplugin-go/blob/88f24f4a34243598cc973f16038e6acfff3c8f69/check/response_writer.go#L88-L100).

doriable:
This PR adds a back-up file path to `responseWriter.AddProtosourceAnnotation`,
so that in cases where `bufprotosource.Location` is `nil` (e.g. when the input does
not provide source code info), annotations will fallback and use the file path.

Currently, when no source code info is available, we get the following output:

```
<input>:1:1:RPC "Echo" has the same type "foo.v1.Message" for the request and response.
<input>:1:1:RPC request type "Message" should be named "EchoRequest" or "EchoServiceEchoRequest".
<input>:1:1:RPC response type "Message" should be named "EchoResponse" or "EchoServiceEchoResponse".
```

With this change, we will get:

```
foo/v1/foo.proto:1:1:RPC "Echo" has the same type "foo.v1.Message" for the request and response.
foo/v1/foo.proto:1:1:RPC request type "Message" should be named "EchoRequest" or "EchoServiceEchoRequest".
foo/v1/foo.proto:1:1:RPC response type "Message" should be named "EchoResponse" or "EchoServiceEchoResponse".
```

This is useful, because it allows us to respect `ignore` and `ignore_only`
configurations, since we will be able to filter the file annotations using the
file path.

Fixes #3715

### DIFF
--- a/private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverutil/response_writer.go
+++ b/private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverutil/response_writer.go
@@ -23,13 +23,14 @@ import (
 type ResponseWriter interface {
 	check.ResponseWriter
 
-	// AddProtosourceAnnotation adds a [check.Annotation] for the given
-	// [bufprotosource.Location]s. The input file name is provided, if available, as a
-	// back-up to populate the input file name in cases where source code info is not
-	// available, and so the [bufprotosource.Location]s are nil. In the case where the
-	// [bufprotosource.Location] is available, the input file name is ignored.
-	// The against location is not required, so we don't need to pass a back-up file path
-	// for against.
+	// AddProtosourceAnnotation adds a [check.Annotation] for the specified [bufprotosource.Location]s.
+	//
+	// If location is non-nil, it is used to populate the annotation's source code information,
+	// and inputFileName is ignored. If location is nil and inputFileName is provided, the
+	// latter is used as a fallback to set the input file name in the annotation.
+	// againstLocation is optional and may be nil.
+	//
+	// The format string and args are used to construct the annotation message.
 	AddProtosourceAnnotation(
 		location bufprotosource.Location,
 		againstLocation bufprotosource.Location,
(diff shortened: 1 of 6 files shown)


private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverutil/response_writer.go
```diff
@@ -23,10 +23,17 @@ import (
...
 	check.ResponseWriter
 
-	// AddProtosourceAnnotation adds a check.Annotation for bufprotosource.Locations.
+	// AddProtosourceAnnotation adds a [check.Annotation] for the given
+	// [bufprotosource.Location]s. The input file name is provided, if available, as a
+	// back-up to populate the input file name in cases where source code info is not
+	// available, and so the [bufprotosource.Location]s are nil. In the case where the
+	// [bufprotosource.Location] is available, the input file name is ignored.
+	// The against location is not required, so we don't need to pass a back-up file path
+	// for against.
```
emcfarlane, line 32:
```suggestion
	// AddProtosourceAnnotation adds a [check.Annotation] for the specified [bufprotosource.Location]s.
	//
	// If location is non-nil, it is used to populate the annotation's source code information,
	// and inputFileName is ignored. If location is nil and inputFileName is provided, the
	// latter is used as a fallback to set the input file name in the annotation.
	// againstLocation is optional and may be nil.
	//
	// The format string and args are used to construct the annotation message.
```

emcfarlane (approved):

PR #3737, opened 2025-04-02
Add `Policy` audit events, and missing `Plugin` events and fields

unmultimedio:
Add some events that the `plugins` don’t have but `repositories` do, for consistency:
- plugin deprecated
- plugin undeprecated
- plugin visibility changed

Add missing field `visibility` for `plugins` events to match the ones we’re already logging for `repositories` events.

Add the same events, with the same fields, for `policies`.

### DIFF
--- a/proto/buf/alpha/audit/v1alpha1/event.proto
+++ b/proto/buf/alpha/audit/v1alpha1/event.proto
@@ -49,6 +49,7 @@ enum ResourceType {
   RESOURCE_TYPE_REPOSITORY = 4;
   RESOURCE_TYPE_REPOSITORY_CONTRIBUTOR = 5;
   RESOURCE_TYPE_REPOSITORY_COMMIT = 6;
+  RESOURCE_TYPE_POLICY = 17;
   RESOURCE_TYPE_PLUGIN = 7;
   RESOURCE_TYPE_PLUGIN_COMMIT = 15;
   RESOURCE_TYPE_CURATED_PLUGIN = 8;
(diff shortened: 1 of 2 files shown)


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -419,6 +422,17 @@ message PayloadRepositoryDefaultBranchChanged {
...
 
+message PayloadPolicyCreated {
+  // owner_id is the id of the owner of the policy.
+  string owner_id = 1;
+  // owner_name is the name of the owner of the policy.
+  string owner_name = 2;
+  // visibility is the visibility of the policy.
+  buf.alpha.registry.v1alpha1.Visibility visibility = 3;
+  // description of the policy.
+  string description = 4;
```
emcfarlane, line 433: Why are visibility and description set?

unmultimedio: It's additional useful fields in the payload. We already track visibility in other events like [the `CreateRepository` payload](https://github.com/bufbuild/buf/blob/bb812fa169445ab73d8eabad26d8f8a39f99478e/proto/buf/alpha/audit/v1alpha1/event.proto#L291-L292), and I'd be inclined to add it to [the `CreatePlugin` event](https://github.com/bufbuild/buf/blob/bb812fa169445ab73d8eabad26d8f8a39f99478e/proto/buf/alpha/audit/v1alpha1/event.proto#L436) too.

emcfarlane: These shared fields should be the same between these entities.

bufdev (approved):
I defer to @emcfarlane

doriable (approved):

PR #3740, opened 2025-04-04
Init bufpolicy pkg for policies

emcfarlane:
This initializes the bufpolicy pkg to manage polices. The pkg is not currently used and theres no change in behaviour.

### DIFF
--- a/private/bufpkg/bufpolicy/errors.go
+++ b/private/bufpkg/bufpolicy/errors.go
@@ -1,0 +1,70 @@
+// Copyright 2020-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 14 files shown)


private/bufpkg/bufpolicy/errors.go
```diff
@@ -0,0 +1,70 @@
...
+)
+
+// DigestMismatchError is the error returned if the Digest of a downloaded Policy
+// does not match the expected digest in a buf.lock file.
+type DigestMismatchError struct {
+	FullName       bufparse.FullName
+	CommitID       uuid.UUID
+	ExpectedDigest Digest
+	ActualDigest   Digest
+}
```
doriable, line 32: I don't think this is something that should happen in this PR, but thoughts on moving `DigestMismatchError` into a shared package under `bufpkg`, similar to `bufparse`?

private/bufpkg/bufpolicy/policy_visibility.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+)
+
+const (
+	// PolicyVisibilityPublic says the Policy is public on the registry.
+	PolicyVisibilityPublic = iota + 1
+	// PolicyVisibilityPrivate says the Policy is private on the registry.
+	PolicyVisibilityPrivate
+)
```
doriable, line 26: Same thing with visibility as `DigestMismatchError`.

doriable (approved):

PR #3741, opened 2025-04-04
Pass CRC32 option to Zip

unmultimedio:


unmultimedio:
Closing, CRC value is already written in the header by the underlying zip library.

PR #3759, opened 2025-04-09
Add `UpdateOrganizationGroup`

saquibmian:
Primarily to support updating the role override. Also add associated audit event.

### DIFF
--- a/proto/buf/alpha/audit/v1alpha1/event.proto
+++ b/proto/buf/alpha/audit/v1alpha1/event.proto
@@ -81,6 +81,7 @@ enum EventType {
   EVENT_TYPE_ORGANIZATION_MEMBER_ROLES_CHANGED = 50;
   EVENT_TYPE_ORGANIZATION_MEMBER_REMOVED = 5;
   EVENT_TYPE_ORGANIZATION_IDP_GROUP_ADDED = 21;
+  EVENT_TYPE_ORGANIZATION_IDP_GROUP_UPDATED = 59;
   EVENT_TYPE_ORGANIZATION_IDP_GROUP_REMOVED = 22;
   EVENT_TYPE_REPOSITORY_CREATED = 6;
   EVENT_TYPE_REPOSITORY_DEPRECATED = 48;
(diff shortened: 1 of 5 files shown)


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -292,6 +294,19 @@ message PayloadOrganizationIDPGroupAdded {
   buf.alpha.registry.v1alpha1.OrganizationRole role_override = 3;
 }
 
+message PayloadOrganizationIDPGroupUpdated {
+  // organization_id is the id of the organization with the updated IDP group.
+  string organization_id = 1;
+  // organization_name is the name of the organization with the updated IDP group.
+  string organization_name = 2;
+  // old_role_override is the role override associated with the IDP updated group. Only set if the
```
mfridman, line 302: Wouldn't these always be set, otherwise what are we auditing? I.e., and organization_id/name without a role change means nothing?

saquibmian: We might go from `<nothing>` to `<something>`, in which case `old_role_override` will not be set, and vice versa.

mfridman: Ah okay, so at least one of these is guaranteed to be set.

saquibmian: Yup. I could document that... Should I rephrase to use `UNSPECIFIED` instead?

proto/buf/alpha/registry/v1alpha1/organization.proto
```diff
@@ -278,6 +282,20 @@ message AddOrganizationGroupRequest {
...
 message AddOrganizationGroupResponse {}
 
+message UpdateOrganizationGroupRequest {
+  // The ID of the organization to which this group belongs.
+  string organization_id = 1;
+  // The name of the group to update.
+  string group_name = 2;
+  // The role to associate with this organization group.
+  //
+  // Setting this to 'UNSPECIFIED' will remove the override. Leaving this unset will not update this
```
mfridman, line 292: Should this error if it's unset, make it mandatory either explicitly UNSPECIFIED or the desired updated role override?

saquibmian: Generally, the mutable fields in an Update request are optional so that users don't have to update everything in one shot, they can update just the one thing they care about. It would be more obvious in this case if there was more than one thing to update, but I didn't want to deviate from that style. See https://github.com/bufbuild/registry-proto/blob/2f065e825828d75db601227b33d824a952b9363e/buf/registry/owner/v1/user_service.proto#L164-L180 for example.

mfridman (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
mfridman: I think it's fine, I misread the comment initially.

PR #3760, opened 2025-04-09
Filter out `GIT_DIR` from cloner environment variables

### LINKED ISSUE #3752: buf breaking in pre-commit will set core.bare to 'true' and failed

_No response_

### What's up?

When I run `pre-commit run --all-files` or `buf breaking --against "ssh://git@github.com/provistaai/protobuf.git#branch=main"`, It's all right. But when I run `git commit`, It will auto set core.bare to 'true' and failed.

Image

It's a submodule and my `.pre-commit-config.yaml` is:
```yaml
repos:
  - repo: local
    hooks:
      - id: buf-format
        name: Format Protobuf Files
        entry: buf format -w
        language: system
        files: \.proto$
        pass_filenames: false

      - id: buf-breaking
        name: Check Protobuf Breaking Changes
        entry: buf breaking --against "ssh://git@github.com/provistaai/protobuf.git#branch=main"
        language: system
        files: \.proto$
        pass_filenames: false
```
Is that a bug? What should I do? Thx a lot

doriable:
In the case where `buf` is being run in an environment where `GIT_DIR`
is set, e.g. within a submodule, `git init` will defer to using the `GIT_DIR`
rather than creating its own. We want to avoid mutating an existing `GIT_DIR`
through `CloneToBucket`, and should instead be treating it as a stand-alone
clone, so we filter out `GIT_DIR` from the environment variables.

An example of this came up in #3752 where a submodule is running `buf` in
repository-local pre-commit hooks.

Fixes #3752

### DIFF
--- a/private/pkg/git/cloner.go
+++ b/private/pkg/git/cloner.go
@@ -76,16 +76,12 @@ func (c *cloner) CloneToBucket(
 
 	depthArg := strconv.Itoa(int(depth))
 
-	// In the case where this is being run in an environment where GIT_DIR is set, e.g. within
-	// a submodule, we want to treat this as a stand-alone clone rather than interacting with
-	// an existing GIT_DIR. So we filter out GIT_DIR from our environment variables.
-	filteredEnv := make(map[string]string)
-	envContainer.ForEachEnv(func(key, value string) {
-		if key != "GIT_DIR" {
-			filteredEnv[key] = value
-		}
+	envContainer = app.NewEnvContainerWithOverrides(envContainer, map[string]string{
+		// In the case where this is being run in an environment where GIT_DIR is set, e.g. within
+		// a submodule, we want to treat this as a stand-alone clone rather than interacting with
+		// an existing GIT_DIR. So we filter out GIT_DIR from our environment variables.
+		"GIT_DIR": "",
 	})
-	envContainer = app.NewEnvContainer(filteredEnv)
 
 	baseDir, err := tmp.NewDir(ctx)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


doriable:
Testing against the repro provided in #3752:

The behaviour prior to the change, when attempting to commit in the submodule, it would result in the parent `.git/modules/app/protos/config` to set `core.bare = true`.
```
$ git commit
Format Protobuf Files....................................................Passed
Check Protobuf Breaking Changes..........................................Failed
- hook id: buf-breaking
- exit code: 1

Failure: could not clone https://github.com/Wu-Jiayang/test-b.git: exit status 3
warning: core.bare and core.worktree do not make sense
error: remote origin already exists.

$ cat .git/modules/app/protos/config
[core]
        repositoryformatversion = 0
        filemode = true
        bare = true
...
```
This is because `git init` is a `GIT_DIR` path set to the submodule, which contains a `.git` file pointing to the parent `.git` directory:

```
GIT_DIR=/Users/dkeung/src/playground/buf-3752/test-a/.git/modules/app/protos

~/src/playground/buf-3752/test-a/app/protos @ $ cat .git
gitdir: ../../.git/modules/app/protos
```

And thus the subsequent fetches/checkouts affected the parent config.

With the change in the PR applied, we are able to get a stand-alone clone and the pre-commit hook was able to run:

```
~/src/playground/buf-3752/test-a/app/protos @ $ git commit
Format Protobuf Files....................................................Passed
Check Protobuf Breaking Changes..........................................Failed
- hook id: buf-breaking
- exit code: 100
- files were modified by this hook

content.proto:17:3:Field "1" with name "category_name" on message "CategoryStayDuration" changed type from "string" to "int64".
```

PR #3769, opened 2025-04-14
Add buf policy YAML file config

emcfarlane:
This adds support for buf policy configuration. A new field in `buf.yaml` files (`policies`)  takes an array of `PolicyConfig` and a new external file type `BufPolicyYAMLFile` for defining a policy.


private/bufpkg/bufconfig/buf_policy_yaml_file_test.go
```diff
@@ -0,0 +1,114 @@
...
+`,
+		// expected output
+		`version: v2
+`,
+	)
+
+	testReadWriteBufPolicyYAMLFileRoundTrip(
+		t,
+		// input
+		`version: v2
```
unmultimedio, line 42: Curious, this is the first policy file version, why we're starting at v2?

emcfarlane: The file will consume v2 lint/breaking config options, so we went with v2 to clarify its related to the v2 spec.

private/bufpkg/bufconfig/buf_policy_yaml_file.go
```diff
@@ -0,0 +1,284 @@
...
+var (
+	bufPolicyYAMLFileNames                       = []string{defaultBufPolicyYAMLFileName}
+	bufPolicyYAMLFileNameToSupportedFileVersions = map[string]map[FileVersion]struct{}{
+		defaultBufPolicyYAMLFileName: {
+			FileVersionV2: struct{}{},
+		},
+	}
+)
+
+// BufPolicyYAMLFile represents a buf.policy.yaml file.
```
doriable, line 42: Instead of `buf.policy.yaml`, could we say something like:

```
BufPolicyYAMLFile represents a policy configuration YAML file.
```

Just because this is meant to be generic.

doriable (commented):
I don't think `buf_policy_yaml_file.go` should live under `bufpkg/bufconfig`, because it is not a a configuration file for `buf`. It is much closer in nature to `bufremotepluginconfig` (https://github.com/bufbuild/buf/tree/main/private/bufpkg/bufremoteplugin/bufremotepluginconfig) where it is the configuration of an specific entity. In this case, it is also the content of said entity. So I would move the policy YAML under `bufpolicy` as a sub-package, `bufpolicyconfig`, and go from there (`policy_config.go` and the changes to `buf.yaml` would stay in `bufconfig`).

doriable (approved):

PR #3775, opened 2025-04-17
Run modernize tool to update to newer APIs

pkwarren:
Run the modernize tool on buf's codebase to migrate to newer APIs and perform other migrations (e.g. `interface{}` to `any`).

### DIFF
--- a/private/buf/bufcli/flags_args.go
+++ b/private/buf/bufcli/flags_args.go
@@ -17,6 +17,7 @@ package bufcli
 import (
 	"errors"
 	"fmt"
+	"slices"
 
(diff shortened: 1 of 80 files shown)


private/buf/bufcli/flags_args.go
```diff
@@ -17,6 +17,7 @@ package bufcli
 import (
 	"errors"
 	"fmt"
+	"slices"
```
pkwarren, line 20: FYI - I had this waiting a while but noticed that formatting wasn't being enforced properly so wanted to fix https://github.com/bufbuild/buf/pull/3774 first.

emcfarlane (approved):

bufdev (approved):

PR #3776, opened 2025-04-17
Add repository idp group audit resource type

saquibmian:
Missed this in the previous PR.


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -49,6 +49,7 @@ enum ResourceType {
   RESOURCE_TYPE_REPOSITORY = 4;
   RESOURCE_TYPE_REPOSITORY_CONTRIBUTOR = 5;
   RESOURCE_TYPE_REPOSITORY_COMMIT = 6;
+  RESOURCE_TYPE_REPOSITORY_IDP_GROUP = 18;
```
mfridman, line 52: Most a forward-thinking question while this is fresh in our minds. Do you think this could be collapsed to just a `RESOURCE_TYPE_IDP_GROUP`, afaics there's nothing differentiating org and repo, and I don't see anything in the future that could?

mfridman (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
saquibmian: They are distinct resources internally, with different ID (but the same name). Technically when we update a repository idp group, it's not the idp group that's updated, it's the mapping that's updated. If we wanted to reduce the surface area here, could definitely do it, we would re-design the API to be idp-group centric and just map an idp group to a repository, with some optional configuration. Then resource type would be idp group, and the events would be idpgroupmappingadded, idpgroupmappingupdated, etc.

I don't know if that makes sense...

mfridman: Yeah okay, mainly jotting this down as a thought if we ever consider porting this over to bufbuild/registry-proto.

I suspect the IdP group could be a well-defined resource in that world, and you could map that to organizations, repositories, etc.

I don't think it's worth refactoring v1alpha1 just for repository mapping, so the duplication is the least-worst option for now.

bufdev (approved):

PR #3780, opened 2025-04-21
Add support for policies in buf.lock

emcfarlane:
This PR adds support for policies in the `buf.lock` file. The lock file includes a new field `policies` which declares a set of policy keys and their plugin keys. The workspace exposes two new methods: `RemotePolicyKeys ` returns the policy keys, and `RemotePluginKeysForPolicy` returns the plugin keys for the policy, if any.


doriable (approved):
The PR looks good, stamping under the assumption that the merge conflict will be resolved :)

PR #3784, opened 2025-04-22
Rename `AddRepositoryGroup.repository_role` to `role_override`

saquibmian:
This matches all the other places where we refer to this, must have been stale on my end.

No users of this API, so this is safe to do.

### DIFF
--- a/proto/buf/alpha/registry/v1alpha1/repository.proto
+++ b/proto/buf/alpha/registry/v1alpha1/repository.proto
@@ -406,7 +406,7 @@ message AddRepositoryGroupRequest {
   // The name of the group to add.
   string group_name = 2;
   // The role to associate with any user who is added to the repository via this group.
-  RepositoryRole repository_role = 3;
+  RepositoryRole role_override = 3;
 }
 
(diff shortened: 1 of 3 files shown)


proto/buf/alpha/registry/v1alpha1/repository.proto
```diff
@@ -406,7 +406,7 @@ message AddRepositoryGroupRequest {
   // The name of the group to add.
   string group_name = 2;
   // The role to associate with any user who is added to the repository via this group.
-  RepositoryRole repository_role = 3;
+  RepositoryRole role_override = 3;
```
mfridman, line 409: Can you add a quick comment update what happens if this is not set. Something closer to organizations,

```proto
// Any user who maps to this organization via this group will be granted this role. If no role is
// set, the user will be granted the 'member' role.
OrganizationRole role_override = 3;
```

saquibmian: Yup

mfridman (approved):

PR #3785, opened 2025-04-22
Run buf Policies locally

emcfarlane:
This PR adds support for running local buf Policies. Lint and breaking checks now invoke the set of policy rules alongside the set of configured rules. Only local policies are currently supported, remote will be added with lock file support. The annotations emitted now include the policy name alongside the plugin (e.g. `<annotation> (buf-plugin-name, buf-policy-name)`). An inbuilt plugin for a policy will include only the policy name. 

As an example a policy can be expressed for the `buf-plugin-suffix.wasm` plugin in a `policy.yaml` file:
```yaml
version: v2
lint:
  use:
    - SERVICE_BANNED_SUFFIXES
    - RPC_BANNED_SUFFIXES
plugins:
  - plugin: buf-plugin-suffix.wasm
    options:
      service_banned_suffixes:
        - Mock
        - Test
      rpc_banned_suffixes:
        - Element
```
Which can then be used in a `buf.yaml` file:
```
version: v2
lint:
  disable_builtin: true
policies:
  - policy: policy.yaml
```
And run with `buf lint` showing the output annotations for a simple test proto:
```
a.proto:8:1:Service name "AMock" has banned suffix "Mock". (buf-plugin-suffix.wasm, policy.yaml)
a.proto:9:3:Method name "AMock.GetAllAElement" has banned suffix "Element". (buf-plugin-suffix.wasm, policy.yaml)
```


private/buf/bufctl/controller.go
```diff
@@ -801,6 +808,13 @@ func (c *controller) GetCheckClientForWorkspace(
 	if err != nil {
 		return nil, err
 	}
+	bucket, err := c.storageosProvider.NewReadWriteBucket(
+		".",
+		storageos.ReadWriteBucketWithSymlinksIfSupported(),
+	)
```
doriable: A couple of things here:
- This currently implies that the local policies configured in the `buf.yaml` are read relative to `.`, which is the working directory of the command. This should not be the case, right? It should be read relative to the root of the workspace (basically where the `buf.yaml` is). In the `bufcheck` tests below, you're using workspace/module read bucket, which is correct.
- We should be able to map the storage bucket to `.wasm` and `.yaml` files only as a small optimization.

private/buf/bufworkspace/workspace.go
```diff
@@ -81,6 +82,14 @@ type Workspace interface {
 	//
 	// These come from the buf.lock file. Only v2 supports plugins.
 	RemotePluginKeys() []bufplugin.PluginKey
+	// PolicyConfigs gets the configured PolicyConfigs of the Workspace.
+	//
+	// These come from the buf.yaml files.
+	PolicyConfigs() []bufconfig.PolicyConfig
```
doriable, line 88: The above comment may be moot depending on this discussion ^^"
So right now, we are resolving local `PolicyData` at the `bufcheck.Client` layer... which is mostly reasonable, except that it results in the passing of a storage bucket, which is kind of awkward. I think it mostly makes sense to only fetch `PolicyData` if/when we need it, but we should not be passing random buckets around like this (since it's prone to bugs, like the comment above).

I think the way to address this is that in `workspace_provider.go`, when we set `PluginConfig` and `PolicyConfig`, the paths should be mapped so that they can be read based on the execution context of the command. Then you won't need to pass the bucket around. This may also be something that should be addressed as a part of workspace targeting/mapping the modules, happy to discuss this in detail!

doriable (commented):
This mostly makes sense to me, left some comments around the plugin/policy config paths and avoiding passing the read bucket.

private/buf/bufctl/controller.go
emcfarlane: Loading has been made to match plugin loading. A loader func is provided. Local paths are not relative to the workspace but the cwd. This follows plugins which may be also absolute (and in the PATH).

private/buf/bufworkspace/workspace.go
doriable: Resolving this comment based on our discussion: we are going to follow the same path logic as plugins.

private/bufpkg/bufcheck/lint_test.go
```diff
@@ -1382,13 +1545,19 @@ func testLintWithOptions(
 		logger,
 		bufcheck.ClientWithRunnerProvider(bufcheck.NewLocalRunnerProvider(wasmRuntime)),
 		bufcheck.ClientWithLocalWasmPluginsFromOS(),
+		bufcheck.ClientWithLocalPoliciesFromOS(),
```
doriable: We overwrite this below, can remove ^^"

private/bufpkg/bufcheck/bufcheck.go
```diff
@@ -267,6 +275,23 @@ func ClientWithRemoteWasmPlugins(
...
 }
 
+// ClientWithLocalPolicies returns a new ClientOption that specifies reading local policies.
+//
+// The readFile is used to read the policy data from the filesystem.
+func ClientWithLocalPolicies(readFile func(string) ([]byte, error)) ClientOption {
+	return func(clientOptions *clientOptions) {
+		clientOptions.policyReadFile = readFile
+	}
+}
```
doriable, line 285: Nit: should we specify in the godoc that this option is used to support testing, and that we should use `ClientWithLocalPoliciesFromOS`?

emcfarlane: It could be used by the BSR to load policies from storage. This may need to change to a more structured form to avoid serializing again since we no longer use the serialized form in transport/storage.

PR #3788, opened 2025-04-23
Support local bufplugins for protoc-gen plugins

### LINKED ISSUE #3786: Support custom plugins in `protoc-gen-buf-*`
### Feature

Right now, the protoc plugins do not support custom lint/breaking change plugins:

https://github.com/bufbuild/buf/blob/a8ec9a6b7397b037819b5c6959102993f2d84303/private/buf/cmd/protoc-gen-buf-lint/lint.go#L99-L103

`rules_buf` uses those plugins, so by extension it is not possible to use custom plugins with `rules_buf` rules.

emcfarlane:
This PR adds support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint` protoc plugins. Remote plugins (e.g. `buf.build/acme/plugin`) are not supported.

Closes #3786


doriable:
This looks reasonable for supporting local bufplugins for `protoc-gen-{lint, breaking}`, pending some discussions around whether we can support remote plugins also.

doriable (approved):
Based on our discussions, this makes sense, and would unlock the use of custom bufplugins with Bazel rules, so stamping. Follow-up work will be done to support the local plugin executables in the Bazel rules.

mkanat:
Looks like this never got threaded through into the rules? I don't see any way to use plugins, still.

emcfarlane:
@mkanat see the [check plugin docs](https://buf.build/docs/bsr/checks/plugins/) on how to create and use custom plugins.

mkanat:
Yes, I know how to make them. But this was not piped through to the bazel rules for buf linting.

PR #3789, opened 2025-04-23
Release binaries for riscv64 linux platform

kxxt:
This PR adds binary releases for riscv64 linux platform.

I find running `make bufrelease` a little slow. So as a bonus, I made the build parallel in a separate commit(Though it doesn't bring much benefits for GitHub CI because it only offers poor runners with merely 4 cores).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add `CSR` category to breaking rules.
 - Add support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
+- Add RISC-V (64-bit) binaries for Linux to releases.
 
 ## [v1.53.0] - 2025-04-21
 
--- a/make/buf/scripts/release.bash
+++ b/make/buf/scripts/release.bash
@@ -25,6 +25,7 @@ goos() {
 goarch() {
   case "${1}" in
     x86_64) echo amd64 ;;
+    riscv64) echo riscv64 ;;
     arm64) echo arm64 ;;
     aarch64) echo arm64 ;;
     armv7) echo arm ;;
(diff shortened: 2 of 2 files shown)


kxxt:
Hi @emcfarlane, could you review this pull request? Thanks!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Add `CSR` category to breaking rules.
 - Add support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
+- Add binary releases for riscv64 linux platform.
```
doriable, line 7:
```suggestion
- Add RISC-V (64-bit) binaries for Linux to releases.
```

doriable (commented):
Thank you for contributing this PR, after going through it, we're happing to add `RISC-V` linux binaries to our release process. I left a small nit on the changelog statement and we would prefer not to add the parallelism in this case, given that it would not provide a material gain in GitHub Actions, which is the primary way we orchestrate our release process, so I'd like to request that we remove it.

Once these changes are in, we can get this merged, and the binaries will be available for our next release. Thanks!

kxxt:
> Thank you for contributing this PR, after going through it, we're happing to add `RISC-V` linux binaries to our release process. 

Thanks very much!

> I left a small nit on the changelog statement and we would prefer not to add the parallelism in this case, given that it would not provide a material gain in GitHub Actions, which is the primary way we orchestrate our release process, so I'd like to request that we remove it.

I have dropped that commit. If github provides larger runners in the future, you are welcome to pick it up.

> Once these changes are in, we can get this merged, and the binaries will be available for our next release. Thanks!

Thanks! This will unblock some use cases of bazel on riscv64 as `rules_buf` uses prebuilt binary in this repo.

doriable (approved):

PR #3790, opened 2025-04-23
Add policy pushed event

unmultimedio:
It includes the `[#next-free-field: x]` annotation for large audit events that keep growing as we audit more events. Seems to be a standard annotation: https://github.com/search?q=content%3Anext-free-field+language%3Aproto&type=code


doriable (approved):
I think it's reasonable to keep track of `#next-free-field` since the audit log structures are becoming quite large. lgtm!

PR #3792, opened 2025-04-23
Add `GetSDKInfo` rpc to `ResolveService`

doriable:
This adds a `GetSDKInfo` rpc to the `ResolveService`.
This RPC will provide the SDK information based on the provided
module reference, plugin information, and optionally SDK version string.

If a specific module reference (commit or label) is and/or a plugin version
is specified, then the SDK for the specified information is resolved. If the
SDK version string is provided, then it will be validated with the module and
plugin information.

This allows us to replace the use of the subsequent RPCs in `buf registry sdk version`,
and make a single call to `GetSDKInfo` to resolve the SDK version information.
It also allows us to easily provide module commit and plugin version
information given the module, plugin, and SDK version string.

### DIFF
--- a/proto/buf/alpha/registry/v1alpha1/resolve.proto
+++ b/proto/buf/alpha/registry/v1alpha1/resolve.proto
@@ -311,5 +311,5 @@ message GetSDKInfoResponse {
   PluginInfo plugin_info = 2;
   // The SDK version string. The format is based on the SDK registry supported by the
   // provided plugin.
-  string version = 3;
+  string sdk_version = 3;
 }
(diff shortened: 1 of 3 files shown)


mfridman (approved):

proto/buf/alpha/registry/v1alpha1/resolve.proto
```diff
@@ -257,3 +273,43 @@ message GetRemotePackageVersionPlugin {
...
+    // The semver plugin version. This will always be valid semver.
+    string version = 3;
+    // The plugin revision.
+    uint32 plugin_revision = 4;
+  }
+  ModuleInfo module_info = 1;
+  PluginInfo plugin_info = 2;
+  // The SDK version string. The format is based on the SDK registry supported by the
+  // provided plugin.
+  string version = 3;
```
mfridman, line 314:
```suggestion
  string sdk_version = 3;
```

mfridman (approved):

PR #3804, opened 2025-04-29
Add buf policy push command

emcfarlane:
This PR adds the command `buf policy push` for uploading Policies to the BSR.

### DIFF
--- a/private/buf/cmd/buf/command/policy/policypush/policypush.go
+++ b/private/buf/cmd/buf/command/policy/policypush/policypush.go
@@ -1,0 +1,240 @@
+// Copyright 2020-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


private/buf/cmd/buf/command/policy/policypush/policypush.go
```diff
@@ -0,0 +1,204 @@
...
+	if len(flags.Labels) > 0 {
+		options = append(options, bufpolicy.UploadWithLabels(flags.Labels...))
+	}
+	if flags.SourceControlURL != "" {
+		options = append(options, bufpolicy.UploadWithSourceControlURL(flags.SourceControlURL))
+	}
+	commits, err := uploader.Upload(ctx, []bufpolicy.Policy{policy}, options...)
+	if err != nil {
+		return nil, err
+	}
```
doriable, line 166: Ah apologies, I think I missed this in the initial review:

https://github.com/bufbuild/buf/blob/b4e76be7fd0af4a9164d82d9ffe6b867918e15fd/private/bufpkg/bufpolicy/bufpolicyapi/uploader.go#L275

We should not be using `policy.Description()` for the policy description for the create request, `Description` in this context is similar to that of `Module.Description()`, which is used to provide a descriptive output in the context of the CLI.

Similar to the module uploader, we should create the policy with an empty description.

emcfarlane: Thanks! Fixed

PR #3805, opened 2025-05-01
Update protovalidate to v0.12.0

jchadwick-buf:
The only actual change that matters is to the `resolve` helpers which are now fallible. However, even though the `resolve` helpers are technically fallible, right now the "correct" way to handle it is essentially to treat it the same as absence.


private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -163,7 +163,7 @@ func checkField(
 	if err != nil {
 		return err
 	}
-	constraints := resolve.FieldRules(fieldDescriptor)
+	constraints, _ := resolve.FieldRules(fieldDescriptor)
```
nicksnyder, line 166: We should update the code to handle potential future errors. The fact that it never returns an error today is an implementation detail we shouldn’t rely on as a client.

jchadwick-buf: Definitely not a time-bomb: the `resolve` helpers actually can already return a non-`nil` error today, it's just that from a protovalidate perspective it isn't interesting, we treat it as-if there are simply no rules at all (same before and after this PR.) As far as Buf goes, returning an error here will definitely change the existing behavior, but probably not for the better, because the only case where this should ever return an error versus just `nil, nil` is the case where the protobuf can't be re-cast. Since we don't and have never made breaking wire changes to the protovalidate schema, the only reason this can happen is because of an extension number conflict, i.e. a proto file is using the same extension number as protovalidate but for an incompatible proto. In this case, the errors that we might get could be rather unpredictable depending on how things line up. In the rather unlikely event that someone runs across this error case, it will almost assuredly be an inactionable false positive.

The error cases might be interesting to someone but to us it just tells us that we're definitely not dealing with protovalidate annotations.

doriable: This is super helpful context -- any chance we could capture a super short summary of this in a comment along with the change?

emcfarlane: Mind expanding on the error cases? Presumably this would fail if the user put the wrong type or something in the  protovalidate cases ([from the marshalling here](https://github.com/bufbuild/protovalidate-go/blob/8fa707802a9ac7e6fa9860b23c0b09b76344cf29/resolve/resolve.go#L57-L81)), would we not want to error as bufcheck should check those cases?

doriable (approved):
Looks good! I just left a quick comment about annotating the call-sites to say why we are absorbing the error (so we don't need to check blame for context, haha). I'm pre-approving, since I think this is good to merge with a reasonable comment. Thank you for taking care of this!

EDIT: based on the discussion in the thread, let's return the error from the resolver layer. While I think it makes sense to treat that defensively in terms of the lint UX when making this change, we are already returning errors through our check layers, so given that the underlying resolver is introducing this error, it's safe/reasonable to return it here.

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
emcfarlane: I guess my main concern is if the error chance is really low, why not return the error? Having a silent error swallow is always difficult to debug as even if unrelated it's a potential path. An extension conflict seems really unlikely, so why not error in that case?

doriable: > I guess my main concern is if the error chance is really low, why not return the error?

I think this is a very fair point. I think @jchadwick-buf was taking the conservative approach of not breaking the `buf lint` UX, since from this is something introduced to protovalidate's resolve behaviour and is an error that would only concern protovalidate's resolve API. My point was mostly that we would need to document that in-line carefully if we want to merge this as-is.

But also, I think it's fair to say that this is very much an edge case error, and that since we are already surfacing errors at this layer, it's also fine to surface, and maybe this is just a far clearer behaviour, at least in code. Given this comment thread, it might be prudent to surface the error in lieu of additional documentation instead. I'll edit my top-level comment also.

emcfarlane (approved):

jchadwick-buf:
I believe I have resolved all of the concerns so I will merge. Thanks.

PR #3810, opened 2025-05-02
Fix config migration for removed rules

emcfarlane:
This PR fixes unknown rules after a config migration by filtering any used IDs by the target v2 rule and categories ID set.


doriable (approved):
Looks good to me! We should add a changelog for this.

PR #3811, opened 2025-05-02
Add ppc64le architecture to release script

Jenkins-J:
Add ppc64le architecture to releases. This is necessary for the rules_buf bazel module to work properly on the ppc64le architecture as a binary release is required by the rules_buf_toolchains target.


doriable:
This has been superseded by #3840 which adds a `ppc64le` Linux binary to the releases, so closing this.
The build will be available in the upcoming release!

PR #3819, opened 2025-05-05
Fix exclude filter behavior with empty image file

### LINKED ISSUE #3795: Empty file causes buf generate failure
In my buf module, I have a file `acme/v1/stub.proto`:

```
syntax = "proto3";

package acme.v1;

import "google/rpc/code.proto";
import "google/rpc/error_details.proto";
import "google/rpc/status.proto";
``` 

Buf cli will crash on it with `Failure: missing "acme/v1/stub.proto"`. I'm assuming this is because there are no types in there.

doriable:
When we filter an image, if a file has no types, instead of error-ing,
we should simply continue since it is valid for a file after filtering to
contain no types. If the file should be included, then we should allow
for including an empty file.

Fixes #3795

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - Add `CSR` category to breaking rules.
 - Add support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
 - Add RISC-V (64-bit) binaries for Linux to releases.
-- Fix issue with excluding types and empty files when running `buf generate`.
+- Fix type filtering on `buf generate` for empty files, files with no declared types.
 
 ## [v1.53.0] - 2025-04-21
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add `CSR` category to breaking rules.
 - Add support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
 - Add RISC-V (64-bit) binaries for Linux to releases.
+- Fix issue with excluding types and empty files when running `buf generate`.
 
 ## [v1.53.0] - 2025-04-21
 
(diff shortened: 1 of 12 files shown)


doriable:
If we want the behaviour to still include the empty file, then we should not error on this condition:

https://github.com/bufbuild/buf/blob/5f3a4d77f40852436473fa04d616619a76085d95/private/bufpkg/bufimage/bufimageutil/bufimageutil.go#L453-L456

and instead just allow for file descriptors with no types defined with them.

doriable:
> If we want the behaviour to still include the empty file, then we should not error on this condition:
> 
> https://github.com/bufbuild/buf/blob/5f3a4d77f40852436473fa04d616619a76085d95/private/bufpkg/bufimage/bufimageutil/bufimageutil.go#L453-L456
> 
> and instead just allow for file descriptors with no types defined with them.

We are taking this approach, since including an empty file should be valid, and so we should not be error-ing here. Instead, we should simply move forward with processing the file.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -451,15 +451,16 @@ func (t *transitiveClosure) addElement(
...
-		}
-		// A file includes all elements. The types are resolved in the image index
-		// to ensure all nested types are included.
-		for _, typeName := range typeNames {
-			typeInfo := imageIndex.ByName[typeName]
-			if err := t.addElement(typeInfo.element, "", false, imageIndex, opts); err != nil {
-				return err
+		// If the file we are attempting to include is empty (has no types defined), it is still
+		// valid for inclusion.
+		if ok {
```
emcfarlane, line 456: It's fine to range over the empty slice, can remove `ok` var and indentation.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Add `CSR` category to breaking rules.
 - Add support for local bufplugins for `protoc-gen-buf-breaking` and `protoc-gen-buf-lint`.
 - Add RISC-V (64-bit) binaries for Linux to releases.
+- Fix issue with excluding types and empty files when running `buf generate`.
```
emcfarlane, line 8: Nit: Fix type filtering on `buf generate` for empty files, files with no declared types.

emcfarlane (approved):

emcfarlane (approved):

PR #3820, opened 2025-05-06
Update appcmdtesting to use functional options

bufdev:
The old style is a 2019 relic.
Port testing to use new `appcmdtesting.Run` with the requisite options.


bufdev:
All lint failures are due to me marking the `appcmdtesting.RunCommand.*` functions as deprecated (and they should be removed before this PR is merged).

doriable (approved):

emcfarlane (approved):

PR #3831, opened 2025-05-09
Allow users to set examples without constraints in `PROTOVALIDATE` lint rule

### LINKED ISSUE #3830: Still check example values set with protovalidate even if there are no constraints to validate

_No response_

### What's up?

It is a pretty reasonable use-case for protovalidate examples (e.g. https://buf.build/docs/reference/protovalidate/rules/string_rules/#example) to use them without protovalidate constraints.

Currently, we surface a lint violation:

```
{"path":"...,"start_line":#,"start_column":#,"end_line"#:,"end_column":#,"type":"PROTOVALIDATE","message":"example value is specified by there are no constraints defined"}
```

We should still validate the example set, but no longer surface this file annotation.

doriable:
Currently when a user attempts to set protovalidate examples
without constraints, `buf lint` returns a violation:

```
{"path":"...,"start_line":#,"start_column":#,"end_line"#:,"end_column":#,"type":"PROTOVALIDATE","message":"example value is specified by there are no constraints defined"}
```

However, it should be valid to set examples without constraints.
This removes the constraints check when validating examples.
If a user sets a valid example with no constraints, no violation is surfaced.
If a user sets an invalid example type, then this is caught in the constraint
type check.

Fixes #3830


private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
```diff
@@ -758,11 +758,6 @@ func checkExampleValues(
 			return true
 		})
 	}
-	if !hasRules {
```
oliversun9, line 761: nit: if there are no rules, could we return early here?

Alfus (approved):

emcfarlane (approved):

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/field.go
doriable: Yeah -- given that we already check that the appropriate example type is set, all other errors would be compilation errors when building the image (e.g. if I set something like `string id = 1 [(buf.validate.field).string.example = 42];`). So yeah, this makes sense, pushed an update!

oliversun9 (approved):

PR #3864, opened 2025-06-05
Allow workspaces adding new module(s) with no module-specifc breaking configurations to run `buf breaking`

### LINKED ISSUE #3641: Support adding new modules with buf breaking
### Feature

Adding a module to a workspace is not technically a breaking change. When creating a new module and running `buf breaking` it will conservatively fail with an error message like:
```
Failure: input contained <N> images, whereas against contained <N+X> images
```

The current workaround is to skip breaking change on the creation of new modules. This issue often occurs on first setup of a repository too, when the against is empty.  Ideally we can avoid failing breaking change detection by matching the set of previous images and allowing any unmatched images to pass without error.

doriable:
This PR adds support to `buf breaking` for workspace inputs that
are adding new module(s). Formerly, we enforced the condition that
the count of images in the input must match the against:

```
Failure: input contained <N+X> images, whereas against contained <N> images
```

This is because of module-specific breaking configurations. However it was
not very ergonomic for users adding new modules to their workspaces. To
improve the UX, for the case where the input contains new module(s), if the
input does not have module-specific breaking configurations, we attempt to
map the against images to the input images, and any images in the input not
mapped are considered "newly added" and ignored.

We still error in the following cases:
- if there are more against images than input images
- if not all against images are mapped to an input image
- if there are any unique module-specific breaking configurations in the input
  that we need to respect

Fixes #3641

This PR also removes some `TODO` from `workspace_test.go` to investigate
path errors -- the errors are being returned from our storage packages,
and so we expected normalised paths.


private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -339,3 +351,103 @@ func validateFlags(flags *flags) error {
...
+	}
+	for ignoreID, paths1 := range breakingConfig1.IgnoreIDOrCategoryToPaths() {
+		paths2, ok := breakingConfig2.IgnoreIDOrCategoryToPaths()[ignoreID]
+		if !ok {
+			return false
+		}
+		if slices.Compare(paths1, paths2) != 0 {
+			return false
+		}
+	}
```
emcfarlane: I think you could use here:
```go
maps.EqualFunc(
    breakingConfig1.IgnoreIDOrCategoryToPaths(),
    breakingConfig2.IgnoreIDOrCategoryToPaths(), 
    slices.Equal[[]string],
)
```
And then squash the if statements to lines separated by `&&` if you wanted.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -339,3 +351,103 @@ func validateFlags(flags *flags) error {
...
+// Checks if the specified [bufconfig.BreakingConfig]s are equal. Returns true if both
+// [bufconfig.BreakingConfig] have the same configuration parameters.
+func equalBreakingConfig(breakingConfig1, breakingConfig2 bufconfig.BreakingConfig) bool {
+	if breakingConfig1.Disabled() != breakingConfig2.Disabled() {
+		return false
+	}
+	if breakingConfig1.FileVersion() != breakingConfig2.FileVersion() {
+		return false
+	}
+	if slices.Compare(breakingConfig1.UseIDsAndCategories(), breakingConfig2.UseIDsAndCategories()) != 0 {
```
emcfarlane: Should use `slices.Equal`

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -339,3 +351,103 @@ func validateFlags(flags *flags) error {
...
+
+// A helper function for filtering out images from [imagesWithConfig] if there is no corresponding
+// image in [againstImages]. We determine this based on image file path.
+//
+// This assumes that len(imageWithConfigs) > len(againstImages).
+// We also expect that each image in [againstImages] is mapped only once to a single
+// imageWithConfig in [imagesWithConfig]. If an againstImage is found, then we don't check
+// it again. We also validate that each image in [againstImages] is mapped to an imageWithConfig
+// from [imageWithConfigs].
+func filterImagesNotInAgainstImages(
```
emcfarlane: Helper name include with configs: `filterImageWithConfigsNotInAgainstImages`

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -339,3 +351,103 @@ func validateFlags(flags *flags) error {
...
+	imageWithConfigs []bufctl.ImageWithConfig,
+	againstImages []bufimage.Image,
+) ([]bufctl.ImageWithConfig, error) {
+	foundAgainstImageIndices := make(map[int]struct{})
+	var filteredImageWithConfigs []bufctl.ImageWithConfig
+	for _, imageWithConfig := range imageWithConfigs {
+		for _, imageFile := range imageWithConfig.Files() {
```
emcfarlane, line 410: This is stable per invocation, but could change depending on the DAG order of the files. For example, if the first file matches against A, but the second file matches against B. Do we need to exhaustively check that all the against paths match the same image or no image at all (deleted)?

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -339,3 +351,103 @@ func validateFlags(flags *flags) error {
...
+			if foundImage != nil {
+				filteredImageWithConfigs = append(filteredImageWithConfigs, imageWithConfig)
+				break
+			}
+		}
+	}
+	// If we are unsuccessful in mapping all againstImages to a unique imageWithConfig, then
+	// we return the same
+	if len(foundAgainstImageIndices) != len(againstImages) || len(againstImages) != len(filteredImageWithConfigs) {
+		return nil, newInputAgainstImageCountError(len(imageWithConfigs), len(againstImages))
```
emcfarlane, line 431: Maybe this should be more informative that the image was deleted/unmatched (and is therefore a breaking change?).

private/buf/cmd/buf/command/breaking/breaking.go
doriable: So I considered this issue, but I think this is a problem/instability we have independent of whether we are doing matching. Even in a scenario where we have the same number of images, the roots can change and you could end up in a state where a file path has changed modules.

I had an implementation that exhaustively checked all the filed paths at first, but I couldn't come to a sane UX for handling roots shifting, so I figured it would be better to surface those as breaking changes (e.g. a file was "deleted" from the module). Sorry this explanation might be a bit confusing, let me know if there's a way I can better clarify it ^^"

private/buf/cmd/buf/command/breaking/breaking.go
doriable: Hmm, so I waffled on whether I wanted to provide a different error message here or if simply providing the known error message was better. Since this isn't actually part of the check, I wasn't sure if returning an error message that appears to be a "breaking change" makes sense -- especially since it would only happen in a very specific subset of cases.

emcfarlane (approved):

PR #3871, opened 2025-06-13
Add support for plugin path overrides in `protoc-gen-buf-{lint,breaking}`

### LINKED ISSUE #3867: Add a way to supply custom buf plugin paths to `protoc-gen-buf-*`
### Feature

`rules_buf` uses `protoc-gen-buf-*` plugins to support lint and breaking checks in Bazel. To be able to use custom buf plugins we will need a way to pass the custom plugin paths to `protoc-gen-buf-*`.

One way to do this is to use a map to override the path to custom plugins in `protoc-gen-buf-*` config. We can add a new config option called `plugin_paths` or `plugin_overrides` that is a map of plugin names (as they appear in `buf.yaml`) to the path of the plugin. This will let `rules_buf` pass the plugin paths.

This feature is similar to `protoc`'s `--plugin` option.

doriable:
This PR adds support to `protoc-gen-buf-{lint,breaking}` for overriding
local plugin paths. `protoc-gen-buf-{lint,breaking}` do not support running
remote check plugins, only local.

The primary use-case for this is Bazel, which runs `protoc-gen-buf-{lint,breaking}`
in a sandboxed environment. `rules_buf` needs to be able to provide
path overrides to local check plugin binaries.

If an override for a plugin config is not provided, if the plugin config is
a remote plugin, we return an error, otherwise we attempt to use the
plugin config as-is.

Fixes #3867

### DIFF
--- a/private/buf/cmd/internal/internal.go
+++ b/private/buf/cmd/internal/internal.go
@@ -19,6 +19,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"path/filepath"
 
 	"buf.build/go/app"
 	"buf.build/go/app/appext"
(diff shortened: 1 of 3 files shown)


private/buf/cmd/internal/internal.go
```diff
@@ -28,6 +29,71 @@ import (
...
+	case 0:
+		return nil, nil, nil, fmt.Errorf("no module found for %q", module)
+	case 1:
+		return moduleConfigsFound[0], pluginConfigs, allCheckConfigs, nil
+	default:
+		return nil, nil, nil, fmt.Errorf("multiple modules found at %q, specify its full name as <remote/owner/module> instead", module)
+	}
+}
+
 // GetModuleConfigForProtocPlugin gets ModuleConfigs for the protoc plugin implementations.
```
emcfarlane: Is this method still used?

doriable: Oh yes, it is not -- I need to remove it,  thank you!

srikrsna-buf (approved):

doriable:
Tested this against a local Bazel test setup, things look good. Going to merge.

PR #3886, opened 2025-06-17
Add `--all` flag for `buf export`

### LINKED ISSUE #3870: Add `buf export --all` flag that provides all files
### Feature

Right now `buf export` only returns proto sources. We want to add an `--all` flag that would return all other files included for the input, such as docs and license files.

`buf export` by default includes imports -- for imports, non-proto files will _not_ be included with the `--all` flag. The `--all` flag is meant only for the input/target module(s).

`--all` is compatible with `--exclude-imports`: non-proto files for the input will be returned, and the imports will be excluded in that case.

doriable:
Add `buf export --all` flag for exporting non-proto sources,
such as documentation and license files with `buf export`.

Fixes #3870

### DIFF
--- a/private/buf/cmd/buf/command/export/export.go
+++ b/private/buf/cmd/buf/command/export/export.go
@@ -17,8 +17,10 @@ package export
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io/fs"
 	"os"
+	"strings"
 
 	"buf.build/go/app/appcmd"
 	"buf.build/go/app/appext"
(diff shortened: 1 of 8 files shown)


bufdev (approved):

private/buf/cmd/buf/command/export/export.go
```diff
@@ -121,6 +126,12 @@ func (f *flags) Bind(flagSet *pflag.FlagSet) {
 		"",
 		`The buf.yaml file or data to use for configuration`,
 	)
+	flagSet.BoolVar(
+		&f.All,
+		allFlagName,
+		false,
+		`When set, include any available documentation and license files for the exported input. If the input has more than one module, then the documentation and license file names will be prefixed with the module name.`,
```
emcfarlane:
```suggestion
		`When set, include any available documentation and license files for the exported input. If the input has more than one module, then the documentation and license file names will be suffixed with the module name.`,
```

emcfarlane (approved):

PR #3887, opened 2025-06-17
Ensure pre-commit hooks are kept up-to-date

### LINKED ISSUE #3882: Update pre-commit hook language version and check for pre-commit hook compatibility during CI
### Feature

We've had a couple of breakages with pre-commit hooks in the past, where the `go.mod` versions have been incremented, but the language versions in pre-commit hooks were not updated, #3880, #3750.

We should include updating pre-commit hooks as part of our deps upgrade process and add checks in CI to ensure compatibility.

doriable:
This PR adds a script to check that pre-commit hooks are up-to-date
with `go.mod` versions and if not, update them, as a part of `make upgrade`.

Fixes #3882

### DIFF
--- a/make/buf/all.mk
+++ b/make/buf/all.mk
@@ -204,8 +204,8 @@ newtodos:
 newtodofiles:
 	@bash make/buf/scripts/newtodos.bash | grep -v FUTURE | cut -f 1 -d : | sort | uniq
 
-.PHONY: checkprecommithooks
-checkprecommithooks:
-	@bash make/buf/scripts/checkprecommithooks.bash
+.PHONY: checkandupdateprecommithooks
+checkandupdateprecommithooks:
+	@bash make/buf/scripts/checkandupdateprecommithooks.bash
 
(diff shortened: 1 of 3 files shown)


emcfarlane (approved):

stefanvanburen:
> WIP: add a script for updating the `.pre-commit-hooks.yaml` in the case
> where it is out-of-date.

drive-by, but maybe could add to the [`updategoversion` Makefile target](https://github.com/bufbuild/buf/blob/120922eba6cdbff66802d85d0bab3d615df60c36/make/buf/all.mk#L190-L192)?

doriable:
> drive-by, but maybe could add to the [updategoversion Makefile target](https://github.com/bufbuild/buf/blob/120922eba6cdbff66802d85d0bab3d615df60c36/make/buf/all.mk#L190-L192)?

That's a good call, going to update, thank you!

emcfarlane (approved):

PR #3896, opened 2025-06-23
Fix breaking against registry for empty repositories

emcfarlane:
This fixes `buf breaking --against-registry` for newly created modlues when the default branch contains no commits. Treat the module as empty, therefore no breaking changes.

Related to #3641


private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -258,6 +261,15 @@ func run(
 				bufctl.WithConfigOverride(flags.AgainstConfig),
 			)
 			if err != nil {
+				if connect.CodeOf(err) == connect.CodeFailedPrecondition && strings.Contains(err.Error(), "no commits in the default label") {
```
bufdev: There must be a way to do this that is structured using ErrorDetails. Doing a string match on an error message doesn't seem right.

emcfarlane: Pushed a workaround for now. Will update the registry-protos to annotate the use of ErrorDetails and update the Registry to use them.

private/buf/cmd/buf/command/breaking/breaking.go
```diff
@@ -251,9 +261,42 @@ func run(
...
+										Owner:  moduleFullName.Owner(),
+										Module: moduleFullName.Name(),
+									},
+								},
+							},
+						},
+					},
+				),
+			); err != nil {
+				if connect.CodeOf(err) == connect.CodeFailedPrecondition {
```
bufdev, line 286: Why is an empty response returning failed precondition? Why not not found?

Also I would probably use list with a page size of 1, and see if the returned list has 0 elements.

emcfarlane: Failed precondition is due to being able to resolve the resource (a Module), but not able to return a Commit for it. The ListCommits endpoint was not used as it needs to resolve the default label, then using List would return a 404, as the default Label resource does not exist, and can't resolve.

doriable (approved):
This looks good to me, thank you! As discussed and noted in the comment, once error details become available, the check will be updated.

PR #3898, opened 2025-06-24
Export MarshalPolicyConfigAsJSON for policies

emcfarlane:
This exposes the stable JSON marshal for the policy configuration.

### DIFF
--- a/private/bufpkg/bufpolicy/digest.go
+++ b/private/bufpkg/bufpolicy/digest.go
@@ -17,18 +17,12 @@ package bufpolicy
 import (
 	"bytes"
 	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"slices"
 	"strconv"
 	"strings"
 
-	pluginoptionv1 "buf.build/gen/go/bufbuild/bufplugin/protocolbuffers/go/buf/plugin/option/v1"
-	"buf.build/go/bufplugin/option"
-	"buf.build/go/standard/xslices"
 	"github.com/bufbuild/buf/private/bufpkg/bufcas"
-	"github.com/bufbuild/buf/private/bufpkg/bufconfig"
 	"github.com/bufbuild/buf/private/bufpkg/bufparse"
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufpolicy/policy_config.go
```diff
@@ -27,3 +36,204 @@ type PolicyConfig interface {
...
+// It is a valid JSON encoding of the type buf.registry.policy.v1beta1.PolicyConfig.
+func MarshalPolicyConfigAsJSON(policyConfig PolicyConfig) ([]byte, error) {
+	return marshalPolicyConfigAsJSON(policyConfig)
+}
+
+/// *** PRIVATE ***
+
+// marshalPolicyConfigAsJSON marshals the given PolicyConfig to a stable JSON representation.
+//
+// This is used for the O1 digest and should not be used for other purposes.
```
unmultimedio: May be this part of the comment is not necessary? We're exporting it precisely because we'll use it in the BSR

private/bufpkg/bufpolicy/digest.go
```diff
@@ -221,7 +215,7 @@ func (*digest) isDigest() {}
 
 // getO1Digest returns the O1 digest for the given PolicyConfig.
 func getO1Digest(policyConfig PolicyConfig) (Digest, error) {
```
unmultimedio, line 217: Pretty sure we're going to be needing this one too? it's easy enough to do it ourselves with the `bufcas` pkg too. I guess it's just a matter of making sure use use the same `DigestType` everywhere.

emcfarlane: You should get the digest through the `Policy.Digest(o1)` method, so don't think it's necessary.

private/bufpkg/bufpolicy/policy_config.go
emcfarlane: Fixed.

unmultimedio (approved):

PR #3900, opened 2025-06-27
Add s390x binaries for Linux to releases and fix ppc64le release binaries

jonathan-albrecht-ibm:
Add s390x binaries for Linux to releases.

@doriable and @Jenkins-J I also fixed the ppc64le binaries which are being released as x86_64 binaries since I needed to change the same code. In a previous PR #3840 the bash function `goarch()` wasn't updated with the `ppc64le` value so the default `x86_64` ended up being used. 

Note that the `goarch()` function returns 1 if the arch wasn't found but since `goarch()` is only executed in a subshell, it doesn't fail and exit the script which I think was the intent. So I also made the `goarch()` and `goos()` functions return the string `"unsupported"` if the arch or os is not found so that the script will fail and exit. I'm happy to provide more details if you have more questions on this. It took me a while to figure out exactly what was going on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 - Add `buf export --all` flag to include non-proto source files.
+- Add s390x binaries for Linux to releases.
+- Fix ppc64le binaries for Linux released as x86_64 binaries.
 
 ## [v1.55.1] - 2025-06-17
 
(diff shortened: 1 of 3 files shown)


doriable:
Thank you for submitting this change, lgtm! Going to merge and the binary will be available on the next release.

jonathan-albrecht-ibm:
TFTR @doriable!

PR #3909, opened 2025-07-01
Fix usage formatting in webpages command

stefanvanburen:
Fixes a minor issue where usage containing `%` characters would be assumed to be a Go `fmt` format character. Instead, write out the raw contents and avoid any attempt at formatting (since we never provide args here).


private/bufpkg/bufcobra/markdown.go
```diff
@@ -262,7 +262,9 @@ func writeFlags(f *pflag.FlagSet, writer io.Writer) error {
 		}
 		p(" {#%s}", flag.Name)
 		p("\n")
-		p(usage)
```
emcfarlane, line 265: `p("%s", usage)` would work too?

emcfarlane (approved):

PR #3912, opened 2025-07-04
Add audit log for policy enforcements

emcfarlane:
This adds audit log events for policy enforcements. The resource type is either server, owner (user/organization) or module. This aligns with current policy event types like `EVENT_TYPE_SERVER_UNIQUENESS_POLICY_ENABLED` that apply to the server resource.

### DIFF
--- a/proto/buf/alpha/audit/v1alpha1/event.proto
+++ b/proto/buf/alpha/audit/v1alpha1/event.proto
@@ -157,7 +157,7 @@ enum EventType {
   EVENT_TYPE_POLICY_ENFORCEMENT_CREATED = 74;
   EVENT_TYPE_POLICY_ENFORCEMENT_DELETED = 75;
   EVENT_TYPE_POLICY_ENFORCEMENT_POLICY_CHANGED = 76;
-  EVENT_TYPE_POLICY_ENFORCEMENT_INCLUDE_TARGETS_CHANGED = 77;
+  EVENT_TYPE_POLICY_ENFORCEMENT_TARGETS_CHANGED = 77;
 }
 
 // EventMetadata provides additional details about the audited event.
@@ -997,8 +997,10 @@ message PayloadPolicyEnforcementCreated {
   string enforcement_id = 1;
   // policy_id is the id of the policy that is enforced.
   string policy_id = 2;
+  // policy_ref is the reference of the policy that is enforced.
+  string policy_ref = 3;
   // include_targets are the include targets of the policy enforcement.
-  repeated string include_targets = 3;
+  repeated string include_targets = 4;
 }
 
 message PayloadPolicyEnforcementDeleted {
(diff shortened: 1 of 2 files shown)


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -154,6 +154,10 @@ enum EventType {
   EVENT_TYPE_SDK_PLUGIN_CONSTRAINT_CREATED = 71;
   EVENT_TYPE_SDK_PLUGIN_CONSTRAINT_UPDATED = 72;
   EVENT_TYPE_SDK_PLUGIN_CONSTRAINT_DELETED = 73;
+  EVENT_TYPE_POLICY_ENFORCEMENT_CREATED = 74;
+  EVENT_TYPE_POLICY_ENFORCEMENT_DELETED = 75;
+  EVENT_TYPE_POLICY_ENFORCEMENT_POLICY_CHANGED = 76;
+  EVENT_TYPE_POLICY_ENFORCEMENT_INCLUDE_TARGETS_CHANGED = 77;
```
unmultimedio: Wondering if we should have just a "targets changed" event, and whenever/if we add "exclude targets", it's just a few more fields we add to the appropriate events.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -983,3 +991,35 @@ message PayloadSDKPluginConstraintDeleted {
   // plugin_name is the name of the plugin of the SDK constraint.
   string plugin_name = 8;
 }
+
+message PayloadPolicyEnforcementCreated {
+  // enforcement_id is the id of the policy enforcement.
+  string enforcement_id = 1;
+  // policy_id is the id of the policy that is enforced.
+  string policy_id = 2;
```
unmultimedio, line 999: We need `policy_ref` too.

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -983,3 +991,35 @@ message PayloadSDKPluginConstraintDeleted {
...
+  string enforcement_id = 1;
+}
+
+message PayloadPolicyEnforcementPolicyChanged {
+  // enforcement_id is the id of the policy enforcement.
+  string enforcement_id = 1;
+  // old_policy_id is the id of the old policy.
+  string old_policy_id = 2;
+  // new_policy_id is the id of the new policy.
+  string new_policy_id = 3;
```
unmultimedio, line 1017: Same here, missing the refs. It's possible that the change is only that:

```diff
- buf.build/acme/some-policy:foo
+ buf.build/acme/some-policy:bar
```

proto/buf/alpha/audit/v1alpha1/event.proto
emcfarlane: I like it, fixed!

proto/buf/alpha/audit/v1alpha1/event.proto
emcfarlane: Added

unmultimedio (approved):

doriable (approved):

bufdev (approved):

PR #3916, opened 2025-07-09
Move bufpolicy PolicyConfig methods types to bufpolicy

emcfarlane:
This PR defines the `PolicyConfig` types `LingConfig`, `BreakingConfig` and `PluginConfig` in the `bufpolicy` package. Methods to convert to the `bufconfig` type are provided in `bufpolicyconfig`. Unblocks https://github.com/bufbuild/buf/pull/3780 see https://github.com/bufbuild/buf/pull/3780#pullrequestreview-2994454558

### DIFF
--- a/private/bufpkg/bufpolicy/policy_config.go
+++ b/private/bufpkg/bufpolicy/policy_config.go
@@ -179,6 +179,13 @@ func MarshalPolicyConfigAsJSON(policyConfig PolicyConfig) ([]byte, error) {
 	return marshalPolicyConfigAsJSON(policyConfig)
 }
 
+// UnmarshalJSONPolicyConfig unmarshals the given JSON data into a PolicyConfig.
+//
+// Data is a valid JSON encoding of the type buf.registry.policy.v1beta1.PolicyConfig.
+func UnmarshalJSONPolicyConfig(registry string, data []byte) (PolicyConfig, error) {
+	return unmarshalJSONPolicyConfig(registry, data)
+}
+
 // *** PRIVATE ***
 
 type policyConfig struct {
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufpolicy/digest_test.go
```diff
@@ -17,65 +17,84 @@ package bufpolicy
...
-	}
-	testPolicyConfigO1Digest(t, policyConfig, "o1:db2906a09cca66da39f800207c75a8a2134d1c7918eca793e19ab07d76bea7a8f1282827205a6b7013ee0c1196af4ae7ef3abc6c41dab039bc24153d2e2dc4af")
+	require.NoError(t, err)
+	policyConfig, err := newPolicyConfig(lintConfig, breakingConfig, nil)
+	require.NoError(t, err)
+
+	testPolicyConfigO1Digest(
+		t,
+		policyConfig,
+		"o1:960327c2f0e3382b3e2caee5a5b8043b7a17287e82dc06a5c3a5ed76773e8622c18ddeead96bb5abe44b340965b49103462f4e5a1c11dbd05d1188d9afd7cff8",
```
unmultimedio, line 50: Curious, why did it change if the passed IDs are the same as the previous test?

private/bufpkg/bufpolicy/policy_config_test.go
```diff
@@ -0,0 +1,150 @@
...
+
+func testRoundTripPolicyConfigAsJSON(t *testing.T, policyConfigV1Beta1 *policyv1beta1.PolicyConfig) {
+	const registry = "bufbuild.test"
+	policyConfig, err := bufpolicyapi.V1Beta1ProtoToPolicyConfig(registry, policyConfigV1Beta1)
+	require.NoError(t, err)
+	data, err := bufpolicy.MarshalPolicyConfigAsJSON(policyConfig)
+	require.NoError(t, err)
+	require.NotEmpty(t, data)
+	var policyConfigV1Beta1Copy policyv1beta1.PolicyConfig
+	require.NoError(t, protoencoding.NewJSONUnmarshaler(nil, protoencoding.JSONUnmarshalerWithDisallowUnknown()).Unmarshal(data, &policyConfigV1Beta1Copy))
```
unmultimedio, line 147: I think we should we encapsulate this, mirroring the existing one.

```go
bufpolicy.MarshalPolicyConfigAsJSON(PolicyConfig) ([]byte, error)
bufpolicy.UnmarshalJSONPolicyConfig([]byte) (PolicyConfig, error)
```

private/bufpkg/bufpolicy/digest_test.go
emcfarlane: This changed from the JSON representation.

unmultimedio: Ah right, the snake -> camel case. Thx!

doriable (approved):

private/bufpkg/bufpolicy/policy_config_test.go
emcfarlane: I don't think this would be useful in the general case, as we need the proto form for API responses (and in this test for comparison against the input).

unmultimedio: Right, but we also need the go type, when retrieving the policy data stored as JSON, and when running an actual policy.

emcfarlane: Added methods to convert between the marshalled and the go type.

private/bufpkg/bufpolicy/policy_config.go
```diff
@@ -390,6 +397,68 @@ func marshalPolicyConfigAsJSON(policyConfig PolicyConfig) ([]byte, error) {
 	return data, nil
 }
 
+func unmarshalJSONPolicyConfig(registry string, data []byte) (PolicyConfig, error) {
```
unmultimedio, line 400: This works, thanks! I wouldn't have a problem either if this returned the `*policyv1beta1.PolicyConfig` type instead.

PR #3934, opened 2025-07-23
Add bufcheck support for remote policies

emcfarlane:
This PR adds support to bufcheck.Clients to invoke remote Policies with remote Plugins. Remote Policies are pulled from a registry and stored locally in the buf cache. Remote plugins are supported for both Local and Remote Policies. These are resolved through the policy's plugin providers (see the [buf.lock file changes](https://github.com/bufbuild/buf/pull/3780)).

Invoking a remote policy will show the PolicyRef, and the PluginRef if a custom plugin, per each lint annotation. For example with a remote policy that enforces a custom plugin:
```
$ buf lint
local.proto:9:1:Service name "acme.v1.example.LibraryServiceTest" has banned suffix "Test". (buf.build/acme/buf-plugin-suffix, buf.build/acme/buf-policy-suffix:v1)
```
Policies that invoke just a built-in rule do not have a plugin reference `(buf.build/acme/buf-policy-suffix:v1)`

### DIFF
--- a/private/buf/bufworkspace/workspace_provider.go
+++ b/private/buf/bufworkspace/workspace_provider.go
@@ -206,6 +206,13 @@ func (w *workspaceProvider) GetWorkspaceForModuleKey(
 					return nil, err
 				}
 			}
+			// TODO: Supporting Policies for ModuleKeys requires resolving each policies remote
+			// plugins.  This is not currently supported. Instead of requiring the resolved Keys for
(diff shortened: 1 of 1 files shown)


doriable (commented):
This looks good, a quick follow-up comment (can be done here or in a subsequent PR): would be nice to have a couple of integration tests in `buf_test.go` that runs against a policy, and we can check things like the input being a subdir of a workspace, etc. But I think that the lint and breaking tests on the file annotations are good for this PR.

PR #3955, opened 2025-08-11
Missing scope in enforcement audit events

unmultimedio:


bufdev (approved):

doriable (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -1004,13 +1031,19 @@ message PayloadPolicyEnforcementCreated {
...
-  // enforcement_id is the id of the policy enforcement.
-  string enforcement_id = 1;
+  // scope is the scope to which the policy was enforced.
+  PolicyEnforcementScope scope = 1;
+  // policy_id is the id of the policy that was enforced.
+  string policy_id = 2;
+  // policy_ref is the reference of the policy that was enforced.
+  string policy_ref = 3;
+  // include_targets were the include targets of the policy enforcement.
+  repeated string include_targets = 4;
```
emcfarlane, line 1041: For other archive/delete we don't include the full set of fields, just the changeset. Why is this needed?

emcfarlane (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: I included the enforcement content (what's being deleted) to be consistent with plugin delete event [here](https://github.com/bufbuild/core/pull/20204/files#diff-de2b5ac5a6ddf4c473d346530ab908ee47f9e4aff1132c2745f80fd3e6605aabR38), and with repository delete event [here](https://github.com/bufbuild/buf/blob/5f4c37bbb6c7e80cf8612ac90b89d6ad9901d986/proto/buf/alpha/audit/v1alpha1/event.proto#L366-L376).

For enforcements, I'd rather over-audit than under-audit, since they don't evolve with commits/labels.

Issue #51, opened 2020-04-27
"is outside context directory" false positive


AlecBenzer:
[The check as-implemented](https://github.com/bufbuild/buf/blob/0795ff36a6409ff5596da80ed43643979d78d2d2/internal/pkg/storage/storagepath/storagepath.go#L72) just checks to see if the string `..` exists anywhere in the full relative filepath, which does not necessarily mean the path is outside of the context. E.g., a directory or file could be named `this..is..a..file`.

I'm running into this now, where it seems like a `node_modules` directory somewhere in our repo contains a file called `..d` and is causing buf to complain.

My case should be solvable with excludes, but in general a more robust check should maybe be used here?

bufdev:
This is a good point - need to see if there's a better way to do this. One way would be to calculate the absolute paths, and then see if the normalized path is a strict subset of the input path, but this may be expensive on a continual basis.

bufdev:
@AlecBenzer correct me if I'm wrong, but would this be fixed (and correctly fixed) if we checked against `".." + string(os.PathSeparator)` instead of `".."`? I might be crazy but I think that would work.

AlecBenzer:
@bufdev I think you might actually just be able to do `strings.StartsWith(".." + os.PathSeparator)`?

`filepath.Clean` (called from `Normalize`) should ["Eliminate each inner .. path name element (the parent directory) along with the non-.. element that precedes it."](https://pkg.go.dev/path/filepath?tab=doc#Clean), so I think the only place a `..` could even appear is at the start of the `Clean`ed path.

https://play.golang.org/p/vtUgmJk0Fta

bufdev:
https://github.com/bufbuild/buf/pull/54

bufdev:
I think this is fixed on master (and thanks for waiting on this, sorry for the delay). We'll work to do a proper release with this soon. Let me know if there's any other issues.

PR #3956, opened 2025-08-11
Fix single directory traversal for normalpath

emcfarlane:
This fixes an edge case when validating normalpaths. A path consisting of just ".." is now rejected. The func filepath.IsLocal is used to assert the path is local. The func NormalizeAndValidate is now no longer OS specific.

### DIFF
--- a/private/pkg/normalpath/normalpath.go
+++ b/private/pkg/normalpath/normalpath.go
@@ -37,8 +37,6 @@ const (
 	Absolute PathType = 2
 
 	stringOSPathSeparator = string(os.PathSeparator)
-	// This has to be with "/" instead of os.PathSeparator as we use this on normalized paths
-	normalizedRelPathJumpContextPrefix = "../"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


private/pkg/normalpath/normalpath.go
```diff
@@ -109,6 +107,28 @@ func Normalize(path string) string {
...
+func NormalizeAndValidate(path string) (string, error) {
+	normalizedPath := Normalize(path)
+	unnormalizedPath := Unnormalize(normalizedPath)
+	// Paths cannot be absolute.
+	if filepath.IsAbs(unnormalizedPath) {
+		return "", NewError(path, errNotRelative)
+	}
+	// Use IsLocal to validate the path is relative and does not jump context.
+	//
+	// The string ".." is not matched against, see https://github.com/bufbuild/buf/issues/51
```
bufdev: Isn't this in direct contradiction to https://github.com/bufbuild/buf/issues/51?

emcfarlane: Before we errored on `..` present in the path, which was the issue in #51. `IsLocal` does check for this condition (`../`) but also captures the `..` case:
- Unix: https://cs.opensource.google/go/go/+/refs/tags/go1.24.6:src/internal/filepathlite/path.go;l=161-163
- Windows: https://cs.opensource.google/go/go/+/refs/tags/go1.24.6:src/internal/filepathlite/path_windows.go;l=49-51

This could also be handled by https://pkg.go.dev/io/fs#ValidPath . I'll revert and just add the string check for `..`.

bufdev (approved):

PR #3965, opened 2025-08-18
Add audit logging event for update curated plugin status

cyinma:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -719,6 +721,15 @@ message PayloadCuratedPluginDeleted {
   string owner_name = 2;
 }
 
+message PayloadCuratedPluginStatusUpdated {
```
mfridman, line 724: Should this include the name of the plugin?

cyinma: the audit logging event will include the plugin id and name as part of the `Resource`, but not the owner id and name so we explicitly include those here in payload

PR #3983, opened 2025-08-26
Add policy support to buf config commands 

emcfarlane:
This adds policy support to `buf config` commands `ls-breaking-rules` and `ls-lint-rules`. These now return the related policy rule configuration. When listing all rules (without `--configured-only` set) only the default rules for the main configuration are returned, as all policies are expected to match it. With `--configured-only` set, each policy will define the set of rules used.

### DIFF
--- a/private/buf/cmd/buf/command/config/internal/internal.go
+++ b/private/buf/cmd/buf/command/config/internal/internal.go
@@ -257,6 +257,7 @@ func lsRun(
 		}
 		configuredRuleOptions := []bufcheck.ConfiguredRulesOption{
 			bufcheck.WithPluginConfigs(bufYAMLFile.PluginConfigs()...),
+			bufcheck.WithPolicyConfigs(bufYAMLFile.PolicyConfigs()...),
 			bufcheck.WithRelatedCheckConfigs(allCheckConfigs...),
 		}
 		rules, err = checkClient.ConfiguredRules(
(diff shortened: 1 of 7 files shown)


private/buf/cmd/buf/command/config/internal/internal.go
```diff
@@ -281,6 +283,11 @@ func lsRun(
 		if err != nil {
 			return err
 		}
+		// Filter policy rules to only those that come from plugins. Each policy rule
+		// will match the default buf.yaml configuration.
+		rules = xslices.Filter(rules, func(rule bufcheck.Rule) bool {
+			return rule.PolicyName() == "" || rule.PluginName() != ""
```
doriable: Should this be `rule.Policy.Name() != "" && rule.PluginName() != ""` for rules that come from policies and a plugin?

emcfarlane: We want to remove all policy builtin rules:
- !policy, !plugin (builtins) == true || false => true
- !policy, plugin (custom plugin) == true || true => true
- policy, !plugin (policy builtin) == false || false => false
- policy, plugin (policy's custom plugin) == false || true = true

doriable: Ah yeah, okay, that's fair, for the condition I stated, it would be the inverse of it, and this is just with the inverse factored through ^^"

emcfarlane:
@doriable removed the filtering, added a testcase for `--configured-only`.

PR #4021, opened 2025-09-19
Add rule to check `IGNORE_IF_ZERO_VALUE` on fields that track presence

srikrsna-buf:
As [documented](https://buf.build/bufbuild/protovalidate/docs/main:buf.validate#buf.validate.Ignore), `IGNORE_IF_ZERO_VALUE` is a no-op on fields that track presence. We add a check to the existing `PROTOVALIDATE` rule.

### DIFF
--- a/private/bufpkg/bufcheck/testdata/lint/protovalidate/proto/extension.proto
+++ b/private/bufpkg/bufcheck/testdata/lint/protovalidate/proto/extension.proto
@@ -9,7 +9,7 @@ message ExtensionTest {
     (buf.validate.field).string.min_len = 3,
     (buf.validate.field).required = true
   ];
-  optional string valid_ignore_empty = 2 [
+  string valid_ignore_empty = 2 [
     (buf.validate.field).string.min_len = 3,
     (buf.validate.field).ignore = IGNORE_IF_ZERO_VALUE
   ];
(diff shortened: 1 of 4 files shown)


emcfarlane (approved):

doriable (commented):
We should probably add a changelog for this behaviour also, but seems reasonable! Thank you for adding this!

private/bufpkg/bufcheck/testdata/lint/protovalidate/proto/extension.proto
```diff
@@ -9,7 +9,7 @@ message ExtensionTest {
     (buf.validate.field).string.min_len = 3,
     (buf.validate.field).required = true
   ];
-  optional string valid_ignore_empty = 2 [
+  string valid_ignore_empty = 2 [
```
doriable: Since this file is `proto2`, we'd still want to keep these here, right?
It's just adding the test case to the `proto3` file below.

private/bufpkg/bufcheck/testdata/lint/protovalidate/proto/extension.proto
```diff
@@ -9,10 +9,6 @@ message ExtensionTest {
     (buf.validate.field).string.min_len = 3,
     (buf.validate.field).required = true
   ];
-  optional string valid_ignore_empty = 2 [
```
emcfarlane, line 12: Proto2 always has presence, never valid to set `IGNORE_IF_ZERO_VALUE`.

private/bufpkg/bufcheck/testdata/lint/protovalidate/proto/extension.proto
doriable: @emcfarlane's fixes address this comment, thank you!

doriable (approved):
Looks good, just needs the changelog, approving for now!

Issue #3847, opened 2025-05-18
Buf CLI hangs indefintely


antspy:
Hi, 

I am having problem with my `buf cli`. 
I have the following `buf.yaml` file: 

```yaml
version: v2
modules:
  - path: src
lint:
  use:
    - STANDARD
```

This works as expect, and running `buf lint src/foo/bar.proto` works correctly. 
However, I want the 'base' dir to be the workspace root dir, not `src` [0]. So I modified as such: 

```yaml
version: v2
modules:
  - path: .
lint:
  use:
    - STANDARD
```

Now calling `buf lint src/foo/bar.proto` hangs forever (5min +). My guess is that `buf` is trying to scan every folder underneath the root, which contains things like node_modules, python virtual environments, bazel output dir, etc. etc. Ok, so I tried using the `includes` property: 

```yaml
version: v2
modules:
  - path: .
     includes:
       - src 
lint:
  use:
    - STANDARD
```

This has two problems: 
1. The `yaml` configuration file does not recognize the `includes` property (`excludes` is available, but not `includes`, for some reason). This might be a red herring, as it might simply be the editor not recognizing the right format for the `yaml` file. 
2. The bigger issue is that `buf lint src/foo/bar.proto` hangs once again. It looks like the `includes` property is being ignored. 

Do you have any ideas what could be happening, or how could I debug further? 

Thanks! 

[0] I am using bazel, and bazel understands imports as starting from the root dir. So if I have to `import src/foo/bar2.proto`, I need to use `src` in the path. But `buf` would not recognize that path if I set `path: src` in the `buf.yaml` config.

doriable:
I just want to make sure that I can get to a clear representation of what you're trying to do to get some sort of repro going.

> I have the following buf.yaml file

Where is the `buf.yaml` file currently? At root of your project's source repo?

> However, I want the 'base' dir to be the workspace root dir, not src [0].

Could you please elaborate on "base" in this context? The `modules` key in the `buf.yaml` is used to declare distinct "protobuf roots" in your source repository. So for example, if your source repository looks roughly something like:

```
.
├── README.md
├── buf.yaml
├── node_modules # contains vendored code
├── proto # contains your proto definitions
└── src # contains your source code
```

And you have some proto files in `proto/`, for example, `proto/abc/v1/xyz.proto`, and the import path for this proto file would be `abc/v1/xyz.proto`, then the protobuf root here is the `proto/` directory, and your `buf.yaml` would contain something that looks like:

```
[...]

PR #4025, opened 2025-09-23
Add PrefixFilterReadBucket for optimized filesystem walking

emcfarlane:
This adds PrefixFilterReadBucket and PrefixFilterReadBucketCloser that filter ReadBucket operations using include and exclude prefixes. This optimizes Walk on a bucket to traverse only the include prefixes, avoiding unrelated directories.

Fixes https://github.com/bufbuild/buf/issues/3847


doriable (commented):
So I think that the major benefit to addressing the issue at this layer is that we can get a nice tested API for the `PrefixFilterReadBucket`. However, the API has a bit of a code smell -- it's a prefix filtered, but because it conforms to the `ReadBucket` API, the walk still takes a `prefix`, which is a bit of indirection.

I know it makes it less testable, but unfortunately I think this should be addressed at the [workspace targeting layer by setting a root bucket for each includes](https://github.com/bufbuild/buf/blob/7471eb6817e604885bd7632cd8472de8940bec48/private/buf/bufworkspace/workspace_targeting.go#L683).

I think this would be a follow-up, but if we wanted to do better testing, etc. we can introduce benchmarks/reorganised our integration tests in a way where we can better understand/find performance regressions.

emcfarlane:
@doriable is the API code smell related to naming? An alternative name might be `MaskReadBucket`. The `Walk` function taking a prefix I don't think is related to the name of the entity, you wan't to be able to walk from some directory. It is not conflicting the `ReadBucket` API.

For setting it at the workspace level I think this would break the optimization. For example, a targeted module that has untargeted dependencies all from the root workspace would need to ensure that the walk is optimized to include all their include/exclude values. I'm not sure if that logic can be pushed into the target paths (or if we want it to as maybe this conflates what a "target" is). 

Tests were manually run on large workspaces. Tests that replicate that are part of the bucket that assert the walk calls on the delegate. A benchmark would require a large initial state to test effectively.

emcfarlane:
Ah maybe misunderstood your `root buckets`, that should be a quick way to test. I'll close this  for now.

doriable:
All good, we can always re-open this if we do end up going down this route, but yeah, I probably should've linked a more clear set of lines in the source, lol -- I was thinking of refactoring the logic in the lines above the one I linked to set root buckets based on `includes` (since that is effectively what is going on -- you want to constraint the roots to your `includes`).

> An alternative name might be MaskReadBucket. The Walk function taking a prefix I don't think is related to the name of the entity, you wan't to be able to walk from some directory. It is not conflicting the ReadBucket API.

Maybe that is a better name, tbh -- yeah, it's not strictly conflicting, I think it's more that the read bucket API needs to have clarity on where it cares about the "root of the bucket" vs. "prefixes" (and sorry, it doesn't help that the code that does use these layers currently doesn't present that clarity very well either, that one is on me :<).

I think if we do end up re-opening this PR, I think this renaming would be helpful :D

emcfarlane:
Reopening the PR. We don't currently have an API in `storage` to combine rooted buckets at an include and map them back to their module root. It would be the opposite of the [storage.MapOnPrefix](https://pkg.go.dev/github.com/bufbuild/buf@v1.57.2/private/pkg/storage#MapOnPrefix), which would then be combine with multiple buckets.

doriable (approved):
Looks good to me, tracks with what we chatted about offline, thank you!

PR #4026, opened 2025-09-23
Use new `protocompile/experimental` for LSP and implement `textDocument/References`

doriable:
This PR ports the existing LSP logic to use our new 
`protocompile/experimental` compiler.
This uses the new compiler's query model to get the IR, which
is used for symbol indexing.

In addition to porting over to the new compiler, this also implements
`textDocument/References`, which finds all references for a definition,
when applicable.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -329,7 +329,7 @@ func commentToMarkdown(comment string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(comment, "/*"), "*/")
 }
 
-// compareRanges compares two ranges for lexicographic ordering.
+// comparePositions compares two ranges for lexicographic ordering.
 func comparePositions(a, b protocol.Position) int {
 	diff := int(a.Line) - int(b.Line)
 	if diff == 0 {
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -529,7 +529,7 @@ func (f *file) IndexSymbols(ctx context.Context) {
 		f.referenceableSymbols[def.ast.Name().Canonicalized()] = sym
 	}
 
-	// TODO: this could us a refactor, probably.
+	// TODO: this could use a refactor, probably.
 
 	// Resolve all unresolved symbols from this file
 	for _, sym := range unresolved {


private/buf/buflsp/symbol.go
```diff
@@ -380,208 +170,132 @@ func (s *symbol) LogValue() slog.Value {
...
-		fmt.Fprintf(&tooltip, "```proto\nbuiltin %s\n```\n", kind.name)
-		for _, line := range builtinDocs[kind.name] {
-			fmt.Fprintln(&tooltip, line)
-		}
-
-		fmt.Fprintln(&tooltip)
-		fmt.Fprintf(
-			&tooltip,
-			"This symbol is a Protobuf builtin. [Learn more on protobuf.com.](https://protobuf.com/docs/language-spec#field-types)",
-		)
```
stefanvanburen, line 401: @doriable @emcfarlane just wondering, why did we drop these protobuf.com links? they seem pretty valuable for people looking for more details on stuff like these builtins.

private/buf/buflsp/symbol.go
```diff
@@ -380,208 +170,132 @@ func (s *symbol) LogValue() slog.Value {
...
-			bsrTooltip = pkg + "." + strings.Join(path, ".")
-		} else {
-			bsrTooltip = pkg + "." + strings.Join(path[:len(path)-1], ".")
-		}
-		bsrAnchor = bsrTooltip
-		if isExtension {
-			bsrAnchor = "extensions"
-		}
-
-		fmt.Fprintf(&tooltip, "[`%s` on the Buf Schema Registry](https://%s/docs/main:%s#%s)\n\n", bsrTooltip, bsrHost, pkg, bsrAnchor)
```
stefanvanburen, line 559: similarly, I would think that we'd want to link to the BSR if possible?

private/buf/buflsp/symbol.go
doriable: Honestly, I think that might've just been an oversight on my part -- I will add these back in.

private/buf/buflsp/symbol.go
doriable: Agreed on this -- going to add this back in also.

doriable: https://github.com/bufbuild/buf/pull/4137

PR #4036, opened 2025-10-05
Only set HTTP1 when HTTP/2 Prior Knowledge flag is false in `buf curl`

edaniels:
As of v1.57.2, calling `~/go/bin/buf curl -vvvvvv --debug --http2-prior-knowledge --protocol grpc http://127.0.0.1:55555 --list-methods` would result in the TLS handshake hanging. After bisecting down to v1.57.1 being the first bad tag, I found the regression here https://github.com/bufbuild/buf/commit/2d0ab7a0f731b8d06f1e764e775ec70b32c41b24#diff-05c240e5d4191f595d5b747c112c314cf14ee4d77505cee99aac9500eed39527L1135-R1144. I think there are lacking test cases for the buf curl command (unless you have some other acceptance test suite internally), so this was an easy miss. With this, we should be back to insecurely calling gRPC server methods 8).

### DIFF
--- a/private/buf/cmd/buf/command/curl/curl.go
+++ b/private/buf/cmd/buf/command/curl/curl.go
@@ -1131,7 +1131,7 @@ func makeHTTPRoundTripper(f *flags, isSecure bool, authority string, printer ver
 		}
 	}
 	protocols := new(http.Protocols)
-	protocols.SetHTTP1(!f.HTTP2PriorKnowledge && !f.HTTP3)
+	protocols.SetHTTP1(!f.HTTP2PriorKnowledge)
 	protocols.SetHTTP2(true)
 	protocols.SetUnencryptedHTTP2(f.HTTP2PriorKnowledge && !isSecure)
 	return &http.Transport{


edaniels:
@emcfarlane tagging you on this since you'll probably have the most context. Thanks!

s4lmon:
LGTM @edaniels

private/buf/cmd/buf/command/curl/curl.go
```diff
@@ -1131,7 +1131,7 @@ func makeHTTPRoundTripper(f *flags, isSecure bool, authority string, printer ver
 		}
 	}
 	protocols := new(http.Protocols)
-	protocols.SetHTTP1(true)
+	protocols.SetHTTP1(!f.HTTP2PriorKnowledge && !f.HTTP3)
```
doriable:
```suggestion
	protocols.SetHTTP1(!f.HTTP2PriorKnowledge)
```

`f.HTTP3` is a redundant check, since this is checked at the top of `makeHTTPRoundTripper` and will always be `false` at this level.

As the documentation for the `--http2-prior-knowledge` flag states:

> This flag can be used to indicate that HTTP/2 should be used. **Without** this, HTTP 1.1
> will be used with URLs with an http scheme, and protocol negotiation will be used to
> choose either HTTP 1.1 or HTTP/2 for URLs with an https scheme. **With this flag set,**
> **HTTP/2 is always used, even over plain-text.**

So I think if we adjust this to only set `HTTP1` in the case where prior knowledge is false (thus opting into protocol negotiation), this would do the trick.

edaniels: Yep that will be fine. I did this as a paranoia check in case that above check ever changes in lieu of tests. I will make the change though!

doriable:
This looks reasonable to me, before approving, I'm going to let @emcfarlane a chance to go over this as well.

PR #4064, opened 2025-10-15
Add visibility control audit events

unmultimedio:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -1102,3 +1106,38 @@ message PayloadPolicyEnforcementTargetsChanged {
...
+
+message PayloadServerResourceVisibilityControlChanged {
+  // The resource type that had its visibility control changed.
+  ResourceType resource_type = 1;
+  // The old visibility control for resources of that type under the server instance. If not
+  // present, then this visibility control was just created.
+  optional ResourceVisibilityControl old_visibility_control = 2;
+  // The new visibility control for resources of that type under the server instance. If not
+  // present, then this visibility control was just deleted.
+  optional ResourceVisibilityControl new_visibility_control = 3;
```
cyinma, line 1131: I think this is no longer optional

unmultimedio: Wanted to leave it as it is in case we consider deleting controls in the future.

PR #4066, opened 2025-10-16
LSP fix positional encoding for diagnostics

emcfarlane:
This PR fixes the offset calculation for the utf-16 encoding of the LSP protocol for diagnostics generated by bufanalysis and the current proto compiler.


private/buf/buflsp/diagnostic.go
```diff
@@ -86,3 +87,25 @@ func reportDiagnosticToProtocolDiagnostic(
...
+// This is needed when converting positions from the old protocompile API or bufanalysis
+// which use byte offsets, to LSP positions which use UTF-16 code units.
+func byteOffsetToUTF16Column(text string, byteOffset int) int {
+	// Find the start of the line containing the byte offset.
+	lineStart := strings.LastIndex(text[:byteOffset], "\n")
+	if lineStart == -1 {
+		lineStart = 0
+	} else {
+		lineStart++ // Skip the newline character.
+	}
```
mcy: This can just be lineStart++

private/buf/buflsp/file.go
```diff
@@ -1131,6 +1135,25 @@ func (f *file) appendLintErrors(source string, err error) bool {
 	return true
 }
 
+// lineColumnToByteOffset converts a 1-indexed line and 1-indexed byte-based column to a byte offset in the file.
+func (f *file) lineColumnToByteOffset(line, col int) int {
+	currentLine := 1
+	byteOffset := 0
+
+	// Find the start of the target line.
+	for i := 0; i < len(f.text) && currentLine < line; i++ {
```
mcy: This could be accidentally quadratic :(

Do you have a `report.File` lying around that you can use to avoid this?

private/buf/buflsp/diagnostic.go
emcfarlane: Removed, kept a workaround for reading the file annotations. Maybe we should just skip this case and leave as-is for now, wdyt?

private/buf/buflsp/file.go
emcfarlane: Fixed, use `InverseLocation` now.

private/buf/buflsp/file.go
```diff
@@ -1149,3 +1150,11 @@ func (f *file) PublishDiagnostics(ctx context.Context) {
 type wktObjectInfo struct {
 	storage.ObjectInfo
 }
+
+func readAllAsString(reader io.Reader) (string, error) {
```
mcy, line 1154: Seriously isn't already in `io`? Sigh.

mcy (approved):

PR #4067, opened 2025-10-16
LSP add autocompletion for basic keywords, syntax, package and imports

emcfarlane:
This PR adds basic autocompletion features for keywords types (limited to top-level), syntax (proto or edition), package name, and imports.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -18,6 +18,7 @@ package buflsp
 
 import (
 	"context"
+	"fmt"
 	"iter"
 	"log/slog"
 	"slices"
@@ -122,7 +123,7 @@ func completionItemsForSyntax(ctx context.Context, file *file, syntaxDecl ast.De
 	var items []protocol.CompletionItem
 	for syntax := range syntaxes {
 		items = append(items, protocol.CompletionItem{
-			Label: prefix + "\"" + syntax.String() + "\";",
+			Label: fmt.Sprintf("%s%q;", prefix, syntax),
 			Kind:  protocol.CompletionItemKindValue,
 		})
 	}


private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		file.lsp.logger.DebugContext(
+			ctx,
+			"no AST found for completion",
+			slog.String("file", file.uri.Filename()),
+		)
+		return nil
+	}
+
+	// Convert LSP position (UTF-16 code units) to byte offset.
+	offset := protocolPositionToOffset(file.text, position)
```
mcy: Replace `file.text` with a `*report.File`, then use the native position calculation from that.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+
+	// Walk backwards from the cursor position.
+	start := offset
+	for start > 0 {
+		// Get the byte before current position.
+		b := text[start-1]
+
+		// Check if it's a valid identifier character.
+		// Valid: letters, digits, underscore, dot
+		if isIdentifierChar(b) {
```
mcy: Can you query the token stream instead? I don't like the fact that you're parsing an identifier here, because:

1. This is incorrect because it will treat `123` as an identifier.
2. It does not handle identifiers with non-ASCII characters, which protocompile supports but diagnoses.
3. If we ever change the identifier grammar (such as to add a raw identifier syntax like `#keyword`), we need to update this code too.

Another problem w this is that `buf. validate` is a valid path, but this will not handle that correctly. This happens if you're e.g. having to break a path into multiple lines. Also, you don't handle paths like `(buf.validate.field).rule`, which is something we really should support. Parsing this correctly without help from the lexer will be quite painful...

Finally, consider `repeated .foo.Bar`. This will incorrectly include `repeated` in the path name. I'm not sure that there's an especially good way to cope with this, though..

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+			if !method.Name.IsZero() {
+				nameSpan := method.Name.Span()
+				if offset >= nameSpan.Start && offset <= nameSpan.End {
+					return true
+				}
+			}
+		}
+
+	case ast.DeclKindSyntax:
+		// Cursor in syntax declaration - likely in the syntax value, not a good place for completions.
```
mcy: Are you sure? Makes sense to complete `syntax = |` with possible editions, listed from the ast/syntax packge.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		// No declaration found, return top-level keywords.
+		return topLevelCompletionItems()
+	}
+
+	// Return context-specific completions based on declaration type.
+	switch decl.Kind() {
+	case ast.DeclKindDef:
+		return completionItemsForDef(file, decl.AsDef(), prefix)
+	case ast.DeclKindSyntax:
+		// Inside syntax declaration - could suggest syntax values.
```
mcy: I would make this a TODO.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+				}
+			}
+		}
+
+	case ast.DeclKindSyntax:
+		// Cursor in syntax declaration - likely in the syntax value, not a good place for completions.
+		return true
+
+	case ast.DeclKindPackage:
+		// Cursor in package declaration - package name is unique.
```
mcy: Not true: multiple files can have the same package. We should suggest a name based on the file's path or something like that. You can have a TODO.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		keyword.RPC,
+		keyword.Option,
+	}
+	return xslices.Map(serviceKeywords, keywordToCompletionItem)
+}
+
+// completionItemsForImport returns completion items for import declarations.
+func completionItemsForImport(file *file) []protocol.CompletionItem {
+	// Suggest importable file paths.
+	paths := xslices.MapKeysToSlice(file.importToFile)
```
mcy: This allocates an extra slice. I would just iterate over the map instead of using this.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+	})
+}
+
+// referenceableTypeCompletionItems returns completion items for user-defined types (messages, enums).
+// The prefix parameter filters types and determines what to insert (suffix after prefix).
+func referenceableTypeCompletionItems(file *file, prefix string) []protocol.CompletionItem {
+	var items []protocol.CompletionItem
+
+	// Add types from the current file.
+	for _, symbol := range file.referenceableSymbols {
```
mcy: At some point it might be good to eliminate buflsp.symbol entirely...

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		if !symbol.ir.Kind().IsType() {
+			continue
+		}
+
+		// For same-file types, use short name.
+		fullName := string(symbol.ir.FullName())
+		shortName := symbol.ir.FullName().Name()
+
+		// Filter by prefix if provided.
+		if prefix != "" {
```
mcy: Redundant check, HasPrefix(s, "") is always true.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+				continue
+			}
+		}
+
+		// Determine label and insert text.
+		label := shortName
+		insertText := shortName
+		if prefix != "" {
+			// If user typed a prefix, show full context but insert only suffix.
+			if strings.HasPrefix(fullName, prefix) {
```
mcy: CutPrefix here and below.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+			insertText := shortName
+			if imported.ir.Package() != file.ir.Package() {
+				// Different package - use fully qualified name.
+				label = fullName
+				insertText = fullName
+			}
+
+			// If user typed a prefix, adjust insert text to be just the suffix.
+			if prefix != "" {
+				if strings.HasPrefix(fullName, prefix) {
```
mcy: CutPrefix

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		for _, symbol := range imported.referenceableSymbols {
+			if !symbol.ir.Kind().IsType() {
+				continue
+			}
+
+			fullName := string(symbol.ir.FullName())
+			shortName := symbol.ir.FullName().Name()
+
+			// Filter by prefix if provided.
+			if prefix != "" {
```
mcy: Redundant.

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+		items = append(items, protocol.CompletionItem{
+			Label:      label,
+			InsertText: insertText,
+			Kind:       protocol.CompletionItemKindTypeParameter,
+		})
+	}
+
+	// Add types from imported files.
+	for _, imported := range file.importToFile {
+		for _, symbol := range imported.referenceableSymbols {
```
mcy: This looks very similar to the above loop body...

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,533 @@
...
+func getDeclForPosition(body ast.DeclBody, offset int) ast.DeclAny {
+	return findSmallestDecl(body, offset)
+}
+
+// findSmallestDecl recursively searches for the smallest declaration containing the offset.
+func findSmallestDecl(body ast.DeclBody, offset int) ast.DeclAny {
+	var smallest ast.DeclAny
+	smallestSize := -1
+
+	for _, decl := range seq.All(body.Decls()) {
```
mcy: seq.Values

private/buf/buflsp/server.go
```diff
@@ -422,6 +426,30 @@ func (s *server) References(
...
+) (*protocol.CompletionList, error) {
+	file := s.fileManager.Get(params.TextDocument.URI)
+	if file == nil {
+		return nil, nil
+	}
+
+	items := getCompletionItems(ctx, file, params.Position)
+	return &protocol.CompletionList{
+		Items: items,
+	}, nil
```
mcy: make this one line?

private/buf/buflsp/completion.go
doriable: In the [code completion branch I have up](https://github.com/bufbuild/buf/compare/main...BSR-5472-code-completion), there's a code path for suggesting package names -- this should be based on the file path, especially since that's our [lint recommendation](https://buf.build/docs/lint/rules/#package_directory_match).

private/buf/buflsp/completion.go
doriable: There should be a code path for this in the [code completion branch](https://github.com/bufbuild/buf/compare/main...BSR-5472-code-completion) also.

private/buf/buflsp/completion.go
emcfarlane: Fixed.

private/buf/buflsp/completion.go
emcfarlane: Found a fixed but discarded, removed from the PR for now as is may not be needed/was a side affect of the returned completion item type.

private/buf/buflsp/completion.go
emcfarlane: Done in https://github.com/bufbuild/buf/pull/4066

private/buf/buflsp/completion.go
```diff
@@ -0,0 +1,307 @@
...
+	} else {
+		syntaxes = func(yield func(syntax.Syntax) bool) {
+			_ = yield(syntax.Proto2) &&
+				yield(syntax.Proto3)
+		}
+	}
+	var items []protocol.CompletionItem
+	for syntax := range syntaxes {
+		items = append(items, protocol.CompletionItem{
+			Label: prefix + "\"" + syntax.String() + "\";",
```
mcy: `fmt.Sprintf("%q;", syntax)`

PR #4070, opened 2025-10-16
LSP add workspace symbol queries

emcfarlane:
This adds basic support for workspace symbol queries. Symbols can be serached by name, with the parent listed as the container name. For example searching given a message type `buf.acme.User` with a field `name`. Searching for `user` returns `User` in the container `buf.acme`. Searching for `name` returns `name` in the container `buf.acme.User`.

### DIFF
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -517,10 +517,10 @@ func (s *server) Symbols(
 	query := strings.ToLower(params.Query)
 
 	var results []protocol.SymbolInformation
-	s.fileManager.uriToFile.Range(func(uri protocol.URI, file *file) bool {
+	for uri, file := range s.fileManager.uriToFile.Range {
 		if file.ir.IsZero() {
 			s.lsp.logger.DebugContext(ctx, fmt.Sprintf("workspace symbol: skipping file without IR: %s", uri))
-			return true
+			continue
 		}
 
 		// Search through all symbols in this file.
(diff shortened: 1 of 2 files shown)


private/buf/buflsp/server.go
```diff
@@ -505,3 +507,49 @@ func (s *server) SemanticTokensFull(
...
+// Symbols is the entry point for workspace-wide symbol search.
+func (s *server) Symbols(
+	ctx context.Context,
+	params *protocol.WorkspaceSymbolParams,
+) ([]protocol.SymbolInformation, error) {
+	const maxResults = 1000 // Limit results to avoid overwhelming clients
+	query := strings.ToLower(params.Query)
+
+	var results []protocol.SymbolInformation
+	s.fileManager.uriToFile.Range(func(uri protocol.URI, file *file) bool {
```
mcy: You can do `for uri, file := range s.fileManager.uriToFile.Range`

private/buf/buflsp/symbol.go
```diff
@@ -257,6 +257,55 @@ func (s *symbol) FormatDocs(ctx context.Context) string {
...
+	case ir.SymbolKindEnum:
+		kind = protocol.SymbolKindEnum
+	case ir.SymbolKindEnumValue:
+		kind = protocol.SymbolKindEnumMember
+	case ir.SymbolKindField:
+		kind = protocol.SymbolKindField
+	case ir.SymbolKindExtension:
+		kind = protocol.SymbolKindField
+	case ir.SymbolKindOneof:
+		kind = protocol.SymbolKindField
```
mcy: mmmm tbh I would use class here? Up to you I think

mcy (approved):

PR #4075, opened 2025-10-17
Remove extra space when writing first hunk when outside of context

paul-sachs:
When the first lines are outside of the change header, there was an extra space being injected in the first line. Fixed the conditions slightly to ensure that we don't get an extra space.

### DIFF
--- a/private/pkg/diff/diffmyers/diffmyers.go
+++ b/private/pkg/diff/diffmyers/diffmyers.go
@@ -144,8 +144,10 @@ func Print(from, to [][]byte, edits []Edit) ([]byte, error) {
 	var buffer bytes.Buffer
 	buffer.Grow(bufferSize)
 	for _, line := range out {
-		if line.hunk && len(line.line) > 0 {
-			buffer.Write(line.line)
+		if line.hunk {
+			if len(line.line) > 0 {
+				buffer.Write(line.line)
+			}
 			continue
 		}
 		switch line.EditKind {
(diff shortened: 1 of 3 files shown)


srikrsna-buf (approved):

private/pkg/diff/diffmyers/diffmyers.go
```diff
@@ -144,8 +144,10 @@ func Print(from, to [][]byte, edits []Edit) ([]byte, error) {
 	var buffer bytes.Buffer
 	buffer.Grow(bufferSize)
 	for _, line := range out {
-		if line.hunk && len(line.line) > 0 {
-			buffer.Write(line.line)
+		if line.hunk {
+			if len(line.line) > 0 {
+				buffer.Write(line.line)
+			}
```
unmultimedio, line 150: Isn't this exactly the same? `if a { if b { do }}` vs `if a && b { do }`

paul-sachs: Nope, since there's a `continue` after the check `if len(line.line) > 0 {` block. Otherwise yes it would be

paul-sachs: Which means we continue if `a` is true, whereas before, we only continued if `a && b` were true

unmultimedio: yeah I missed that `continue`, thanks!

PR #4089, opened 2025-10-20
Deprecate and hide the timeout flag

bufdev:
This PR deprecates and hides the global `--timeout` flag.

This specifically came up with `buf lsp serve`, where this command should not have a timeout. As such, we'd want to move `--timeout` to non-global anyways, but looking at the problem, it's not entirely clear what we're solving by having this flag in the first place. It was originally added because "all RPCs should have a timeout" but I'm not sure if this really desired or necessary.

As such, this just turns it into a no-op. You could theoretically argue that users expecting timeouts today will no longer get them, but this is generally how flags are deprecated, and you get a warning.


bufdev:
Alternative: https://github.com/bufbuild/app-go/pull/3

PR #4092, opened 2025-10-21
Change HoverProvider on import to show path on disk

stefanvanburen:
This is more in-line with how other LSP servers behave.


stefanvanburen:
Hovering a file in the module:

image

Hovering an imported file:

image

Hovering a WKT import:

image

emcfarlane (approved):

PR #4094, opened 2025-10-21
Fix several issues with import completion

stefanvanburen:
* Removes an extra space we were inserting during completion
* Supports filtering on the existing import "prefix" in the text
* Avoids adding a `;` when inserting path within quotes (but will send an AdditionalTextEdit to add the `;` at the end of line, if it doesn't exist)
* Removes unnecessary sorting (this is done by LSP clients)
* Avoids an error in clients on attempting completion with a nil return

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -247,11 +247,12 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
 				// We're in a situation where we have `import goo`, in which case we want to make sure that
 				// we fix the entire import to be wrapped in quotes.
 				// Send an AdditionalTextEdit to add the front quote.
-				if len(currentImportPathText) > math.MaxUint32 {
+				importPathLength := len(currentImportPathText)
+				if importPathLength < 0 || importPathLength > math.MaxUint32 {
 					file.lsp.logger.DebugContext(ctx, "completion: current import path text > math.MaxUint32")
 					return nil
 				}
-				characterForInsert := position.Character - uint32(len(currentImportPathText))
+				characterForInsert := position.Character - uint32(importPathLength)
 				positionForInsert := protocol.Position{
 					Line:      position.Line,
 					Character: characterForInsert,
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -184,11 +184,6 @@ linters:
         path: private/buf/buflsp/symbol.go
         text: 'G115:'
       - linters:
-          - gosec
-        # G115 checks for use of truncating conversions.
-        path: private/buf/buflsp/completion.go
-        text: 'G115'
-      - linters:
           - containedctx
         # Type must implement an interface whose methods do not accept context. But this
         # implementation makes RPC calls, which need a context. So we allow creator of the


private/buf/buflsp/completion.go
```diff
@@ -208,19 +208,52 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
+			_, _ = newText.WriteString(`"`)
+			// Only suggest a finishing `;` character if one doesn't already exist, and we're writing out
+			// a `"` before it.
+			// TODO: Should we add an AdditionalTextEdit for adding the `;` at the end of the line?
+			// e.g., if we're doing:
+			//   import "‸"
+			// And we select a value, is it valuable to also insert the semicolon?
+			if declImport.Semicolon().IsZero() {
+				_, _ = newText.WriteString(";")
+			}
```
stefanvanburen: I'm not really sure if:

a. this is desirable
b. this is the intention of `AdditionalTextEdits`

Happy to try it out if it sounds valuable, though.

private/buf/buflsp/server.go
```diff
@@ -429,6 +429,9 @@ func (s *server) Completion(
 		return nil, nil
 	}
 	items := getCompletionItems(ctx, file, params.Position)
+	if items == nil {
+		return nil, nil
+	}
```
stefanvanburen, line 434: without this, running a completion on the first character of an empty line would result in an error in nvim:

Image

private/buf/buflsp/completion.go
```diff
@@ -208,19 +208,52 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
+				Range: protocol.Range{
+					Start: position,
+					End:   position,
+				},
+			},
 		})
 	}
-	slices.SortFunc(items, func(a, b protocol.CompletionItem) int {
-		return strings.Compare(strings.ToLower(a.Label), strings.ToLower(b.Label))
-	})
```
stefanvanburen, line 223: Sorting is left up to the client IIUC: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#:~:text=to%20achieve%20consistency%20across%20languages%20and%20to%20honor%20different%20clients%20usually%20the%20client%20is%20responsible%20for%20filtering%20and%20sorting.

emcfarlane: Yeah, my though was it's just nicer to have deterministic output anyway.

emcfarlane:
Can you check the output vscode? In nvim I get some partial autocompletes:

Screenshot 2025-10-21 at 9 47 18 PM
Screenshot 2025-10-21 at 9 47 13 PM

And in vscode its not doing an inset to replace a partial, and also doesn't insert a space:
Screenshot 2025-10-21 at 9 52 45 PM
Screenshot 2025-10-21 at 9 48 30 PM

stefanvanburen:
@emcfarlane I hadn't thought to test:

```protobuf
import "google/‸/empty.proto";
```

but truthfully I would think that would come up rarely in practice. I sorta wonder if we just want to balk in that scenario? Feels weird to search through the prefix paths & also find a matching suffix, but maybe I'm being dense.

And I imagine the VS Code space thing is why we were adding it in the first place here :(. Clients!

emcfarlane:
My testcase was `"google/_";` and selecting a autocomplete corrupted the output to `"google//empty.proto";`

stefanvanburen:
> My testcase was `"google/_";` and selecting a autocomplete corrupted the output to `"google//empty.proto";`

ah, thanks for the context — I'll make sure that's fixed.

stefanvanburen:
> also doesn't insert a space:

Fixed in https://github.com/bufbuild/buf/pull/4094/commits/05f8f87a2a3a795c87b4d8e67bf383d747e462d3 (maybe not the cleanest solution; open to suggestions).

stefanvanburen:
> My testcase was `"google/_";` and selecting a autocomplete corrupted the output to `"google//empty.proto";`

Haven't been able to reproduce this in VS Code / nvim (with [mini.completion](https://github.com/nvim-mini/mini.nvim/blob/main/readmes/mini-completion.md)). Can you try that again on your end?

private/buf/buflsp/completion.go
stefanvanburen: ok, added this in https://github.com/bufbuild/buf/pull/4094/commits/1335d8f88eec781c5990dc52e9bc36fa2771b086, which worked well in both nvim and vscode.

private/buf/buflsp/completion.go
stefanvanburen: happy to put it back if you prefer — just figured we'd be spending extra CPU cycles and trying to keep the LSP as lean as possible.

private/buf/buflsp/server.go
emcfarlane: nice catch!

emcfarlane:
Space issue looks fixed. Noticed this issue that autocomplete only shows without quotes now, and this shows the previous corruption.

Screenshot 2025-10-22 at 6 39 38 PM
Screenshot 2025-10-22 at 6 39 24 PM
Screenshot 2025-10-22 at 6 42 08 PM
Screenshot 2025-10-22 at 6 42 04 PM

private/buf/buflsp/server.go
emcfarlane: Better check would be `if len(items) == 0 {`.

emcfarlane: Maybe `getCompletions` should return an empty non nil value instead? Does that fix it too?

stefanvanburen:
> Noticed this issue that autocomplete only shows without quotes now,

Thought the quotes were mostly noise since they wrapped all the results, but we can put them back if you'd like.

stefanvanburen:
I think the issue you're seeing is from not having the import path wrapped in a `""`, so we don't have an [ImportPath](https://github.com/bufbuild/buf/blob/1335d8f88eec781c5990dc52e9bc36fa2771b086/private/buf/buflsp/completion.go#L214) in the AST. I guess we could look for the last token on the line instead?

private/buf/buflsp/completion.go
```diff
@@ -208,19 +209,116 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
 // completionItemsForImport returns completion items for import declarations.
 //
 // Suggest all importable files.
-func completionItemsForImport(ctx context.Context, file *file) []protocol.CompletionItem {
+func completionItemsForImport(ctx context.Context, file *file, declImport ast.DeclImport, position protocol.Position) []protocol.CompletionItem {
```
emcfarlane, line 212: This logic is complicated and still doesn't feel right:
- Completions don't work on valid imports within quotes: `import "goog_"; 
- Doesn't sort/eliminate completion that are invalid for the prefix `import google_` shows to many

Image
Image

stefanvanburen: I'm not sure I'm seeing the same thing as you; trying it in VSCode:

1. `import "goog_";`, then <kbd>ctrl</kbd>+<kbd>space</kbd>:
image

selecting an option results in the right import statement:

image

---

2. `import google_`, then <kbd>ctrl</kbd>+<kbd>space</kbd>:

image

selecting an option:

image

stefanvanburen:
Closing in favor of #4103.

PR #4096, opened 2025-10-22
Add LSP DocumentSymbolProvider support

stefanvanburen:
Seemed straightforward to add.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add `textDocument/documentSymbol` support for `buf lsp serve`.
 
 ## [v1.59.0] - 2025-10-20
 


private/buf/buflsp/server.go
```diff
@@ -529,45 +531,64 @@ func (s *server) Symbols(
...
+	}
+	// Fix the return type.
+	anyResults := make([]any, len(results))
+	for i, result := range results {
+		anyResults[i] = result
+	}
+	return anyResults, nil
+}
+
+func getFileSymbols(f *file, query string) []protocol.SymbolInformation {
```
stefanvanburen: I guess this could be an iterator, which would let us stop at maxResults instead of processing the rest of the file?

emcfarlane: +1 to iterator, and maybe it should live as a method on file?

private/buf/buflsp/server.go
```diff
@@ -529,45 +531,64 @@ func (s *server) Symbols(
...
+			// Trim the slice if we have too many symbols.
+			results = results[:maxSymbolResults]
+			break
+		}
+	}
+	return results, nil
+}
+
+// DocumentSymbol is the entry point for document symbol search.
+func (s *server) DocumentSymbol(ctx context.Context, params *protocol.DocumentSymbolParams) (result []any, err error) {
```
emcfarlane: Can you comment the return type. This does the `[]SymbolInformation` not `[]DocumentInformation`.

stefanvanburen: added a comment in https://github.com/bufbuild/buf/pull/4096/commits/a3dca4aec4ff8b1923141de7897f30176b023857

private/buf/buflsp/server.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4096/commits/a3dca4aec4ff8b1923141de7897f30176b023857

PR #4097, opened 2025-10-22
LSP add workspace manager

emcfarlane:
This PR adds a workspace manager to handle multiple files sharing the same workspace. This is used to resolve cache files to their parent workspace. A file in the cache is now able to to resolve its imports with the parents dependencies. Fixes the issue describe in #4091 with support for the LSP within the cached dep.

Closes https://github.com/bufbuild/buf/pull/4091

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fix LSP published diagnostics to filter to the opened file.
 - Add `textDocument/documentSymbol` support for `buf lsp serve`.
+- Fix LSP navigation for cached modules which could cause import paths to become unresolvable.
 
 ## [v1.59.0] - 2025-10-20
 


private/buf/buflsp/workspace.go
```diff
@@ -0,0 +1,199 @@
...
+type errUnresolvableWorkspace protocol.URI
+
+func (e errUnresolvableWorkspace) Error() string {
+	return fmt.Sprintf("unresolvable workspace for %q", string(e))
+}
+
+// workspaceManager tracks all workspaces the LSP is currently handling, per file.
+type workspaceManager struct {
+	lsp        *lsp
+	workspaces []*workspace
```
stefanvanburen, line 40: do we need a mutex around this field?

private/buf/buflsp/workspace.go
```diff
@@ -0,0 +1,199 @@
...
+type workspace struct {
+	lsp *lsp
+
+	// refCount counts all the files that currently reference this workspace.
+	// A refCount of zero will be removed by the workspaceManager on cleanup.
+	refCount           int
+	workspaceURI       protocol.URI // File that created this workspace.
+	workspace          bufworkspace.Workspace
+	fileNameToFileInfo map[string]bufmodule.FileInfo
+	checkClient        bufcheck.Client
```
stefanvanburen, line 125: similar for these - do we need a mutex to handle multiple callers? or is this all synchronized at a higher level?

private/buf/buflsp/workspace.go
emcfarlane: All the methods should be sync'd through this: https://github.com/bufbuild/buf/blob/fc3f71d519d605a2ec8f5e0dfb4a5ed005f0cdc4/private/buf/buflsp/buflsp.go#L146-L148
But we do have checks run async. Let me check that doesn't break this model.

emcfarlane: https://github.com/bufbuild/buf/blob/fc3f71d519d605a2ec8f5e0dfb4a5ed005f0cdc4/private/buf/buflsp/file.go#L886-L887

Ensures checks are serialized too.

PR #4103, opened 2025-10-23
LSP import completion expression fixes

emcfarlane:
This makes the import completion expression more robust. Completions are provided on entering the import expression. 

Partial fix for #4094

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -215,20 +215,28 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
 	offset := positionLocation.Offset
 
 	importPathSpan := declImport.ImportPath().Span()
+	start, end := importPathSpan.Start, importPathSpan.End
 	importPathText := importPathSpan.Text()
-	if importPathSpan.Start > offset || offset > importPathSpan.End {
+
+	// Break on newlines to split an unintended capture.
+	if index := strings.IndexByte(importPathText, '\n'); index != -1 {
+		end = start + index
+		importPathText = importPathText[:index]
+	}
+
+	if start > offset || offset > end {
 		file.lsp.logger.Debug("completion: outside import expr range",
 			slog.String("import", importPathText),
-			slog.Int("start", importPathSpan.Start),
+			slog.Int("start", start),
 			slog.Int("offset", offset),
-			slog.Int("end", importPathSpan.End),
+			slog.Int("end", end),
 			slog.Int("line", int(position.Line)),
 			slog.Int("character", int(position.Character)),
 		)
 		return nil
 	}
 
-	index := offset - importPathSpan.Start
+	index := offset - start
 	prefix := importPathText[:index]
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -208,19 +208,61 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
-func completionItemsForImport(ctx context.Context, file *file) []protocol.CompletionItem {
+func completionItemsForImport(ctx context.Context, file *file, declImport ast.DeclImport, position protocol.Position) []protocol.CompletionItem {
 	file.lsp.logger.DebugContext(ctx, "completion: import declaration", slog.Int("importable_count", len(file.importToFile)))
 
+	positionLocation := file.file.InverseLocation(int(position.Line)+1, int(position.Character)+1, positionalEncoding)
+	offset := positionLocation.Offset
+
+	importPathSpan := declImport.ImportPath().Span()
+	importPathText := importPathSpan.Text()
+	if importPathSpan.Start > offset || offset > importPathSpan.End {
```
emcfarlane: @stefanvanburen I think we can follow this pattern hopefully in other completions:
- Get the AST node for the current position
- Figure out the context of the position in the current expression
- Suggest completions that match the current expression

private/buf/buflsp/server.go
```diff
@@ -123,14 +123,11 @@ func (s *server) Initialize(
 			},
 			CompletionProvider: &protocol.CompletionOptions{
 				ResolveProvider:   true,
-				TriggerCharacters: []string{" ", ".", "("},
+				TriggerCharacters: []string{" ", ".", "(", "\"", "/"},
```
stefanvanburen, line 126: nice, good idea.

stefanvanburen (approved):
happy to land this as it's a big improvement already; two things from #4094 that we don't have in this one:

* the additionalTextEdit for the final `;` if missing (it'd be nice to get this in here or subsequently, since otherwise a user would need to type it themselves)
* support for completion without quotes (e.g. `import goo`), but it sounds like we're intentionally avoiding that by looking at the `ImportPath().Span()`. I think that's reasonable, to be clear; most people are going to type `import "|";` and want completions, and don't necessarily need to have them

private/buf/buflsp/server.go
```diff
@@ -455,6 +433,9 @@ func (s *server) Completion(
 		return nil, nil
 	}
 	items := getCompletionItems(ctx, file, params.Position)
+	if len(items) == 0 {
+		return nil, nil
+	}
```
stefanvanburen, line 438: had you seen an error relating to this? (mostly curious; I hadn't myself, but seems in-line with the way we do returns elsewhere.)

emcfarlane: It was copied over from your branch. I thought it's what was causing the fix. Either way feels like we should do these nil checks everywhere to protect against some clients :(

PR #4105, opened 2025-10-23
Provide all references for LSP References

stefanvanburen:
In the case that we have a reference (e.g., a field with a message type), we want references on that message type to return the other usages of that message type. Also, include the definition (if it exists) in the list of references, similar to `gopls`.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -27,7 +27,6 @@ import (
 	"slices"
 	"strings"
 
-	"buf.build/go/standard/xslices"
 	"github.com/bufbuild/protocompile/experimental/ast"
 	"github.com/bufbuild/protocompile/experimental/ast/predeclared"
 	"github.com/bufbuild/protocompile/experimental/ir"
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/symbol.go
```diff
@@ -124,22 +124,38 @@ func (s *symbol) Definition() protocol.Location {
...
+	var references []protocol.Location
+	referenceableKind, ok := s.kind.(*referenceable)
+	if !ok && s.def != nil {
+		// If the symbol isn't referenceable itself, but has a referenceable definition, use the
+		// definition for the references.
+		referenceableKind, ok = s.def.kind.(*referenceable)
+	}
+	if ok {
+		references = append(references, xslices.Map(referenceableKind.references, func(sym *symbol) protocol.Location {
+			return protocol.Location{
```
emcfarlane: Cheaper to range and append instead of building up the slice map.

stefanvanburen: https://github.com/bufbuild/buf/pull/4105/commits/795560d4295f4dd765cdd5d80570f4beace52d2f

emcfarlane (approved):

stefanvanburen:
> On a field it returns the use of that field, plus definition. On definition, all uses of the type.

yeah, you're right, I should have been more precise: I should have said "field type" instead of field. (In your case above with `gopls`, it looks like you're hovering on the field (name), not the field type; I specifically want the field type to include all other usages of that type, plus the definition.)

stefanvanburen:
(Still going to merge as I think this is an improvement regardless, but we can iterate more 😄.)

PR #4108, opened 2025-10-24
LSP add basic autocomplete for nested definitions

emcfarlane:
This improves the autocomplete to handle nested definitions. Support for keywords within definition declarations is now supported. Basic support for predeclared types. Follow up will improve type support to autocomplete type sequences and modify existing type declarations.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -220,13 +220,7 @@ func completionItemsForKeyword(ctx context.Context, file *file, declPath []ast.D
 	}
 
 	tokenSpan := extractAroundToken(file, offset)
-	prefix, suffix := splitSpan(tokenSpan, offset)
-	file.lsp.logger.DebugContext(
-		ctx, "completion: keyword items",
-		slog.String("span", span.Text()),
-		slog.String("prefix", prefix),
-		slog.String("suffix", suffix),
-	)
+	file.lsp.logger.DebugContext(ctx, "completion: keyword items", slog.String("span", span.Text()))
 
 	// If this is an invalid definition at the top level, return top-level keywords.
 	if len(declPath) == 1 {
@@ -234,9 +228,8 @@ func completionItemsForKeyword(ctx context.Context, file *file, declPath []ast.D
 		return slices.Collect(keywordToCompletionItem(
 			topLevelKeywords(),
 			protocol.CompletionItemKindKeyword,
-			position,
-			prefix,
-			suffix,
+			tokenSpan,
+			offset,
 		))
 	}
 
@@ -258,23 +251,20 @@ func completionItemsForKeyword(ctx context.Context, file *file, declPath []ast.D
 			keywordToCompletionItem(
 				messageLevelKeywords(isProto2),
 				protocol.CompletionItemKindKeyword,
-				position,
-				prefix,
-				suffix,
+				tokenSpan,
+				offset,
 			),
 			keywordToCompletionItem(
 				messageLevelFieldKeywords(),
 				protocol.CompletionItemKindKeyword,
-				position,
-				prefix,
-				suffix,
+				tokenSpan,
+				offset,
 			),
 			keywordToCompletionItem(
 				predeclaredTypeKeywords(),
 				protocol.CompletionItemKindClass,
-				position,
-				prefix,
-				suffix,
+				tokenSpan,
+				offset,
 			),
 			// TODO: add custom types.
 		)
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -187,6 +189,154 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
+		return slices.Collect(keywordToCompletionItem(
+			topLevelKeywords(),
+			protocol.CompletionItemKindKeyword,
+			position,
+			prefix,
+			suffix,
+		))
+	}
+
+	parent := declPath[len(declPath)-2]
```
stefanvanburen, line 236: is `len(declPath)` for sure not `0` here?

private/buf/buflsp/completion.go
```diff
@@ -274,26 +411,122 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
...
+func messageLevelKeywords(isProto2 bool) iter.Seq[keyword.Keyword] {
+	return func(yield func(keyword.Keyword) bool) {
+		ok := yield(keyword.Message) &&
+			yield(keyword.Enum) &&
+			yield(keyword.Option) &&
+			yield(keyword.Extend) &&
+			yield(keyword.Oneof) &&
+			yield(keyword.Extensions) &&
+			yield(keyword.Reserved)
+		_ = ok && isProto2 &&
```
stefanvanburen, line 428: lol. have not seen something like this with iterators; fun.

private/buf/buflsp/completion.go
```diff
@@ -187,6 +189,154 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
+	)
+	if parent.Kind() != ast.DeclKindDef {
+		return nil
+	}
+
+	var items iter.Seq[protocol.CompletionItem]
+	parentDef := parent.AsDef()
+	switch parentDef.Classify() {
+	case ast.DefKindMessage:
+		isProto2 := true // TODO: disable !proto2.
```
stefanvanburen: what's the plan for fixing this? It should be easy to determine if we're in a proto2 file, right?

private/buf/buflsp/completion.go
```diff
@@ -187,6 +189,154 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
...
+			keywordToCompletionItem(
+				messageLevelFieldKeywords(),
+				protocol.CompletionItemKindKeyword,
+				position,
+				prefix,
+				suffix,
+			),
+			keywordToCompletionItem(
+				predeclaredTypeKeywords(),
+				protocol.CompletionItemKindTypeParameter,
```
stefanvanburen: nit: I think I'd also make this an `ItemKindKeyword`, since these predeclared type keywords are builtin to protobuf. (I think TypeParameter is maybe for generics?)

```suggestion
				protocol.CompletionItemKindKeyword,
```

private/buf/buflsp/completion.go
```diff
@@ -274,26 +411,122 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
...
+) iter.Seq[protocol.CompletionItem] {
+	return func(yield func(protocol.CompletionItem) bool) {
+		for keyword := range keywords {
+			suggest := keyword.String()
+			if !strings.HasPrefix(suggest, prefix) || !strings.HasSuffix(suggest, suffix) {
+				continue
+			}
+			if !yield(protocol.CompletionItem{
+				Label: suggest,
+				Kind:  kind,
```
stefanvanburen, line 495: nit: I think it would be nice to add a `Detail` or `Documentation` field to these CompletionItems if we have it, that would match info shown in hover? (I'm not sure which field is correct and/or better.) Feel free to file this as a follow-up.

Once we get into custom types as well, I think we'll want to populate the `Deprecated` field if we know it.

Image

Image

stefanvanburen (approved):
tried this out locally and didn't see anything amiss; a couple of small suggestions

private/buf/buflsp/completion.go
emcfarlane: Yep, we must always find at least one, so guarded by above.

private/buf/buflsp/completion.go
emcfarlane: Loop through and find it in the IR. Could also use the image?

private/buf/buflsp/completion.go
emcfarlane: Switched to class to be consistent with gopls (is uses it for inbuilt types like `int` and defined ones). We could use struct for message types?

stefanvanburen: class is fine with me. I think message types should match whatever `gopls` uses for structs, probably struct?

PR #4109, opened 2025-10-24
Run lints on open/save/change; remove breaking checks

stefanvanburen:
This removes `RunChecks` and its background goroutine in favor of just running BuildImage and RunLints if the file is open in the editor. For now, this removes the breaking check entirely; we may add this back later as an LSP Code Action.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -52,9 +52,8 @@ import (
 //
 // Mutating a file is thread-safe.
 type file struct {
-	lsp       *lsp
-	uri       protocol.URI
-	checkWork chan<- struct{}
+	lsp *lsp
+	uri protocol.URI
 
 	file *report.File
 	// Version is an opaque version identifier given to us by the LSP client. This
@@ -120,10 +119,6 @@ func (f *file) Reset(ctx context.Context) {
 // for this file.
 func (f *file) Close(ctx context.Context) {
 	f.Manager().Close(ctx, f.uri)
-	if f.checkWork != nil {
-		close(f.checkWork)
-		f.checkWork = nil
-	}
 	if f.workspace != nil {
 		f.workspace.Release()
 		f.workspace = nil


private/buf/buflsp/file.go
```diff
@@ -298,7 +292,10 @@ func (f *file) Refresh(ctx context.Context) {
 	f.IndexSymbols(ctx)
 
 	progress.Report(ctx, "Running Checks", 5.0/5)
-	f.RunChecks(ctx)
+	if f.IsOpenInEditor() {
+		f.BuildImage(ctx)
+		f.RunLints(ctx)
+	}
```
emcfarlane, line 293: This needs to be moved out of `Refresh`. It's called on `DidChange`, `DidSave` and `DidOpen`.

private/buf/buflsp/file.go
```diff
@@ -936,57 +879,11 @@ func (f *file) newFileOpener() fileOpener {
...
-
-		if data == nil || errors.Is(err, git.ErrInvalidGitCheckout) {
-			return os.Open(file.objectInfo.LocalPath())
-		}
-
-		return xio.CompositeReadCloser(bytes.NewReader(data), xio.NopCloser), err
-	}
-}
-
 // BuildImages builds Buf Images for this file, to be used with linting
```
```suggestion
// BuildImage builds an image for linting.
```

private/buf/buflsp/file.go
stefanvanburen: tbh, I think it's fine to run lints on all of those events; we should optimize for running more often I think? Or are you concerned about how expensive these checks are?

e.g., I wouldn't want to wait until a literal "save" event to see lints in a file; it should show me immediately (`DidOpen`).

If you agree, I'll update the title of the PR to not specifically say "on save".

private/buf/buflsp/file.go
emcfarlane: I think checkWork is unused now.

stefanvanburen: would be nice to have a linter check for "fields with no writes ever" 😅

private/buf/buflsp/file.go
emcfarlane: I guess my main worry is the `BuildImage`. For some projects this could be slow. I think fine to merge as is now. But be nice for this not to block the LSP.

My ideal experience would be to have this run on save. Then if we make a change, discard the current run, and re-run as needed. Be nice for it to be async, and non-blocking with the main loop.

stefanvanburen: sounds good. yep, if it turns out that `BuildImage` ends up being quite expensive we can definitely rework this further.

PR #4110, opened 2025-10-24
Add additionalTextEdit for trailing `;` in import statement

stefanvanburen:
It's nice to send this along if it doesn't exist; makes the import valid.

Ref: https://github.com/bufbuild/buf/pull/4103#pullrequestreview-3370771747

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"iter"
 	"log/slog"
+	"math"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -240,6 +241,26 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
...
+			NewText: ";",
+			// End of line.
+			Range: protocol.Range{
+				Start: protocol.Position{
+					Line:      position.Line,
+					Character: math.MaxInt32 - 1,
+				},
+				End: protocol.Position{
+					Line:      position.Line,
+					Character: math.MaxUint32,
```
emcfarlane, line 258: Doesn't seem to work in neovim/vim. I tried the following change:
```diff
+		importRange := reportSpanToProtocolRange(declImport.Span())

-				Start: protocol.Position{
-					Line:      position.Line,
-					Character: math.MaxInt32 - 1,
-				},
-				End: protocol.Position{
-					Line:      position.Line,
-					Character: math.MaxUint32,
-				},
+				Start: importRange.End,
+				End:   importRange.End,

```

But that doesn't look like it works either (works in VSCode).

stefanvanburen: It works for me in neovim, at least (I can check vim); with both mini.completion & nvim-cmp. (The trick is to make sure you're using the plugin's mapping for complete, e.g. [`<C-y>` for nvim-cmp](https://github.com/hrsh7th/nvim-cmp/blob/a7bcf1d88069fc67c9ace8a62ba480b8fe879025/lua/cmp/config/mapping.lua#L64-L66), assuming you've set up the defaults.)

stefanvanburen: unfortunately, with vim + https://github.com/prabirshrestha/asyncomplete.vim + https://github.com/prabirshrestha/asyncomplete-lsp.vim, it looks like the plugin has a bug where it applies the additionalTextEdit, but then also inserts the completion again?:

```proto
import "google/protobuf/type.proto";import "google/protobuf/type.proto"
```

Going to see about reporting upstream.

stefanvanburen: https://github.com/prabirshrestha/asyncomplete-lsp.vim/issues/51

PR #4112, opened 2025-10-27
LSP add support for type completions

emcfarlane:
This add support for completion of types within the fields of messages in the LSP.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -342,7 +342,7 @@ func completionItemsForField(ctx context.Context, file *file, declPath []ast.Dec
 	for range strings.FieldsSeq(strings.TrimPrefix(typeSuffix, tokenSuffix)) {
 		suffixCount++
 	}
-	// Limit completions based on the following hyeristics:
+	// Limit completions based on the following heuristic:
 	// - Show modifiers for the first two types
 	// - Only show types on the final type
 	showModifiers := prefixCount <= 2


private/buf/buflsp/completion.go
```diff
@@ -504,9 +562,75 @@ func keywordToCompletionItem(
...
+				continue // Unsupported kind, skip it.
+			}
+			if !yield(protocol.CompletionItem{
+				Label: label,
+				Kind:  kind,
+				TextEdit: &protocol.TextEdit{
+					Range:   editRange,
+					NewText: label,
+				},
+				// TODO: If this type's file is not currently imported add an additional edit.
```
stefanvanburen, line 633: this will be awesome :)

private/buf/buflsp/completion.go
```diff
@@ -504,9 +562,75 @@ func keywordToCompletionItem(
...
+				kind = protocol.CompletionItemKindModule
+			case ir.SymbolKindField, ir.SymbolKindEnumValue, ir.SymbolKindExtension, ir.SymbolKindOneof, ir.SymbolKindMethod:
+				// These should be skipped by IsType() filter.
+			}
+			if kind == 0 {
+				continue // Unsupported kind, skip it.
+			}
+			if !yield(protocol.CompletionItem{
+				Label: label,
+				Kind:  kind,
```
stefanvanburen, line 627: Can we add the `Deprecated:` field here similar to the others?

private/buf/buflsp/completion.go
```diff
@@ -504,9 +562,75 @@ func keywordToCompletionItem(
...
 
-// joinCompletionItems returns a sequence of sequences.
-func joinCompletionItems(itemIters ...iter.Seq[protocol.CompletionItem]) iter.Seq[protocol.CompletionItem] {
+// typeReferencesToCompletionItems returns completion items for user-defined types (messages, enums, etc).
+func typeReferencesToCompletionItems(
+	current *file,
+	span report.Span,
+	offset int,
+) iter.Seq[protocol.CompletionItem] {
+	fileSymobolTypesIter := func(yield func(*file, *symbol) bool) {
```
stefanvanburen: nit: typo: fileSymbolTypesIter

private/buf/buflsp/completion.go
```diff
@@ -504,9 +562,75 @@ func keywordToCompletionItem(
...
+			case ir.SymbolKindMessage:
+				kind = protocol.CompletionItemKindStruct
+			case ir.SymbolKindEnum:
+				kind = protocol.CompletionItemKindEnum
+			case ir.SymbolKindService:
+				kind = protocol.CompletionItemKindInterface
+			case ir.SymbolKindScalar:
+				kind = protocol.CompletionItemKindClass
+			case ir.SymbolKindPackage:
+				kind = protocol.CompletionItemKindModule
```
stefanvanburen, line 614: It's a little too bad that there are all sorts of different types for the same thing; we now have a similar mapping for both [symbols](https://github.com/bufbuild/buf/blob/a3723f83ad2bf79283ce55678548ede4fd203e6b/private/buf/buflsp/symbol.go#L297-L316) and [semantic types](https://github.com/bufbuild/buf/blob/a3723f83ad2bf79283ce55678548ede4fd203e6b/private/buf/buflsp/server.go#L506-L528). It would be great if we were confirming they all lined up somehow.

private/buf/buflsp/completion.go
```diff
@@ -308,33 +315,84 @@ func completionItemsForField(ctx context.Context, file *file, declPath []ast.Dec
...
+	prefixCount := 0
+	for range strings.FieldsSeq(strings.TrimSuffix(typePrefix, tokenPrefix)) {
+		prefixCount++
+	}
+	suffixCount := 0
+	for range strings.FieldsSeq(strings.TrimPrefix(typeSuffix, tokenSuffix)) {
+		suffixCount++
+	}
+	showModifiers := prefixCount <= 2
+	showTypes := suffixCount == 0
```
stefanvanburen, line 349: can we add comments here as to the cases we're trying to handle here? I think I mostly understand it now but would be helpful later.

private/buf/buflsp/completion.go
emcfarlane: Need some symbol to a "kind" that is then mapped to each of these protocol kinds. Leaving for now, but would be good.

PR #4113, opened 2025-10-27
Refactor LSP checks to run at workspace level

emcfarlane:
This PR significantly refactors how checks are run in the LSP to improve performance and reduce locking. Previously, each file maintained its own check state and would independently run checks on each update. Checks are now moved onto the workspace level, which allows us to batch update lint checks for all files within the workspace. Workspace checks run asynchronously without holding the LSP lock to also eliminate blocking.

The concurrency model has been improved. Before a 3 second debounce was used. This added latency and and could show stale diagnostics if files were edited during the debounce window. Now, we instantly cancel any in-progress checks when a file is edited to immediately start fresh check run. This ensures diagnostics are always accurate for the current file version and responds much faster to changes.

Currently we refresh the workspace on each file update. This is excessive as workspaces only update when the filesystem is written to. Workspace refreshes have been moved to file open and save. On workspace update we would also re-read all files from disk. This is unnecessary as the workspace contains all file contents. This rehydrating logic now simply uses the workspace to read the file contents.

This PR continues on from https://github.com/bufbuild/buf/pull/4109


emcfarlane:
Closed by #4121 and #4120

PR #4120, opened 2025-10-28
LSP checks run async

emcfarlane:
This PR updates checks to run asynchronously, preventing the LSP from blocking on long running lint checks. Both image building and lint checks are now run outside the main LSP lock, only acquiring it again when publishing diagnostics. Any checks running from a previous update are cancelled to avoid stale diagnostics.

### DIFF
--- a/private/buf/buflsp/image.go
+++ b/private/buf/buflsp/image.go
@@ -90,9 +90,10 @@ func buildImage(
 			})
 		}
 	}
-	if compiled[0] == nil {
-		return nil, nil
+	if len(compiled) == 0 || compiled[0] == nil {
+		return nil, nil // Image failed to build.
 	}
+	compiledFile := compiled[0]
 
 	syntaxMissing := make(map[string]bool)
 	pathToUnusedImports := make(map[string]map[string]bool)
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/file.go
```diff
@@ -683,96 +679,99 @@ func (f *file) SymbolAt(ctx context.Context, cursor protocol.Position) *symbol {
...
+			return
+		default:
+		}
+
+		f.lsp.lock.Lock()
+		defer f.lsp.lock.Unlock()
+
+		select {
+		case <-ctx.Done():
+			f.lsp.logger.DebugContext(ctx, "checks cancelled after waiting ", slog.String("uri", f.uri.Filename()), xslog.ErrorAttr(ctx.Err()))
```
stefanvanburen: nit:

```suggestion
			f.lsp.logger.DebugContext(ctx, "checks cancelled after waiting for file lock", slog.String("uri", f.uri.Filename()), xslog.ErrorAttr(ctx.Err()))
```

private/buf/buflsp/file.go
```diff
@@ -683,96 +679,99 @@ func (f *file) SymbolAt(ctx context.Context, cursor protocol.Position) *symbol {
...
+	f.cancelChecks = cancel
+
+	go func() {
+		image, diagnostics := buildImage(ctx, path, f.lsp.logger, opener)
+		if image == nil {
+			f.lsp.logger.DebugContext(ctx, "check cancelled on image build", slog.String("uri", f.uri.Filename()))
+			return
+		}
+
+		f.lsp.logger.Debug(fmt.Sprintf("check running lint for %q in %v", f.uri, module.FullName()))
```
stefanvanburen: nit: want to fix this to avoid `fmt.Sprintf` in a logger, and use `DebugContext` instead?

PR #4121, opened 2025-10-28
LSP only refresh workspace on open/save

emcfarlane:
This updates the LSP to only refresh the file's workspace on open or save. Previously, every update would trigger a full workspace rebuild. This was unnecessary since file tree changes (additions/deletions) only occur when files are opened or saved. With #4120 and this change we remove the progress updates. These are no longer necessary as the LSP latency is reduced.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -685,14 +685,14 @@ func (f *file) RunChecks(ctx context.Context) {
 	}
 	f.CancelChecks(ctx)
 
-	path := f.objectInfo.Path()
 	workspace := f.workspace.Workspace()
 	module := f.workspace.GetModule(f.uri)
 	checkClient := f.workspace.CheckClient()
-	if workspace == nil || module == nil || checkClient == nil {
+	if workspace == nil || module == nil || checkClient == nil || f.objectInfo == nil {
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/file.go
```diff
@@ -307,12 +279,19 @@ func (f *file) IndexSymbols(ctx context.Context) {
...
-	f.referenceSymbols = nil
-	f.referenceableSymbols = make(map[string]*symbol)
+	clear(f.symbols)
+	f.symbols = f.symbols[:0]
+	clear(f.referenceSymbols)
+	f.referenceSymbols = f.referenceSymbols[:0]
+	clear(f.referenceableSymbols)
+	if f.referenceableSymbols == nil {
+		f.referenceableSymbols = make(map[string]*symbol)
+	}
```
stefanvanburen: is this actually fixing something, or are we just trying to assert here that we can definitely gc the old values? feels like the old version is clearer and less reliant on knowing the tricks to make gc run?

stefanvanburen:
(I think this looks fine FWIW; may just want to land some of your other PRs and get this all merged and I'll take another look.)

private/buf/buflsp/file.go
emcfarlane: Reverted to cleaner version

PR #4124, opened 2025-10-28
Add import of missing file on completion

stefanvanburen:
If the completion item is not currently imported in the file, this will add the missing import.

We can't necessarily assume that the file is already formatted by `buf format`, so instead, we just either shove the import directly after any existing imports (which will be reordered by `buf format`, if used); if no imports, right after `package`; if no package, right after `syntax`, and otherwise just balk.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -642,7 +642,6 @@ func typeReferencesToCompletionItems(
 		case !current.ir.AST().Syntax().IsZero():
 			line = current.ir.AST().Syntax().Span().EndLoc().Line
 		default:
-			// We won't add an import in this case.
 			line = 0
 		}
 		if line < 0 || line > math.MaxUint32 {
@@ -697,8 +696,7 @@ func typeReferencesToCompletionItems(
 			symbolFile := symbol.ir.File().Path()
 			_, hasImport := currentImportPaths[symbolFile]
 			var additionalTextEdits []protocol.TextEdit
-			var zero protocol.Position
-			if !hasImport && insertPosition != zero {
+			if !hasImport {
 				additionalTextEdits = append(additionalTextEdits, protocol.TextEdit{
 					NewText: "import " + `"` + symbolFile + `";` + "\n",
 					Range: protocol.Range{


private/buf/buflsp/completion.go
```diff
@@ -647,16 +651,74 @@ func typeReferencesToCompletionItems(
 			if _, ok := symbol.ir.Deprecated().AsBool(); ok {
 				isDeprecated = true
 			}
+			symbolFile := symbol.ir.File().Path()
+			missingImport := true
+			var lastImportLine uint32
+			for currentFileImport := range seq.Values(current.ir.Imports()) {
```
emcfarlane: This should be moved above the yield func. It's the same across all symbols. edit: thought it was building up the map.

private/buf/buflsp/completion.go
```diff
@@ -647,16 +651,74 @@ func typeReferencesToCompletionItems(
...
+					var line int
+					switch {
+					case !current.ir.AST().Package().IsZero():
+						line = current.ir.AST().Package().Span().StartLoc().Line
+					case !current.ir.AST().Syntax().IsZero():
+						line = current.ir.AST().Syntax().Span().StartLoc().Line
+					default:
+						// We won't add an import in this case.
+						line = 0
+					}
```
emcfarlane: Same here: this is same across all symbols.

private/buf/buflsp/completion.go
```diff
@@ -647,16 +651,74 @@ func typeReferencesToCompletionItems(
...
+			symbolFile := symbol.ir.File().Path()
+			missingImport := true
+			var lastImportLine uint32
+			for currentFileImport := range seq.Values(current.ir.Imports()) {
+				l := currentFileImport.Decl.Span().StartLoc().Line
+				if l < 0 || l > math.MaxUint32 {
+					return
+				}
+				lastImportLine = max(uint32(l), lastImportLine)
+				if currentFileImport.Path() == symbolFile {
```
emcfarlane: Maybe should be a `map[string]struct{}` comparison?

private/buf/buflsp/completion.go
stefanvanburen: hm, I can pull it out into a func, but it [needs the `symbolFile` to determine if we're missing the import](https://github.com/bufbuild/buf/blob/3e703f63ac719d2cdf20c9821f4427d90747af24/private/buf/buflsp/completion.go#L654). so not the same across all symbols?

stefanvanburen: ah, I see what you mean: grab all the imports up front (and the last location), and if we're missing the current symbol's import do that. let me restructure.

private/buf/buflsp/completion.go
stefanvanburen: lifted it out and made it a map instead: https://github.com/bufbuild/buf/pull/4124/commits/d2d93364bf666cb55af7c8aeb5d25c84c473b6fb

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/d2d93364bf666cb55af7c8aeb5d25c84c473b6fb

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/d2d93364bf666cb55af7c8aeb5d25c84c473b6fb

private/buf/buflsp/completion.go
```diff
@@ -612,6 +612,47 @@ func typeReferencesToCompletionItems(
 			}
 		}
 	}
+	var lastImportLine uint32
+	currentImportPaths := map[string]struct{}{}
+	for currentFileImport := range seq.Values(current.ir.Imports()) {
+		l := currentFileImport.Decl.Span().StartLoc().Line
+		if l < 0 || l > math.MaxUint32 {
+			return nil
```
emcfarlane: Ah this shouldn't return here?

stefanvanburen: I figured this was kind of an exceptional situation 😅. I guess just `continue` instead?

private/buf/buflsp/completion.go
```diff
@@ -612,6 +612,47 @@ func typeReferencesToCompletionItems(
...
+		case !current.ir.AST().Package().IsZero():
+			line = current.ir.AST().Package().Span().StartLoc().Line
+		case !current.ir.AST().Syntax().IsZero():
+			line = current.ir.AST().Syntax().Span().StartLoc().Line
+		default:
+			// We won't add an import in this case.
+			line = 0
+		}
+		if line < 0 || line > math.MaxUint32 {
+			return nil
```
emcfarlane: And here.

private/buf/buflsp/completion.go
emcfarlane: This is only checking for gosec? Also does the iterator over imports not iterate in with line number increasing?

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/1da0f9e72e25000acefe9e41f22077807b8508e6

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/1da0f9e72e25000acefe9e41f22077807b8508e6

private/buf/buflsp/completion.go
```diff
@@ -612,6 +612,47 @@ func typeReferencesToCompletionItems(
...
+		}
+	} else {
+		// If lastImportLine is 0, we have no imports in this file; put it after `package`, if
+		// package exists. Otherwise, after `syntax`, if it exists. Otherwise, balk.
+		// NOTE: We simply want to add the import on the next line (which may not be how `buf
+		// format` would format the file); we leave the overall file formatting to `buf format`.
+		var line int
+		switch {
+		case !current.ir.AST().Package().IsZero():
+			line = current.ir.AST().Package().Span().StartLoc().Line
```
emcfarlane: Should be `EndLoc()`. Most likely on the same line, but should just use end.

private/buf/buflsp/completion.go
```diff
@@ -612,6 +612,47 @@ func typeReferencesToCompletionItems(
...
+		case !current.ir.AST().Syntax().IsZero():
+			line = current.ir.AST().Syntax().Span().StartLoc().Line
+		default:
+			// We won't add an import in this case.
+			line = 0
```
emcfarlane, line 645: Inserting at the top of the file feels find if theres nothing else (unlikely).

private/buf/buflsp/completion.go
stefanvanburen: > Also does the iterator over imports not iterate in with line number increasing?

I guess it does; I wasn't sure if we ought to rely on that behavior in practice, though. I didn't know if a `seq.Indexer` ordering could be relied on.

emcfarlane (approved):

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/f7f9ae7e2fc92122f6d98759fa52cfe8a808d421

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4124/commits/1b59c6976b4ab02eec16ff15933c6f1839f6c4a2

PR #4126, opened 2025-10-28
LSP fix triggers for autocomplete

emcfarlane:
This fixes trigger values for autocomplete to work across more editors, the `" "` char was causing issues in vim. Also a more robust handling for syntax completion.


stefanvanburen (approved):

private/buf/buflsp/completion.go
```diff
@@ -825,7 +914,7 @@ func splitSpan(span report.Span, offset int) (prefix string, suffix string) {
 }
 
 func offsetInSpan(span report.Span, offset int) bool {
```
emcfarlane, line 892: This was just broken.

private/buf/buflsp/completion.go
```diff
@@ -106,13 +106,68 @@ func completionItemsForDeclPath(ctx context.Context, file *file, declPath []ast.
...
-	var prefix string
+	positionLocation := file.file.InverseLocation(int(position.Line)+1, int(position.Character)+1, positionalEncoding)
+	offset := positionLocation.Offset
+
+	valueSpan := syntaxDecl.Value().Span()
+	start, end := valueSpan.Start, valueSpan.End
+	valueText := valueSpan.Text()
+
+	// Break on newlines to split an unintended capture.
+	if index := strings.IndexByte(valueText, '\n'); index != -1 {
```
emcfarlane, line 120: This is needed on `syntax = "_` where theres no closing `"` the token reports the whole file as the value capture. Must split. This is the same logic as import statements so we can probably factor this bit out too.

stefanvanburen (changes requested):
I think this is missing something as we don't get the completion for the files in different packages/files(?) now; this PR:

Image

main:

Image

emcfarlane:
@stefanvanburen this looks fixed by [60fdaef](https://github.com/bufbuild/buf/pull/4126/commits/60fdaefdc7c13103d2d38357fb0c16128943d6d4) and I think also addresses the `.` continuation syntax. Tried in vsccode/nvim/vim.

private/buf/buflsp/completion.go
```diff
@@ -641,7 +714,9 @@ func typeReferencesToCompletionItems(
 	packagePrefix := string(current.ir.Package()) + "."
 	return func(yield func(protocol.CompletionItem) bool) {
 		editRange := reportSpanToProtocolRange(span)
-		prefix, suffix := splitSpan(span, offset)
+		prefix, _ := splitSpan(span, offset)
+		// Prefix filter on the trigger character '.', if present.
+		prefix = prefix[:strings.LastIndexByte(prefix, '.')+1]
```
stefanvanburen, line 719:
```suggestion
```

I don't think this is necessary

private/buf/buflsp/completion.go
```diff
@@ -660,7 +735,7 @@ func typeReferencesToCompletionItems(
 			} else if strings.HasPrefix(label, packagePrefix) {
 				label = label[len(packagePrefix):]
 			}
-			if !strings.HasPrefix(label, prefix) || !strings.HasSuffix(label, suffix) {
+			if !strings.HasPrefix(label, prefix) {
```
stefanvanburen, line 738: I think this should be `!strings.Contains(label, prefix)`, so that we can match types that don't necessarily start with the "prefix" (i.e., already typed text):

Image

stefanvanburen (commented):
two small things that I think would get this working as I would expect

private/buf/buflsp/completion.go
stefanvanburen: (This is showing what the completions are with that change applied; not sure why GitHub won't let me suggest it.)

emcfarlane: We don't control how the client shows the list response. In nvim with the recommend config (plus `vim.o.completeopt = 'menuone,noselect,popup'` for sanity). If we remove _all_ prefix filtering and ask for a completion on `C_` we get the following behaviour:

https://github.com/user-attachments/assets/b880fa30-1310-4d1e-9367-e13c41b5d5bb

So we could shorten the given list to the client on contains, but as soon as they type it will be filtered away on the client side.

private/buf/buflsp/completion.go
emcfarlane: This is necessary for fixing the above https://github.com/bufbuild/buf/pull/4126#pullrequestreview-3391161129 and working on `google._`.

private/buf/buflsp/completion.go
emcfarlane: In VSCode this is shown as: 
Screenshot 2025-10-29 at 5 58 44 PM

stefanvanburen: > So we could shorten the given list to the client on contains, but as soon as they type it will be filtered away on the client side.

I think that's fine: if they've typed `C`, we give them everything matching `C`; the client can then filter down without making another request based on the local list they've got. If they want further filtering (server-side) they can manually make another request?

private/buf/buflsp/completion.go
stefanvanburen: hm, I don't think so; locally I've dropped this line (and applied the below change) and I get the results I expect:

image

image

I won't hold up on this, I just don't think it's necessary.

private/buf/buflsp/completion.go
stefanvanburen: > In VSCode this is shown as

looks right to me? or are you expecting something else?

private/buf/buflsp/completion.go
emcfarlane: Screenshot 2025-10-29 at 6 04 34 PM

This would break this style of completion in VScode. If we `strings.Contains(label, "buf.Commit")` it won't find any matches. Its the client that is responsible to filter the list.

emcfarlane: So instead of the `Contains` we break on the trigger char `.` where an autocomplete event was requested and prefix filter on that but let the client handle the completion.

stefanvanburen: ah, ok, that makes sense to me, thanks for the example!

private/buf/buflsp/completion.go
emcfarlane: It's showing the different behaviour between VSCode and nvim omnifunc completions.

PR #4127, opened 2025-10-28
Skip adding import for symbols in same file

stefanvanburen:
Otherwise, we'll get circular imports by trying to import ourselves 😸.


private/buf/buflsp/completion.go
emcfarlane, line 616:
```suggestion
	currentImportPaths := map[string]struct{}{current.ir.Path(): struct{}{}}
```

emcfarlane (approved):

private/buf/buflsp/completion.go
stefanvanburen: hah, tricky. I guess we could represent it that way, I just typically don't think of the current file as "imported" by itself. Gonna just leave as is, but LMK if you feel strongly and we can switch it.

PR #4128, opened 2025-10-29
Tighten up item kinds

stefanvanburen:
* predeclared types (`string`, `int64`) are more like keywords than classes
* simplify typeReferencesToCompletionItems logic to only map types
* add scalar handling to symbols

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -262,7 +262,7 @@ func completionItemsForKeyword(ctx context.Context, file *file, declPath []ast.D
 			),
 			keywordToCompletionItem(
 				predeclaredTypeKeywords(),
-				protocol.CompletionItemKindClass,
+				protocol.CompletionItemKindKeyword,
 				tokenSpan,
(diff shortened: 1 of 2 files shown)


private/buf/buflsp/completion.go
```diff
@@ -643,8 +643,16 @@ func typeReferencesToCompletionItems(
...
 		prefix, suffix := splitSpan(span, offset)
 		for _, symbol := range fileSymbolTypesIter {
-			if !symbol.ir.Kind().IsType() {
-				continue
+			// We only support types in this completion instance, and not scalar values, which leaves us
+			// with messages and enums.
+			var kind protocol.CompletionItemKind
+			switch symbol.ir.Kind() {
+			case ir.SymbolKindMessage:
+				kind = protocol.CompletionItemKindClass // Messages are like classes
```
stefanvanburen, line 651: we have message -> class mapping in symbols.go, which @pkwarren reminded me of; we should be consistent here whichever way we go.

PR #4130, opened 2025-10-29
Support modifiers in LSP semantic tokens

stefanvanburen:
Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokenModifiers

We may as well populate these if we can while we're at it. Starting with deprecated and "defaultLibrary", aka builtin symbols, seems useful to report to clients, which can colorize these differently in the editor to call out their difference.

### DIFF
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -36,6 +36,9 @@ const (
 	maxSymbolResults = 1000
 )
 
+// The subset of SemanticTokenTypes that we support.
+// Must match the order of [semanticTypeLegend].
+// https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_semanticTokens
 const (
 	semanticTypeType = iota
 	semanticTypeStruct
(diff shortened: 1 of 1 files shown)


stefanvanburen:
nvim colorizes the deprecated `GetLabels` RPC differently than the default:

image

emcfarlane (approved):

private/buf/buflsp/server.go
```diff
@@ -45,15 +45,25 @@ const (
 	semanticTypeInterface
 	semanticTypeMethod
 	semanticTypeDecorator
+	semanticTypeNamespace
+)
+
+const (
+	semanticModifierDeprecated = 1 << iota
```
emcfarlane, line 59: Maybe just link the docs here for reference.

stefanvanburen: added a bit more context in https://github.com/bufbuild/buf/pull/4130/commits/933762fb68e7e261d8f0a2b061f88fadf8b230a6.

PR #4132, opened 2025-10-30
LSP add goto definition on parent names

emcfarlane:
This PR adds support for goto definition based on the cursor position in the fields FullName. For example: `google.protobuf.Timestamp` can go to both the type `Timestamp` or the package `google.protobuf` within the file `google/protobuf/timestamp.proto`. We may wish to extend this behaviour to also cover going to every file within the package `google.protobuf`.

### DIFF
--- a/private/buf/buflsp/server.go
+++ b/private/buf/buflsp/server.go
@@ -371,9 +371,7 @@ func (s *server) definition(ctx context.Context, uri protocol.URI, position prot
 		return nil, nil
 	}
 
-	return []protocol.Location{
-		symbol.Definition(),
-	}, nil
+	return symbol.Definitions(position), nil
 }
 
 // References is the entry point for get-all-references.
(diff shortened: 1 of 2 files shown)


stefanvanburen (commented):
two things, and one weird thing I noticed:

* I think hover should have the same logic as "go-to definition"; that is, if I'm hovering on a symbol and get docs, I would expect go-to definition to bring me to the location where that doc comment comes from. (I think we should add this logic to hover.)
* The UX feels a little weird on the parent types, given the option `(buf.validate.field).string.max_len`, using go-to-definition:
  * `max_len` does sorta what I expect (it brings me to `BytesRules.max_len`; it should bring me to `StringRules.max_len` 😅)
  * `string` does sorta what I expect (it brings me to `BytesRules`, which also defines a `max_len`, but should bring me to `StringRules`)
  * `field` takes me to the `package buf.validate` line in validate.proto;
* (likewise with `google.protobuf.Timestamp`, I'm not sure if it's better UX that `google` doesn't do anything and `protobuf` just brings me to a file, rather than have the whole thing go to `Timestamp`?)

emcfarlane:
Closing in favour of breaking the symbols up, cc @doriable

PR #4134, opened 2025-10-30
LSP fix type completion on prefix modifiers

emcfarlane:
This improves the type completion to be more robust. Previously, we would incorrectly ignore type completions when a type modifier was used on the last line of a declaration. This was due to the declaration identifying the value as the name value. Now, we split on the line and capture the values from the token stream. This avoids issues with invalid declaration spans.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -233,40 +233,6 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
 		slog.Int("path_depth", len(declPath)),
 	)
 
-	// Extract the span of the definition to include the type and name.
-	// We only complete in types, but the type may be capture in the name for an unfinished def.
-	span := def.Span()
-	if !def.Type().IsZero() {
-		span.Start = def.Type().Span().Start
-	}
-	if !def.Name().IsZero() {
-		span.End = def.Name().Span().End
-	}
-	positionLocation := file.file.InverseLocation(int(position.Line)+1, int(position.Character)+1, positionalEncoding)
-	offset := positionLocation.Offset
-	if !offsetInSpan(span, offset) {
-		file.lsp.logger.DebugContext(
-			ctx,
-			"completion: ignoring definition outside span",
-			slog.String("kind", def.Classify().String()),
-		)
-		return nil
-	}
-
-	tokenSpan := extractAroundToken(file, offset)
-	tokenPrefix, tokenSuffix := splitSpan(tokenSpan, offset)
-	typeSpan := extractLine(span, offset)
-	typePrefix, typeSuffix := splitSpan(typeSpan, offset)
-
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -224,79 +224,67 @@ func completionItemsForPackage(ctx context.Context, file *file, syntaxPackage as
...
-		return nil
-	case ast.DefKindEnum:
-		return nil
-	case ast.DefKindField:
-		return completionItemsForField(ctx, file, declPath, def, position)
-	case ast.DefKindInvalid:
-		return completionItemsForKeyword(ctx, file, declPath, def, position)
+	case ast.DefKindMessage, ast.DefKindService, ast.DefKindEnum:
+		return nil // ignored
+	case ast.DefKindField, ast.DefKindInvalid:
```
stefanvanburen, line 239: I realize this isn't changing, but why do we handle `Invalid` here? (might be a useful comment.)

private/buf/buflsp/completion.go
```diff
@@ -565,6 +470,18 @@ func completionItemsForImport(ctx context.Context, file *file, declImport ast.De
 	return items
 }
 
+// declarationSet is the set of keywords that starts a declaration.
+var declarationSet = func() map[string]struct{} {
```
stefanvanburen: fancy

private/buf/buflsp/completion.go
emcfarlane: Added a doc comment 👍

stefanvanburen: thanks!

private/buf/buflsp/completion.go
```diff
@@ -462,14 +354,46 @@ func completionItemsForField(ctx context.Context, file *file, declPath []ast.Dec
...
+		}
+	case ast.DefKindMethod:
+		if showKeywords {
+			iters = append(iters,
+				keywordToCompletionItem(
+					optionKeyword(),
+					protocol.CompletionItemKindKeyword,
+					tokenSpan,
+					offset,
+				),
```
stefanvanburen, line 409: ~~should we also show `stream` here? (and message types?)~~

Ah, nvm, played with it: this is for within the brackets, where `option` is:

```proto
  rpc Download(DownloadRequest) returns (DownloadResponse) {
  }
```

emcfarlane: Yeah working on this next, I just pushed up an improvement for capturing the type/name. I noticed this change before was now recommended types too often. Hopefully fine now but let me know if you see any odd behaviour.

It did this because the span of the field declaration wasn't capturing the line:
```
   string _ 
   string field = 1;
```
This cursor `_` was being interpreted as a field declaration with type `_ string field = 1;` so would show types. We now just ignore the declaration span entirely, and walk the token stream to build up the line as: `string _` and then use the heuristics of no field modifier so must be the name value.

PR #4137, opened 2025-10-31
Add links in docs for predeclared types, remote dependencies, and WKTs

doriable:
This adds the links back for predeclared types, remote deps,
and WKTs in the LSP doc hovers.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -191,7 +191,6 @@ func (s *symbol) LogValue() slog.Value {
 //
 // Returns the empty string if no docs are available.
 func (s *symbol) FormatDocs() string {
-	var def ast.DeclDef
 	switch s.kind.(type) {
 	case *imported:
 		imported, _ := s.kind.(*imported)
@@ -261,22 +260,18 @@ func (s *symbol) FormatDocs() string {
 			comments = append(
 				comments,
 				"",
-				"This symbol is a Protobuf builtin. [Learn more on protobuf.com.](https://protobuf.com/docs/language-spec#field-types)",
+				fmt.Sprintf(
+					"`%s` is a Protobuf builtin. [Learn more on protobuf.com.](https://protobuf.com/docs/language-spec#field-types)",
+					builtin.predeclared,
+				),
 			)
 			return strings.Join(comments, "\n")
 		}
 		return ""
-	case *referenceable:
-		referenceable, _ := s.kind.(*referenceable)
-		def = referenceable.ast
-	case *static:
-		static, _ := s.kind.(*static)
-		def = static.ast
-	case *reference:
-		reference, _ := s.kind.(*reference)
-		def = reference.def
+	case *referenceable, *static, *reference:
+		return s.getDocsFromComments()
 	}
-	return s.def.getComments(def)
+	return ""
 }
 
 // GetSymbolInformation returns the protocol symbol information for the symbol.
@@ -349,10 +344,27 @@ func protowireTypeForPredeclared(name predeclared.Name) protowire.Type {
 	return protowire.BytesType
 }
 
-func (s *symbol) getComments(def ast.DeclDef) string {
+// getDocsFromComments is a helper function that gets the doc string from the comments from
+// the definition AST, if available.
+// This helper function expects that imports, tags, and predeclared (builtin) types are
+// already handled, since those types currently do not get docs from their comments.
+func (s *symbol) getDocsFromComments() string {
+	var def ast.DeclDef
+	switch s.kind.(type) {
+	case *referenceable:
+		referenceable, _ := s.kind.(*referenceable)
+		def = referenceable.ast
+	case *static:
+		static, _ := s.kind.(*static)
+		def = static.ast
+	case *reference:
+		reference, _ := s.kind.(*reference)
+		def = reference.def
+	}
 	if def.IsZero() {
 		return ""
 	}
+
 	var comments []string
 	// We drop the other side of "Around" because we only care about the beginning -- we're
 	// traversing backwards for leading comemnts only.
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/symbol.go
```diff
@@ -339,7 +345,7 @@ func protowireTypeForPredeclared(name predeclared.Name) protowire.Type {
 	return protowire.BytesType
 }
 
-func getCommentsFromDef(def ast.DeclDef) string {
+func (s *symbol) getComments(def ast.DeclDef) string {
```
emcfarlane: This func signature is a bit odd because it presumes that the def must be the same symbol?

private/buf/buflsp/symbol.go
```diff
@@ -253,6 +254,11 @@ func (s *symbol) FormatDocs() string {
 		builtin, _ := s.kind.(*builtin)
 		comments, ok := builtinDocs[builtin.predeclared.String()]
 		if ok {
+			comments = append(
+				comments,
+				"",
+				"This symbol is a Protobuf builtin. [Learn more on protobuf.com.](https://protobuf.com/docs/language-spec#field-types)",
```
stefanvanburen: nit: should we `s/This symbol/<symbol-name>`, for consistency with the below link?

private/buf/buflsp/symbol.go
```diff
@@ -361,6 +367,41 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
+		}
+
+		var bsrHost, bsrAnchor, bsrTooltip string
+		if s.file.IsWKT() {
+			bsrHost = "buf.build/protocolbuffers/wellknowntypes"
+		} else if fileInfo, ok := s.file.objectInfo.(bufmodule.FileInfo); ok {
+			bsrHost = fileInfo.Module().FullName().String()
+		}
+		bsrTooltip = string(s.ir.FullName())
+		if !hasAnchor {
```
emcfarlane, line 412: What does `hasAnchor` mean here?

private/buf/buflsp/symbol.go
```diff
@@ -361,6 +367,41 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
+			components := strings.Split(bsrTooltip, ".")
+			bsrTooltip = strings.Join(components[:len(components)-1], ".")
```
emcfarlane: Don't need the string manipulation if you keep the type as the `ir.FullName`, then `bsrTooltip.Parent()`.

private/buf/buflsp/symbol.go
```diff
@@ -361,6 +367,41 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
+		}
+		bsrAnchor = bsrTooltip
+		if isExtension {
+			bsrAnchor = "extensions"
```
emcfarlane: Feels like these should be constants.

private/buf/buflsp/symbol.go
```diff
@@ -361,6 +367,41 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
+			comments = append(comments, fmt.Sprintf(
+				"\n[`%s` on the Buf Schema Registry](https://%s/docs/main:%s#%s)",
+				bsrTooltip,
+				bsrHost,
+				s.file.ir.Package(),
+				bsrAnchor,
+			))
+		}
+	}
 	return strings.Join(comments, "")
```
emcfarlane: Maybe a `strings.Builder` here would be better?

private/buf/buflsp/symbol.go
emcfarlane: Ah, thought this was an LSP protocol constant. See it's just for building the URL.

private/buf/buflsp/symbol.go
```diff
@@ -361,6 +367,41 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
+			components := strings.Split(bsrTooltip, ".")
+			bsrTooltip = strings.Join(components[:len(components)-1], ".")
+		}
+		bsrAnchor = bsrTooltip
+		if isExtension {
+			bsrAnchor = "extensions"
+		}
+		if bsrHost != "" {
+			comments = append(comments, fmt.Sprintf(
+				"\n[`%s` on the Buf Schema Registry](https://%s/docs/main:%s#%s)",
```
stefanvanburen, line 422: nit: I'm not immediately seeing why, but something here seems to leave a trailing newline (or _maybe_, it's the absence of a trailing newline) in neovim. (seems fine in VS Code though.)

Image

private/buf/buflsp/symbol.go
doriable: Yeah... I refactored this to be a little bit clearer, hopefully, let me know what you think!

private/buf/buflsp/symbol.go
doriable: Added comments earlier for this, hopefully this is a bit more clear.

private/buf/buflsp/symbol.go
doriable: This makes sense to me, updated.

private/buf/buflsp/symbol.go
doriable: Hmm, so maybe I have my hover configured slightly differently, but I'm not seeing a trailing newline... different clients are handling spacing a little differently though, and that is bothering me, so going to take a look.

Right now, we are writing the comments directly into the `strings.Builder` (EDIT: kind of -- we have the markdown processing, but we aren't adding whitespace there...), and then only adding a newline before the link, so I'm not sure how much more we should adjust that.

private/buf/buflsp/symbol.go
```diff
@@ -365,7 +383,51 @@ func getCommentsFromDef(def ast.DeclDef) string {
...
 	slices.Reverse(comments)
-	return strings.Join(comments, "")
+
+	var docs strings.Builder
+	for _, comment := range comments {
+		docs.WriteString(comment)
+	}
+
+	// If the file is a remote dependency, link to BSR docs.
+	if s.def.file != nil && !s.def.file.IsLocal() {
```
emcfarlane, line 393: Image 

Same issue as commented here: https://github.com/bufbuild/buf/pull/4132#pullrequestreview-3400102295 . The docs for the string type link to the bytes type?

private/buf/buflsp/symbol.go
stefanvanburen: Looks fixed on the latest commit; resolving this now.

Image

private/buf/buflsp/symbol.go
doriable: Hmm, this is an issue with the symbol resolution logic -- fwiw, I believe #4139 fixes this.

PR #4139, opened 2025-11-01
Split spans for nested message value elements

doriable:
This PR fixes an issue where we were not
splitting the nested message values for options
and indexing the symbols accordingly. For example,
the following field option:

```
message Foo {
  string module_id = 4 [(buf.validate.field).required = true];
}
```

We now correctly index the following symbols:

```
(buf.validate.field) -> definition is FieldRules
.required -> definition is the required field on FieldRules
```

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -651,13 +651,21 @@ func (f *file) messageToSymbolsHelper(msg ir.MessageValue, index int, parents []
 		// For both examples, we would want to create a separate symbol for each referenceable
 		// part of each path, e.g. (option), .message, .field_a, etc.
 		//
-		// To do this, we recursively iterate through the message value, e.g.
-		//   (option) -> (option).message -> (option).message.field_a
+		// An option is represented as a [ir.MessageValue] that is accessed recursively, so in
+		// both examples above, we have:
 		//
-		// and we set a symbol for each along the way. Then we also check the path we have to
-		// make sure we have a symbol for the corresponding span, e.g. (option).message.field_b
-		// we need to make sure we have a symbol for (option) and .message, even though we
-		// recursed into the message value through the top-level message value.
+		//  (option) -> (option).message -> (option).message.field_a / field_a
+		//                               -> (option).message.field_b / field_b
+		//
+		// As we walk the message recursively, we set a symbol for each message/field along the
+		// way. Because we are accessing each message from the top-level message, we need to
+		// make sure that we capture a symbol for each corresponding path span along the way.
+		//
+		// In the example, in the second definition, (option) and .message for field_b has a
+		// separate span from (option) and .message for field_a, but when we walk the mesasge
+		// tree, we get the span for (option) and .mesage for the first field. So we check the
+		// symbols we've collected so far in parents and make sure we have captured a symbol for
+		// each path component.
 		for element := range seq.Values(field.Elements()) {
 			key := field.KeyASTs().At(element.ValueNodeIndex())
 			components := slices.Collect(key.AsPath().Components)
@@ -704,6 +712,7 @@ func (f *file) messageToSymbolsHelper(msg ir.MessageValue, index int, parents []
 					for _, parent := range parents {
 						if parent.span == component.Span() {
 							found = true
+							break
 						}
 					}
 					if !found {


private/buf/buflsp/file.go
```diff
@@ -622,16 +622,30 @@ func (f *file) importToSymbol(imp ir.Import) *symbol {
 	}
 }
 
-func (f *file) messageToSymbols(msg ir.MessageValue) []*symbol {
+func (f *file) messageToSymbols(msg ir.MessageValue, index int, parents []*symbol) []*symbol {
```
emcfarlane: Can we make the recursive args into helper method. Both `index`, and `parents` shouldn't be set by callers.

private/buf/buflsp/file.go
```diff
@@ -622,16 +622,30 @@ func (f *file) importToSymbol(imp ir.Import) *symbol {
...
+			//    field_b: 100
+			//  }]
+			//
+			// `field_a` is a path that only has one component, but index = 2, since we have
+			// (option).message.field_a. So, in this case, we use the last component.
+			if index > len(components)-1 {
+				span = components[len(components)-1].Span()
+			} else {
+				span = components[index].Span()
+			}
```
emcfarlane: I'm not sure if this logic is required. With:
```
		for element := range seq.Values(field.Elements()) {
			for elementKey := range seq.Values(element.Value().KeyASTs()) {
// ...
			}
		}
```
This seems to handle the `field_a` and `field_b` cases.

private/buf/buflsp/file.go
```diff
@@ -640,9 +654,52 @@ func (f *file) messageToSymbols(msg ir.MessageValue) []*symbol {
...
+			// We check back along the path to make sure that we have a symbol for each component.
+			// This can happen if we have an option structure such as the following:
+			// [
+			//    (option).message.field_a = 2
+			//    (option).message.field_b = 100
+			//  ]
+			//
+			// We need to ensure that (option) for (option).message.field_b has a symbol defined
+			// among the parent symbols.
+			if len(components) > 1 {
```
emcfarlane, line 705: This took a lot of time to reason through. My understanding now is that on this recursive iteration we get the following:
1. `(option)`
2. `(option).message`
3. `(option).message.field_a`
4. `(option).message.field_b`

So we always register the parents of the first, but fail on the second one for the partial types. I think this case can be handled by only checking on leaf nodes. And using the key components from:
```
astPath := elementKey.AsPath()
components := astPath.Components()
```

private/buf/buflsp/file.go
```diff
@@ -640,9 +654,52 @@ func (f *file) messageToSymbols(msg ir.MessageValue) []*symbol {
...
+				parentType := element.Value().Container()
```
emcfarlane, line 706: Is the initial parent not the current `msg` value?

private/buf/buflsp/file.go
doriable: Yeah, so we are able to simplify the entire thing, other than the case of:

```
(option).message = {
    field_a: 2
    field_b: 100
}]
```

Since that won't be captured by the path components... which is why we cannot simply iterate through the path components... in those cases, `field_a` and `field_b` are the leaf nodes, and so `(option)` and `.message` are never captured by the path components.

I can abstract the recursive portion of it into a separate function though.

private/buf/buflsp/file.go
doriable: Yeah, the initial type is, but we skip `component.IsLast()`, so that is basically the first parent we care about. This could be clearer...

private/buf/buflsp/file.go
doriable: I'm not sure I follow this, `field_a` and `field_b` are separate elements, and we get the key for each one and handle. Iterating through the keys of the element's containing value would be different...

private/buf/buflsp/file.go
emcfarlane: But we can iterate through the list and only when we hit a leaf, then do this check. I think I'm no understanding the need for the count and index?

emcfarlane: Maybe adding to the docs here on why this is necessary.

doriable: What would you consider as a leaf? If we walk the `ir.MessageValue`, e.g.:

```
for field := range msg.Fields() {
  for element := range seq.Values(field.Elements()) {
    if !element.AsMessage().IsZero() {
      // Not a "leaf", so we recurse into the message
      symbols = append(symbols, f.messageToSymbols(element.AsMessage())...)
    }
    // Handle the element
  }
}
```

You would hit the fields `field_a` and `field_b`, but because the paths are disconnected, when you iterate through the paths for `field_a` and `field_b`, you don't get `(options).message`... symbols would need to be added for those in some way...

So you basically need to add symbols along the way, and then back fill any for the other case:

```
[
  (option).message.field_a = 1,
  (option).message.field_b = 2,
]
```

doriable:
> Maybe adding to the docs here on why this is necessary.

Yeah, I'll clean up the docs (and maybe some of the variable names) for this, the options stuff is super finicky, and I definitely want to minimise the confusion on how it works when we look back on this ^^"

private/buf/buflsp/file.go
```diff
@@ -622,16 +622,36 @@ func (f *file) importToSymbol(imp ir.Import) *symbol {
...
 	var symbols []*symbol
 	for field := range msg.Fields() {
 		if field.ValueAST().IsZero() {
 			continue
 		}
 		for element := range seq.Values(field.Elements()) {
-			span := element.Value().KeyASTs().At(element.ValueNodeIndex()).Span()
-			elem := &symbol{
-				// NOTE: no [ir.Symbol] for option elements
+			key := element.Value().KeyASTs().At(element.ValueNodeIndex())
```
```suggestion
			key := field.KeyASTs().At(element.ValueNodeIndex())
```
The `element.Value()` I think is just the `field` here?

private/buf/buflsp/file.go
```diff
@@ -640,9 +680,47 @@ func (f *file) messageToSymbols(msg ir.MessageValue) []*symbol {
...
+			if len(components) > 1 {
+				parentType := element.Value().Container()
+				for _, component := range slices.Backward(components) {
+					if component.IsLast() {
+						continue
+					}
+					found := false
+					for _, parent := range parents {
+						if parent.span == component.Span() {
+							found = true
```
emcfarlane, line 714: nit: break

private/buf/buflsp/file.go
emcfarlane: The comment doesn't explain why the recursion doesn't match the representation in the AST.

doriable: I added some clarifications to the top level comment, lmk what you think! :D

PR #4146, opened 2025-11-04
Audit organization uniqueness changes

unmultimedio:


proto/buf/alpha/audit/v1alpha1/event.proto
```diff
@@ -92,6 +92,8 @@ enum EventType {
   EVENT_TYPE_ORGANIZATION_IDP_GROUP_UPDATED = 59;
   EVENT_TYPE_ORGANIZATION_IDP_GROUP_REMOVED = 22;
   EVENT_TYPE_ORGANIZATION_RESOURCE_VISIBILITY_CONTROL_CHANGED = 81;
+  EVENT_TYPE_ORGANIZATION_UNIQUENESS_POLICY_ENABLED = 82;
+  EVENT_TYPE_ORGANIZATION_UNIQUENESS_POLICY_DISABLED = 83;
```
cyinma, line 96: how about a change event with the enable/disable being in the payload? I can see we kind of have both practices here so not a strong opinion

cyinma (approved):

proto/buf/alpha/audit/v1alpha1/event.proto
unmultimedio: Also thought about that, but wanted to keep it consistent with the server events https://github.com/bufbuild/buf/blob/f5061cceef2f27672a35cc82fec60908d5f3a007/proto/buf/alpha/audit/v1alpha1/event.proto#L149-L150

PR #4147, opened 2025-11-04
LSP add completion for service methods

emcfarlane:
This adds completions to the LSP for service methods. The service keywords (`rpc`, `returns`) and its input and output types with a field modifier (`stream`) are all completed. Type completions are restricted to matching parenthesis to avoid parser ambiguities.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -908,7 +908,8 @@ func isTokenParen(tok token.Token) bool {
 func isTokenTypeDelimiter(tok token.Token) bool {
 	kind := tok.Kind()
 	return (kind == token.Unrecognized && tok.IsZero()) ||
-		(kind == token.Space && strings.IndexByte(tok.Text(), '\n') != -1)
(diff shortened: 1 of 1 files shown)


emcfarlane:
@stefanvanburen thanks, removed scalars. For completion on `(`, it'd need the completion trigger char `(` but we also currently ignore any invalid definitions of an `rpc` method for type completion. The AST will parse `rpc Something` as an invalid declaration, so the bounds check on input and output spans isn't available. The completion would then be required to parse the stream to figure out the span of this. This is complicated as the file may be in an invalid state (and parenthesis fuse tokens). So currently the completion is limited to what the AST deems a valid rpc declaration, which will always be something like `rpc Something()` or `rpc Something() returns ()`.  

Edit: Was wrong on the restriction of the AST. The current restriction is needing to solve the different capture groups (capture the definition, vs capture the type arg).

stefanvanburen:
> thanks, removed scalars.

sweet, let's also make sure to remove enums since those can't be used in RPC I/O 😄

I'm fine on punting on the completions in `()` for now.

doriable (approved):

PR #4149, opened 2025-11-05
Upgrade protocompile and use span to index referenceable symbols

doriable:
This upgrades protocompile and fixes the breaking changes from
the upgrade. It also changes the index for referenceable symbols
to use `report.Span` instead of the canonicalized name, since the
canonicalized name is local to the container of the type.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -54,6 +54,8 @@ func getCompletionItems(
 		return nil
 	}
 
+	// This grabs the contents of the file as the top-level [ast.DeclBody], see [ast.File].Decls()
+	// for reference.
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -53,7 +54,7 @@ func getCompletionItems(
 		return nil
 	}
 
-	declPath := getDeclForPosition(file.ir.AST().DeclBody, position)
+	declPath := getDeclForPosition(id.Wrap(file.ir.AST(), id.ID[ast.DeclBody](1)), position)
```
stefanvanburen, line 59: oof

doriable: Yeah ^^" that one is a little... I should document that.

stefanvanburen (approved):

PR #4158, opened 2025-11-10
LSP add completion for options

emcfarlane:
This adds completions for options. Options are completed with the type path declaration and the value for both option declarations and compaction options.

Screenshot 2025-11-10 at 10 13 46 PM
Screenshot 2025-11-10 at 10 13 21 PM
Screenshot 2025-11-10 at 10 13 06 PM
Screenshot 2025-11-10 at 10 12 55 PM
Screenshot 2025-11-10 at 10 17 28 PM

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -594,7 +594,6 @@ func completionItemsForOptions(
 	offset int,
 ) []protocol.CompletionItem {
 	file.lsp.logger.DebugContext(ctx, "completion: options", slog.Int("offset", offset))
-
 	optionSpan, optionSpanParts := parseOptionSpan(file, offset)
 	if optionSpan.IsZero() {
 		file.lsp.logger.DebugContext(
@@ -604,7 +603,6 @@ func completionItemsForOptions(
 		)
 		return nil
 	}
-
 	optionsTypeName, targetKind := defKindToOptionType(parentDef.Classify())
 	if targetKind == ir.OptionTargetInvalid {
 		file.lsp.logger.DebugContext(
@@ -614,7 +612,6 @@ func completionItemsForOptions(
 		)
 		return nil
 	}
-
 	// Find the options message type in the workspace by looking through all types.
 	var optionsType ir.Type
 	for _, importedFile := range file.workspace.PathToFile() {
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -832,6 +1095,295 @@ func typeReferencesToCompletionItems(
...
+				}
+				item := protocol.CompletionItem{
+					Label: label,
+					Kind:  protocol.CompletionItemKindProperty,
+					TextEdit: &protocol.TextEdit{
+						Range:   reportSpanToProtocolRange(span),
+						NewText: label,
+					},
+					Deprecated: isDeprecated,
+					Detail:     "extension",
```
stefanvanburen: can we add useful documentation to these with `Documentation: symbol.FormatDocs()`? (I realize these aren't symbols, but maybe there's a way to transform them?)

(similar comment to the other `protocol.CompletionItem`s here.)

doriable: Perhaps it makes sense to do this in a general way, I think it's reasonable to log this as a follow-up.

doriable:
Did some testing in VS Code, it looks good generally :) I get different behaviours from @stefanvanburen (who appears to be testing on nvim?):

- The open and closing of parens does not seem to be an issue on VS Code, I wonder if this is because the client handles open/close parens for the user (we have [auto-closing pairs](https://github.com/bufbuild/vscode-buf/blob/main/protobuf-language-configuration.json) configured in the exte
- However, I don't get the list of suggestions when typing `.` 
Screenshot 2025-11-10 at 18 16 07

I think overall, this works pretty well, and I think we're mostly good to merge. It might be good for us (@emcfarlane, @stefanvanburen) to chat a little about cleaning up the experience for a specific set of clients (and codifying that somewhere for continuing to develop features for the LSP).

private/buf/buflsp/completion.go
emcfarlane: Added in [db78491](https://github.com/bufbuild/buf/pull/4158/commits/db78491a79de0a145118f21e07aa010f1808d38a)

private/buf/buflsp/symbol.go
```diff
@@ -462,6 +411,64 @@ func commentToMarkdown(comment string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(comment, "/*"), "*/")
 }
 
+// irMemberDoc returns the documentation for a message field, enum value or extension field.
```
doriable, line 417: Super tiny nit, this is only for the number:

```suggestion
// irMemberDoc returns the documentation for a message field, enum value or extension field number.
```

PR #4161, opened 2025-11-11
Resolve definition and type definition separately for options

doriable:
Previously, we were returning the same thing for `textDocument/definition`
and `textDocument/typeDefinition`, which is reasonable for most things in
proto. However, this is not the case with options, the behaviour we want is
for an example option definition, `(option).message.field_a` would be:

  | Definition | Type Definition
-- | -- | --
(option) | field option in extend declaration | message option { … }
.message | field message in the message option | message Message { … }
field_a | field field_a in message Message | This an int, so it should not go anywhere.

This PR also fixes symbol indexing for extensions, they should be considered
referenceable symbols.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -768,13 +768,7 @@ func (f *file) resolveASTDefinition(def ast.DeclDef, defName ir.FullName) *symbo
 		}
 		file = f
 	}
-	sym, ok := file.referenceableSymbols[defName]
-	if !ok {
-		// This could happen in the case where we are in the cache for example, and we do not
-		// have access to a buildable workspace.
-		return nil
-	}
-	return sym
+	return file.referenceableSymbols[defName]
 }
 
 // SymbolAt finds a symbol in this file at the given cursor position, if one exists.


private/buf/buflsp/file.go
```diff
@@ -315,47 +316,56 @@ func (f *file) IndexSymbols(ctx context.Context) {
...
+				referenceable, ok := def.kind.(*referenceable)
+				if !ok {
+					// This shouldn't happen, logging a warning
+					f.lsp.logger.Warn(
+						"found non-referenceable symbol in index",
+						slog.String("file", f.uri.Filename()),
+						slog.Any("symbol", def),
+					)
+				}
+				referenceable.references = append(referenceable.references, sym)
```
emcfarlane: This will panic on `referenceable` with a null pointer deref when `ok` is `false`.

private/buf/buflsp/file.go
```diff
@@ -315,47 +316,56 @@ func (f *file) IndexSymbols(ctx context.Context) {
...
-		}
-		file, ok := f.workspace.PathToFile()[ref.def.Span().Path()]
-		if !ok {
-			// Check current file
-			if ref.def.Span().Path() != f.objectInfo.Path() {
-				// This can happen if this references a predeclared type or if the file we are
-				// checking has not indexed imports.
+		switch sym.kind.(type) {
+		case *reference:
+			ref, _ := sym.kind.(*reference)
```
```suggestion
		switch kind := sym.kind.(type); kind {
		case *reference:
```
Use the type switch to avoid type casting again.

private/buf/buflsp/file.go
```diff
@@ -364,20 +374,30 @@ func (f *file) IndexSymbols(ctx context.Context) {
 			continue // ignore self
 		}
 		for _, sym := range file.referenceSymbols {
-			ref, ok := sym.kind.(*reference)
-			if !ok {
+			var fullName ir.FullName
+			switch sym.kind.(type) {
+			case *reference:
+				ref, _ := sym.kind.(*reference)
```
emcfarlane: Same here, use type switch arg.

emcfarlane (approved):
Tested and behaviour looks good, comments around type cast behaviour.

private/buf/buflsp/file.go
```diff
@@ -736,6 +753,29 @@ func (f *file) messageToSymbolsHelper(msg ir.MessageValue, index int, parents []
...
+		}
+		file = f
+	}
+	sym, ok := file.referenceableSymbols[defName]
+	if !ok {
+		// This could happen in the case where we are in the cache for example, and we do not
+		// have access to a buildable workspace.
+		return nil
+	}
+	return sym
```
```suggestion
	return file.referenceableSymbols[defName]
```
Not sure it can be missing for a valid name, but can just return the map value directly.

private/buf/buflsp/file.go
doriable: Ah yeah, this should be wrapped in an `else`, fixed!

PR #4163, opened 2025-11-11
Remove prefixes for field type ASTs

doriable:
This PR fixes an issue where we want to handle the field
type without prefixes.


emcfarlane (approved):
Mind expanding on what the issue is in the PR comment?

PR #4164, opened 2025-11-12
Check for newlines in parsing docs comments

doriable:
This PR checks if a space token contains in a newline
when parsing the token stream for leading comments/docs.

This is to prevent picking up comments that are separated
from a declaration by some amount of whitespace
containing a newline, for example:

```
// This is just a random note

// Message Foo...
message Foo { ... }
```

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -357,8 +357,16 @@ func (s *symbol) getDocsFromComments() string {
 		case token.Comment:
 			comments = append(comments, commentToMarkdown(t.Text()))
 		}
-		if !cursor.PeekPrevSkippable().Kind().IsSkippable() {
+		prev := cursor.PeekPrevSkippable()
+		if !prev.Kind().IsSkippable() {
 			break
+		}
+		if prev.Kind() == token.Space {
+			// Check if the whitespace contains a newline. If so, then we break. This is to prevent
+			// picking up comments that are not contiguous to the declaration.
+			if strings.Contains(prev.Text(), "\n") {
+				break
+			}
 		}
 		t = cursor.PrevSkippable()
 	}


private/buf/buflsp/symbol.go
```diff
@@ -357,9 +357,17 @@ func (s *symbol) getDocsFromComments() string {
 		case token.Comment:
 			comments = append(comments, commentToMarkdown(t.Text()))
 		}
-		if !cursor.PeekPrevSkippable().Kind().IsSkippable() {
+		prev := cursor.PeekPrevSkippable()
+		if !prev.Kind().IsSkippable() {
```
emcfarlane, line 361: If we are just collecting comments can this be simplified to?
```
for t := cursor.PrevSkippable(); t.Kind() == token.Comment; t = cursor.PrevSkippable() {
    comments = append(comments, commentToMarkdown(t.Text())
}
```
Does it need to go through non-comment tokens first?

emcfarlane (approved):

private/buf/buflsp/symbol.go
doriable: Hmm, this is a good point -- so when I originally wrote this, I had considered adding the `token.Space` as well, and I think I'm still considering it (I'm looking at a different issue that is a follow-up to this, where different LSP clients handle whitespace differently). Basically I'm considering whitespace tokens that may not be exclusively newlines for separating comments. In this case, I think I might merge this as-is first, so I can play around with the whitespace token for that issue, and pull in this simplification there as-needed? ^^"

emcfarlane: You may be able to avoid building the comment `[]string` list too, and just use the token start and end offset to capture the full span of the comment.

doriable: That might not work because the following:

```
// A foo is a...
//
// Foo is used for...
message Foo {
```

Would provide 3 comment tokens, so we need to collect all contiguous comment tokens, right?

emcfarlane:
```
var start, end token.Token
for t := cursor.PrevSkippable(); t.Kind() == token.Comment; t = cursor.PrevSkippable() {
    if end.IsZero() { end = t }
    start = t
}
var comment string
if !start.IsZero() {
    commentSpan := start.Span()
    commentSpan.End = end.Span().End
    comment = commentToMarkdown(commentSpan.Text())
}
```
😬

PR #4165, opened 2025-11-12
Line up comments due to different LSP client behaviours

doriable:
There is an issue where different LSP clients handle leading
whitespaces in rendered documentation differently. We don't
trim leading whitespace unilaterally because Commonmark
indented code blocks depend on them. Instead we will apply
a heuristic where if all lines containing one or more non-space
characters all begin with a leading space character, we trim it.

This PR also applies a nice simplification for collecting the comment
tokens from https://github.com/bufbuild/buf/pull/4164#discussion_r2519520004

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -462,7 +462,7 @@ func lineUpComments(comments []string) []string {
 			return !unicode.IsSpace(r)
 		}) {
 			// We are only checking for " " and do not count nbsp's.
-			if strings.Index(comment, " ") != 0 {
+			if !strings.HasPrefix(comment, " ") {
 				return comments
 			}
 			linedUp[i] = strings.TrimPrefix(comment, " ")


private/buf/buflsp/symbol.go
```diff
@@ -450,6 +435,44 @@ func commentToMarkdown(comment string) string {
 	return strings.TrimSuffix(strings.TrimPrefix(comment, "/*"), "*/")
 }
 
+// lineUpComments is a helper function for lining up the comments for docs, since some users
+// may start their comments with spaces, e.g.
+//
+//	// Foo is a ...
+//	//
+//	// Foo is used for ...
+//	message Foo { ... }
```
stefanvanburen, line 444: can you check what VS Code does for comments with multiple leading `" "` characters? e.g.:

```proto
//  These comments have two characters of leading whitespace.
//  Foo is a ...
//
//  Foo is used for ...
message Foo { ... }
```

Basically, I'm wondering if we ought to repeat this until all of the leading whitespace is "evened-out", skipping only-whitespace lines.

private/buf/buflsp/symbol.go
```diff
@@ -450,6 +435,44 @@ func commentToMarkdown(comment string) string {
...
+// If all lines containing one or more non-space characters all start with a single space
+// character, we trim the space character for each of these lines. Otherwise, do nothing.
+func lineUpComments(comments []string) []string {
+	linedUp := make([]string, len(comments))
+	for i, comment := range comments {
+		if strings.ContainsFunc(comment, func(r rune) bool {
+			return !unicode.IsSpace(r)
+		}) {
+			// We are only checking for " " and do not count nbsp's.
+			if strings.Index(comment, " ") != 0 {
```
stefanvanburen: nit: `!strings.HasPrefix(comment, " ")` may be clearer, but this is fine.

stefanvanburen (approved):
looks basically right to me, would just want to make sure we do the right thing for multiple matching leading spaces?

private/buf/buflsp/symbol.go
doriable: Ah yeah, no, that is better, I originally had some `strings.IndexAny` logic to account for nbsp's, and I decided against it, let me update that ^^"

private/buf/buflsp/symbol.go
doriable: I guess what you're suggesting is recursing (until "no changes"), and in theory, any indented code blocks should account for the extra spaces? Hmm... I wonder if that's covering for too niche of an edge case... since this heuristic is already a little bit shaky, and I feel like the most common cases are either a single space or no space. And then less commonly, tabs (which we don't account for) and multiple spaces (e.g. 4 spaces) in lieu of tabs, which recursing would cover...

I am of the opinion that we either account for tabs as well if we're going to recurse, or we simply forego both of those cases congruently, and I'm leaning towards the latter... ^^"

stefanvanburen: that's fine, we can leave it out for now IMO. maybe this will come up in practice and we can refine the heuristic then?

stefanvanburen (approved):

PR #4168, opened 2025-11-14
Upgrade deps

doriable:
This PR upgrades our deps in preparation for release,
including pulling in the latest version of protocompile,
which introduces some API changes.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -375,12 +375,12 @@ func (f *file) IndexSymbols(ctx context.Context) {
 			var fullName ir.FullName
 			switch kind := sym.kind.(type) {
 			case *reference:
-				if f.localPathToPath(kind.def.Span().Path()) != f.objectInfo.Path() {
+				if kind.def.Span().Path() != f.objectInfo.LocalPath() {
 					continue
 				}
 				fullName = kind.fullName
 			case *option:
-				if f.localPathToPath(kind.def.Span().Path()) != f.objectInfo.Path() {
+				if kind.def.Span().Path() != f.objectInfo.LocalPath() {
 					continue
 				}
 				fullName = kind.defFullName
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/file.go
```diff
@@ -771,6 +771,20 @@ func (f *file) resolveASTDefinition(def ast.DeclDef, defName ir.FullName) *symbo
 	return file.referenceableSymbols[defName]
 }
 
+// localPathToPath is a helper function for converting a local (on-disk) path to the importable
+// path, if available. If not, this will return an empty string. If the workspace is nil,
+// then we also return an empty string.
+func (f *file) localPathToPath(localPath string) string {
+	if f.workspace == nil {
+		return ""
```
emcfarlane: Should default to the path if it's not in a workspace.
```suggestion
		return localPath
```

private/buf/buflsp/file.go
```diff
@@ -771,6 +771,20 @@ func (f *file) resolveASTDefinition(def ast.DeclDef, defName ir.FullName) *symbo
...
+	}
+	fileInfo, ok := f.workspace.fileNameToFileInfo[localPath]
+	if !ok {
+		return ""
```
emcfarlane: Same here. Although this should not happen, it''s an error if so.

private/buf/buflsp/file.go
```diff
@@ -375,12 +375,12 @@ func (f *file) IndexSymbols(ctx context.Context) {
 			var fullName ir.FullName
 			switch kind := sym.kind.(type) {
 			case *reference:
-				if kind.def.Span().Path() != f.objectInfo.Path() {
+				if f.localPathToPath(kind.def.Span().Path()) != f.objectInfo.Path() {
```
emcfarlane: What about using `LocalPath` here? `kind.def.Span().Path() != f.objectInfo.LocalPath()`

private/buf/buflsp/file.go
```diff
@@ -757,7 +757,7 @@ func (f *file) messageToSymbolsHelper(msg ir.MessageValue, index int, parents []
 // resolveASTDefinition is a helper for resolving the [ast.DeclDef] to the *[symbol], if
 // there is a matching indexed *[symbol].
 func (f *file) resolveASTDefinition(def ast.DeclDef, defName ir.FullName) *symbol {
-	file, ok := f.workspace.PathToFile()[def.Span().Path()]
+	file, ok := f.workspace.PathToFile()[f.localPathToPath(def.Span().Path())]
```
emcfarlane: Maybe a method on workspace would be cleaner then the maps we expose here:
```
f.workspace.LocalPathToFile(def.Span().Path())
```

private/buf/buflsp/file.go
doriable: Hmm... I guess that is... fine? It's a weird thing because, if there is no workspace, then returning anything is somewhat meaningless. But I guess your reasoning is to deal with empty string situations/checks?

private/buf/buflsp/file.go
doriable: Hmm, maybe, although we need the map for setting up the opener o_o

private/buf/buflsp/file.go
emcfarlane: Or if there's no workspace the path is just the file name? I think that might be more accurate.

doriable: Actually, with some of the other changes, there would be a single call-site for this, going to just collapse it into the AST decl resolution logic, so it's reasonable.

emcfarlane (approved):

PR #4169, opened 2025-11-14
Disable format on files with errors

emcfarlane:
This disables format on files that contain an error. Replicates the check in the [LSP](https://github.com/bufbuild/buf/blob/810f1dd307b1724d401287bdd0259a06d998c51d/private/buf/buflsp/server.go#L277-L283). With a file given an unsupported edition this will now fail avoiding stripping/corrupting the unknown syntax. 

```
$ buf format
Failure: local.proto:1:11: edition "2024" not yet fully supported; latest supported edition "2023"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Disable format on unknown or invalid syntax.
 
 ## [v1.60.0] - 2025-11-14
 


emcfarlane:
@doriable if this change looks fine. Can go through and remove/correct the failing testcases and add a changelog.

doriable:
Yeah, this looks like a reasonable fix for the current formatter, thank you!

Issue #4173, opened 2025-11-17
Unable to open browser when logging into registry in WSL2


F21:
### GitHub repository with your minimal reproducible example (do not leave this field blank or fill out this field with "github.com/bufbuild/buf" or we will automatically close your issue, see the instructions above!)

None

### Commands

```shell
buf registry login
```

### Output

```shell
/usr/bin/xdg-open: 882: x-www-browser: not found
/usr/bin/xdg-open: 882: firefox: not found
/usr/bin/xdg-open: 882: iceweasel: not found
/usr/bin/xdg-open: 882: seamonkey: not found
/usr/bin/xdg-open: 882: mozilla: not found
/usr/bin/xdg-open: 882: epiphany: not found
/usr/bin/xdg-open: 882: konqueror: not found
/usr/bin/xdg-open: 882: chromium: not found
/usr/bin/xdg-open: 882: chromium-browser: not found
/usr/bin/xdg-open: 882: google-chrome: not found
/usr/bin/xdg-open: 882: www-browser: not found
/usr/bin/xdg-open: 882: links2: not found
/usr/bin/xdg-open: 882: elinks: not found
/usr/bin/xdg-open: 882: links: not found
/usr/bin/xdg-open: 882: lynx: not found
/usr/bin/xdg-open: 882: w3m: not found
xdg-open: no method available for opening 'https://buf.build/settings/user/device?code=REDACTED'
Failure: unable to complete authorize device grant: failed to open browser: exit status 3
```

### Expected Output

The browser should open and login should be successful.

### Anything else?

I am using Ubuntu 24.04 in WSL2 on Windows 11.

emcfarlane:
As a workaround you can use the the following workflow which will prompt for a token:
```
buf registry login --prompt
```
For the issue: I'm unable to test this locally, but you might be able to set the BROWSER environment var:
```
export BROWSER='/mnt/c/Windows/explorer.exe'
```

F21:
I tried setting `BROWSER` to `/mnt/c/Windows/explorer.exe` and `/mnt/c/Program Files/Google/Chrome/Application/chrome.exe`, but both did not work and return this error:

```
/usr/bin/xdg-open: 882: /mnt/c/Program: not found
/usr/bin/xdg-open: 882: /mnt/c/Program: not found
xdg-open: no method available for opening 'https://buf.build/settings/user/device?code=REDACTED'
Failure: unable to complete authorize device grant: failed to open browser: exit status 3
```

[...]

PR #4178, opened 2025-11-20
Switch to github.com/cli/browser

emcfarlane:
Package https://github.com/cli/browser is better supported.

Closes https://github.com/bufbuild/buf/issues/4173


emcfarlane:
@doriable would you have any way to check the WSL side of things?

doriable:
> @doriable would you have any way to check the WSL side of things?

Yep! I can validate on WSL and either leave a comment and/or stamp :)

PR #4186, opened 2025-11-24
Fix regressions for well-known types in LSP

doriable:
This PR fixes a regression to LSP features with WKTs
that was introduced when a change to protocompile's
API requires us to check for local path.

`workspace.fileNameToFileInfos` does not include WKTs
in module file infos, so we need to check `workspace.fileInfos(ctx)`
instead. This PR also fixes an issue with `workspace.fileInfos(ctx)`
iterator, where the use of `storage.ReadBucket.Walk` did not
break when `yield` returned `false`, so we now return an
`io.EOF` to break the walk.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -765,28 +765,11 @@ func (f *file) resolveASTDefinition(ctx context.Context, def ast.DeclDef, defNam
 	if f.workspace == nil {
 		return nil
 	}
-	for objectInfo := range f.workspace.fileInfos(ctx) {
-		if objectInfo.LocalPath() == def.Span().Path() {
-			file, ok := f.workspace.PathToFile()[objectInfo.Path()]
-			if !ok {
-				// This shouldn't happen, if it does, we log an error and optimistically continue
-				// to walk file infos.
-				f.lsp.logger.Error(
-					"found local path that does not resolve to a file in the workspace",
-					slog.String("file", f.uri.Filename()),
-					slog.String("local path", def.Span().Path()),
-				)
-			}
+	for _, file := range f.workspace.PathToFile() {
+		if file.file.Path() == def.Span().Path() {
 			return file.referenceableSymbols[defName]
 		}
 	}
-	// Unable to resolve an importable path for the file in our workspace, log a debug
-	// statement and return no definition.
-	f.lsp.logger.Debug(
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/file.go
```diff
@@ -756,36 +756,38 @@ func (f *file) messageToSymbolsHelper(msg ir.MessageValue, index int, parents []
...
+				f.lsp.logger.Error(
+					"found local path that does not resolve to a file in the workspace",
+					slog.String("file", f.uri.Filename()),
+					slog.String("local path", def.Span().Path()),
+				)
+			}
+			return file.referenceableSymbols[defName]
 		}
-		file = f
 	}
```
```suggestion
	for _, file := range f.workspace.PathToFile() {
		if file.file.Path() != def.Span().Path() {
			continue
		}
		return file.referenceableSymbols[defName]
	}
```

Or you may still be able to do the lookup but:
```
	fileInfo, ok := f.workspace.fileNameToFileInfo[def.Span().Path()]
	if !ok {
		// Handle WKT
		fileInfo, _ := w.lsp.wktBucket.Stat(ctx, def.Span().Path())
	}
```

private/buf/buflsp/workspace.go
```diff
@@ -296,11 +298,11 @@ func (w *workspace) fileInfos(ctx context.Context) iter.Seq[storage.ObjectInfo]
 		if err := w.lsp.wktBucket.Walk(ctx, "", func(objectInfo storage.ObjectInfo) error {
 			if _, ok := seen[objectInfo.Path()]; !ok {
 				if !yield(wktObjectInfo{objectInfo}) {
-					return nil
+					return io.EOF
```
emcfarlane, line 301: Nice catch!

private/buf/buflsp/file.go
doriable: The only issue with the alternative is that we can't stat the local path from `def.Span().Path()`, since the bucket abstraction expects a normalized and validated path. I can adjust to iterate through the map instead though.

EDIT: also we need to always be comparing to `LocalPath()` ^^"

emcfarlane: `file.file.Path()` is the local path. Logging it does a lot of lookups, we maybe should just make a mapping here. Also looks like sometimes `def.Span().Path()` is empty.

emcfarlane (approved):

PR #4189, opened 2025-11-25
Check for existing modules before attempting to create new ones

### LINKED ISSUE #4157: Failure: failed_precondition: feature "user-owned repository" is not enabled on the server

https://github.com/nvx/subspace-relay/actions/runs/19225434662/job/54951827780

### Commands

```shell
My CI pipeline to push to buf is now randomly failing with `Failure: failed_precondition: feature "user-owned repository" is not enabled on the server`

There appears to be no documentation around what this error from the buf registry means or how to resolve it.

The repo is public with the failing action run: https://github.com/nvx/subspace-relay/actions/runs/19225434662/job/54951827780

An earlier run where it succeeded: https://github.com/nvx/subspace-relay/actions/runs/18626992441/job/53106389344
```

### Output

```shell
Failure: failed_precondition: feature "user-owned repository" is not enabled on the server
```

### Expected Output

buf.build/nvxx/subspace-relay:<hash>

### Anything else?

_No response_

doriable:
This PR makes a fix to check modules _first_, before attempting
to create the modules when the `--create` flag is set for `buf push`.

This is because if a module already exists, we shouldn't bother to
attempt to create it, and we should allow users to continue pushing
to it.

This also allows users that have pre-existing user-owned modules
on registries that now disallow creating _new_ user-owned modules
to continue to push to their modules.

Fixes #4157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,7 @@
 - Fix regression in LSP functionality for well-known types.
 - Fix browser open for `buf registry login` in WSL2.
 - Fix panic in LSP for EOF lookups.
-- Allow users to push to pre-existing user-owned modules when the `--create` flag
-  is set for `buf push` on registries that disallow user-owned modules.
+- Fix `--create` flag for `buf push` to avoid errors on already existing modules if create is disallowed.
 
 ## [v1.60.0] - 2025-11-14
 


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - Fix regression in LSP functionality for well-known types.
 - Fix browser open for `buf registry login` in WSL2.
 - Fix panic in LSP for EOF lookups.
+- Allow users to push to pre-existing user-owned modules when the `--create` flag
+  is set for `buf push` on registries that disallow user-owned modules.
```
emcfarlane:
```suggestion
- Fix `--create` flag for `buf push` to avoid errors on already existing modules if create is disallowed.
```

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
```diff
@@ -371,48 +371,51 @@ func (a *uploader) createContentModuleIfNotExist(
...
-		),
-	)
-	if err != nil {
-		if connect.CodeOf(err) == connect.CodeAlreadyExists {
-			// If a module already existed, then we check validate its contents.
-			modules, err := a.validateContentModulesExist(ctx, primaryRegistry, []bufmodule.Module{contentModule})
+				),
+			)
 			if err != nil {
 				return nil, err
```
emcfarlane, line 404: Should we wrap the error here to inform the user that it failed on `--create`?

emcfarlane (approved):

private/bufpkg/bufmodule/bufmoduleapi/uploader.go
doriable: Hmmm... that's a good question... so my only concern is that the uploader is "generic"... and while this command is really the only use-case for it... I would like to avoid it if possible. This is already the error that the API returns, I'm fine with respecting it? But I don't know...

PR #4201, opened 2025-12-03
YAML error improvement

bufdev:
Quick and dirty.

Before:

```
Failure: decode buf.yaml: invalid as version v2: could not unmarshal as YAML: yaml: unmarshal errors:
  line 14: field foo not found in type bufconfig.externalBufYAMLFileV2
  line 15: field baz not found in type bufconfig.externalBufYAMLFileV2
```

After:

```
Failure: decode buf.yaml: yaml: unmarshal errors:
  line 14: field foo not found
  line 15: field baz not found
```


emcfarlane (approved):

bufdev:
Error has to do with the fact that 1.25.4 is being used for some reason (GHA hasn't updated to 1.25.5 maybe? cache?) and our builds don't appear to be deterministic in some way w.r.t. `golangci-lint`? The windows failure looks to be flaky. @emcfarlane can you take a look?

emcfarlane:
Added `check-latest` so the lastest patch is fetched. This avoids using the local cache and golangci-lint failing on the older version issue: https://github.com/bufbuild/buf/pull/4202

Issue #4213, opened 2025-12-09
Support swift_prefix override


q42jaap:
### Feature

The file option  swift_prefix isn't available as a buf.gen.yaml managed override.

When I try, buf refuses to parse the config:
```
version: v2
inputs:
  - directory: api
managed:
  enabled: true
  override:
    - path: user/v1
      file_option: swift_prefix
      value: UserV1
```
Results in the error.
```
Failure: decode config file: unknown file_option: "swift_prefix"
```

`private/bufpkg/bufconfig/generate_managed_option.go` has a limited list of file options where `swift_prefix` is absent.

`swift_prefix` is used by  `buf.build/connectrpc/swift` and `buf.build/apple/swift`

Please add it to buf.gen.yaml config, now every file needs to have the `swift_prefix` repeated.

doriable:
Thank you for filing this issue! Currently, [managed mode](https://buf.build/docs/generate/managed-mode/) prioritises support for the built-in protoc plugins, however, I think this is a very reasonable request. We’re currently not able to prioritise this work immediately, but I’m happy to chat with the team and look into getting this implemented in the future.

PR #4215, opened 2025-12-10
Add swift_prefix to managed mode

bufdev:
Fixes https://github.com/bufbuild/buf/issues/4213.

This PR adds `swift_prefix` to managed mode, e.g.

```
version: v2
inputs:
  - directory: api
managed:
  enabled: true
  override:
    - path: user/v1
      file_option: swift_prefix
      value: UserV1
```

If `managed.enabled: true` is set, but no override is provided, the
default behavior for `swift_prefix` is to take the pascal case of the
package name and use a `_` delimited prefix, e.g. `foo.empty` -> `Foo_Empty_`.
This emulates the default behavior for `apple/swift` when `swift_prefix`
is not set.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Add `swift_prefix` to managed mode.
+- Add `textDocument/rename` and `textDocument/prepareRename` support for `buf lsp serve`.
 
 ## [v1.61.0] - 2025-11-25
 
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -588,10 +588,12 @@ func (f *file) irToSymbols(irSymbol ir.Symbol) ([]*symbol, []*symbol) {
 		resolved = append(resolved, method)
 
 		input, _ := irSymbol.AsMethod().Input()
+		// Method input must be a single message type.
+		inputAST := irSymbol.AsMethod().AST().AsMethod().Signature.Inputs().At(0)
 		inputSym := &symbol{
 			ir:   irSymbol,
 			file: f,
-			span: irSymbol.AsMethod().AST().AsMethod().Signature.Inputs().Span(),
+			span: inputAST.RemovePrefixes().Span(), // We always strip prefixes in case of streaming.
 			kind: &reference{
 				def:      input.AST(), // Only messages can be method inputs and outputs
 				fullName: input.FullName(),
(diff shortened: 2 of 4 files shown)


private/bufpkg/bufimage/bufimagemodify/override.go
```diff
@@ -422,3 +422,10 @@ func rubyPackageValue(imageFile bufimage.ImageFile) string {
 	}
 	return strings.Join(packageParts, "::")
 }
+
+// swiftPrefixValue returns the swift_prefix for the given ImageFile based on its
+// package declaration. If the image file doesn't have a package declaration, an
+// empty string is returned.
+func swiftPrefixValue(imageFile bufimage.ImageFile) string {
+	panic("TODO")
```
bufdev: This TODO

doriable: Pushed an update for this TODO, thanks for setting this PR up!

doriable:
Will open a PR for docs so that we have the docs ready to go when this feature is merged and released.

private/bufpkg/bufimage/bufimagemodify/override.go
```diff
@@ -422,3 +422,18 @@ func rubyPackageValue(imageFile bufimage.ImageFile) string {
...
 }
+
+// swiftPrefixValue returns the swift_prefix for the given ImageFile based on its
+// package declaration. If the image file doesn't have a package declaration, an
+// empty string is returned.
+func swiftPrefixValue(imageFile bufimage.ImageFile) string {
+	pkg := imageFile.FileDescriptorProto().GetPackage()
+	if pkg == "" {
+		return ""
+	}
```
bufdev, line 433: Just confirming you checked this is right (I'm not sure if there's anything special to verify or. Ot based on what protoc's behavior is, maybe not)

doriable: I actually had a question about this/wanted your thoughts on the behaviour.

I checked against apple/swift, and I think the following behaviour makes sense before and after this change:

- When the package is not set and `swift_prefix` is not set, there will be no prefix (same before and after)
- When `swift_prefix` is set (in the file before the change and by managed mode after the change), the  struct will get the configured prefix.

The only thing I'm unsure about is the default behaviour/what we want to set the default value for `swift_prefix` in managed mode to be, since `swift_prefix` is an optional file option for apple/swift. So when a file does not have `swift_prefix` at all, apple/swift will still generate code and will set the prefix of each struct to a `_` delimited version of the package in pascal case, e.g.

```
syntax = "proto3";

package weather.acme.v1;

message Foo {...}
```

```
struct Weather_Acme_V1_Foo: ...
```

So I wasn't sure if we wanted to set an opinionated version of this, e.g. `WeatherAcmeV1`, or if we should effectively treat `swift_prefix` as unset when `managed.enabled: true` but no override is given to `swift_prefix`. I'm starting to think we should err on the side of the latter, just pushed up an update, but interested to get your opinion on this.

emcfarlane (approved):

PR #4216, opened 2025-12-11
Implement `textDocument/rename` and `textDocument/prepareRename` for LSP

doriable:
This PR adds [renaming capabilities ](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_rename)to the Buf LSP, with
[prepare rename request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) support.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -310,7 +310,6 @@ func (s *symbol) GetSymbolInformation() protocol.SymbolInformation {
 // Rename returns the [protocol.WorkspaceEdit] for renaming the symbol.
 func (s *symbol) Rename(newName string) (*protocol.WorkspaceEdit, error) {
 	var edits protocol.WorkspaceEdit
-	edits.Changes = make(map[protocol.DocumentURI][]protocol.TextEdit)
 	switch s.kind.(type) {
 	case *referenceable:
 		changes, err := renameChangesForReferenceableSymbol(s, newName)


doriable:
For valid symbol renames:

Screenshot 2025-12-11 at 17 19 13

Clients will handle when `textDocument/prepareRename` returns an empty range:

Screenshot 2025-12-11 at 17 18 19

emcfarlane (approved):

private/buf/buflsp/symbol.go
```diff
@@ -307,6 +307,62 @@ func (s *symbol) GetSymbolInformation() protocol.SymbolInformation {
 	}
 }
 
+// Rename returns the [protocol.WorkspaceEdit] for renaming the symbol.
+func (s *symbol) Rename(newName string) (*protocol.WorkspaceEdit, error) {
+	var edits protocol.WorkspaceEdit
+	edits.Changes = make(map[protocol.DocumentURI][]protocol.TextEdit)
```
emcfarlane: Could be moved into the `default:` case of the switch.

doriable: I think this line basically just isn't needed, lol, since I refactored out the helper function, so I just removed it. ^^"

stefanvanburen:
my only question on this one: should we error when renaming a symbol to something that would conflict with another symbol in the scope?

e.g., if in gopls I attempt to [rename `lock` to `shutdown` here](https://github.com/bufbuild/buf/blob/b641342ffe6a2262bc76dc3b38389134f40d21e7/private/buf/buflsp/buflsp.go#L94-L96) (where the `shutdown` symbol already exists), it'll fail and log an error:

image

It seems like we'll happily rename to a conflicting name, and just get a lint error after:

image

doriable:
> my only question on this one: should we error when renaming a symbol to something that would conflict with another symbol in the scope?

Yeah, I think this is a good idea -- will put up a quick follow-up for this.

PR #4217, opened 2025-12-12
Add LSP request and response snippet

emcfarlane:
This adds a snippet for completion of `rpc` methods. It will insert a snippet with `rpc ${1:Name}(${1:Name}Request) returns (${1:Name}Response);` as well as the message types `message ${1:Name}Request` and `message ${1:Name}Response`. Additional text does not support snippet completions, we therefore replace the service definition to ensure a single update for this snippet.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -444,6 +444,7 @@ func completionItemsForDef(ctx context.Context, file *file, declPath []ast.DeclA
 						tokenSpan,
 						offset,
 					),
+					rpcRequestResponseSnippetCompletionItem(parentDef, tokenSpan),
 				)
 			}
 		} else if showReturnKeyword {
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -867,6 +868,50 @@ func keywordToCompletionItem(
...
+			Detail:           "Insert RPC method with Request and Response messages",
+			Documentation:    "Creates an RPC method declaration and corresponding Request/Response message types",
+			TextEdit: &protocol.TextEdit{
+				Range:   reportSpanToProtocolRange(tokenSpan),
+				NewText: sb.String(),
+			},
+			AdditionalTextEdits: []protocol.TextEdit{
+				{
+					Range:   reportSpanToProtocolRange(expandedSpan),
+					NewText: "", // Delete from cursor to end of service
```
emcfarlane, line 908: Delete is required for VSCode. If the `TextEdit.Range` spans the `tokenSpan` to the `serviceOffset` it rejects the completion item.

PR #4229, opened 2025-12-15
Block renaming a symbol to an existing symbol name

doriable:
This checks if the proposed rename conflicts with an existing
symbol name in the workspace.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -387,7 +387,7 @@ func checkRenameConflicts(target *symbol, newName string) error {
 		// We first check the current symbol's file.
 		if slices.ContainsFunc(target.file.symbols, containsFunc) {
 			return fmt.Errorf(
-				"Renaming %q to %q would conflict with existing symbol at %q:%d:%d",
+				"Renaming %q to %q would conflict with existing symbol at %s:%d:%d",
 				target.ir.FullName().Name(),
 				newName,
 				target.file.ir.Path(),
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/symbol.go
```diff
@@ -362,6 +371,46 @@ func renameChangesForReferenceableSymbol(s *symbol, newName string) (map[protoco
...
+		containsFunc := func(s *symbol) bool {
+			existing = s.span
+			return s.ir.FullName() == newFullName
+		}
+		// We check all files in the workspace for a conflict, since a package can span an arbitrary
+		// number of files.
+		// We first check the current symbol's file.
+		if slices.ContainsFunc(target.file.symbols, containsFunc) {
+			return fmt.Errorf(
+				"Renaming %q to %q would conflict with existing symbol at %q:%d:%d",
```
stefanvanburen: not sure if the initial string should be in quotes, maybe just a string here for the file?:

Image

```suggestion
				"Renaming %q to %q would conflict with existing symbol at %s:%d:%d",
```

private/buf/buflsp/symbol.go
```diff
@@ -362,6 +371,46 @@ func renameChangesForReferenceableSymbol(s *symbol, newName string) (map[protoco
...
+				newName,
+				target.file.ir.Path(),
+				existing.StartLoc().Line,
+				existing.StartLoc().Column,
+			)
+		}
+		for _, file := range target.file.workspace.PathToFile() {
+			if slices.ContainsFunc(file.symbols, containsFunc) {
+				return fmt.Errorf(
+					"Renaming %q to %q would conflict with existing symbol at %q:%d:%d",
```
stefanvanburen: Assuming you agree with the above

```suggestion
					"Renaming %q to %q would conflict with existing symbol at %s:%d:%d",
```

stefanvanburen (commented):
looks good to me, just linking back to [my comment](https://github.com/bufbuild/buf/pull/4216#issuecomment-3656293370) for context. One small nit on the error text.

private/buf/buflsp/symbol.go
doriable: Yeah, I was undecided whether to have the quotes or not ^^" I'm happy with no quotes, going to push up the change.

stefanvanburen (approved):

PR #4230, opened 2025-12-16
Reuse `source.Opener` and `ir.Session` to fix executor memory leaks in LSP

### LINKED ISSUE #4225: Buf 1.61 vscode plugin memory leak

no

### Commands

```shell
I'm using the buf vscode plugin at my dev container.
I'm running it on my private repo, so I can provide a sample repo.
But I can provide some details to address the problem.
```

### Output

```shell
No exception.
```

### Expected Output

No.

### Anything else?

# buf --version
Image

# buf memory

Image

Image

doriable:
This PR moves the `source.Opener` and `ir.Session` to
the top-level `lsp` structure and caches it for re-use when
executing queries with `protocompile`. It also adds eviction
for stale queries from the executor. This fixes the query
caching (which caused memory leaks) in the executor.

Fixes #4225

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -217,6 +217,14 @@ func (f *file) RefreshWorkspace(ctx context.Context) {
 //
 // This operation requires a workspace.
 func (f *file) RefreshIR(ctx context.Context) {
+	if f.objectInfo == nil {
+		// This can happen if the file that is opened is excluded from the workspace through
+		// includes/excludes in the build. In those cases, we ignore the file, the same way
+		// the rest of our tools would.
+		// In the future, we may want to rework this behavior in the LSP.
+		return
+	}
+
 	f.lsp.logger.Info(
 		"parsing IR for file",
 		slog.String("uri", string(f.uri)),


private/buf/buflsp/file.go
```diff
@@ -221,19 +222,31 @@ func (f *file) RefreshIR(ctx context.Context) {
...
+		seen = append(seen, path)
 		files = append(files, file)
 	}
-	opener := source.NewMap(openerMap)
+	// Remove paths that are no longer in the current workspace.
+	for key := range maps.Keys(openerMap) {
+		if !slices.Contains(seen, key) {
+			delete(openerMap, key)
+		}
+	}
```
emcfarlane: This looks like we can do a simple lookup in the `pathToFiles` map:
```
for path := range openerMap {
   if _, ok := pathToFile(path); !ok { delete(openerMap, path) }
}
```

private/buf/buflsp/buflsp.go
```diff
@@ -90,6 +94,8 @@ type lsp struct {
 	fileManager      *fileManager
 	workspaceManager *workspaceManager
 	queryExecutor    *incremental.Executor
+	opener           source.Map
+	irSession        *ir.Session
```
emcfarlane, line 98: nit: can this just be `&lsp.irSession`, so the field type is a non pointer `irSession ir.Session`.

emcfarlane (commented):
Does this still leak memory just not as much as we now re-use the files? What was keeping the previous `ir.FIle`s active?

doriable:
> Does this still leak memory just not as much as we now re-use the files? What was keeping the previous ir.FIles active?

Hmm yeah -- I'm going to make a change to this to cache the queries for each file on the file itself, and then if there are changes, we can evict the query key from the executor.

private/buf/buflsp/file.go
```diff
@@ -221,19 +222,29 @@ func (f *file) RefreshIR(ctx context.Context) {
...
+		// If there is no entry for the current path or if the file content has changed, we
+		// update the opener.
+		if current == nil || current.Text() != file.file.Text() {
+			openerMap[path] = file.file
+		}
 		files = append(files, file)
 	}
-	opener := source.NewMap(openerMap)
+	// Remove paths that are no longer in the current workspace.
+	for path := range maps.Keys(openerMap) {
```
emcfarlane: This does not need to be converted to an iterator.
```suggestion
	for path := range openerMap {
```

emcfarlane (approved):

PR #4231, opened 2025-12-16
Support multi-arch image manifests for image digest in `buf beta registry plugin push`

hampusohlsson:
Use ref.Context() to get the repository instead of manually parsing the reference name with strings.Cut(). The previous approach only worked for digest references (repo@sha256:...) but failed for tag references (repo:tag) when resolving OCI image indexes to platform-specific manifests.

Also add validation to return a clear error when no valid platform manifest is found in an image index.


doriable:
Thank you for submitting this, looks good. It will be available at the next release!

PR #4249, opened 2025-12-31
Respect TLS configuration in registry browser login

mfridman:
This PR updates `doBrowserLogin` to use `http://` when TLS is disabled instead of always defaulting to `https://`.

This ensures browser login works correctly in environments where TLS is not configured.

The intention is to match how we initialize clients elsewhere, e.g.,

https://github.com/bufbuild/buf/blob/e6b77ae450b2212a335203b9462004259f140f8a/private/buf/bufcli/connectclient_config.go#L74-L77

### DIFF
--- a/cmd/buf/internal/command/registry/registrylogin/registrylogin.go
+++ b/cmd/buf/internal/command/registry/registrylogin/registrylogin.go
@@ -28,6 +28,7 @@ import (
 	"github.com/bufbuild/buf/private/buf/bufapp"
 	"github.com/bufbuild/buf/private/buf/bufcli"
 	"github.com/bufbuild/buf/private/bufpkg/bufconnect"
+	"github.com/bufbuild/buf/private/bufpkg/buftransport"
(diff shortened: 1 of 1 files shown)


cmd/buf/internal/command/registry/registrylogin/registrylogin.go
```diff
@@ -284,6 +284,10 @@ func doBrowserLogin(
 	if err != nil {
 		return "", err
 	}
+	baseURL := buftransport.PrependHTTPS(remote)
+	if appConfig.TLS == nil {
+		baseURL = buftransport.PrependHTTP(remote)
+	}
```
stefanvanburen, line 290: drive-by, but: you want to `if` / `else` this, otherwise we'd double-prepend in the http case, right?

mfridman: We're not modifying the `remote` variable, only overriding the `baseURL` set above. And 99.99% of the time, the HTTPS case (the existing default) should be used.

stefanvanburen: ah, right, ignore me.

PR #4255, opened 2026-01-05
Add LSP tests for definition / type definition

stefanvanburen:
Follow up to #4254. I know this sort of test is somewhat contrived, but they're fairly cheap to write and maintain, and should give us the freedom to make larger changes knowing that we aren't breaking things under the covers.

### DIFF
--- a/private/buf/buflsp/definition_test.go
+++ b/private/buf/buflsp/definition_test.go
@@ -107,7 +107,7 @@ func TestDefinition(t *testing.T) {
 			line:               96, // Line with "Account account = 1;" in CreateAccountRequest
 			character:          2,  // On "Account" type
 			expectedDefURI:     testURI,
-			expectedDefLine:    7,  // message Account definition
+			expectedDefLine:    7, // message Account definition
 			expectedDefCharMin: 8,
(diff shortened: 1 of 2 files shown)


stefanvanburen:
FWIW, pausing on adding any other tests for now; I know we talked about specifically tests for completion and semantic tokens, but IMO we ought to have some basic coverage over [all of our advertised "providers"](https://github.com/bufbuild/buf/blob/db27d98b273761120a27bac9f2a5f870c7255af4/private/buf/buflsp/server.go#L146-L166), so we'd also want to add formatting, rename, references and symbols as well at some point.

private/buf/buflsp/testdata/definition/definition.proto
stefanvanburen, line 1: I figured it was better for the tests to all have their own separate "projects" in testdata; didn't want to have to update every test when we made changes to further test a single feature.

doriable: Yeah, I agree with this sentiment and it's consistent with the way we structure test data elsewhere in the code base.

doriable (approved):

PR #4256, opened 2026-01-05
Add LSP test for completions

emcfarlane:
This adds basic autocompletion tests to the LSP.

### DIFF
--- a/private/buf/buflsp/completion_test.go
+++ b/private/buf/buflsp/completion_test.go
@@ -55,19 +55,19 @@ func TestCompletion(t *testing.T) {
 		},
 		{
 			name:             "complete_builtin_service",
-			line:             14, // Line 15 in file (1-indexed) = line 14 in LSP (0-indexed) with "// complete_builtin_service"
-			character:        2,  // Indented position where "rpc" would be
+			line:             14,
+			character:        2, // Indented position where "rpc" would be
 			expectedContains: []string{"rpc", "option"},
 		},
 		{
 			name:             "complete_rpc_request_type",
-			line:             13, // Line 14 in file (1-indexed) = line 13 in LSP (0-indexed) with RPC definition
+			line:             13,
 			character:        uint32(len("  rpc GetUser(Get") - 1),
 			expectedContains: []string{"GetUserRequest", "GetUserResponse"},
 		},
 		{
 			name:             "complete_rpc_response_type",
-			line:             13, // Line 14 in file (1-indexed) = line 13 in LSP (0-indexed) with RPC definition
+			line:             13,
 			character:        uint32(len("  rpc GetUser(Get) returns (Get") - 1),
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion_test.go
```diff
@@ -0,0 +1,109 @@
...
+				return
+			}
+			require.NotNil(t, completionList, "expected completion list to be non-nil")
+			labels := make([]string, 0, len(completionList.Items))
+			for _, item := range completionList.Items {
+				labels = append(labels, item.Label)
+			}
+			for _, expected := range tt.expectedContains {
+				assert.Contains(t, labels, expected, "expected completion list to contain %q", expected)
+			}
```
stefanvanburen, line 103: I realize I may have done a similar thing in my tests, but would it make sense to do an `assert.ElementsMatch` (ignoring ordering) so we know if the full set of completions changes? or is that too noisy (i.e., too many completion suggestions)?

private/buf/buflsp/completion_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	clientJSONConn, testURI := setupLSPServer(t, testProtoPath)
+
+	tests := []struct {
+		name                string
+		line                uint32
+		character           uint32
+		expectedContains    []string
+		expectedNotContains []string
+		expectNoCompletions bool
```
stefanvanburen, line 42: nit: we don't use either of these — not sure if expectedNotContains is useful if we switch to using `assert.ElementsMatch` below, but can probably remove it, and it might be nice to have a couple of negative tests for places where we don't have completion suggestions (using `expectNoCompletions`).

stefanvanburen (approved):
I'm happy to iterate on these (and realize I didn't necessarily exhaust all of these in my own tests), but I think it'd also be good to have tests covering [syntax, packages and imports](https://github.com/bufbuild/buf/blob/74f276893fab49e90ea19da4932fe1060fd7aff0/private/buf/buflsp/completion.go#L100-L105), [options](https://github.com/bufbuild/buf/blob/74f276893fab49e90ea19da4932fe1060fd7aff0/private/buf/buflsp/completion.go#L253), and any of the [specialized cases we have based on the parentDef](https://github.com/bufbuild/buf/blob/74f276893fab49e90ea19da4932fe1060fd7aff0/private/buf/buflsp/completion.go#L368-L496). (We can follow-up on missing ones, though, especially if we're working through fixing some of this.)

private/buf/buflsp/completion_test.go
emcfarlane: I think theres too many.

private/buf/buflsp/completion_test.go
emcfarlane: We are missing a bunch here, was going to follow up with some fixes needed. Theres an issue with state which isn't captured in this test. I think we need to have an update, followed by a completion to show this issue.

PR #4257, opened 2026-01-05
Evict `queries.File` with `ReportError` set to both `true` and `false`

doriable:
This PR fixes buggy behaviour found in autocompletion
after the memory leak was merged (#4230) due to stale
state being returned from `queries.Executor`. This is because
we were only evicting `queries.File` where `ReportError` is set
to true, but `queries.IR` depends on `queries.File` with `ReportError`
set to false, so a stale version of the `File` was being parsed for IR.

This also updates protocompile to the latest version to pull in fixes
for zero values.

### DIFF
--- a/private/buf/buflsp/hover_test.go
+++ b/private/buf/buflsp/hover_test.go
@@ -34,17 +34,18 @@ func TestHover(t *testing.T) {
 	clientJSONConn, testURI := setupLSPServer(t, testProtoPath)
 
 	tests := []struct {
-		name             string
-		line             uint32
-		character        uint32
-		expectedContains string
-		expectNoHover    bool
+		name                string
+		line                uint32
+		character           uint32
+		expectedContains    string
+		expectedNotContains string
+		expectNoHover       bool
 	}{
 		{
 			name:             "hover_on_user_message",
-			line:             7, // Line with "message User {"
-			character:        8, // On the word "User"
-			expectedContains: "User represents a user in the system",
+			line:             7,               // Line with "message User {"
+			character:        8,               // On the word "User"
+			expectedContains: "system.\nThis", // Ensure newline between comment lines
 		},
 		{
 			name:             "hover_on_id_field",
@@ -77,6 +78,16 @@ func TestHover(t *testing.T) {
 			expectedContains: "GetUser retrieves a user by their ID",
 		},
 		{
+			name:      "hover_on_deprecated_option",
+			line:      37, // Line with "option deprecated = true;"
+			character: 11, // On "deprecated"
+			// We don't want the hover info to include the floating comment that is separated by newlines
+			// from the comment above the option.
+			// Ref: https://buf.build/protocolbuffers/wellknowntypes/file/main:google/protobuf/descriptor.proto#L946
+			expectedNotContains: "Buffers.", // From last line of the previous floating comment.
+			expectedContains:    "Is this method deprecated?",
+		},
+		{
 			name:             "hover_on_status_type_reference",
 			line:             15, // Line with "Status status = 3;"
 			character:        2,  // On "Status" type
@@ -84,7 +95,7 @@ func TestHover(t *testing.T) {
 		},
 		{
 			name:             "hover_on_user_type_reference",
-			line:             45, // Line with "User user = 1;"
+			line:             50, // Line with "User user = 1;"
 			character:        2,  // On "User" type
 			expectedContains: "User represents a user in the system",
 		},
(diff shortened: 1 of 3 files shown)


private/buf/buflsp/file.go
```diff
@@ -245,16 +243,22 @@ func (f *file) RefreshIR(ctx context.Context) {
...
-			}
-			file.fileQuery = queries.File{
-				Opener:      file.lsp.opener,
-				Path:        path,
-				ReportError: true, // [queries.AST] sets this to be true.
+			evictQueryKeys = append(evictQueryKeys, file.evictQueryKeys...)
+
+			// Set new queries on the file. We must evict [queries.File] with ReportError set to
+			// both true and false, since [queries.AST] depends on ReportError set to true, and
+			// [queries.IR] depends on ReportError set to false.
```
emcfarlane: This doesn't look correct, key requires the [opener, session and path](https://github.com/bufbuild/protocompile/blob/4d8d90b1c6b81a3d7cbd55df836c33b55d0c88c6/experimental/incremental/queries/ir.go#L48) not the report error.

doriable: We're evicting `queries.File` since `queries.IR` and `queries.AST` both depend on it https://github.com/bufbuild/protocompile/blob/4d8d90b1c6b81a3d7cbd55df836c33b55d0c88c6/experimental/incremental/queries/file.go#L46-L48

private/buf/buflsp/file.go
```diff
@@ -245,16 +243,22 @@ func (f *file) RefreshIR(ctx context.Context) {
...
+			file.evictQueryKeys = []any{
+				queries.File{
+					Opener:      file.lsp.opener,
```
emcfarlane: I think we need to make this opener based on the workspace. The path is not unique per invocation of the LSP. Maybe this could simply be the `workspace` implements `source.Opener`?

private/buf/buflsp/file.go
```diff
@@ -245,16 +243,22 @@ func (f *file) RefreshIR(ctx context.Context) {
 		// update the opener and set a new query.
 		if current == nil || current.Text() != file.file.Text() {
```
emcfarlane, line 242: `current == nil` is not possible? Also this is going through both the LSP map and the workspace map. Must only be based on the workspace.

private/buf/buflsp/file.go
emcfarlane: If we evict on the `queries.IR` does this not evict child nodes?

private/buf/buflsp/file.go
doriable: Yeah, we _could_, and then basically have the workspace manager manage the state of the files instead. In this case, I kinda took the approach of requesting the most recent state of files as late as possible to get updated IR information, but I suppose based on the top-level locking of the LSP moving it to workspace _should_ be fine...

private/buf/buflsp/file.go
doriable: It's possible when the IR is being requested for the very first time -- the LSP does not prepopulate the map. Slightly related to my response to the other comment.

private/buf/buflsp/file.go
doriable: No, we only [evict the things that call it](https://github.com/bufbuild/protocompile/blob/4d8d90b1c6b81a3d7cbd55df836c33b55d0c88c6/experimental/incremental/executor.go#L236-L238) -- since its dependencies may be the dependencies of other, unrelated queries, so they would be cached.

private/buf/buflsp/file.go
```diff
@@ -63,7 +63,7 @@ type file struct {
 
 	ir                   *ir.File
 	irQuery              queries.IR
-	fileQuery            queries.File
+	evictQueryKeys       []any
```
emcfarlane: Can we remove these in favour of methods, this state is derived from the file fields:
```go
func (f *file) queryIR() queries.IR {
	return queries.IR{
		Opener:  f.workspace,
		Path:    f.objectInfo.Path(),
		Session: f.lsp.irSession,
	}
}
func (f *file) queryFileKeys() []any {
	return []any{
		queries.File{
			Opener:      f.workspace,
			Path:        f.objectInfo.Path(),
			ReportError: true,
		}.Key(),
		queries.File{
			Opener:      f.workspace,
			Path:        f.objectInfo.Path(),
			ReportError: false,
		}.Key(),
	}
}
```

private/buf/buflsp/file.go
```diff
@@ -245,16 +243,22 @@ func (f *file) RefreshIR(ctx context.Context) {
...
-			// the [queries.File] query since this allows the executor to evict all dependent
-			// queries, e.g. AST and IR.
-			if file.fileQuery.Path != "" {
-				evictQueryKeys = append(evictQueryKeys, file.fileQuery.Key())
-			}
-			file.fileQuery = queries.File{
-				Opener:      file.lsp.opener,
-				Path:        path,
-				ReportError: true, // [queries.AST] sets this to be true.
+			evictQueryKeys = append(evictQueryKeys, file.evictQueryKeys...)
```
emcfarlane: I don't think we need to evict all keys here. Just the update from this file.

doriable: We still need to evict the keys here because if the keys don't get evicted, when the queries are executed, they will go straight to the cached file query results.

private/buf/buflsp/file.go
doriable: Yep, updated!

private/buf/buflsp/file.go
emcfarlane: I think we only need to evict the changes, which is the current file.

doriable: Ah yeah, this makes sense -- will adjust this to check for the non-nil current case.

private/buf/buflsp/file.go
emcfarlane: Okay, this effectively creates the cached view introduced here: https://github.com/bufbuild/buf/commit/9ac413fcab3c23157e7b391c43fd46df8aacb0e7
But we need to store the map so evict keys can be correct. So on each workspace switch, the map is mutated to look like the cached view. I don't think this has any side effects.

PR #4258, opened 2026-01-05
Fix LSP symbol resolve on missing rpc types

emcfarlane:
This fixes a panic on an invalid file where the `rpc` method has an empty request or response type.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -629,7 +629,7 @@ func (f *file) irToSymbols(irSymbol ir.Symbol) ([]*symbol, []*symbol) {
 
 		input, _ := irSymbol.AsMethod().Input()
 		// Method input must be a single message type.
-		if astInputs := irSymbol.AsMethod().AST().AsMethod().Signature.Inputs(); !astInputs.IsZero() {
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/file.go
```diff
@@ -629,31 +629,34 @@ func (f *file) irToSymbols(irSymbol ir.Symbol) ([]*symbol, []*symbol) {
...
-		outputAST := irSymbol.AsMethod().AST().AsMethod().Signature.Outputs().At(0)
-		outputSym := &symbol{
-			ir:   irSymbol,
-			file: f,
-			span: outputAST.RemovePrefixes().Span(), // We always strip prefixes in case of streaming.
-			kind: &reference{
-				def:      output.AST(), // Only messages can be method inputs and outputs
-				fullName: output.FullName(),
-			},
+		if astOutputs := irSymbol.AsMethod().AST().AsMethod().Signature.Outputs(); !astOutputs.IsZero() {
```
doriable: I think this should actually be a check on `astOutputs.Len() > 0`, since all AST accessors already have zero value checks on them (e.g. https://github.com/bufbuild/protocompile/blob/4d8d90b1c6b81a3d7cbd55df836c33b55d0c88c6/experimental/ast/type_generic.go#L147)

emcfarlane: Assert == 1.

PR #4259, opened 2026-01-06
Stop collecting floating comments in LSP hover

stefanvanburen:
Comments that are separated by multiple newlines ([e.g.](https://buf.build/protocolbuffers/wellknowntypes/file/main:google/protobuf/descriptor.proto#L946)) are "floating comments" that typically don't refer directly to the next field. We ought to not include them in the hover documentation.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -457,26 +457,22 @@ func (s *symbol) getDocsFromComments() string {
 	tok, _ := def.Context().Stream().Around(def.Span().Start)
 	cursor := token.NewCursorAt(tok)
 	var seenNewline bool
-	var addNewline bool
 	for {
 		t := cursor.PrevSkippable()
 		if t.Kind() == token.Comment {
 			text := commentToMarkdown(t.Text())
-			if addNewline {
+			if seenNewline {
 				text += "\n"
-				addNewline = false
 			}
 			seenNewline = false
 			comments = append(comments, text)
 		} else if t.Kind() == token.Space {
-			isNewline := strings.Contains(t.Text(), "\n")
-			if isNewline && seenNewline {
-				break
+			if strings.Contains(t.Text(), "\n") {
+				if seenNewline {
+					break
+				}
+				seenNewline = true
 			}
-			if isNewline {
-				addNewline = true
-			}
-			seenNewline = seenNewline || isNewline
 		} else {
 			break
 		}


private/buf/buflsp/symbol.go
```diff
@@ -459,17 +459,20 @@ func (s *symbol) getDocsFromComments() string {
...
 				addNewline = false
 			}
 			comments = append(comments, commentToMarkdown(text))
-		case token.Space:
-			// If the space token only contains spaces (e.g. code indentation), then we drop it.
-			// If the space token ends in a new-line, we append it to the comment above for formatting.
+		} else if t.Kind() == token.Space {
+			// If the space token contains multiple newlines (blank line), stop collecting
+			// comments since they are separated and not part of the same documentation block.
+			if strings.Count(t.Text(), "\n") > 1 {
```
emcfarlane: I thought the newline was now moved into the comment? This doesn''t look like it works:

Image

I thought we want the comment to now just be `This is now the comment`.

stefanvanburen: oof, good catch. Fixing!

stefanvanburen: ok, should be fixed now: 
image

stefanvanburen: https://github.com/bufbuild/buf/pull/4259/commits/dd79cbf2cb207c7e85fa560fa7cb6ded051cb0b7

stefanvanburen: (I had screwed up the assertions so that the failing one wasn't running, fixed in https://github.com/bufbuild/buf/pull/4259/commits/8e1e0729e2d5bc5145c42e5c311bad73d40ff7d0. 🤦)

doriable: So BSR docs uses leading comments (and effectively drops detached leading comments), which is the behaviour we want here -- do you mind adding a test case that captures this?

stefanvanburen: I think we have that with the existing test, which covers the `MethodOptions.deprecated` option; code [here](https://buf.build/protocolbuffers/wellknowntypes/file/main:google/protobuf/descriptor.proto#L946), docs [here](https://buf.build/protocolbuffers/wellknowntypes/docs/main:google.protobuf#google.protobuf.MethodOptions) (specifically pointing at the comment that doesn't exist in the docs).

private/buf/buflsp/symbol.go
emcfarlane: Played with this. I think we can get around the mutating newlines in the loop and have one var to track. What do you think of:
```go
	tok, _ := def.Context().Stream().Around(def.Span().Start)
	cursor := token.NewCursorAt(tok)
	var seenNewline bool
	for {
		t := cursor.PrevSkippable()
		if t.Kind() == token.Space {
			isNewline := strings.Contains(t.Text(), "\n")
			if isNewline && seenNewline {
				break
			}
			seenNewline = seenNewline || isNewline
		} else if t.Kind() == token.Comment {
			seenNewline = false
			comments = append(comments, commentToMarkdown(t.Text()))
		} else {
			break
		}
	}
	comments = lineUpComments(comments)
	// Reverse the list and return joined.
	slices.Reverse(comments)

	var docs strings.Builder
	for _, comment := range comments {
		docs.WriteString(comment)
		docs.WriteRune('\n')
	}
```

emcfarlane (approved):

private/buf/buflsp/symbol.go
stefanvanburen: basically stole your suggestion, slightly edited, here: https://github.com/bufbuild/buf/pull/4259/commits/781057be35727672d9d2f43527b5e2d931a2405d

private/buf/buflsp/symbol.go
```diff
@@ -453,32 +453,24 @@ func (s *symbol) getDocsFromComments() string {
...
-				addNewline = false
+	// traversing backwards for leading comments only.
+	tok, _ := def.Context().Stream().Around(def.Span().Start)
+	cursor := token.NewCursorAt(tok)
+	var seenNewline bool
+	for {
+		t := cursor.PrevSkippable()
+		if t.Kind() == token.Comment {
+			seenNewline = false
+			comments = append(comments, commentToMarkdown(t.Text()))
```
emcfarlane: Might be missing the `\n` newline in the comment, can append here or below to write directly to `strings.Builer`.

stefanvanburen: let me check...

stefanvanburen: yeah, you're right. let me make sure we have a test that covers that and fix it up.

stefanvanburen: ok, added a failing test in https://github.com/bufbuild/buf/pull/4259/commits/362ccca9d2a2887fa41b361041f76749b0b69667 and put some logic back in https://github.com/bufbuild/buf/pull/4259/commits/2646678a548ee08bce0f96dc2f674c8533bac334. (maybe could be simplified further?)

looks right to me now:

image

stefanvanburen: simplified a little further: https://github.com/bufbuild/buf/pull/4259/commits/2b4297c5224a4536015d978ed33018614143e9fc

doriable (approved):

private/buf/buflsp/symbol.go
emcfarlane: 👨‍🍳 2b4297c5224a4536015d978ed33018614143e9fc

PR #4260, opened 2026-01-06
Add LSP completion for field numbers

stefanvanburen:
Suggests the next available field number, taking into account existing field numbers in the message as well as reserved field numbers. We want this so that people get suggested the next lowest field number, from [protobuf.com][1]:

> When creating messages, authors should prefer lower numbers:
> unmarshaling a message is typically optimized for smaller field
> numbers. Also, smaller values are more efficiently encoded and decoded:
> field numbers 1 to 15 can be encoded in a single byte. So it is
> recommended to always start numbering fields at 1 and to use smaller
> values for more commonly used fields when possible.

Should work with or without the semicolon after, e.g.:

```proto
message ABC {
  string xyz = |
}
```

```proto
message ABC {
  string xyz = |;
}
```

[1]: https://protobuf.com/docs/language-spec#field-numbers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Update `PROTOVALIDATE` lint rule to support field mask rules.
+- Add LSP completion for field numbers.
 
 ## [v1.62.1] - 2025-12-29
 


private/buf/buflsp/completion.go
```diff
@@ -1497,6 +1517,25 @@ func defToOptionMessage(file *file, def ast.DeclDef) ir.MessageValue {
...
+		return false
+	}
+
+	before, _ := file.ir.AST().Stream().Around(offset)
+	cursor := token.NewCursorAt(before)
+
+	// Skip backwards over whitespace to find the previous non-whitespace token
+	for !before.IsZero() && before.Kind() == token.Space {
+		before = cursor.PrevSkippable()
+	}
```
emcfarlane: Good to use `isTokenSpace` here to avoid newlines.
```suggestion
	for isTokenSpace(before) {
		before = cursor.PrevSkippable()
	}
```
I don't think we need this in a loop, could probably get away with:
```go
if isTokenSpace(before) {
   before = cursor.PrevSkippable()
}
return isTokenEqual(before)
```

private/buf/buflsp/completion.go
```diff
@@ -1634,6 +1673,75 @@ func getOptionValueType(file *file, ctx context.Context, optionValue ir.MessageV
...
+		}
+	}
+
+	nextNumber := int32(1)
+	for {
+		if !usedNumbers[nextNumber] && !reservedNumbers[nextNumber] {
+			break
+		}
+		nextNumber++
+		if nextNumber > maxProtobufFieldNumber {
```
emcfarlane: lol

private/buf/buflsp/completion.go
```diff
@@ -1634,6 +1673,75 @@ func getOptionValueType(file *file, ctx context.Context, optionValue ir.MessageV
...
+						reservedNumbers[int32(num)] = true
+					}
+				}
+			}
+		}
+	}
+
+	nextNumber := int32(1)
+	for {
+		if !usedNumbers[nextNumber] && !reservedNumbers[nextNumber] {
```
emcfarlane: nit: just combine these sets to one.

private/buf/buflsp/completion.go
```diff
@@ -37,6 +38,9 @@ import (
 	"go.lsp.dev/protocol"
 )
 
+// Ref: https://protobuf.com/docs/language-spec#field-numbers
+const maxProtobufFieldNumber = 536_870_911
```
doriable: We can get this from `protowire`. We also need to respect the reserved range.

https://github.com/protocolbuffers/protobuf-go/blob/96a179180f0ad6bba9b1e7b6e38d0affb0168e9a/encoding/protowire/wire.go#L23-L29

private/buf/buflsp/completion.go
```diff
@@ -1497,6 +1517,25 @@ func defToOptionMessage(file *file, def ast.DeclDef) ir.MessageValue {
...
+	before, _ := file.ir.AST().Stream().Around(offset)
+	cursor := token.NewCursorAt(before)
+
+	// Skip backwards over whitespace to find the previous non-whitespace token
+	for !before.IsZero() && before.Kind() == token.Space {
+		before = cursor.PrevSkippable()
+	}
+
+	// Check if the first non-whitespace token is an equals sign
+	return !before.IsZero() && before.Kind() == token.Keyword && before.Text() == "="
```
doriable: Can check this against `keyword.Assign`.

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4260/commits/078a641bed561a1b22409f9cc4c4db6569262fda

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4260/commits/a10fd98e0170e8e8e2075d15030ee1782db80311

private/buf/buflsp/completion.go
```diff
@@ -1634,6 +1675,71 @@ func getOptionValueType(file *file, ctx context.Context, optionValue ir.MessageV
 	return ir.Type{}, false
 }
 
+func completionItemsForFieldNumber(
+	parentDef ast.DeclDef,
```
doriable, line 1682: So, I know that completions are based off the AST, and this has to do with resolving arbitrary and potentially invalid locations in the token stream, but given that we have the `parentDef` here, an alternative approach we can take is resolving the `parentDef`'s `symbol`, which would give us access to the `ir.Symbol`. And since the `parentDef` is asserted to be a valid message, the `ir.Symbol` should always be a `ir.Type`, which has a nice API for [accessing the reserved ranges](https://github.com/bufbuild/protocompile/blob/a57debf106713fee6498d10eb0a65bd4d6edbd40/experimental/ir/ir_type.go#L432), without us re-doing the parsing here.

There might be a catch with this approach, but maybe it's worth looking into?

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4260/commits/8aed814c451dea09fc5a2fd38963855733ef5b1b

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4260/commits/36b9ae851739fe472f253636b668baa2f0e48da4. Probably don't even need to check if the token is a keyword?

private/buf/buflsp/completion.go
stefanvanburen: https://github.com/bufbuild/buf/pull/4260/commits/7d7ca0784a07daf110ac57640e0a359acc75ff2b - there might be a simpler way to resolve the parentDef's symbol, though?

stefanvanburen: looks like maybe not, so just consolidated the callsites into the same helper: https://github.com/bufbuild/buf/pull/4260/commits/c383ffe9feabb21405bef7965ccddb5f473c4880

doriable: ^ honestly, this is pretty much exactly what I had in mind, lol. it's a little ugly because you need to pass the file around, but i feel like it's not the _worst_? given that it's always going to be scoped to the current file, it's pretty safe?

doriable (approved):
Looks good, I pulled in `main` and tested locally. Going to pre-approve, but do you mind putting the feature into the changelog?

stefanvanburen:
> Looks good, I pulled in `main` and tested locally. Going to pre-approve, but do you mind putting the feature into the changelog?

Added in https://github.com/bufbuild/buf/pull/4260/commits/e3776f58ddc98434f2b7c53234901586f3aff7db; merging, but feel free to wordsmith separately!

PR #4263, opened 2026-01-07
Add LSP references / {workspace,document} symbol tests

stefanvanburen:
Just continuing to pick away at coverage here. These are fairly straightforward, so I figured reviewing them together was reasonable.


stefanvanburen:
(From my [previous comment](https://github.com/bufbuild/buf/pull/4255#issuecomment-3711857259), after this I think we just have rename, formatting, semantic tokens left? (alongside any additional testing we might want on the other features.))

emcfarlane (approved):

PR #4264, opened 2026-01-07
Add LSP rename/formatting tests

stefanvanburen:
Tests rename/prepare rename with a variety of types, and a quick set of formatting tests.

### DIFF
--- a/private/buf/buflsp/format_test.go
+++ b/private/buf/buflsp/format_test.go
@@ -1,0 +1,89 @@
+// Copyright 2020-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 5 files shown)


private/buf/buflsp/format_test.go
```diff
@@ -0,0 +1,89 @@
...
+					URI: testURI,
+				},
+			}, &textEdits)
+			require.NoError(t, formatErr)
+			assert.Len(t, textEdits, tt.expectNumEdits)
+			if tt.expectEdits {
+				expectedFormatted := getExpectedFormattedContent(t, ctx, testProtoPath)
+				assert.Equal(t, expectedFormatted, textEdits[0].NewText)
+				assert.Equal(t, uint32(0), textEdits[0].Range.Start.Line)
+				assert.Equal(t, uint32(0), textEdits[0].Range.Start.Character)
```
stefanvanburen, line 73: If we get around to just sending diffs to the client, we can tighten up these assertions a little bit; seems ok for now, though.

emcfarlane (approved):

PR #4265, opened 2026-01-07
Add simple LSP semantic tokens test

stefanvanburen:
Ideally, we'd wire this up with some sort of golden-file approach for simplicity in updates, but I don't believe we have a standardized way of doing that. For now, we just trust what the server gives us, so we'll know if the value changes. We can manually update the asserted data if we add new supported fields.


stefanvanburen:
(Should wrap up the high-level suite of LSP features, based on my [previous comment](https://github.com/bufbuild/buf/pull/4263). Hopefully should help us catch issues in the future!)

Open to other approaches on this, but didn't feel like we needed to get in the business of parsing the data.

emcfarlane (approved):

PR #4267, opened 2026-01-09
Add LSP document highlighting

stefanvanburen:
This adds the LSP document highlighting feature, which highlights all references to a "symbol" in a given file.

In the protobuf case, I've taken this to mean messages and enums. (I think highlighting scalars is too noisy; RPCs and services only appear once. There may be other cases worth highlighting, but starting out conservative.)

Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentHighlight

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Add LSP document highlighting support.
+- Add LSP completion for fully-qualified type references.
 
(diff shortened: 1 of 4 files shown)


stefanvanburen:
I'm open to suggestions on what to highlight; take a peek and see if it makes sense to you.

video:

https://github.com/user-attachments/assets/296885e0-6285-4e3d-a3b2-2dd7d7a716af

PR #4268, opened 2026-01-09
Add LSP completion for fully-qualified references

stefanvanburen:
It's valid for a type to be referred to as either e.g. "google.protobuf.Timestamp" or ".google.protobuf.Timestamp"; add the fully-qualified case to completions.

Ref: https://protobuf.com/docs/language-spec#fully-qualified-references

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add LSP completion for fully-qualified type references.
 
 ## [v1.63.0] - 2026-01-06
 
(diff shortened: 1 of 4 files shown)


emcfarlane:
Even though it's valid I think preferring the non prefixed "." would avoid lots of duplicate completions, is this not an issue as we trim the completion list by a prefix?

emcfarlane (approved):

stefanvanburen:
> Even though it's valid I think preferring the non prefixed "." would avoid lots of duplicate completions, is this not an issue as we trim the completion list by a prefix?

yeah, we consider the prefix when doing completions (asserted in the test): 
image

but it _is_ an unbounded list of all visible types if you just start with `.`. tbh, I don't think that's an issue, and is actually desirable; sometimes you just want to see all the types available to you:

image

(does that cover your question?)

emcfarlane:
Yeah lgtm

PR #4270, opened 2026-01-09
Improve LSP semantic tokens

stefanvanburen:
The existing semantic tokens implementation had a couple of issues - for example, `message MyMessage` was all considered a semantic token of type "struct", when only "MyMessage" should be considered a struct and "message" should be a keyword.

Iterating on that, I've added some more testcases and handling for the various protobuf types. To handle things that aren't directly in the file's symbol's IR, I reworked the main loop to just collect the tokens up front, and then figure out sorting and encoding once we've grabbed all of the tokens.

---

I'm fairly certain this can be improved more, but think it's already a decent improvement on what we have, and lays the groundwork for future tweaks with additional testing.

For testing this, I highly recommend checking out this PR, then running `make installbuf`, and in neovim, using the `:Inspect` command on different tokens under the cursor, which should show a "Semantic Tokens" section that contains the semantic token under the cursor.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Fix LSP completion for options.
 - Add LSP document highlighting support.
 - Add LSP completion for fully-qualified type references.
+- Improve LSP semantic tokens implementation (for syntax highlighting).
 
 ## [v1.63.0] - 2026-01-06
 


private/buf/buflsp/semantic_tokens.go
stefanvanburen, line 1: the impl in server.go was getting unwieldy, and the constants only pertained to semantic tokens, so I figured a separate file was reasonable now.

private/buf/buflsp/semantic_tokens.go
```diff
@@ -0,0 +1,342 @@
...
+		if tok.Kind() == token.Comment {
+			collectToken(tok.Span(), semanticTypeComment, 0, keyword.Unknown)
+		}
+		kw := tok.Keyword()
+		switch kw {
+		// These keywords seemingly are not easy to reach via the IR.
+		case keyword.Option, keyword.Reserved, keyword.To, keyword.Returns:
+			collectToken(tok.Span(), semanticTypeKeyword, 0, kw)
+		}
+	}
```
stefanvanburen, line 139: originally I had this grabbing _all_ the keywords, but it's too greedy: syntax operators like `=`, `;`, `{` and `}` and all considered keywords in the taxa, and it makes the highlighting too confusing (everything becomes bolded as a keyword).

Instead, I tried to be more specific and go through the individual elements below. Again, this could probably be improved; I'm doubtless missing elements that we could probably add.

private/buf/buflsp/semantic_tokens_test.go
```diff
@@ -16,39 +16,273 @@ package buflsp_test
...
+			notExpectedTokens: []notExpectedToken{
+				// = and ; should not be keywords
+				{0, 7, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{0, 17, 1, semanticTypeKeyword, "';' should not be keyword"},
+				{10, 20, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{10, 26, 1, semanticTypeKeyword, "';' should not be keyword"},
+				{13, 14, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{13, 17, 1, semanticTypeKeyword, "';' should not be keyword"},
+				// Invalid import should not have string token (only resolved imports get string tokens)
+				{6, 12, 13, semanticTypeString, "invalid import path should not have string token"},
```
stefanvanburen, line 183: There's a test-case with an import to a file that doesn't exist, so the IR doesn't have it; seems fine to me — the alternative is uglier.

private/buf/buflsp/semantic_tokens_test.go
```diff
@@ -16,39 +16,273 @@ package buflsp_test
...
+				{50, 43, 17, semanticTypeStruct, "'ListUsersResponse' type"},
+			},
+			notExpectedTokens: []notExpectedToken{
+				// = and ; should not be keywords
+				{0, 7, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{0, 17, 1, semanticTypeKeyword, "';' should not be keyword"},
+				{10, 20, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{10, 26, 1, semanticTypeKeyword, "';' should not be keyword"},
+				{13, 14, 1, semanticTypeKeyword, "'=' should not be keyword"},
+				{13, 17, 1, semanticTypeKeyword, "';' should not be keyword"},
```
stefanvanburen, line 181: I don't think any of these "operators"(?) ought to be considered keywords; just leaving these negative tests around.

private/buf/buflsp/semantic_tokens_test.go
```diff
@@ -16,39 +16,273 @@ package buflsp_test
...
+				// import declarations (only resolved imports are tokenized)
+				{5, 0, 6, semanticTypeKeyword, "'import' keyword"},
+				{5, 7, 6, semanticTypeModifier, "'public' modifier"},
+				{5, 14, 34, semanticTypeString, "import path string"},
+				// option keyword
+				{10, 2, 6, semanticTypeKeyword, "'option' keyword"},
+				// decorator (option)
+				{10, 9, 10, semanticTypeDecorator, "'deprecated' as decorator"},
+				// built-in type
+				{13, 2, 6, semanticTypeProperty, "'string' as property"},
```
stefanvanburen: Need to revisit this; not sure that in `string name = 1;`, the `string` is a "property" (`name` is the property); maybe more just a `type` or `keyword`?:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#textDocument_semanticTokens

private/buf/buflsp/semantic_tokens_test.go
doriable: I think it would be too noisy to consider the operators in proto for keyword tokens, since it's mostly `=` for alignment and `;` for breaking decl statements. I like this as-is.

private/buf/buflsp/semantic_tokens_test.go
doriable: I like that idea of having built-in scalars being distinct from types declared for protos, but also, they aren't exactly keywords (e.g. I think `bool` is a type, and `true` and `false` are keywords). Since there isn't really a way to differentiate the types, maybe the scalar types should just be `type`... based on the spec.

stefanvanburen: yeah, I think `type` is reasonable; I wish there was a scalarType or something in the spec - oh well!: https://github.com/bufbuild/buf/pull/4270/commits/66a7a88646b176720317b2c0c7a8d4321cf0efaa

PR #4277, opened 2026-01-13
Add LSP organize imports code action

emcfarlane:
Add LSP organize imports code action to add missing imports, remove unused imports, and sort imports alphabetically. If imports match expected the code action is not shown. It uses tagged diagnostics from the compiler to find the unknown types and unused imports, this avoids re-parsing the file.

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -1542,7 +1542,7 @@ func isTokenSpace(tok token.Token) bool {
 	return tok.Kind() == token.Space && strings.IndexByte(tok.Text(), '\n') == -1
 }
 
-// isTokenNewline returns true for spaces, excluding newlines.
+// isTokenNewline returns true if tok is a newline.
 func isTokenNewline(tok token.Token) bool {
 	return tok.Kind() == token.Space && strings.Count(tok.Text(), "\n") == 1
 }
--- a/private/buf/buflsp/organize_imports.go
+++ b/private/buf/buflsp/organize_imports.go
@@ -47,8 +47,6 @@ func (s *server) getOrganizeImportsCodeAction(ctx context.Context, file *file) *
 		return nil
 	}
 
-	s.logger.Debug("code action: checking IR diagnostics", "count", len(file.irReport.Diagnostics))
-
 	// Find all unresolved type references and unused imports from the IR diagnostics
 	unresolvedRefs := make(map[ir.FullName]bool)
 	unusedImports := make(map[string]bool)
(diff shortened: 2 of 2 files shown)


private/buf/buflsp/organize_imports.go
```diff
@@ -0,0 +1,251 @@
...
+		if typ.FullName() == fullName {
+			return true
+		}
+	}
+	return false
+}
+
+// captureImportSpan expands the import span declaration to include both
+// leading and trailing comments and any trailing whitespace to the next
+// declaration.
```
doriable, line 231: I think this is generally reasonable and this is mostly a note for myself, but it might be worth extracting the comment attribution code path I have right now beyond the file descriptor/source code info use-case, so we have consistent comment attribution for the LSP.

stefanvanburen:
The only thing I think we ought to do here is remove duplicate imports. `gopls` seems to do: 

1. if no trailing comments on any, remove all except one
2. if one has a trailing import, keep that one
3. if multiple imports have a trailing comments, fail to run (we don't want to remove comments in basically any case). (or maybe we should just remove all except the ones that have comments on them, and leave it up to the user, since they'll get a diagnostic anyway?)

I guess this could be a follow-up, but seems valuable to include here.

emcfarlane:
@stefanvanburen added support to remove duplicates and handle unknowns. Duplicates are only removed if the text content of the span is different, which ensures we always keep at least one version of the commented import around.

private/buf/buflsp/organize_imports.go
emcfarlane: Would be nice to have a general handle for this.

private/buf/buflsp/completion.go
```diff
@@ -1542,6 +1542,11 @@ func isTokenSpace(tok token.Token) bool {
 	return tok.Kind() == token.Space && strings.IndexByte(tok.Text(), '\n') == -1
 }
 
+// isTokenNewline returns true for spaces, excluding newlines.
```
stefanvanburen: nit:

```suggestion
// isTokenNewline returns true if tok is a newline.
```

? Seems like it's saying the opposite now.

private/buf/buflsp/organize_imports.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+// getOrganizeImportsCodeAction generates a code action for organizing imports.
+// It uses IR diagnostics to find unresolved types, unused imports, and organizes all imports.
+func (s *server) getOrganizeImportsCodeAction(ctx context.Context, file *file) *protocol.CodeAction {
+	if file.workspace == nil || file.irReport == nil {
+		return nil
+	}
+
+	s.logger.Debug("code action: checking IR diagnostics", "count", len(file.irReport.Diagnostics))
```
stefanvanburen: nit: if we want to leave around the logs, shall we use `DebugContext` since we have a `ctx` available to us?

stefanvanburen (approved):
tried it out and worked well for me. when this merges, let's make sure to follow up w/ adding this to the docs (add to features; how to enable in vim)

private/buf/buflsp/organize_imports.go
emcfarlane: Removed.

PR #4278, opened 2026-01-13
Fix bugs in LSP option renaming

stefanvanburen:
Noticed that our renaming of option types was resulting in invalid syntax; e.g., renaming `test` to `testing` in `(example).test` was resulting in `(example)testing`. Also, renames to messages weren't being picked up in usages in `extend` blocks. Lastly, renaming was previously not including the fully-qualified name of the extension.

Notably, this specifically does _not_ allow for renaming from option "call-sites" (references). I think we can fix this later with further work.

Verified in [registry-proto](https://github.com/bufbuild/registry-proto).

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -423,29 +423,32 @@ func renameChangesForReferenceableSymbol(s *symbol, newName string) (map[protoco
 	if ok {
 		for _, reference := range referenceableKind.references {
 			newText := newName
-			// For option references (extension usages), use the fully-qualified package name from
-			// defFullName and preserve parentheses.
-			// e.g., if renaming "(testing)" to "validated", result is "(rename.v1.validated)"
-			// e.g., if renaming "(subpkg.testing)" to "validated", result is "(rename.v1.subpkg.validated)"
-			if opt, isOption := reference.kind.(*option); isOption {
+			// For option references (extension usages), preserve package qualification and parentheses.
+			// e.g., if renaming "(subpkg.testing)" to "validated", result should be "(subpkg.validated)"
+			if _, isOption := reference.kind.(*option); isOption {
 				spanText := reference.span.Text()
-				// Extract parentheses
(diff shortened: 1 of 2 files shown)


private/buf/buflsp/symbol.go
```diff
@@ -413,11 +413,48 @@ func renameChangesForReferenceableSymbol(s *symbol, newName string) (map[protoco
...
+		// If the symbol isn't referenceable itself, but has a referenceable definition, use the
+		// definition for the references.
+		referenceableKind, ok = s.def.kind.(*referenceable)
+	}
+	if ok {
+		for _, reference := range referenceableKind.references {
+			newText := newName
+			// For option references (extension usages), preserve package qualification and parentheses.
+			// e.g., if renaming "(subpkg.testing)" to "validated", result should be "(subpkg.validated)"
+			if _, isOption := reference.kind.(*option); isOption {
```
doriable: Rather than parsing the option name in this manner through the token, are we not able to use the full name of the option set on the kind? That would give us the `Parent()` and `Components()` APIs, which we can parse? Not sure, but might be worth trying.

stefanvanburen: I think the only trick with that is that we'd be potentially making a relative reference into an absolute one, because we're not retaining the textual information about what's there as-is. but maybe it doesn't really matter?

stefanvanburen: check out https://github.com/bufbuild/buf/pull/4278/commits/ab8be6bf0c900d20abd1d794c3259af4be3b4b65 for what I'm talking about, specifically the relative options becoming fully-qualified in the tests. maybe we could work around that, though?

doriable: Ah yeah, because this always uses the full name in that case... alright, I like what you had :D

stefanvanburen: https://github.com/bufbuild/buf/pull/4278/commits/f4eb952aed8f1c8a7ad1abc908f2da20158cf144 (reverted)

PR #4279, opened 2026-01-13
Add LSP DocumentLink support

stefanvanburen:
Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentLink
Ref: https://go.dev/gopls/features/passive#document-link

This does two things: identifies (for now, `https://`) links in comments to create clickable URLs in editors, and adds links to import statements (that currently go to the BSR files tab if the import seems to be a BSR module (with a special-case for WKTs, like hover), otherwise falls back to the local file URI).

### DIFF
--- a/private/buf/buflsp/document_link_test.go
+++ b/private/buf/buflsp/document_link_test.go
@@ -30,9 +30,9 @@ func TestDocumentLink(t *testing.T) {
 	ctx := t.Context()
 
 	tests := []struct {
-		name           string
-		protoFile      string
-		expectedLinks  []expectedLink
+		name          string
+		protoFile     string
+		expectedLinks []expectedLink
 	}{
 		{
 			name:      "local_import_and_comment_urls",
@@ -105,7 +105,7 @@ func TestDocumentLink(t *testing.T) {
 
 			for i, expected := range tt.expectedLinks {
 				link := links[i]
-				assert.Equal(t, uint32(expected.line), link.Range.Start.Line, "link %d (%s): wrong line", i, expected.description)
+				assert.Equal(t, expected.line, link.Range.Start.Line, "link %d (%s): wrong line", i, expected.description)
 
 				switch expected.targetType {
 				case linkTargetTypeLocal:


stefanvanburen:
For testing, I'd check out VS Code; you should be able to command-click the links (either in comments or the imports) to open them up.

The one weird thing I've noticed with VS Code is command-clicking an import both opens up the link, _and_ opens the local file; I'm not sure what is up with that behavior (maybe it's also doing a go-to-definition on command+click?).

Support for neovim has [stalled out](https://github.com/neovim/neovim/pull/33500), but [this plugin](https://github.com/icholy/lsplinks.nvim) adds support.

private/buf/buflsp/document_link.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+
+			links = append(links, protocol.DocumentLink{
+				Range:  reportSpanToProtocolRange(urlSpan),
+				Target: protocol.DocumentURI(url),
+			})
+
+			offset = urlEnd
+		}
+	}
```
stefanvanburen, line 144: this is pretty janky; I could see dropping this support for now (even though it seems reasonable), or using a well-tested/small package like https://pkg.go.dev/mvdan.cc/xurls/v2#StrictMatchingScheme to do this.

private/buf/buflsp/document_link.go
```diff
@@ -0,0 +1,150 @@
...
+			var targetURI protocol.DocumentURI
+
+			// Try to get BSR module information
+			if file.workspace != nil && imported.file.ir != nil {
+				module := file.workspace.GetModule(imported.file.uri)
+				packageName := string(imported.file.ir.Package())
+				bsrURL := bsrDocsURL(module, packageName, "")
+				if bsrURL != "" {
+					targetURI = protocol.DocumentURI(bsrURL)
+				}
```
stefanvanburen: I could make the argument that instead of going to the package docs tab, an import (pointing directly at a file) should go to the code tab of the file, but that feels weird because that's not much better than viewing the code locally, and the hover docs go to the docs tab instead of the code tab.

private/buf/buflsp/document_link.go
stefanvanburen: swapped this out for `xurls` in https://github.com/bufbuild/buf/pull/4279/commits/c742a00963d7ea797555221363f376d1ef4f10fa; I think it's probably better than doing this ourselves, but also open to just removing the support for now.

private/buf/buflsp/symbol.go
```diff
@@ -591,29 +592,46 @@ func (s *symbol) getDocsFromComments() string {
...
-				defFullName,
-				bsrHost,
-				s.def.file.ir.Package(),
-				bsrAnchor,
-			))
+			packageName := string(s.def.file.ir.Package())
+			var bsrURL string
+			if s.def.file.IsWKT() {
+				// WKT uses special bsrHost format
+				bsrURL = "https://" + bsrHost + "/docs/main:" + packageName
```
emcfarlane: Are WKT imports mean to use this link? Currently goes to cache.

stefanvanburen: Ah yeah, that's wrong. Let me fix it up.

stefanvanburen: ok, in the process of [fixing this](https://github.com/bufbuild/buf/pull/4279/commits/d88fec5aaf07c7207bd06fd800f7fbd40066bd82), I made the decision that these import links should go to the code tab instead (since they have raw file paths in them), which I had [mentioned above](https://github.com/bufbuild/buf/pull/4279#discussion_r2688204174). I think that behavior is more correct for imports, since they are raw file paths, not referring to packages.

private/buf/buflsp/document_link.go
stefanvanburen: ended up going with this approach; see [this comment below](https://github.com/bufbuild/buf/pull/4279#discussion_r2691125616).

emcfarlane (approved):

PR #4280, opened 2026-01-14
Add LSP folding range support

stefanvanburen:
Ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_foldingRange
Ref: https://go.dev/gopls/features/passive#folding-range

This adds folding range support to the LSP. Folding ranges apply to messages, nested messages, enums, extensions, options, import blocks, multi-line comments, oneofs, services and methods.

By default, it looks like VS Code might use tree-sitter to compute folds, which isn't too bad, but this approach is more precise. neovim also supports either tree-sitter or LSP as a fold provider. For testing, reference: https://neovim.io/doc/user/lsp.html#vim.lsp.foldexpr().


doriable (approved):
This looks reasonable to me -- going to stamp under the assumption that the merge will be resolved. Thank you!

PR #4281, opened 2026-01-14
Support multiple file paths in `buf format`

### LINKED ISSUE #4233: `buf format` should accept multiple file paths as positional arguments
## Summary

The `buf format` command currently accepts only a single `<source>` positional argument. This makes it incompatible with tools like [lint-staged](https://github.com/lint-staged/lint-staged), which pass multiple file paths as positional arguments to formatters. Supporting multiple file paths would greatly improve the developer experience for teams using `buf format` in pre-commit hooks.

## Current Behavior

```bash
$ buf format -w file1.proto file2.proto file3.proto
# Error: accepts at most 1 arg(s), received 3
```

The command fails when given multiple files as positional arguments.

## Expected Behavior

```bash
$ buf format -w file1.proto file2.proto file3.proto
# Successfully formats all three files
```

The command should accept multiple file paths and format each of them.

## Use Case

Many teams use **lint-staged** with **husky** (or similar tools like **pre-commit**) to run formatters only on staged files during git commits. This is a common pattern for maintaining code quality without formatting the entire codebase on every commit.

lint-staged works by passing the list of staged files as positional arguments to the configured command:

```json
{
  "**/*.proto": "buf format -w"
}
```

When multiple `.proto` files are staged, lint-staged executes:

```bash
buf format -w path/to/file1.proto path/to/file2.proto path/to/file3.proto
```

This pattern works out-of-the-box with most formatters:

- `prettier --write` ✅
- `biome check --write` ✅
- `eslint --fix` ✅
- `black` (Python) ✅
- `gofmt -w` ✅
- `buf format -w` ❌

## Current Workaround

The workaround is to use a bash wrapper that converts positional arguments to `--path` flags:

```json
{
  "**/*.proto": "bash -c 'args=(); for f in \"$@\"; do args+=(--path \"$f\"); done; buf format -w . \"${args[@]}\"' _"
}
```

While functional, this workaround:

- Is not intuitive or discoverable
- Adds complexity to the configuration
- May not work on all platforms (Windows without bash)
- Requires understanding of bash array manipulation

## Proposed Solution

Accept multiple file paths as positional arguments:

```bash
# Format multiple specific files
buf format -w file1.proto file2.proto file3.proto

# Equivalent to current behavior with --path
buf format -w . --path file1.proto --path file2.proto --path file3.proto
```

### Alternative: Accept files via stdin

Another option would be to accept file paths from stdin, similar to `xargs`:

```bash
echo "file1.proto file2.proto" | buf format -w --stdin
```

## Additional Context

This is a common pattern request for CLI tools that integrate with the JavaScript/TypeScript ecosystem's pre-commit tooling. The lint-staged project has over 13k GitHub stars and is widely used, so improving compatibility would benefit many developers working with Protocol Buffers.

Related tools that have addressed similar requests:

- Prettier accepts multiple files: `prettier --write file1 file2 file3`
- ESLint accepts multiple files: `eslint --fix file1 file2 file3`
- Biome accepts multiple files: `biome check --write file1 file2 file3`

stefanvanburen:
This matches the behavior of `gofmt`, as well as other tools mentioned in #4233.

The only real complexity here is that we're enforcing `-o` (if used) specifies a directory for now if multiple files are passed (I don't think there's a use-case for formatting multiple files that output to a single file, or a non-matching number of output files), and we disallow using `--path` with multiple input files.

Fixes #4233.

### DIFF
--- a/cmd/buf/buf_test.go
+++ b/cmd/buf/buf_test.go
@@ -4401,7 +4401,7 @@ func TestFormatMultipleFiles(t *testing.T) {
 	t.Parallel()
 	tempDir := t.TempDir()
 	formattedDir := filepath.Join(tempDir, "multifile.formatted")
-	// First, format the multiple file paths to a temp directory
+	// Format the multiple file paths to a temp directory
 	testRunStdout(
 		t,
 		nil,
@@ -4413,29 +4413,6 @@ func TestFormatMultipleFiles(t *testing.T) {
 		filepath.Join("testdata", "format", "multifile", "file3.proto"),
 		"-o",
 		formattedDir,
-	)
-	// Verify that multiple file paths can be passed as positional arguments
-	// and formatting with -d produces no diff
-	testRunStdout(
-		t,
-		nil,
-		0,
-		``,
-		"format",
-		filepath.Join("testdata", "format", "multifile", "file1.proto"),
-		filepath.Join("testdata", "format", "multifile", "file2.proto"),
-		filepath.Join("testdata", "format", "multifile", "file3.proto"),
-		"-d",
-	)
-	// Verify formatting the whole directory with -d produces no diff
-	testRunStdout(
-		t,
-		nil,
-		0,
-		``,
-		"format",
-		formattedDir,
-		"-d",
 	)
 }
 


doriable (commented):
This approach makes sense to me -- it's non-breaking to existing customers, and I think the most reasonable approach is not to mix the use of `--path` with multiple path arguments to avoid confusion.

doriable:
After some internal discussions, I think we're not going to be pursuing this feature now. I will update the issue also. Thank you!

PR #4283, opened 2026-01-15
Add LSP completion for next field in enums

stefanvanburen:
Similar to how this works for messages, but with an enum-specific flair; enum values must be an int32 (including negative values).

Also simplifies the CompletionItem to only include a Label; we don't need to specify `InsertText` if it's the same value.

(We mentioned this feature specifically in [the LSP blog post][1] and I realized I had totally forgotten about doing it alongside message field completion in #4260.)

[1]: https://buf.build/blog/protobuf-lsp

### DIFF
--- a/private/buf/buflsp/completion.go
+++ b/private/buf/buflsp/completion.go
@@ -39,6 +39,12 @@ import (
 	"google.golang.org/protobuf/encoding/protowire"
 )
 
+// numberRange represents a range of numbers (inclusive) used for field/enum number completion.
+type numberRange[T int32 | uint64] struct {
+	start T
+	end   T
+}
+
 // getCompletionItems returns completion items for the given position in the file.
 //
 // This function is called by the Completion handler in server.go.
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/completion.go
```diff
@@ -1738,12 +1754,63 @@ func completionItemsForFieldNumber(
...
+		for member := range seq.Values(irType.Members()) {
+			// For enums, all numbers are valid including 0 (unlike fields where 0 is invalid)
+			num := member.Number()
+			usedOrReservedEnumNumbers[num] = true
+		}
+
+		// Collect reserved ranges
+		for reservedRange := range seq.Values(irType.ReservedRanges()) {
+			start, end := reservedRange.Range()
+			for i := start; i <= end; i++ {
```
emcfarlane: This could cause an issue if a large range size is specified.

stefanvanburen: yeah, let me fix that (and also in the message impl which does the same thing)

stefanvanburen: a [little more complex](https://github.com/bufbuild/buf/pull/4283/commits/f6eab85ad7abe8a4d3740b72800b926940bdbfab), but should still be quick since we skip over entire ranges if they conflict.

emcfarlane: Asked claude with some guidance to flesh out an alternative to avoid the nested loops. Haven't validated fully:
```
type numberRange struct {
    start uint64
    end   uint64
}

func findNextAvailableFieldNumber(irType IRType) *uint64 {
    var ranges []numberRange
    
    // Add protobuf reserved range
    ranges = append(ranges, numberRange{
        start: uint64(protowire.FirstReservedNumber),
        end:   uint64(protowire.LastReservedNumber),
    })
    
    if !irType.IsZero() {
        // Add each member as a range of one
        for member := range seq.Values(irType.Members()) {
            num := uint64(member.Number())
            ranges = append(ranges, numberRange{start: num, end: num})
        }
        
        // Add reserved ranges
        for reservedRange := range seq.Values(irType.ReservedRanges()) {
            start, end := reservedRange.Range()
            if end == math.MaxInt32 {
                return nil // Can't allocate beyond open-ended range
            }
            ranges = append(ranges, numberRange{
                start: uint64(start),
                end:   uint64(end),
            })
        }
    }
    
    // Sort by start position
    sort.Slice(ranges, func(i, j int) bool {
        return ranges[i].start < ranges[j].start
    })
    
    // Find first gap starting from 1
    nextNumber := uint64(1)
    for _, r := range ranges {
        if nextNumber < r.start {
            // Found a gap before this range
            return &nextNumber
        }
        // Move past this range
        if r.end >= nextNumber {
            nextNumber = r.end + 1
        }
    }
    
    // Check if we're still within valid range
    if nextNumber <= uint64(protowire.MaxValidNumber) {
        return &nextNumber
    }
    
    return nil
}
```

stefanvanburen: ended up being pretty close, but didn't genericize into a function because we have different max values and all that: https://github.com/bufbuild/buf/pull/4283/commits/ddb61df298a0f50f4e85c017a3e3f8da89bcff0c

(probably could, but I think it's fine)

emcfarlane: lgtm!

Issue #3207, opened 2024-08-04
Add post-processing commands


benedikt-bartscher:
### Feature

I would love to see an option for the buf generator which adds custom post-processing commands.
For example running ruff after generating code with python-betterproto.

Slack ref: https://bufbuild.slack.com/archives/CRZ680FUH/p1722782692615029

benedikt-bartscher:
Commands could be specified per plugin. It should be possible to pass the out directory to the commands, f.e. like this:

buf.gen.yaml
```yaml
plugins:
  - local: protoc-gen-python_betterproto
    out: gen
    post:
      - "ruff --fix $out"
```

andreas-04:
I would like to see this as well, curious why it isn't an option

doriable:
Hi, just responding to this after discussing this feature internally. This feature is outside of the scope of `buf` and `buf generate`. I understand the concern raised in this issue, in that some plugins may not generate code that conform to the standards of a project and require additional post-processing (e.g. formatting), however this should be managed holistically across a project -- the same thing that ensures that a developer's code is correctly formatted should be the same thing that ensures generated code is correctly formatted, and these checks should be managed through the project's higher-level CI processes, etc.

Closing this issue, since as mentioned, this feature is considered outside of the scope of `buf generate`.

PR #4287, opened 2026-01-18
feat: add `postprocess_cmd` option to run commands after code generation

amithkk:
This PR adds a new `postprocess_cmd` option to plugin configurations in `buf.gen.yaml`. This allows users to specify commands that run after code generation for each plugin, enabling automatic formatting, linting, or other post-processing of generated files.

### Motivation

Solves #3207

Many protoc plugins generate code that doesn't conform to a project's formatting standards and some languages need postprocessing (eg. python) for the modules to be importable. Currently, users must manually run formatters and cleanup scripts after `buf generate`, or maintain separate scripts. This feature allows integrating post-processing directly into the generation workflow.

Common use cases:
- Running `gofmt` or `goimports` on generated Go code
- Running `proletariat` on generated python code to fix imports

### Usage

```yaml
version: v2
plugins:
  - local: protoc-gen-python_betterproto
    out: gen/python
    postprocess_cmd:
      - "ruff check --fix $out"
      - "black $out"
```

### Supported Variables

Commands support variable substitution:
- `$out` - output directory (resolved with any base output path)
- `$name` - plugin name
- `$opt` - plugin options (comma-separated)
- `$path` - plugin path (for local plugins)
- `$strategy` - generation strategy ("directory" or "all")

Unit tests have been added

### DIFF
--- a/private/buf/bufgen/postprocess_test.go
+++ b/private/buf/bufgen/postprocess_test.go
@@ -1,0 +1,126 @@
+// Copyright 2020-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


doriable:
Thank you for submitting this PR. As I responded in the issue, this feature is unfortunately outside of the scope of `buf generate`, and we do not plan on implementing this.

PR #4296, opened 2026-01-20
Improve LSP map support

stefanvanburen:
This improves LSP handling of `map` types, including completions for the keyword and types within `<` angle brackets `>`. It also fixes up semantic tokens to properly tokenize maps, and removes document highlighting from map types (which aren't meant to be reference-able), and adds hover documentation for maps.

### DIFF
--- a/private/buf/buflsp/completion_test.go
+++ b/private/buf/buflsp/completion_test.go
@@ -283,10 +283,10 @@ func TestCompletionMaps(t *testing.T) {
 		},
 		{
 			name:                "complete_map_value_type_incomplete_syntax",
-			line:                34,                                    // Line with "map<int32, > attributes2 = 31;"
-			character:           13,                                    // Right after "  map<int32, " (before >)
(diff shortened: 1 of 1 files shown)


stefanvanburen:
hover:

image

completion for keys (only allowed types):

image

completion for values (lol, nvim display is bad):

image

document highlight works:

image

semantic tokens are valid now, etc.

private/buf/buflsp/completion.go
```diff
@@ -1590,6 +1652,68 @@ func isTokenTypeDelimiter(tok token.Token) bool {
...
+		case ',':
+			if angleDepth == 0 {
+				commaFound = true
+			}
+		case '\n', ';':
+			return false, false
+		}
+	}
+
+	return false, false
```
stefanvanburen, line 1714: this is pretty unfortunate, but not too dissimilar from what we do in other incomplete field situations

PR #4300, opened 2026-01-23
Add number of deprecated types to protostat

cyinma:


private/pkg/protostat/protostat_test.go
```diff
@@ -0,0 +1,257 @@
...
+			name: "outer deprecated but nested not",
+			content: `
+				syntax = "proto3";
+				message Outer {
+					option deprecated = true;
+					message Inner {}
+				}
+			`,
+			expectedTypes:           2,
+			expectedDeprecatedTypes: 1,
```
cyinma, line 107: This might be the most ambiguous case, but I think each nested type should be marked as deprecated explicitly in order to be considered deprecated

PR #4306, opened 2026-01-26
Add buf source edit deprecate command

emcfarlane:
Adds a new `buf source edit deprecate` command that edits Protobuf source files in-place. The command supports `--name` flags that adds `option deprecated = true;` to types matching the FQN prefix.


cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
```diff
@@ -0,0 +1,271 @@
...
+
+const (
+	configFlagName          = "config"
+	diffFlagName            = "diff"
+	diffFlagShortName       = "d"
+	disableSymlinksFlagName = "disable-symlinks"
+	errorFormatFlagName     = "error-format"
+	excludePathsFlagName    = "exclude-path"
+	pathsFlagName           = "path"
+	nameFlagName            = "name"
```
doriable: Should we call this flag "prefix" instead? It can be an exact match to a name or a prefix, right?

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
```diff
@@ -0,0 +1,271 @@
...
+		diffBuffer,
+		originalReadBucket,
+		formattedReadBucket,
+		storage.DiffWithExternalPaths(),
+	)
+	if err != nil {
+		return err
+	}
+	diffExists := diffBuffer.Len() > 0
+
```
doriable: We should only bother doing the work of calculating the diff `if flags.Diff {...}` -- unlike `format`, we are not looking to return an error code based on the diff existing.

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
```diff
@@ -0,0 +1,271 @@
...
+		Short: "Deprecate Protobuf types",
+		Long: `
+Deprecate Protobuf types by adding the 'deprecated = true' option.
+
+The --name flag is required and specifies the fully-qualified name prefix of the
+types to deprecate. All types whose fully-qualified name starts with this prefix
+will have the 'deprecated = true' option added. For fields and enum values, only
+exact matches are deprecated.
+
+By default, the source is the current directory and files are formatted and rewritten in-place.
```
doriable, line 70: The UX of this command as it currently stands is that even if nothing is deprecated (either they're already deprecated or there are no matches), this will always format the file. I'm undecided whether or not this is desirable... I think it's fine? In which case, related to my comment on the diff, maybe we do want to handle some kind of exit code when any diff exists.

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
stefanvanburen: `prefix` is probably more correct, although it does end up being more to type :).

Are we expecting in the future to add different flags to this command, like `--file=<name>` to deprecate a file? I guess I'm wondering if this ought to be a flag at all, and not just a positional arg?

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
stefanvanburen: I guess my vague notion would be that we ought to not format anything if we don't match, and perhaps even return an exit code of `1`, in the same vein of doing an `rg <match>` that doesn't match anything?

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
emcfarlane: Use it for avoiding writes, getting the changedPaths, changed diff to output to io.Discard if not using it.

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
emcfarlane: Added a doc comment about the error handling. We now error if no provided prefix matches. If prefix match but already deprecated/format command succeeds. If prefix match but not formatted, still formats.

cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate/sourceeditdeprecate.go
emcfarlane: Moved to `prefix`. I think it's preferred to have a flag with the one arg being the input.

stefanvanburen: ah right, I keep forgetting about the `<source>` arg. ignore me.

PR #4310, opened 2026-01-30
Fix edge case in LSP symbol doc hover

stefanvanburen:
See the added testcase; we don't want the trailing comment on the import to be included as a doc comment for the message without a comment.

### DIFF
--- a/private/buf/buflsp/symbol.go
+++ b/private/buf/buflsp/symbol.go
@@ -567,10 +567,10 @@ func (s *symbol) getDocsFromComments() string {
 	for {
 		t := cursor.PrevSkippable()
 		if t.Kind() == token.Comment {
-			text := commentToMarkdown(t.Text())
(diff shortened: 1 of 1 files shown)


private/buf/buflsp/testdata/hover/unused_import.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
+
+import "google/protobuf/cpp_features.proto"; // unused import
+
+message TestTopLevel {
```
emcfarlane, line 5: We should also handle:
```
import "google/protobuf/cpp_features.proto"; // unused import
message TestTopLevel {}
```
I think we need the comment handling to be aware that it is on it's own newline.

stefanvanburen: good call. let me make sure we handle that.

PR #4311, opened 2026-01-30
Add tags to LSP diagnostics

stefanvanburen:
Also removes the diagnostic.Data from the new diagnostics, which AFAICS is unused; we can save ourselves the json marshaling.

As a part of this, added a first pass at adding tests for diagnostics; we don't currently have any. These are a bit different because we haven't implemented [PullDiagnostics][1] (we can add this, but we'd need a custom implementation, or wait until [go.lsp.dev upgrades to 3.17][2]). Instead, we intercept the notifications from the server. Lastly, since we don't actually want the tests to wait around, adds synctest and only runs the tests on Go 1.25 for now (1.26 should be released next month and we'll be able to upgrade to 1.25 as our minimum).

This is in preparation to eventually landing relatedInformation (other spans) in diagnostics, from bufbuild/protocompile#659.

Also pulls in the latest protocompile@main to bring in the new deprecated tag.

[1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics
[2]: go-language-server/protocol#52

### DIFF
--- a/private/buf/buflsp/diagnostic.go
+++ b/private/buf/buflsp/diagnostic.go
@@ -40,7 +40,7 @@ var reportLevelToDiagnosticSeverity = map[report.Level]protocol.DiagnosticSeveri
 // corresponding [protocol.Diagnostic].
 func reportDiagnosticToProtocolDiagnostic(
 	reportDiagnostic report.Diagnostic,
-) (protocol.Diagnostic, error) {
+) protocol.Diagnostic {
 	diagnostic := protocol.Diagnostic{
 		Source:   serverName,
 		Severity: reportLevelToDiagnosticSeverity[reportDiagnostic.Level()],
@@ -70,5 +70,5 @@ func reportDiagnosticToProtocolDiagnostic(
 			protocol.DiagnosticTagDeprecated,
 		}
 	}
-	return diagnostic, nil
+	return diagnostic
 }
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -290,17 +290,7 @@ func (f *file) RefreshIR(ctx context.Context) {
 	fileDiagnostics := xslices.Filter(diagnosticReport.Diagnostics, func(d report.Diagnostic) bool {
 		return d.Primary().Path() == f.objectInfo.Path()
 	})
-	diagnostics, err := xslices.MapError(
-		fileDiagnostics,
-		reportDiagnosticToProtocolDiagnostic,
-	)
-	if err != nil {
-		f.lsp.logger.Error(
-			"failed to parse report diagnostics",
-			xslog.ErrorAttr(err),
-		)
-	}
-	f.diagnostics = diagnostics
+	f.diagnostics = xslices.Map(fileDiagnostics, reportDiagnosticToProtocolDiagnostic)
 	f.lsp.logger.DebugContext(
 		ctx, "ir diagnostic(s)",
 		slog.String("uri", f.uri.Filename()),


private/buf/buflsp/buflsp_test.go
```diff
@@ -121,26 +121,20 @@ func setupLSPServer(
...
+		ctx,
+		"test",
+		wktBucket,
+		appextContainer,
+		controller,
+		wasmRuntime,
+		stream,
+		queryExecutor,
+	)
+	require.NoError(t, err)
```
stefanvanburen, line 134: just simplified; `Serve` will start running in the background, so we don't need the extra goroutine.

private/buf/buflsp/diagnostic.go
```diff
@@ -70,20 +60,15 @@ func reportDiagnosticToProtocolDiagnostic(
...
-	}
-	bytes, err := json.Marshal(data)
-	if err != nil {
-		return protocol.Diagnostic{}, err
-	}
-	if bytes != nil {
-		// We serialize the bytes into a string before providing the structure to diagnostic.Data
-		// because diagnostic.Data is an interface{}, which is treated as a JSON "any", which
-		// will not cleanly deserialize.
-		diagnostic.Data = string(bytes)
```
stefanvanburen, line 86: we aren't using this data; it's just included so it can be passed back to us in a subsequent codeAction (ref: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic). I think we can leave it off, for now.

private/buf/buflsp/file.go
```diff
@@ -1148,7 +1148,13 @@ func (f *file) appendAnnotations(source string, annotations []bufanalysis.FileAn
 			Severity: protocol.DiagnosticSeverityWarning,
 			Source:   source,
 			Message:  annotation.Message(),
-		})
+		}
+		if annotation.Type() == "IMPORT_USED" {
+			diagnostic.Tags = []protocol.DiagnosticTag{
+				protocol.DiagnosticTagUnnecessary,
+			}
+		}
```
stefanvanburen, line 1146: this is the old diagnostic from the lint server that we're currently using; the [const is here (unexported)](https://github.com/bufbuild/buf/blob/c4f9a5ea2af656fcf4ac66c05a0fb88b17f45be7/private/bufpkg/bufcheck/bufcheckserver/internal/bufcheckserverbuild/bufcheckserverbuild.go#L704); if it changes out from under us, the test will catch it, but I suspect we're not going to change anything in that package in the near future.

private/buf/buflsp/symbol.go
```diff
@@ -364,8 +364,10 @@ func (s *symbol) GetSymbolInformation() protocol.SymbolInformation {
 		Kind:          kind,
 		Location:      location,
 		ContainerName: containerName,
-		// TODO: Use Tags with a protocol.CompletionItemTagDeprecated if the client supports tags.
-		Deprecated: isDeprecated,
+		Deprecated:    isDeprecated,
+		Tags: []protocol.SymbolTag{
+			protocol.SymbolTagDeprecated,
+		},
```
stefanvanburen, line 370: I still think we ought to have a more generalized idea of client capabilities, but it doesn't seem like a bad idea to just populate both of these values; figured I'd nix the TODO.

private/buf/buflsp/testdata/diagnostics/unused_import.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package diagnostics.v1;
+
+import "google/protobuf/timestamp.proto";
```
stefanvanburen, line 5: rendered in grey, from the "unnecessary" tag:

Image

doriable (approved):

PR #4318, opened 2026-02-04
Add LSP diagnostic when compilation fails

stefanvanburen:
Noticed that we weren't generating diagnostics for an import of a non-existent file, which made it confusing when the file looked fine otherwise.

Also, fixes the diagnostic to use proper .Start() and .End() for the diagnostic range, instead of a zero-width range consisting of only the start. (I might be missing something here, but worked for me; this was seemingly added way back in bufbuild/protocompile#205.)

### DIFF
--- a/private/buf/buflsp/image.go
+++ b/private/buf/buflsp/image.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"io/fs"
 	"log/slog"
+	"slices"
 	"strings"
 
 	"buf.build/go/standard/xlog/xslog"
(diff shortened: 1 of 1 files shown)


stefanvanburen:
Looks like:

image

private/buf/buflsp/image.go
```diff
@@ -88,6 +89,13 @@ func buildImage(
 			diagnostics = xslices.Map(errorsWithPos, func(errorWithPos reporter.ErrorWithPos) protocol.Diagnostic {
 				return newDiagnostic(errorWithPos, false, opener, logger)
 			})
+		} else {
```
emcfarlane: Just checking this isn't required in addition to the diagnostics? Should they be appended together?

emcfarlane: Is it part of the `warningErrorsWithPos`? Why don't we include those upfront?

stefanvanburen: I just checked; it's not a part of `warningErrorsWithPos`. (It looks like `warningErrorsWithPos` is basically a dupe of `buf lint` diagnostics, hence why we don't actually use those values here.)

stefanvanburen: > Just checking this isn't required in addition to the diagnostics? Should they be appended together?

we might as well append them together, I think.

stefanvanburen: @emcfarlane does https://github.com/bufbuild/buf/pull/4318/commits/6965cea5699ead47a1d5100813eb05d6834a1f16 look like what you were thinking?

PR #4322, opened 2026-02-09
Fix bufimageutil issues regarding weak imports

jhump:
I was porting a lot of this code to a new stand-alone package (for now, in the bufbuild/extra repo) and when I was comparing test output in those ported tests to the original golden files in this package, there were some discrepancies. I was able to understand all of the discrepancies except ones related to weak imports in the "testdata/sourcecodeinfo" files. It turns out that there was a bug in the logic, that happened to manifest slightly differently between this original code and the ported code in extra.

So this fixes the logic. The issue is that it was using the "from" and "to" indexes in the `dependencies` field of `FileDescriptorProto` when updating source code info for the `weak_dependencies` field 😞. So this fixes that.

I also updated the "testdata/sourcecodeinfo" golden files to include the actual filtered sources, instead of only the source code info blob, in order to compare the filtered source code info to those filtered sources when debugging. I updated a handful of other small things, too, like fixing stale symbols in doc comments.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -476,6 +476,12 @@ linters:
         # and should be considered safe.
         path: private/bufpkg/bufcobra/markdown.go
         text: "printf: non-constant format string in call to p"
+      - linters:
+          - gosec
+        # This converts slice indexes in a FileDescriptorProto to int32,
+        # which are not an actual risk of overflow.
+        path: private/bufpkg/bufimage/bufimageutil/image_filter.go
+        text: "G115:"
 
 issues:
   max-same-issues: 0


private/bufpkg/bufimage/bufimageutil/testdata/sourcecodeinfo/NestedFoo.txtar
```diff
@@ -1,3 +1,192 @@
...
+  // Must be a valid serialized protocol buffer of the above specified type.
+  bytes value = 2;
+}
+-- test.proto --
+// Keep 1: comment on syntax
+syntax = "proto2";
+// Keep 2: comment on package
+package foo.bar;
+// Keep if NestedFoo: comment on import any.proto
+import "google/protobuf/any.proto";
```
jhump, line 165: Annoyingly, this should be `import weak`. The actual filtered descriptor and source code info is correct. The issue turns out to be the way we package up the descriptor protos into `protoreflect.FileDescriptor` instances: those types no longer have any support for weak imports as of v1.36.5. 🤷

private/bufpkg/bufimage/bufimageutil/image_filter.go
```diff
@@ -314,14 +319,16 @@ func (b *sourcePathsBuilder) remapDependencies(
...
+			path := append(weakDependencyPath, weakFrom)
 			indexTo := dependencyChanges[indexFrom]
 			if indexTo == -1 {
 				sourcePathsRemap.markDeleted(path)
 			} else {
-				if indexTo != indexFrom {
-					sourcePathsRemap.markMoved(path, indexTo)
+				weakTo := int32(len(newWeakDependencies))
+				if weakTo != weakFrom {
+					sourcePathsRemap.markMoved(path, weakTo)
```
jhump, line 331: This is the actual fix. Other stuff in this PR is incidental changes/improvement (trying to leave everything in better shape than when I arrived) and the updated test outputs.

private/bufpkg/bufimage/bufimageutil/image_filter.go
```diff
@@ -254,6 +255,10 @@ func (b *sourcePathsBuilder) remapDependencies(
 	dependencies := fileDescriptor.GetDependency()
 	publicDependencies := fileDescriptor.GetPublicDependency()
 	weakDependencies := fileDescriptor.GetWeakDependency()
+	// TODO: Add support for option dependencies when buf CLI supports edition 2024.
+	if len(fileDescriptor.GetOptionDependency()) > 0 {
+		return nil, nil, nil, false, errors.New("edition 2024 not yet supported: cannot filter a file descriptor with option dependencies")
+	}
```
doriable, line 261: This is mostly a note for myself, I appreciate the `TODO`, I'm porting over the new compiler for edition 2024 support right now, so this is helpful for me when I rebase on this. Thank you!

doriable (approved):

private/bufpkg/bufimage/bufimageutil/testdata/sourcecodeinfo/NestedFoo.txtar
doriable: Yeah, noticed that when we did the upgrade here :< https://github.com/bufbuild/buf/pull/3653

PR #4325, opened 2026-02-12
Add LSP workaround for filepaths with `@` symbols

stefanvanburen:
It appears that vscode has it's own URI implementation that will [escape `@` characters][1]. By default, our go.lsp.dev package follows Go's semantics and does not escape these characters. (I plan to report this inconsistency upstream to go.lsp.dev, but don't expect a quick response.)

Fixes bufbuild/vscode-buf#569.

[1]: https://github.com/microsoft/vscode-uri/blob/65786c7aef8aa1d142fedfde76073cc3549736d2/src/uri.ts#L462


stefanvanburen:
Using this `buf` locally, I can get hover info / semantic tokens / all that jazz in VS Code:

image

PR #4338, opened 2026-02-17
Fix breaking against matching on new modules

emcfarlane:
This fixes `buf breaking` when adding a new module. Before it could incorrectly filter images based on imported files. For example with module A and B with a.proto and b.proto a new module C could be added with a dependency on A. If this module was ordered before A and B (by opaqueID), it would capture module A, as C would compare `a.proto`, and cause a breaking change between them.


stefanvanburen (approved):

stefanvanburen (approved):
thanks for adding the test!

PR #4341, opened 2026-02-18
Add LSP hover support for CEL documentation

stefanvanburen:
This adds hover support to LSP for CEL expressions, adding documentation for the various functions/operators/etc. when writing raw CEL using protovalidate. _Most_ of the documentation comes from cel-go; we do have tests over the output so we'll know when things change.

I think we can follow up with some links to either protovalidate.com or celbyexample.com in the hovers, similar to our links to protobuf.com and buf.build for protobuf cases.

### DIFF
--- a/private/buf/buflsp/cel.go
+++ b/private/buf/buflsp/cel.go
@@ -26,15 +26,12 @@ import (
 // isCELKeyword returns true if the identifier is a CEL reserved keyword.
 // See https://github.com/google/cel-spec/blob/master/doc/langdef.md#syntax
 func isCELKeyword(name string) bool {
-	keywords := map[string]bool{
-		// Literals
-		"true":  true,
-		"false": true,
-		"null":  true,
-		// Special identifiers
-		"this": true,
+	switch name {
+	case "true", "false", "null", // literals
+		"this": // special identifier
+		return true
 	}
-	return keywords[name]
+	return false
 }
 
 // isCELMacroFunction returns true if the function name is a CEL macro (comprehension or special function).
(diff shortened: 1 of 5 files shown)


private/buf/buflsp/cel.go
stefanvanburen, line 1: tried to consolidate the shared cel bits from semantic tokens and hover here.

private/buf/buflsp/hover_cel_test.go
```diff
@@ -0,0 +1,471 @@
...
+		line     uint32 // 0-indexed
+		char     uint32 // Character offset where the token starts
+		expected string // Full expected hover content
+	}{
+		// Keywords
+		{
+			name:     "keyword: this",
+			line:     19,
+			char:     17,
+			expected: "**Special variable**\n\nRefers to the current message or field being validated.\n\nIn field-level rules, `this` refers to the field value.\nIn message-level rules, `this` refers to the entire message.",
```
stefanvanburen, line 46: good to look through `expected` here for a sense of how these look. Could certainly make some subjective improvements, but they feel alright to me.

For context, this test should go over the comprehensive list of cel hovers, so that we lock in the output. 🤷

private/buf/buflsp/hover_cel.go
```diff
@@ -0,0 +1,951 @@
...
+						fmt.Fprintf(&result, ": %s", child.Description)
+					}
+				}
+			}
+		} else {
+			// Children are examples (e.g., for macros)
+			result.WriteString("\n\n**Examples**:")
+			for _, child := range doc.Children {
+				if child.Description != "" {
+					fmt.Fprintf(&result, "\n```cel\n%s\n```", child.Description)
```
stefanvanburen, line 833: I'd love to get `cel` supported by tree-sitter in nvim so these docs in markdown show up with syntax highlighting :)

private/buf/buflsp/cel.go
```diff
@@ -0,0 +1,173 @@
...
+			continue
+		}
+
+		// Found opening quote - scan the string content
+		i++
+		for i < len(spanText) && spanText[i] != '"' {
+			// Check if we've found the token start
+			if celPos == celStart {
+				fileStart := multilineSpan.Start + i
+				fileEnd := fileStart + (celEnd - celStart)
```
emcfarlane: Little worried we have a mismatch of CEL offset to LSP utf-16 here. Does it matter?

private/buf/buflsp/hover_cel.go
emcfarlane: that'd be neat

private/buf/buflsp/cel.go
stefanvanburen: good call; we had bugs w/ both unicode and multi-line CEL expressions; added tests and fixes in the last two commits.

PR #4345, opened 2026-02-20
Replace : with %3A after file:/// scheme in URIs to match VS Code's LSP client

yjoer:
Closes https://github.com/bufbuild/vscode-buf/pull/575

### DIFF
--- a/private/buf/buflsp/diagnostics_test.go
+++ b/private/buf/buflsp/diagnostics_test.go
@@ -141,7 +141,7 @@ func setupLSPServerWithDiagnostics(
 	})
 
 	testWorkspaceDir := filepath.Dir(testProtoPath)
-	testURI := uri.New(testProtoPath)
+	testURI := buflsp.FilePathToURI(testProtoPath)
 	var initResult protocol.InitializeResult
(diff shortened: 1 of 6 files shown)


stefanvanburen:
hey @yjoer, looks pretty good to me (I might just push a commit w/ some tests after).

Just to confirm, have you run this buf version locally and confirmed it fixes the issues? (You should be able to install it locally with `make installbuf` from this repo.)

yjoer:
There isn't make on Windows :(

I build it with `go build -a -ldflags "-s -w" -trimpath -buildvcs=false -o buf.exe ./cmd/buf`. The version of the binary matches the one shown at the status bar. It should be using the binary I built. Before this, I reverted the VS Code extension to the one from the marketplace and made sure the highlighting isn't applying.

```
> buf --version
1.65.1-dev
```

https://github.com/user-attachments/assets/14c61c99-f330-4f48-a056-dc90af07a521

stefanvanburen:
@yjoer thanks for confirming! just fixing up the tests and we should be able to land this.

stefanvanburen:
Just as a note for later, this relates to #4325.

stefanvanburen:
@yjoer thanks for your contribution; this will go out in the next `buf` release!

PR #4347, opened 2026-02-20
Fix test flake

stefanvanburen:
We saw a test flake in a CI run on main; this fixes that issue.

Tested with:

```console
$ go test -run TestCodeAction_Deprecate -count=1000 ./private/buf/buflsp/
```

Ref: https://github.com/bufbuild/buf/actions/runs/22233794677/attempts/1

### DIFF
--- a/private/buf/buflsp/image.go
+++ b/private/buf/buflsp/image.go
@@ -85,7 +85,9 @@ func buildImage(
 	var diagnostics []protocol.Diagnostic
 	compiled, err := compiler.Compile(ctx, path)
 	if err != nil {
-		logger.WarnContext(ctx, "error building image", slog.String("path", path), xslog.ErrorAttr(err))
+		if !errors.Is(err, context.Canceled) {
+			logger.WarnContext(ctx, "error building image", slog.String("path", path), xslog.ErrorAttr(err))
+		}
 		var errorWithPos reporter.ErrorWithPos
 		if errors.As(err, &errorWithPos) {
 			diagnostics = []protocol.Diagnostic{newDiagnostic(errorWithPos, false, opener, logger)}


private/buf/buflsp/file.go
```diff
@@ -1079,7 +1079,7 @@ func (f *file) RunChecks(ctx context.Context) {
 			); err != nil {
 				var fileAnnotationSet bufanalysis.FileAnnotationSet
 				if !errors.As(err, &fileAnnotationSet) {
-					if errors.Is(err, context.Canceled) {
+					if errors.Is(err, context.Canceled) || ctx.Err() != nil {
```
stefanvanburen, line 1082: related to this: https://github.com/pluginrpc/pluginrpc-go/pull/17

private/buf/buflsp/file.go
```diff
@@ -1062,7 +1062,7 @@ func (f *file) RunChecks(ctx context.Context) {
 	}
 
 	const checkTimeout = 30 * time.Second
-	ctx, cancel := context.WithTimeout(context.WithoutCancel(ctx), checkTimeout)
+	ctx, cancel := context.WithTimeout(f.lsp.connCtx, checkTimeout)
```
emcfarlane, line 1065: Why does `context.WithoutCancel(ctx)` not work here?

stefanvanburen: the issue is that we're logging after the tests are complete form the goroutine; the `ctx` here doesn't get canceled when the tests stop?

private/buf/buflsp/buflsp.go
```diff
@@ -61,6 +61,17 @@ func Serve(
 	logger.Info("starting LSP server")
 
 	conn := jsonrpc2.NewConn(stream)
+	// connCtx is a context scoped to the connection's lifetime. It is cancelled
+	// when the connection is done (or when ctx is cancelled), so that background
+	// goroutines (e.g. RunChecks) do not outlive the connection.
+	connCtx, connCancel := context.WithCancel(context.Background())
```
emcfarlane, line 67: This feels like a test context issue? Why is the test canceling the conn context before the end of the test?

private/buf/buflsp/file.go
emcfarlane: This still an issue: https://github.com/bufbuild/buf/actions/runs/22236404191/job/64328996626#step:5:132 ?
Right so the logging is firing as part of the *testing.T handler?

private/buf/buflsp/buflsp.go
stefanvanburen: hm, I guess I'm not seeing it in the test; maybe something in the ordering in [buflsp_test](https://github.com/bufbuild/buf/blob/main/private/buf/buflsp/buflsp_test.go)?

At the end of the day, we're disconnecting from the request context, but we're not tying the goroutine context to the running server in any way, either?

stefanvanburen:
For context, the core fix for the flake is just [avoiding logging](https://github.com/bufbuild/buf/pull/4347/commits/ce2d395b05ebb5984ce5e312e52924f49ab92bbe) during the goroutine after the `err`, because the `err` can be `context.Canceled` from the test ending. But this also structures our shutdown a little better, and wraps the handler in the `protocol.CancelHandler`, which should handler some of the annoying `$/cancelRequest` logging we see in the LSP.

PR #4357, opened 2026-02-25
Forward the input image resolver to protovalidate so it can support predefined rule

jeremie-stripe:
This simple change allow the `--validate` option of `buf convert` to locate custom rules (as documented in https://protovalidate.com/schemas/predefined-rules/) as long as they are included in the input (e.g. via a protoset).

### DIFF
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -751,7 +751,9 @@ func (c *controller) GetMessage(
 	}
 	var validator protoyaml.Validator
 	if functionOptions.messageValidation {
-		protovalidateValidator, err := protovalidate.New()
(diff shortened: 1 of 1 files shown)


stefanvanburen:
hey @jeremie-stripe, looks good to me — could you just change your PR title's first word to not end in "s", please? (Sorry, it's a [silly regex](https://github.com/bufbuild/base-workflows/blob/ed5fd209ec01d4db149246a7dd7db5193354cf3c/.github/workflows/pr-title.yaml#L25-L29) but it helps us keep in the imperative tone.)

jeremie-stripe:
@stefanvanburen did "Pass" --> "Forward", hopefully that does the trick!

PR #4358, opened 2026-02-26
Add LSP completion for CEL expressions

stefanvanburen:
This adds contextual completion for CEL expressions for protovalidate, in the same spirit as our completion approach for protobuf.

I strongly suggest reviewing this by also installing this branch locally and trying out the completion in the testcases, to see how it feels to use locally.

### DIFF
--- a/private/buf/buflsp/completion_cel.go
+++ b/private/buf/buflsp/completion_cel.go
@@ -1,0 +1,850 @@
+// Copyright 2020-2025 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


private/buf/buflsp/completion_cel.go
```diff
@@ -0,0 +1,850 @@
...
+// It finds the leftmost (outermost) such macro in celContent so that callers can
+// process comprehension layers from outer to inner by calling it in a loop.
+// Returns the range expression, the iteration variable name, and the body content
+// (the expression fragment after the separating comma), or ok=false if not matched.
+// Examples:
+//
+//	"this.items.filter(addr, addr."  → ("this.items", "addr", "addr.", true)
+//	"items.all(x, x.name"            → ("items", "x", "x.name", true)
+//	"this.filter("                   → ("", "", "", false)  — no comma yet
+func celParseIteratorBody(celContent string) (rangeExpr, iterVar, bodyContent string, ok bool) {
```
emcfarlane, line 297: In the furture we should be able to use protocompile here to parse these CEL expressions right?

stefanvanburen: not sure; are we planning to add CEL support there? (there's probably some stuff in cel.go or throughout this package that could be useful to upstream.)

stefanvanburen:
> Minor issue: completions for `in` complete to a function call `in()`, should just be membership test (`this in that`).

cool, I will follow up with a fix for that. thanks for checking it out!

PR #4363, opened 2026-03-02
Add support for Go features (features.(pb.go).*) in buf

NexZhu:
This commit adds support for configuring Go features in buf.yaml for managed mode code generation and breaking change detection.

Changes:
- Add ResolveGoFeature and ResolveGoFeatureForEnum functions in customfeatures
- Add FIELD_SAME_GO_STRIP_ENUM_PREFIX breaking change rule
- Add FileOptionGoApiLevel for managed mode (features.(pb.go).api_level)
- Add test for ResolveGoFeature

The Go features use the external google.golang.org/protobuf/types/gofeaturespb package to avoid namespace conflicts.

Note: The strip_enum_prefix breaking rule test is skipped as it requires edition 2024 which is not yet fully supported in buf.

### DIFF
--- a/private/bufpkg/bufcheck/breaking_test.go
+++ b/private/bufpkg/bufcheck/breaking_test.go
@@ -379,6 +379,18 @@ func TestRunBreakingFieldSameJavaUTF8Validation(t *testing.T) {
 	)
 }
 
+func TestRunBreakingFieldSameGoStripEnumPrefix(t *testing.T) {
+	t.Skip("TODO: enable when edition 2024 is supported")
+	t.Parallel()
+	testBreaking(
+		t,
+		"breaking_field_same_go_strip_enum_prefix",
+		bufanalysistesting.NewFileAnnotation(t, "1.proto", 8, 1, 8, 7, "FIELD_SAME_GO_STRIP_ENUM_PREFIX"),
+		bufanalysistesting.NewFileAnnotation(t, "1.proto", 15, 1, 15, 12, "FIELD_SAME_GO_STRIP_ENUM_PREFIX"),
+		bufanalysistesting.NewFileAnnotation(t, "1.proto", 22, 1, 22, 17, "FIELD_SAME_GO_STRIP_ENUM_PREFIX"),
+	)
+}
+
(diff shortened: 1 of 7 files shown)


doriable:
Thank you for submitting this change -- I think this request seems very reasonable. Internally, our current priority is getting Editions 2024 support up and running with our new compiler, and would prefer to review this when it can be completed with the required tests. So transparently, I won't be able to review this until that body of work is complete, but if it is okay with you, I am happy to keep this PR open and will provide feedback once we are ready to accept this request.

NexZhu:
> Thank you for submitting this change -- I think this request seems very reasonable. Internally, our current priority is getting Editions 2024 support up and running with our new compiler, and would prefer to review this when it can be completed with the required tests. So transparently, I won't be able to review this until that body of work is complete, but if it is okay with you, I am happy to keep this PR open and will provide feedback once we are ready to accept this request.

Sure

themavik (commented):
`ResolveGoFeatureForEnum` plus `handleBreakingFieldSameGoStripEnumPrefix` wires `(pb.go).strip_enum_prefix` into `FIELD_SAME_GO_STRIP_ENUM_PREFIX` in the same pattern as the other feature rules.

`TestRunBreakingFieldSameGoStripEnumPrefix` stays skipped until edition 2024 lands; maybe link the enabling issue in the skip comment so it does not rot.

AdrianDuan:
@doriable Any updates about this PR? It seems that buf supports edition 2024 now. I'm looking forward to use this new feature in managed mode

doriable:
Apologies for the late response on this, unfortunately after reviewing these changes with the team, we have decided to not support the changes in this PR:
- We are not currently looking to support `FeatureSet` as a part of managed mode, and we should not implement partial support for this
- The breaking change rule implemented here has not actually been added to rule spec, so this would not actually run at `buf breaking` time, but also, we are not looking to add this to our existing rules. If you'd like this custom check, this rule can be implemented as a [custom check plugin](https://buf.build/docs/bsr/checks/plugins/) that you can run with `buf breaking`, there is a [tutorial for writing and running a plugin](https://buf.build/docs/bsr/checks/plugins/tutorial/)

PR #4361, opened 2026-02-27
protoc_builtin: Support `--rbs_out`

sorah:
protoc v34.0 gained `--rbs_out`, type signature file for Ruby.

- https://github.com/protocolbuffers/protobuf/releases/tag/v34.0
- https://github.com/protocolbuffers/protobuf/pull/15633

PR #4382, opened 2026-03-10
Add support for `--rbs_out`

stefanvanburen:
Originally in #4361, but we need to add our typical version checks and testing.

Ref: https://github.com/protocolbuffers/protobuf/releases/tag/v34.0
Ref: protocolbuffers/protobuf#15633

Supersedes #4361.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Add support for `--rbs_out` as a `protoc_builtin` plugin (requires protoc v34.0+).
 
 ## [v1.66.1] - 2026-03-09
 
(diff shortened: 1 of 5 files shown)


stefanvanburen:
@doriable had mentioned having `claude` take a look at this; my prompt was:

> look at `gh pr view 4361`; I think we want to support rbs_out, but first we need to upgrade the version of protoc we're testing (maybe this is in the Makefile somewhere)? Can we do that so we make sure we're testing it?

And it basically one-shotted this, although it also bumped the `PROTOC_VERSION` in the Makefile, which I thought about more and didn't think was necessary, so I followed up with:

> do we actually need to bump the PROTOC_VERSION for this? do tests fail without it?

which caused it to revert the Makefile changes, and now we're here. (I don't see any obvious integration testing involving calling `protoc` with these args, so unless I'm missing something this should be how the others work and are tested?)

doriable:
So the answer is yes and no, lol ^^"

Basically, `PROTOC_VERSION` ensures that this is the `protoc` version used when developing in `bufbuild/buf` through `make` and it also updates the WKTs. So in this case, we should update this version if we plan on writing tests for it, but it's not strictly necessary to update this version to support the features from the latest `protoc` (since it would invoke the Ruby codegen through the user's local `protoc` as a codegen plugin, effectively).

PR #4387, opened 2026-03-12
Switch docker APIs to use new moby/client

stefanvanburen:
Gets us away from the +incompatible dependency (and drops a few dependencies in the process).


stefanvanburen:
(I need to test this more fully before taking it out of draft.)

stefanvanburen:
ok, in [plugins](https://github.com/bufbuild/plugins) I was able to run:

```console
$ make push BUF_PLUGIN_PUSH_ARGS="--override-remote bufbuild.internal --visibility=public" PLUGINS="apple/swift:v1.35.0 grpc/swift-protobuf:v2.2.0"
```

Which runs `buf beta registry plugin push` under the hood and hits these codepaths. Also passed some of our internal tests.

private/bufpkg/bufremoteplugin/bufremoteplugindocker/docker.go
```diff
@@ -242,35 +221,6 @@ func (d *dockerAPIClient) Close() error {
 	return d.cli.Close()
 }
 
-func (d *dockerAPIClient) negotiateVersion(ctx context.Context) error {
```
emcfarlane, line 245: Mind just commenting why this is no longer needed? (On GitHub not code). This now handled by the library?

stefanvanburen: I was curious about that myself while working on this with `claude` and here's what it said:

> The old docker/docker/client required WithAPIVersionNegotiation() as an opt-in — without it, the client used a hardcoded fallback version. The manual Ping + NegotiateAPIVersionPing dance was the workaround: ping to discover what the server supports, then tell the client to use that version.

> The new moby/moby/client flips this: negotiation is opt-out. On the first real API call, the client automatically pings and adjusts. WithAPIVersion(...) is what disables it (which is why tests still work — they set a fixed version deliberately).

And here's some source links, where the negotiate option is deprecated: https://github.com/moby/moby/blob/c2f80b072ca4754a19b721e9a4d75ab851cac925/client/client_options.go#L340-L341, and to the client behavior: https://github.com/moby/moby/blob/c2f80b072ca4754a19b721e9a4d75ab851cac925/client/client.go#L176-L180

emcfarlane: neat, thanks. Nice that we can get rid of all of this

doriable: Based on what I'm seeing in moby/client -- API negotiation is enabled by default?

doriable: Lol, sorry, comment at the same time, I think ^^"

evankanderson:
Hello, there is a Docker CVE announced that this PR resolves: https://github.com/moby/moby/security/advisories/GHSA-v23v-6jw2-98fq 

Any chance of a v1.66.2?

doriable:
Yep, we're planning on making a new release tomorrow in the morning, EST.

Issue #3126, opened 2024-07-04
Make buf generate --clean a per-plugin option that does a "smart" clean


bufdev:
See https://github.com/bufbuild/buf/pull/3124#issuecomment-2209407669 for an explanation.

doriable:
> For zip/tar files: sum the file beforehand, generate the new file, if there is a diff, overwrite, if the filename changed (somehow), delete the old one
> For directories: sum each file beforehand, generate the new files, for each file that no longer exists, delete, for each file that changed, overwrite

There are a few things happening here and I wonder if some of this should be separated out:
- For files that have not changed, ignore them
- For files that no longer exist, delete them
- For files where contents have changed, overwrite them
- For net new files, add them

For the first one, files that have not changed being ignored, would it be reasonable to implement this as the default behaviour of `buf generate`, regardless of the `--clean` flag/config? This seems like it would be backwards compatible.

Then `--clean` would only need to handle deleting files that no longer exist/have been renamed, overwriting files with new/different content, and adding new any new files.

PR #4396, opened 2026-03-16
Skip writing unchanged generated files to preserve mtimes

iainmcgin:
Contributes towards #3126.

## Summary

`buf generate` previously wrote output files unconditionally, touching mtimes even when content was identical to what was already on disk. For mtime-based build systems (cargo, make, ninja, file-watching dev servers), a no-op `buf generate` would cascade into unnecessary downstream rebuilds.

The protoc plugin protocol gives plugins no visibility into the output directory—they emit `File{name, content}` on stdout and the invoking tool handles filesystem writes—so this fix lives in buf's write path rather than in individual plugins.

## Approach

Replaces the unconditional `storage.Copy` in the directory output path (`bufprotopluginos.writeDirectory`) with a compare-then-write loop. For each generated file:

- Read existing content from the OS bucket
- If it matches the new content, skip the write (mtime preserved)
- Otherwise write as before

Read errors during comparison (file doesn't exist, permissions, etc.) fall through to write, so worst-case behavior is identical to before.

## Scope

- **Directory output only.** zip/jar output continues to write unconditionally—comparing archive content would require unzip-compare-rezip, which isn't motivated by the use cases in #3126 and can be a follow-up if needed.
- **Default-on, no flag.** Per the discussion in #3126, this is backwards-compatible: it's never wrong to skip writing identical content, and the cost is one extra read per existing file (zero cost on first run).
- **`--clean` interaction is already correct.** Clean deletes first, so comparison finds nothing and writes everything.

## Testing

- 4 unit tests in `response_writer_test.go` covering unchanged/changed/new/mixed file scenarios
- 1 integration test in `generate_test.go` that runs `buf generate` twice and verifies the output mtime is preserved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add support for `--rbs_out` as a `protoc_builtin` plugin (requires protoc v34.0+).
 - Add relevant links from CEL LSP hover documentation to either <celbyexample.com> or <protovalidate.com>
+- `buf generate` now skips writing output files when the content matches what's already on disk, preserving modification times for mtime-based build systems like cargo and make.
 
 ## [v1.66.1] - 2026-03-09
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Add support for `--rbs_out` as a `protoc_builtin` plugin (requires protoc v34.0+).
 - Add relevant links from CEL LSP hover documentation to either <celbyexample.com> or <protovalidate.com>
-- `buf generate` now skips writing output files when the content matches what's already on disk, preserving modification times for mtime-based build systems like cargo and make.
+- Skip writing unchanged output files in `buf generate` to preserve modification times
 
 ## [v1.66.1] - 2026-03-09
 
--- a/cmd/buf/internal/command/generate/generate_test.go
+++ b/cmd/buf/internal/command/generate/generate_test.go
@@ -25,6 +25,7 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Add support for `--rbs_out` as a `protoc_builtin` plugin (requires protoc v34.0+).
 - Add relevant links from CEL LSP hover documentation to either <celbyexample.com> or <protovalidate.com>
+- `buf generate` now skips writing output files when the content matches what's already on disk, preserving modification times for mtime-based build systems like cargo and make.
```
emcfarlane:
```suggestion
- Skip writing unchanged output files in `buf generate` to preserve modification times
```

emcfarlane (approved):
Thanks for this contribution! The skip-unchanged behavior is a good improvement.

However, this doesn't fully close #3126. That issue is about making --clean a per-plugin "smart" clean, including stale file deletion. This PR is a useful step toward that, but the core of the issue remains open.

Could you update the PR description to say "Contributes to #3126" instead of "Closes #3126" so the issue stays open?

doriable (approved):
Looks good to me, thanks for putting this up! +1 to Ed's comments on fixing the PR description.

iainmcgin:
Changed to "contributes towards" as requested

jtszalay:
> this is backwards-compatible

image

We had a workflow that detected files with a pattern and recorded the mtimes. Any files that _didn't_ get updated mtimes get deleted. This change has now resulted in our workflow deleting all files when no changes occur. We did this because we have manually written files in the output directories that we don't want deleted by buf.

Edit: we are trying to work around this on our own. Just caught us by surprise as we updated.

PR #4397, opened 2026-03-17
Escape HTML in generated CLI docs

stefanvanburen:
I noticed on our docs, the generated CLI descriptions for some flags on `buf curl` were mis-rendering because of describing arguments like `<input>`. We need to HTML escape these.

Additionally, remove the `include-front-matter` flag, which is no longer used (as our docs site is not a Docusaurus site any longer).

Lastly, unconditionally add the `title:` frontmatter for now, which should help when rendering sub-flags: instead of rendering "Dep" or "Beta" (corresponding to `buf dep` and `buf beta`, respectively), we'll render the actual `buf beta` name, which should read better on the docs site.

Ref: https://buf.build/docs/reference/cli/buf/curl/#schema

### DIFF
--- a/private/bufpkg/bufcobra/markdown.go
+++ b/private/bufpkg/bufcobra/markdown.go
@@ -155,10 +155,6 @@ func generateMarkdownPage(
 	return err
 }
 
-func websitePageIDForCommand(cmd *cobra.Command) string {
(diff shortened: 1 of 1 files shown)


stefanvanburen:
For example, here's the current status, where the literal `<input>` text becomes a text input (https://buf.build/docs/reference/cli/buf/curl/#schema):

image

PR #4398, opened 2026-03-17
Use overlay bucket on buf.lock updates

emcfarlane:
This updates `buf dep update` to use an overlay bucket to update the `buf.lock` file before persisting to disk. A workspace option is added for the overlay configuration. Completes the TODOs. No change in behaviour.


emcfarlane:
@themavik the plugin and policy updates aren't validated during a workspace build in the same way the `buf build` command would fail. We may be able to add better validation here, but removed to simplify for now.

PR #4402, opened 2026-03-19
Add support for URL completions in `buf curl`

### LINKED ISSUE #2044: feature request: tab completions for `buf curl`
Thanks for creating `buf curl`! I think it's a very promising tool that will make it a lot easier to work with gRPC and Connect APIs. 

An "obvious" and killer feature that's missing from `buf curl` is tab completion of services and methods. It would make for a fantastic developer experience if `buf curl https://my-api.com/<TAB>` could list out the available endpoints and methods. 

A few ideas for how this could be done: 

1. Check and see if we're in a folder (or sub-folder) that contains a Buf module or workspace. In that case the module could be built, and the endpoints and methods enumerated. 
2. Attempt to do reflection on the written URL (and with any header flags, if passed into the command). 
3. If the `--schema` flag is passed, inspect that to find the data. 

I don't know how `buf curl` is typically used. But for _my personal usage_, I'm mostly using it against local services I'm actively working on, for manual testing and debugging. Prior to `buf curl` being developed, I've used a Fish script called `bufcurl` that wraps `grpcurl` and uses the approach I described in point 1. I've implemented tab completion for that (it only lists services, I haven't gotten around to implementing the methods). Pasting this below, for inspiration/context on what I'm thinking. 

A final note: I'm configuring my `bufcurl` script through environment variables. In the folder containing services I'm working on I typically set these environment variables using an environment file. I think this is crucial to providing the *best* possible DX, as it simplifies the command needed to be invoked. 

<details>
<summary>The script itself</summary>
```fish
function bufcurl
    if test -n "$BUFCURL_PLAINTEXT" 
        set plaintext_arg '-plaintext'
    end

    # Check if we get passed a data arg, with
    # a trailing `=`
    if string match -q -- "*-d=*" "$argv"
        set idx (contains --index -- -d= $argv)
        set data_args $argv[$idx] 

        # only do it once, because it's the same arg 
        set --erase argv[$idx]

    # Check if we get a data arg, without 
    # trailing `=`
    else if contains -- -d $argv
        set idx (contains --index -- -d $argv)
        set data_args $argv[$idx] $argv[(math "$idx + 1")]

        # do it twice, both flag and value
        set --erase argv[$idx]
        set --erase argv[$idx]

    end

    if test -n "$BUFCURL_HEADER" 
        set header_arg -H="$BUFCURL_HEADER"
    end

    if test -n "$BUFCURL_VERBOSE"
        set verbose_arg "-v"
    end

    # If we get passed a target, this overwrites 
    # env var target. 
    if string match -q -- "*localhost*" "$argv" 
        set --erase BUFCURL_TARGET
    end

    set working_dir (pwd)
    while not test -e buf.yaml -o -e buf.yml -o -e buf.work.yaml -o -e buf.work.yml 
        if test (pwd) = "/"
            echo could not find Buf config! >&2
            cd $working_dir
            return 1
        end

        __bufcurl_log pwd=(pwd), going backwards
        
        cd ..
    end

    __bufcurl_log found buf_dir=(pwd)
    set buf_dir (pwd)
    cd $working_dir

    __bufcurl_log plaintext: $plaintext_arg
    __bufcurl_log target: $BUFCURL_TARGET
    grpcurl $verbose_arg $plaintext_arg $header_arg $data_args -protoset (buf build $buf_dir -o - | psub) $BUFCURL_TARGET $argv
end
</details>

<details>
<summary>The tab completions</summary>
```fish
❯ cat $HOME/.config/fish/completions/bufcurl.fish
function __bufcurl_list_services
    buf build --exclude-source-info --output -#format=json | \
        jq --raw-output --compact-output \
        '.file | map ({service: .service, package: .package} | select(.service) | {service: .service[0].name, package: .package} | .package +"."+ .service  ) | .[]'
end

# Not implemented...
function __bufcurl_list_methods --argument-names service

end

function __bufcurl_complete
    # no buf setup!
    if not test -e buf.yaml && not test -e buf.work.yaml
        return 
    end

    set -l cmd (commandline -poc)
    set -e cmd[1]

    if not __fish_seen_subcommand_from (__bufcurl_list_services)
        __bufcurl_list_services
        return 
    end

    __bufcurl_list_methods



 
end

complete -c bufcurl --no-files -a "(__bufcurl_complete)"
```
</details>

stefanvanburen:
First mentioned in #2044, which is long something I've wanted. This adds support for completing URLs based on the available RPCs, which either comes from the existing `--schema` flag or gRPC reflection.

The completion itself works somewhat similar to LSP completion, in that it attempts to complete as far as it can before giving the user an option between the remaining values to disambiguate. It will complete up to the entire service URL, and then provide completions for RPCs within the service (if there are multiple).

This requires one fix upstream in app-go so that subcommands that define `ModifyCobra` actually run: bufbuild/app-go#5. We'll want to land upstream to `main` before landing this.

Open to suggestions on the completion UX. It feels fairly natural to me currently, but I'm sure there are edge cases.

Future work here could include better completions for `--schema` values, either using local directories or BSR modules, and completions for `--data` values (if we know the schema and the specific RPC targeted by the URL, we know the shape of the JSON for the `--data` flag). Also, following this pattern of using ModifyCobra, other commands could be made to have better contextual completion.

Also fixes the buf curl help examples to have consistent indentation.

Resolves #2044.

### DIFF
--- a/cmd/buf/internal/command/curl/curl.go
+++ b/cmd/buf/internal/command/curl/curl.go
@@ -196,29 +196,8 @@ exit code that is the gRPC code, shifted three bits to the left.
 				return run(ctx, container, flags)
 			},
 		),
-		BindFlags: flags.Bind,
-		ModifyCobra: func(cmd *cobra.Command) error {
-			if err := errors.Join(
-				cmd.RegisterFlagCompletionFunc(
-					protocolFlagName,
-					cobra.FixedCompletions([]string{
-						connect.ProtocolConnect,
-						connect.ProtocolGRPC,
-						connect.ProtocolGRPCWeb,
-					}, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder),
-				),
-				cmd.RegisterFlagCompletionFunc(
-					reflectProtocolFlagName,
-					cobra.FixedCompletions(bufcurl.AllKnownReflectProtocolStrings, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder),
-				),
-				cmd.RegisterFlagCompletionFunc(headerFlagName, cobra.NoFileCompletions),
-				cmd.RegisterFlagCompletionFunc(reflectHeaderFlagName, cobra.NoFileCompletions),
-			); err != nil {
-				return err
-			}
-			cmd.ValidArgsFunction = completeURL
-			return nil
-		},
+		BindFlags:   flags.Bind,
+		ModifyCobra: completeCurlCommand,
 	}
 }
 
@@ -756,6 +735,29 @@ func (f *flags) getTLSConfig(authority string, printer verbose.Printer) (*tls.Co
 		HTTP2PriorKnowledge: f.HTTP2PriorKnowledge,
 		HTTP3:               f.HTTP3,
 	}, authority, printer)
+}
+
+func completeCurlCommand(cmd *cobra.Command) error {
+	if err := errors.Join(
+		cmd.RegisterFlagCompletionFunc(
+			protocolFlagName,
+			cobra.FixedCompletions([]string{
+				connect.ProtocolConnect,
+				connect.ProtocolGRPC,
+				connect.ProtocolGRPCWeb,
+			}, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder),
+		),
+		cmd.RegisterFlagCompletionFunc(
+			reflectProtocolFlagName,
+			cobra.FixedCompletions(bufcurl.AllKnownReflectProtocolStrings, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder),
+		),
+		cmd.RegisterFlagCompletionFunc(headerFlagName, cobra.NoFileCompletions),
+		cmd.RegisterFlagCompletionFunc(reflectHeaderFlagName, cobra.NoFileCompletions),
+	); err != nil {
+		return err
+	}
+	cmd.ValidArgsFunction = completeURL
+	return nil
 }
 
 func promptForPassword(ctx context.Context, container app.Container, prompt string) (string, error) {
--- a/cmd/buf/internal/command/curl/curl_completion_test.go
+++ b/cmd/buf/internal/command/curl/curl_completion_test.go
@@ -137,13 +137,13 @@ func TestFlagCompletions(t *testing.T) {
 	protocolFn, _ := cmd.GetFlagCompletionFunc(protocolFlagName)
 	require.NotNil(t, protocolFn)
 	protocols, directive := protocolFn(cmd, nil, "")
-	assert.Equal(t, cobra.ShellCompDirectiveNoFileComp, directive)
+	assert.Equal(t, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder, directive)
 	assert.ElementsMatch(t, []string{connect.ProtocolConnect, connect.ProtocolGRPC, connect.ProtocolGRPCWeb}, protocols)
 
 	reflectFn, _ := cmd.GetFlagCompletionFunc(reflectProtocolFlagName)
 	require.NotNil(t, reflectFn)
 	reflectProtocols, directive := reflectFn(cmd, nil, "")
-	assert.Equal(t, cobra.ShellCompDirectiveNoFileComp, directive)
+	assert.Equal(t, cobra.ShellCompDirectiveNoFileComp|cobra.ShellCompDirectiveKeepOrder, directive)
 	assert.ElementsMatch(t, []string{"grpc-v1", "grpc-v1alpha"}, reflectProtocols)
 }
 


stefanvanburen:
ok, as a quick demo gif:
curl-completion

using the following [vhs tape](https://github.com/charmbracelet/vhs):

```vhs
Output curl-completion.gif

Set Shell fish
Set FontSize 14
Set Width 900
Set Height 400
Set Padding 16
Set Theme "Catppuccin Mocha"

Env PATH "/Users/stefanvanburen/.cache/buf/Darwin/arm64/gobin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin"

Hide
Sleep 1s
Type "source ~/.config/fish/completions/buf.fish"
Enter
Sleep 200ms
Show
Ctrl+L
Sleep 200ms

# --protocol flag: static completions (connect, grpc, grpcweb)
Type "buf curl --protocol "
Sleep 500ms
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# --reflect-protocol flag: static completions (grpc-v1, grpc-v1alpha)
Type "buf curl --reflect-protocol "
Sleep 500ms
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# URL via reflection — Tab 1: unambiguous prefix collapses to full service name
Type "buf curl https://demo.connectrpc.com/"
Sleep 500ms
Tab
Sleep 4s

# Tab 2: method completions (Converse, Introduce, Say)
Tab
Sleep 2s

# Select a method by typing a prefix and completing
Type "Say"
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# Same via --schema (no running server needed) — Tab 1: service name
Type "buf curl --schema buf.build/connectrpc/eliza https://demo.connectrpc.com/"
Sleep 500ms
Tab
Sleep 4s

# Tab 2: method completions
Tab
Sleep 2s
Ctrl+C
```

stefanvanburen:
Quick UX update: in https://github.com/bufbuild/buf/pull/4402/commits/20a16eaef5c57491275ae66d5d63c8e9d0311d2e I added descriptions to the RPC completions about where the schema is coming from (local module, reflection, --schema flag), and also added some basic error handling.

Updated gif:

curl-completion

via this tape:

```vhs
Output curl-completion.gif

Set Shell fish
Set FontSize 14
Set Width 900
Set Height 400
Set Padding 16
Set Theme "Catppuccin Mocha"

Env PATH "/Users/stefanvanburen/.cache/buf/Darwin/arm64/gobin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin"

Hide
Sleep 1s
Type "source ~/.config/fish/completions/buf.fish"
Enter
Sleep 200ms
Show
Ctrl+L
Sleep 200ms

# --protocol flag: static completions (connect, grpc, grpcweb)
Type "buf curl --protocol "
Sleep 500ms
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# --reflect-protocol flag: static completions (grpc-v1, grpc-v1alpha)
Type "buf curl --reflect-protocol "
Sleep 500ms
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# URL via reflection — Tab 1: unambiguous prefix collapses to full service name
Type "buf curl https://demo.connectrpc.com/"
Sleep 500ms
Tab
Sleep 4s

# Tab 2: method completions (Converse, Introduce, Say)
Tab
Sleep 2s

# Select a method by typing a prefix and completing
Type "Say"
Tab
Sleep 2s
Ctrl+C
Sleep 1s

# Same via --schema (no running server needed) — Tab 1: service name
Type "buf curl --schema buf.build/connectrpc/eliza https://demo.connectrpc.com/"
Sleep 500ms
Tab
Sleep 4s

# Tab 2: method completions
Tab
Sleep 2s
Ctrl+C
```

cmd/buf/internal/command/curl/curl.go
```diff
@@ -1193,3 +1196,306 @@ func makeHTTP3RoundTripper(f *flags, authority string, printer verbose.Printer)
...
+
+	// 1. Explicit --schema flag.
+	schemas, _ := cmd.Flags().GetStringSlice(schemaFlagName)
+	if len(schemas) > 0 {
+		return completeURLFromSchema(ctx, schemas, baseURL, rawPath, "--schema")
+	}
+
+	// 2. Live server reflection.
+	isSecure := parsed.Scheme == "https"
+	if httpClient, ok := makeCompletionHTTPClient(cmd, isSecure); ok {
```
doriable: Would we be instantiating a new `httpClient` each time we are resolving a completion? When using reflection to resolve schemas at the top-level, we reuse the same transport as the invoker.

cmd/buf/internal/command/curl/curl.go
```diff
@@ -1193,3 +1196,306 @@ func makeHTTP3RoundTripper(f *flags, authority string, printer verbose.Printer)
...
+	return nil
+}
+
+// completeURL provides shell completions for the <url> positional argument.
+// It completes service and method name path components using the first source
+// that succeeds, tried in order:
+//
+//  1. --schema flag (explicit schemas provided by the user)
+//  2. Live server reflection against the URL being completed
+//  3. The buf module found by walking up from the current working directory
```
doriable, line 1229: So this pattern makes sense to me, and mirrors the way `buf curl` works in general, but I wonder if we can leverage and refactor the existing code paths to support this.

Basically, the command does some validation upfront, sets up the URL, and then sets up the client + resolvers for the schema. I wonder if we can refactor a little bit so that we set up the resolvers for the schema after the validation, and then use that for completions for the URL.

That way, we have a single, consistent code path, and also uses the same `container`, client, transport, etc.

stefanvanburen: added in https://github.com/bufbuild/buf/pull/4402/commits/c48c0a6aab11f623245ab2e54c38991bdeae47cf if I'm understanding properly!

cmd/buf/internal/command/curl/curl.go
stefanvanburen: yes, we'd be instantiating a new client each time the user presses <kbd>tab</kbd> (or whatever their keyboard shortcut for invoking completions is)?

PR #4403, opened 2026-03-20
Remove xslog.NopLogger usage

stefanvanburen:
Dropped upstream in favor of standard library.

Ref: bufbuild/standard-go#4

### DIFF
--- a/make/buf/all.mk
+++ b/make/buf/all.mk
@@ -1,6 +1,7 @@
 GO_ALL_REPO_PKGS := ./cmd/... ./private/...
 GO_GET_PKGS := $(GO_GET_PKGS) \
-	github.com/bufbuild/protocompile@main
+	github.com/bufbuild/protocompile@main \
+	buf.build/go/standard@main
 GO_BINS := $(GO_BINS) \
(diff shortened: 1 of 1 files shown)


go.mod
```diff
@@ -13,7 +13,7 @@ require (
 	buf.build/go/protovalidate v1.1.3
 	buf.build/go/protoyaml v0.6.0
 	buf.build/go/spdx v0.2.0
-	buf.build/go/standard v0.1.0
+	buf.build/go/standard v0.1.1-0.20260320140628-2996a887cf13
```
emcfarlane, line 16: I think we need to upgrade this in the Makefile to avoid `Make upgrade` failing by not resolving the latest main. Add to `GO_GET_PKGS`

stefanvanburen: https://github.com/bufbuild/buf/pull/4403/commits/5832e011512a3a0b4a80931bd1dd988f4a73d90e

stefanvanburen: yeah, you're right; tested locally that without that commit, it will downgrade; now it should leave it alone.

PR #4407, opened 2026-03-23
Use the new compiler for building images

### LINKED ISSUE #4193: When will edition 2024 features be supported?
### Feature

I see that ProtoBuf has released Edition 2024, but when using buf v1.61.0, it still indicates that it only supports 2023. Some features of Edition 2024 are unavailable. Are there any other ways to support 2024, such as setting up a trial version?

doriable:
This ports our image building path to use the new compiler and handles all the necessary test changes, etc.

The new compiler brings a lot of improvements:
- Richer diagnostics and error reporting
- Support for Editions 2024 features
- Incremental compilation (although this is mostly for the LSP, which is already using the new compiler)

Fixes #4193

### DIFF
--- a/private/buf/bufctl/controller.go
+++ b/private/buf/bufctl/controller.go
@@ -1376,10 +1376,11 @@ func (c *controller) handleFileAnnotationSetRetError(retErrAddr *error) {
 		} else {
 			// When writing to stderr, check if the input is TTY, if so, allow printing the
 			// diagnostic report.
-			file, ok := c.container.Stdin().(*os.File)
-			if ok {
-				printDiagnosticReport = term.IsTerminal(int(file.Fd()))
-			}
+			//
+			// HACK: We need to check os.Stderr rather than writer, since c.container.Stderr()
+			// returns wrapped writer, which only implements io.Writer. A fix needs to be made
+			// upstream in order to address this.
+			printDiagnosticReport = term.IsTerminal(int(os.Stderr.Fd()))
 		}
 		if err := bufanalysis.PrintFileAnnotationSet(
 			writer,
--- a/private/bufpkg/bufimage/build_image_test.go
+++ b/private/bufpkg/bufimage/build_image_test.go
@@ -431,7 +431,7 @@ func testFileAnnotations(t *testing.T, relDirPath string, parallelism bool, want
 	for i, annotation := range fileAnnotations {
 		got[i] = annotation.String()
 	}
-	assert.Equal(t, len(want), len(got))
+	require.Equal(t, len(want), len(got))
 	for i := range want {
 		options := strings.Split(want[i], "||")
 		matched := false


private/bufpkg/bufimage/build_image_test.go
```diff
@@ -267,29 +268,26 @@ func TestCyclicImport(t *testing.T) {
...
 		"cyclicimport",
 		false,
-		// Since the compiler is multi-threaded, order of file compilation can happen one of two ways
-		fmt.Sprintf(`%s:5:8:cycle found in imports: "a/a.proto" -> "b/b.proto" -> "a/a.proto"
-				|| %s:5:8:cycle found in imports: "b/b.proto" -> "a/a.proto" -> "b/b.proto"`,
-			filepath.FromSlash("testdata/cyclicimport/a/a.proto"),
+		fmt.Sprintf(
+			`%s:5:1:detected cyclic import while importing "a/a.proto"`,
 			filepath.FromSlash("testdata/cyclicimport/b/b.proto"),
 		),
```
doriable, line 274: I don't love this diagnostic, and the fact that the diagnostics in general are unable to provide the more robust details in a single `FileAnnotation`... e.g. this is the full report shows a much clearer cycle: https://github.com/bufbuild/protocompile/blob/main/experimental/ir/testdata/imports/cycle.proto.yaml.stderr.txt

That being said, if the user is printing the output to text, then they'll get the full information. This is only on the individual `FileAnnotation` (which would be used for things like in-line github annotations, the `json` output for analysis, etc.) -- and would provide a clear place where the import cycle was detected (although more details, in this case in particular would be nice). Will noodle this one a little.

private/bufpkg/bufimage/build_image.go
```diff
@@ -49,283 +50,221 @@ func buildImage(
...
+			moduleFileResolver,
 			excludeSourceCodeInfo,
-			fileDescriptor,
-			parserAccessorHandler,
-			syntaxUnspecifiedFilenames,
-			filenameToUnusedDependencyFilenames,
-			alreadySeen,
-			nonImportFilenames,
+			seen,
+			xslices.ToStructMap(paths),
```
stefanvanburen: nit: call this once outside the loop?

private/bufpkg/bufimage/module_file_resolver.go
```diff
@@ -0,0 +1,195 @@
...
+			}
+			return readObjectCloserToSourceFile(path, wktModuleFile)
+		}
+		return nil, moduleErr
+	}
+	defer func() {
+		if retErr != nil {
+			retErr = errors.Join(retErr, moduleFile.Close())
+		}
+	}()
```
stefanvanburen, line 88: I think we want to close the moduleFile regardless of if we have an error, because `readObjectCloserToSourceFile` will copy over its contents but not close it.

```suggestion
	defer moduleFile.Close()
```

private/bufpkg/bufimage/build_image.go
```diff
@@ -49,283 +50,221 @@ func buildImage(
...
-	moduleReadBucket = bufmodule.ModuleReadBucketWithOnlyProtoFiles(moduleReadBucket)
-	parserAccessorHandler := newParserAccessorHandler(ctx, moduleReadBucket)
+
 	targetFileInfos, err := bufmodule.GetTargetFileInfos(ctx, moduleReadBucket)
 	if err != nil {
 		return nil, err
 	}
+
 	if len(targetFileInfos) == 0 {
 		// If we had no no target files within the module after path filtering, this is an error.
```
stefanvanburen:
```suggestion
		// If we had no target files within the module after path filtering, this is an error.
```

Just since we're nearby.

cmd/buf/buf_test.go
```diff
@@ -4493,7 +4493,29 @@ func TestProtoFileNoWorkspaceOrModule(t *testing.T) {
 		nil,
 		bufctl.ExitCodeFileAnnotation,
 		"", // no stdout
-		filepath.FromSlash(`testdata/protofileref/noworkspaceormodule/fail/import.proto:3:8:import "`)+`google/type/date.proto": file does not exist`,
+		fmt.Sprintf(
```
bufdev, line 4496: This (and all others) is likely a breaking change. I am not sure if this matters on stderr, I can't remember, but buf was designed to output errors in the standard file:line:column:message format, which is relied upon by many tools.

doriable: Did a sweep, and I think it's mostly ALE -> stdout. That being said, this is still a valid concern and @stefanvanburen had a good suggestion, which I am fully supportive of:

- Add a new `error-format=pretty`, which would provide the new error report format
- Detect if the output is TTY, which we [already do for logging](https://github.com/bufbuild/buf/blob/f2fc7cec115a9b75f94b1b30dced30800cc5e3fd/private/pkg/slogapp/console.go#L204), and if it is, then we default to `error-format=pretty`, and other wise, keep the default as `error-format=text`. This would keep CI's consistent, ALE, etc.
- This would apply to anything that produces compiler diagnostics and file annotations, so I can do a bit of work to structure this for `lint, breaking`

Let me know if this approach seems sound to you!

doriable: Discussed this offline, putting some notes here:
- No need for a new format name, instead we'll just default to this report format for TTY (and keep the old format for non-TTY)
- No changes to `lint, breaking`'s `stdout` output -- this will only be used for stderr compiler output

private/buf/bufctl/controller.go
```diff
@@ -1365,15 +1368,25 @@ func (c *controller) handleFileAnnotationSetRetError(retErrAddr *error) {
...
 	var fileAnnotationSet bufanalysis.FileAnnotationSet
+	var printDiagnosticReport bool
 	if errors.As(*retErrAddr, &fileAnnotationSet) {
 		writer := c.container.Stderr()
 		if c.fileAnnotationsToStdout {
 			writer = c.container.Stdout()
+		} else {
+			// When writing to stderr, check if the input is TTY, if so, allow printing the
+			// diagnostic report.
+			file, ok := c.container.Stdin().(*os.File)
```
stefanvanburen: nit: should be stderr to match the comment?
```suggestion
			file, ok := c.container.Stderr().(*os.File)
```

private/bufpkg/bufimage/build_image_test.go
```diff
@@ -433,7 +431,7 @@ func testFileAnnotations(t *testing.T, relDirPath string, parallelism bool, want
 	for i, annotation := range fileAnnotations {
 		got[i] = annotation.String()
 	}
-	require.Equal(t, len(want), len(got))
+	assert.Equal(t, len(want), len(got))
```
stefanvanburen: I think `testifylint` would flag this (we will enable someday!), but I think this should go back to `require`, because we will panic below if len(got) < len(want) when trying to access.

```suggestion
	require.Equal(t, len(want), len(got))
```

private/buf/bufctl/controller.go
doriable: Leaving a note here before updating: chatted about this -- there is a limitation with `app-go`'s wrapping of `os.Stderr` where it only implements `io.Writer`, and not `os.File`. We're going to put a well-documented workaround and follow-up after.

PR #4411, opened 2026-03-24
Sync PROTOC_VERSION with makego

stefanvanburen:
This had fallen out of sync with makego mistakenly from #4383; bumped to latest version in bufbuild/makego#295 and now pulled in here.


stefanvanburen:
I forgot this from the previous PR:

> Bumps PROTOC_VERSION (and related bits) to 33.3, **which is the latest version that protobuf-go supports.**

We need a new protobuf-go release to upgrade to 34.1.

For now, I'm just going to leave this de-synced; we ought to update makego to just 33.3, but not looking to churn here more.

PR #4413, opened 2026-03-24
Implement smart clean

### LINKED ISSUE #3126: Make buf generate --clean a per-plugin option that does a "smart" clean

emcfarlane:
This implements smart clean for `buf generate --clean` as described in this [comment](https://github.com/bufbuild/buf/pull/3124#issuecomment-2209407669), building on the mtime-preserving skip-unchanged behavior from #4396.

With --clean, buf generate now handles each output directory intelligently:
- Unchanged files are skipped (mtime preserved)
- Changed files are overwritten
- New files are created
- Stale files that are no longer generated are deleted
- Empty directories left behind are cleaned up

This replaces the previous delete-everything-then-regenerate approach with a compare-and-reconcile model, which is safer for mtime-based build systems (make, cargo, ninja) and avoids unnecessary downstream rebuilds.

Architecture change: the `ResponseWriter` is now created once per `Generate()` call rather than per image. This is required so the writer can see the full set of generated paths across all images before deciding what to delete. A side effect is improved atomicity: if any plugin fails, no partial output is written. The trade-off is holding all images' output in
memory until flush, which should be negligible in practice.

Fixes #3126

### DIFF
--- a/private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
+++ b/private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
@@ -289,7 +289,12 @@ func (w *responseWriter) writeZip(
 		if err == nil && bytes.Equal(existingContent, newContent) {
 			return nil
 		}
-		return os.WriteFile(outFilePath, newContent, 0666)
+		file, err := os.Create(outFilePath)
+		if err != nil {
+			return err
+		}
+		_, writeErr := file.Write(newContent)
+		return errors.Join(writeErr, file.Close())
 	})
 	return nil
 }


private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
```diff
@@ -225,18 +271,21 @@ func (w *responseWriter) writeZip(
...
-			retErr = errors.Join(retErr, file.Close())
-		}()
-		// protoc does not compress.
-		return storagearchive.Zip(ctx, readWriteBucket, file, false)
+		newContent := buf.Bytes()
+		existingContent, err := os.ReadFile(outFilePath)
+		if err == nil && bytes.Equal(existingContent, newContent) {
+			return nil
+		}
+		return os.WriteFile(outFilePath, newContent, 0600)
```
stefanvanburen: just checking: is this matching the behavior of `os.Create`? just wondering about the `0600`

private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
```diff
@@ -325,10 +374,129 @@ func (w *responseWriter) copySkipUnchanged(
...
+	if err != nil {
+		return "", err
+	}
+	return filepath.EvalSymlinks(path)
+}
+
+// isArchivePath returns true if the given path has a .zip or .jar extension.
+func isArchivePath(path string) bool {
+	ext := filepath.Ext(path)
+	return ext == ".zip" || ext == ".jar"
```
stefanvanburen: seems like we should consolidate these extensions into some consts/seqs so we don't miss updating them?

private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
```diff
@@ -325,10 +374,129 @@ func (w *responseWriter) copySkipUnchanged(
 	return thread.Parallelize(ctx, jobs)
 }
 
+// deleteStaleFilesAndEmptyDirs deletes files present in outDirPath that are
+// not in retainPaths, then removes any directories that are now empty.
+func (w *responseWriter) deleteStaleFilesAndEmptyDirs(
+	outDirPath string,
+	retainPaths map[string]struct{},
+) error {
+	ctx := context.Background()
```
stefanvanburen: nit: should thread through the `ctx`, even if `Close` doesn't take one?

private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
emcfarlane: Fixed, 0666.

private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
emcfarlane: Made consts in this package. It is used outside, but rarely.

private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
emcfarlane: Changed interface to `Close(ctx context.Context) error`

stefanvanburen: fwiw, I'd be fine with leaving them as `io.Closer`s and just having a single `context.Background()` that gets passed down within the Closer itself. But the update you made also makes sense to me. Just figured we ought to share one.

PR #4418, opened 2026-03-30
Add `buf beta scaffold` command to setup buf.yaml in a git repository

### LINKED ISSUE #3159: `buf scaffold` command to setup a new Proto definition git repository
### Feature

For ease of `buf` adoption (migrating existing definitions and/or starting new Protobuf projects with `buf`), we should have a `buf scaffold` command that sets up a new git repository for Proto definitions.

cyinma:
fix #3159

### DIFF
--- a/cmd/buf/buf.go
+++ b/cmd/buf/buf.go
@@ -158,7 +158,6 @@ func newRootCommand(name string) *appcmd.Command {
 			push.NewCommand("push", builder),
 			convert.NewCommand("convert", builder),
 			curl.NewCommand("curl", builder),
-			scaffold.NewCommand("scaffold", builder),
 			{
 				Use:   "dep",
 				Short: "Work with dependencies",
@@ -409,6 +408,7 @@ func newRootCommand(name string) *appcmd.Command {
 				SubCommands: []*appcmd.Command{
 					lspserve.NewCommand("lsp", builder, deprecatedMessage("buf lsp serve", "buf beta lsp"), true, true),
 					price.NewCommand("price", builder),
+					scaffold.NewCommand("scaffold", builder),
 					bufpluginv1beta1.NewCommand("buf-plugin-v1beta1", builder),
 					bufpluginv1.NewCommand("buf-plugin-v1", builder),
 					bufpluginv2.NewCommand("buf-plugin-v2", builder),
--- a/cmd/buf/buf.go
+++ b/cmd/buf/buf.go
@@ -116,6 +116,7 @@ import (
 	"github.com/bufbuild/buf/cmd/buf/internal/command/registry/sdk/sdkinfo"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/registry/sdk/version"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/registry/whoami"
+	"github.com/bufbuild/buf/cmd/buf/internal/command/scaffold"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/source/sourceedit/sourceeditdeprecate"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/stats"
 	"github.com/bufbuild/buf/private/buf/bufcli"
@@ -157,6 +158,7 @@ func newRootCommand(name string) *appcmd.Command {
 			push.NewCommand("push", builder),
 			convert.NewCommand("convert", builder),
 			curl.NewCommand("curl", builder),
+			scaffold.NewCommand("scaffold", builder),
 			{
 				Use:   "dep",
 				Short: "Work with dependencies",
--- a/cmd/buf/internal/command/scaffold/scaffold.go
+++ b/cmd/buf/internal/command/scaffold/scaffold.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"io/fs"
 	"os"
 	"path"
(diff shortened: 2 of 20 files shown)


cmd/buf/buf.go
```diff
@@ -157,6 +158,7 @@ func newRootCommand(name string) *appcmd.Command {
 			push.NewCommand("push", builder),
 			convert.NewCommand("convert", builder),
 			curl.NewCommand("curl", builder),
+			scaffold.NewCommand("scaffold", builder),
```
cyinma: I am thinking should this be in beta first, but I don't think we want this to be `buf config scaffold`

cmd/buf/internal/command/scaffold/scaffold.go
```diff
@@ -0,0 +1,223 @@
...
+		return fmt.Errorf("%s is not the root of a git repository", dirPath)
+	}
+	exists, err := bufcli.BufYAMLFileExistsForDirPath(ctx, dirPath)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return fmt.Errorf("buf.yaml already exists in %s, will not overwrite", dirPath)
+	}
+	fsys := os.DirFS(dirPath)
```
cyinma: not using the `storage.ReadBucket` abstraction since in the end we need to read the raw content for parser

doriable: I think we should still use `storageosProvider.NewReadBucket`, just for consistency in the implementation patterns for these code paths (e.g. handling the symlinks, etc.).

cmd/buf/internal/command/scaffold/scaffold.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+type protoFileInfo struct {
+	filePath    string
+	packageName string
+}
+
+// walkAndParseProtoFiles walks the filesystem for .proto files and parses each one.
+// Files that fail to parse or have no package declaration are silently skipped.
+func walkAndParseProtoFiles(fsys fs.FS) ([]protoFileInfo, error) {
```
doriable: So the current behaviour is, on the walk, for each file, we synchronously parse using `parser.Parse` directly. I think it might make sense to:
- Walk, collect each proto source file along the way, [dump them into a `source.Map`](https://github.com/bufbuild/protocompile/blob/a86a8ee991d748fd73ebe3ecdd32d93d66a59ed6/experimental/source/opener.go#L41-L47) so that you get the `source.Opener` implementation and create a `queries.AST` for each file (https://github.com/bufbuild/protocompile/blob/main/experimental/incremental/queries/ast.go)
- Then, you can run `incremental.Run` on all the queries, and this will parse/resolve the AST in parallel
- And then `inferModuleRoots`'s signature can just be `func inferModuleRoots(files ...ast.File) []string`

cmd/buf/internal/command/scaffold/scaffold.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	return fileInfos, nil
+}
+
+// inferModuleRoots determines module root directories by deriving each file's
+// expected import path from its package declaration, then comparing against its
+// file path. The difference is the module root.
+//
+// Returns sorted, unique module root paths. Returns nil if no roots can be inferred.
+func inferModuleRoots(files []protoFileInfo) []string {
```
doriable: So I think here, we can get a bit of a "freebie" based on existing imports. I think the algorithm I would do is, assuming `inferModuleRoots(files ...ast.File) []string`:

- Iterate through each file and gather up the [imports](https://github.com/bufbuild/protocompile/blob/a86a8ee991d748fd73ebe3ecdd32d93d66a59ed6/experimental/ast/context.go#L89-L97) in a set
- Then you iterate through files again and compare the normalised `Path` (which should be what you get from the bucket) against the import paths that we have (I think the way to do this is to compare "tail to front", e.g. for `local/path/to/abc/v1/xyz.proto`, you would find the closest match, starting from `xyz.proto` -> `v1/xyz.proto` -> `abc/v1/xyz.proto`, etc.)
- For any files that have not been imported at all, we can run a fallback to the package name check

Otherwise this is entirely dependent on well-formed package names.

cmd/buf/internal/command/scaffold/scaffold.go
```diff
@@ -0,0 +1,223 @@
...
+		moduleConfigs,
+		nil,
+		nil,
+		nil,
+		bufconfig.BufYAMLFileWithIncludeDocsLink(),
+	)
+	if err != nil {
+		return err
+	}
+	return bufcli.PutBufYAMLFileForDirPath(ctx, dirPath, bufYAMLFile)
```
doriable: I think before we write the `bufYAMLFile` to disk permanently, we should validate that this produces a workspace that compiles, which is a little tricky. With the way our APIs are currently structured, I think I would temporarily write the `buf.yaml` to disk, and then use `Controller` to attempt to `GetImage` based on the `dirPath` (since `buf.yaml` is written in the same place as `dirPath`, which would be the "input" in this case), and if it succeds, great, keep it written and return. If it fails, we would not write the the `buf.yaml`, but maybe we have a verbose option of providing the roots we found or something? Would need to look into the UX of the failing case a little but, but yeah... I think some validation step is useful.

cmd/buf/internal/command/scaffold/scaffold.go
cyinma: make sense, and I opt to not supporting the `--disable-symlinks` flag, which is aligned with other commands that take a directory as input

cmd/buf/internal/command/scaffold/scaffold.go
cyinma: added the build validation, but yes we might need to see how to present the build error as otherwise this can be confusing

cmd/buf/buf.go
cyinma: moved this to beta

doriable:
Closing this for now, since there are additional edge-cases that need to be considered and this is not currently needed.

PR #4437, opened 2026-04-08
Add `--debug-address` flag for debug and profiling of the LSP

emcfarlane:
Add `--debug-address` flag to `buf lsp serve` that starts a debug HTTP server (similar to gopls's --debug flag). When set, it serves a browsable debug UI with:

- Server info and build details
- Memory stats (runtime.MemStats)
- pprof endpoints (goroutines, heap, allocs, block, mutex)

Usage: start the LSP with `--debug-address=localhost:0` and find the assigned port in the LSP logs. Open the address in a browser to inspect server state, memory usage, goroutines, and collect CPU/heap profiles to help debug issues.


emcfarlane:
Screenshot 2026-04-08 at 6 01 35 PM
Screenshot 2026-04-08 at 6 01 31 PM
Screenshot 2026-04-08 at 6 01 27 PM
Screenshot 2026-04-08 at 6 01 25 PM

Screenshots of the different views (and the familiar /debug/pprof handler).

doriable (approved):
This looks good to me, and I think that it will help us gather some better stats on the LSP. We can also iterate on this over time.

As for the flags, I was debating whether we wanted a default behaviour, e.g. `--debug-address=""`, but this emulates the behaviour of `gopls` and other similar tooling, so I think having the user provide, at minimum, `localhost:0` makes sense to me.

Approving ahead of merge conflict resolution.

doriable (approved):

PR #4438, opened 2026-04-09
Add LSP codelenses for buf.yaml dep upgrades

stefanvanburen:
This adds an initial set of LSP codelenses for buf.yaml files that act similarly to `gopls`' handling of `go.mod` files. There are currently two supported lenses on the `deps:` line, that correspond to upgrading all dependencies (resulting in a changed `buf.lock` if necessary), or checking for updates (resulting in a diagnostic on the dependency line with details on the upgrade).

In order for this to work, clients (e.g., vscode-buf) must be updated to handle `buf.yaml` (rather than just `.proto`) files; I plan on making the necessary updates to those repositories. For now, if you're testing in e.g. `nvim`, you'd need to update your [buf_ls configuration][1] to include `filetypes = { 'proto', 'yaml' }`.

This is the first in a series of LSP features that I think we can add to supporting buf configuration files; follow-ups include better hover documentation (currently, most editors integrate with our JSON Schema, but we can do better with LSP hovers), document links (to BSR modules), inlay hints and perhaps completions.

[1]: https://github.com/neovim/nvim-lspconfig/blob/master/lsp/buf_ls.lua


stefanvanburen:
For context, here's `gopls` in buf, showing the neovim codelenses:

https://github.com/user-attachments/assets/88541c4d-2b16-44d7-a3e5-23e4ed5727fe

And here's `buf` in registry-proto, showing a similar update:

https://github.com/user-attachments/assets/8e17fb07-8ea6-451a-b456-f699714ff067

private/buf/buflsp/buf_yaml.go
```diff
@@ -0,0 +1,385 @@
...
+func newBufYAMLManager(lsp *lsp) *bufYAMLManager {
+	return &bufYAMLManager{
+		lsp:       lsp,
+		uriToFile: make(map[protocol.URI]*bufYAMLFile),
+	}
+}
+
+// bufYAMLFile holds the parsed state of an open buf.yaml file.
+type bufYAMLFile struct {
+	depsKeyLine uint32 // 0-indexed line of the "deps:" key
```
stefanvanburen, line 64: I'd prefer to use our internal types to handle this, but unfortunately they don't give us the info we need regarding positioning of yaml nodes within the file, which we need here.

private/buf/buflsp/buf_yaml_lsp_test.go
```diff
@@ -0,0 +1,483 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.lsp.dev/jsonrpc2"
+	"go.lsp.dev/protocol"
+	"go.lsp.dev/uri"
+)
+
+// mockModuleKeyProvider is a ModuleKeyProvider that returns fixed latest commits for testing.
+type mockModuleKeyProvider struct {
```
emcfarlane: This could be moved into `bufmodule` as a `bufmodule.StaticKeyProvider` following the `bufplugin` and `bufpolicy` static key providers.

emcfarlane (approved):

private/buf/buflsp/buf_yaml_lsp_test.go
stefanvanburen: @emcfarlane just double-checking that https://github.com/bufbuild/buf/pull/4438/commits/9613ec8dcc77a572207bf926e04080ec63ce8fdc is what you had in mind?

emcfarlane (approved):

PR #4439, opened 2026-04-09
Handle files that have not completed lowering in LSP

doriable:
This implements more graceful behaviour in the LSP for when
a file is unable to go through the full lowering process, and therefore
cannot/does not have fully resolved symbols for LSP functionality, and
attempts to return the most helpful possible diagnostic to the user to
communicate this.

Fixes #4434

PR #4445, opened 2026-04-10
Extend diagnostics to include notes/help/debug

stefanvanburen:
Quick follow-up to #4439. In that PR, we mentioned that these [fields could become code actions][1], which is probably true with some refactoring / exporting of more of the underlying structure so we don't have just `string`s without more structured data. For now, it seems valuable to just expose these as additional information added to the hover states. See the added test-cases for how these look.

[1]: https://github.com/bufbuild/buf/pull/4439#discussion_r3060536313

### DIFF
--- a/private/buf/buflsp/diagnostic.go
+++ b/private/buf/buflsp/diagnostic.go
@@ -43,12 +43,21 @@ var reportLevelToDiagnosticSeverity = map[report.Level]protocol.DiagnosticSeveri
 func reportDiagnosticToProtocolDiagnostic(
 	reportDiagnostic report.Diagnostic,
 ) protocol.Diagnostic {
-	message := reportDiagnostic.Message()
-	if reportDiagnostic.Level() == report.ICE {
-		// Include notes for ICE
-		notes := append([]string{message}, reportDiagnostic.Notes()...)
-		message = strings.Join(notes, " ")
+	parts := []string{reportDiagnostic.Message()}
+	for _, note := range reportDiagnostic.Notes() {
+		parts = append(parts, "note: "+note)
 	}
+	for _, help := range reportDiagnostic.Help() {
+		parts = append(parts, "help: "+help)
+	}
+	// Debug info is implementation-level detail; only include it for ICE where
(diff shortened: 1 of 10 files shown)


private/buf/buflsp/diagnostic.go
```diff
@@ -43,12 +43,21 @@ var reportLevelToDiagnosticSeverity = map[report.Level]protocol.DiagnosticSeveri
...
-	message := reportDiagnostic.Message()
+	parts := []string{reportDiagnostic.Message()}
+	for _, note := range reportDiagnostic.Notes() {
+		parts = append(parts, "note: "+note)
+	}
+	for _, help := range reportDiagnostic.Help() {
+		parts = append(parts, "help: "+help)
+	}
+	// Debug info is implementation-level detail; only include it for ICE where
+	// all available context helps diagnose the unexpected failure.
```
stefanvanburen, line 54: was just checking where we `Debugf` in protocompile; this logic is the same as before, just updates the format to `debug: ` similar to the other added types.

private/buf/buflsp/diagnostics_test.go
```diff
@@ -237,6 +237,147 @@ func TestDiagnostics(t *testing.T) {
...
+			protoFile: "testdata/diagnostics/enum_map_key.proto",
+			expectedDiagnostics: []protocol.Diagnostic{
+				{
+					Range: protocol.Range{
+						Start: protocol.Position{Line: 10, Character: 6},
+						End:   protocol.Position{Line: 10, Character: 12},
+					},
+					Severity: protocol.DiagnosticSeverityError,
+					Source:   "buf-lsp",
+					Message:  "expected map key type, found enum type `diagnostics.v1.Status`\nhelp: valid map key types are integer types, `string`, and `bool`\nhelp: counterintuitively, user-defined enum types cannot be used as keys",
```
stefanvanburen, line 266: this sort of thing seems very valuable to include in our hover diagnostics, as it really points to how to solve the issue. (I do think _another_ follow up of adding code actions to resolve these, where possible, would be great, though.)

doriable (approved):

PR #4461, opened 2026-04-13
Switch from x/crypto to stdlib sha3 hash

pkwarren:
Starting in Go 1.24, sha3 is now part of the stdlib. Switch to it instead of consuming via x/crypto.

### DIFF
--- a/private/pkg/shake256/shake256.go
+++ b/private/pkg/shake256/shake256.go
@@ -16,11 +16,10 @@
 package shake256
 
 import (
+	"crypto/sha3"
 	"fmt"
 	"io"
 	"slices"
-
-	"golang.org/x/crypto/sha3"
 )
 
 const shake256Length = 64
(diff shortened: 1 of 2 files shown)


private/pkg/shake256/shake256.go
```diff
@@ -39,9 +38,7 @@ func NewDigest(value []byte) (Digest, error) {
 
 // NewDigestForContent returns a new Digest for the content read from the Reader.
 func NewDigestForContent(reader io.Reader) (Digest, error) {
-	shakeHash := sha3.NewShake256()
-	// TODO FUTURE: remove in the future, this should have no effect
-	shakeHash.Reset()
+	shakeHash := sha3.NewSHAKE256()
```
pkwarren, line 41: https://cs.opensource.google/go/x/crypto/+/master:sha3/shake.go;l=37-42;drc=cf29fa96f8b66328e59829f064539321159bfa5b

The x/crypto one wraps this implementation from the stdlib.

private/pkg/shake256/shake256.go
```diff
@@ -39,9 +38,7 @@ func NewDigest(value []byte) (Digest, error) {
 
 // NewDigestForContent returns a new Digest for the content read from the Reader.
 func NewDigestForContent(reader io.Reader) (Digest, error) {
-	shakeHash := sha3.NewShake256()
-	// TODO FUTURE: remove in the future, this should have no effect
-	shakeHash.Reset()
```
pkwarren, line 44: Wasn't sure why this was there since we were creating it just before.

stefanvanburen (approved):

doriable (approved):

PR #4463, opened 2026-04-14
Add LSP document link support for remaining buf config files

stefanvanburen:
Similar to #4442, but for the remaining config types:

* `buf.gen.yaml`: `plugins.remote`, `inputs.module`
* `buf.policy.yaml`: top-level `name`, `plugins.plugin` (that look like BSR references)
* `buf.lock`: `deps.name`


CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- Add LSP document links for `buf.yaml` dep entries, making each dependency a clickable link to its BSR module page.
+- Add LSP document links for `buf.yaml` deps, `buf.gen.yaml` remote plugins and input modules, `buf.policy.yaml` name and BSR plugins, and `buf.lock` dep names, making each a clickable link to its BSR page.
```
stefanvanburen, line 5: figured we could consolidate the CHANGELOG if this makes it into the release. Otherwise, I'll split it out.

private/buf/buflsp/document_link.go
```diff
@@ -81,6 +82,17 @@ func (s *server) documentLink(file *file) []protocol.DocumentLink {
 	return links
 }
 
+// bsrRefDocURL builds a BSR page URL from a parsed reference.
+// When ref.Ref() is non-empty, the URL points to the versioned /docs/<ref> page.
+func bsrRefDocURL(ref bufparse.Ref) string {
+	fn := ref.FullName()
+	url := "https://" + fn.Registry() + "/" + fn.Owner() + "/" + fn.Name()
+	if ref.Ref() != "" {
+		url += "/docs/" + ref.Ref()
```
emcfarlane, line 91: nit: would expect the fallback to `https://buf.build/docs/`.

emcfarlane (approved):

private/buf/buflsp/document_link.go
stefanvanburen: ah, this is going to https://buf.build/bufbuild/buf/docs/v1.67.0 (versus just a regular https://buf.build/bufbuild/buf on no ref), so I think this is correct as-is.

PR #4476, opened 2026-04-15
Consolidate CEL compilation errors to single line

stefanvanburen:
Matching our default `buf lint` format. Still, when cel-go reports multiple issues, we'll include all of them. Also strips out the leading "Syntax error: ", which IMO seems superfluous.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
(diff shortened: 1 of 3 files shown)


doriable (approved):
I think this is worth changelog-ing -- it seems like this was an issue for a while now, thank you! (I agree with the decision to trim `Syntax error`, since that is the pathological case for CEL expressions. If other errors crop up, they'll be properly prefixed)

PR #4478, opened 2026-04-16
Handle vendored descriptor.proto gracefully

emcfarlane:
Upgrade protocompile to handle vendored descriptor.proto.

Fix resolver construction when compiled files include a vendored `descriptor.proto`. The buf descriptor extension (field `536000000` on FileDescriptorSet) fails protodesc validation when FileDescriptorSet is resolved from the vendored file rather than left as a placeholder, because it has no declared extension ranges. Skip prepending the buf descriptor proto when `descriptor.proto` is already in the compiled output.

Depends on https://github.com/bufbuild/protocompile/pull/710

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [Unreleased]
 
+- Fix build failures for modules with a vendored `descriptor.proto`.
 - Fix CEL compilation error messages in `buf lint` to use the structured error API instead of parsing cel-go's text output.
 
 ## [v1.68.1] - 2026-04-14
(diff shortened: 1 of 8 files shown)


emcfarlane:
As an e2e example. The LSP will show errors when using features that are not suported/outside the provided `descriptor.proto` version. Here I use the dependency `buf.build/protocolbuffers/wellknowntypes:v27.1` with an edition 2024 file with the some example features:

Screenshot 2026-04-17 at 11 49 14 AM

The build will also fail with the following errors:
```
custom.proto:11:16:cannot find message field `enforce_naming_style` in `google.protobuf.FeatureSet`
custom.proto:12:16:cannot find message field `default_symbol_visibility` in `google.protobuf.FeatureSet`
custom.proto:24:28:cannot find message field `large_enum` in `pb.JavaFeatures`
```

doriable (approved):

PR #4493, opened 2026-04-20
Reduce allocations in protoencoding.ReparseExtensions

pkwarren:
The protoencoding.ReparseExtensions method is showing up in memory profiles as a large source of allocations. Update the method to add a per-call cache to prevent allocations when revisiting the same types.

PR #4494, opened 2026-04-21
Reduce allocations in protoencoding.ReparseExtensions, take 2

jhump:
This is an alternative solution to #4493. It improves the "RealWorld" benchmark by approximately the same amount (in both runtime and allocations), but is better on the "Synthetic" case. The value of the "Synthetic" case is that it has a shape more like small modules with fewer fields and/or modules where nearly all elements are annotated. For the former, this solution elides the map that caches relevant fields for all messages (whose cost to construct can outweight savings for small modules where there are fewer elements to amortize the cost of the cache). For the latter, it hoists the lambda and all accessed state up into a struct so it can be allocated once at the start of the operation instead of per-recursive-call (so if there are a very large number of extensions, we still beat the baseline in `main`).

Here's the output of `benchstat`:
```
goos: darwin
goarch: arm64
pkg: github.com/bufbuild/buf/private/pkg/protoencoding
cpu: Apple M1 Max
                               │ old_bench.txt │            new_bench.txt            │
                               │    sec/op     │    sec/op     vs base               │
ReparseExtensions_Synthetic-10    5.124m ± ∞ ¹   4.354m ± ∞ ¹  -15.03% (p=0.008 n=5)
ReparseExtensions_RealWorld-10    5.205m ± ∞ ¹   3.238m ± ∞ ¹  -37.80% (p=0.008 n=5)
geomean                           5.165m         3.755m        -27.30%
¹ need >= 6 samples for confidence interval at level 0.95

                               │ old_bench.txt │            new_bench.txt             │
                               │     B/op      │     B/op       vs base               │
ReparseExtensions_Synthetic-10   3.629Mi ± ∞ ¹   3.322Mi ± ∞ ¹   -8.47% (p=0.008 n=5)
ReparseExtensions_RealWorld-10   4.635Mi ± ∞ ¹   2.961Mi ± ∞ ¹  -36.12% (p=0.008 n=5)
geomean                          4.101Mi         3.136Mi        -23.53%
¹ need >= 6 samples for confidence interval at level 0.95

                               │ old_bench.txt │            new_bench.txt            │
                               │   allocs/op   │  allocs/op    vs base               │
ReparseExtensions_Synthetic-10    57.63k ± ∞ ¹   43.53k ± ∞ ¹  -24.47% (p=0.008 n=5)
ReparseExtensions_RealWorld-10   111.35k ± ∞ ¹   60.30k ± ∞ ¹  -45.84% (p=0.008 n=5)
geomean                           80.11k         51.24k        -36.04%
¹ need >= 6 samples for confidence interval at level 0.95
```

### DIFF
--- a/private/pkg/protoencoding/reparse_extensions.go
+++ b/private/pkg/protoencoding/reparse_extensions.go
@@ -80,6 +80,9 @@ func (r *reparser) reparseExtensions(reflectMessage protoreflect.Message) {
 		r.reflectMessage = reflectMessage
 
 		reflectMessage.Range(r.visitExt)
+		if r.err != nil {
+			return
+		}
 		if r.msgExts != nil {
 			options := proto.MarshalOptions{AllowPartial: true}
 			var err error
@@ -121,7 +124,6 @@ func (r *reparser) reparseExtensions(reflectMessage protoreflect.Message) {
 			return
 		}
 	}
-	return
 }
 
 func (r *reparser) reparseInMessageField(msg protoreflect.Message, field protoreflect.FieldDescriptor) {
--- a/private/pkg/protoencoding/reparse_extensions.go
+++ b/private/pkg/protoencoding/reparse_extensions.go
@@ -25,47 +25,68 @@ func ReparseExtensions(resolver Resolver, reflectMessage protoreflect.Message) e
 	if resolver == nil {
 		return nil
 	}
-	return reparseExtensions(
-		resolver,
-		map[protoreflect.MessageDescriptor][]protoreflect.FieldDescriptor{},
-		reflectMessage,
-	)
+	rep := newReparser(resolver)
+	rep.reparseExtensions(reflectMessage)
+	return rep.err
 }
 
-// reparseExtensions implements ReparseExtensions with a shared cache of
-// recursion-worthy fields per descriptor. The cache lives for the duration of
-// the top-level call, so memory is bounded by the descriptor types visited in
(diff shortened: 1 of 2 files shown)


private/pkg/protoencoding/reparse_extensions.go
```diff
@@ -25,81 +25,123 @@ func ReparseExtensions(resolver Resolver, reflectMessage protoreflect.Message) e
...
+		reflectMessage.Range(r.visitExt)
+		if r.msgExts != nil {
 			options := proto.MarshalOptions{AllowPartial: true}
 			var err error
-			reparseBytes, err = options.MarshalAppend(reparseBytes, msgExts.Interface())
+			reparseBytes, err = options.MarshalAppend(reparseBytes, r.msgExts.Interface())
 			if err != nil {
-				return err
+				r.err = err
+				return
```
pkwarren, line 92: Any benefit to short circuiting above - `if r.err != nil { return }`?

pkwarren (approved):
There's a lint failure but overall looks good.

private/pkg/protoencoding/reparse_extensions.go
jhump: I can add it, as a just in case. It's not needed today since `visitExt` never sets `r.err`. But that's a good check to have in place in case we ever touch the `visitExt` function and add an error case to it.

pkwarren (approved):

Issue #3985, opened 2025-08-27
Provide feedback to users on invalid configurations in `buf.yaml`


doriable:
### Feature

Saw a question/request around whether `buf lint` returns an error for invalid `buf.yaml` configurations, such as an incorrectly configured `ignore` or a stale `ignore_only`.
We currently do not provide that feedback in `buf {lint,breaking}` commands.

It would be nice if either these commands can surface these issues (either as warnings or annotations) or if we had some kind of utility for validating the config, `buf.yaml`, e.g. a `buf config` subcommand, such as `validate`.

Some prior art worth noting: [Biome](https://biomejs.dev/) provides errors on an ignore that is invalid/unnecessary.

PR #4498, opened 2026-04-22
Add LSP warning diagnostics for invalid buf.yaml ignore paths

stefanvanburen:
Warn when a path listed under `lint.ignore` or `breaking.ignore` in buf.yaml does not match any file in the workspace.

Relates to #3985.


stefanvanburen:
Looks like:

image

doriable (approved):

PR #4504, opened 2026-04-24
feat: add BUF_WASM_MAX_MEMORY_BYTES env var to configure WASM memory limit

paultyng:
The WASM plugin sandbox is hardcoded to 512 MiB. Larger check plugins (e.g., those embedding the AEP api-linter with transitive gRPC deps) can OOM during init before any proto analysis begins.

Add BUF_WASM_MAX_MEMORY_BYTES environment variable to override the default. The WithMaxMemoryBytes runtime option already exists but was never exposed to users.

### DIFF
--- a/private/buf/bufcli/cache.go
+++ b/private/buf/bufcli/cache.go
@@ -21,6 +21,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"strconv"
 
 	"buf.build/go/app/appext"
 	"github.com/bufbuild/buf/private/buf/bufwkt/bufwktstore"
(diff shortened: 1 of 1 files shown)


emcfarlane:
Hey @paultyng, thanks for the PR. From your issue this morning we bumped the memory limits from 512MiB to 1GiB: https://github.com/bufbuild/buf/pull/4503. This will get set in the next release. Looked into exposing it as a variable but this would give us potentially different limits across the CLI and the BSR. Ideally plugins work the same wherever they are run. For your use case the AEP should work out of the box now. Does this approach work for you?

paultyng:
@emcfarlane sorry I missed that! thank you!

PR #4505, opened 2026-04-24
Reduce allocations in shake256.NewDigestForContent

emcfarlane:
Both `shake256.NewDigestForContent` and `bufimage.readObjectCloserToSourceFile` use `io.Copy` into an in-memory writer. With no `WriterTo`/`ReaderFrom` fast path available, `io.Copy` allocates its 32 KiB scratch buffer per call, much larger then many protobuf files.

Updated to pass an explicit 1 KiB buffer to `io.CopyBuffer`. This reduces bytes allocated.

Two new benchmarks (`BenchmarkBucketToDigest`, `BenchmarkBuildImage`) cover the digest and image-build paths over googleapis. Per pass:

- `BucketToDigest`: 78.9 MiB -> 15.4 MiB allocated (-80%)
- `BuildImage`:    578.6 MiB -> 544.4 MiB allocated (-6%)


pkwarren (approved):
Should we add a benchmark for this, or is it covered by other benchmarks? I'd be interested to see the effect of say building a manifest/module for the buf CLI's own proto/ sources.

emcfarlane:
Updated to remove the pools and use a small buffer. This is very similar in perf but this change is maybe not worth it. Its a minimal improvement only improving the `B/op` but not a real impact in cpu or allocations.
```
$ benchstat new.txt old.txt pool.txt
goos: darwin
goarch: arm64
pkg: github.com/bufbuild/buf/private/pkg/cas
cpu: Apple M2
                 │    new.txt    │             old.txt             │            pool.txt             │
                 │    sec/op     │    sec/op     vs base           │    sec/op     vs base           │
BucketToDigest-8   105.08m ± ∞ ¹   97.74m ± ∞ ¹  ~ (p=0.100 n=3) ²   99.71m ± ∞ ¹  ~ (p=0.100 n=3) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                 │    new.txt    │             old.txt              │             pool.txt             │
                 │     B/op      │     B/op       vs base           │     B/op       vs base           │
BucketToDigest-8   15.37Mi ± ∞ ¹   78.92Mi ± ∞ ¹  ~ (p=0.100 n=3) ²   12.91Mi ± ∞ ¹  ~ (p=0.100 n=3) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05

                 │    new.txt    │             old.txt              │            pool.txt             │
                 │   allocs/op   │   allocs/op    vs base           │  allocs/op    vs base           │
BucketToDigest-8   100.15k ± ∞ ¹   100.15k ± ∞ ¹  ~ (p=1.000 n=3) ²   95.99k ± ∞ ¹  ~ (p=0.100 n=3) ²
¹ need >= 6 samples for confidence interval at level 0.95
² need >= 4 samples to detect a difference at alpha level 0.05
```

private/bufpkg/bufimage/build_image_bench_test.go
```diff
@@ -0,0 +1,52 @@
...
+// BenchmarkBuildImage builds an image from the googleapis corpus (1574 proto
+// files), matching TestGoogleapis. Exercises the proto-source read path that
+// allocates an io.Copy fallback buffer per file.
+func BenchmarkBuildImage(b *testing.B) {
+	googleapisDirPath := buftesting.GetGoogleapisDirPath(b, buftestingDirPath)
+	moduleSet, err := bufmoduletesting.NewModuleSetForDirPath(googleapisDirPath)
+	if err != nil {
+		b.Fatal(err)
+	}
+	logger := slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: slog.LevelError}))
```
pkwarren: Should this use `slogtestext`?

private/bufpkg/bufimage/module_file_resolver.go
```diff
@@ -182,14 +182,18 @@ func (m *moduleFileResolver) addPath(
 // readObjectCloserToSourceFile is a helper function that takes a given [storage.ReadObjectCloser]
 // and returns the corresponding [*source.File].
 func readObjectCloserToSourceFile(readObjectCloser storage.ReadObjectCloser) (*source.File, error) {
-	var buf strings.Builder
-	if _, err := io.Copy(&buf, readObjectCloser); err != nil {
+	var sb strings.Builder
+	// Use a small fixed buffer instead of io.Copy's 32 KiB default. Proto
+	// files are typically small (median a few KiB), so 1 KiB reads most files
+	// in a few iterations while keeping per-call allocation minimal.
+	buf := make([]byte, 1024)
```
pkwarren, line 189: If the src supports WriterTo or the dst supports ReaderFrom, there is no allocation performed in `io.Copy`: https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/io/io.go;l=408-416

Do we want to do the same here?

doriable (approved):

private/bufpkg/bufimage/module_file_resolver.go
emcfarlane: The `io.WriterTo` on the os bucket has this unfortunate fallback: https://cs.opensource.google/go/go/+/refs/tags/go1.26.2:src/os/file.go;l=273 which reverts this optimization. So we could just implement WriterTo for the inmemory bucket?

emcfarlane: Merged as-is for now but would be great to get some improvements here. My plumbing of io.WriterTo through was not a clear win and rolleback some of these buffer improvements.

PR #4514, opened 2026-04-27
Clear stale LSP diagnostics when a file is closed or deleted

### LINKED ISSUE #626: Deleted file linting issues persist as problems in VSC after file deletion
### Buf Latest Version

- [x] I'm using the latest version of `buf`.

### steps

```shell
- create a file (i.e. joey.proto)
- VSC problems tab should show:
  - Files must have a package defined.
  - Files must have a syntax explicitly specified. If no syntax is specified, the file defaults to "proto2".
- delete joey.proto
```

### Output

Image

problems are still present despite file missing

Image

### Expected Output

I expect that if a proto file isn't present anymore, it no longer have its problems listed, even if those problems were legitimate at time of file existence

### Anything else?

buf --version output is 1.64.0

stefanvanburen:
When a file was closed or deleted, the LSP server never published empty diagnostics to clear the client's Problems panel.

Root cause: the workspace's `indexFiles` path calls `fileManager.Track` for every file in the module, so the ref count for an editor-open file is typically 2 (one from `DidOpen`, one from the workspace). When `DidClose` decremented the count to 1, `Reset` was never called and diagnostics were never cleared.

Fix by making `file.Close` call `clearEditorState` before decrementing the ref count. `clearEditorState` cancels in-flight lint checks, publishes empty diagnostics (only when non-empty, to skip the no-op network write), and sets `version=-1` so subsequent `RunChecks` won't re-publish lint diagnostics.

Also implement `DidDeleteFiles` (`workspace/didDeleteFiles`), previously a NYI stub, so files deleted from disk via the OS or IDE file explorer are handled the same way.

Also extract the `"buf lint"` diagnostic source string into a named constant `lintSourceName`.

Fixes https://github.com/bufbuild/vscode-buf/issues/626.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Increase check plugin WASM memory limits to 1GiB.
+- Fix LSP stale diagnostics persisting in the Problems panel after a file is closed or deleted.
 
 ## [v1.68.4] - 2026-04-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 - Increase check plugin WASM memory limits to 1GiB.
-- Fix LSP stale diagnostics persisting in the Problems panel after a file is closed or deleted.
+- Fix LSP stale diagnostics persisting after a file is closed or deleted.
 
 ## [v1.68.4] - 2026-04-22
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [Unreleased]
 
 - Increase check plugin WASM memory limits to 1GiB.
+- Fix LSP stale diagnostics persisting in the Problems panel after a file is closed or deleted.
```
stefanvanburen:
```suggestion
- Fix LSP stale diagnostics persisting after a file is closed or deleted.
```

emcfarlane (approved):

doriable (approved):

PR #4517, opened 2026-04-28
Propagate DuplicateProtoPathError as typed error from new compiler path

stefanvanburen:
When a proto file is contained in multiple modules (`DuplicateProtoPathError`), the new experimental compiler was silently swallowing the typed error: FDS did not check the fatal error from its Link sub-query, producing an empty FileDescriptorSet instead of propagating the error upward. This meant callers of `compileImage` could not inspect it via `errors.As`, and the downstream "nil FileDescriptor" panic was unhelpful.

The fix is in protocompile (https://github.com/bufbuild/protocompile/pull/722): `FDS.Execute` now checks `linkResult[0].Fatal` before iterating over IR files, propagating the Link fatal as its own. With this, `results[0].Fatal` in compileImage is DuplicateProtoPathError, which is returned directly so callers retain its type information and the existing "Failure: " / exit code 1 behavior is preserved.

### DIFF
--- a/cmd/buf/workspace_test.go
+++ b/cmd/buf/workspace_test.go
@@ -1269,9 +1269,9 @@ func TestWorkspaceDuplicateFailSpecificModule(t *testing.T) {
 	testRunStdoutStderrNoWarn(
 		t,
 		nil,
-		100,
+		1,
 		``,
-		`foo.proto:1:1:foo.proto is contained in multiple modules:
+		`Failure: foo.proto is contained in multiple modules:
   path: "other/proto"
   path: "proto"`,
 		"build",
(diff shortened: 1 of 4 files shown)


emcfarlane (approved):

cmd/buf/workspace_test.go
```diff
@@ -1269,9 +1269,9 @@ func TestWorkspaceDuplicateFailSpecificModule(t *testing.T) {
 	testRunStdoutStderrNoWarn(
 		t,
 		nil,
-		1,
+		100,
```
doriable: Urgh, sorry for being annoying about this, and this gives me a little bit of extra insight on how the compiler diagnostics have always been... so unfortunately, this would be considered a breaking change (exit code + stdout/stdwarn behaviour)...

So I think the approach we need to take here is:
- We continue to suppress the diagnostics with no primary span
- And then have `DuplicateProtoPathError` continue to be surfaced through the module build

We should document this in the code path where we do this pretty carefully, since this can be confusing to the reader of the code, but also, not introducing breaking behaviours is the priority.

Ideally, we would do an audit of the diagnostic surface area now that we have the new compiler in, but that may not be feasible right now -- in the short term, we need to prioritise retaining consistency in the high-level behaviour...

stefanvanburen:
@doriable I'll update the PR description if you agree, but PTAL at https://github.com/bufbuild/buf/pull/4517/commits/0808f448f4c57a52611356ad506f587c5b6597e0, which pulls in https://github.com/bufbuild/protocompile/pull/722 instead and handles the fatal errors first, which I think should be what we want to do...

doriable:
@stefanvanburen Ah yeah, that is much better! We should go with that approach, feel free to update and I'll stamp :)

PR #4531, opened 2026-05-07
Fix LSP documentSymbol marking every symbol as deprecated

### LINKED ISSUE #4530: LSP documentSymbol marks every symbol with SymbolTag.Deprecated, regardless of whether it is deprecated

https://github.com/garrettr/buf-lsp-symboltag-deprecated-repro

### Commands

```shell
python3 lsp_probe.py


(`lsp_probe.py` is a dependency-free Python LSP client that opens
`proto/example/v1/example.proto`, requests `textDocument/documentSymbol`,
and prints each returned symbol's `tags` field. It just shells out to
`buf lsp serve`.)
```

### Output

```shell
$ python3 lsp_probe.py
buf version: 1.67.0
symbols returned: 9
symbols with SymbolTag.Deprecated (tags=[1]): 9

  name=example.v1.Greeting                      tags=[1]
  name=example.v1.Greeting.text                 tags=[1]
  name=example.v1.Greeting.count                tags=[1]
  name=example.v1.GreeterService                tags=[1]
  name=example.v1.GreeterService.Greet          tags=[1]
  name=example.v1.GreetRequest                  tags=[1]
  name=example.v1.GreetRequest.name             tags=[1]
  name=example.v1.GreetResponse                 tags=[1]
  name=example.v1.GreetResponse.greeting        tags=[1]
```

### Expected Output

`example.proto` contains no `option deprecated = true;` or `[deprecated = true]`
anywhere, so no symbol should carry `SymbolTag.Deprecated`. Expected:

```shell
buf version: 1.67.0
symbols returned: 9
symbols with SymbolTag.Deprecated (tags=[1]): 0

  name=example.v1.Greeting                      tags=None
  name=example.v1.Greeting.text                 tags=None
  ... (etc.)
```

### Anything else?

The user-visible effect is that every symbol in every `.proto` file is rendered
with a strikethrough in editors that honor `SymbolTag.Deprecated` from
`textDocument/documentSymbol` — in VS Code that's the Outline panel, breadcrumbs,
sticky scroll, and Go to Symbol (`Cmd+Shift+O`). Editor body text is unaffected
because the semantic-token codepath flags deprecated tokens correctly.

**Cause.** Introduced in #4311 (`Add tags to LSP diagnostics`, merged
2026-02-03), first released in `v1.65.0`, still present on `main` as of this
writing. `private/buf/buflsp/symbol.go` populates `Tags` unconditionally:

```go
var isDeprecated bool
if _, ok := s.ir.Deprecated().AsBool(); ok {
    isDeprecated = true
}
return protocol.SymbolInformation{
    ...
    Deprecated:    isDeprecated,
    Tags: []protocol.SymbolTag{          // <-- always set
        protocol.SymbolTagDeprecated,
    },
}
```

Per the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolInformation),
clients prefer `tags` over the deprecated `deprecated` field, so the
unconditional tag wins. `Tags` should only include `SymbolTagDeprecated` when
`isDeprecated` is true, e.g.:

```go
var tags []protocol.SymbolTag
if isDeprecated {
    tags = []protocol.SymbolTag{protocol.SymbolTagDeprecated}
}
```

A second, smaller bug: the `if` discards the bool returned by `AsBool()` and
only checks the second `ok` return, so `isDeprecated` (and the legacy
`Deprecated` field) would also be wrong for an explicit
`option deprecated = false;`. It should be:

```go
isDeprecated, _ := s.ir.Deprecated().AsBool()
```

Affected: `v1.65.0`–`v1.69.0` (and `main`). Last unaffected release: `v1.64.0`.
Both `buf lsp serve` and `buf beta lsp` are affected (they share `symbol.go`).
Reproduced on Linux x86_64.

garrettr:
## What

`textDocument/documentSymbol` responses unconditionally set `Tags` to `[SymbolTagDeprecated]` for every symbol. Per the LSP spec, clients prefer `tags` over the legacy `deprecated` boolean, so the unconditional tag won and every symbol in every `.proto` file rendered with a strikethrough in editors that honor symbol tags — in VS Code that's the Outline panel, breadcrumbs, sticky scroll, and Go to Symbol (`Cmd+Shift+O`).

This change only includes `SymbolTagDeprecated` in `Tags` when the symbol is actually deprecated.

It also fixes the deprecation check itself: it previously discarded the value returned by `AsBool()` and only checked the `ok` return, so an explicit `option deprecated = false;` would also be reported as deprecated. Use the returned boolean value directly.

## Why

Introduced in #4311, first released in v1.65.0. Reproduction and full details in #4530.

Fixes #4530


doriable:
This fix looks good, thank you for catching and submitting this!
A quick comment: would you mind adding a regression test case for this in `diagnostic_test.go`? As an aside, I'll look into the CVE's, it seems we'll need to upgrade our go version to address that, I can handle that from our end.

(EDIT: a PR has been opened to upgrade the Go version #4532)

garrettr:
Thanks for the review! Added a regression test in `document_symbol_test.go` (the existing `textDocument/documentSymbol` test — there isn't a `diagnostic_test.go` and the bug is in the documentSymbol path rather than diagnostics, so I extended the test that was already exercising it).

The existing test already had a deprecated `LegacyDocument` case but only asserted on the legacy `Deprecated` boolean, not `Tags`, which is why it didn't catch this. The test now asserts `Tags` for every symbol, and I added an `ExplicitlyNotDeprecated` message with `option deprecated = false;` to pin down both the unset and explicitly-false cases.

Verified the new assertions fail against the pre-fix `symbol.go` ("symbol X has wrong tags" for every non-deprecated symbol) and pass with the fix.

garrettr:
I'm still waiting for my employer to approve the CLA. I will sign it (or close the PR without signing) when I hear back from them.

garrettr:
Just signed the CLA after getting permission from my company's legal department. @doriable Would you please take another look?

doriable (approved):
This looks good, thanks for pushing up the fix! I'll merge for you and this will be available on the next release.

PR #4540, opened 2026-05-12
Lint MessageRules.oneof in PROTOVALIDATE check

jhump:
Validate that each MessageOneofRule entry names at least one field, contains no duplicate names, references only fields defined in the message, and that no field appears in more than one entry.


doriable (approved):
This makes sense to me -- I know there's a lint exception that needs to be added for the `int32` conversion for field tags, approving in advance of that. Super tiny note: do you mind adding a changelog entry for this?

jhump:
> This makes sense to me -- I know there's a lint exception that needs to be added for the int32 conversion for field tags, approving in advance of that. Super tiny note: do you mind adding a changelog entry for this?

Sorry it took me so long to get back to this. Fixed the lint rule and also added the changelog entry.

PR #4541, opened 2026-05-14
Add buf beta registry cargo credential provider

kjvalencik:
## Summary

Adds a new `buf beta registry cargo` subcommand that implements Cargo's [`cargo:token-from-stdout` credential-provider protocol](https://doc.rust-lang.org/cargo/reference/registry-authentication.html#cargotoken-from-stdout). With this in place, Cargo can resolve a Bearer token for a BSR-hosted Cargo registry by shelling out to `buf`, instead of users having to copy a token into `~/.cargo/credentials.toml` and keep two copies in sync.

The command is registered under `beta` so we have room to iterate on the UX before promoting it.

## Motivation

Today, a developer pulling crates from a BSR-hosted Cargo registry has to:

1. Run `buf registry login buf.build` to authenticate `buf`.
2. Separately store the same token in `~/.cargo/credentials.toml` (or `cargo login`) so Cargo can use it.

Two stores for one secret, two places to rotate, two places to leak. This change lets users delegate the second step to `buf`:

```toml
# ~/.cargo/config.toml
[registry]
global-credential-providers = ["cargo:token-from-stdout buf beta registry cargo"]
```

Now any `cargo` operation against a BSR-hosted registry invokes `buf beta registry cargo`, which resolves the token from the same chain `buf` already uses (`BUF_TOKEN` env var, then `~/.netrc`).

## Configuration

Default — public BSR at `buf.build`:

```toml
[registry]
global-credential-providers = ["cargo:token-from-stdout buf beta registry cargo"]
```

Enterprise BSR (positional args replace the default; `buf.build` is **not** implicitly included):

```toml
[registry]
global-credential-providers = ["cargo:token-from-stdout buf beta registry cargo bsr.example.com"]
```

Multiple hosts:

```toml
[registry]
global-credential-providers = ["cargo:token-from-stdout buf beta registry cargo buf.build bsr.example.com"]
```

Token management continues to flow through `buf registry login <host>`; no Cargo-specific login is required.

## How it works

On each invocation Cargo sets `CARGO_REGISTRY_INDEX_URL` (and other env vars per the spec) and reads a single line from stdout. The command:

1. Extracts the host from `CARGO_REGISTRY_INDEX_URL` (strips an optional `sparse+` prefix, parses as a URL, lowercases, drops any port).
2. Checks the host against the positional-argument allow-list (default `[buf.build]`). Allow-list entries are also lowercased and port-stripped so `bsr.example.com:8443` matches `sparse+https://bsr.example.com:8443/...`.
3. Looks up a token via `bufconnect.NewTokenProviderFromContainer` (the `BUF_TOKEN` env var) and falls through to `bufconnect.NewNetrcTokenProvider` (`~/.netrc`).
4. On success, writes `Bearer <token>\n` to stdout and exits `0`.

No Connect clients are constructed and no network calls are made — credential providers are on the hot path of every dependency resolution.

## Failure semantics

| Situation | Behavior | Why |
|---|---|---|
| `CARGO_REGISTRY_INDEX_URL` unset, unparseable, or host not in the allow-list | Exit non-zero with empty stdout/stderr | Lets Cargo fall through to its next configured credential provider |
| Host in the allow-list but no token resolves | Exit non-zero; stderr says `Failure: no token found for <host>. Run "buf registry login <host>" ...` | The user opted in to this host but `buf` has no credentials for it — surface a fix |
| `BUF_TOKEN` is malformed | Exit non-zero; stderr says `Failure: the BUF_TOKEN environment variable could not be parsed. Either unset BUF_TOKEN, or run "buf registry login <host>" ...`. Underlying parser error is debug-only | The underlying `bufconnect` parse error embeds the raw token value, so the user-visible message is deliberately redacted |

Pass `--debug` to log host extraction, allow-list matching, and (when present) the raw `BUF_TOKEN` parser error.

## Out of scope

- No `cargo login` / `cargo logout` support — the `cargo:token-from-stdout` protocol explicitly doesn't support them.
- No changes to `bufconnect` or any existing authentication primitive; this is a thin orchestrator over what's already there.

## Test plan

- [x] `go test -race ./cmd/buf/internal/command/beta/registry/registrycargo/...` (15 unit + command-level tests covering the contract matrix: unset URL, default and explicit allow-lists, scoped vs unscoped `BUF_TOKEN`, scoped-miss → netrc fallthrough, env-wins-over-netrc, netrc match + miss, uppercase URL host, uppercase positional, positional `host:port`, malformed `BUF_TOKEN` with leak-prevention `NotContains` assertion).
- [x] `golangci-lint run ./cmd/buf/internal/command/beta/registry/registrycargo/...` — 0 issues.
- [x] `go build ./...` and `go vet ./...` clean.
- [x] Manual smoke tests against the built binary:
  - Silent fallback for a non-allow-listed `CARGO_REGISTRY_INDEX_URL`.
  - Success path returns `Bearer <token>` and exits 0.
  - Loud failure with full advice when no token resolves for an allow-listed host.
  - `--debug` surfaces structured logs and (only with `--debug`) the raw `BUF_TOKEN` parser error.
  - `BUF_TOKEN=secret@host,secret@host` → user-facing stderr does **not** contain `secret`; only the debug log does.

## Notes for review

- Single new package: `cmd/buf/internal/command/beta/registry/registrycargo`. The only touches outside it are a 2-line wiring change in `cmd/buf/buf.go` and a 1-line `CHANGELOG.md` entry.
- The command uses the empty-message `errSilent` sentinel for silent exits, following the same pattern as `private/pkg/bandeps/cmd/bandeps/main.go`. The top-level `wrapError` in `cmd/buf/buf.go` already bypasses its `Failure: ` prefix when `err.Error() == ""`.
- Placed under `beta` per the design discussion; promote to `buf registry cargo` once we've shipped to early users.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add LSP completion for `buf.gen.yaml`, `buf.yaml`, and `buf.policy.yaml` files.
 - Add error for when a dependency is added to `buf.yaml` and is missing from `buf.lock`.
+- Add `buf beta registry cargo` subcommand that acts as a Cargo `cargo:token-from-stdout` credential provider for BSR-hosted Cargo registries.
 
 ## [v1.69.0] - 2026-04-29
 
--- a/cmd/buf/buf.go
+++ b/cmd/buf/buf.go
@@ -36,6 +36,7 @@ import (
 	"github.com/bufbuild/buf/cmd/buf/internal/command/beta/price"
 	betaplugindelete "github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/plugin/plugindelete"
 	betapluginpush "github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/plugin/pluginpush"
+	"github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/registrycargo"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/webhook/webhookcreate"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/webhook/webhookdelete"
 	"github.com/bufbuild/buf/cmd/buf/internal/command/beta/registry/webhook/webhooklist"
@@ -415,6 +416,7 @@ func newRootCommand(name string) *appcmd.Command {
 						Use:   "registry",
 						Short: "Manage assets on the Buf Schema Registry",
 						SubCommands: []*appcmd.Command{
+							registrycargo.NewCommand("cargo", builder),
 							{
 								Use:   "webhook",
 								Short: "Manage webhooks for a repository on the Buf Schema Registry",
--- a/cmd/buf/internal/command/beta/registry/registrycargo/registrycargo.go
+++ b/cmd/buf/internal/command/beta/registry/registrycargo/registrycargo.go
@@ -1,0 +1,208 @@
+// Copyright 2020-2026 Buf Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package registrycargo implements the "buf beta registry cargo" command,
+// which acts as a Cargo "cargo:token-from-stdout" credential provider for
+// BSR-hosted Cargo registries.
+package registrycargo
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"net/url"
+	"slices"
+	"strings"
+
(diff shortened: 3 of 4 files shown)


doriable:
Thanks for submitting this PR, I've reviewed it and unfortunately, I don't think we'll be merging this feature right now. So my stance on this PR is the UX of the proposed command here is narrow/assumes that the use has run `buf registry login` ahead of time, and has Cargo-specific implementation details bleed into `buf`. My stance is also that auth through `.netrc` should be addressed through upstream solution.

I was wondering if you have explored using a credential provider that pulls from `.netrc`. An example of this would be using `cargo-credential-netrc` (https://crates.io/crates/cargo-credential-netrc) and a `credential-alias`, then setting something like the following in `.cargo/config.toml`:


```
[credential-alias]
cargo-credential-bsr = ["cargo-credential-netrc", "Bearer {{password}}"]

[registries.buf]
index = "sparse+https://buf.build/gen/cargo/"
credential-provider = "cargo-credential-bsr"
```

For now, I am closing this PR, but please feel free to follow-up with any questions and if there is anything else I can help with to get to a solution for you. Thanks again!

kjvalencik:
@doriable I think the key part is that it IS `buf` specific. Buf will check for credentials in places _other_ than `.netrc` (e.g., environment variables) which rules otu something generic like `cargo-credential-netrc`.

Would you consider something that dropped the `cargo` functionality and simply emitted the credential? FWIW, `npm` has a similar problem and could benefit from this.

PR #4542, opened 2026-05-14
Update the `PROTOVALIDATE` lint rule to check NaN

oliversun9:
Updates the `PROTOVALIDATE` rule to detect `NaN` in `const`, `lt`, `lte`, `gt`, `gte`, `in` and `not_in` in both floating point number types.

Doesn't check for infinities.


private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/numeric.go
```diff
@@ -55,6 +56,88 @@ func checkNumberRules[
...
+			}
+			if math.IsNaN(float64(v)) {
+				adder.addForPathf(
+					[]int32{ruleFieldNumber, fieldNumber},
+					"Field %q has %s set to NaN. Comparisons with NaN are always false, so this rule can never be satisfied.",
+					adder.fieldName(),
+					adder.getFieldRuleName(ruleFieldNumber, fieldNumber),
+				)
+			}
+		case "in":
```
jonbodner-buf: Since the bodies of "in" and "not_in" are the same except for part of the error message, I would merge the cases and build different strings depending on the value of field.Name()

jonbodner-buf (changes requested):

private/bufpkg/bufcheck/bufcheckserver/internal/buflintvalidate/numeric.go
oliversun9: Fixed in [2d611e3](https://github.com/bufbuild/buf/pull/4542/commits/2d611e311bd3c9f2012b13af691cd30aa822debb)

jonbodner-buf (approved):

jonbodner-buf (approved):

PR #4549, opened 2026-05-18
Filter out GIT_WORK_TREE from cloner environment variables

jasondamour:
## Summary

Follows the same pattern as #3760 (`GIT_DIR`) and #3813 (`GIT_INDEX_FILE`) to fix the same class of issue (#3752) for git worktrees.

When `buf breaking --against` is run inside a git worktree (e.g. via a pre-commit hook in a Cursor/VS Code worktree), the shell sets `GIT_WORK_TREE` pointing at the worktree's working directory. Passing this variable into the cloner's `git init` / `git fetch` / `git checkout` sequence causes git to reject the operation:

```
fatal: GIT_WORK_TREE (or --work-tree=<directory>) not allowed without specifying GIT_DIR (or --git-dir=<directory>).
```

The fix is a one-liner: add `"GIT_WORK_TREE": ""` to the existing env-sanitization map in `private/pkg/git/cloner.go`, so the cloner always starts with a clean git environment regardless of how buf was invoked.

## Test plan

- [x] Run `buf breaking --against` from inside a git worktree (e.g. `.cursor/worktrees/...`) and confirm no `GIT_WORK_TREE` error
- [x] Existing unit/integration tests pass (`go test ./private/pkg/git/...`)


doriable:
This generally looks good -- any chance you could merge `main` to pull in the test cases from #4550 and implement a similar test case for `GIT_WORK_TREE`?

jasondamour:
@doriable done!

doriable (approved):

doriable:
Looks good, this is ready to merge, thank you for your contribution!

PR #4550, opened 2026-05-19
Add env override test cases to git package

doriable:
This PR adds test cases for ensuring that the env variable
overrides in the `git` package are respected.


doriable:
This adds some tests for the env overrides, once these merge, #4549 should add a test case for `GIT_WORK_TREE`.

emcfarlane (approved):

PR #4551, opened 2026-05-19
Add merge_base option to buf breaking --against

### LINKED ISSUE #4495: Add a merge-base option to the buf-breaking check
### Feature

If I have some changes on a branch that has diverged from the main branch, to check for breaking changes, I probably want to compare against not the main branch itself, but the merge-base of the main branch and my branch. So it would be nice if the `--against` flag accepted `.git#merge-base=main` or something like that.

jasondamour:
Closes #4495

## Summary

Adds a `merge_base=<ref>` option to the `--against` flag of `buf breaking`, so users can write:

```
buf breaking --against ".git#merge_base=main"
```

instead of the shell workaround:

```
buf breaking --against ".git#ref=$(git merge-base HEAD main)"
```

This is especially useful in pre-commit hook tools (e.g. [prek](https://prek.j178.dev/)) that don't expose the merge-base commit as an environment variable.

## How it works

When `merge_base=<ref>` is specified:

1. The option is parsed through the existing `buffetch` ref-parsing pipeline and stored on the `GitRef` interface as `GitMergeBase() string`.
2. At clone time in `getGitBucket()`, buf runs `git merge-base HEAD <ref>` against the local repository to resolve the common ancestor commit hash.
3. That commit hash is used as the checkout target (equivalent to `ref=<hash>`), so the comparison is against the exact point where the current branch diverged from `<ref>`.

`merge_base` is **only supported for local git references** (e.g. `.git#merge_base=main`). For remote repos, the existing shell workaround `ref=$(git merge-base HEAD origin/main)` remains the approach.

## Constraints

- Cannot be combined with `branch`, `commit`, `tag`, or `ref` (returns a clear error)
- Defaults to `depth=50` (same as `ref=`)
- Only valid for local git schemes

## Changes

| File | Change |
|------|--------|
| `private/pkg/git/git.go` | Add `GetMergeBase()` helper |
| `private/buf/buffetch/internal/errors.go` | Add 2 error constructors |
| `private/buf/buffetch/internal/internal.go` | Add `GitMergeBase` to `RawRef` and `GitRef` interface |
| `private/buf/buffetch/internal/git_ref.go` | Add `gitMergeBase` field and accessor |
| `private/buf/buffetch/internal/ref_parser.go` | Parse, validate, and thread `merge_base` through |
| `private/buf/buffetch/internal/reader.go` | Resolve merge-base commit before cloning |
| `private/buf/buffetch/internal/ref_parser_test.go` | Unit tests for parsing and validation |

## Test plan

- [x] Unit tests added in `ref_parser_test.go` covering valid parsing and all invalid combinations
- [x] `buf breaking . --against ".git#merge_base=main"` produces identical output to `buf breaking . --against ".git#ref=$(git merge-base HEAD main)"`
- [x] All existing `buffetch` and `git` package tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/private/buf/buffetch/internal/errors.go
+++ b/private/buf/buffetch/internal/errors.go
@@ -48,6 +48,16 @@ func NewCannotSpecifyCommitOrTagWithRefError() error {
 	return errors.New(`cannot specify "commit" or "tag" with "ref"`)
 }
 
+// NewCannotSpecifyMergeBaseWithOtherGitOptionsError is a fetch error.
+func NewCannotSpecifyMergeBaseWithOtherGitOptionsError() error {
+	return errors.New(`"merge_base" cannot be specified with "branch", "commit", "tag", or "ref"`)
+}
+
+// NewMergeBaseOnlyForLocalGitError is a fetch error.
+func NewMergeBaseOnlyForLocalGitError() error {
+	return errors.New(`"merge_base" is only supported for local git references (e.g. ".git#merge_base=main")`)
+}
+
 // NewDepthParseError is a fetch error.
 func NewDepthParseError(s string) error {
 	return fmt.Errorf(`could not parse "depth" value %q`, s)
--- a/private/buf/buffetch/internal/git_ref.go
+++ b/private/buf/buffetch/internal/git_ref.go
@@ -43,6 +43,7 @@ type gitRef struct {
 	recurseSubmodules bool
 	subDirPath        string
 	filter            string
+	gitMergeBase      string
 }
 
 func newGitRef(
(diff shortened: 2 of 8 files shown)


doriable:
thank you for submitting this PR! unfortunately, i am reticent to accept an extension to the git input for this functionality right now — the UX of a `merge_base` option goes against the composable nature of the other options (and is constrained to local git references), and now presents side effects as a part of the input. imo, this should be handled by downstream tooling. 

i understand that some tooling, such as the pre-commit hooks use-case cited in the original issue, do not currently provide an easy way to handle using a merge-commit as the base, i can respond to the issue with a workaround for specifically that case, but because of the UX of this option, we cannot accept this.

PR #4571, opened 2026-06-11
Support recursive glob (.**) in bufimageutil type filters

jhump:
A trailing ".**" on an include or exclude type name now matches the named element and every symbol nested beneath it: a package and all its sub-packages and types, or a message and all its nested types.

(Note that exclusion was already recursive if the root type was a message that contain nested types: we can't realistically keep the nested types if we remove their container.)

This addresses a long-standing TODO since the question of whether including a type should include its children/descendants definitely came up when this functionality was first implemented.

### DIFF
--- a/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
+++ b/private/bufpkg/bufimage/bufimageutil/bufimageutil.go
@@ -107,6 +107,11 @@ func WithAllowIncludeOfImportedType() ImageFilterOption {
 // For example, "google.protobuf.Any" or "buf.validate". Types may be nested,
 // and can be any package, message, enum, extension, service or method name.
 //
+// A name ending in ".**" is a recursive glob that matches the named element and
+// every symbol nested beneath it. For example, "acme.foo.**" matches the package
+// "acme.foo", all of its sub-packages, and all of their types; "acme.Foo.**"
+// matches the message "acme.Foo" and all of its nested types.
+//
 // If the type does not exist in the image, an error wrapping
 // [ErrImageFilterTypeNotFound] will be returned.
 func WithIncludeTypes(typeNames ...string) ImageFilterOption {
@@ -127,7 +132,14 @@ func WithIncludeTypes(typeNames ...string) ImageFilterOption {
 // For example, "google.protobuf.Any" or "buf.validate". Types may be nested,
 // and can be any package, message, enum, extension, service or method name.
 //
-// If the  type does not exist in the image, an error wrapping
+// A name ending in ".**" is a recursive glob that matches the named element and
+// every symbol nested beneath it. For a package, this additionally excludes all
+// of its sub-packages; without the glob, only the named package's own types are
+// excluded. Note that, even without the glob suffix, excluding a message always
+// excludes any nested types, too. This is because we can't realistically preserve
+// the nested types if we've excluded their container.
+//
+// If the type does not exist in the image, an error wrapping
 // [ErrImageFilterTypeNotFound] will be returned.
 func WithExcludeTypes(typeNames ...string) ImageFilterOption {
 	return func(opts *imageFilterOptions) {
@@ -345,6 +357,7 @@ func (t *transitiveClosure) hasOption(
 
 func (t *transitiveClosure) includeType(
 	typeName protoreflect.FullName,
+	recursive bool,
 	imageIndex *imageIndex,
 	options *imageFilterOptions,
 ) error {
@@ -372,6 +385,12 @@ func (t *transitiveClosure) includeType(
 		if err := t.addElement(descriptorInfo.element, "", false, imageIndex, options); err != nil {
 			return fmt.Errorf("inclusion of type %q: %w", typeName, err)
 		}
+		if recursive {
+			// Auto-include any symbol nested beneath the named element.
+			if err := t.includeChildElements(descriptorInfo.element, imageIndex, options); err != nil {
+				return fmt.Errorf("inclusion of type %q: %w", typeName, err)
+			}
+		}
 		return nil
 	}
 	// It could be a package name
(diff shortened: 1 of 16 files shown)


mfridman:
Wdyt about adding a test where the globbed package has no files directly in it?

Something like

- acme/order/v1/order.proto (package acme.order.v1)
- acme/order/v2/order.proto   (package acme.order.v2)

And `WithIncludeTypes("acme.order.**")`

With versioned packages this is the most common shape and nobody puts files directly in acme.order, so the glob target has zero files of its own and everything lives in sub-packages. afaics all packages in testdata have their own files, so this path is never exercised (which is the one we care about).

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -381,9 +400,14 @@ func (t *transitiveClosure) includeType(
 		return fmt.Errorf("inclusion of type %q: %w", typeName, ErrImageFilterTypeNotFound)
 	}
 	if !options.allowImportedTypes {
-		// if package includes only imported files, then reject
+		// If the package contains only imported files, then reject. For a
```
mfridman: Do we need/want improved error messages to cover this new functionality? E.g.,

Say we have something like

```sh
$ ls proto/
acme/order/v1/order.proto    # package acme.order.v1
```

And the user runs

```sh
$ buf build --type acme.order
Failure: inclusion of type "acme.order": type declared in imported module
```

Would a guard before the `if onlyImported` be useful, and we could surface something like:

```
Failure: package "acme.order" contains no types; did you mean "acme.order.**"?
```

Otherwise that original error is a bit misleading.

private/bufpkg/bufimage/bufimageutil/image_filter.go
```diff
@@ -35,18 +35,19 @@ func filterImage(image bufimage.Image, options *imageFilterOptions) (bufimage.Im
...
-		if err := closure.excludeType(excludeType, imageIndex, options); err != nil {
+		excludeType, recursive := strings.CutSuffix(excludeType, ".**")
+		if err := closure.excludeType(protoreflect.FullName(excludeType), recursive, imageIndex, options); err != nil {
 			return nil, err
 		}
 	}
 	for includeType := range options.includeTypes {
-		includeType := protoreflect.FullName(includeType)
-		if err := closure.includeType(includeType, imageIndex, options); err != nil {
+		includeType, recursive := strings.CutSuffix(includeType, ".**")
```
mfridman: Should we be a bit more defensive here to prevent things like `"foo.**.bar"` or `".**"` and surface a more meaningful error?

Right now `"foo.**.bar"` falls through to a literal lookup and produces a confusing not found, and `".**"` cuts to an empty name that matches the root package entry, so it silently includes (or excludes) the entire image.The latter is prob unlikely, but the former I could see tripping someone up.

```go
includeType, recursive := strings.CutSuffix(includeType, ".**")
if !protoreflect.FullName(includeType).IsValid() {
    return nil, fmt.Errorf("invalid type name %q", includeType)
}
```

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -393,18 +417,91 @@ func (t *transitiveClosure) includeType(
...
+		if err := t.includePackage(subPackage, recursive, false, imageIndex, options); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// includeChildElements recursively adds the symbols nested beneath the given
+// element to the closure. Only messages have such children: their nested
+// messages and enums. Other element kinds have nothing to expand.
```
emcfarlane: I think extensions are missing here. A services methods and a enums values are always included.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
```diff
@@ -393,18 +417,91 @@ func (t *transitiveClosure) includeType(
...
+	options *imageFilterOptions,
+) error {
 	for _, file := range pkg.files {
 		fileDescriptor := file.FileDescriptorProto()
 		if mode := t.elements[fileDescriptor]; mode == inclusionModeExcluded {
-			return fmt.Errorf("inclusion of excluded package %q", typeName)
+			if !isTopLevel {
+				continue
+			}
+			return fmt.Errorf("inclusion of excluded package %q", pkg.fullName)
```
mfridman: Do we want a sentinel error to match on, similar to ErrImageFilterTypeNotFound, but something like ErrImageFilterTypeExcluded.

That way the caller of this package (assume an API) can surface this as an invalid argument error for contradictory filters. There's probably other places too of this kind of cateogry.

emcfarlane (commented):
Maybe glob of `a.**` where `a.b` is excluded and `a` is empty should error. And we should maybe provide a better error on something like `foo.**.bar` where a user tries to use an unsupported glob syntax. Do we need to doc this change in the `--type strings` flags?

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
jhump: Yeah, good catch. I thought about this when creating this branch and somehow convinced myself it wasn't necessary. But you're right that it is in fact needed to be correct.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
jhump: Done

private/bufpkg/bufimage/bufimageutil/image_filter.go
jhump: Done. Also added a hint about valid wildcards if the invalid name contains an asterisk.

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
jhump: Done

private/bufpkg/bufimage/bufimageutil/bufimageutil.go
jhump: Done

jhump:
>Wdyt about adding a test where the globbed package has no files directly in it?
>
>Something like
>
>* acme/order/v1/order.proto (package acme.order.v1)
>* acme/order/v2/order.proto (package acme.order.v2)
>
>And `WithIncludeTypes("acme.order.**")`

@mfridman, done.

> Maybe glob of `a.**` where `a.b` is excluded and a is empty should error.

@emcfarlane, done. If an include pattern adds zero types to the set (due to exclusions), it will be reported as an error.

> And we should maybe provide a better error on something like `foo.**.bar` where a user tries to use an unsupported glob syntax.

Yep, @mfridman also pointed this out [here](https://github.com/bufbuild/buf/pull/4571#discussion_r3399026488). Done.

> Do we need to doc this change in the `--type strings` flags?

Done.

emcfarlane (approved):

PR #4580, opened 2026-06-15
Validate UUID assert not nil

emcfarlane:
This ensures no `uuid.UUID` taken as `00000000-0000-0000-0000-000000000000` is marked valid. Parsing a UUID allows for `uuid.Nil` but should not be when validating.

### DIFF
--- a/private/pkg/uuidutil/uuidutil.go
+++ b/private/pkg/uuidutil/uuidutil.go
@@ -58,19 +58,25 @@ func FromDashless(dashless string) (uuid.UUID, error) {
 // Validate determines if the given UUID string is valid and not [uuid.Nil].
 func Validate(s string) error {
 	id, err := FromString(s)
+	if err != nil {
+		return err
+	}
 	if id == uuid.Nil {
 		return fmt.Errorf("must not be nil: %s", id)
 	}
-	return err
+	return nil
 }
 
 // ValidateDashless validates the dashless uuid is valid and not [uuid.Nil].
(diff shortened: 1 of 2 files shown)


private/pkg/uuidutil/uuidutil.go
```diff
@@ -55,15 +55,21 @@ func FromDashless(dashless string) (uuid.UUID, error) {
 	return FromString(dashless[0:8] + "-" + dashless[8:12] + "-" + dashless[12:16] + "-" + dashless[16:20] + "-" + dashless[20:])
 }
 
-// Validate determines if the given UUID string is valid.
+// Validate determines if the given UUID string is valid and not [uuid.Nil].
 func Validate(s string) error {
-	_, err := FromString(s)
+	id, err := FromString(s)
+	if id == uuid.Nil {
```
DMarby, line 64: Do we not need to check `err` here first?

emcfarlane: oof yes

doriable (approved):
Looks reasonable as long as all the callsites agree on this semantic (I checked `bufbuild/buf`, and this isn't something we use locally here).

DMarby (approved):

emcfarlane:
I went through use cases. There is some behaviour that checks for the inverse of this, `if err == nil`, to assert it must not be a uuid. This will need handling on update to keep the same strictness. Othercases this always increased the validation restriction. Which is what this PR implements.

PR #4592, opened 2026-06-22
Fix webhook create/list output to use protojson

nikohofmann:
Fix `buf beta registry webhook create` and `buf beta registry webhook list` printing empty objects (`{}` / `[{}]`) instead of actual webhook data.

Root cause: both commands used `encoding/json` to marshal proto-generated structs. Proto structs store field data in unexported fields, invisible to `encoding/json`, so all fields silently serialize as zero values. Fix is to replace `encoding/json.MarshalIndent` with `protojson.MarshalOptions{Multiline: true, Indent: "\t"}.Marshal`, which uses the protobuf reflection API and correctly reads proto field values.

Note: the `create` output shape changes from a bare `{...}` to `{"webhook": {...}}` to match the RPC response structure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
+- Fix `buf beta registry webhook create` and `buf beta registry webhook list` to emit proto JSON output.
 
 ## [v1.71.0] - 2026-06-16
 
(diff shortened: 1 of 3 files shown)


doriable (approved):
Looks good to me, thank you for putting this up! We also chatted about testing out this fix etc.

Stamping ahead of time, but do you mind adding a changelog statement for this, since this fixes a user-facing issue with webhooks? Thank you!

PR #4615, opened 2026-07-13
Fix `buf generate` clean to not delete nested output

doriable:
This PR fixes an issue where if a plugin writes to
the child directory of another plugin's out directory,
the file will end up being cleaned, since it is not part
of the latter plugin's output.

Instead of tracking per-directory by individual buckets
for each plugin's generated output, we track the full paths
of all generated files, and then clean based on that. A test
case has also been added for this.

### DIFF
--- a/private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
+++ b/private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
@@ -144,24 +144,32 @@ func (w *responseWriter) Close(ctx context.Context) error {
 			return err
 		}
 	}
-	// Collect the set of generated paths per directory output before
-	// clearing state, so the delete phase knows which files to keep.
-	var dirOutputPaths map[string]map[string]struct{}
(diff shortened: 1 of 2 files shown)


private/bufpkg/bufprotoplugin/bufprotopluginos/response_writer.go
```diff
@@ -431,27 +435,6 @@ func (w *responseWriter) deleteStaleFilesAndEmptyDirs(
 	return removeEmptyDirs(outDirPath)
 }
 
-// nestedOutputDirs returns the set of relative directory paths for output
-// directories in dirOutputPaths that are nested under parentDir. Both
-// parentDir and the map keys are OS-native absolute paths; the returned
-// paths are normalized (forward slashes) for use with storage bucket paths.
-func nestedOutputDirs(parentDir string, dirOutputPaths map[string]map[string]struct{}) map[string]struct{} {
```
timostamm, line 438: Nice.

timostamm (approved):

PR #4625, opened 2026-07-22
Upgrade protocompile

emcfarlane:
This upgrades protocompile to bring in the fix for `buf format`. Adds a CHANGELOG entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## [Unreleased]
 
-- No changes yet.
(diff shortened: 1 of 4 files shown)


emcfarlane:
@doriable this golden looks like it was incorrect before the new format changes: https://github.com/bufbuild/buf/blob/af0c352877828472f4dbd98678846d1ff7b61585/private/buf/bufformat/testdata/proto3/all/v1/all.golden#L94

bufdev (approved):

PR #4636, opened 2026-08-04
Fix symbols re-exported via `import public` not being found

### LINKED ISSUE #4633: Since v1.68.0, a symbol that is re-exported via `import public` is not found by importers. v1.67.0 and protoc accept the same files.

https://github.com/smnbbrv/buf-public-import-repro

### Commands

```shell
npx -y -p @bufbuild/buf@1.72.0 buf build .   # broken: errors
npx -y -p @bufbuild/buf@1.67.0 buf build .   # good: accepts the same files
```

### Output

```shell
proto/a.proto:5:3:cannot find `google.protobuf.Timestamp` in this scope
```

### Expected Output

None

### Anything else?

_No response_

bufdev:
Fixes #4633.

Since v1.68.0, which switched `buf build` to the new compiler, a symbol re-exported through `import public` is not always found by importers. `protoc` and v1.67.0 accept the same files.

From the reporter's repro:

```protobuf
// a.proto
import "b.proto";
message A { google.protobuf.Timestamp ts = 1; }

// b.proto
import public "act.proto";                      // imports timestamp.proto non-publicly
import public "c.proto";                        // imports timestamp.proto publicly

// c.proto
import public "google/protobuf/timestamp.proto";
```

```
proto/a.proto:5:3:cannot find `google.protobuf.Timestamp` in this scope
```

Nothing to do with well-known types; swapping the order of `b.proto`'s two imports makes it compile.

## Root cause

In protocompile, `imports.Recurse` classified each transitive import by looking only at whichever direct import happened to pull it in first, so a file reachable through several direct imports with differing public-ness could be misclassified. Public direct imports are ordered first specifically so the public path wins, but that does not disambiguate when *both* direct imports are public: `act.proto` is visited first, reaches `timestamp.proto` non-publicly, and `timestamp.proto` lands in the trailing non-public segment, so `b.proto` stops re-exporting it.

A second, related shape was also broken: a plain `import` of a file that is *also* re-exported by a public import. That file has to stay in a direct segment, and `Transitive()` derived public-ness purely from the segment offsets, so it could not represent the case at all.

Fixed in bufbuild/protocompile#754, which classifies every transitive import up front over all direct imports and records public-ness on the import rather than deriving it from segment offsets. This matches `protoc`'s `DescriptorBuilder::RecordPublicDependencies`, which is a closure over `public_dependency` edges and therefore order-independent.

## Testing

`TestComparePublicImports` builds `testdata/publicimports` and diffs the resulting `FileDescriptorSet` against real `protoc`. It covers both shapes independently (each has its own consumer file, so one working mechanism cannot mask the other breaking), and both fail before the protocompile fix:

```
testdata/publicimports/consumer_diamond.proto:11:3:cannot find `Leaf` in this scope
testdata/publicimports/consumer_shadowed.proto:10:3:cannot find `Leaf` in this scope
```

Beyond that, the fix was validated by differential fuzzing against `protoc`: 700 random import DAGs of up to 12 files, each file referencing every symbol `protoc`'s rule says is visible (zero rejections), plus 838 pairwise accept/reject probes agreeing with `protoc` in both directions. The same fuzzer flags the bug on an unpatched build within ~30 seeds.

## Note

`go.mod` currently pins protocompile to the branch commit for bufbuild/protocompile#754. Needs a re-pin to `main` once that merges.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## [Unreleased]
 
 - Fix `buf format` dropping comments next to commas or semicolons in message literals.
+- Fix compilation failing to resolve symbols re-exported through `import public` when the
+  re-exporting file also reaches those symbols through a non-public import.
 
 ## [v1.72.0] - 2026-07-17
 
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	connectrpc.com/connect v1.20.0
 	connectrpc.com/grpcreflect v1.3.0
 	connectrpc.com/otelconnect v0.9.0
-	github.com/bufbuild/protocompile v0.14.2-0.20260723175035-caf54d45c839
+	github.com/bufbuild/protocompile v0.14.2-0.20260804015511-0bf9b3f60f22
 	github.com/bufbuild/protoplugin v0.0.0-20260414125817-25d1d281b46b
 	github.com/cli/browser v1.3.0
 	github.com/gofrs/flock v0.13.0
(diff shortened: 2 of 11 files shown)


emcfarlane (approved):

doriable (approved):
This is fine to merge as-is -- I can update `protocompile` to follow `main` once the upstream has merged as well.

PR #4645, opened 2026-08-18
Fix LSP panic on imports within external files

pkwarren:
Currently, performing a hover or go to definition on an import within a file outside the workspace (like a WKT) panics. Fix path resolution to fall back to using the LSP opener if the file isn't found in the workspace. Update a few places to guard against a nil file.

### DIFF
--- a/private/buf/buflsp/file.go
+++ b/private/buf/buflsp/file.go
@@ -878,9 +878,21 @@ func (f *file) importToSymbol(imp ir.Import) *symbol {
 		file: f,
 		span: imp.Decl.ImportPath().Span(),
 		kind: &imported{
-			file: f.workspace.PathToFile()[imp.File.Path()],
+			file: f.importedFile(imp),
 		},
 	}
+}
+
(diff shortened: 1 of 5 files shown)


doriable (approved):

private/buf/buflsp/file.go
```diff
@@ -878,11 +878,23 @@ func (f *file) importToSymbol(imp ir.Import) *symbol {
...
 }
 
+func (f *file) importedFile(imp ir.Import) *file {
+	path := imp.File.Path()
+	if imported, ok := f.workspace.PathToFile()[path]; ok {
+		return imported
+	}
+	sourceFile, ok := f.lsp.opener.Get()[path]
+	if !ok {
+		return nil
```
emcfarlane, line 893: Can this still panic on `nil` here? What triggered the panic?

emcfarlane: Ah missed you added more nil checks.

emcfarlane (approved):

PR #4655, opened 2026-08-25
Bump protoc version to v35.1

### LINKED ISSUE #4595: Bump built-in version of WKT in buf

https://github.com/Mr3zee/buf-wkt-behind-repro

### Commands

```shell
buf build
```

### Output

```shell
proto/feature_support_removal_error.proto:42:7:cannot resolve message field name for `google.protobuf.FieldOptions.FeatureSupport`
```

### Expected Output

No failures

### Anything else?

Originally found to be failing with https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/unittest_features.proto and https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/unittest_custom_features.proto

This probably means that buf's built-in WKT types are too behind current v35.1

### LINKED ISSUE #4650: buf format v1.71+ regressions: non-idempotent inline block comments and stripped interior indentation
### Description

`buf format` v1.71.0 introduced two regressions with the "Port new formatter" change:

**Bug 1: Non-idempotent formatting of trailing line comments**

Trailing `//` comments on option fields are converted to `/* */` block comments on the first pass (kept inline), then moved to a separate line on the second pass. The output never stabilizes in a single pass.

**Repro (`example.proto`):**
```protobuf
syntax = "proto3";

message Foo {
  bool bar = 1; // keep this inline
}
```

**First `buf format` pass:**
```protobuf
  bool bar = 1; /* keep this inline */
```

**Second `buf format` pass:**
```protobuf
  bool bar = 1;
  /* keep this inline */
```

Running `buf format` twice produces different output — formatting is not idempotent.

**Bug 2: Interior indentation alignment stripped from block comments**

The formatter strips intentional interior indentation, destroying semantic alignment. Previously fixed in PR #1615 (v1.10.0) but reintroduced by the v1.71.0 new formatter port.

**Repro:**
```protobuf
syntax = "proto3";

service ExampleService {
  /**
     Reverse lookup a user from an identifier.

     - OK - response contains the resolved user id
     - NOT_FOUND - identifier was not found
     - INVALID_ARGUMENT - invalid identifier type
                        - missing identifier
                        - provided identifier is already USER_ID
  */
  rpc GetUserId(GetUserIdRequest) returns (GetUserIdResponse);
}

message GetUserIdRequest {}
message GetUserIdResponse {}
```

**`buf format` output (v1.72.0):**
```protobuf
  /**
   Reverse lookup a user from an identifier.

   - OK - response contains the resolved user id
   - NOT_FOUND - identifier was not found
   - INVALID_ARGUMENT - invalid identifier type
   - missing identifier
   - provided identifier is already USER_ID
  */
```

The aligned sub-items under `INVALID_ARGUMENT` are flattened to the same indentation level as top-level list items, destroying the semantic grouping. "missing identifier" and "provided identifier is already USER_ID" are sub-cases of `INVALID_ARGUMENT`, not independent items.

### Expected behavior

1. Formatting should be idempotent — running `buf format` twice should produce the same output.
2. Interior indentation within block comments should be preserved (relative indentation maintained, as it was in v1.70.0 and earlier).

### Versions affected

Both bugs affect v1.71.0+ (tested on v1.72.0). v1.70.0 does not have either issue.

emcfarlane:
This pulls in the latest protocompile version and bumps to the related protoc version.

Closes https://github.com/bufbuild/buf/issues/4595, closes https://github.com/bufbuild/buf/issues/4650


make/go/dep_protoc.mk
```diff
@@ -10,10 +10,10 @@ $(call _assert_var,CACHE_INCLUDE)
 $(call _assert_var,CACHE_BIN)
 
 # Settable
-# https://github.com/protocolbuffers/protobuf/releases 20260109 checked 20260427
+# https://github.com/protocolbuffers/protobuf/releases 20260611 checked 20260825
 # NOTE: Set to version compatible with genproto source code (only used in tests).
 # This version must be supported by the matching https://github.com/protocolbuffers/protobuf-go version.
-PROTOC_VERSION ?= 33.3
+PROTOC_VERSION ?= 35.1
```
pkwarren, line 16: Should make this change upstream in https://github.com/bufbuild/makego.

pkwarren (approved):

PR #4656, opened 2026-08-26
Add PyPI wheel publishing flow

anuraaga:
This brings in the publishing workflow for https://pypi.org/project/buf-bin/. I considered building the wheels as part of release preparation, but it would mean pypi has wheels before the GitHub release itself is published, abandoning the release for whatever reason would become awkward. So this continues to download and package binaries after a release is published. The workflow and its logs will be tracable from pypi.

This is ready to review the workflow but before merging it I will make sure the GitHub repo environment and pypi publishing settings are set up.


.github/workflows/publish-pypi.yaml
```diff
@@ -0,0 +1,62 @@
...
+  build:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          persist-credentials: false
+
+      - uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
```
anuraaga, line 19: I followed the repo style of floating tags as much as possible, but uv only publishes pins, no floating tag, so this line will probably result in an increase in dependabot PRs

anuraaga:
@doriable The pypi and github environment settings should be in place so I think this is good to go now

packaging/python/out/bin/.gitkeep
doriable: Is there a reason why we need to keep the `out/bin` directory and why this wouldn't be ephemeral after the publish?

anuraaga: `uv sync` which we use to download the tools we use for generating the release binary fails if pyproject references an empty directory. There are other ways around this such as `UV_NO_PROJECT`, `uvx`, but I feel it's worth the reliability of working with the normal checkout without worrying about env.

So I added a doc into this file to comment on this since indeed that wasn't obvious.

doriable (approved):

PR #4664, opened 2026-09-01
Stop setting java_multiple_files for edition 2024 sources

pkwarren:
Setting `java_multiple_files = true` on an edition 2024 or later proto file causes generation to fail. Update managed mode to skip setting this.

### DIFF
--- a/private/bufpkg/bufimage/bufimagemodify/file_option.go
+++ b/private/bufpkg/bufimage/bufimagemodify/file_option.go
@@ -433,6 +433,11 @@ func modifyJavaMultipleFiles(
 	config bufconfig.GenerateManagedConfig,
 	options ...ModifyOption,
 ) error {
+	// The option was removed in edition 2024, where its behavior became the
+	// default. The latest protoc plugins reject any file that still sets it.
+	if imageFile.FileDescriptorProto().GetEdition() >= descriptorpb.Edition_EDITION_2024 {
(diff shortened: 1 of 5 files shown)


private/bufpkg/bufimage/bufimagemodify/file_option.go
```diff
@@ -433,6 +433,11 @@ func modifyJavaMultipleFiles(
 	config bufconfig.GenerateManagedConfig,
 	options ...ModifyOption,
 ) error {
+	// The option was removed in edition 2024, where its behavior became the
+	// default. The latest protoc plugins reject any file that still sets it.
```
pkwarren, line 437: Error from `buf generate`:

```
> buf generate
Failure: invalid_argument: plugin "buf.build/protocolbuffers/java:v36.1" exited with non-zero status 1
stderr:
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1788296414.665177       1 descriptor.cc:1493] foo/v1/foo.proto foo/v1/foo.proto: google.protobuf.FileOptions.java_multiple_files has been removed in edition 2024: This behavior is enabled by default in editions 2024 and above. To disable it, you can set `features.(pb.java).nest_in_file_class = YES` on individual messages, enums, or services.
```

emcfarlane (approved):

PR #4671, opened 2026-09-09
Update error message for canceled root context

oliversun9:
This updates the error message when the user presses ctrl-C.

Before
```
buf registry organization info buf.build/bufbuild
^CFailure: canceled: context canceled
```

After

```
buf registry organization info buf.build/bufbuild
^CFailure: interrupted
```


doriable (approved):

doriable:
This is a more indicative message of what is happening, so I'm on board with this.

PR #4680, opened 2026-09-15
Automatically attempt HTTP/2 prior knowledge without flag when required

### LINKED ISSUE #4629: Avoid having to pass `--http2-prior-knowledge` for every `http` URL
### Feature

Most of the time I'm using `buf curl` it is against services I am actively developing/testing, locally on my machine. Those services are exposing their servers over HTTP, not HTTPS. It is annoying that I have to pass `--http2-prior-knowledge` all the time! I'd love to be able to do `buf curl http://localhost:8888/my.service/Endpoint`. There are several different approaches that could solve this: 

1. Providing an env var that lets me set `--http2-prior-knowledge`, such that I don't need to pass it when writing out the command myself
2. Provide a configuration file that does the same thing
3. Automatically trying HTTP2, without me having to pass the flag

I'm unsure if there's a way of detecting HTTP2 prior to making the actual request. But intuitively, I'd expect there to be some sort of way! You could also argue that it'd even make sense to automatically enable HTTP if reflection is active, as HTTP2 is a prerequisite for reflection to even work.

anuraaga:
Currently, `--http2-prior-knowledge` must always be set with http urls where HTTP/2 is required. The reality is it's generally well known if a request will require HTTP/2 or not, so this change detects those cases and implicitly enables the flag, for gRPC protocol, server reflection enabled, or bidi streaming.

After this change, users should generally not need to set `--http2-prior-knowledge` anymore.

Fixes #4629


cmd/buf/internal/command/curl/curl.go
```diff
@@ -152,7 +156,7 @@ Examples:
 Issue a unary RPC to a plain-text (i.e. "h2c") gRPC server, where the schema for the service is
 in a Buf module in the current directory, using an empty request message:
 
-    $ buf curl --schema . --protocol grpc --http2-prior-knowledge \
+    $ buf curl --schema . --protocol grpc \
```
anuraaga, line 159: Clearest demonstration of the win

cmd/buf/internal/command/curl/curl.go
```diff
@@ -878,6 +886,31 @@ func parseEndpointURL(urlArg string) (service, method, baseURL string, err error
...
 }
 
+// wrapPlainTextHTTP2Error is a best effort to return a more helpful error if err indicates
+// that the server answered an HTTP/2 prior knowledge (h2c) connection with an
+// HTTP 1.1 response, which means it does not support HTTP/2 over plain-text.
+// Otherwise, err is returned unchanged.
+//
+// Without this, the CLI's error interceptor would render this without details as:
+//
+//	Failure: the server hosted at that remote is unavailable.
```
anuraaga, line 896: While this function is a bit hacky, it felt worth it to improve on this error message

cmd/buf/internal/command/curl/curl.go
```diff
@@ -878,6 +886,31 @@ func parseEndpointURL(urlArg string) (service, method, baseURL string, err error
...
+func wrapPlainTextHTTP2Error(err error, f *flags, host string, isSecure bool) error {
+	if err == nil || isSecure || !f.HTTP2PriorKnowledge ||
+		// The stdlib does not expose a structured way of knowing the error is from a
+		// HTTP/1.1-like response so do a string match, meaning this function is
+		// best-effort across Go versions.
+		!strings.Contains(err.Error(), "frame header looked like an HTTP/1.1 header") {
+		return err
+	}
+	// Format with %v rather than %w on purpose: the CLI's error interceptor
```
anuraaga, line 905: This seemed reasonable to me, the alternative was going to be much more invasive

cmd/buf/internal/command/curl/curl.go
```diff
@@ -878,6 +886,31 @@ func parseEndpointURL(urlArg string) (service, method, baseURL string, err error
...
 
+// wrapPlainTextHTTP2Error is a best effort to return a more helpful error if err indicates
+// that the server answered an HTTP/2 prior knowledge (h2c) connection with an
+// HTTP 1.1 response, which means it does not support HTTP/2 over plain-text.
+// Otherwise, err is returned unchanged.
+//
+// Without this, the CLI's error interceptor would render this without details as:
+//
+//	Failure: the server hosted at that remote is unavailable.
+func wrapPlainTextHTTP2Error(err error, f *flags, host string, isSecure bool) error {
```
jonbodner-buf: should this function taken in `f` or a `bool` with the value of  `f.HTTP2PriorKnowledge`?

cmd/buf/internal/command/curl/curl.go
```diff
@@ -878,6 +886,31 @@ func parseEndpointURL(urlArg string) (service, method, baseURL string, err error
...
+	if err == nil || isSecure || !f.HTTP2PriorKnowledge ||
+		// The stdlib does not expose a structured way of knowing the error is from a
+		// HTTP/1.1-like response so do a string match, meaning this function is
+		// best-effort across Go versions.
+		!strings.Contains(err.Error(), "frame header looked like an HTTP/1.1 header") {
```
jonbodner-buf, line 902: I wonder if using a structured error for this case would be a good contribution to the go standard library/(or really, golang.org/x/net). There are two places in the standard library that would need to be changed and one new exported error type as a sentinel.

jonbodner-buf (approved):

cmd/buf/internal/command/curl/curl.go
anuraaga: ~~It already does doesn't it (read on 898)?~~

Oops got it makes sense

PR #4681, opened 2026-09-15
Add support for formatting from stdin

### LINKED ISSUE #1035: Format accepts input from stdin

anuraaga:
This adds the `--stdin-filepath` flag which when set, indicates a filepath to use to identify input passed via stdin instead. One use case is simple editor integration, though I suspect since the issue was filed LSP integration on both sides may have improved enough that this is less of an issue. Another still important one is integration with meta formatters, the Gradle spotless plugin in the linked issue invokes other formatters and unfortunately needs to prepare a temp file only for protos, and rumdl also rejected the tempfile approach for formatting protos within markdown files. With this change, we can eventually have formatting of proto in markdown files which is pretty great for docs pages.

There are various names for a similar flag in many formatters, `--filename` etc. I went with this one only because it matches [Zed's docs](https://zed.dev/docs/reference/all-settings) which feels like a neutral third party.

Fixes #1035

/cc @stefanvanburen


cmd/buf/internal/command/format/format.go
```diff
@@ -399,6 +436,138 @@ func run(
...
+	}
+	externalPath := flags.StdinFilePath
+	path := normalpath.Base(normalpath.Normalize(externalPath))
+	if normalpath.Ext(path) != ".proto" {
+		return appcmd.NewInvalidArgumentErrorf(
+			"--%s must be a path to a .proto file",
+			stdinFilePathFlagName,
+		)
+	}
+	data, err := io.ReadAll(container.Stdin())
```
jonbodner-buf, line 456: Should there be some sort of limit on this? https://pkg.go.dev/io#LimitReader

jonbodner-buf: If I understand this correctly, the contents of `data` are being written to `externalPath`. Rather than buffer the complete stdin contents in memory via `data`, you could just use `io.Copy` to read from stdin and write to the file.

cmd/buf/internal/command/format/format.go
```diff
@@ -399,6 +436,138 @@ func run(
...
+	}
+	originalReadBucket, err := newStdinReadBucket(ctx, path, externalPath, data)
+	if err != nil {
+		return err
+	}
+	formattedReadBucket, err := bufformat.FormatBucket(ctx, originalReadBucket)
+	if err != nil {
+		return err
+	}
+	diffBuffer := bytes.NewBuffer(nil)
```
jonbodner-buf, line 468: When creating an initially empty buffer, usually `var diffBuffer bytes.Buffer` is used, and then you use `&diffBuffer`.

jonbodner-buf (commented):
one question and one suggestion.

cmd/buf/internal/command/format/format.go
emcfarlane: io.ReadAll matches the current file behavior. Don't think theres a reasonable limit to apply.

cmd/buf/internal/command/format/format.go
emcfarlane: `diffBuffer := bytes.NewBuffer(nil)` matches the current file.

emcfarlane (approved):

PR #4683, opened 2026-09-16
Add --locked flag to verify lockfile is in sync

### LINKED ISSUE #4003: buf does not warn about out of sync issue with buf.yaml and buf.lock

https://github.com/josejulio/buf-dep-reproducer?tab=readme-ov-file

### Commands

```shell
buf build
```

### Output

```shell
# No output
```

### Expected Output

```
WARNING: buf.yaml dependency `buf.build/bufbuild/protovalidate:v0.14.1` is out of sync with buf.lock file - using `buf.build/bufbuild/protovalidate:v0.7.1` instead. Run `buf dep update` to update the dependencies
```

### Anything else?

I was using an old version of the dependency for a while without really knowing it, until some issues appeared months later and we confirmed we were using an old version of protovalidate.

It would be nice if ~~`buf generate` and~~ `buf build` would check if the lock file is still on sync with `buf.yaml`.

It would also be useful to have a command to check if the files are in sync without having to run `buf build` ~~or `buf generate`~~ (for ci purposes).

anuraaga:
Most if not all package managers have a way to verify a lock file is in sync. For example, `uv` provides `--locked` flag to all its operations which does it, and this PR follows the same model. It is quite common to add this in CI to make sure a dep update PR is consistent, for example we even do it in [protobuf-py](https://github.com/bufbuild/protobuf-py/blob/main/.github/workflows/ci.yaml#L52) and I think a lot of CI workflows using buf will take advantage of this. Without such a check, buf.yaml can be updated without a lock and users will basically never realize the update didn't take effect until a problem (buf uses the versions in the lockfile, not yaml file), as we see in the attached issue.

`--locked` is added to all commands that reference dependencies, including `build` and `generate`, and when set, will check that the versions in buf.yaml match the hashes in the lock. It will not check dependencies in buf.yaml with no version set.

This does not add a separate command that only checks the lockfile without running something, in uv this is `uv lock --checked`. One idea was adding a `check` subcommand, but it results in three `buf dep check`, `buf plugin check`, `buf policy check`, which is tedious for users. But then it seems like a new top-level command `buf lock check` is needed, and I think it's a bigger change than needed for now.

Fixes #4003


emcfarlane (commented):
The `--locked` requires network access and I think this behavior will break with labels like `v1` which can resolve to any commit in the version history. This would break in CI. What about storing the ref in the lock file and simply comparing it to the resolved ref from the yaml file? This can be done without network access and without a flag.

anuraaga:
Thanks for the idea @emcfarlane! Yeah that does seem like a good idea - there's no point in recording ref without checking it so I will incorporate it into this change.

anuraaga:
I have things working well with the lock file approach, but it depends on #4685 I think to prevent version lock issues. I will go ahead and close this for now to wait for if/after that gets out in the wild for some time

PR #4684, opened 2026-09-16
Resolve ignore comments for group fields

### LINKED ISSUE #4187: Comment ignores are not being picked up by `group` fields

https://github.com/bufbuild/buf

### Commands

```shell
buf {lint, breaking}

message GroupFoo {
  // buf:lint:ignore FIELD_NOT_REQUIRED
  required group Metadata = 1 { /* fields in group */ }
}
```

### Output

```shell
proto/test/foo.proto:8:12:field Foo.Metadata.id: field has no label; proto2 requires explicit 'optional' label
```

### Expected Output

Expect the comment ignore to be respected.

### Anything else?

I did some investigation and this has to do with the synthetic type produced for `group` fields. The leading comment is being attributed to the synthetic message declaration rather than the field declaration.

My suggested fix for the current implementation would be to resolve the descriptor from the given source path. This is tricky because we would need to parse the source path and effectively walk the descriptor tree, determine whether the source path is a field, check if it's a `group` field, and then finally resolve the associated synthetic message.

For now, there is a workaround, to apply the comment ignore to the containing message of the `group` field. This isn't great, since the comment ignore would apply to all fields in the containing message.

anuraaga:
Currently ignore comments are missed when a group is defined inline on a field. This PR takes the approach exactly outlined, to resolve a source path in the descriptor, check if it's a group, and check it's message type for comments instead of only the field. Not ideal, but it's only needed on errors and should still generally be a relatively short traversal.

Fixes #4187


private/bufpkg/bufcheck/client.go
```diff
@@ -824,18 +825,125 @@ func ignoreFileLocation(
...
+	for i := 0; i < len(sourcePath)-2; i += 2 {
+		tag, index := sourcePath[i], int(sourcePath[i+1])
+		var messageDescriptors protoreflect.MessageDescriptors
+		switch {
+		case parentMessageDescriptor == nil && tag == fileMessagesTag:
+			messageDescriptors = fileDescriptor.Messages()
+		case parentMessageDescriptor != nil && tag == messageNestedMessagesTag:
+			messageDescriptors = parentMessageDescriptor.Messages()
+		default:
+			return nil
```
anuraaga: While I think source paths are supposed to be valid here, I kept things lenient here since it seemed better that way

private/bufpkg/bufcheck/client.go
```diff
@@ -824,18 +825,125 @@ func ignoreFileLocation(
...
 		for _, associatedSourcePath := range associatedSourcePaths {
 			sourceLocation := sourceLocations.ByPath(associatedSourcePath)
-			if leadingComments := sourceLocation.LeadingComments; leadingComments != "" {
-				for _, line := range xstrings.SplitTrimLinesNoEmpty(leadingComments) {
-					if checkCommentLineForCheckIgnore(line, config.CommentIgnorePrefix, ruleID) {
-						return true, nil
-					}
+			if leadingCommentsHaveCheckIgnore(sourceLocation.LeadingComments, config.CommentIgnorePrefix, ruleID) {
+				return true, nil
+			}
```
emcfarlane, line 830: This. `leadingCommentsHaveCheckIgnore` doesn't look like it changed the inline code logic? And the `if` check seems minor anyway as `for` will have zero iterations.

private/bufpkg/bufcheck/client.go
```diff
@@ -824,18 +825,125 @@ func ignoreFileLocation(
...
+		if index < parentMessageDescriptor.Fields().Len() {
+			return parentMessageDescriptor.Fields().Get(index)
+		}
+	case messageExtensionsTag:
+		if index < parentMessageDescriptor.Extensions().Len() {
+			return parentMessageDescriptor.Extensions().Get(index)
+		}
+	}
+	return nil
+}
```
emcfarlane: Played with a generic traversal idea.

```suggestion
func groupFieldSyntheticMessage(
	fileDescriptor protoreflect.FileDescriptor,
	sourcePath protoreflect.SourcePath,
) protoreflect.MessageDescriptor {
	fieldDescriptor, ok := descriptorForSourcePath(fileDescriptor, sourcePath).(protoreflect.FieldDescriptor)
	if !ok || fieldDescriptor.Kind() != protoreflect.GroupKind {
		return nil
	}
	return fieldDescriptor.Message()
}

// Source path tags for the declarations traversed when resolving a source path.
const (
	// FileDescriptorProto.message_type.
	fileMessagesTag = int32(4)
	// FileDescriptorProto.extension.
	fileExtensionsTag = int32(7)
	// DescriptorProto.field.
	messageFieldsTag = int32(2)
	// DescriptorProto.nested_type.
	messageNestedMessagesTag = int32(3)
	// DescriptorProto.extension.
	messageExtensionsTag = int32(6)
)

// descriptorList is the shape shared by protoreflect's descriptor list types, such as
// protoreflect.MessageDescriptors and protoreflect.FieldDescriptors.
type descriptorList[D protoreflect.Descriptor] interface {
	Len() int
	Get(i int) D
}

// descriptorAtIndex returns the descriptor at the given index, or nil if the index is out
// of range.
func descriptorAtIndex[D protoreflect.Descriptor, L descriptorList[D]](
	descriptors L,
	index int,
) protoreflect.Descriptor {
	if index < 0 || index >= descriptors.Len() {
		return nil
	}
	return descriptors.Get(index)
}

// descriptorForSourcePath returns the declaration at the given source path, or nil if the
// source path does not point to a declaration.
//
// A source path alternates a tag and an index, descending through the declarations of a
// file, for example [4, 0, 3, 1, 2, 0] for .message_type(0).nested_type(1).field(0). Only
// the tags needed to reach a field or extension declaration are resolved.
func descriptorForSourcePath(
	fileDescriptor protoreflect.FileDescriptor,
	sourcePath protoreflect.SourcePath,
) protoreflect.Descriptor {
	if len(sourcePath) == 0 || len(sourcePath)%2 != 0 {
		return nil
	}
	descriptor := protoreflect.Descriptor(fileDescriptor)
	for ; len(sourcePath) > 0; sourcePath = sourcePath[2:] {
		tag, index := sourcePath[0], int(sourcePath[1])
		switch typedDescriptor := descriptor.(type) {
		case protoreflect.FileDescriptor:
			switch tag {
			case fileMessagesTag:
				descriptor = descriptorAtIndex(typedDescriptor.Messages(), index)
			case fileExtensionsTag:
				descriptor = descriptorAtIndex(typedDescriptor.Extensions(), index)
			default:
				return nil
			}
		case protoreflect.MessageDescriptor:
			switch tag {
			case messageNestedMessagesTag:
				descriptor = descriptorAtIndex(typedDescriptor.Messages(), index)
			case messageFieldsTag:
				descriptor = descriptorAtIndex(typedDescriptor.Fields(), index)
			case messageExtensionsTag:
				descriptor = descriptorAtIndex(typedDescriptor.Extensions(), index)
			default:
				return nil
			}
		default:
			return nil
		}
		if descriptor == nil {
			return nil
		}
	}
	return descriptor
}
```

emcfarlane (approved):
lgtm! Small suggestions but feel free to resolve.

private/bufpkg/bufcheck/client.go
doriable: This seems like a reasonable approach.

private/bufpkg/bufcheck/client.go
```diff
@@ -824,18 +825,125 @@ func ignoreFileLocation(
...
+	messageExtensionsTag = int32(6)
+)
+
+// fieldDescriptorForSourcePath returns the field declaration at the given source path, or
+// nil if the source path does not point to a field declaration.
+//
+// A source path for a field declaration alternates a tag and an index, descending through
+// message declarations before terminating at a field or an extension field, for example
+// [4, 0, 3, 1, 2, 0] for .message_type(0).nested_type(1).field(0).
+func fieldDescriptorForSourcePath(
```
doriable: Would it be helpful to export a utility from `protosourcepath` for source path parsing/traversal, since those mechanisms already exist? I think that might help simplify the logic required for maintenance at this layer.

private/bufpkg/bufcheck/client.go
anuraaga: Yup it didn't change - this is to reuse it for the second check after resolving a group field. It seemed like enough logic to warrant the helper, but I removed the extra if

private/bufpkg/bufcheck/client.go
anuraaga: Thanks - I have extracted a helper, in doing so made it a bit more general (previously we only traversed for a group field, not any type of descriptor

PR #4685, opened 2026-09-16
Parse lock file leniently

anuraaga:
When looking at https://github.com/bufbuild/buf/pull/4683#issuecomment-5694621749 I noticed buf.lock is parsed strictly, so an older buf would fail to run any command with a newer lock file format. Maybe buf.lock should be managed with a .proto :-)

But anyways as there was no test verifying failure with unknown fields, I feel it's probably unintentional, and being able to additively update the lockfile seems useful. I left v1 unchanged since it seems best to keep that behavior frozen. What do you think?


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 - Add `--stdin-filepath` flag to `buf format`, which reads a single `.proto` file from
   stdin and writes the formatted result to stdout. The path is not read from disk, and is
   only used to report parse errors and diffs.
+- Parse v2 `buf.lock` files non-strictly, ignoring unknown fields. A `buf.lock` written by a
+  newer `buf` that records additional fields no longer fails to parse on an older `buf`.
```
emcfarlane, line 12: Maybe doens't need a changelog as not user facing.

emcfarlane (approved):
Makes sense to me, but introducing any field is still going to be a breaking change upto this versions release.

doriable:
So I think this was intentional behaviour, tbh, to push folks towards updating to later CLI versions when the lock file updates. I'm not sure we want to change that behaviour right now...

anuraaga:
Thanks for the notes - I think either way either we stick to encouraging CLI update, or we should do the opposite and consider the leniency change breaking and potentially a v3. I think this PR doesn't help with either so will avoid it for now.

PR #4686, opened 2026-09-16
Deduplicate remote fetches for generate inputs

### LINKED ISSUE #4657: Deduplicate identical remote fetches across `inputs:` entries in buf.gen.yaml
### Feature

When a single `buf.gen.yaml` (v2) declares multiple `inputs:` entries that resolve to the same remote (`tarball`, `zip_archive`, `binary_image`, etc...), the URL is re-fetched over and over.

It would be nice if this could be deduplicated to be a little more efficient and also avoid spamming the servers.

### Usecase
In one of our mono-repos this is a non-issue because our services are all namespace well and we have a single `buf.yaml` representing everything as one module, but in others each service is treated as an independent module with a `buf.yaml` in the root listing each by path. When we package up an artifact containing all our `.protos` as a `.tar.gz` , or build into a `.binpb` we have to effectively reuse the same URL for each module we include in our `buf.gen.yaml` and this results in a bunch of unnecessary rapid hits to our hosts.

We also looked at fetching direct from github too (which would be really nice to avoid publishing artifacts at all), but the same refetching occurs for `git_repo` and limits us there too. Arguably its worse for that one because our repos are huge and re-pulling over and over is painfully slow.

### Resources
While this isn't a bug, I did decide to go ahead and put together a repo to showcase the behavior incase it helps:
https://github.com/is-jonreeves/ghissues-buf-dedupe-remote-inputs

emcfarlane:
This adds caching to remote references so buf generate now fetches each remote input once. Multiple inputs in a buf.gen.yaml that resolve to the same archive, git repository, or image previously fetched once per input. The cache is opt-in and set per command required.

Closes https://github.com/bufbuild/buf/issues/4657


doriable (commented):
Since this would be using an in-memory cache, we need to do some validation for the LSP on memory consumption and include that in some combination of the PR description or in a test.

emcfarlane:
@doriable thanks added option config for setting per command.

doriable (approved):
Since we're gating this to only be used, this looks reasonable to me. Approving in advance of merge conflicts being resolved.

PR #4687, opened 2026-09-16
Improve buf curl error message on trailing input

### LINKED ISSUE #3944: Provide specific error message for multiple-request messages in `buf curl`
### Feature

I was using the `buf curl` CLI and did not notice a trailing `}` in a JSON request body. The below error message which is given from the command did not fully identify the issue at hand, as the target was a Server-Streaming RPC. The message was initially confusing as it it insinuated an incorrect schema whereas any single-request RPC may give this message, not just pure-unary RPCs.

https://github.com/bufbuild/buf/blob/93b4aebf8944a58850ce4a4347992b6856355f29/private/buf/bufcurl/invoker.go#L130

As a suggestion, a more accurate message might be:
```go
return fmt.Errorf("method %s requires a single request message, but input contained more than one request message (check for trailing content)", inv.md.Name()) 
```

emcfarlane:
Small improvement to the error message for `buf curl` on trailing data misinterpreted as a multi-stream message.

Closes https://github.com/bufbuild/buf/issues/3944


private/buf/bufcurl/invoker_test.go
```diff
@@ -86,3 +71,48 @@ func TestCountUnrecognized(t *testing.T) {
 	unrecognized := countUnrecognized(msg)
 	assert.Equal(t, expectedUnrecognized, unrecognized)
 }
+
+func TestVerifySingleRequest(t *testing.T) {
+	t.Parallel()
+	resolver := newTestResolver(t)
+	// Download is server-streaming: it still accepts only a single request, and
```
anuraaga: Perhaps it's in a different file since I don't see it in this one, otherwise should we have a test case for normal unary too? Similarly maybe cases for second message succeeding to parse unless they're elsewhere

anuraaga (approved):

PR #4688, opened 2026-09-17
Reject config files with !-prefixed local YAML tags

### LINKED ISSUE #3973: `buf` silently consumes v1 configuration errors

https://github.com/engnatha/buf_config_validation_repro

### Commands

```shell
`bazel test //src/...`
```

### Output

```shell
# When using the existing configuration
$ bazel test //src/...
INFO: Analyzed 6 targets (0 packages loaded, 0 targets configured).
INFO: Found 3 targets and 3 test targets...
INFO: Elapsed time: 0.195s, Critical Path: 0.11s
INFO: 4 processes: 6 linux-sandbox.
INFO: Build completed successfully, 4 total actions
//src/demo_one:bar_test                                                  PASSED in 0.1s
//src/demo_one:foo_test                                                  PASSED in 0.1s
//src/demo_two:foo_test                                                  PASSED in 0.1s

Executed 3 out of 3 tests: 3 tests pass.
```

### Expected Output

An error saying that the config specified was invalid or identifying the linting errors in `src/demo_two:foo_test`

### Anything else?

This is largely mitigated by migrating to v2. That wasn't feasible for our setup at the time, but it's something we'll prioritize going forward. This left us vulnerable to linting on the v1 config for a number of months because all our tests were silently getting ignored.

A couple potential solutions to this bug would be
* Improving the config validation for v1
* Supporting glob syntax rather than simple directories/files and handling the yaml tagging syntax intentionally

anuraaga:
An unquoted string starting with `!` is a YAML type tag to force the following value to a specific type registered by the decoder (as opposed to a built-in type that starts with `!!`). We don't register any such types so there is no valid usage for this. On the flip side, there are places where a string starting with `!` is appropriate such as globs, but when forgetting quotes, it gets parsed as a type tag, which effectively silently replaces the value with an empty string which we then use as is. For globs, this means ignoring `!` unexclude globs completely, things that shouldn't be excluded are. This actually affects both v1 and v2 to different degrees.

The approach is to parse the YAML once to a YAML node to first to be able to distinguish type tags from values to find and reject them, before parsing the YAML into the config struct. Unfortunately, there seems to be no way to do it in a single pass since a config struct doesn't preserve type info and no decoder, even in other libraries, have an option to reject local tags. Parse only happens once per operation and is still microseconds for very large configs, I think the problematic case (silent corruption) is significant enough that we need to eat it.

Fixes #3973


emcfarlane (approved):
Neat.

doriable (approved):

PR #4689, opened 2026-09-17
Split imports by directory when generating with include_imports

### LINKED ISSUE #3056: include_imports does not respect 'strategy: directory'

anuraaga:
This splits imports by directory during generation in the same way as non-imports when `include_imports` is set with `strategy: directory`. Everything else effectively picks up the change, notably computing the generator requests will first populate `nonImportPaths` with the imports that have been added to the images which prevents them from being promoted for generation in arbitrary images that import them instead.

Fixes #3056


doriable (approved):
The overall approach and test makes sense to me, approving ahead of merge conflicts.

